PR #1146, opened 2019-04-06
Refactor collecting RestrictInfo

svenklemm:
Building RestrictInfo for plan time chunk exclusion only collected
expressions in WHERE clause but not constraints present in ON clause
of JOINs. This patch adds support for collecting constraints from
JOIN conditions.

This patch also changes the expression_tree_mutator to an
expression_tree_walker because the modifications we do do not require
using a mutator.

### DIFF
--- a/src/plan_expand_hypertable.c
+++ b/src/plan_expand_hypertable.c
@@ -62,15 +62,11 @@ init_chunk_exclusion_func()
 }
 
 static bool
-is_chunk_exclusion_func(Node *node)
+is_chunk_exclusion_func(Expr *node)
 {
-	if (IsA(node, FuncExpr))
-	{
-		FuncExpr *explicit_exclusion = (FuncExpr *) node;
+	if (IsA(node, FuncExpr) && castNode(FuncExpr, node)->funcid == chunk_exclusion_func)
+		return true;
 
-		if (explicit_exclusion->funcid == chunk_exclusion_func)
-			return true;
-	}
 	return false;
 }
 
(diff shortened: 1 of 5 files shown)


src/plan_expand_hypertable.c
```diff
@@ -80,100 +80,95 @@ is_chunk_exclusion_func(Node *node)
...
+			if (!IsA(linitial(func_expr->args), Var))
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg("first parameter for chunks_in function needs to be a record")));
+
+			ctx->chunk_exclusion_func = func_expr;
+			quals = (Node *) list_delete_cell((List *) quals, lc, prev);
+			lc = prev;
+		}
+		else
```
cevian: should this be `else if(ctx->chunk_exclusion_func == NULL)`

src/plan_expand_hypertable.c
```diff
@@ -80,100 +80,95 @@ is_chunk_exclusion_func(Node *node)
...
+	if (IsA(node, FromExpr))
+	{
+		FromExpr *f = castNode(FromExpr, node);
+		f->quals = process_quals(f->quals, ctx);
+	}
+	else if (IsA(node, JoinExpr))
+	{
+		JoinExpr *j = castNode(JoinExpr, node);
+		j->quals = process_quals(j->quals, ctx);
 	}
```
cevian: (optional) add `if (ctx->chunk_exclusion_func != NULL) return node;` to short-cut for efficiency?

svenklemm: we still need to process all quals to detect duplicate chunks_in calls

src/plan_expand_hypertable.c
svenklemm: I've changed this because this leads to inconsistent results. Now it will always build restrictinfo for all quals that qualify, previously it would stop building restrictinfo depending on the position of the chunks_in call in the quals

src/plan_expand_hypertable.c
```diff
@@ -80,100 +76,86 @@ is_chunk_exclusion_func(Node *node)
...
+
+			/* validation */
+			Assert(func_expr->args->length == 2);
+			if (!IsA(linitial(func_expr->args), Var))
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg("first parameter for chunks_in function needs to be record")));
+
+			ctx->chunk_exclusion_func = func_expr;
+			quals = (Node *) list_delete_cell((List *) quals, lc, prev);
```
cevian: suggestion: set ctx->restrictions = Nil here?

PR #1147, opened 2019-04-07
Mark ddl_hook test as DEBUG test

svenklemm:


cevian:
Maybe mention that it cannot be run in non-debug mode in the commit body?

PR #1153, opened 2019-04-09
Add time_bucket support to chunk exclusion

svenklemm:
This patch adds support for chunk exclusion for time_bucket
expressions in the WHERE clause. The following transformation
is done when building RestrictInfo:

Transform time_bucket calls of the following form in WHERE clause:

time_bucket(width, column) OP value

Since time_bucket always returns the lower bound of the bucket
for lower bound comparisons the width is not relevant and the
following transformation can be applied:

time_bucket(width, column) > value
column > value

Example with values:

time_bucket(10, column) > 109
column > 109

For upper bound comparisons width needs to be taken into account
and we need to extend the upper bound by width to capture all
possible values.

time_bucket(width, column) < value
column < value + width

Example with values:

time_bucket(10, column) < 100
column < 100 + 10

This allows chunk exclusions to work for views with aggregations.

### DIFF
--- a/src/plan_expand_hypertable.c
+++ b/src/plan_expand_hypertable.c
@@ -6,6 +6,7 @@
 #include <postgres.h>
 #include <nodes/relation.h>
 #include <parser/parsetree.h>
+#include <optimizer/clauses.h>
 #include <optimizer/var.h>
 #include <optimizer/restrictinfo.h>
 #include <nodes/plannodes.h>
@@ -15,6 +16,7 @@
 
 #include <catalog/pg_constraint.h>
 #include <catalog/pg_inherits.h>
+#include <catalog/pg_namespace.h>
 #include "compat.h"
 #if PG96 || PG10 /* PG11 consolidates pg_foo_fn.h -> pg_foo.h */
 #include <catalog/pg_constraint_fn.h>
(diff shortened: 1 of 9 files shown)


src/plan_expand_hypertable.c
```diff
@@ -70,6 +72,124 @@ is_chunk_exclusion_func(Expr *node)
...
+
+/*
+ * Transform time_bucket calls of the following form in WHERE clause:
+ *
+ * time_bucket(width, column) > value
+ * column > value - width
+ *
+ * time_bucket(width, column) < value
+ * column < value + width
+ *
```
JLockerman: writing down notes on why this works:
`time_bucket` rounds down, so `column - width < time_bucket(width, column) <= column`
for the first equation, with any non-negative `width`
```
column >= time_bucket(width, column) > value
column > value
column > value - width
```
for the second
```
column - width < time_bucket(width, column)  < value
column - width < value
column < value + width
```
so we might be more conservative than we need to be, but it should work.

JLockerman: I think we need more of the justification in this comment.

src/plan_expand_hypertable.c
```diff
@@ -70,6 +72,124 @@ is_chunk_exclusion_func(Expr *node)
...
+ * This type of WHERE clause is used when querying continuous aggregation
+ * views.
+ */
+static OpExpr *
```
JLockerman: would you add a note saying `this function only works with the two argument time_bucket, it's the callers responsibility to ensure that it is only called with such` or add checks of such into the function?

src/plan_expand_hypertable.c
```diff
@@ -70,6 +72,124 @@ is_chunk_exclusion_func(Expr *node)
...
+ * Transform time_bucket calls of the following form in WHERE clause:
+ *
+ * time_bucket(width, column) > value
+ * column > value - width
+ *
+ * time_bucket(width, column) < value
+ * column < value + width
+ *
+ * This type of WHERE clause is used when querying continuous aggregation
+ * views.
```
cevian: As an outside reviewer, I found Josh's derivation above much easier to verify. Please include it (or something like it) in the comment.

cevian: Especially the >= vs > is easier to follow

src/plan_expand_hypertable.c
```diff
@@ -70,6 +72,124 @@ is_chunk_exclusion_func(Expr *node)
...
+		{
+			/*
+			 * TIMESTAMPTZ OP INTERVAL is marked stable and unsafe
+			 * to evaluate at plan time unless it only has a time
+			 * component
+			 */
+			Interval *interval = DatumGetIntervalP(width->constvalue);
+
+			if (interval->day == 0 && interval->month == 0)
+				subst = (Expr *) estimate_expression_value(root, (Node *) subst);
```
cevian: don't we need to return op in an else? won't this just chug merrily along otherwise?

src/plan_expand_hypertable.c
```diff
@@ -70,6 +72,124 @@ is_chunk_exclusion_func(Expr *node)
...
+			if (!OidIsValid(cast_func))
+				return op;
+
+			subst = (Expr *)
+				makeFuncExpr(cast_func, tce->type_id, list_make1(subst), InvalidOid, InvalidOid, 0);
+		}
+		else if (tce->type_id == TIMESTAMPTZOID && width->consttype == INTERVALOID)
+		{
+			/*
+			 * TIMESTAMPTZ OP INTERVAL is marked stable and unsafe
```
cevian: probably not for this PR but I do wonder if we could just increase the interval by 2 days to be safe here?

src/plan_expand_hypertable.c
svenklemm: done

src/plan_expand_hypertable.c
svenklemm: for lower bound comparison the width does not actually matter so i adjusted the logic and also tried documenting it a bit more

src/plan_expand_hypertable.c
svenklemm: thats intentional because estimate_expression_value is not safe for those but subst still has  a valid expression we can use we just dont try to estimate it

PR #1155, opened 2019-04-10
Limit initial max_open_chunks_per_insert to PG_INT16_MAX

svenklemm:
When the initial value for max_open_chunks_per_insert is calculated
it may be bigger then the max for high values of work_mem. This
patch limits the initial value to the max of the GUC.

### DIFF
--- a/src/guc.c
+++ b/src/guc.c
@@ -133,15 +133,16 @@ _guc_init(void)
 							"Maximum open chunks per insert",
 							"Maximum number of open chunk tables per insert",
 							&ts_guc_max_open_chunks_per_insert,
-							work_mem * 1024L / 25000L, /* Measurements via
-														* `MemoryContextStats(TopMemoryContext)`
-														* show chunk insert
-														* state memory context
-														* takes up ~25K bytes
-														* (work_mem is in
-														* kbytes) */
+							Min(work_mem * INT64CONST(1024) / INT64CONST(25000),
+								PG_INT16_MAX), /* Measurements via
+												* `MemoryContextStats(TopMemoryContext)`
+												* show chunk insert
+												* state memory context
+												* takes up ~25K bytes
+												* (work_mem is in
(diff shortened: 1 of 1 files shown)


src/guc.c
```diff
@@ -133,13 +133,14 @@ _guc_init(void)
...
 							"Maximum number of open chunk tables per insert",
 							&ts_guc_max_open_chunks_per_insert,
-							work_mem * 1024L / 25000L, /* Measurements via
-														* `MemoryContextStats(TopMemoryContext)`
-														* show chunk insert
-														* state memory context
-														* takes up ~25K bytes
-														* (work_mem is in
-														* kbytes) */
+							Min(work_mem * 1024L / 25000L,
```
JLockerman:
```suggestion
							Min(work_mem * INT64CONST(1024) / INT64CONST(25000L),
```
I think otherwise it can still overflow the multiplication

RobAtticus:
Would just suggest a nit to the commit subject line:
```
Limit initial max_open_chunks_per_insert to 16-bit max
```
Slightly over 50 chars, but it's clearer.

Or:
```
Limit initial GUC value to 16-bit max
```

Since you mention its name in the body

src/guc.c
```diff
@@ -133,13 +133,14 @@ _guc_init(void)
...
+								65536L), /* Measurements via
```
JLockerman:
```suggestion
								PG_INT16_MAX), /* Measurements via
```

RobAtticus:
Or since it's also the GUC max:
```
Limit initial max_open_chunks_per_insert to GUC max
```

```
Limit initial GUC value to its max value
```

Basically I was confused by what `Max` in the summary meant

svenklemm:
I'm not limiting to 16-bit max though but the max for that specific GUC

RobAtticus:
Yeah just noticed, so I suggested two others ^

PR #1156, opened 2019-04-10
Add ability to work with aggregate partials

cevian:
The ability to get aggregate partials instead of the final state
is important for both continuous aggregation and clustering.

This commit adds the ability to work with aggregate partials.
Namely a function called _timescaledb_internal.partialize_agg
can now wrap an aggregate and return the partial results as a bytea.

The _timescaledb_internal.finalize_agg aggregate allows you to combine
and finalize partials.

The partialize_agg function works as a marker in the planner to force
the planner to return partial result.

Unfortunately, we could not get the planner to modify the plan directly
to aggregate partials. Instead, the finalize_agg is a real aggregate
that performs aggregation on the partial state. Note that it is not
yet parallel.

Aggregate that use FINALFUNC_EXTRA are currently not supported.

Co-authored-by: gayyappan <gayathri@timescale.com>
Co-authored-by: David Kohn <david@timescale.com>


tsl/test/sql/partial_aggs.sql
```diff
@@ -0,0 +1,178 @@
...
... be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k. this must be over 2k.
```
JLockerman: could we do something like `repeat('this must be over 2k. ', 1100)` for these, and do an equality check on the output?

tsl/test/expected/partial_aggs.out
```diff
@@ -0,0 +1,224 @@
...
+create or replace  view v1(a, partialb, partialtv) as select a, _timescaledb_internal.partialize_agg( max(b) ), _timescaledb_internal.partialize_agg( min(toastval)) from foo group by a;
+create table t1 as select * from v1;
+-- 2145 min(text), 2127 max(timestamp)
+-- select regprocedureout( 2127);
+insert into t1 select * from v1;
+select a, _timescaledb_internal.finalize_agg( 'max(timestamp with time zone)', null, null, partialb, null::timestamptz ) maxb, substring(_timescaledb_internal.finalize_agg( 'min(text)', 'pg_catalog', 'default', partialtv, null::text ) from 1 for 20) mintv
+from t1 group by a order by a;
+ a |             maxb             |        mintv         
+---+------------------------------+----------------------
+ 1 | Wed Oct 19 10:23:54 2005 PDT |  I am a tall big gir
```
JLockerman: we should check that the outputted string is equal to the inputted one, or at least has the same length

src/plan_partialize.c
```diff
@@ -0,0 +1,114 @@
...
+		state->found_partialize = true;
+		state->looking_for_agg = true;
+	}
+
+	return expression_tree_walker((Node *) node, partialize_function_call_walker, state);
+}
+
+/* We currently cannot handle cases like
+ *     SELECT sum(i), partialize(sum(i)) ...
+ * instead we use this function to ensure that if any of the aggregates in a staetment are
```
svenklemm: staetment => statement

src/plan_partialize.c
```diff
@@ -0,0 +1,114 @@
...
+									.fnoid = InvalidOid };
+	ListCell *lc;
+
+	if (CMD_SELECT != parse->commandType)
+		return;
+	partialfnoid = get_function_oid(TS_PARTIALFN, INTERNAL_SCHEMA_NAME, lengthof(argtyp), argtyp);
+	Assert(partialfnoid != InvalidOid);
+
+	state.fnoid = partialfnoid;
+	expression_tree_walker((Node *) parse->targetList, partialize_function_call_walker, &state);
```
svenklemm: this should be
partialize_function_call_walker((Node *) parse->targetList, &state);

src/plan_partialize.c
```diff
@@ -0,0 +1,114 @@
...
+
+	if (state.found_partialize)
+	{
+		expression_tree_walker((Node *) parse->targetList, ensure_only_partials, NULL);
```
svenklemm: this should be
ensure_only_partials((Node *) parse->targetList, NULL);

sql/updates/latest-dev.sql
```diff
@@ -2,3 +2,21 @@
 -- we add an addition optional argument to locf
 DROP FUNCTION IF EXISTS locf(ANYELEMENT,ANYELEMENT);
 
+CREATE OR REPLACE FUNCTION _timescaledb_internal.finalize_agg_sfunc(
+tstate internal, aggfn TEXT, inner_agg_collation_schema NAME, inner_agg_collation_name NAME, inner_agg_serialized_state BYTEA, return_type_dummy_val ANYELEMENT)
```
svenklemm: why is aggfn not REGPROC?

JLockerman:
Otherwise LGTM

tsl/test/sql/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ set(TEST_FILES
     edition.sql
     gapfill.sql
     reorder.sql
+    ddl_hook.sql
```
svenklemm: merge error, ddl_hook needs to be in TEST_FILES_DEBUG

tsl/test/sql/partial_aggs.sql
```diff
@@ -0,0 +1,178 @@
+-- This file and its contents are licensed under the Timescale License.
```
svenklemm: i'd like to see tests with non-toplevel calls for this

sql/updates/latest-dev.sql
cevian: because that's not safe for dump/restore.

src/plan_partialize.c
```diff
@@ -0,0 +1,114 @@
...
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+#include <postgres.h>
+#include <catalog/pg_type.h>
+#include <nodes/nodeFuncs.h>
+#include <optimizer/planner.h>
+#include <utils/lsyscache.h>
+#include <plan_partialize.h>
```
svenklemm: this should use
#include "plan_partialize.h"

davidkohn88:
LGTM, btw.

PR #1175, opened 2019-04-19
Fix ts_chunk_for_tuple

JLockerman:
ts_chunk_for_tuple should use the chunk cache.
ts_chunk_for_tuple should be marked stable.
These fixes markedly improve performance.

### DIFF
--- a/sql/chunk.sql
+++ b/sql/chunk.sql
@@ -49,4 +49,4 @@ AS '@MODULE_PATHNAME@', 'ts_chunks_in' LANGUAGE C VOLATILE STRICT;
 -- returns the chunk id for the chunk that would contain a given tuple, or NULL if no such chunk
 -- currently exists in the hypertable. tuple must be of the hypertable's row type.
 CREATE OR REPLACE FUNCTION _timescaledb_internal.chunk_for_tuple(hypertable_id INTEGER, tuple ANYELEMENT) RETURNS INTEGER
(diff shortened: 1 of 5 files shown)


sql/chunk.sql
```diff
@@ -49,4 +49,4 @@ AS '@MODULE_PATHNAME@', 'ts_chunks_in' LANGUAGE C VOLATILE STRICT;
 -- returns the chunk id for the chunk that would contain a given tuple, or NULL if no such chunk
 -- currently exists in the hypertable. tuple must be of the hypertable's row type.
 CREATE OR REPLACE FUNCTION _timescaledb_internal.chunk_for_tuple(hypertable_id INTEGER, tuple ANYELEMENT) RETURNS INTEGER
-AS '@MODULE_PATHNAME@', 'ts_chunk_for_tuple' LANGUAGE C VOLATILE STRICT;
+AS '@MODULE_PATHNAME@', 'ts_chunk_for_tuple' LANGUAGE C STABLE STRICT;
```
cevian: Should also be parallel safe

src/hypertable.c
```diff
@@ -724,39 +724,54 @@ hypertable_chunk_store_add(Hypertable *h, Chunk *chunk)
 	return cse;
 }
 
-Chunk *
-ts_hypertable_get_chunk(Hypertable *h, Point *point)
+static inline Chunk *
+hypertable_get_chunk(Hypertable *h, Point *point, bool create_on_miss)
```
cevian: this should be `create_if_not_exists` It does not always create on cache miss.

PR #1182, opened 2019-04-23
Enable optimizing SELECTs within INSERTs

JLockerman:
Before this PR only SELECTs would be optimized to exclude unneeded
chunks by our planner. This PR enables such optimizations on SELECTs
found within an INSERT as well. This should speed up commands of the
form

    INSERT INTO <hypertable> (SELECT ... FROM <hyepertable> WHERE ...)

We would like to enable this for all commands, but currently DELETE and
UPDATE can not handle them, and cause errors when the optimizations are
enabled.

This commit also fixes an issue that would occur if we tried to exclude
chunks based off of infinite time values.

### DIFF
--- a/src/hypertable_restrict_info.c
+++ b/src/hypertable_restrict_info.c
@@ -102,7 +102,7 @@ dimension_restrict_info_open_add(DimensionRestrictInfoOpen *dri, StrategyNumber 
 												   PointerGetDatum(lfirst(item)),
 												   dimvalues->type,
 												   &restype);
-		int64 value = ts_time_value_to_internal(datum, restype);
+		int64 value = ts_time_value_to_internal_or_infinite(datum, restype, NULL);
 
 		switch (strategy)
 		{
--- a/src/utils.c
+++ b/src/utils.c
@@ -182,6 +182,80 @@ ts_integer_to_internal(Datum time_val, Oid type_oid)
 	}
 }
 
+int64
+ts_time_value_to_internal_or_infinite(Datum time_val, Oid type_oid,
+									  TimevalInfinity *is_infinite_out)
+{
+	switch (type_oid)
+	{
+		case TIMESTAMPOID:
+		{
+			Timestamp ts = DatumGetTimestamp(time_val);
+			if (TIMESTAMP_NOT_FINITE(ts))
+			{
+				if (TIMESTAMP_IS_NOBEGIN(ts))
+				{
+					if (is_infinite_out != NULL)
+						*is_infinite_out = TimevalNegInfinity;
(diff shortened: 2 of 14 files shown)


src/utils.h
```diff
@@ -31,7 +38,11 @@ extern bool ts_type_is_int8_binary_compatible(Oid sourcetype);
  * Will throw an error for that, or other conversion issues.
  */
 extern TSDLLEXPORT int64 ts_time_value_to_internal(Datum time_val, Oid type);
+extern int64 ts_time_value_to_internal_or_infinite(Datum time_val, Oid type_oid,
```
gayyappan: why do we need two variants of this function? shouldn't callers always be using the ts_time_value_to_internal_or_infinite variant.Is that change for another PR?

src/utils.c
```diff
@@ -182,6 +182,131 @@ ts_integer_to_internal(Datum time_val, Oid type_oid)
 	}
 }
 
+int64
+ts_time_value_to_internal_or_infinite(Datum time_val, Oid type_oid,
+									  TimevalInfinity *is_infinite_out)
+{
+	switch (type_oid)
+	{
+		case TIMESTAMPOID:
```
svenklemm: i think you should merge TIMESTAMPOID and TIMESTAMPTZOID branch since the code is identical

src/utils.h
JLockerman: we should not really store the result of `ts_time_value_to_internal_or_infinite` since most postgres functions cannot handle infinities, it's really mostly useful for our own computation, where we know how to handle it.

src/utils.c
JLockerman: not quite identical, they're still different types

src/hypertable_restrict_info.c
```diff
@@ -98,11 +98,12 @@ dimension_restrict_info_open_add(DimensionRestrictInfoOpen *dri, StrategyNumber
 	foreach (item, dimvalues->values)
 	{
 		Oid restype;
+		TimevalInfinity is_infinite;
```
svenklemm: This is never checked so why is it part of the signature? I'm wondering whether macros would be more useful to check if an internal value is finite

src/utils.c
```diff
@@ -182,6 +182,131 @@ ts_integer_to_internal(Datum time_val, Oid type_oid)
...
+		{
+			Timestamp ts = DatumGetTimestamp(time_val);
+			if (TIMESTAMP_NOT_FINITE(ts))
```
svenklemm: None of the NOT_FINITE code pathes are hit by any of the tests

PR #1183, opened 2019-04-24
Handle NULL in a check constraint name

erthalion:
Hi,

I've got a strange situation. When I perform a DDL on a hypertable with adding
a column and an invalid check constraing, at the validation time it crashes. My
investigation resulted in this small change.

Check constraints are propagated to child chunks by default, but they're
missing from _timescaledb_catalog.chunk_constraint table. It can lead to a
situation, when such a constraint will be added as not valid, and then at
validation time validate_hypertable_constraint will construct a chunk_cmd with
a NULL name (since it was not there and
ts_chunk_constraint_get_name_from_hypertable_constraint returns NULL), what
eventually leads to a crash.

I haven't fixed any tests tests (looks like they're broken just because more
constraints are registered), because I would like to get a confirmation first
that my conclusions are correct or I'm missing something. Also I see two tests
from constraint.sql for constraint validation, but they're only for the
negative path.

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -1141,6 +1141,9 @@ validate_hypertable_constraint(Hypertable *ht, Oid chunk_relid, void *arg)
 	chunk_cmd->name =
 		ts_chunk_constraint_get_name_from_hypertable_constraint(chunk_relid, cmd->name);
 
+	if (chunk_cmd->name == NULL)
+		return;
+
 	/* do not pass down the VALIDATE RECURSE subtype */
 	chunk_cmd->subtype = AT_ValidateConstraint;
 	AlterTableInternal(chunk_relid, list_make1(chunk_cmd), false);
--- a/test/expected/constraint.out
+++ b/test/expected/constraint.out
@@ -783,3 +783,19 @@ SELECT * FROM create_hypertable('hyper_unique', 'time', chunk_time_interval => 1
 INSERT INTO hyper_unique(time, device_id,sensor_1) VALUES (1257987700000000000, 'dev2', 11);
 ALTER SCHEMA my_associated_schema RENAME TO new_associated_schema;
 ALTER TABLE hyper_unique DROP CONSTRAINT hyper_unique_time_key;
+-- test for constraint validation crash, see #1183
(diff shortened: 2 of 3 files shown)


cevian:
Hi @erthalion, 
Thank you very much for this PR. Sorry that it has taken a few days to respond. 

This is indeed a bug. I wonder if this is the correct approach for a fix though. Postgres itself already handles check constraints on an inherited table which is why we don't have these in our catalog tables. My fear is that adding it in will cause us to do work twice. For example with the current approach I think the validation will run twice, once in Postgres core and once in `validate_hypertable_constraint`.

An alternative fix is to add 
```
if(chunk_cmd->name == NULL)
   return
```
to  `validate_hypertable_constraint`. That way we simply get out of Postgres's way in these cases.

What do you think of that alternative?

erthalion:
@cevian Thank you for reply!

> An alternative fix is to add
>
> if(chunk_cmd->name == NULL)
>   return
> to validate_hypertable_constraint. That way we simply get out of Postgres's way in these cases.
>
> What do you think of that alternative?

Yes, of course that would work too. My only concern is that it fixes only the crash itself, and there could be some other code, that relies in a similar way on those checks being present in chunk_constraint. Although I'm not familiar with the code base good enough, so if you say that chances for that are rather small, it's good enough for me (then I can update the code to verify whether the name is null and add a test case)

cevian:
@erthalion Yes I think that should work. I think most other code paths that use check constraints are already tested. If you could change the PR and add some tests that would be greatly appreciated. 

Just an FYI, the 1.3.0 release is scheduled for Monday, so to get in this PR must make it in tomorrow. Otherwise, 1.3.1 will be a ~2 weeks out.

Thanks again, Mat.

erthalion:
I believe that's how it should be, more or less. But I'm not sure why travis fails on PG11 with the regression.diff pointing to the Postgres isolation tests and how is it related to this change.

cevian:
@erthalion  It was just a flaky test. Worked once restarted. Can we remove the last commit and merge

erthalion:
@cevian I've removed the empty commit.

cevian:
@erthalion Thanks again! We really appreciate it!

erthalion:
Thank you!

PR #1184, opened 2019-04-24
Add continuous aggregate functionality.

cevian:


JLockerman:
(note, the failure we're seeing on travis now is an unrelated issue that occurs when you have  something like ```SELECT time_bucket(n, big_int_max) as time FROM ... WHERE time < foo``` we had an unprincipled workaround in the private branch, but may want to wait for a more complete one before merging this)

PR #1186, opened 2019-04-25
Improve overflow handling in time_bucket transformation

svenklemm:
The expression generated when transforming time_bucket comparison
in qual can overflow leading to postgres throwing an error for an
otherwise valid query. This patch catches any errors when evaluating
the expressions and skips adding the transformation if an error
is thrown.

### DIFF
--- a/src/plan_expand_hypertable.c
+++ b/src/plan_expand_hypertable.c
@@ -169,6 +169,7 @@ transform_time_bucket_comparison(PlannerInfo *root, OpExpr *op)
 	{
 		/* column < value + width */
 		Expr *subst;
+		volatile Expr *estimate;
 		Oid resulttype;
 		Oid subst_opno = get_operator("+",
 									  PG_CATALOG_NAMESPACE,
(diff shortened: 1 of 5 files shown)


src/plan_expand_hypertable.c
```diff
@@ -169,6 +169,7 @@ transform_time_bucket_comparison(PlannerInfo *root, OpExpr *op)
 	{
 		/* column < value + width */
 		Expr *subst;
```
cevian: I think `subst` also needs to be declared volatile since it's used in both try and catch

src/plan_expand_hypertable.c
```diff
@@ -245,7 +253,48 @@ transform_time_bucket_comparison(PlannerInfo *root, OpExpr *op)
...
 
 			if (interval->day == 0 && interval->month == 0)
-				subst = (Expr *) estimate_expression_value(root, (Node *) subst);
+			{
+				/*
+				 * wrap estimation in try/catch in case it overflows
+				 */
+				PG_TRY();
+				{
+					subst = (Expr *) estimate_expression_value(root, (Node *) subst);
```
cevian: here is where subst is used in both try and catch

test/sql/include/plan_expand_hypertable_query.sql
```diff
@@ -146,6 +146,11 @@ SELECT * FROM cte ORDER BY value;
 :PREFIX SELECT * FROM hyper WHERE 10::bigint > time_bucket(10, time) ORDER BY time;
 :PREFIX SELECT * FROM hyper WHERE 11::bigint > time_bucket(10, time) ORDER BY time;
 
+-- test overflow behaviour of time_bucket exclusion
+SET client_min_messages TO error;
+:PREFIX SELECT * FROM hyper WHERE time > 950 AND time_bucket(10, time) < '9223372036854775807'::bigint ORDER BY time;
```
cevian: is it possible to add one overflow case for timestamptz and interval?

svenklemm: added more tests

PR #1189, opened 2019-04-26
Add default index for continuous aggregates

gayyappan:
Add indexes for materialization table created by continuous aggregates.
This behavior can be turned on/off by using timescaledb.create_groupcol_index parameter
of the WITH clause when the continuous agg is created.

### DIFF
--- a/src/continuous_agg.c
+++ b/src/continuous_agg.c
@@ -50,6 +50,11 @@ static const WithClauseDefinition continuous_aggregate_with_clause_def[] = {
 			.arg_name = "max_interval_per_job",
 			.type_id = TEXTOID,
 		},
+		[ContinuousViewOptionCreateGroupIndex] = {
+			.arg_name = "create_group_indexes",
+			.type_id = BOOLOID,
+			.default_val = BoolGetDatum(true),
+		},
 };
 
 WithClauseResult *
--- a/tsl/src/continuous_aggs/create.c
+++ b/tsl/src/continuous_aggs/create.c
@@ -19,12 +19,15 @@
 #include <nodes/parsenodes.h>
 #include <nodes/makefuncs.h>
 #include <nodes/nodeFuncs.h>
+#include <catalog/index.h>
 #include <catalog/pg_type.h>
 #include <catalog/pg_aggregate.h>
 #include <catalog/toasting.h>
 #include <catalog/pg_collation.h>
 #include <catalog/pg_trigger.h>
+#include <commands/defrem.h>
 #include <commands/tablecmds.h>
+#include <commands/tablespace.h>
 #include <commands/trigger.h>
 #include <commands/view.h>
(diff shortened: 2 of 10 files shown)


src/continuous_agg.c
```diff
@@ -50,6 +50,11 @@ static const WithClauseDefinition continuous_aggregate_with_clause_def[] = {
 			.arg_name = "max_interval_per_job",
 			.type_id = TEXTOID,
 		},
+		[ContinuousViewOptionCreateGroupIndex] = {
+			.arg_name = "create_groupcol_index",
```
davidkohn88: I would name this `create_group_indexes` or something like that.

tsl/src/continuous_aggs/create.c
```diff
@@ -134,8 +137,12 @@ typedef struct MatTableColumnInfo
 	List *matcollist;		 /* column defns for materialization tbl*/
 	List *partial_seltlist;  /* tlist entries for populating the materialization table columns */
 	List *partial_grouplist; /* group clauses used for populating the materialization table */
-	int matpartcolno;		 /*index of partitioning column in matcollist */
-	char *matpartcolname;	/*name of the partition column */
+	List *mat_groupcolname_list; /* names of columns that are populated by the group-by clause
```
davidkohn88: So, currently it looks like we do it over the mat_groupcolname_list, which is all the group by columns, even if they don't actually end up in the materialization, do we only want to create the indexes if they also appear in the final view? (though this seems like a bit of an edge case)

tsl/src/continuous_aggs/create.c
```diff
@@ -316,6 +324,42 @@ cagg_add_trigger_hypertable(Oid relid, char *trigarg)
...
+		.relation = makeRangeVar(NameStr(ht->fd.schema_name), NameStr(ht->fd.table_name), 0),
+		.tableSpace = get_tablespace_name(get_rel_tablespace(ht->main_table_relid)),
+	};
+	IndexElem timeelem = { .type = T_IndexElem,
+						   .name = matcolinfo->matpartcolname,
+						   .ordering = SORTBY_DESC };
+	ListCell *le = NULL;
+	foreach (le, matcolinfo->mat_groupcolname_list)
+	{
+		char *grpcolname = (char *) lfirst(le);
```
cevian: foreach index created we should elog a NOTICE `adding group index named %s BTREE(%s, %s)`

src/continuous_agg.c
cevian: I do like `create_group_indexes`

gayyappan: modified.

tsl/src/continuous_aggs/create.c
gayyappan: defer this for later.

PR #1191, opened 2019-04-27
Chunk for tuple optimization

cevian:


src/utils.c
```diff
@@ -538,6 +538,12 @@ ts_function_types_equal(Oid left[], Oid right[], int nargs)
 	return true;
 }
 
+/* Get a function oid, throwing an error if not found.
+ *
+ * Note: This uses the same cached lookup as the parser, so should
+ * be pretty effecient.
+ */
+
```
svenklemm: i think we should get rid of this function and use LookupFuncName

src/chunk.c
```diff
@@ -1536,6 +1537,160 @@ ts_chunk_get_by_relid(Oid relid, int16 num_constraints, bool fail_if_not_found)
...
+		te->expr = (Expr *)
+			makeConst(INT4OID, -1, InvalidOid, 4, Int32GetDatum(chunk->fd.id), false, true);
+
+		/* we only expect one such entry so it's ok to return here */
+		return true;
+	}
+	return false;
+}
+
+/* Recurse throughout the plan to do this optimization */
```
svenklemm: you are not recursing properly here and your logic will only work for toplevel elements, this needs more tests where toplevel is not append, i think mergeappend will not work with the current implementation

cevian:
Closed in favor of #1193 which cut down the execution time of getting the chunk_id so much that this complexity of an optimization is no longer worth it

PR #1192, opened 2019-04-29
Fix release build warning

svenklemm:
The address variable in process_create_trigger_start is only used
in an assert in debug builds so compiler will warn about this
variable in non-debug builds. This patch sets the unused attribute
for the variable to silence the warning.

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -2588,7 +2588,7 @@ process_create_trigger_start(ProcessUtilityArgs *args)
 	CreateTrigStmt *stmt = (CreateTrigStmt *) args->parsetree;
 	Cache *hcache;
 	Hypertable *ht;
-	ObjectAddress address;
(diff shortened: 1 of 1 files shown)


src/process_utility.c
```diff
@@ -2588,7 +2588,7 @@ process_create_trigger_start(ProcessUtilityArgs *args)
 	CreateTrigStmt *stmt = (CreateTrigStmt *) args->parsetree;
 	Cache *hcache;
 	Hypertable *ht;
-	ObjectAddress address;
+	ObjectAddress __attribute__((unused)) address;
```
JLockerman:
```suggestion
	ObjectAddress pg_attribute_unused() address;
```
`__attribute__((unused))` is not well supported

JLockerman:
```suggestion
	ObjectAddress PG_USED_FOR_ASSERTS_ONLY address;
```
actually

PR #1195, opened 2019-04-29
 Fix cascade in scheduled drop chunks, and allow scheduled drop_chunks to cascade to aggs

JLockerman:
Before this PR, cascade was unconditionally set to true, as stored if the pointer to the inputted form data was non-null instead of the value stored within it. This PR changes that, so the intended setting for cascade is stored instead.

This PR also adds a `cascade_to_materializations` flag to the scheduled version of drop_chunks that behaves much like the one from manual `drop_chunks`: if a hypertable that has a continuous aggregate tries to drop chunks, and this flag is not set, the chunks will not be dropped

### DIFF
--- a/tsl/test/expected/tsl_tables.out
+++ b/tsl/test/expected/tsl_tables.out
@@ -168,6 +168,12 @@ WARNING:  could not add drop_chunks policy due to existing policy on hypertable 
                      -1
 (1 row)
 
+select * from _timescaledb_config.bgw_policy_drop_chunks;
+ job_id | hypertable_id | older_than | cascade 
+--------+---------------+------------+---------
+   1002 |             1 | @ 3 mons   | t
+(1 row)
+
 \set ON_ERROR_STOP 0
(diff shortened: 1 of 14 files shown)


tsl/test/expected/tsl_tables.out
```diff
@@ -168,6 +168,13 @@ WARNING:  could not add drop_chunks policy due to existing policy on hypertable
                      -1
 (1 row)
 
+select add_drop_chunks_policy('test_table', INTERVAL '3 days', if_not_exists => true, cascade_to_materializations => true);
+WARNING:  could not add drop_chunks policy due to existing policy on hypertable with different arguments
+ add_drop_chunks_policy 
+------------------------
+                     -1
+(1 row)
+
```
cevian: could we add a `select * from _timescaledb_config.bgw_policy_drop_chunks;` here. I don't see any other tests that show the column with a value of 'true'. Would be nice to avoid the same class of error we saw before.

PR #1200, opened 2019-05-01
Coverity fixes for 1.3

cevian:


src/plan_expand_hypertable.c
```diff
@@ -236,11 +236,13 @@ transform_time_bucket_comparison(PlannerInfo *root, OpExpr *op)
 				if (interval->month != 0)
 					return op;
 				if (DatumGetDateADT(castNode(Const, value)->constvalue) >=
-					DATEVAL_NOEND - interval->day + ceil(interval->time / USECS_PER_DAY))
+					DATEVAL_NOEND - interval->day +
+						ceil((double) interval->time / (double) USECS_PER_DAY))
```
JLockerman: Has @svenklemm looked at this?

tsl/src/continuous_aggs/create.c
```diff
@@ -519,8 +517,9 @@ get_timebucketfnoid()
 	Oid funcoid;
 	const char *funcname = TIMEBUCKETFN;
 	CatCList *catlist = SearchSysCacheList1(PROCNAMEARGSNSP, CStringGetDatum(funcname));
+	int i;
 
-	for (int i = 0; i < catlist->n_members; i++)
+	for (i = 0; i < catlist->n_members; i++)
```
RobAtticus: Can we figure out how to get the compiler to fail (either locally or on Travis) for this kind of thing? Sometimes it doesn't get caught and it causes problems when building on Debian/Ubuntu

cevian: Well  coverity fails but not noisily. I can ask @JLockerman to take a look. But not for this PR.

JLockerman: disable the warning everywhere 😜 
more seriously, the warning isn't widely available, I think it's only some versions of GCC that offer it.

src/plan_expand_hypertable.c
svenklemm: LGTM

PR #1201, opened 2019-05-01
Add telemetry for policies: drop_chunk & reorder

cevian:
The commit will add telemetry for the number of drop_chunks and
reorder policies in the database. This will allow us to see
how much these features are used.

### DIFF
--- a/src/telemetry/telemetry.c
+++ b/src/telemetry/telemetry.c
@@ -23,6 +23,8 @@
 #include "net/http.h"
 #include "jsonb_utils.h"
 #include "license_guc.h"
+#include "bgw_policy/drop_chunks.h"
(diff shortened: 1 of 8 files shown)


src/telemetry/telemetry.c
```diff
@@ -173,6 +177,24 @@ get_num_continuous_aggs()
...
+	return buf->data;
+}
+
+static char *
+get_num_reorder_policies()
+{
+	StringInfo buf = makeStringInfo();
+
+	appendStringInfo(buf, "%d", ts_bgw_policy_reorder_count());
+	return buf->data;
```
JLockerman: Hey, that's clever, we should do this more often

PR #1203, opened 2019-05-02
Rename telemetry_metadata table to just metadata

WireBaron:
This change renames the _timescale_catalog.telemetry_metadata to
_timescale_catalog.metadata.  It also adds a new boolean column to this
table which is used to flag data which should be included in telemetry.

It also renamed the src/telemetry/metadata.{h,c} files to
src/telemetry/telemetry_metadata.{h,c} and updated the API to reflect
this.  Finally it also includes the logic to use the new boolean column
when populating the telemetry parse state.


sql/updates/latest-dev.sql
```diff
@@ -41,8 +41,8 @@ CREATE AGGREGATE _timescaledb_internal.finalize_agg(agg_name TEXT,  inner_agg_co
     FINALFUNC_EXTRA
 );
 
-ALTER TABLE _timescaledb_catalog.installation_metadata RENAME TO telemetry_metadata;
-ALTER INDEX _timescaledb_catalog.installation_metadata_pkey RENAME TO telemetry_metadata_pkey;
+ALTER TABLE _timescaledb_catalog.installation_metadata RENAME TO metadata;
```
erimatnor: I would just keep the old name `installation_metadata` and remove these lines from the update script.

src/telemetry/metadata.h
```diff
@@ -1,17 +0,0 @@
-/*
```
erimatnor: Why isn't this file simply renamed instead of deleted and re-added? Now there's an unnecessarily large diff.

You should try to do 

```
git mv  src/telemetry/metadata.h src/telemetry/telemetry_metadata.h
```

Same goes for `metadata.c`.

(Actually unclear why these needs to be renamed since already namespaced by the telemetry dir).

WireBaron: This actually was done as a `git mv`, and was shown as a rename in my local `git status`.  Somehow that got lost when I pushed the change to github.

sql/updates/latest-dev.sql
WireBaron: Unfortunately the `installation_metadata` to `telemetry_metadata` rename is going out in 1.3 which is already frozen.  So I'll have to change this to rename from `telemetry_metadata` to whatever we want the new name to be.  

As far as whether we want to go with `installation_metadata` again, I actually found that name to be confusing as I was reading as metadata needed or created at install time (I realize now that was my parsing error).  I still prefer simply `metadata` as a generic catch-all table name, but if you have a strong preference we can have a broader discussion.

src/telemetry/metadata.h
erimatnor: can it be fixed somehow. I think it helps things like rebasing other branches.

sql/updates/latest-dev.sql
svenklemm: I would prefer to keep the telemetry metadata separate from metadata required for operation.

Also the 1.3.0 release PR will empty this file so this change has to be added after 1.3.0 has been released

cevian:
@WireBaron I think this needs some changes now that 1.3 is out.

WireBaron:
@erimatnor @cevian 
This should be ready for re-review now that it has the 1.3 latest-dev.sql changes.  Please let me know if there's more I need to do to handle updates.

src/metadata.c
```diff
@@ -15,11 +15,11 @@
 #include <utils/datum.h>
 
 #include "catalog.h"
-#include "telemetry_metadata.h"
+#include "metadata.h"
 #include "scanner.h"
 
 #define TYPE_ERROR(inout, typeid)                                                                  \
-	elog(ERROR, "ts::telemetry_metadata: no %s function for type %u", inout, typeid);
+	elog(ERROR, "ts::metadata: no %s function for type %u", inout, typeid);
```
erimatnor: Maybe change this to `ts_metadata` now that we have decided on that prefix.

src/telemetry/metadata.c
```diff
@@ -1,108 +0,0 @@
-/*
```
erimatnor: It is still not clear to me why this file is deleted rather than renamed like other files. Not sure why it needs renaming either, given that it is namespaced by the telemetry dir already. In any case, I'd like to see this file moved rather than deleted if possible because I believe it helps rebasing.

WireBaron: Git just seems incapable of recognizing a move if the file is sufficiently modified with the same change.  Nothing I try here seems to make any difference.  I split the move and modifications into two changes, both of which looked correct when view by themselves, but when viewing a PR of the pair of changes, it still presented it as a delete and new file.

src/metadata.c
WireBaron: Done

WireBaron:
@cevian @erimatnor @svenklemm 
Fixed a bug in the update code, if someone wants to look at the latest-dev.sql changes where I'm actually adding the new column.  I'll merge this later today if no one objects.

PR #1207, opened 2019-05-02
Fix partial view targetlist for continuous aggregates

gayyappan:
The partial view should always project the time_bucket expression related
column as this is a special column for the materialization table. The partial
view failed to project it when the user query's SELECT targetlist did not
contain the time_bucket expression. The materialization fails in this
scenario.


cevian:
Great catch!

PR #1215, opened 2019-05-07
Add scheduled indexes

JLockerman:
This commit adds scheduled indexes. A scheduled index is not created on
new chunks when the chunk is created, but instead created by a
background worker at some point in the future. The goal is that such
indexes will not slow down INSERTs as much for the live chunks, but can
still be used to speed up queries.


tsl/src/bgw_policy/job.c
```diff
@@ -219,20 +221,125 @@ bgw_policy_job_requires_enterprise_license(BgwJob *job)
...
+				(errcode(ERRCODE_TS_INTERNAL_ERROR),
+				 errmsg("could not run scheduled_index policy #%d because no args in policy table",
+						job_id)));
+
+	ht = ts_hypertable_get_by_id(args->fd.hypertable_id);
+
+	/* Find a chunk needing the index */
+	/* currently just use the same logic as reorder */
+	// chunk_id = get_chunk_id_needing_schedule_index(args->fd.job_id, ht);
+	chunk_id = get_chunk_id_to_reorder(args->fd.job_id, ht);
```
cevian: need to rename function plus should probably make the number of recent chunks to skip (ie`REORDER_SKIP_RECENT_DIM_SLICES_N`) a param to the function.

tsl/src/bgw_policy/job.c
```diff
@@ -219,20 +221,125 @@ bgw_policy_job_requires_enterprise_license(BgwJob *job)
...
+		elog(NOTICE,
+			 "no chunks need a scheduled index for hypertable %s.%s",
+			 ht->fd.schema_name.data,
+			 ht->fd.table_name.data);
+		goto commit;
+	}
+
+	chunk = ts_chunk_get_by_id(chunk_id, 0, false);
+	elog(LOG,
+		 "creating a scheduled index on chunk %s.%s",
```
cevian: add args->fd.hypertable_index_name to this message

tsl/src/bgw_policy/job.c
```diff
@@ -219,20 +221,125 @@ bgw_policy_job_requires_enterprise_license(BgwJob *job)
...
+		 chunk->fd.schema_name.data,
+		 chunk->fd.table_name.data);
+
+	/* Now update chunk_stats table */
+	ts_bgw_policy_chunk_stats_record_job_run(args->fd.job_id,
+											 chunk_id,
+											 ts_timer_get_current_timestamp());
+
+	// if (fast_continue && get_chunk_id_needing_schedule_index(args->fd.job_id, ht) != -1)
+	if (fast_continue && get_chunk_id_to_reorder(args->fd.job_id, ht) != -1)
```
cevian: don't we have a nicer function for fast continue now?

tsl/src/bgw_policy/scheduled_index_api.c
```diff
@@ -0,0 +1,245 @@
...
+							   NameGetDatum(&optional_index_info->fd.hypertable_index_name)));
+
+	// TODO: use update when optional_index_info already exists
+	ts_indexing_optional_info_insert(optional_index_info);
+
+	PG_RETURN_INT32(job_id);
+}
+
+Datum
+scheduled_index_remove_policy(PG_FUNCTION_ARGS)
```
cevian: if a policy is removed shouldn't we add indexes to any chunks that don't have them?

sql/pre_install/tables.sql
```diff
@@ -203,6 +211,13 @@ CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_drop_chunks (
 );
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_policy_drop_chunks', '');
 
+CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_scheduled_index (
+    job_id          		INTEGER     PRIMARY KEY REFERENCES _timescaledb_config.bgw_job(id) ON DELETE CASCADE,
+    hypertable_id   		INTEGER     UNIQUE NOT NULL    REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
```
cevian: the unique here makes no sense

tsl/src/bgw_policy/scheduled_index_api.c
```diff
@@ -0,0 +1,245 @@
...
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+				 errmsg("could not add scheduled_index policy because \"%s\" is not a hypertable",
+						get_rel_name(ht_oid))));
+
+	/* Now verify that the index is an actual index on that hypertable */
+	check_valid_index(ht, index_name);
+
+	/* Make sure that an existing policy doesn't exist on this hypertable */
+	existing = ts_bgw_policy_scheduled_index_find_by_hypertable(ts_hypertable_relid_to_id(ht_oid));
```
cevian: I think this should check per hypertable index not hypertable.

tsl/src/bgw_policy/scheduled_index_api.c
cevian: This should take a particular hypertable_index

erimatnor:
I am closing this PR since it hasn't been worked on for quite some time.

PR #1235, opened 2019-05-20
Add missing header guard

RobAtticus:
Code checking website lgtm.com flagged this as an alert. As far as
I can tell this is the only header file missing the guard.

### DIFF
--- a/src/plan_partialize.h
+++ b/src/plan_partialize.h
@@ -3,7 +3,11 @@
  * Please see the included NOTICE for copyright information and
  * LICENSE-APACHE for a copy of the license.
  */
+#ifndef TIMESCALEDB_PLAN_PARTIALIZE_H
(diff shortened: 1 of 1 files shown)


RobAtticus:
Here is our full `lgtm.com` page, fwiw:
https://lgtm.com/projects/g/timescale/timescaledb

src/plan_partialize.h
```diff
@@ -3,7 +3,11 @@
  * Please see the included NOTICE for copyright information and
  * LICENSE-APACHE for a copy of the license.
  */
+#ifndef TIMESCALEDB_PLANNER_PARTIALIZE_H
```
svenklemm: hmm i think this should be TIMESCALEDB_PLAN_PARTIALIZE_H to be consistent

RobAtticus: Oops, yep

Issue #1218, opened 2019-05-08
REFRESH cannot run inside a transaction block


dvdrozdov:
**Relevant system information:**
 - OS: Ubuntu 18.04
 - PostgreSQL version: 10.7
 - TimescaleDB version: 1.3.0
 - Installation method: "apt install"

**Describe the bug**
After timescaledb 1.3.0 installation refreshing of a materialized view inside a transaction fails with the following error:
```
REFRESH cannot run inside a transaction block
```

**To Reproduce**
```
createdb -U postgres test

test=# create materialized view mvtest as select 1;
SELECT 1

test=# select * from mvtest;
 ?column? 
----------
        1
(1 row)

test=# refresh materialized view mvtest;
REFRESH MATERIALIZED VIEW
test=# begin;
BEGIN
test=# refresh materialized view mvtest;
REFRESH MATERIALIZED VIEW
test=# commit;
COMMIT

CREATE EXTENSION IF NOT EXISTS timescaledb CASCADE;
...
Running version 1.3.0

test=# begin;
BEGIN
test=# refresh materialized view mvtest;
ERROR:  REFRESH cannot run inside a transaction block
test=# commit;
ROLLBACK

test=# select version();
                                                             version                                                              
----------------------------------------------------------------------------------------------------------------------------------
 PostgreSQL 10.7 (Ubuntu 10.7-1.pgdg18.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0, 64-bit
(1 row)

dpkg -l | grep timesc
ii  timescaledb-loader-postgresql-10       1.3.0~ubuntu18.04                  amd64        The loader for TimescaleDB to load individual versions.
ii  timescaledb-postgresql-10              1.3.0~ubuntu18.04                  amd64        An open-source time-series database based on PostgreSQL, as an extension.
ii  timescaledb-tools                      0.6.0~ubuntu18.04                  amd64        A suite of tools that can be used with TimescaleDB.
```

od0:
seeing this as well

cevian:
Solved by #1236. Will be part of 1.3.1

vi-vsn:
Hello,

I have originally ran into the issue reported in this thread while running version 1.3.0 of TimescaleDb.  This happened a few days ago. Today I have upgraded to 1.4.0 believing that this will resolve the problem, however it is now crashing in the exactly same routine, but with a different error message:

[...]

PR #1236, opened 2019-05-20
Stop preventing REFRESH in transaction blocks

JLockerman:
We prevent REFRESHing continuous aggregates within a transaction block
since we COMMIT and start our own transactions during this operation,
making it inherently non-transactional. This commit ensures we only
prevent REFRESHing continuous aggregates, and not REFRESH on other
materialized views.

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -2674,8 +2674,6 @@ process_refresh_mat_view_start(ProcessUtilityArgs *args, Node *parsetree)
 	NameData view_name;
 	NameData view_schema;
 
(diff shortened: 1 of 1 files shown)


cevian:
This addresses  #1218

t0k4rt:
Hello Timescale team, 
Do you have an idea when the fix will be available ?
Or is it possible to downgrade the extension safely ? 
Thanks for your help

JLockerman:
@t0k4rt we're planning to release early next week. If it's a problem until then you can disable timescale by setting `timescaledb.restoring`

t0k4rt:
Thanks for your help !

PR #1245, opened 2019-05-23
Fixes for PG 11.3

cevian:


src/constraint_aware_append.c
```diff
@@ -563,10 +563,17 @@ static CustomPathMethods constraint_aware_append_path_methods = {
 };
 
 Path *
-ts_constraint_aware_append_path_create(PlannerInfo *root, Hypertable *ht, Path *subpath)
+ts_constraint_aware_append_path_create(PlannerInfo *root, Hypertable *ht, Path *subpath,
+									   RelOptInfo *rel)
 {
 	ConstraintAwareAppendPath *path;
 
+#if !(PG96 || PG10)
```
erimatnor: I guess this could be a short-term fix, but it seems we are forgoing potential optimizations for partitioning in order to do runtime pruning. 

Also wondering if this is the right place to turn this off as it doesn't seem to be this node's business to make this decision. Seems like it should be a more top-level decision by the planner, e.g., at the same level it is deciding whether it should optimize append. Now this seems like a hidden non-obvious side-effect.

One other thing to note here is that PG partitioning has its own runtime pruning, so wondering if there's a way to hook into that in the future rather than doing our own thing.

src/plan_ordered_append.c
```diff
@@ -106,6 +106,12 @@ ts_ordered_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht
 	if (list_length(merge->subpaths) == 0)
 		return NULL;
 
+#if !(PG96 || PG10)
+	/* Turn off partitioning. Otherwise, `apply_scanjoin_target_to_paths` will blow away these paths
+	 * and will not call hooks to redo this */
+	rel->nparts = 0;
```
erimatnor: Again, seems like this should be more of a top-level decision by the planner rather than delegated to individual plan nodes. A top-level decision would also avoid having to do this at multiple places in plan nodes.

PR #1247, opened 2019-05-27
Add ordered append support for MergeJoin

svenklemm:
When doing a JOIN with a hypertable ordered append required the
ORDER BY clause to have the time partitioning column as first
expression. This patch loosens that restriction.
If the ORDER BY clause of a query does not reference our hypertable
but has an equality condition on the time partitioning column
of our hypertable we will do an ordered append because
it allows the planner to skip the Sort step for MergeJoin.

### DIFF
--- a/src/chunk_append/chunk_append.c
+++ b/src/chunk_append/chunk_append.c
@@ -18,6 +18,8 @@
 
 static bool contain_param_exec(Node *node);
 static bool contain_param_exec_walker(Node *node, void *context);
+static Var *find_equality_join_var(Var *sort_var, Index ht_relid, Oid eq_opr,
+								   List *join_conditions);
 
 static CustomPathMethods chunk_append_path_methods = {
 	.CustomName = "ChunkAppend",
@@ -81,6 +83,8 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
 			children = castNode(AppendPath, subpath)->subpaths;
 			break;
 		case T_MergeAppendPath:
+			if (!ordered)
+				return subpath;
 			children = castNode(MergeAppendPath, subpath)->subpaths;
 			path->cpath.path.pathkeys = subpath->pathkeys;
 			break;
@@ -89,8 +93,7 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
 			break;
 	}
 
-	Assert(ht->space->num_dimensions <= 2);
-	if (ht->space->num_dimensions == 1)
+	if (!ordered || ht->space->num_dimensions == 1)
 		path->cpath.custom_paths = children;
 	else
 	{
@@ -118,6 +121,8 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
 		ListCell *flat = list_head(children);
 		List *nested_children = NIL;
 		bool has_scan_childs = false;
+
+		Assert(ht->space->num_dimensions <= 2);
 
 		foreach (lc, nested_oids)
 		{
@@ -203,13 +208,18 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
  * Check if conditions for doing ordered append optimization are fulfilled
  */
 bool
-ts_ordered_append_should_optimize(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, bool *reverse)
+ts_ordered_append_should_optimize(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht,
+								  List *join_conditions, bool *reverse)
 {
 	SortGroupClause *sort = linitial(root->parse->sortClause);
 	TargetEntry *tle = get_sortgroupref_tle(sort->tleSortGroupRef, root->parse->targetList);
 	RangeTblEntry *rte = root->simple_rte_array[rel->relid];
 	TypeCacheEntry *tce;
 	char *column;
+	Index ht_relid = rel->relid;
+	Index sort_relid;
+	Var *ht_var;
+	Var *sort_var;
 
 	/* these are checked in caller so we only Assert */
 	Assert(!ts_guc_disable_optimizations && ts_guc_enable_ordered_append &&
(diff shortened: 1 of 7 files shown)


src/chunk_append/chunk_append.c
```diff
@@ -119,6 +120,8 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht,
 		List *nested_children = NIL;
 		bool has_scan_childs = false;
 
+		Assert(ht->space->num_dimensions <= 2);
```
JLockerman: What happens if the table has more than 2 space dimensions?

svenklemm: then you won't get ordered append plan, this is checked properly in caller that's why it's only asserted there

src/chunk_append/chunk_append.c
```diff
@@ -281,3 +291,63 @@ contain_param_exec_walker(Node *node, void *context)
...
+{
+	Var *sort_var;
+	Var *ht_var;
+	Index sort_relid;
+	Index ht_relid;
+	Oid eq_opr;
+} find_equality_join_var_context;
+
+static bool
+find_equality_join_var_walker(Node *node, find_equality_join_var_context *context)
```
gayyappan: should have a test with  a) > 2 hypertables involved in a  join, b) more than 2 hypertables and a non-hypertable involved in join. (not sure if this is possible since looks like we need the time partition column).

src/chunk_append/chunk_append.c
```diff
@@ -281,3 +291,63 @@ contain_param_exec_walker(Node *node, void *context)
...
+{
+	if (node == NULL)
+		return false;
+
+	if (IsA(node, OpExpr))
```
gayyappan: Would this work if we have a subquery pulled up? Wondering if the jointree walker will handle it correctly.

src/chunk_append/chunk_append.c
```diff
@@ -82,15 +83,15 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht,
...
+				path->cpath.path.pathkeys = subpath->pathkeys;
 			break;
 		default:
 			elog(ERROR, "invalid child of chunk append: %u", nodeTag(subpath));
 			break;
 	}
 
-	Assert(ht->space->num_dimensions <= 2);
-	if (ht->space->num_dimensions == 1)
+	if (path->cpath.path.pathkeys == NULL || ht->space->num_dimensions == 1)
```
cevian: instead of `path->cpath.path.pathkeys == NULL` is `!ordered` more clear? Is that the right logic?

src/chunk_append/chunk_append.c
```diff
@@ -281,3 +291,63 @@ contain_param_exec_walker(Node *node, void *context)
...
+	{
+		OpExpr *op = castNode(OpExpr, node);
+		Var *left_arg = linitial(op->args);
+		Var *right_arg = lsecond(op->args);
+
+		if (op->opno != context->eq_opr || !IsA(left_arg, Var) || !IsA(right_arg, Var))
+			return false;
+
+		/* Is this a join condition referencing our hypertable */
+		if ((left_arg->varno == context->sort_relid &&
```
cevian: nit: this would be easier to read if the checks were in the same order for both sides of the OR. right now it's varno, varattno, varno vs varno, varno, varattno.

test/expected/plan_ordered_append-10.out
```diff
@@ -764,10 +764,10 @@ LIMIT 2;
                ->  Index Scan using _hyper_3_7_chunk_dimension_last_time_idx on _hyper_3_7_chunk
          ->  Materialize
                ->  Append
```
cevian: the order changed here. I guess any order is technically correct here but the new order is more efficient? Can @svenklemm confirm?

cevian: also will this append later become a chunk append too? What's the plan

test/expected/plan_ordered_append-10.out
```diff
@@ -764,10 +764,10 @@ LIMIT 2;
...
-                     ->  Seq Scan on _hyper_4_11_chunk
-                     ->  Seq Scan on _hyper_4_12_chunk
-                     ->  Seq Scan on _hyper_4_13_chunk
                      ->  Seq Scan on _hyper_4_14_chunk
+                     ->  Seq Scan on _hyper_4_13_chunk
+                     ->  Seq Scan on _hyper_4_12_chunk
+                     ->  Seq Scan on _hyper_4_11_chunk
 (15 rows)
 
 -- test INNER JOIN against non-hypertable
```
cevian: maybe add another tests `ORDER BY dimension_last.time ASC` to double check the order changes?

test/expected/plan_ordered_append-10.out
```diff
@@ -1170,11 +1170,11 @@ ORDER BY o1.time;
          ->  Index Scan Backward using _hyper_1_3_chunk_ordered_append_time_idx on _hyper_1_3_chunk o1_3 (actual rows=23043 loops=1)
                Index Cond: ("time" < 'Tue Feb 01 00:00:00 2000 PST'::timestamp with time zone)
    ->  Materialize (actual rows=220327 loops=1)
-         ->  Merge Append (actual rows=73443 loops=1)
-               Sort Key: o2."time"
-               ->  Index Scan Backward using _hyper_1_1_chunk_ordered_append_time_idx on _hyper_1_1_chunk o2 (actual rows=20160 loops=1)
-               ->  Index Scan Backward using _hyper_1_2_chunk_ordered_append_time_idx on _hyper_1_2_chunk o2_1 (actual rows=30240 loops=1)
-               ->  Index Scan Backward using _hyper_1_3_chunk_ordered_append_time_idx on _hyper_1_3_chunk o2_2 (actual rows=23043 loops=1)
+         ->  Custom Scan (ChunkAppend) (actual rows=73443 loops=1)
```
cevian: is the comment above `no exclusion on joined table because its not ChunkAppend node` out of date?

test/expected/plan_ordered_append-10.out
```diff
@@ -1435,5 +1435,154 @@ LEFT OUTER JOIN LATERAL(
...
+         ->  Materialize (actual rows=298 loops=1)
+               ->  Custom Scan (ChunkAppend) (actual rows=100 loops=1)
+                     Hypertable: ordered_append
+                     ->  Index Scan Backward using _hyper_1_1_chunk_ordered_append_time_idx on _hyper_1_1_chunk o2_1 (actual rows=100 loops=1)
+                     ->  Index Scan Backward using _hyper_1_2_chunk_ordered_append_time_idx on _hyper_1_2_chunk o2_2 (never executed)
+                     ->  Index Scan Backward using _hyper_1_3_chunk_ordered_append_time_idx on _hyper_1_3_chunk o2_3 (never executed)
+(16 rows)
+
+-- test JOIN on device_id
+-- should not use ordered append for 2nd hypertable
```
cevian: this comment is weird, the plan below does use an append node right? So it is ordered in a way?

test/expected/plan_ordered_append-10.out
```diff
@@ -1435,5 +1435,154 @@ LEFT OUTER JOIN LATERAL(
...
+               ->  Index Scan Backward using _hyper_1_2_chunk_ordered_append_time_idx on _hyper_1_2_chunk o1_2 (never executed)
+               ->  Index Scan Backward using _hyper_1_3_chunk_ordered_append_time_idx on _hyper_1_3_chunk o1_3 (never executed)
+         ->  Materialize (actual rows=100 loops=1)
+               ->  Custom Scan (ChunkAppend) (actual rows=34 loops=1)
+                     Hypertable: ordered_append
+                     ->  Index Scan Backward using _hyper_1_1_chunk_ordered_append_time_idx on _hyper_1_1_chunk o2_1 (actual rows=34 loops=1)
+                     ->  Index Scan Backward using _hyper_1_2_chunk_ordered_append_time_idx on _hyper_1_2_chunk o2_2 (never executed)
+                     ->  Index Scan Backward using _hyper_1_3_chunk_ordered_append_time_idx on _hyper_1_3_chunk o2_3 (never executed)
+(14 rows)
+
```
cevian: should we add some tests with space partitions?

test/expected/plan_ordered_append-10.out
```diff
@@ -1435,5 +1435,154 @@ LEFT OUTER JOIN LATERAL(
...
+                                                               QUERY PLAN                                                               
+----------------------------------------------------------------------------------------------------------------------------------------
+ Limit (actual rows=100 loops=1)
+   ->  Nested Loop (actual rows=100 loops=1)
+         ->  Custom Scan (ChunkAppend) (actual rows=1 loops=1)
+               Hypertable: ordered_append
+               ->  Index Scan Backward using _hyper_1_1_chunk_ordered_append_time_idx on _hyper_1_1_chunk o1_1 (actual rows=1 loops=1)
+               ->  Index Scan Backward using _hyper_1_2_chunk_ordered_append_time_idx on _hyper_1_2_chunk o1_2 (never executed)
+               ->  Index Scan Backward using _hyper_1_3_chunk_ordered_append_time_idx on _hyper_1_3_chunk o1_3 (never executed)
+         ->  Append (actual rows=100 loops=1)
```
cevian: is an append here correct what happens if this table has duplicate device ids?

test/expected/plan_ordered_append-10.out
svenklemm: it will be chunk append later but it wont be ordered append because of the nested loop join

svenklemm: the reason why the order changes is because expand_hypertable thinks this is gonna become ordered append, but because postgres  is doing nested loop join it never creates the mergeappend plan that would be transformed into ordered append, efficiency-wise it shouldnt matter and once the append becomes a chunk append runtime exclusion should kick in and the nested loop join should only ever target 1 chunk per iteration

test/expected/plan_ordered_append-10.out
svenklemm: done

PR #1253, opened 2019-05-29
Fix telemetry test

cevian:
The current telemetry test fails in ApacheOnly mode because it uses
community functions. The PR separates out the community from the
apache only testing and add a telemetry_community test that only
runs if not in ApacheOnly mode.

Also adjust the travis file to print telemetry output for the
telemetry test run without openssl.

### DIFF
--- a/test/expected/telemetry.out
+++ b/test/expected/telemetry.out
@@ -312,12 +312,6 @@ WHERE key != 'os_name_pretty';
  num_drop_chunks_policies
 (22 rows)
 
-SELECT json_object_field(get_telemetry_report()::json,'num_continuous_aggs');
- json_object_field 
--------------------
- "0"
-(1 row)
-
 -- check telemetry picks up flagged content from metadata
 SELECT json_object_field(get_telemetry_report()::json,'db_metadata');
  json_object_field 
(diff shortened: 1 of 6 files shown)


test/expected/telemetry.out
```diff
@@ -332,33 +326,15 @@ SELECT json_object_field(get_telemetry_report()::json,'instance_metadata');
...
-  min(observation_time)
-FROM
-  device_readings
-GROUP BY bucket;
-SELECT json_object_field(get_telemetry_report()::json,'num_continuous_aggs');
+SELECT json_object_field(get_telemetry_report()::json,'num_hypertables');
  json_object_field 
 -------------------
- "1"
+ "0"
```
svenklemm: should at least have 1 hypertable here, or print num_hypertables, create hypertable and then print it again

test/sql/telemetry_community.sql
```diff
@@ -0,0 +1,28 @@
...
+WITH (timescaledb.continuous) AS
+SELECT
+  time_bucket('1 hour', observation_time) as bucket,
+  min(observation_time)
+FROM
+  device_readings
+GROUP BY bucket;
+
+SELECT json_object_field(get_telemetry_report()::json,'num_continuous_aggs');
+SELECT json_object_field(get_telemetry_report()::json,'num_hypertables')
```
svenklemm: missing semicolon

PR #1260, opened 2019-05-31
Update release notes for 1.3.1

cevian:
Add bugfix PRs and thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,28 @@
 accidentally triggering the load of a previous DB version.**
 
 
-## 1.3.1 (unreleased)
+## 1.3.1 (2019-06-10)
+
+This maintenance release contains bugfixes since the 1.3.0 release.
+We deem it low-to-moderate priority for upgrading.
+
+In particular, the fixes contained in this maintenance release do not address any
+security vulnerabilities, while the only one affecting system stability is related
+to TimescaleDB running on PostgreSQL 11.  More details below.
 
 **Bugfixes**
 * #1220 Fix detecting JOINs for continuous aggs
+* #1221 Fix segfault in VACUUM on PG11
+* #1228 ARM32 Fix: Pass int64 using Int64GetDatum when a Datum is required
 * #1232 Allow Param as time_bucket_gapfill arguments
+* #1236 Stop preventing REFRESH in transaction blocks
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,13 +7,20 @@ accidentally triggering the load of a previous DB version.**
 
 ## 1.3.1 (unreleased)
 
+This release contains bugfixes.
```
RobAtticus: Open question -- should we have something a bit more descriptive here? On some release platforms like Ubuntu/Debian, I have to fill out a bullet point changelog, and usually I put whatever is here as those bullets. I guess I could use just the things listed under Bugfixes

CHANGELOG.md
```diff
@@ -7,13 +7,20 @@ accidentally triggering the load of a previous DB version.**
...
+
 **Bugfixes**
 * #1220 Fix detecting JOINs for continuous aggs
+* #1221 Fix segfault in VACUUM on PG11
+* #1228 Pass int64 using Int64GetDatum when a Datum is required
```
mfreed: This is for ARM32, no?  If so, let's specify that here to give some context?

CHANGELOG.md
```diff
@@ -7,13 +7,20 @@ accidentally triggering the load of a previous DB version.**
...
 * #1232 Allow Param as time_bucket_gapfill arguments
+* #1236 Stop preventing REFRESH in transaction blocks
 
 **Thanks**
 * @od0 for reporting an error with continuous aggs and JOINs
 * @rickbatka for reporting an error when using time_bucket_gapfill in functions
+* @OneMoreSec for reporting the bug with VACUUM
+* @dvdrozdov @od0 @t0k4rtfor reporting the issue with REFRESH in transaction blocks
```
mfreed: Should this be "@t0k4rt for" -- missing a space before "for"?

CHANGELOG.md
mfreed: Yes.  I think one particular relevance is that whether any are security vulnerabilities (I think not), stability bug fixes (maybe just for PG11 given VACUUM?), etc.  In other words, it would be useful if this header could give user a sense of how important would it be for them to upgrade?

dianasaur323: Sounds like a great idea

PR #1261, opened 2019-05-31
Improve ChunkAppend explain

svenklemm:
Change EXPLAIN output to include the order if its an ordered append

### DIFF
--- a/src/chunk_append/explain.c
+++ b/src/chunk_append/explain.c
@@ -1,0 +1,163 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+
+#include <postgres.h>
+#include <catalog/pg_collation.h>
+#include <commands/explain.h>
+#include <nodes/execnodes.h>
+#include <nodes/nodeFuncs.h>
+#include <nodes/parsenodes.h>
+#include <parser/parsetree.h>
+#include <utils/builtins.h>
+#include <utils/lsyscache.h>
+#include <utils/ruleutils.h>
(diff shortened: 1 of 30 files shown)


src/chunk_append/explain.c
```diff
@@ -0,0 +1,163 @@
...
+	typentry = lookup_type_cache(sortcoltype, TYPECACHE_LT_OPR | TYPECACHE_GT_OPR);
+
+	/*
+	 * Print COLLATE if it's not default.  There are some cases where this is
+	 * redundant, eg if expression is a column whose declared collation is
+	 * that collation, but it's hard to distinguish that here.
+	 */
+	if (OidIsValid(collation) && collation != DEFAULT_COLLATION_OID)
+	{
+		char *collname = get_collation_name(collation);
```
cevian: This whole branch is never tested. Please test non-default collations

src/chunk_append/explain.c
```diff
@@ -0,0 +1,163 @@
...
+			elog(ERROR, "cache lookup failed for operator %u", sortOperator);
+		appendStringInfo(buf, " USING %s", opname);
+		/* Determine whether operator would be considered ASC or DESC */
+		(void) get_equality_op_for_ordering_op(sortOperator, &reverse);
+	}
+
+	/* Add NULLS FIRST/LAST only if it wouldn't be default */
+	if (nullsFirst && !reverse)
+	{
+		appendStringInfoString(buf, " NULLS FIRST");
```
cevian: This branch is never hit in tests

src/chunk_append/explain.c
svenklemm: can't be tested in our ci because alpine does not support locales

PR #1273, opened 2019-06-05
Propagate quals to joined hypertables

svenklemm:
When a hypertable is joined on any column with equality references
another tables all constraints that reference the same column will be
propagated to the hypertable.

### DIFF
--- a/test/expected/plan_expand_hypertable-11.out
+++ b/test/expected/plan_expand_hypertable-11.out
@@ -77,7 +77,7 @@ psql:include/plan_expand_hypertable_load.sql:62: NOTICE:  adding not-null constr
 (1 row)
 
 INSERT INTO metrics_timestamp SELECT generate_series('2000-01-01'::timestamp,'2000-02-01'::timestamp,'1d'::interval);
-CREATE TABLE metrics_timestamptz(time timestamptz);
+CREATE TABLE metrics_timestamptz(time timestamptz, device_id int);
 SELECT create_hypertable('metrics_timestamptz','time');
 psql:include/plan_expand_hypertable_load.sql:66: NOTICE:  adding not-null constraint to column "time"
         create_hypertable         
@@ -85,10 +85,12 @@ psql:include/plan_expand_hypertable_load.sql:66: NOTICE:  adding not-null constr
  (6,public,metrics_timestamptz,t)
 (1 row)
 
-INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval);
+INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 1;
+INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 2;
+INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 3;
 CREATE TABLE metrics_date(time date);
 SELECT create_hypertable('metrics_date','time');
-psql:include/plan_expand_hypertable_load.sql:70: NOTICE:  adding not-null constraint to column "time"
+psql:include/plan_expand_hypertable_load.sql:72: NOTICE:  adding not-null constraint to column "time"
      create_hypertable     
 ---------------------------
  (7,public,metrics_date,t)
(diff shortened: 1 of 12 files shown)


test/expected/plan_expand_hypertable-11.out
```diff
@@ -1315,6 +1320,558 @@ psql:include/plan_expand_hypertable_query.sql:165: ERROR:  timestamp out of rang
...
+   ->  Materialize
+         ->  Custom Scan (ChunkAppend) on metrics_timestamptz m2
+               Order: m2."time"
+               ->  Index Only Scan Backward using _hyper_6_160_chunk_metrics_timestamptz_time_idx on _hyper_6_160_chunk m2_1
+                     Index Cond: (("time" < 'Mon Jan 10 00:00:00 2000 PST'::timestamp with time zone) AND ("time" > 'Sat Jan 01 00:00:00 2000 PST'::timestamp with time zone))
+               ->  Index Only Scan Backward using _hyper_6_161_chunk_metrics_timestamptz_time_idx on _hyper_6_161_chunk m2_2
+                     Index Cond: (("time" < 'Mon Jan 10 00:00:00 2000 PST'::timestamp with time zone) AND ("time" > 'Sat Jan 01 00:00:00 2000 PST'::timestamp with time zone))
+(15 rows)
+
+:PREFIX SELECT m1.time FROM metrics_timestamptz m1 LEFT JOIN metrics_timestamptz m2 ON m1.time = m2.time WHERE m1.time > '2000-01-01' AND m2.time < '2000-01-10' ORDER BY m1.time;
```
cevian: I don't see why with a LEFT JOIN it's ok to propagate m2.time < '2000-01-10' to m1. Shouldn't m1 include all values where  m1.time > '2000-01-01' with NULL values on the m2 side for m2  > '2000-01-10'

test/expected/plan_expand_hypertable-11.out
```diff
@@ -1315,6 +1320,558 @@ psql:include/plan_expand_hypertable_query.sql:165: ERROR:  timestamp out of rang
...
+   ->  Materialize
+         ->  Custom Scan (ChunkAppend) on metrics_timestamptz m2
+               Order: m2."time"
+               ->  Index Only Scan Backward using _hyper_6_160_chunk_metrics_timestamptz_time_idx on _hyper_6_160_chunk m2_1
+                     Index Cond: (("time" < 'Mon Jan 10 00:00:00 2000 PST'::timestamp with time zone) AND ("time" > 'Sat Jan 01 00:00:00 2000 PST'::timestamp with time zone))
+               ->  Index Only Scan Backward using _hyper_6_161_chunk_metrics_timestamptz_time_idx on _hyper_6_161_chunk m2_2
+                     Index Cond: (("time" < 'Mon Jan 10 00:00:00 2000 PST'::timestamp with time zone) AND ("time" > 'Sat Jan 01 00:00:00 2000 PST'::timestamp with time zone))
+(15 rows)
+
+:PREFIX SELECT m1.time FROM metrics_timestamptz m1 RIGHT JOIN metrics_timestamptz m2 ON m1.time = m2.time WHERE m1.time > '2000-01-01' AND m2.time < '2000-01-10' ORDER BY m1.time;
```
cevian: same question here I don't think we should propagate here

test/expected/plan_expand_hypertable-11.out
```diff
@@ -1315,6 +1320,558 @@ psql:include/plan_expand_hypertable_query.sql:165: ERROR:  timestamp out of rang
...
+         ->  Custom Scan (ChunkAppend) on metrics_timestamptz m2
+               Order: m2."time"
+               Chunks excluded during startup: 3
+               ->  Index Only Scan Backward using _hyper_6_160_chunk_metrics_timestamptz_time_idx on _hyper_6_160_chunk m2_1
+                     Index Cond: ("time" < ('2000-01-10'::cstring)::timestamp with time zone)
+               ->  Index Only Scan Backward using _hyper_6_161_chunk_metrics_timestamptz_time_idx on _hyper_6_161_chunk m2_2
+                     Index Cond: ("time" < ('2000-01-10'::cstring)::timestamp with time zone)
+(17 rows)
+
+-- test now()
```
cevian: can we also add a test with a volatile function?

test/expected/plan_expand_hypertable-11.out
svenklemm: since the qual is toplevel there is an implied not null

test/expected/plan_expand_hypertable-11.out
svenklemm: since the qual is toplevel there is an implied not null

PR #1275, opened 2019-06-05
Add CMAKE option for default telemetry setting

cevian:
Adds a CMAKE option to turn telemetry off by default by specifying
-DSEND_TELEMETRY_DEFAULT=NO. The default is YES or on.


RobAtticus:
My only thought is that an option to set the specific level might be more versatile/future proof. But also I guess a bit harder on the validation side (aka making sure its not set to an invalid value)?

RobAtticus:
(Not a blocker though)

PR #1284, opened 2019-06-10
Release 1.3.1

cevian:
Commit for the 1.3.1 release.


sql/updates/1.3.0--1.3.1.sql
```diff
@@ -0,0 +1 @@
+--No SQL Changes
```
RobAtticus: Think we usually omit this file if there's no changes.

PR #1287, opened 2019-06-11
Test PRs on the latest stable patch of Postgres

cevian:
Make Travis test PRs and other branches on the latest released
PG patch version for each major version. Cron still tests the earliest
supported patch version (and the ABI tests run by cron cover the
tip of each major version).


cevian:
@svenklemm I don't think that's right. If a PR will break overnight then our ABI tests will break as soon as the next cron job is run after the merge. I'd rather catch problems before merge than after.

svenklemm:
ABI tests only cover merged code so they are not relevant for PRs.
Consider the following scenario:
Time 0: PR gets created, approved and all tests pass
Time 1: new postgres version gets released that breaks stuff in that PR
Time 2: PR can still be merged since the tests are not invalidated
Time 3: master is broken even though PR was all green when merging

PR #1293, opened 2019-06-12
Fix test output and compiler warning

erimatnor:
This PR contains two commits:

1. Fix column name in show_columns
2. Fix compiler warning in `ChunkAppend`

See individual commits for further descriptions.

### DIFF
--- a/src/chunk_append/exec.c
+++ b/src/chunk_append/exec.c
@@ -15,6 +15,7 @@
 #include <parser/parsetree.h>
 #include <rewrite/rewriteManip.h>
 #include <utils/typcache.h>
+#include <utils/memutils.h>
 
 #include "chunk_append/chunk_append.h"
 #include "chunk_append/exec.h"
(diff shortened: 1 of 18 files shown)


src/chunk_append/exec.c
```diff
@@ -15,6 +15,7 @@
 #include <parser/parsetree.h>
 #include <rewrite/rewriteManip.h>
 #include <utils/typcache.h>
+#include <utils/memutils.h>
```
svenklemm: i think executor/executor.h should be included here instead which i'm already doing in another PR

erimatnor: That's what I figured too, but it doesn't work because it appears the issue is that on some PG versions executor doesn't include memutils.h. I can include both, if that make sense, but only memutils.h won't quench the warning.

svenklemm: since i'm adding executor.h in another pr this is fine, which compiler flags do you use since this wasn't picked up by our ci

erimatnor: It happens on MacOS X with Clang so I guess that's why it is not picked up on CI.

PR #1295, opened 2019-06-13
Remove option REQUIRED from CMake find_program

k-rus:
Remove option `REQUIRED` from calls to `find_program` in CMake files, 
since such options doesn't exist and is ignored.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,8 +134,7 @@ find_program(PG_CONFIG pg_config
   PATH_SUFFIXES
   bin
   DOC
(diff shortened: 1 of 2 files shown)


k-rus:
> The code looks good but the commit message doesn't follow our guidelines: https://chris.beams.io/posts/git-commit/

Thanks! Fixed the commit message to be imperative without period and fit 50 chars. Let me know what I am still missing.

k-rus:
Addressed an offline comment from @erimatnor and added body to the commit (after the commit was shorten).

PR #1296, opened 2019-06-13
Expose a function cache for query optimizations

erimatnor:
A number of TimescaleDB query optimizations involve operations on
functions. This refactor exposes a function cache that can be used to
quickly identify important functions and get access to relevant
auxiliary functionality and/or information. In particular, certain
functions apply to some type of (time) bucketing expression, e.g.,
expressions involving our own `time_bucket` function or PostgreSQL's
`date_trunc`.

This change recognizes the importance of time bucketing and uses the
function cache to access custom functionality around time bucketing
used in query optimizations. For example, both grouping estimates for
hash aggregates and sort transforms can be quickly accessed to make
better use of indexes when bucketing on a time column.

This refactor is also done with anticipation that this will be useful
going forward when other types of optimizations are implemented on
time bucketing expressions, or other functions that can benefit from
this cache.


erimatnor:
@svenklemm I discussed this with @cevian and we decided to _not_ do a generic cache for now (baby steps). One reason is that bucketing is "special" as it is works on a time dimensions and is in most cases compatible with "standard" time (e.g. sorting). I guess `time_bucket_gapfill` might fit, but if we start adding functions with different semantics we'd have to add additional filters to identify the semantics of the particular function you have a hit on in the cache. For instance, in the multinode case, we'd need this to identify bucketing functions on partitioning keys to push down to remote nodes.

I think this PR should be considered a step in the right direction. Then we'd have to figure out if we want one monolithic cache for all functions are several caches with different semantics.

src/func_cache.c
```diff
@@ -0,0 +1,366 @@
...
+										   (double) ts_date_trunc_interval_period_approx(interval));
+}
+
+typedef struct FuncEntry
+{
+	Oid funcid;
+	FuncInfo *funcinfo;
+} FuncEntry;
+
+/* Definitions of functions with a custom group estimate.
```
cevian: is this comment stale now?

src/sort_transform.c
```diff
@@ -241,18 +189,23 @@ transform_int_op_const(OpExpr *op)
...
+				return orig_expr;
+
+			return finfo->sort_transform(func);
+		}
 
-		if (strncmp(func_name, "date_trunc", NAMEDATALEN) == 0)
-			return transform_date_trunc(func);
-		if (strncmp(func_name, "time_bucket", NAMEDATALEN) == 0)
-			return transform_time_bucket(func);
 		if (strncmp(func_name, "timestamp", NAMEDATALEN) == 0)
```
cevian: I guess these can now go into the cache? Also OK to leave for a separate PR.

erimatnor: You mean the `timestamp` function? Figured it didn't have the right semantics to be considered a bucketing function, but if we'd like to turn this into a generic cache then I guess we can put it there. I'd rather do that separately though.

PR #1057, opened 2019-02-08
Fix chunk exclusion constraint type inference

Ngalstyan4:
The existing constraint exclusion code creates issues on 32bit processors when constraint given
in a query does not exactly match the type of the table dimension but could be promoted to it.
For example, if hyper_w_space has bigint time type, the following query will pass 10 as
Const variable of type int32 (that has Oid INT4OID).

SELECT * FROM hyper_w_space WHERE time < 10

If this type info is not propagated correctly, later Const variable type is inferred as bigint (int64) and
DatumGetInt64 is used to get its value. In 64 bit systems, DatumGetInt32 and DatumGetInt64 are identical and so there
are no issues, however this makes a difference in 32 bit systems as there 32 bit values are passed by value while
64 bit values are passed by reference.

PR #1299, opened 2019-06-14
Fix Datum to int cast issue

Ngalstyan4:


Ngalstyan4:
`(Const*) value`  itself has an attribute `consttype` which stores its type information. This PR makes sure that we use this type information to accordingly extract value from Datum and not rely on the type information of the column.
When executing a query such as 
`SELECT * FROM hyper WHERE time_bucket(10, time) < 100`
where time column has type `bigint`

the current code incorrectly assumes that the `Const*` type storing the value `100` is of type `bigint` but it does not have to be if the value could fit in a smaller type regardless of the type of the column The fix makes sure we use the type info of `100` and not that of `time` column in this case to extract the value from relevant Datum.

The issue is very similar to [this](https://github.com/timescale/timescaledb/pull/1057)

I will add this info to the commit message. Just made a PR to get it tested on all postgres versions on ARM then remembered that I need to push to prerelease_tests for that.

src/plan_expand_hypertable.c
```diff
@@ -92,6 +92,29 @@ is_time_bucket_function(Expr *node)
...
+		case INT8OID:
+			return DatumGetInt64(cnst->constvalue);
+	}
+
+	ereport(ERROR,
+			(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+			 errmsg("can only use const_datum_get_int with integer types")));
+
+	Assert(false);
+	return 0;
```
svenklemm: this should be pg_unreachable();

PR #1300, opened 2019-06-18
Update get_telemetry_report to expected behavior

spolcyn:
Previously, get_telemetry_report() returns full report even if telemetry is disabled.
Now, when telemetry is disabled, get_telemetry_report() informs user telemetry is 
disabled and how to view the report locally.

### DIFF
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -2,7 +2,6 @@ ALTER TABLE _timescaledb_catalog.telemetry_metadata ADD COLUMN include_in_teleme
 ALTER TABLE _timescaledb_catalog.telemetry_metadata ALTER COLUMN include_in_telemetry DROP DEFAULT;
 ALTER TABLE _timescaledb_catalog.telemetry_metadata RENAME TO metadata;
 ALTER INDEX _timescaledb_catalog.telemetry_metadata_pkey RENAME TO metadata_pkey;
-
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_materialization_invalidation_log(
     materialization_id INTEGER PRIMARY KEY
         REFERENCES _timescaledb_catalog.continuous_agg(mat_hypertable_id)
@@ -16,3 +15,4 @@ CREATE INDEX continuous_aggs_materialization_invalidation_log_idx
     ON _timescaledb_catalog.continuous_aggs_materialization_invalidation_log (materialization_id, lowest_modified_value ASC);
 
 GRANT SELECT ON _timescaledb_catalog.continuous_aggs_materialization_invalidation_log TO PUBLIC;
+DROP FUNCTION IF EXISTS get_telemetry_report();
--- a/sql/version.sql
+++ b/sql/version.sql
@@ -9,5 +9,5 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.get_os_info()
     RETURNS TABLE(sysname TEXT, version TEXT, release TEXT, version_pretty TEXT)
     AS '@MODULE_PATHNAME@', 'ts_get_os_info' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
-CREATE OR REPLACE FUNCTION get_telemetry_report() RETURNS TEXT
-    AS '@MODULE_PATHNAME@', 'ts_get_telemetry_report' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE OR REPLACE FUNCTION get_telemetry_report(always_display_report boolean DEFAULT false) RETURNS TEXT
+    AS '@MODULE_PATHNAME@', 'ts_get_telemetry_report' LANGUAGE C STABLE PARALLEL SAFE;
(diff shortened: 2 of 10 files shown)


src/telemetry/telemetry.c
```diff
@@ -64,6 +64,17 @@
...
+<<<<<<< HEAD
+=======
+#define TS_TELEMETRY_LEVEL "timescaledb.telemetry_level"
+#define TS_TELEMETRY_LEVEL_OFF "off"
+#define TS_TELEMETRY_REPORT_OVERRIDE_ARG "always_display_report := true"
+>>>>>>> 884781c7... Update get_telemetry_report to expected behavior
+=======
+#define TS_TELEMETRY_LEVEL "timescaledb.telemetry_level"
+#define TS_TELEMETRY_LEVEL_OFF "off"
+>>>>>>> 718ac72b... Show error when telemetry requested but set off
```
JLockerman: The diff markers got committed

JLockerman:
cc @dianasaur323 what do you think of the new behavior?

src/telemetry/telemetry.c
```diff
@@ -475,7 +486,21 @@ TS_FUNCTION_INFO_V1(ts_get_telemetry_report);
 Datum
 ts_get_telemetry_report(PG_FUNCTION_ARGS)
 {
-	StringInfo request = build_version_body();
-
-	return CStringGetTextDatum(request->data);
+    StringInfo request;
+
+    /* Show error message if telemetry is disabled and no override argument is passed */
+    if (strcmp(TS_TELEMETRY_LEVEL_OFF, GetConfigOptionByName(TS_TELEMETRY_LEVEL, NULL, false)) == 0)
```
JLockerman: you should just use `ts_telemetry_on` (from `guc.h`)

JLockerman:
this is also going to have slightly weird behavior in that `get_telemetry_report(NULL)` will return `NULL`, I think we want to remove the `STRICT`

dianasaur323:
@JLockerman lgtm! @spolcyn thanks for the fix.

sql/version.sql
```diff
@@ -9,5 +9,5 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.get_os_info()
     RETURNS TABLE(sysname TEXT, version TEXT, release TEXT, version_pretty TEXT)
     AS '@MODULE_PATHNAME@', 'ts_get_os_info' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
 
-CREATE OR REPLACE FUNCTION get_telemetry_report() RETURNS TEXT
-    AS '@MODULE_PATHNAME@', 'ts_get_telemetry_report' LANGUAGE C IMMUTABLE STRICT PARALLEL SAFE;
+CREATE OR REPLACE FUNCTION get_telemetry_report(always_display_report boolean DEFAULT false) RETURNS TEXT
```
svenklemm: if you drop the STRICT this needs to be dropped in the upgrade script latest-dev.sql

src/telemetry/telemetry.c
```diff
@@ -475,7 +476,21 @@ TS_FUNCTION_INFO_V1(ts_get_telemetry_report);
...
 {
-	StringInfo request = build_version_body();
-
-	return CStringGetTextDatum(request->data);
+    StringInfo request;
+
+    /* Show error message if telemetry is disabled and no override argument is passed */
+    if (!ts_telemetry_on())
+    {        
+        if(PG_GETARG_BOOL(0) != true)
```
svenklemm: you should check for NULL args here and number of args passed

svenklemm: should probably add a test for calling it with NULL

sql/version.sql
spolcyn: This line has been added.

sql/version.sql
```diff
@@ -9,5 +9,5 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.get_os_info()
...
+    AS '@MODULE_PATHNAME@', 'ts_get_telemetry_report' LANGUAGE C IMMUTABLE PARALLEL SAFE;
```
svenklemm: IMMUTABLE is wrong here, should be STABLE or VOLATILE

sql/updates/latest-dev.sql
```diff
@@ -2,3 +2,4 @@ ALTER TABLE _timescaledb_catalog.telemetry_metadata ADD COLUMN include_in_teleme
 ALTER TABLE _timescaledb_catalog.telemetry_metadata ALTER COLUMN include_in_telemetry DROP DEFAULT;
 ALTER TABLE _timescaledb_catalog.telemetry_metadata RENAME TO metadata;
 ALTER INDEX _timescaledb_catalog.telemetry_metadata_pkey RENAME TO metadata_pkey;
+DROP FUNCTION get_telemetry_report();
```
svenklemm: we usually do DROP FUNCTION IF EXISTS here

sql/version.sql
spolcyn: Function will be set as STABLE.

PR #1303, opened 2019-06-19
Fix handling of types with custom partitioning

### LINKED ISSUE #1250: The column type is used instead of return type of `time_partitioning_func` in many functions
 - OS: Fedora 30
 - TimescaleDB version (output of `\dx` in `psql`): 1.3.0
 - Installation method: Docker

**Describe the bug**
I created a hypertable with `time_partitioning_func` function to return the time value used in partitioning. However, functions such as `drop_chunks()`,`show_chunks()` and `set_chunk_time_interval()` doesn't work correctly as they are apparently looking into the type of the column rather than return value of `time_partitioning_func` function.

**To Reproduce**
1. Create a sample hypertable using `time_partitioning_func`:

```
CREATE TYPE report AS (last_seen timestamp with time zone, contents text);

CREATE TABLE customer_reports(reports report[] NOT NULL);

CREATE FUNCTION first_report_last_seen (report[])
RETURNS timestamptz
LANGUAGE SQL
IMMUTABLE
AS 'SELECT $1[1].last_seen';

SELECT create_hypertable(
    main_table => 'customer_reports',
    time_column_name => 'reports',
    time_partitioning_func => 'first_report_last_seen'
);
```

2. Now, try to run one of the mentioned functions, and it'll produce an error:
```
# SELECT set_chunk_time_interval('terminated_flows', interval '0.080000 hour');
ERROR:  invalid dimension type: "reports" must be an integer, date or timestamp
```

**Expected behavior**
The functions should properly work in this case.

**Actual behavior**
The error message appears and function doesn't do anything

JLockerman:
In various places, most notably drop_chunks and show_chunks, we
dispatch based on the type of the "time" column of the hypertable, for
things such as determining which interval type to use. With a custom
partition function, this logic is incorrect, as we should instead be
determining this based on the return type of the partitioning function.

This commit changes all relevant access of dimension.column_type to a
new function, ts_dimension_get_partition_type, which has the correct
behavior: it returns the partitioning function's return type, if one
exists, and only otherwise uses the column type. After this commit, all
references to column_type directly should have a comment explaining why
this is appropriate.

fixes #1250

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -1080,10 +1080,10 @@ chunks_typecheck_and_find_all_in_range_limit(Hyperspace *hs, Dimension *time_dim
 
 	if (older_than_type != InvalidOid)
 	{
-		ts_dimension_open_typecheck(older_than_type, time_dim->fd.column_type, caller_name);
+		Oid partitioning_type = ts_dimension_get_partition_type(time_dim);
+		ts_dimension_open_typecheck(older_than_type, partitioning_type, caller_name);
 		if (older_than_type == INTERVALOID)
-			older_than =
-				ts_interval_from_now_to_internal(older_than_datum, time_dim->fd.column_type);
+			older_than = ts_interval_from_now_to_internal(older_than_datum, partitioning_type);
 		else
 			older_than = ts_time_value_to_internal(older_than_datum, older_than_type);
 		end_strategy = BTLessStrategyNumber;
@@ -1091,10 +1091,10 @@ chunks_typecheck_and_find_all_in_range_limit(Hyperspace *hs, Dimension *time_dim
 
 	if (newer_than_type != InvalidOid)
 	{
-		ts_dimension_open_typecheck(newer_than_type, time_dim->fd.column_type, caller_name);
+		Oid partitioning_type = ts_dimension_get_partition_type(time_dim);
+		ts_dimension_open_typecheck(newer_than_type, partitioning_type, caller_name);
 		if (newer_than_type == INTERVALOID)
-			newer_than =
-				ts_interval_from_now_to_internal(newer_than_datum, time_dim->fd.column_type);
+			newer_than = ts_interval_from_now_to_internal(newer_than_datum, partitioning_type);
 		else
 			newer_than = ts_time_value_to_internal(newer_than_datum, newer_than_type);
 		start_strategy = BTGreaterEqualStrategyNumber;
@@ -1267,7 +1267,7 @@ chunk_get_chunks_in_time_range(Oid table_relid, Datum older_than_datum, Datum ne
 		time_dim = hyperspace_get_open_dimension(ht->space, 0);
 
 		if (time_dim_type == InvalidOid)
-			time_dim_type = time_dim->fd.column_type;
+			time_dim_type = ts_dimension_get_partition_type(time_dim);
 
 		/*
 		 * Even though internally all time columns are represented as bigints,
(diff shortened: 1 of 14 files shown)


mfreed:
@JLockerman Can you confirm whether this PR fixes this issue as well?    https://github.com/timescale/timescaledb/issues/1250#issuecomment-503658988

test/sql/chunk_utils.sql
```diff
@@ -347,6 +347,8 @@ BEGIN;
     SELECT * FROM test.show_subtables('drop_chunk_test_date');
 ROLLBACK;
 
+SET timezone TO '+0';
```
svenklemm: why are you setting timezone to 0 i'd rather see tests run with a timezone to make sure this is handled properly

test/sql/chunk_utils.sql
```diff
@@ -369,3 +371,58 @@ SELECT _timescaledb_internal.chunk_id_from_relid(tableoid) FROM chunk_id_from_re
...
+SELECT * FROM test.show_subtables('test_weird_type');
+SELECT show_chunks('test_weird_type', older_than=>'2019/06/06 4:00+0'::TIMESTAMPTZ);
+SELECT show_chunks('test_weird_type', older_than=>'2019/06/06 10:00+0'::TIMESTAMPTZ);
+
+SELECT drop_chunks('2019/06/06 6:00+0'::TIMESTAMPTZ, 'test_weird_type');
+SELECT * FROM test.show_subtables('test_weird_type');
+SELECT show_chunks('test_weird_type', older_than=>'2019/06/06 10:00+0'::TIMESTAMPTZ);
+
+DROP TABLE test_weird_type;
+DROP FUNCTION extract_time;
```
svenklemm: this won't work on 9.6 you need to specify the argument types

src/chunk.c
```diff
@@ -1266,6 +1266,7 @@ chunk_get_chunks_in_time_range(Oid table_relid, Datum older_than_datum, Datum ne
 
 		time_dim = hyperspace_get_open_dimension(ht->space, 0);
 
+		// FIXME should this be partitioning_type instead?
```
svenklemm: this should probably be fixed or removed :)

src/hypertable.c
```diff
@@ -214,6 +214,7 @@ ts_hypertable_get_time_type(PG_FUNCTION_ARGS)
 	time_dimension = hyperspace_get_open_dimension(ht->space, 0);
 	if (time_dimension == NULL)
 		PG_RETURN_NULL();
+	/* This is deliberately column_type not partitioning_type */
```
svenklemm: should probably explain why here

src/chunk.c
JLockerman: @svenklemm @cevian  well should it? (if we don't know, the `FIXME` is better than having us forget about it)

JLockerman:
@mfreed that one won't be fixed, but due to how we handle partitioning functions, I'm a little leery of trying to fix it right now (we need to make sure that it's still a valid input to the function).

JLockerman:
In fact, I'm going to block altering the type of open dimensions with a custom partitioning function.

src/chunk.c
svenklemm: oh i thought you left this as todo for yourself

svenklemm: i think it should be partitioning_type though because column_type makes no sense

JLockerman: @svenklemm mayyybe, but that would be a strictly weaker condition than we have here; it would allow multiple chunks which have different time-column types to be dropped together.

Honestly, I'm inclined to say that allowing `drop_chunks` to affect multiple tables is a miss-feature, and I don't care if it works; pretty much anything will break it anyway.

PR #1309, opened 2019-06-22
Skip runtime exclusion when Param is not partitioning column

svenklemm:
When the column a Param references is not a partitioning column
the constraint is not useful for excluding chunks so we skip
enabling runtime exclusion for those cases.

### DIFF
--- a/src/chunk_append/chunk_append.c
+++ b/src/chunk_append/chunk_append.c
@@ -9,9 +9,11 @@
 #include <optimizer/pathnode.h>
 #include <optimizer/paths.h>
 #include <optimizer/tlist.h>
+#include <optimizer/var.h>
 #include <utils/builtins.h>
 #include <utils/typcache.h>
 
+#include "hypertable.h"
 #include "chunk_append/chunk_append.h"
 #include "chunk_append/planner.h"
 #include "compat.h"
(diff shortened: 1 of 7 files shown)


src/chunk_append/chunk_append.c
```diff
@@ -304,6 +329,19 @@ contain_param_exec_walker(Node *node, void *context)
 	return expression_tree_walker(node, contain_param_exec_walker, context);
 }
 
+static bool
+is_partitioning_column(Hypertable *ht, Index varattno)
```
cevian: Consider putting this into hypertable.c/.h

test/expected/append-10.out
```diff
@@ -1044,6 +1044,30 @@ ORDER BY time DESC;
...
+                                                          QUERY PLAN                                                          
+------------------------------------------------------------------------------------------------------------------------------
+ Nested Loop Left Join (actual rows=5 loops=1)
+   ->  Custom Scan (ChunkAppend) on append_test a1 (actual rows=5 loops=1)
+         Order: a1."time"
+         ->  Index Scan Backward using _hyper_1_1_chunk_append_test_time_idx on _hyper_1_1_chunk a1_1 (actual rows=2 loops=1)
+         ->  Index Scan Backward using _hyper_1_2_chunk_append_test_time_idx on _hyper_1_2_chunk a1_2 (actual rows=2 loops=1)
+         ->  Index Scan Backward using _hyper_1_3_chunk_append_test_time_idx on _hyper_1_3_chunk a1_3 (actual rows=1 loops=1)
+   ->  Limit (actual rows=1 loops=5)
+         ->  Custom Scan (ChunkAppend) on append_test a2 (actual rows=1 loops=5)
```
cevian: How does this show that runtime exclusion is off as opposed to that it just didn't exclude anything

svenklemm: It would have a line like this: `Chunks excluded during runtime: 4`

PR #1315, opened 2019-06-24
Fix potentially lost invalidations in continuous aggs

JLockerman:
We use the InvalidationThreshold as a barrier to ensure that all transactions see an invalidation threshold. Upon transaction commit, all transactions grab a lock on said table, while during the first phase of materialization, the materializer grabs an AccessExclusive lock on the table. This is supposed to ensure that all INSERT/UPDATE/DELETEs are ordered strictly before or after the first phase of materialization, and thus any mutations to newly materialized data will be part the new materialization, or append an invalidation. Unfortunately, the scanner does not hold its locks until the end of the transaction, and since that was the only manner we were taking a lock on the InvalidationLog in mutations, it was not functioning as a barrier. To fix this, we now take an explicit lock.

### DIFF
--- a/tsl/src/continuous_aggs/insert.c
+++ b/tsl/src/continuous_aggs/insert.c
@@ -13,6 +13,7 @@
 #include <miscadmin.h>
 #include <utils/hsearch.h>
 #include <access/tupconvert.h>
+#include <storage/lmgr.h>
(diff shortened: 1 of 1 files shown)


JLockerman:
cc @cevian this should be the final one for 1.3.2

PR #1317, opened 2019-06-24
Support time bucket functions in Ordered Append

svenklemm:
The initial implementation for Ordered Append required the ORDER BY
clause expression to match the time partitioning column.
This patch loosenes that restriction and will apply the Ordered
Append optimizations for queries with ORDER BY time_bucket as well.

### DIFF
--- a/src/chunk_append/chunk_append.c
+++ b/src/chunk_append/chunk_append.c
@@ -17,6 +17,7 @@
 #include "chunk_append/chunk_append.h"
 #include "chunk_append/planner.h"
 #include "compat.h"
+#include "func_cache.h"
 #include "guc.h"
 
 static bool contain_param_exec(Node *node);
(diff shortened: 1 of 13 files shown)


src/chunk_append/chunk_append.c
```diff
@@ -261,11 +262,37 @@ ts_ordered_append_should_optimize(PlannerInfo *root, RelOptInfo *rel, Hypertable
...
 	Assert(root->parse->sortClause != NIL);
 
-	/* we only support direct column references here */
-	if (!IsA(tle->expr, Var))
+	if (IsA(tle->expr, Var))
+	{
+		/* direct column reference */
+		sort_var = castNode(Var, tle->expr);
+	}
+	else if (IsA(tle->expr, FuncExpr) && list_length(root->parse->sortClause) == 1)
```
cevian: why `list_length(root->parse->sortClause) == 1` here?

svenklemm: added an example why this is required

PR #1321, opened 2019-06-26
Fix SIGHUP handling in scheduler and launcher

cevian:
SIGHUP's can be dropped in between background worker startup and
the call into the background worker entrypoint. Namely, Postgres
calls `pqsignal(SIGHUP, SIG_IGN)` inside of `StartBackgroundWorker`.
Thus, SIGHUPs will be ignored before the call to the entrypoint.
This creates a possible race condition where a config file change
is not correctly processed (and is instead ignored). We prevent
this by always processing the config file after we set our own
signal handler.

We also fix the tests here in two ways:
1) We disable background workers in this test (or, rather, delete the
line that starts them back up).

2) We put in a new mock timer option that calls the standard timer wait.
This allows us to test proper latch processing in the SIGHUP case.

We believe that this resolves some flakiness in our tests as well.


src/bgw/scheduler.c
```diff
@@ -725,6 +725,10 @@ ts_bgw_scheduler_register_signal_handlers(void)
 	 */
 	pqsignal(SIGTERM, handle_sigterm);
 	pqsignal(SIGHUP, handle_sighup);
+
+	/* Some SIGHUPS may already have been dropped, so we must load the file here */
+	got_SIGHUP = false;
```
JLockerman: why do we set `got_SIGHUP = false`?

JLockerman: This is the usual sigatomic handling.

PR #1323, opened 2019-06-27
Implement adding a notion of now for hypertables with integer time columns

Ngalstyan4:
This PR implements functionality for users to give a custom definition of now() for integer open dimension typed hypertables. It also implements the necessary changes in drop chunks policy to use this now() function when handling drop_chunk background worker tasks.

Below is an outline for the PR.

- [x] Add columns to the dimension catalog for the schema/function name of the current_time function.
  - [x] a213746 regression test output changes
  - [x] cb78532 catalog table definition changes

- [x] Modify the current drop_chunks_policy table to use the internal representation rather than an actual interval. 
   - [x] 65874bb  regression test output changes
   - [x] 62d3131  catalog table definition changes

- [x] da4dc1a  Implement set_integer_now_func logic 

- [x] Teach drop_chunks policy as well as any others that use now() to use the current_time function instead for integer hypertables.
  - [x] 91f6fb4 add_drop_chunks_policy
  - (postponed) Reorder chunks

- [x] Modify update scripts accordingly
- [x] Add more tests
  - [ ] More thoroughly test background job failure cases and get appropriate regression test output
  - [x] Test functional cases when integer_now function and partitioning function are both set on a hypertable

I will fixup the commits later but thought detailed breakdown may be useful for reviewing.

### DIFF
--- a/sql/bgw_scheduler.sql
+++ b/sql/bgw_scheduler.sql
@@ -21,7 +21,7 @@ INSERT INTO _timescaledb_config.bgw_job (id, application_name, job_type, schedul
 (1, 'Telemetry Reporter', 'telemetry_and_version_check_if_enabled', INTERVAL '24h', INTERVAL '100s', -1, INTERVAL '1h')
 ON CONFLICT (id) DO NOTHING;
 
-CREATE OR REPLACE FUNCTION add_drop_chunks_policy(hypertable REGCLASS, older_than INTERVAL, cascade BOOL = FALSE, if_not_exists BOOL = false, cascade_to_materializations BOOL = false)
+CREATE OR REPLACE FUNCTION add_drop_chunks_policy(hypertable REGCLASS, older_than "any", cascade BOOL = FALSE, if_not_exists BOOL = false, cascade_to_materializations BOOL = false)
 RETURNS INTEGER
 AS '@MODULE_PATHNAME@', 'ts_add_drop_chunks_policy'
 LANGUAGE C VOLATILE STRICT;
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -78,6 +78,8 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.dimension (
     partitioning_func           NAME     NULL,
     -- open dimensions (e.g., time)
     interval_length             BIGINT   NULL CHECK(interval_length IS NULL OR interval_length > 0),
+    integer_now_func_schema     NAME     NULL,
+    integer_now_func            NAME     NULL,
     CHECK (
         (partitioning_func_schema IS NULL AND partitioning_func IS NULL) OR
         (partitioning_func_schema IS NOT NULL AND partitioning_func IS NOT NULL)
@@ -85,6 +87,10 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.dimension (
     CHECK (
         (num_slices IS NULL AND interval_length IS NOT NULL) OR
         (num_slices IS NOT NULL AND interval_length IS NULL)
+    ),
+    CHECK (
+        (integer_now_func_schema IS NULL AND integer_now_func IS NULL) OR
+        (integer_now_func_schema IS NOT NULL AND integer_now_func IS NOT NULL)
     ),
     UNIQUE (hypertable_id, column_name)
 );
@@ -195,11 +201,12 @@ CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_reorder (
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_policy_reorder', '');
 
 CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_drop_chunks (
-    job_id          		INTEGER     PRIMARY KEY REFERENCES _timescaledb_config.bgw_job(id) ON DELETE CASCADE,
-    hypertable_id   		INTEGER     UNIQUE NOT NULL REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
-	older_than				INTERVAL    NOT NULL,
-	cascade					BOOLEAN,
-    cascade_to_materializations BOOLEAN
+    job_id          		    INTEGER                 PRIMARY KEY REFERENCES _timescaledb_config.bgw_job(id) ON DELETE CASCADE,
+    hypertable_id   		    INTEGER     UNIQUE      NOT NULL REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
+    older_than	    _timescaledb_catalog.ts_interval    NOT NULL,
+	cascade					    BOOLEAN                 NOT NULL,
+    cascade_to_materializations BOOLEAN                 NOT NULL,
+    CONSTRAINT valid_older_than CHECK(_timescaledb_internal.valid_ts_interval(older_than))
 );
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_policy_drop_chunks', '');
 
--- a/sql/pre_install/types.sql
+++ b/sql/pre_install/types.sql
@@ -1,0 +1,22 @@
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+
+-- Postgres has special types such as timestamp, date, timestamptz, etc. to
+-- represent logical time values. On top of these, TimescaleDB allows integers
+-- to represent logical time values. Postgres INTERVAL types are suited only for
+-- logical time types in postgres. The type defined below is an INTERVAL equivalent
+-- for TimescaleDB and enables to represent time intervals for both postgres time
+-- type valued and integer valued time columns.
+CREATE TYPE _timescaledb_catalog.ts_interval AS (
+    is_time_interval        BOOLEAN,
+    time_interval		    INTERVAL,
+    integer_interval        BIGINT
+    );
+
+-- PostgreSQL composite types do not support constraint checks. That is why any table having a ts_interval column must use the following
+-- function for constraint validation.
+-- This function needs to be defined before executing pre_install/tables.sql because it is used as
+-- validation constraint for columns of type ts_interval.
+CREATE OR REPLACE FUNCTION _timescaledb_internal.valid_ts_interval(invl _timescaledb_catalog.ts_interval)
+RETURNS BOOLEAN AS '@MODULE_PATHNAME@', 'ts_valid_ts_interval' LANGUAGE C VOLATILE STRICT;
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -1,1 +1,60 @@
 DROP FUNCTION IF EXISTS drop_chunks("any",name,name,boolean,"any",boolean,boolean);
+DROP FUNCTION IF EXISTS add_drop_chunks_policy(REGCLASS,INTERVAL,BOOL,BOOL,BOOL);
+
+ALTER TABLE _timescaledb_catalog.dimension
+ADD COLUMN integer_now_func_schema     NAME     NULL;
+
+ALTER TABLE _timescaledb_catalog.dimension
+ADD COLUMN integer_now_func            NAME     NULL;
+
+ALTER TABLE _timescaledb_catalog.dimension
+ADD CONSTRAINT dimension_check2
+CHECK (
+        (integer_now_func_schema IS NULL AND integer_now_func IS NULL) OR
+        (integer_now_func_schema IS NOT NULL AND integer_now_func IS NOT NULL)
+);
+-- ----------------------
+CREATE TYPE _timescaledb_catalog.ts_interval AS (
+    is_time_interval        BOOLEAN,
+    time_interval		    INTERVAL,
+    integer_interval        BIGINT
+    );
+
+-- q -- todo:: this is probably necessary if we keep the validation constraint in the table definition.
+CREATE OR REPLACE FUNCTION _timescaledb_internal.valid_ts_interval(invl _timescaledb_catalog.ts_interval)
+RETURNS BOOLEAN AS '@MODULE_PATHNAME@', 'ts_valid_ts_interval' LANGUAGE C VOLATILE STRICT;
+
+DROP VIEW IF EXISTS timescaledb_information.drop_chunks_policies;
+DROP VIEW IF EXISTS timescaledb_information.policy_stats;
+
+CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_drop_chunks_tmp (
+    job_id          		    INTEGER                 PRIMARY KEY REFERENCES _timescaledb_config.bgw_job(id) ON DELETE CASCADE,
+    hypertable_id   		    INTEGER     UNIQUE      NOT NULL REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
+    older_than	    _timescaledb_catalog.ts_interval    NOT NULL,
+	cascade					    BOOLEAN                 NOT NULL,
+    cascade_to_materializations BOOLEAN                 NOT NULL,
+    CONSTRAINT valid_older_than CHECK(_timescaledb_internal.valid_ts_interval(older_than))
+);
+
+INSERT INTO _timescaledb_config.bgw_policy_drop_chunks_tmp
+(SELECT job_id, hypertable_id, ROW('t',older_than,NULL)::_timescaledb_catalog.ts_interval  as older_than, cascade, cascade_to_materializations
+FROM _timescaledb_config.bgw_policy_drop_chunks);
+
+ALTER EXTENSION timescaledb DROP TABLE _timescaledb_config.bgw_policy_drop_chunks;
+DROP TABLE _timescaledb_config.bgw_policy_drop_chunks;
+
+CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_drop_chunks (
+    job_id          		    INTEGER                 PRIMARY KEY REFERENCES _timescaledb_config.bgw_job(id) ON DELETE CASCADE,
+    hypertable_id   		    INTEGER     UNIQUE      NOT NULL REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
+    older_than	    _timescaledb_catalog.ts_interval    NOT NULL,
+	cascade					    BOOLEAN                 NOT NULL,
+    cascade_to_materializations BOOLEAN                 NOT NULL,
+    CONSTRAINT valid_older_than CHECK(_timescaledb_internal.valid_ts_interval(older_than))
+);
+
+INSERT INTO _timescaledb_config.bgw_policy_drop_chunks
+(SELECT * FROM _timescaledb_config.bgw_policy_drop_chunks_tmp);
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_policy_drop_chunks', '');
+DROP TABLE _timescaledb_config.bgw_policy_drop_chunks_tmp;
+GRANT SELECT ON _timescaledb_config.bgw_policy_drop_chunks TO PUBLIC;
--- a/src/bgw_policy/drop_chunks.c
+++ b/src/bgw_policy/drop_chunks.c
@@ -9,6 +9,7 @@
 #include <utils/timestamp.h>
 #include <utils/lsyscache.h>
 #include <utils/syscache.h>
+#include <funcapi.h>
 
 #include "catalog.h"
 #include "policy.h"
@@ -18,16 +19,42 @@
 #include "hypertable.h"
 #include "bgw/job.h"
 #include "scan_iterator.h"
+#include "interval.h"
 
 static ScanTupleResult
 bgw_policy_drop_chunks_tuple_found(TupleInfo *ti, void *const data)
 {
 	BgwPolicyDropChunks **policy = data;
+	bool nulls[Natts_bgw_policy_drop_chunks];
+	Datum values[Natts_bgw_policy_drop_chunks];
 
-	*policy = STRUCT_FROM_TUPLE(ti->tuple,
-								ti->mctx,
-								BgwPolicyDropChunks,
-								FormData_bgw_policy_drop_chunks);
+	/* cannot use STRUCT_FROM_TUPLE because bgw_policy_drop_chunks
+	 * includes a column(older_than) that has a composite data type
+	 */
+	heap_deform_tuple(ti->tuple, ti->desc, values, nulls);
+
+	*policy = MemoryContextAllocZero(ti->mctx, sizeof(BgwPolicyDropChunks));
+	Assert(!nulls[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_job_id)]);
+	(*policy)->fd.job_id =
+		DatumGetInt32(values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_job_id)]);
+
+	Assert(!nulls[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_hypertable_id)]);
+	(*policy)->fd.hypertable_id =
+		DatumGetInt32(values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_hypertable_id)]);
+
+	Assert(!nulls[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_older_than)]);
+
+	(*policy)->fd.older_than = *ts_interval_from_tuple(
+		values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_older_than)]);
+
+	Assert(!nulls[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade)]);
+	(*policy)->fd.cascade =
+		DatumGetBool(values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade)]);
+
+	Assert(
+		!nulls[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade_to_materializations)]);
+	(*policy)->fd.cascade_to_materializations = DatumGetBool(
+		values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade_to_materializations)]);
 
 	return SCAN_CONTINUE;
 }
@@ -113,6 +140,7 @@ ts_bgw_policy_drop_chunks_insert_with_relation(Relation rel, BgwPolicyDropChunks
 	CatalogSecurityContext sec_ctx;
 	Datum values[Natts_bgw_policy_drop_chunks];
 	bool nulls[Natts_bgw_policy_drop_chunks] = { false };
+	HeapTuple ht_older_than;
 
 	tupdesc = RelationGetDescr(rel);
 
@@ -120,8 +148,12 @@ ts_bgw_policy_drop_chunks_insert_with_relation(Relation rel, BgwPolicyDropChunks
 		Int32GetDatum(policy->fd.job_id);
 	values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_hypertable_id)] =
 		Int32GetDatum(policy->fd.hypertable_id);
+
+	ht_older_than = ts_interval_form_heaptuple(&policy->fd.older_than);
+
 	values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_older_than)] =
-		IntervalPGetDatum(&policy->fd.older_than);
+		HeapTupleGetDatum(ht_older_than);
+
 	values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade)] =
 		BoolGetDatum(policy->fd.cascade);
 	values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade_to_materializations)] =
@@ -130,6 +162,7 @@ ts_bgw_policy_drop_chunks_insert_with_relation(Relation rel, BgwPolicyDropChunks
 	ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
 	ts_catalog_insert_values(rel, tupdesc, values, nulls);
 	ts_catalog_restore_user(&sec_ctx);
+	heap_freetuple(ht_older_than);
 }
 
 void
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -172,6 +172,8 @@ enum Anum_dimension
 	Anum_dimension_partitioning_func_schema,
 	Anum_dimension_partitioning_func,
 	Anum_dimension_interval_length,
+	Anum_dimension_integer_now_func_schema,
+	Anum_dimension_integer_now_func,
 	_Anum_dimension_max,
 };
 
@@ -190,6 +192,8 @@ typedef struct FormData_dimension
 	NameData partitioning_func;
 	/* open (time) columns */
 	int64 interval_length;
+	NameData integer_now_func_schema;
+	NameData integer_now_func;
 } FormData_dimension;
 
 typedef FormData_dimension *Form_dimension;
@@ -691,6 +695,23 @@ typedef struct FormData_bgw_policy_reorder_hypertable_id_idx
  *
  ******************************************/
 #define BGW_POLICY_DROP_CHUNKS_TABLE_NAME "bgw_policy_drop_chunks"
+
+typedef enum Anum_ts_interval
+{
+	Anum_is_time_interval = 1,
+	Anum_time_interval,
+	Anum_integer_interval,
+	_Anum_ts_interval_max
+} Anum_ts_interval;
+
+#define Natts_ts_interval (_Anum_ts_interval_max - 1)
+typedef struct FormData_ts_interval
+{
+	bool is_time_interval;
+	Interval time_interval;
+	int64 integer_interval;
+} FormData_ts_interval;
+
 typedef enum Anum_bgw_policy_drop_chunks
 {
 	Anum_bgw_policy_drop_chunks_job_id = 1,
@@ -707,7 +728,7 @@ typedef struct FormData_bgw_policy_drop_chunks
 {
 	int32 job_id;
 	int32 hypertable_id;
-	Interval older_than;
+	FormData_ts_interval older_than;
 	bool cascade;
 	bool cascade_to_materializations;
 } FormData_bgw_policy_drop_chunks;
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -50,6 +50,7 @@
 #include "trigger.h"
 #include "compat.h"
 #include "utils.h"
+#include "interval.h"
 #include "hypertable_cache.h"
 #include "cache.h"
 #include "bgw_policy/chunk_stats.h"
@@ -1055,6 +1056,8 @@ chunks_typecheck_and_find_all_in_range_limit(Hyperspace *hs, Dimension *time_dim
 											 uint64 *num_found)
 {
 	ChunkScanCtx *chunk_ctx = NULL;
+	FormData_ts_interval *invl;
+
 	int64 older_than = -1;
 	int64 newer_than = -1;
 
@@ -1071,10 +1074,22 @@ chunks_typecheck_and_find_all_in_range_limit(Hyperspace *hs, Dimension *time_dim
 	{
 		Oid partitioning_type = ts_dimension_get_partition_type(time_dim);
 		ts_dimension_open_typecheck(older_than_type, partitioning_type, caller_name);
+
 		if (older_than_type == INTERVALOID)
-			older_than = ts_interval_from_now_to_internal(older_than_datum, partitioning_type);
+		{
+			invl = ts_interval_from_sql_input(hs->main_table_relid,
+											  older_than_datum,
+											  older_than_type,
+											  "older_than",
+											  caller_name);
+			older_than = ts_time_value_to_internal(ts_interval_subtract_from_now(invl, time_dim),
+												   partitioning_type);
+		}
 		else
+		{
 			older_than = ts_time_value_to_internal(older_than_datum, older_than_type);
+		}
+
 		end_strategy = BTLessStrategyNumber;
 	}
 
@@ -1082,10 +1097,22 @@ chunks_typecheck_and_find_all_in_range_limit(Hyperspace *hs, Dimension *time_dim
 	{
 		Oid partitioning_type = ts_dimension_get_partition_type(time_dim);
 		ts_dimension_open_typecheck(newer_than_type, partitioning_type, caller_name);
+
 		if (newer_than_type == INTERVALOID)
-			newer_than = ts_interval_from_now_to_internal(newer_than_datum, partitioning_type);
+		{
+			invl = ts_interval_from_sql_input(hs->main_table_relid,
+											  newer_than_datum,
+											  newer_than_type,
+											  "newer_than",
+											  caller_name);
+			newer_than = ts_time_value_to_internal(ts_interval_subtract_from_now(invl, time_dim),
+												   partitioning_type);
+		}
 		else
+		{
 			newer_than = ts_time_value_to_internal(newer_than_datum, newer_than_type);
+		}
+
 		start_strategy = BTGreaterEqualStrategyNumber;
 	}
 
--- a/src/cross_module_fn.c
+++ b/src/cross_module_fn.c
@@ -13,6 +13,8 @@
 #include "guc.h"
 #include "bgw/job.h"
 
+TS_FUNCTION_INFO_V1(ts_set_integer_now_func);
+TS_FUNCTION_INFO_V1(ts_valid_ts_interval);
 TS_FUNCTION_INFO_V1(ts_add_drop_chunks_policy);
 TS_FUNCTION_INFO_V1(ts_add_reorder_policy);
 TS_FUNCTION_INFO_V1(ts_remove_drop_chunks_policy);
@@ -23,6 +25,18 @@ TS_FUNCTION_INFO_V1(ts_partialize_agg);
 TS_FUNCTION_INFO_V1(ts_finalize_agg_sfunc);
 TS_FUNCTION_INFO_V1(ts_finalize_agg_ffunc);
 TS_FUNCTION_INFO_V1(continuous_agg_invalidation_trigger);
+
+Datum
+ts_set_integer_now_func(PG_FUNCTION_ARGS)
+{
+	PG_RETURN_DATUM(ts_cm_functions->set_integer_now_func(fcinfo));
+}
+
+Datum
+ts_valid_ts_interval(PG_FUNCTION_ARGS)
+{
+	PG_RETURN_DATUM(ts_cm_functions->valid_ts_interval(fcinfo));
+}
 
 Datum
 ts_add_drop_chunks_policy(PG_FUNCTION_ARGS)
@@ -222,6 +236,8 @@ TSDLLEXPORT CrossModuleFunctions ts_cm_functions_default = {
 	.add_tsl_license_info_telemetry = add_telemetry_default,
 	.bgw_policy_job_execute = bgw_policy_job_execute_default_fn,
 	.continuous_agg_materialize = cagg_materialize_default_fn,
+	.set_integer_now_func = error_no_default_fn_pg_enterprise,
+	.valid_ts_interval = error_no_default_fn_pg_enterprise,
 	.add_drop_chunks_policy = error_no_default_fn_pg_enterprise,
 	.add_reorder_policy = error_no_default_fn_pg_enterprise,
 	.remove_drop_chunks_policy = error_no_default_fn_pg_enterprise,
--- a/src/dimension.c
+++ b/src/dimension.c
@@ -165,6 +165,16 @@ dimension_fill_in_from_tuple(Dimension *d, TupleInfo *ti, Oid main_table_relid)
 		MemoryContextSwitchTo(old);
 	}
 
+	if (!isnull[Anum_dimension_integer_now_func_schema - 1] &&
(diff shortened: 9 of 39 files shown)


src/dimension.c
```diff
@@ -1337,6 +1380,7 @@ static ScanTupleResult
 dimension_rename_schema_name(TupleInfo *ti, void *data)
 {
 	HeapTuple tuple = heap_copytuple(ti->tuple);
+	// TODO:: Q:: now Dimension table may have null values so this may not be valid??
 	FormData_dimension *dimension = (FormData_dimension *) GETSTRUCT(tuple);
```
Ngalstyan4: now Dimension table may have null values so is this still valid or should I change it to `heap_getattr` or `heap_deform_tuple` ?

src/dimension.h
```diff
@@ -97,6 +97,10 @@ typedef struct DimensionInfo
 	int64 interval;
 	int32 num_slices;
 	regproc partitioning_func;
+	// todo:: q:: integer_now_func not needed unless we add an argument to create_hypertable
+	// currently it is nevery populated or used
+	// populating it may make code simpler in some places
+	regproc integer_now_func;
```
Ngalstyan4: [DISCUSS]
- [x] should we add a flag in `create_hypertable` to set `integer_now_func`?
if not, this field may be redundant but maintaining it may simplify related code down the road.

src/dimension.h
```diff
@@ -167,6 +171,12 @@ extern void ts_dimension_info_validate(DimensionInfo *info);
 extern void ts_dimension_add_from_info(DimensionInfo *info);
 extern void ts_dimensions_rename_schema_name(char *oldname, char *newname);
 
+/* TODO:: Q:: this function should probably need to be refactored if we agree it should be exposed
+ */
+extern TSDLLEXPORT void dimension_update(FunctionCallInfo fcinfo, Oid table_relid, Name dimname,
```
Ngalstyan4: This used to be a static function in `dimension.c`. Its first `fcinfo` arg is used iff `Datum *interval` is not null and it is used to retrieve interval type. Should the interval type be passed directly instead?

The function is otherwise quite generic and makes it easy to update certain dimension table fields so may be useful to expose it.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,6 +58,137 @@
 										  Int32GetDatum(0),                                        \
 										  Float8GetDatum(0)))
 
+/* TODO:: Q:: These should probably be in its own file? */
+Datum
+ts_integer_from_now_func_get_datum(int64 interval, Oid time_dim_type, Oid now_func)
```
Ngalstyan4: Should these be in a separate file or is here good?

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,6 +58,137 @@
...
+				 errmsg("cache lookup failed for function %u", now_func_oid)));
+
+	now_func = (Form_pg_proc) GETSTRUCT(tuple);
+
+	if ((now_func->provolatile != PROVOLATILE_IMMUTABLE &&
+		 now_func->provolatile != PROVOLATILE_STABLE) ||
+		now_func->pronargs != 0)
+		// q:: todo:: do I have to ReleaseSysCache(tuple); if the next step is an error?
+		// it is done in some places and omited in others.
+		ereport(ERROR,
```
Ngalstyan4: Is `ReleaseSysCache(tuple)` required before an error step? It is done in some cases but skipped in some others.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,6 +58,137 @@
...
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("integer_now_func must take no arguments and it must be STABLE or "
+						"IMMUTABLE")));
+
+	/*TODO:: Q:: do we want to autocast among INT?OID ? types*/
+	if (now_func->prorettype != open_dim_type)
```
Ngalstyan4: Currently, a integer_now function that returns `BIGINT` will not work for a hypertable with time dimension of type `INT`. Is this ok? Do we want some level of autocasting? Will definitely make life harder in implementation side :))

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -91,11 +223,70 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
...
+			if (!IS_INTEGER_TYPE(partitioning_type))
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg("integer older_than argument can only be used with hypertables "
+								"that have integer time dimensions"))); // <<-- time columns? open
+																		// dimensions? TODO::
+
+			/* TODO:: q:: awkward way to check if a NameStr col is null or not but did not find a
+			 * better API */
+			if ('\0' == *NameStr(open_dim->fd.integer_now_func) ||
```
Ngalstyan4: Is there a better way to know if a Name field is set or not?

tsl/src/bgw_policy/job.c
```diff
@@ -159,21 +178,67 @@ execute_drop_chunks_policy(int32 job_id)
...
+					 errmsg("could not set integer_now function because \"%s\" is not a hypertable",
+							get_rel_name(table_relid))));
+
+		open_dim = hyperspace_get_open_dimension(hypertable->space, 0);
+
+		Assert(IS_INTEGER_TYPE(ts_dimension_get_partition_type(open_dim)));
+
+		// todo:: perhaps use FuncnameGetCandidates instead?
+		// Q:: this uniquely identifies a function right?
+		now_func = lookup_proc_filtered(NameStr(open_dim->fd.integer_now_func_schema),
```
Ngalstyan4: This uniquely identifies a function right?
Should I use `FuncnameGetCandidates` instead of exposing `lookup_proc_filtered` ?

`lookup_proc_filtered` is pretty generic and I think it is worth exposing it.

src/partitioning.c
```diff
@@ -47,15 +47,14 @@
...
 /*
  * Find a partitioning function with a given schema and name.
  *
  * The caller can optionally pass a filter function and a type of the argument
  * that the partitioning function should take.
  */
-static regproc
+// Q:: should this be moved to utils.c since it is not specific to partitioning functions ?
+regproc
 lookup_proc_filtered(const char *schema, const char *funcname, Oid *rettype, proc_filter filter,
```
Ngalstyan4: If we decide to expose it, it should probably be moved to `utils.c` since it will no longer be partitioning func specific

src/utils.c
```diff
@@ -513,13 +513,16 @@ ts_type_is_int8_binary_compatible(Oid sourcetype)
...
+ * Also note that the function assumes no NULLs in the tuple.
  */
 void *
 ts_create_struct_from_tuple(HeapTuple tuple, MemoryContext mctx, size_t alloc_size,
 							size_t copy_size)
 {
 	void *struct_ptr = MemoryContextAllocZero(mctx, alloc_size);
 
+	/* Make sure the function is not used when the tuple contains NULLs */
+	Assert(copy_size == tuple->t_len - tuple->t_data->t_hoff);
```
Ngalstyan4: This is not related to the rest of the PR and I am happy to remove it. However, I initially assumed I could use the function even when there are NULL columns and took waay to long to find and fix the bug. A check like this would save a lot of time.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -91,11 +223,70 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
...
+
+	switch (older_than_type)
+	{
+		case INTERVALOID:
+			if (IS_INTEGER_TYPE(partitioning_type))
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg(
+							 "interval older_than argument can only be used with hypertables "
+							 "that have non-integer time dimensions"))); // <<-- time columns? open
```
Ngalstyan4: time dimensions? time columns? open dimensions?

src/bgw_policy/drop_chunks.c
```diff
@@ -23,12 +23,52 @@ static ScanTupleResult
...
+		DatumGetBool(values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade)]);
+
+	Assert(
+		!nulls[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade_to_materializations)]);
+	(*policy)->fd.cascade_to_materializations = DatumGetBool(
+		values[AttrNumberGetAttrOffset(Anum_bgw_policy_drop_chunks_cascade_to_materializations)]);
+
+	// Q:: why scan continue? Is it not wrong to have multiple? And even if there are multiple, this
+	// would just take the last one?
 	return SCAN_CONTINUE;
```
Ngalstyan4: Should this be SCAN_DONE?

src/dimension.c
```diff
@@ -902,9 +929,11 @@ dimension_add_not_null_on_column(Oid table_relid, char *colname)
 	AlterTableInternal(table_relid, list_make1(&cmd), false);
 }
 
-static void
+/*Q:: why are the last args passed by reference? is it to make validation easier?
+ * Q:: is it really necessary to pass all of fcinfo thorugh? */
+void
 dimension_update(FunctionCallInfo fcinfo, Oid table_relid, Name dimname, DimensionType dimtype,
-				 Datum *interval, int16 *num_slices)
+				 Datum *interval, int16 *num_slices, Oid *integer_now_func)
```
Ngalstyan4: Why are the last args passed by reference?

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,6 +58,137 @@
...
+ts_integer_from_now_func_get_datum(int64 interval, Oid time_dim_type, Oid now_func)
+{
+	Datum now;
+
+	AssertArg(IS_INTEGER_TYPE(time_dim_type));
+
+	ts_integer_now_func_validate(now_func, time_dim_type);
+
+	// bgw luncher starts a transactions so here I can assume I am always inside a transaction
+	// right? no need for StartTransactionCommand();
```
Ngalstyan4: bgw luncher starts a transaction so here I can assume I am always inside a transaction right? so no need for StartTransactionCommand();

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,6 +58,137 @@
...
+	/* functions in indexes may want a snapshot set */
+	// Q:: Why is there not one already?
+	PushActiveSnapshot(GetTransactionSnapshot());
```
Ngalstyan4: Why is there not a Snapshot already? This is required to be able to call a user defined function from here.

src/utils.c
cevian: This is great

tsl/src/bgw_policy/drop_chunks_api.c
cevian: this language is fine as is

tsl/src/bgw_policy/drop_chunks_api.c
cevian: you should use strlen. It's fine to set this to an empty string if NULL;

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -91,11 +223,70 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
...
+								"that have integer time dimensions"))); // <<-- time columns? open
+																		// dimensions? TODO::
+
+			/* TODO:: q:: awkward way to check if a NameStr col is null or not but did not find a
+			 * better API */
+			if ('\0' == *NameStr(open_dim->fd.integer_now_func) ||
+				'\0' == *NameStr(open_dim->fd.integer_now_func_schema))
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg("integer older_than argument can only be used with hypertables "
```
cevian: too long for errmsg->move into errhint.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -91,11 +223,70 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
...
+	open_dim = hyperspace_get_open_dimension(hypertable->space, 0);
+	partitioning_type = ts_dimension_get_partition_type(open_dim);
+
+	switch (older_than_type)
+	{
+		case INTERVALOID:
+			if (IS_INTEGER_TYPE(partitioning_type))
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg(
```
cevian: shorten and move bulk to errhint

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,6 +58,137 @@
...
+
+	tuple = SearchSysCache1(PROCOID, ObjectIdGetDatum(now_func_oid));
+	if (!HeapTupleIsValid(tuple))
+		ereport(ERROR,
+				(errcode(ERRCODE_NO_DATA_FOUND),
+				 errmsg("cache lookup failed for function %u", now_func_oid)));
+
+	now_func = (Form_pg_proc) GETSTRUCT(tuple);
+
+	if ((now_func->provolatile != PROVOLATILE_IMMUTABLE &&
```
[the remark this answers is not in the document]
Ngalstyan4: @davidkohn88 So should we explicitly ban the use of IMMUTABLE functions? There is no reason they would not be allowed. And, in fact for regression testing I am using an IMMUTABLE now function as it makes things easier to understand.

tsl/src/bgw_policy/drop_chunks_api.c
Ngalstyan4: Here we do the search by function regproc oid which is passed as an argument so if I understand it correctly there cannot be any ambiguity.

src/bgw_policy/interval.c
```diff
@@ -0,0 +1,53 @@
...
+#include <postgres.h>
+#include <funcapi.h>
+#include <utils/typcache.h>
+#include <parser/parse_type.h>
+#include <access/htup_details.h>
+
+#include "bgw_policy/interval.h"
+
+void
+ts_interval_deform_datum(Datum ts_interval_datum, Datum *values, bool *isnull)
```
cevian: I would think this should return the FormData object instead of filling in the nulls and values. That way the validation logic should be contained in here.

src/bgw_policy/interval.c
```diff
@@ -0,0 +1,53 @@
...
+	tuple.t_data = th;
+
+	heap_deform_tuple(&tuple, rowdesc, values, isnull);
+
+	// lookup_rowtype_tupdesc gives a ref counted pointer
+	DecrTupleDescRefCount(rowdesc);
+}
+
+HeapTuple
+ts_interval_form_heaptuple(Datum *values, bool *isnull)
```
cevian: Should this take the FormData object?

src/bgw_policy/drop_chunks.c
cevian: Its scan continue so that the calling function can check the count and error if > 1. Can you verify this check happens.

tsl/src/bgw_policy/drop_chunks_api.c
cevian: yes its own file would be better

tsl/src/bgw_policy/drop_chunks_api.c
cevian: Yeah we should probably fix any callers that don't push snapshots when starting txns.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,6 +58,134 @@
...
+						"that have integer time dimensions"))); // <<-- time columns? open
+																// dimensions? TODO::
+
+	ts_integer_now_func_validate(now_func_oid, open_dim_type);
+	// elog(INFO, "haha %s %s", NameStr(now_func->proname),
+	// get_namespace_name(now_func->pronamespace));
+
+	// Q:: TODO:: NOTE we actually load almost all of the info about the function in validation
+	// function above and the dimension_update repeats the DB search. this is done to make the APIs
+	// simple but I could optimize it if deemed necessary
```
cevian: no optimization necessary

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -91,11 +220,67 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
 				 errmsg("could not add drop_chunks policy because \"%s\" is not a hypertable",
 						get_rel_name(ht_oid))));
 
+	/* validate that the open dimension uses a time type */
+	open_dim = hyperspace_get_open_dimension(hypertable->space, 0);
```
cevian: can this be refactored into a constructor for the interval type you created. The constructor should return the FormData_ts_interval or another struct representing the interval. It should live in the interval.c file you already created.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -105,10 +290,15 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
 							get_rel_name(ht_oid))));
 		}
 
-		if (!DatumGetBool(DirectFunctionCall2(interval_eq,
-											  IntervalPGetDatum(&existing->fd.older_than),
-											  IntervalPGetDatum(older_than))) ||
-			(existing->fd.cascade != cascade) ||
+		same_interval =
```
cevian: This should also happen in interval.c/.h and be able to check two interval types for equallity

tsl/src/bgw_policy/job.c
```diff
@@ -159,21 +184,66 @@ execute_drop_chunks_policy(int32 job_id)
...
+								(Datum) 0,
+								INTERVALOID,
+								InvalidOid,
+								args->fd.cascade,
+								args->fd.cascade_to_materializations,
+								LOG);
+	}
+	else
+	{
+		hcache = ts_hypertable_cache_pin();
```
cevian: can we also create an api in interval.c/.h takes a ts_interval and a hypertable_id and returns a datum with internal type or something. Not sure here.

sql/bgw_scheduler.sql
```diff
@@ -21,7 +21,11 @@ INSERT INTO _timescaledb_config.bgw_job (id, application_name, job_type, schedul
 (1, 'Telemetry Reporter', 'telemetry_and_version_check_if_enabled', INTERVAL '24h', INTERVAL '100s', -1, INTERVAL '1h')
 ON CONFLICT (id) DO NOTHING;
 
-CREATE OR REPLACE FUNCTION add_drop_chunks_policy(hypertable REGCLASS, older_than INTERVAL, cascade BOOL = FALSE, if_not_exists BOOL = false, cascade_to_materializations BOOL = false)
+CREATE OR REPLACE FUNCTION set_integer_now_func(hypertable REGCLASS, integer_now_func REGPROC, replace_if_exists BOOL = false) RETURNS VOID
+AS '@MODULE_PATHNAME@', 'ts_set_integer_now_func'
+LANGUAGE C VOLATILE STRICT;
+
+CREATE OR REPLACE FUNCTION add_drop_chunks_policy(hypertable REGCLASS, older_than "any", cascade BOOL = FALSE, if_not_exists BOOL = false, cascade_to_materializations BOOL = false)
```
davidkohn88: We have to actually drop the old function if it exists, either here or probably in the update script as otherwise it will be overloaded in a weird way and may break things horribly.

tsl/src/bgw_policy/drop_chunks_api.c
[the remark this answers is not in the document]
davidkohn88: I would, yes. It just doesn't make any sense, there's no reason not to mark your testing function as stable...

tsl/test/expected/bgw_policy.out
```diff
@@ -420,14 +420,16 @@ select check_chunk_oid(:oldest_chunk_id, :reorder_chunk_oid);
...
-CREATE TABLE test_table_int(time int);
+CREATE TABLE test_table_int(time bigint, junk int);
 SELECT create_hypertable('test_table_int', 'time', chunk_time_interval => 1);
 NOTICE:  adding not-null constraint to column "time"
       create_hypertable      
 -----------------------------
  (2,public,test_table_int,t)
 (1 row)
 
+create or replace function dummy_now() returns BIGINT LANGUAGE SQL IMMUTABLE as  'SELECT 1::BIGINT';
```
davidkohn88: So we probably need a test where you have funcs, 1) called dummy_now() which is fine, and another, called dummy_now(int) which is not fine, right now, you will get errors, and this is why we probably shouldn't take in a regproc...

sql/pre_install/tables.sql
```diff
@@ -193,13 +195,22 @@ CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_reorder (
 	hypertable_index_name	NAME		NOT NULL
 );
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_policy_reorder', '');
+CREATE TYPE _timescaledb_catalog.ts_interval AS (
```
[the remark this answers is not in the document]
Ngalstyan4: @cevian Do you think timescale will have more custom types later? 
if so, I could create a `types.sql` preinstall script and also add a separate test file like the `extension` one to list all visible types.

src/bgw_policy/drop_chunks.c
```diff
@@ -18,17 +19,65 @@
...
-								ti->mctx,
-								BgwPolicyDropChunks,
-								FormData_bgw_policy_drop_chunks);
-
+	bool nulls[Natts_bgw_policy_drop_chunks];
+	Datum values[Natts_bgw_policy_drop_chunks];
+	bool nulls_older_than[Natts_ts_interval];
+	Datum values_older_than[Natts_ts_interval];
+
+	/* cannot use STRUCT_FROM_TUPLE because some fields in the result may be NULLs*/
```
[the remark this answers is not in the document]
Ngalstyan4:
```sql
is_time_interval        BOOLEAN,
time_interval	        INTERVAL,
integer_interval        BIGINT
```
in postgres sql I cannot add custom(e.g. NOT NULL) constraints on composite type fields.
I could still add a validation function and enforce that for example `is_time_interval` is never null.

That said, when `is_time_interval` is false, `time_interval` must be null.
Maybe I still get use `STRUCT_FROM_TUPLE` because sometimes postgres does not compress null fields out but I think doing so will rely on some internal postgres implementation.

sql/bgw_scheduler.sql
Ngalstyan4: Will put it in the update script.
 I think Create or replace fails if the function exists with a different signature so worst case it will fail but not result in an unexpected state

tsl/test/expected/bgw_policy.out
Ngalstyan4: not sure what you mean by 'this is why we probably shouldn't take in a regproc...' ?
I thought as long as we catch the error with an informative message, it is ok?

src/bgw_policy/drop_chunks.c
Ngalstyan4: I see, verified, it is being checked

src/bgw_policy/drop_chunks.c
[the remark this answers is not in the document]
Ngalstyan4: resolved in a call, you are right, nothing in `bgw_policy_drop_chunks` table can be null so will change back to `STRUCT_FROM_TUPLE`

src/catalog.h
```diff
@@ -707,7 +728,7 @@ typedef struct FormData_bgw_policy_drop_chunks
 {
 	int32 job_id;
 	int32 hypertable_id;
-	Interval older_than;
+	FormData_ts_interval older_than;
```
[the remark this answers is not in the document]
Ngalstyan4: @cevian Code later is much simpler when the struct is loaded here instead of having a datum pointer. It helps us encapsulate all of low level custom type reading/writing into a few places only. 

If we change this to datum, Any place that needs the `older_than` as a struct, has to do the loading itself.

src/bgw_policy/interval.c
Ngalstyan4: I made the interface identical to `heap_deform_tuple` thinking that consistency with the standard postgres type deform API would be more beneficial  but I switched it to what you suggested.

src/bgw_policy/interval.c
Ngalstyan4: yes!

tsl/src/bgw_policy/drop_chunks_api.c
cevian: ping

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -55,43 +57,148 @@
 										  Int32GetDatum(0),                                        \
 										  Float8GetDatum(0)))
 
+Datum
```
cevian: this should really move somewhere else

tsl/src/bgw_policy/job.c
```diff
@@ -145,39 +149,108 @@ execute_reorder_policy(BgwJob *job, reorder_func reorder, bool fast_continue)
...
+								IntervalPGetDatum(&args->fd.older_than.time_interval),
+								(Datum) 0,
+								INTERVALOID,
+								InvalidOid,
+								args->fd.cascade,
+								args->fd.cascade_to_materializations,
+								LOG);
+	}
+	else
+	{
```
cevian: This branch should be a lot simpler. all the integer - now stuff should be done in interval.c/.h

tsl/test/sql/bgw_policy.sql
```diff
@@ -216,17 +216,79 @@ select check_chunk_oid(:oldest_chunk_id, :reorder_chunk_oid);
...
 
 \set ON_ERROR_STOP 0
--- we cannot add a drop_chunks policy on a table whose open dimension is not time
+-- we cannot add a drop_chunks policy on a table whose open dimension is not time and no now_func is set
 select add_drop_chunks_policy('test_table_int', INTERVAL '4 months', true);
 \set ON_ERROR_STOP 1
 
+INSERT INTO test_table_int VALUES (-2, -2), (-1, -1), (0,0), (1, 1), (2, 2), (3, 3);
+
+select set_integer_now_func('test_table_int', 'dummy_now');
```
cevian: we should add a test that has a permission error on set_integer_now_func if you are not the owner of the hypertable.

sql/updates/latest-dev.sql
```diff
@@ -1 +1,176 @@
...
+CREATE OR REPLACE FUNCTION add_drop_chunks_policy(hypertable REGCLASS, older_than "any", cascade BOOL = FALSE, if_not_exists BOOL = false, cascade_to_materializations BOOL = false)
+RETURNS INTEGER
+AS '@MODULE_PATHNAME@', 'ts_add_drop_chunks_policy'
+LANGUAGE C VOLATILE STRICT;
+
+CREATE OR REPLACE FUNCTION set_integer_now_func(hypertable REGCLASS, integer_now_func REGPROC, replace_if_exists BOOL = false) RETURNS VOID
+AS '@MODULE_PATHNAME@', 'ts_set_integer_now_func'
+LANGUAGE C VOLATILE STRICT;
+
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.dimension_tmp (
```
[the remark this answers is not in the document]
Ngalstyan4: Ok, I thought it made more sense for integer_now_func name and schema came right after partitioning_func ones but if you think it is better to add them in the end, will change that

sql/updates/latest-dev.sql
```diff
@@ -1 +1,176 @@
...
+
+INSERT INTO _timescaledb_config.bgw_policy_drop_chunks_tmp
+(SELECT job_id, hypertable_id, ROW('t',older_than,NULL)::_timescaledb_catalog.ts_interval  as older_than, cascade, cascade_to_materializations
+FROM _timescaledb_config.bgw_policy_drop_chunks);
+
+ALTER EXTENSION timescaledb DROP TABLE _timescaledb_config.bgw_policy_drop_chunks;
+DROP TABLE _timescaledb_config.bgw_policy_drop_chunks;
+
+
+CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_drop_chunks (
```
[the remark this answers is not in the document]
Ngalstyan4: That is what I initially had
but when renaming, indexes and other dependencies are not renamed accordingly and it is a lot messier to do all the renaming necessary by hand.

sql/updates/latest-dev.sql
```diff
@@ -1 +1,176 @@
...
+);
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_policy_drop_chunks', '');
+
+INSERT INTO _timescaledb_config.bgw_policy_drop_chunks
+(SELECT * FROM _timescaledb_config.bgw_policy_drop_chunks_tmp);
+
+DROP TABLE _timescaledb_config.bgw_policy_drop_chunks_tmp;
+
+GRANT SELECT ON ALL TABLES IN SCHEMA _timescaledb_catalog TO PUBLIC;
```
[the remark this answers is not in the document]
Ngalstyan4: hm, ok can change that. 
But looks like tables.sql grants all permissions anyway and looks like there is a good reason for it.
is there any reason it should be different in the update script ?
```
-- Set table permissions
-- We need to grant SELECT to PUBLIC for all tables even those not
-- marked as being dumped because pg_dump will try to access all
-- tables initially to detect inheritance chains and then decide
-- which objects actually need to be dumped.
GRANT SELECT ON ALL TABLES IN SCHEMA _timescaledb_catalog TO PUBLIC;
GRANT SELECT ON ALL TABLES IN SCHEMA _timescaledb_config TO PUBLIC;
GRANT SELECT ON ALL TABLES IN SCHEMA _timescaledb_internal TO PUBLIC;
GRANT SELECT ON ALL SEQUENCES IN SCHEMA _timescaledb_catalog TO PUBLIC;
GRANT SELECT ON ALL SEQUENCES IN SCHEMA _timescaledb_config TO PUBLIC;
GRANT SELECT ON ALL SEQUENCES IN SCHEMA _timescaledb_internal TO PUBLIC;
```

src/bgw_policy/drop_chunks.c
Ngalstyan4: DISCUSS: I later noticed that if I leave older_than as deformed struct inside the `FormData_bgw_policy_drop_chunks` instead of a Datum, then I cannot just use STRUCT_FROM_TUPLE.

The reason to leave older_than deformed is to make the later code using older_than simpler.

src/bgw_policy/interval.c
```diff
@@ -0,0 +1,300 @@
...
+{
+	FormData_ts_interval *invl;
+	if (interval_type == INTERVALOID)
+	{
+		invl = ts_interval_from_sql_input(relid, interval_datum, interval_type, caller_name);
+		return ts_interval_from_now_to_internal(invl, partitioning_type);
+	}
+	else
+	{
+		return ts_time_value_to_internal(interval_datum, interval_type);
```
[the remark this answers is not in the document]
Ngalstyan4: I got rid of the function and inlined just the first part. That is what I had initally but then thought the code might be reusable and put it here.

src/catalog.h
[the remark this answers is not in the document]
Ngalstyan4: DISCUSS

tsl/src/bgw_policy/job.c
Ngalstyan4: it won't mbe much simpler, but can factor out the lookup call into a dedicated interval function?

src/bgw_policy/drop_chunks.c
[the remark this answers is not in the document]
Ngalstyan4: DISCUSS: I later noticed that if I leave older_than as deformed struct inside the FormData_bgw_policy_drop_chunks instead of a Datum, then I cannot just use STRUCT_FROM_TUPLE.

The reason to leave older_than deformed is to make the later code using older_than simpler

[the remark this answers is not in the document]
Ngalstyan4: Also, I tried converting it back to `STRUCT_FROM_TUPLE` just to have the code in case we decide to proceed with that approach. 

I did not manage to get the catalog.c FormData_bgw_policy_drop_chunks struct right. I am not sure how the custom row type must be stored and was unable to find any examples of that being done in Timescale or postgres source code.
It looks like `STRUCT_FROM_TUPLE` expects the struct we are copying into be 55 bytes but if I assume the row type is stored as datum, struct size is 24 and if I assume it is stored as a struct with all of its fields, it is 48 bytes

tsl/src/bgw_policy/job.c
Ngalstyan4: DISCUSS

Ngalstyan4: Some of the complexity comes from the fact that several calls are required to get the open dimension partitioning type of the hypertable. Perhaps we could add another function similar to ts_dimension_get_partition_type():
 `ts_hypertable_get_open_partition_type(Oid relid)`

tsl/src/bgw_policy/job.c
```diff
@@ -22,16 +24,17 @@
 
 #include "errors.h"
 #include "job.h"
-#include "hypertable.h"
+#include "../hypertable.h"
```
Ngalstyan4: DISCUSS

sql/updates/latest-dev.sql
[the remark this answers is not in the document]
Ngalstyan4: DISCUSS

src/dimension.c
```diff
@@ -1394,4 +1476,13 @@ ts_dimensions_rename_schema_name(char *old_name, char *new_name)
 				NameGetDatum(&old_schema_name));
 
 	ts_scanner_scan(&scanctx);
+	// q:: pretty sure the first one does a seq scan since the index is not set
```
Ngalstyan4: q:: todo::

src/dimension.c
```diff
@@ -1355,11 +1400,25 @@ ts_dimension_add(PG_FUNCTION_ARGS)
 static ScanTupleResult
 dimension_rename_schema_name(TupleInfo *ti, void *data)
```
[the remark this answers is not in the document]
Ngalstyan4: done with 2 different scans of the table. Was not sure how to force postgres to just scan all the rows of the table

tsl/test/sql/bgw_policy.sql
```diff
@@ -216,17 +216,118 @@ select check_chunk_oid(:oldest_chunk_id, :reorder_chunk_oid);
...
+insert into part_time_now_func values
+(1.1, 23.4), (2.2, 22.3), (3.3, 42.3);
+select * from part_time_now_func;
+select set_integer_now_func('part_time_now_func', 'dummy_now');
+select add_drop_chunks_policy('part_time_now_func', 0) as drop_chunks_job_id \gset
+select test_drop_chunks(:drop_chunks_job_id);
+select * from part_time_now_func;
+select remove_drop_chunks_policy('part_time_now_func');
+\c :TEST_DBNAME :ROLE_SUPERUSER
+alter schema public rename to new_public;
```
Ngalstyan4: DISCUSS

tsl/src/bgw_policy/job.c
```diff
@@ -150,34 +153,85 @@ execute_drop_chunks_policy(int32 job_id)
...
+					(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+					 errmsg("could not set integer_now function because \"%s\" is not a hypertable",
+							get_rel_name(table_relid))));
+
+		open_dim = hyperspace_get_open_dimension(hypertable->space, 0);
+		rettype = ts_dimension_get_partition_type(open_dim);
+
+		Assert(IS_INTEGER_TYPE(rettype));
+
+		now_func = get_integer_now_func(table_relid);
```
Ngalstyan4: should take dimension and return now function

Ngalstyan4: refactor all to ___(open_Dim Dimension ,invl ts_interval ) : now - older_than_datum

src/dimension.c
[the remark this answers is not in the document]
Ngalstyan4: DISCUSS

tsl/src/bgw_policy/job.c
```diff
@@ -170,61 +167,34 @@ execute_drop_chunks_policy(int32 job_id)
...
-		ts_cache_release(hcache);
-	}
+	table_relid = ts_hypertable_id_to_relid(args->fd.hypertable_id);
+	hcache = ts_hypertable_cache_pin();
+	hypertable = ts_hypertable_cache_get_entry(hcache, table_relid);
+	/* First verify that the hypertable corresponds to a valid table */
+	if (hypertable == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+				 errmsg("could not set integer_now function because \"%s\" is not a hypertable",
```
cevian: wrong error message

src/interval.c
```diff
@@ -256,31 +256,65 @@ ts_interval_from_now_func_get_datum(int64 interval, Oid time_dim_type, Oid now_f
...
+	now_func = lookup_proc_filtered(NameStr(open_dim->fd.integer_now_func_schema),
+									NameStr(open_dim->fd.integer_now_func),
+									NULL,
+									noarg_integer_now_func_filter,
+									&rettype);
+	return now_func;
+}
+
 /*
  * Convert the difference of interval and current timestamp to internal representation
```
cevian: please update the comment to reflect that the return value is a datum of the dimension type

src/chunk.c
```diff
@@ -1104,7 +1105,8 @@ chunks_typecheck_and_find_all_in_range_limit(Hyperspace *hs, Dimension *time_dim
 											  newer_than_type,
 											  "older_than",
```
cevian: shouldn't this be "newer_than"?

sql/pre_install/types.sql
```diff
@@ -0,0 +1,20 @@
...
+-- logical time types in postgres. The type defined below is an INTERVAL equivalent
+-- for TimescaleDB and enables to represent time intervals for both postgres time
+-- type valued and integer valued time columns.
+CREATE TYPE _timescaledb_catalog.ts_interval AS (
+    is_time_interval        BOOLEAN,
+    time_interval		    INTERVAL,
+    integer_interval        BIGINT
+    );
+
+-- PostgreSQL composite types do not support constraint checks like in the comment above. That is why any table having a ts_interval column must use the following
```
cevian: update this comment. there is no "like in the comment above" anymore

tsl/test/expected/bgw_policy.out
```diff
@@ -437,10 +439,256 @@ WARNING:  unexpected interval: smaller than one second
...
+(2 rows)
+
+select remove_drop_chunks_policy('part_time_now_func');
+ remove_drop_chunks_policy 
+---------------------------
+ 
+(1 row)
+
+\c :TEST_DBNAME :ROLE_SUPERUSER
+alter schema public rename to new_public;
```
cevian: can we also test ALTER FUNCTION...RENAME TO and make sure the catalogs are update when you rename the function?

src/interval.c
```diff
@@ -0,0 +1,362 @@
...
+	if (invl1->is_time_interval != invl2->is_time_interval)
+		return false;
+
+	if (invl1->is_time_interval &&
+		!DatumGetBool(DirectFunctionCall2(interval_eq,
+										  IntervalPGetDatum(&invl1->time_interval),
+										  IntervalPGetDatum(&invl2->time_interval))))
+		return false;
+
+	if (!invl1->is_time_interval && invl1->integer_interval != invl2->integer_interval)
```
cevian: This branch is never hit in the tests. Can we make sure to hit it?

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -105,32 +100,27 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
 							get_rel_name(ht_oid))));
 		}
 
-		if (!DatumGetBool(DirectFunctionCall2(interval_eq,
-											  IntervalPGetDatum(&existing->fd.older_than),
-											  IntervalPGetDatum(older_than))) ||
-			(existing->fd.cascade != cascade) ||
-			(existing->fd.cascade_to_materializations != cascade_to_materializations))
+		if (ts_interval_equal(&existing->fd.older_than, older_than) &&
```
cevian: is the not equal case ever hit here in tests?

Ngalstyan4: Both branches are hit in tsl_tables.sql tests
see [ sl_tables.out#L154 ](https://github.com/timescale/timescaledb/blob/2d8fe54d02b65b3e08d5f522bc54c2eab18b35dd/tsl/test/expected/tsl_tables.out#L154) for equal case
https://github.com/timescale/timescaledb/blob/2d8fe54d02b65b3e08d5f522bc54c2eab18b35dd/tsl/test/expected/tsl_tables.out#L154
see [ sl_tables.out#L162 ](https://github.com/timescale/timescaledb/blob/2d8fe54d02b65b3e08d5f522bc54c2eab18b35dd/tsl/test/expected/tsl_tables.out#L162) for not equal case
https://github.com/timescale/timescaledb/blob/2d8fe54d02b65b3e08d5f522bc54c2eab18b35dd/tsl/test/expected/tsl_tables.out#L162

sql/pre_install/types.sql
```diff
@@ -0,0 +1,20 @@
...
+-- function for constraint validation.
+CREATE OR REPLACE FUNCTION _timescaledb_internal.valid_ts_interval(invl _timescaledb_catalog.ts_interval)
```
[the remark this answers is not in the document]
Ngalstyan4: hm, the issue is that tables.sql is in preinstalls and this is a function that is used as a validation constraint inside the tables.sql preinstall file
so, this function has to be present when the table is being created.
or we can create the table without the constraint and alter the table when the function is defined/redefined but I am not sure if we want to routinely modify catalog tables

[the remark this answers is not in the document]
Ngalstyan4: DISCUSS

tsl/test/expected/bgw_policy.out
Ngalstyan4: hm, renaming function does not change the catalogs and of course later attempts to use the function fail. This fails for partitioning function as well.

Ngalstyan4: DISCUSS

sql/pre_install/types.sql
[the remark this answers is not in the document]
cevian: Ahh I see. Ok let's leave it as is but add a comment.

tsl/test/expected/bgw_policy.out
cevian: Maybe better left to a separate PR then

sql/pre_install/tables.sql
```diff
@@ -86,6 +88,12 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.dimension (
         (num_slices IS NULL AND interval_length IS NOT NULL) OR
         (num_slices IS NOT NULL AND interval_length IS NULL)
     ),
+    -- The check added here and not after the partitioning func check to make sure this one gets
```
cevian: want to discuss this.

src/cross_module_fn.c
```diff
@@ -24,6 +26,18 @@ TS_FUNCTION_INFO_V1(ts_finalize_agg_sfunc);
 TS_FUNCTION_INFO_V1(ts_finalize_agg_ffunc);
 TS_FUNCTION_INFO_V1(continuous_agg_invalidation_trigger);
 
+Datum
+ts_set_integer_now_func(PG_FUNCTION_ARGS)
```
svenklemm: Should setting the function be apache-licensed?

src/dimension.c
```diff
@@ -1355,11 +1400,50 @@ ts_dimension_add(PG_FUNCTION_ARGS)
...
 {
-	HeapTuple tuple = heap_copytuple(ti->tuple);
-	FormData_dimension *dimension = (FormData_dimension *) GETSTRUCT(tuple);
+	/* Dimension table may contain null valued calumns that is why we do not use
```
svenklemm: calumns => columns

src/interval.c
```diff
@@ -0,0 +1,371 @@
...
+	Datum now;
+
+	AssertArg(IS_INTEGER_TYPE(time_dim_type));
+
+	ts_interval_now_func_validate(now_func, time_dim_type);
+	now = OidFunctionCall0(now_func);
+
+	switch (time_dim_type)
+	{
+		case INT2OID:
```
svenklemm: this needs underflow checks

src/interval.c
```diff
@@ -0,0 +1,371 @@
...
+
+		if (InvalidOid == now_func)
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("integer_now function must be set")));
+
+		return ts_interval_from_now_func_get_datum(invl->integer_interval, type_oid, now_func);
+	}
+	/* suppress compiler warnings on MSVC */
+	Assert(false);
```
svenklemm: why not pg_unreachable ?

tsl/src/bgw_policy/job.c
svenklemm: why is this needed?

tsl/src/hypertable.c
```diff
@@ -0,0 +1,86 @@
...
+
+#include <postgres.h>
+#include <catalog/pg_type.h>
+#include <utils/builtins.h>
+#include <utils/lsyscache.h>
+#include <utils/syscache.h>
+#include <miscadmin.h>
+#include <access/xact.h>
+#include <cache.h>
+#include <hypertable.h>
```
svenklemm: hypertable.h is not system include and its included again 3 lines below

src/cross_module_fn.c
Ngalstyan4: We discussed this with @cevian and decided against it as anything using the set `now_func` would be enterprise.
Also, I think the function had some dependencies in the tsl folder which could complicate making it Apache licensed

src/interval.c
Ngalstyan4: done, please check!

PR #1326, opened 2019-06-28
Add architecture and bit size to telemetry

cevian:
Add architecture and pointer bit size fields to what is sent in
telemetry. This will allow us to evaluate the product usage on
different architectures.


cevian:
is there a better name for the bit size field?

PR #1327, opened 2019-06-29
Fix chunk exclusion with ordered append

erimatnor:
With ordered append, chunk exclusion occur only along the primary open
"time" dimension, failing to exclude chunks along additional
partitioning dimensions. For instance, a query on a two-dimensional
table "hyper" (time, device), such as

```
SELECT * FROM hyper
WHERE time > '2019-06-11 12:30'
AND device = 1
ORDER BY time;
```

would only exclude chunks based on the "time" column restriction, but
not the "device" column restriction. This causes an unnecessary number
of chunks to be included in the query plan.

The reason this happens is because chunk exclusion during ordered
append is based on pre-sorting the set of slices in the primary
dimension to determine ordering. This is followed by a scan for
chunks slice-by-slice in the order of the sorted slices. Since those
scans do not include the restrictions in other dimensions, chunks that
would otherwise not match are included in the result.

This change fixes this issue by using the "regular" chunk scan that
account for multi-dimensional restrictions. This is followed by a sort
of the resulting chunks along the primary "time" dimension.

While this, sometimes, means sorting a larger set than the initial
slices in the primary "time" dimension, the resulting chunk set is
smaller instead. Sorting chunks also allows doing secondary ordering
on chunk ID for those chunks that belong to the same "time"
slice. While this additional ordering is not required for correct
tuple ordering, it gives slightly nicer EXPLAIN output since chunks
are also ordered by ID.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -763,6 +763,7 @@ chunk_scan_ctx_init(ChunkScanCtx *ctx, Hyperspace *hs, Point *p)
 	ctx->htab = hash_create("chunk-scan-context", 20, &hctl, HASH_ELEM | HASH_CONTEXT | HASH_BLOBS);
 	ctx->space = hs;
 	ctx->point = p;
+	ctx->num_complete_chunks = 0;
 	ctx->early_abort = false;
 	ctx->lockmode = NoLock;
 }
@@ -902,22 +903,10 @@ chunk_scan_ctx_foreach_chunk(ChunkScanCtx *ctx, on_chunk_func on_chunk, uint16 l
 	return num_processed;
 }
 
-/* Returns true if the chunk has a full set of constraints, otherwise
- * false. Used to find a chunk matching a point in an N-dimensional
- * hyperspace. */
-static bool
-chunk_is_complete(ChunkScanCtx *scanctx, Chunk *chunk)
-{
-	if (scanctx->space->num_dimensions != chunk->constraints->num_dimension_constraints)
-		return false;
-
-	return true;
-}
-
 static ChunkResult
 set_complete_chunk(ChunkScanCtx *scanctx, Chunk *chunk)
 {
-	if (chunk_is_complete(scanctx, chunk))
+	if (chunk_is_complete(chunk, scanctx->space))
 	{
 		scanctx->data = chunk;
 #ifdef USE_ASSERT_CHECKING
@@ -1124,7 +1113,7 @@ chunks_typecheck_and_find_all_in_range_limit(Hyperspace *hs, Dimension *time_dim
 static ChunkResult
 append_chunk_oid(ChunkScanCtx *scanctx, Chunk *chunk)
 {
-	if (!chunk_is_complete(scanctx, chunk))
+	if (!chunk_is_complete(chunk, scanctx->space))
 		return CHUNK_IGNORED;
 
 	/* Fill in the rest of the chunk's data from the chunk table */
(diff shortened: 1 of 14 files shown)


src/dimension_slice.c
```diff
@@ -68,21 +68,14 @@ ts_dimension_slice_create(int dimension_id, int64 range_start, int64 range_end)
...
-	{
-		if (left->fd.range_end == right->fd.range_end)
-			return 0;
-
-		if (left->fd.range_end > right->fd.range_end)
-			return 1;
-
-		return -1;
-	}
+	int res = (left->fd.range_start > right->fd.range_start) -
```
svenklemm: Even though this results in less instructions not sure this makes the code more readable

erimatnor: I think it this is an easy to remember pattern to standardize on that is always "correct." Also, not sure it is that hard to understand; it basically just says that either (1) left  is greater than right, or (2) it is smaller, or (3) both are false = equal (0 - 0). Can aid the reader with a comment if that helps.

src/chunk_constraint.c
```diff
@@ -468,10 +469,19 @@ ts_chunk_constraint_scan_by_dimension_slice(DimensionSlice *slice, ChunkScanCtx
 
 		ts_hypercube_add_slice(chunk->cube, slice);
 
-		if (ctx->early_abort && chunk->constraints->num_dimension_constraints == hs->num_dimensions)
+		/* A chunk is complete when we've added slices for all its dimensions,
+		 * i.e., a complete hypercube */
+		chunk_is_complete = chunk->constraints->num_dimension_constraints == hs->num_dimensions;
```
pmwkaa: There are static function `chunk_is_complete()` exist in `chunk.c`. Maybe it is better to make it public and use here instead? If that makes sense

src/chunk.c
```diff
@@ -1134,13 +1135,37 @@ append_chunk_oid(ChunkScanCtx *scanctx, Chunk *chunk)
...
+	/* Fill in the rest of the chunk's data from the chunk table */
+	chunk_fill_stub(chunk, false);
+
+	if (scanctx->lockmode != NoLock)
+		LockRelationOid(chunk->table_id, scanctx->lockmode);
+
+	if (NULL == chunks && scanctx->num_complete_chunks > 0)
+		scanctx->data = chunks = palloc(sizeof(Chunk *) * scanctx->num_complete_chunks);
+
+	if (scanctx->num_complete_chunks > 0)
```
pmwkaa: I probably misunderstood a logic here, but does it makes sense to rely on `num_complete_chunks` counter here, since there are a check to ignore all non complete chunks at the beginning of the function?

src/dimension_slice.c
pmwkaa: Unless this makes it less convenient, maybe we could make a macro for it? Looks like this is a common pattern and used for chunk comparison as well

src/chunk.c
erimatnor: So, what this function does is to collect the "complete" chunks (those that have slices in all dimensions) after a scan. During the scan, we increment `num_complete_chunks` whenever a chunk gets "completed", then we know how many chunks to collect from the hash table at the end. This simply uses the `num_complete_chunks` as an index into the final array.

src/dimension_slice.c
erimatnor: Added some macros and comments to help with this.

src/chunk_constraint.c
erimatnor: Made it static inline in the header.

src/dimension_slice.c
erimatnor: Done.

PR #1330, opened 2019-07-03
Fail when adding space dimension with no partitions

erimatnor:
Calling `create_hypertable` with a space dimension silently succeeds
without actually creating the space dimension if `num_partitions` is
not specified.

This change ensures that we raise an appropriate error when a user
fails to specify the number of partitions.

### DIFF
--- a/src/dimension.c
+++ b/src/dimension.c
@@ -1043,6 +1043,7 @@ ts_dimension_info_create_open(Oid table_relid, Name column_name, Datum interval,
 {
 	DimensionInfo *info = palloc(sizeof(*info));
 	*info = (DimensionInfo){
+		.type = DIMENSION_TYPE_OPEN,
 		.table_relid = table_relid,
 		.colname = column_name,
 		.interval_datum = interval,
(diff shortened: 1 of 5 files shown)


src/dimension.c
```diff
@@ -1067,14 +1069,67 @@ ts_dimension_info_create_closed(Oid table_relid, Name column_name, int32 num_sli
...
+					 errmsg("invalid partitioning function"),
+					 errhint("A valid partitioning function for open (time) dimensions must be "
+							 "IMMUTABLE, "
+							 "take the column type as input, and return an integer or "
+							 "timestamp type.")));
+
+		dimtype = get_func_rettype(info->partitioning_func);
+	}
+
+	info->interval = dimension_interval_to_internal(NameStr(*info->colname),
```
cevian: nit: don't love the validate function having the side-effect of setting a member. But no big deal.

erimatnor: FWIW, this is old code I moved around and is actually prevalent in the validation functions. I think the thing that makes this OK is that this is secondary information that is set and can be considered the "result" of the validation. In other words, the information set is not critical for the validation. 

I guess one could break this up in separate validate and "transform"/process functions, but seems out-of-scope of this PR.

PR #1332, opened 2019-07-03
Add job statistics columns to timescaledb_information.continuous_aggregate_stats view

gayyappan:
Add more information about job history for continuous aggregate
background worker jobs.

### DIFF
--- a/tsl/test/expected/continuous_aggs_usage.out
+++ b/tsl/test/expected/continuous_aggs_usage.out
@@ -99,9 +99,14 @@ completed_threshold    | Sun Dec 30 22:00:00 2018 PST
 invalidation_threshold | Sun Dec 30 22:00:00 2018 PST
 job_id                 | 1000
 last_run_started_at    | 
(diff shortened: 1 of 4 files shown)


tsl/test/expected/continuous_aggs_usage.out
```diff
@@ -102,6 +102,10 @@ last_run_started_at    |
 job_status             | 
 last_run_duration      | 
 next_scheduled_run     | 
+total_runs             | 
+total_successes        | 
+total_failures         | 
+total_crashes          |
```
dianasaur323: Is it possible, or do you think it would be helpful - to also show last successful run? or is that overlapping with completed_threshold?

gayyappan: The catalog tables show the status of the last job last time - did it succeed or fail? We  don't have enough info to pin point when was the last time  the job succeeded. Is that what  you are looking for?

PR #1336, opened 2019-07-06
Avoid starting postmaster when there are no tests

Ngalstyan4:


svenklemm:
the diff coverage looks suspicious seems like not everything is running with this change

Issue #1335, opened 2019-07-06
Continuous aggregates with negative refresh lag don't update


ik9999:
**Relevant system information:**
 - OS: Debian 9
 - PostgreSQL version (output of `postgres --version`): 11
 - TimescaleDB version (output of `\dx` in `psql`): 1.4.0-dev (1.4.x branch - c8f508d40e557abbe22ec18c84d05a0b7361458b commit)
 - Installation method: docker

**Describe the bug**
Continuous aggregates with negative refresh lag don't update themselves and i cant update them manually.

```
# REFRESH MATERIALIZED VIEW table_m1;
INFO:  new materialization range for public.records (time column record_time) (1562338800000000)
INFO:  materializing continuous aggregate public.table_m1: new range up to 1562338800000000
ERROR:  duplicate key value violates unique constraint "continuous_aggs_materialization_invalidation_log_pkey"
DETAIL:  Key (materialization_id)=(4) already exists.
```

This error only happens when i use negative refresh_lag. It works fine with positive refresh_lag.

**To Reproduce**
- Create continuous aggregate with negative refresh lag
```
CREATE VIEW table_m1
WITH (timescaledb.continuous, timescaledb.refresh_lag = '-1 minute')
AS
SELECT
  time_bucket('1 minute', record_time) as initial_time,
  MAX(val) AS high,
  MIN(val) AS low,
  first(val, record_time) AS initial_val,
  last(val, record_time) AS final_val,
  item_id AS item_id
FROM
  records
GROUP BY initial_time, item_id;
```
- Do `REFRESH MATERIALIZED VIEW table_m1;`

**Expected behavior**
Update without errors

mfreed:
Hi @ik9999 Can you post the full schema (and all indexes) -- try `\d+ records` -- on the hypertable on which the cagg is defined?  Thanks!

ik9999:
```
# \d+ records
                                                             Table "public.records"
    Column     |            Type             | Collation | Nullable |              Default                | Storage | Stats target | Description
---------------+-----------------------------+-----------+----------+-------------------------------------+---------+--------------+-------------
 id            | integer                     |           | not null | nextval('records_id_seq'::regclass) | plain   |              |
 record_time   | timestamp without time zone |           | not null |                                     | plain   |              |
 item_id       | integer                     |           | not null |                                     | plain   |              |
 val           | numeric(20,10)              |           | not null |                                     | main    |              |
[...]

PR #1339, opened 2019-07-08
Fix continuous agg catalog table insert failure

gayyappan:
The primary key on continuous_aggs_materialization_invalidation_log
prevents multiple records with the same materialization id. Remove
the primary key to fix this problem.

### DIFF
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -3,7 +3,7 @@ ALTER TABLE _timescaledb_catalog.telemetry_metadata ALTER COLUMN include_in_tele
 ALTER TABLE _timescaledb_catalog.telemetry_metadata RENAME TO metadata;
 ALTER INDEX _timescaledb_catalog.telemetry_metadata_pkey RENAME TO metadata_pkey;
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_materialization_invalidation_log(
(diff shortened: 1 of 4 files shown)


mfreed:
Fixes https://github.com/timescale/timescaledb/issues/1335

sql/updates/latest-dev.sql
```diff
@@ -3,7 +3,7 @@ ALTER TABLE _timescaledb_catalog.telemetry_metadata ALTER COLUMN include_in_tele
 ALTER TABLE _timescaledb_catalog.telemetry_metadata RENAME TO metadata;
 ALTER INDEX _timescaledb_catalog.telemetry_metadata_pkey RENAME TO metadata_pkey;
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_materialization_invalidation_log(
-    materialization_id INTEGER PRIMARY KEY
```
svenklemm: shouldnt this be an alter table in the update script or not be here at all

gayyappan: this is a new table for 1.4. Update does not work if I don't have the CREATE TABLE here.

svenklemm: yes since its a new table the create table command should not be in latest-dev.sql at all because the actual update script for a specific version gets generated and the generated file will already include tables.sql which has the required create table command

PR #1344, opened 2019-07-09
Update continuous agg bgw job start time

gayyappan:
When the continuous aggregate refresh_interval setting is modified,
it does not modify the job schedule until the next scheduled job
runs. This fix addresses it by updating the next_start time for
the bgw job.

### DIFF
--- a/src/bgw/job.c
+++ b/src/bgw/job.c
@@ -9,9 +9,11 @@
 #include <access/xact.h>
 #include <pgstat.h>
 #include <utils/memutils.h>
+#include <utils/timestamp.h>
 #include <miscadmin.h>
 #include <storage/ipc.h>
 #include <tcop/tcopprot.h>
+#include <utils/builtins.h>
 
 #include "job.h"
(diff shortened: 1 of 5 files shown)


src/bgw/job.c
```diff
@@ -597,9 +598,17 @@ bgw_job_tuple_update_by_id(TupleInfo *ti, void *const data)
 	BgwJob *updated_job = (BgwJob *) data;
 	HeapTuple tuple = heap_copytuple(ti->tuple);
 	FormData_bgw_job *fd = (FormData_bgw_job *) GETSTRUCT(tuple);
+	TimestampTz nowts, next_start;
 
 	ts_bgw_job_permission_check(updated_job);
-
+	/* when we update the schedule interval, modify the next start time as well*/
```
cevian: I think this should only happen `if(fd->schedule_interval != updated_job->fd.schedule_interval)`

gayyappan: changed.

PR #1349, opened 2019-07-14
Handle CustomScan node with scanrelid similar to other Scans

svenklemm:
This patch changes ChunkAppend to treat CustomScan nodes which have
scanrelid set similar to postgres native Scan nodes. This enables
constraint exclusion to work for CustomScan nodes that scan a relation.


svenklemm:
we don't have CustomScan node ourselves that scans a relation yet

PR #1362, opened 2019-07-19
Fix ConstraintAwareAppend subquery exclusion

### LINKED ISSUE #1360: ConstraintAwareAppend does not exclude chunks in subquery
 - OS: CentOS 7
 - PostgreSQL version 11.2
 - TimescaleDB version 1.4.0
 - Installation method: source

**Describe the bug**
When there's a subquery and an order by clause, the ConstraintAwareAppend does not exclude chunks in v1.4.0

**To Reproduce**
Easiest way to reproduce is by explicitly setting timescaledb.enable_chunk_append to off, because that forces ConstraintAwareAppend in all cases. I have seen it in cases even with enable_chunk_append=on though, so this setting is just to easily reproduce it.

set timescaledb.enable_chunk_append=off;
drop table a cascade;
create table a (a int, b timestamptz);
select create_hypertable('a', 'b', chunk_time_interval=>interval '1d');
insert into a values (1, now()), (2,now()-interval '1d'), (3, now()-interval '2d');
explain select * from a where b >= now(); -- ok
explain select * from a where b >= now() order by a, b; -- ok
explain select * from (select * from a where b >= now() order by a, b) _; -- not ok

**Expected behavior**
Last query should exclude chunks as the other two queries before it also did.

**Actual behavior**
No chunks get excluded in the third query. 

**Screenshots**
N/A

**Additional context**
This looks like a regression, as this worked in earlier versions.

svenklemm:
The RangeTblEntry index used during planning might be different from
the index used during execution due to flattening. This patch changes
ConstraintAwareAppend to handle queries where the index changed between
planning and execution.

Fixes #1360

### DIFF
--- a/test/expected/append-11.out
+++ b/test/expected/append-11.out
@@ -7,7 +7,7 @@ SELECT format('include/%s_load.sql', :'TEST_BASE_NAME') as "TEST_LOAD_NAME",
        format('%s/results/%s_results_optimized.out', :'TEST_OUTPUT_DIR', :'TEST_BASE_NAME') as "TEST_RESULTS_OPTIMIZED",
        format('%s/results/%s_results_unoptimized.out', :'TEST_OUTPUT_DIR', :'TEST_BASE_NAME') as "TEST_RESULTS_UNOPTIMIZED"
 \gset
-SELECT format('\! diff %s %s', :'TEST_RESULTS_OPTIMIZED', :'TEST_RESULTS_UNOPTIMIZED') as "DIFF_CMD"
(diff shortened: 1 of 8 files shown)


fvannee:
I haven't looked at the code much, but I can confirm this fixes the issue

test/expected/append-11.out
```diff
@@ -1237,49 +1249,105 @@ ORDER BY time DESC;
...
+@@ -1,7 +1,7 @@
+               setting              | value 
+ -----------------------------------+-------
+- timescaledb.disable_optimizations | on
+- timescaledb.enable_chunk_append   | on
++ timescaledb.disable_optimizations | off
++ timescaledb.enable_chunk_append   | off
+ (2 rows)
+ 
+  time | temp | colorid
```
cevian: what is this line doing here?

svenklemm: that is the context from the unified diff, 3 lines before and 3 lines after changed lines

PR #1365, opened 2019-07-22
Clean up agg_bookends test

svenklemm:
Remove unnecessary quotes from identifiers and wrap modifying queries
in transaction to enable multiple executions.


cevian:
nice cleanup

PR #1373, opened 2019-07-25
Only allow PARAM_EXTERN as time_bucket_gapfill arguments

### LINKED ISSUE #1372: `time_bucket_gapfill` does not appear to evaluate subexpressions in `start` and `end` parameters correctly
**Version info**
 - PostgreSQL: 11.3
 - TimescaleDB: 1.4.0

**Issue**

When using a sub-expression as the `start` or `end` parameters with `time_bucket_gapfill`, the sub-expression does not appear to be evaluated correctly.

**Reproduction**

Here are two queries that I should show the issue (simplified examples, the original issue was due to a `LATERAL` query):

1. This query produces the expected result:
```sql
SELECT SUM(duration) FROM (
    SELECT
        time_bucket_gapfill('1m', time, now() - '1h'::interval, now()) AS t,
        '1m'::interval AS duration
    FROM
        (SELECT now() AS time) AS n
    GROUP BY t
) AS inner_query;
```
```sql
   sum
----------
 01:01:00
(1 row)
```
```sql
                                                 QUERY PLAN
------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=0.07..0.07 rows=1 width=16) (actual time=0.032..0.033 rows=1 loops=1)
   ->  Custom Scan (GapFill)  (cost=0.03..0.05 rows=1 width=0) (actual time=0.018..0.023 rows=61 loops=1)
         ->  Group  (cost=0.03..0.05 rows=1 width=0) (actual time=0.016..0.017 rows=1 loops=1)
               Group Key: (time_bucket_gapfill('00:01:00'::interval, now()))
               ->  Sort  (cost=0.03..0.04 rows=1 width=8) (actual time=0.015..0.015 rows=1 loops=1)
                     Sort Key: (time_bucket_gapfill('00:01:00'::interval, now()))
                     Sort Method: quicksort  Memory: 25kB
                     ->  Result  (cost=0.00..0.02 rows=1 width=8) (actual time=0.001..0.001 rows=1 loops=1)
 Planning Time: 0.149 ms
 Execution Time: 0.097 ms
(10 rows)
```

2. This query does *not* produce the correct result:

```sql
SELECT SUM(duration) FROM (
    SELECT 
        time_bucket_gapfill('1m', time, (SELECT now() - '1h'::interval), now()) AS t, 
        '1m'::interval AS duration
    FROM
        (SELECT now() AS time) AS n
    GROUP BY t
) AS inner_query;
```
```sql
     sum
--------------
 171486:45:00
(1 row)
```
```sql
                                                    QUERY PLAN
------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=0.07..0.08 rows=1 width=16) (actual time=1348.695..1348.695 rows=1 loops=1)
   ->  Custom Scan (GapFill)  (cost=0.04..0.06 rows=1 width=0) (actual time=0.021..491.840 rows=10289205 loops=1)
         InitPlan 1 (returns $0)
           ->  Result  (cost=0.00..0.01 rows=1 width=8) (never executed)
         ->  Group  (cost=0.03..0.04 rows=1 width=0) (actual time=0.019..0.023 rows=1 loops=1)
               Group Key: (time_bucket_gapfill('00:01:00'::interval, now()))
               ->  Sort  (cost=0.03..0.03 rows=1 width=8) (actual time=0.017..0.020 rows=1 loops=1)
                     Sort Key: (time_bucket_gapfill('00:01:00'::interval, now()))
                     Sort Method: quicksort  Memory: 25kB
                     ->  Result  (cost=0.00..0.02 rows=1 width=8) (actual time=0.001..0.001 rows=1 loops=1)
 Planning Time: 0.108 ms
 Execution Time: 1348.760 ms
(12 rows)
```

**Notes**

Notably it appears that the start time appears to have been evaluated as  `2000-01-01` which is = `0` in postgres's representation of timestamps.

svenklemm:
Since bucket_width, start and stop of the gapfill node are
initialized during executor startup only PARAM_EXTERN params
are safe to be used.  This patch will make the gapfill node
throw an error if any argument is a param that is not
PARAM_EXTERN. This patch also improves the error message
for negative or 0 bucket_width.

Fixes #1372

### DIFF
--- a/tsl/src/gapfill/exec.c
+++ b/tsl/src/gapfill/exec.c
@@ -142,6 +142,8 @@ gapfill_internal_get_datum(int64 value, Oid type)
 static inline int64
 interval_to_usec(Interval *interval)
 {
+	if (interval == NULL)
+		ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR), errmsg("interval cannot be NULL")));
 	return (interval->month * DAYS_PER_MONTH * USECS_PER_DAY) + (interval->day * USECS_PER_DAY) +
 		   interval->time;
 }
@@ -273,7 +275,10 @@ is_simple_expr_walker(Node *node, void *context)
 		case T_CoerceViaIO:
 		case T_CaseExpr:
 		case T_CaseWhen:
+			break;
 		case T_Param:
+			if (castNode(Param, node)->paramkind != PARAM_EXTERN)
+				return true;
 			break;
 		default:
 			return true;
@@ -530,6 +535,16 @@ gapfill_begin(CustomScanState *node, EState *estate, int eflags)
 
 	state->gapfill_period =
 		gapfill_period_get_internal(func->funcresulttype, exprType(linitial(args)), arg_value);
+
+	/*
+	 * this would error when trying to align start and stop to bucket_width as well below
+	 * but checking this explicitly here will make a nicer error message
+	 */
+	if (state->gapfill_period <= 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg(
+					 "invalid time_bucket_gapfill argument: bucket_width must be greater than 0")));
 
 	/*
 	 * check if gapfill start was left out so we have to infer from WHERE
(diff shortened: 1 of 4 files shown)


tsl/src/gapfill/exec.c
```diff
@@ -142,6 +142,8 @@ gapfill_internal_get_datum(int64 value, Oid type)
 static inline int64
 interval_to_usec(Interval *interval)
 {
+	if (interval == NULL)
+		return 0;
```
JLockerman: should this ERROR on NULL? it seems like a contract violation to pass in no `interval`

JLockerman:
otherwise LGTM

tsl/src/gapfill/exec.c
svenklemm: changed to error

PR #1385, opened 2019-08-02
Run all tests on cron triggered runs

svenklemm:
Change cron triggered travis runs to include all tests.
This patch also adds tagging for coverage information to prevent
faulty coverage differences.


.travis.yml
```diff
@@ -139,44 +143,47 @@ jobs:
         - docker exec docker_arm_emulator /bin/bash -c "cd /build && make install"
       before_script:
       after_failure:
+      after_success:
```
cevian: why do we need this change instead of just setting CODECOV_FLAGS="-F arm"

.travis.yml
```diff
@@ -9,6 +9,7 @@ env:
...
     - secure: "S8WfPsrVGg21vZPxjrmy1X19T7zpHNgaSAoqYGzJdDZcNYuZbI87+tP3VQZo4MxjP3Ag46zTI5K+6aMjCqYqFGKufZdEaSpPSLmppFbXALgSp/rOkzp1M9HhGj0QVP7QcoPbva5Bjnj3zVi9CMAo6MwtnGAvzqBLOsdtnXTxEmKYyeNgwUVbQi2nzO38GeLgabaFL4XD3f3bmHE5jG6QAwZ/L7+jajsD1biH4sOiRLCRshtvermtYxh3SK6rML2kiwATABxTd1xDdMVRO/llmDBN373tHK6mufJyIhrwG9oTprSSDDIJlEbSPCkH0uvJBGcDvOAWf9tr4TGs/beawx3ELcXqR/4EisFqvMiVL4Vpt1gMBg9gue4Y0wpHZYoBOOef2gFtuyL7PjYe9koYJAJxg9DJ29DMxIjuCLTnbHN6yLa9425pdvxcuNEJ2K2zD0ZWoLSVueL0MxLhOZJsqlMkmmrFyI7y5cj1XvJF3vcsa5yK3S7sXroKUzeLu6QMh8hO3jD/2IvdQJ7Huy5uiJk2a+KUceLzruxLbwPF4b8SEoTquJ9NFGtagCG77lbIbLltLs2fUf+JnM8ipdpoMibVRFBOahP8NZwpL/NFzG3WLH3tDncj/c3fneMiNv3FUHLf6ufOVOayQiAEidszcMct2R20qIxggXKymrjZA3k="
     - RETRY_PREFIX=$(if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then echo "travis_retry"; else echo ""; fi)
+    - CODECOV_FLAGS="-F cron"
```
cevian: the default here is cron but in .codecov.yml it's PR. why?

svenklemm: because thats the config for the PR status report so we only want to diff it against coverage that runs on PRs

.travis.yml
svenklemm: i didnt have env variable initially so this was leftover, removed

PR #1388, opened 2019-08-04
Refactor ChunkAppend plan selection

svenklemm:
Move ChunkAppend selection of next subplan into separate function

### DIFF
--- a/src/chunk_append/exec.c
+++ b/src/chunk_append/exec.c
@@ -29,6 +29,11 @@
 #include "chunk_append/planner.h"
 #include "compat.h"
 
+#define INVALID_SUBPLAN_INDEX -1
+#define NO_MATCHING_SUBPLANS -2
+
+static void choose_next_subplan(ChunkAppendState *state);
+
 static TupleTableSlot *chunk_append_exec(CustomScanState *node);
 static void chunk_append_begin(CustomScanState *node, EState *estate, int eflags);
 static void chunk_append_end(CustomScanState *node);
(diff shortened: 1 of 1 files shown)


src/chunk_append/exec.c
```diff
@@ -333,18 +333,44 @@ chunk_append_exec(CustomScanState *node)
...
+		state->current = NO_MATCHING_SUBPLANS;
+		return;
+	}
+
+	if (state->runtime_exclusion)
+	{
+		if (!state->runtime_initialized)
+			initialize_runtime_exclusion(state);
+
+		if (bms_is_empty(state->valid_subplans))
```
JLockerman: how come we do this now even if `state->runtime_initialized`?

svenklemm: runtime exclusion might have excluded all but bms_next_member should come to the same conclusion so this check can be removed

src/chunk_append/exec.c
```diff
@@ -333,18 +333,41 @@ chunk_append_exec(CustomScanState *node)
 		 * more subplans, return the empty slot set up for us by
```
cevian: the part of the comment about returning empty slot is now out of date? (happens above?)

src/chunk_append/exec.c
```diff
@@ -293,9 +293,9 @@ chunk_append_exec(CustomScanState *node)
 
 		CHECK_FOR_INTERRUPTS();
 
-		/*
-		 * figure out which subplan we are currently processing
-		 */
+		if (state->current == NO_MATCHING_SUBPLANS)
+			return ExecClearTuple(node->ss.ps.ps_ResultTupleSlot);
+
```
cevian: can we Assert(state->current != INVALID_SUBPLAN_INDEX) here?

PR #1393, opened 2019-08-08
Add support for moving chunks to different tablespaces

davidkohn88:
Adds a move_chunk function which to a different tablespace. This is
implemented as an extension to the reorder command.
Given that the heap, toast tables, and indexes are being rewritten
during the reorder operation, adding the ability to modify the tablespace
is relatively simple and mostly requires adding parameters to the relevant
functions for the destination tablespace (and index tablespace). The tests
do not focus on further exercising the reorder infrastructure, but instead
ensure that tablespace movement and permissions checks properly occur.

### DIFF
--- a/sql/maintenance_utils.sql
+++ b/sql/maintenance_utils.sql
@@ -10,3 +10,11 @@ CREATE OR REPLACE FUNCTION reorder_chunk(
     index REGCLASS=NULL,
     verbose BOOLEAN=FALSE
 ) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_reorder_chunk' LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE FUNCTION move_chunk(
+    chunk REGCLASS,
+    destination_tablespace Name,
+    index_destination_tablespace Name=NULL,
+    reorder_index REGCLASS=NULL,
+    verbose BOOLEAN=FALSE
+) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_move_chunk' LANGUAGE C VOLATILE;
--- a/tsl/src/reorder.c
+++ b/tsl/src/reorder.c
@@ -32,6 +32,7 @@
 #include <catalog/toasting.h>
 #include <commands/cluster.h>
 #include <commands/tablecmds.h>
+#include <commands/tablespace.h>
 #include <commands/vacuum.h>
 #include <miscadmin.h>
 #include <nodes/pg_list.h>
@@ -61,11 +62,13 @@
 #include "license.h"
 #include "reorder.h"
 
-extern void timescale_reorder_rel(Oid tableOid, Oid indexOid, bool verbose, Oid wait_id);
+extern void timescale_reorder_rel(Oid tableOid, Oid indexOid, bool verbose, Oid wait_id,
+								  Oid destination_tablespace, Oid index_tablespace);
 
 #define REORDER_ACCESS_EXCLUSIVE_DEADLOCK_TIMEOUT "101000"
 
-static void timescale_rebuild_relation(Relation OldHeap, Oid indexOid, bool verbose, Oid wait_id);
+static void timescale_rebuild_relation(Relation OldHeap, Oid indexOid, bool verbose, Oid wait_id,
+									   Oid destination_tablespace, Oid index_tablespace);
 static void copy_heap_data(Oid OIDNewHeap, Oid OIDOldHeap, Oid OIDOldIndex, bool verbose,
 						   bool *pSwapToastByContent, TransactionId *pFreezeXid,
 						   MultiXactId *pCutoffMulti);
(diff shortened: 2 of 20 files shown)


tsl/src/reorder.c
```diff
@@ -101,12 +104,59 @@ tsl_reorder_chunk(PG_FUNCTION_ARGS)
...
 
-	reorder_chunk(chunk_id, index_id, verbose, wait_id);
+	reorder_chunk(chunk_id, index_id, verbose, wait_id, InvalidOid, InvalidOid);
+	PG_RETURN_VOID();
+}
+
+Datum
+tsl_move_chunk(PG_FUNCTION_ARGS)
+{
+	Oid chunk_id = PG_ARGISNULL(0) ? InvalidOid : PG_GETARG_OID(0);
```
cevian: should we check that the heap and indexes are not already in the right namespace? Not sure here because as it is now this will still do a reorder, which may be useful.

sql/maintenance_utils.sql
```diff
@@ -10,3 +10,11 @@ CREATE OR REPLACE FUNCTION reorder_chunk(
     index REGCLASS=NULL,
     verbose BOOLEAN=FALSE
 ) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_reorder_chunk' LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE FUNCTION move_chunk(
```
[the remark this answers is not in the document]
cevian: or maybe a reorder_if_already_moved flag

tsl/test/expected/move.out
```diff
@@ -0,0 +1,469 @@
...
+
+SELECT * FROM test.show_indexes('_timescaledb_internal._hyper_1_2_chunk');
+                              Index                               |  Columns   | Expr | Unique | Primary | Exclusion | Tablespace 
+------------------------------------------------------------------+------------+------+--------+---------+-----------+------------
+ _timescaledb_internal._hyper_1_2_chunk_cluster_test_location_idx | {location} |      | f      | f       | f         | 
+ _timescaledb_internal._hyper_1_2_chunk_cluster_test_time_idx     | {time}     |      | f      | f       | f         | 
+(2 rows)
+
+\set ON_ERROR_STOP 0
+-- cannot move a chunk with no reorder index on first call
```
cevian: maybe add a test for what happens if the previous reorder index is dropped?

tsl/test/expected/move.out
```diff
@@ -0,0 +1,469 @@
...
+(2 rows)
+
+\set ON_ERROR_STOP 0
+-- cannot move a chunk with no reorder index on first call 
+SELECT move_chunk(chunk=>'_timescaledb_internal._hyper_1_2_chunk', destination_tablespace=>'tablespace1', index_destination_tablespace=>'tablespace1');
+WARNING:  Timescale License expired
+ERROR:  there is no previously clustered index for table "_hyper_1_2_chunk"
+-- cannot move a chunk without a destination tablespace set
+SELECT move_chunk(chunk=>'_timescaledb_internal._hyper_1_2_chunk', index_destination_tablespace=>'tablespace1', reorder_index=>'_timescaledb_internal._hyper_1_2_chunk_cluster_test_time_idx');
+ERROR:  function move_chunk(chunk => unknown, index_destination_tablespace => unknown, reorder_index => unknown) does not exist at character 8
```
svenklemm: this is not supposed to error like that is it?

davidkohn88: No, this is fine. It should fail in some way if it's not set, because it's a required argument, i think this is the way it will fail?

davidkohn88: eh, nvm, set it to null explicitly so it would trigger failure.

tsl/test/expected/move.out
davidkohn88: added in reorder tests

tsl/src/reorder.c
davidkohn88: It's unclear to me which behavior would be proper, given that this does also reorder...given that we'd have to loop through all indexes, and then we'd not just move that index, but actually still do the reorder, it seems like it'd be harder to understand than just saying, if you move a table to the tablespace it's already in, it gets reordered and re-written in that tablespace, so don't do it willy-nilly.

sql/maintenance_utils.sql
```diff
@@ -10,3 +10,11 @@ CREATE OR REPLACE FUNCTION reorder_chunk(
...
+    chunk REGCLASS,
+    destination_tablespace Name,
```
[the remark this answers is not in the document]
davidkohn88: I'm open to it? But don't feel strongly either way @mfreed ? @dianasaur323 ?

sql/maintenance_utils.sql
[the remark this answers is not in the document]
davidkohn88: I don't think we want to do this...if people want it we can futz with it later, but I'm not horribly interested in adding complexity to this right now, either in backend implementation or in having more parameters in this function.

sql/maintenance_utils.sql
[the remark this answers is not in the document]
cevian: let's make this change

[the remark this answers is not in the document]
JLockerman: @cevian is this last comment from before or after the in-video conversation?

PR #1407, opened 2019-08-21
Make version mismatch errors FATAL

### LINKED ISSUE #1406: Version mismatch errors should force clients to reconnect
 - OS: :man_shrugging: timesacle cloud / aiven
 - PostgreSQL version (output of `postgres --version`): 11.5
 - TimescaleDB version (output of `\dx` in `psql`): while upgrading from 1.2.1 to 1.4.1
 - Installation method: timescale cloud / aiven

**Describe the bug**
After running `ALTER EXTENSION timescaledb UPDATE` on our production DB we immediately began seeing `extension "timescaledb" version mismatch: shared library version 1.2.2; SQL version 1.4.1`.

With help from Slack support we figured out that we needed to kill all the old connections from our app to the DB so that it could create fresh connections without mismatches. This didn't happen when we upgraded our staging environment DB first, probably because it doesn't have the constant traffic the production DB gets. This makes sense in hindsight, but wasn't obvious to us when we started the upgrade process.

**To Reproduce**
1. Check that the web console says "Everything up to date"
2. `psql $DB_URL -x -c "ALTER EXTENSION timescaledb UPDATE"`
3. Have your app try to reuse connections that were created before step 2

davidkohn88:
We currently check and throw an error if the version loaded in the
client is different from the installed extension version, however
there is no way to recover from this state in a backend. (There is
no way to load the new version as we cannot unload the old and no
commands can be effectively run). Now, we instead throw a
FATAL error which will cause the client to reconnect so it can load
the proper extension version.

Fixes #1406 
@svenklemm Do you agree that this is an appropriate use of a FATAL?

### DIFF
--- a/src/extension.c
+++ b/src/extension.c
@@ -68,7 +68,11 @@ ts_extension_check_version(const char *so_version)
 
 	if (strcmp(sql_version, so_version) != 0)
 	{
-		ereport(ERROR,
+		/*
+		 * Throw a FATAL error here so that clients will be forced to reconnect
+		 * when they have the wrong extension version loaded.
+		 */
+		ereport(FATAL,
 				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 				 errmsg("extension \"%s\" version mismatch: shared library version %s; SQL version "
 						"%s",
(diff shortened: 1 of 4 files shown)


davidkohn88:
@cevian said we should go forward, going to merge with your two reviews.

PR #1420, opened 2019-08-30
Fix OUTER JOIN qual propagation

### LINKED ISSUE #1414: joins with where condition returns incorrect result
 - OS: Ubuntu 19.04
 - PostgreSQL version: 11.5
 - TimescaleDB version: 1.4.1
 - Installation method: docker

**Describe the bug**

I was doing `LEFT OUTER JOIN` with a `WHERE` clause and noticed that certain rows in my LEFT table are missing if there are no corresponding entry (as per the ON condition) on the RIGHT table. Vanilla Postgres gives me these rows with references from RIGHT table set to NULL.

**To Reproduce**
1. Start timescale, create the following schema:
```sql
create table test_table_1 (
  time timestamptz default now(),
  id int,
  name text
);
create table test_table_2 (
  time timestamptz default now(),
  id int,
  name text
);

select create_hypertable('test_table_1', 'time');
select create_hypertable('test_table_2', 'time');

insert into test_table_1 values(now(), 1, 'a');
insert into test_table_1 values(now(), 2, 'b');
insert into test_table_2 values(now(), 1, 'a');
```
2. Make the following query:
```sql
select one.id, two.name from test_table_1 one left outer join test_table_2 two on one.id=two.id where one.id=2;
```

**Expected behavior**
The result should be:
```
postgres=# select one.id, two.name from test_table_1 one left outer join test_table_2 two on one.id=two.id where one.id=2;
 id | name 
----+------
  2 | 
(1 row)
```

**Actual behavior**
The output was:
```
postgres=# select one.id, two.name from test_table_1 one left outer join test_table_2 two on one.id=two.id where one.id=2;
 id | name 
----+------
(0 rows)
```

**Screenshots**
NA

**Additional context**
This expected result is returned by vanilla Postgres and also on Timescale if the table is not a hypertable. `jitans` on Slack tried with Postgres inheritance to see if this is a bug with PG, but got the expected result [ref](https://timescaledb.slack.com/archives/C63MYDZ35/p1566888145148100).

svenklemm:
The code for propagating quals to JOINs was too aggressive and
propagated quals for OUTER JOINs when it was not safe to do so.
This patch fixes this behaviour. This patch also refactors
the qual processing and splits the processing of JOIN related
information into a separate function.

Fixes #1414


src/plan_expand_hypertable.c
```diff
@@ -51,6 +51,7 @@ typedef struct CollectQualCtx
 	List *restrictions;
 	FuncExpr *chunk_exclusion_func;
 	List *join_conditions;
```
cevian: would be nice to add comments explaining the differences between these qual lists

src/plan_expand_hypertable.c
```diff
@@ -556,14 +569,8 @@ collect_quals_walker(Node *node, CollectQualCtx *ctx)
...
-		if (!IS_OUTER_JOIN(j->jointype))
-			j->quals = process_quals(j->quals, ctx, true);
-		else if (j->jointype == JOIN_LEFT && IsA(j->rarg, RangeTblRef) &&
-				 ctx->rel->relid == castNode(RangeTblRef, j->rarg)->rtindex)
-			j->quals = process_quals(j->quals, ctx, false);
-		else if (j->jointype == JOIN_RIGHT && IsA(j->larg, RangeTblRef) &&
-				 ctx->rel->relid == castNode(RangeTblRef, j->larg)->rtindex)
-			j->quals = process_quals(j->quals, ctx, false);
+		j->quals = process_quals(j->quals, ctx);
+		collect_join_quals(j->quals, ctx, !IS_OUTER_JOIN(j->jointype));
```
cevian: The code here is sooo much nicer now!

test/expected/join.out
```diff
@@ -0,0 +1,169 @@
...
+       format('%s/results/%s_results_unoptimized.out', :'TEST_OUTPUT_DIR', :'TEST_BASE_NAME') as "TEST_RESULTS_UNOPTIMIZED"
+\gset
+SELECT format('\! diff -u --label "Unoptimized results" --label "Optimized results" %s %s', :'TEST_RESULTS_UNOPTIMIZED', :'TEST_RESULTS_OPTIMIZED') as "DIFF_CMD"
+\gset
+set client_min_messages to warning;
+\ir :TEST_LOAD_NAME
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+CREATE TABLE INT2_TBL(f1 int2);
```
cevian: nit: maybe make these tables hypertables as well? Dunno

test/sql/include/join_query.sql
```diff
@@ -0,0 +1,2381 @@
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+
```
cevian: nit: add a comment that this is taken and modified from pg source just to give credit where it is due.

test/sql/include/join_load.sql
```diff
@@ -0,0 +1,146 @@
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+
+
```
cevian: nit: add a comment that this is taken and modified from pg source just to give credit where it is due.

PR #1422, opened 2019-09-01
Fix background worker segfaults

svenklemm:
The background worker function to check the owner of a job did not have
proper error handling when the object referenced by a job did not exist
leading to a segfault in that case.  This patch adds proper checking of
return values and errors when the object cannot be found.


svenklemm:
There are 2 ways this could be hit, as a race condition when deleting the object or by manually messing with our catalog, deleting the referenced objects of a job but leaving the job itself untouched.

Writing a test for the first is probably hard and might be prevented by your bgw PR, I dont think we want a test where we intentionally corrupt our catalog.

PR #1424, opened 2019-09-02
Fix ChunkAppend LIMIT pushdown

svenklemm:
Do not push down limit to subplan when the query has operations
that modify the number of tuples returned.  Even if we know a
hard limit overall, it doesn't apply if the query has any
grouping/aggregation operations, or SRFs in the tlist.


test/sql/include/append_query.sql
```diff
@@ -335,3 +335,17 @@ PREPARE prep AS SELECT device_id, time FROM (SELECT device_id, time FROM metrics
 :PREFIX EXECUTE prep;
 DEALLOCATE prep;
 
+-- test LIMIT pushdown
```
cevian: would be nice to add a test in this section for when LIMIT /is/ pushed down.

PR #1430, opened 2019-09-09
Release 1.4.2

svenklemm:
This maintenance release contains bugfixes since the 1.4.1 release. We deem it medium
priority for upgrading.

In particular the fixes contained in this maintenance release address 2 potential
segfaults and no other security vulnerabilities. The bugfixes are related to
background workers, OUTER JOINs, ordered append on space partitioned hypertables
and expression indexes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,23 @@
 `psql` with the `-X` flag to prevent any `.psqlrc` commands from
 accidentally triggering the load of a previous DB version.**
 
-## 1.4.2 (unreleased)
+## 1.4.2 (2019-09-11)
+
+This maintenance release contains bugfixes since the 1.4.1 release. We deem it medium
+priority for upgrading.
+
+In particular the fixes contained in this maintenance release address 2 potential
+segfaults and no other security vulnerabilities. The bugfixes are related to
+background workers, OUTER JOINs, ordered append on space partitioned hypertables
+and expression indexes.
 
 **Bugfixes**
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,7 +21,6 @@ accidentally triggering the load of a previous DB version.**
 * #1408 Fix ChunkAppend space partitioning support for ordered append
 * #1420 Fix OUTER JOIN qual propagation
 * #1422 Fix background worker segfaults
```
mfreed: Is 1423 going in 1.4.2 or not?  If so, should the notes be updated to say "release address one potential segfault" instead of 2?

Also, might be nice to have bugfix description point out where segfault happens in consistent manner so you can scan, which we can use going forward in our release notes:

e.g., `Fix background worker SOMETHIGN (potential segfault)` and `Fix SOMETHING on ARM/32-bit builds (potential segfault).`

Otherwise, LGTM.

svenklemm: 1.4.x is not affected by #1423 but the count is still correct because #1392 was a segfault too

mfreed: Sounds good.  I would then just reiterate that it would be a good thing to have bufgfix description mention that, e.g., `#1392 Fix cagg_agg_validate expression handling (potential segfault)`

I think this is useful when readers see the notes to determine whether upgrade is important for them, i.e., if they aren't using cagg or are ARM32, then nothing here is a segfault.

PR #1433, opened 2019-09-10
Make ChunkAppend parallel aware

svenklemm:
This patch makes ChunkAppend parallel aware allowing ChunkAppend
to coordinate parallel workers.

### DIFF
--- a/src/chunk_append/exec.c
+++ b/src/chunk_append/exec.c
@@ -5,6 +5,7 @@
  */
 
 #include <postgres.h>
+#include <fmgr.h>
 #include <miscadmin.h>
 #include <executor/executor.h>
 #include <executor/nodeSubplan.h>
@@ -27,17 +28,24 @@
 #include "chunk_append/exec.h"
 #include "chunk_append/explain.h"
 #include "chunk_append/planner.h"
+#include "loader/lwlocks.h"
 #include "compat.h"
 
 #define INVALID_SUBPLAN_INDEX -1
 #define NO_MATCHING_SUBPLANS -2
 
-static void choose_next_subplan(ChunkAppendState *state);
-
 static TupleTableSlot *chunk_append_exec(CustomScanState *node);
 static void chunk_append_begin(CustomScanState *node, EState *estate, int eflags);
 static void chunk_append_end(CustomScanState *node);
 static void chunk_append_rescan(CustomScanState *node);
+static Size chunk_append_estimate_dsm(CustomScanState *node, ParallelContext *pcxt);
+static void chunk_append_initialize_dsm(CustomScanState *node, ParallelContext *pcxt,
+										void *coordinate);
+#if !PG96
+static void chunk_append_reinitialize_dsm(CustomScanState *node, ParallelContext *pcxt,
+										  void *coordinate);
+#endif
+static void chunk_append_initialize_worker(CustomScanState *node, shm_toc *toc, void *coordinate);
 
 static CustomExecMethods chunk_append_state_methods = {
 	.BeginCustomScan = chunk_append_begin,
@@ -45,7 +53,17 @@ static CustomExecMethods chunk_append_state_methods = {
 	.EndCustomScan = chunk_append_end,
 	.ReScanCustomScan = chunk_append_rescan,
 	.ExplainCustomScan = chunk_append_explain,
+	.EstimateDSMCustomScan = chunk_append_estimate_dsm,
+	.InitializeDSMCustomScan = chunk_append_initialize_dsm,
+#if !PG96
+	.ReInitializeDSMCustomScan = chunk_append_reinitialize_dsm,
+#endif
+	.InitializeWorkerCustomScan = chunk_append_initialize_worker,
 };
+
+static void choose_next_subplan_non_parallel(ChunkAppendState *state);
+static void choose_next_subplan_for_leader(ChunkAppendState *state);
+static void choose_next_subplan_for_worker(ChunkAppendState *state);
 
 static List *constify_restrictinfos(PlannerInfo *root, List *restrictinfos);
 static bool can_exclude_chunk(List *constraints, List *restrictinfos);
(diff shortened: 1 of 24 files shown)


src/chunk_append/exec.c
```diff
@@ -331,7 +364,7 @@ chunk_append_exec(CustomScanState *node)
 		 * more subplans, return the empty slot set up for us by
 		 * chunk_append_begin.
 		 */
-		choose_next_subplan(state);
+		state->choose_next_subplan(state);
```
cevian: is the comment above this and below this out of date now? We don't return the empty slot here right?

src/chunk_append/exec.c
```diff
@@ -366,6 +396,30 @@ choose_next_subplan(ChunkAppendState *state)
...
+{
+	state->current = NO_MATCHING_SUBPLANS;
+}
+
+static void
+choose_next_subplan_for_worker(ChunkAppendState *state)
+{
+	LWLockAcquire(state->lock, LW_EXCLUSIVE);
+
+	state->current = state->pstate->last_plan;
```
cevian: I think the code is a bit confusing here. maybe choose_next_subplan should take the current index as a parameter?

src/guc.c
```diff
@@ -40,6 +40,7 @@ bool ts_guc_restoring = false;
 bool ts_guc_constraint_aware_append = true;
 bool ts_guc_enable_ordered_append = true;
 bool ts_guc_enable_chunk_append = true;
+bool ts_guc_enable_parallel_chunk_append = true;
```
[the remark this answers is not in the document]
svenklemm: it will still be parallel safe, but the equivalent guc for postgres is `enable_parallel_append` so i think the name is fine

src/loader/chunk_append.c
```diff
@@ -0,0 +1,56 @@
...
+#include "export.h"
+
+#define CHUNK_APPEND_SHARED_STATE_NAME "ts_chunk_append_shared_state"
+#define CHUNK_APPEND_LWLOCK_TRANCHE_NAME "ts_chunk_append_lwlock_tranche"
+
+/*
+ * since shared memory can only be setup in a library loaded as
+ * shared_preload_libraries we have to setup this struct here
+ */
+typedef struct ChunkAppendSharedState
```
cevian: Given that this wrapper struct seems to be necessary for lw locks, I wonder if it's worth making this general as in struct `TsLWLocks { LWLock *chunk_append_lock }` and then make the rendezvous variable point to the struct not the lock. This will make adding additional LWLocks much easier in the future.

src/chunk_append/exec.c
```diff
@@ -355,17 +385,48 @@ choose_next_subplan(ChunkAppendState *state)
...
+static void
+choose_next_subplan_local(ChunkAppendState *state)
+{
+	state->current = get_next_subplan(state, state->current);
+}
+
+static void
+choose_next_subplan_for_leader(ChunkAppendState *state)
+{
+	state->current = NO_MATCHING_SUBPLANS;
```
cevian: I think this is OK if we cannot end up in a state where we have parallel execution with a leader but no workers. Is that possible? I am a bit worried about being unable to make progress in this case.

src/chunk_append/exec.c
```diff
@@ -27,33 +28,51 @@
...
 	.ExplainCustomScan = chunk_append_explain,
+	.EstimateDSMCustomScan = chunk_append_estimate_dsm,
+	.InitializeDSMCustomScan = chunk_append_initialize_dsm,
+#if !PG96
+	.ReInitializeDSMCustomScan = chunk_append_reinitialize_dsm,
+#endif
+	.InitializeWorkerCustomScan = chunk_append_initialize_worker,
 };
 
+static void choose_next_subplan_local(ChunkAppendState *state);
```
cevian: Nit: I think `choose_next_subplan_non_parallel` is a more clear name.

test/expected/parallel-11.out
```diff
@@ -190,33 +190,32 @@ SELECT histogram(i, 10, 100000, 5) FROM "test";
...
-SET max_parallel_workers_per_gather = 1;
-EXPLAIN (costs off) SELECT count(*) FROM "test" WHERE length(version()) > 0;
-                                QUERY PLAN                                 
----------------------------------------------------------------------------
- Finalize Aggregate
-   ->  Gather
-         Workers Planned: 1
-         ->  Partial Aggregate
-               ->  Result
+:PREFIX SELECT count(*) FROM "test" WHERE length(version()) > 0;
```
cevian: This test has #workers == #chunks can we add a test where #workers > #chunks and #workers < #chunks?

src/chunk_append/exec.c
svenklemm: since postgres allows this i dont think this can happen, but i can make leader participation a GUC similar to postgres

src/loader/lwlocks.h
```diff
@@ -0,0 +1,15 @@
...
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+
+#ifndef TIMESCALEDB_LOADER_LWLOCKS_H
+#define TIMESCALEDB_LOADER_LWLOCKS_H
+
+#define RENDEZVOUS_CHUNK_APPEND_LWLOCK "ts_chunk_append_lwlock"
+
+void ts_chunk_append_shmem_startup(void);
```
cevian: shouldn't these function names be updated to be lwlock specific not chunk append specific

src/loader/lwlocks.c
```diff
@@ -0,0 +1,54 @@
...
+	LWLockAcquire(AddinShmemInitLock, LW_EXCLUSIVE);
+	ts_lwlocks = ShmemInitStruct(TS_LWLOCK_SHM_NAME, sizeof(TSLWLocks), &found);
+	if (!found)
+	{
+		memset(ts_lwlocks, 0, sizeof(TSLWLocks));
+		ts_lwlocks->chunk_append = &(GetNamedLWLockTranche(CHUNK_APPEND_LWLOCK_TRANCHE_NAME))->lock;
+	}
+	LWLockRelease(AddinShmemInitLock);
+
+	/* assign to rendezvous variable so it can be found by other libraries */
```
cevian: would be nice to add explanation for why this variable is to specific locks and not to struct

PR #1447, opened 2019-09-20
Fix runtime exclusion memory leak

### LINKED ISSUE #1442: OOM for backend using run-time pruning
 - OS: CentOS7.6
 - PostgreSQL version (output of `postgres --version`): 11.2
 - TimescaleDB version (output of `\dx` in `psql`): 1.4.2
 - Installation method: source

**Describe the bug**
After upgrading to 1.4.2 from 1.2.2, we immediately noticed some of our queries were crashing the database. Turning off chunk_append 'fixed' the issue. I've been able to pinpoint it to runtime chunk exclusion that happens, which is either very unsuitable to use with a large number of chunks, or has a memory leak somewhere - probably the latter. Memory is freed upon query finish, but it seems it's not freed on a per-tuple basis after every nested loop iteration.
The steps below can be used to reproduce. When running the SELECT query you'll see memory is quickly filling up until you hit an OOM error after a few minutes. By varying the "interval '10s'" part, you can change the number of nested loop iterations - choosing a higher interval will make the query more likely to succeed before running out of memory.

**To Reproduce**
The planning parameters hackery is just to make a short easily reproducible example.. We saw this kind of plan without any changes to planning parameters.

```
create table a(ts timestamptz);
select create_hypertable('public.a', 'ts', chunk_time_interval=>interval '1d');
insert into a select * from generate_series('2019-01-01', '2019-12-31', interval '1d');
set random_page_cost=1;
set enable_mergejoin=off;
set enable_hashjoin=off;
set enable_seqscan=off;
set enable_bitmapscan=off;
explain(analyze)
select * from generate_series('2019-05-01', '2019-05-01 23:00', interval '10s') v
cross join lateral (
   select * from a where a.ts < v and a.ts > v::date limit 1
) a
;
```

The SELECT should have a plan similar to this:

```
Nested Loop  (cost=0.16..190.97 rows=1000 width=16)
  ->  Function Scan on generate_series v  (cost=0.00..10.00 rows=1000 width=8)
  ->  Limit  (cost=0.16..0.16 rows=1 width=8)
        ->  Custom Scan (ChunkAppend) on a  (cost=0.16..14.07 rows=4015 width=8)
              ->  Index Only Scan using _hyper_259_45395_chunk_a_ts_idx on _hyper_259_45395_chunk  (cost=0.16..9.38 rows=11 width=8)
                    Index Cond: ((ts < v.v) AND (ts > (v.v)::date))
              ->  Index Only Scan using _hyper_259_45396_chunk_a_ts_idx on _hyper_259_45396_chunk  (cost=0.16..9.38 rows=11 width=8)
                    Index Cond: ((ts < v.v) AND (ts > (v.v)::date))
... many more chunks

```
**Expected behavior**
The query runs to completion without excessive memory growth.

**Actual behavior**
Memory keeps on growing as the query executes until there's an OOM kill (if it happens to complete before 100% memory is used, the memory is released when the query has finished, but memory will have grown to very excessive numbers)

svenklemm:
The intermediate expressions used for runtime exclusions would
only be freed at the end of a query leading to excessive memory
usage for nested loop joins with many loops. This patch changes
runtime exclusion to free intermediate expressions as soon as
they are no longer needed.

Fixes #1442

### DIFF
--- a/src/chunk_append/exec.c
+++ b/src/chunk_append/exec.c
@@ -97,6 +97,10 @@ chunk_append_state_create(CustomScan *cscan)
 	state->current = INVALID_SUBPLAN_INDEX;
 	state->choose_next_subplan = choose_next_subplan_non_parallel;
 
+	state->exclusion_ctx = AllocSetContextCreate(CurrentMemoryContext,
+												 "ChunkApppend exclusion",
+												 ALLOCSET_DEFAULT_SIZES);
+
 	return (Node *) state;
 }
 
@@ -154,6 +158,21 @@ do_startup_exclusion(ChunkAppendState *state)
 
 			if (can_exclude_chunk(lfirst(lc_constraints), restrictinfos))
 				continue;
+
+			/*
+			 * if this node does runtime exclusion we keep the constified
+			 * expressions to save us some work during runtime exclusion
+			 */
+			if (state->runtime_exclusion)
+			{
+				List *const_ri_clauses = NIL;
+				foreach (lc, restrictinfos)
+				{
+					RestrictInfo *ri = lfirst(lc);
+					const_ri_clauses = lappend(const_ri_clauses, ri->clause);
+				}
+				ri_clauses = const_ri_clauses;
+			}
 		}
 
 		filtered_children = lappend(filtered_children, lfirst(lc_plan));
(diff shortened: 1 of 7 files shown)


svenklemm:
Changes a query from ooming on my 16GB laptop to making no dent in memory usage, but it all depends on the query.

PR #1449, opened 2019-09-23
Memory leak fixes

JLockerman:
This series of commits fix memory leaks across the database.
Note that the PR is still WIP, more commits may be added if more leaks are found.

### DIFF
--- a/src/bgw/job.c
+++ b/src/bgw/job.c
@@ -125,13 +125,18 @@ ts_bgw_job_start(BgwJob *job, Oid user_uid)
 {
 	int32 job_id = Int32GetDatum(job->fd.id);
 	StringInfo si = makeStringInfo();
+	BackgroundWorkerHandle *bgw_handle;
 
 	/* Changing this requires changes to ts_bgw_job_entrypoint */
 	appendStringInfo(si, "%u %d", user_uid, job_id);
 
-	return ts_bgw_start_worker(job_entrypoint_function_name,
-							   NameStr(job->fd.application_name),
-							   si->data);
+	bgw_handle = ts_bgw_start_worker(job_entrypoint_function_name,
+									 NameStr(job->fd.application_name),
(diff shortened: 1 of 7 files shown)


test/sql/.gitignore
```diff
@@ -1,10 +1,12 @@
 /agg_bookends-*.sql
+/agg_bookends_optimized-*.sql
```
svenklemm: this should be removed and the other gitignore change too

src/bgw/job.c
```diff
@@ -613,7 +628,10 @@ ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
 										TXN_LOCK,
 										/* block */ false);
 		if (job != NULL)
+		{
 			ts_bgw_job_stat_mark_end(job, JOB_FAILURE);
+			pfree(job);
```
cevian: let's set job = null after the pfree here

src/bgw/job.c
```diff
@@ -636,6 +654,9 @@ ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
 	ts_bgw_job_stat_mark_end(job, res);
 	CommitTransactionCommand();
 
+	if (job != NULL)
+		pfree(job);
```
cevian: also maybe job = NULL here as well?

src/bgw/job.c
```diff
@@ -601,6 +609,13 @@ ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
 			AbortCurrentTransaction();
 		StartTransactionCommand();
 
+		/* Free the old job if it exists, it's no longer needed, and since it's
+		 * in the TopMemoryContext it won't be freed otherwise.
+		 */
+
+		if (job != NULL)
+			pfree(job);
```
cevian: should we set job = null after the pfree?

PR #1450, opened 2019-09-23
Print logs from all test suites on sanitizer failure

svenklemm:


.travis.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         - docker exec -u postgres -it docker_arm_emulator find /build -name postmaster.log -exec cat {} +
       script:
       # allow 50 mins to run
-        - travis_wait 50 docker exec -u postgres docker_arm_emulator /bin/bash -c "cd /build && make installcheck IGNORES='multi_transaction_indexing bgw_db_scheduler bgw_job_delete continuous_aggs_insert continuous_aggs_bgw'"
+        - travis_wait 50 docker exec -u postgres docker_arm_emulator /bin/bash -c "cd /build && make -k installcheck IGNORES='multi_transaction_indexing bgw_db_scheduler bgw_job_delete continuous_aggs_insert continuous_aggs_bgw'"
```
JLockerman: Does this work? I thought the sanitizers tend to bring down the DB when they fire, so the latter tests might not be able to run at all?

PR #1451, opened 2019-09-23
Fix postgres version reported by telemetry

svenklemm:
Telemetry would send the postgres version the extension was compiled
against as postgres_version instead of the postgres version that is
currently running. This patch fixes telemetry to send the version
actually in use.

### DIFF
--- a/src/plan_agg_bookend.c
+++ b/src/plan_agg_bookend.c
@@ -48,6 +48,7 @@
 #include "optimizer/tlist.h"
 #include "parser/parsetree.h"
 #include "parser/parse_clause.h"
+#include "parser/parse_func.h"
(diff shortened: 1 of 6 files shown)


svenklemm:
there is SERVER_VERSION_NAME psql variable but that won't help us here, server_version_string has a lot of additional information like platform, compiler, ... so i think using server_version_num is the easiest to get what we need. The underlying variable is static so we can't access it directly and all the other version specific stuff are macros which we need to avoid because we dont want the version it is compiled against.

JLockerman:
@svenklemm personally, I would love that info, but fair enough, will approve after the last comment is added

svenklemm:
the full version string could indeed be useful but that would have to be added in a separate field

PR #1454, opened 2019-09-25
Enable multiple worker per child for ChunkAppend

svenklemm:
This patch changes ChunkAppend to assign multiple worker per child
if the number of worker exceeds the number of children.

### DIFF
--- a/src/chunk_append/exec.c
+++ b/src/chunk_append/exec.c
@@ -24,6 +24,8 @@
 #include <utils/memutils.h>
 #include <utils/typcache.h>
 
+#include <math.h>
+
 #include "chunk_append/chunk_append.h"
 #include "chunk_append/exec.h"
 #include "chunk_append/explain.h"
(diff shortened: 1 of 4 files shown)


src/chunk_append/exec.c
```diff
@@ -442,8 +444,25 @@ choose_next_subplan_for_worker(ChunkAppendState *state)
...
+		state->pstate->workers_last_plan < state->pstate->workers_per_child)
+	{
+		/*
+		 * if we have more workers then children we will assign
+		 * multiple workers per child
+		 */
+		Assert(state->pstate->workers_per_child > 1);
+		Assert(state->pstate->workers_last_plan >= 1);
+
+		state->current = state->pstate->last_plan;
```
cevian: we will assign all remaining workers to the last plan? This looks wrong? Shouldn't we reassign to the subplan with fewest workers?

svenklemm: we dont, we assign at most workers_per_child to a child

PR #1460, opened 2019-10-03
Test new docker with latest libpq

k-rus:
Test of updated docker for ARM 11.0. Will be removed.


k-rus:
The test succeeded and the docker was updated.

PR #1472, opened 2019-10-09
Disable autovacuum in tests

k-rus:
The tests do not expect that autovacuum is run. The most of the tests
finishes before it runs, but some ARM tests might fail due if autovacuum
is executed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -74,6 +74,8 @@ build_script:
 
     Add-Content "C:\Program Files\postgresql\10\data\postgresql.conf" "max_worker_processes=16"
 
+    Add-Content "C:\Program Files\postgresql\10\data\postgresql.conf" "autovacuum=false"
+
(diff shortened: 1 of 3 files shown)


k-rus:
Successful build with ARM tests enabled: https://travis-ci.org/timescale/timescaledb/builds/595649303?utm_source=github_status&utm_medium=notification

k-rus:
> needs to be set in test/postgresql.conf too for the non-tsl tests and in appveyor.yml for the windows tests

Apologise that I missed to do it. It is obviously needed to be fixed there too.

k-rus:
I think I updated all places. Do I need to update [pgtest.conf](https://github.com/timescale/timescaledb/blob/master/test/pgtest.conf)?

svenklemm:
> I think I updated all places. Do I need to update [pgtest.conf](https://github.com/timescale/timescaledb/blob/master/test/pgtest.conf)?

pgtest.conf is for the postgres testsuite, you could set it there aswell but i'll leave that up to you

PR #1479, opened 2019-10-11
Add next_start option to alter_job_schedule

cevian:
Add the option to set the next start time on a job in the
alter job schedule function. This also adds the ability
to pause jobs by setting next_start to 'infinity'

Also fix the enterprise licence check to only activate for
enterprise jobs.

### DIFF
--- a/tsl/test/expected/tsl_tables.out
+++ b/tsl/test/expected/tsl_tables.out
@@ -776,73 +776,152 @@ select add_reorder_policy('test_table', 'test_table_time_idx') as job_id \gset
 
 -- No change
 select * from alter_job_schedule(:job_id);
- job_id | schedule_interval | max_runtime | max_retries | retry_period 
---------+-------------------+-------------+-------------+--------------
-   1014 | @ 84 hours        | @ 0         |          -1 | @ 1 day
+ job_id | schedule_interval | max_runtime | max_retries | retry_period | next_start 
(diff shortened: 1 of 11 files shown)


tsl/test/expected/tsl_tables.out
```diff
@@ -814,73 +814,152 @@ select add_reorder_policy('test_table', 'test_table_time_idx') as job_id \gset
...
+   1014 | @ 20 mins         | @ 7 mins 36 secs |          20 | @ 33 hours   | Tue Jan 01 01:01:01 2002 PST
+(1 row)
+
+--set the last_finish manually
+\c :TEST_DBNAME :ROLE_SUPERUSER
+UPDATE _timescaledb_internal.bgw_job_stat SET last_finish = '2003-01-01:01:01:01' WHERE job_id = :job_id;
+\c :TEST_DBNAME :ROLE_DEFAULT_PERM_USER
+--not changing the interval doesn't change the next_start
+select * from alter_job_schedule(:job_id, schedule_interval=>'20 min');
+WARNING:  Timescale License expired
```
davidkohn88: we probably shouldn't check license expiration on calls of this function if we've taken it out of enterprise

davidkohn88: but not sure if that's true/if we do it elsewhere...

PR #1480, opened 2019-10-11
Set default retry_period for jobs to 5 minutes

davidkohn88:
The default retry_period for jobs was set too high, causing jobs
that failed because of lack of workers to be restarted too infrequently,
this sets it significantly lower, to 5 minutes, though exponential
backoff will cause that to be larger as multiple failures stack up.

### DIFF
--- a/tsl/src/bgw_policy/drop_chunks_api.c
+++ b/tsl/src/bgw_policy/drop_chunks_api.c
@@ -26,36 +26,15 @@
 
 /* Default scheduled interval for drop_chunks jobs is currently 1 day (24 hours) */
 #define DEFAULT_SCHEDULE_INTERVAL                                                                  \
-	DatumGetIntervalP(DirectFunctionCall7(make_interval,                                           \
-										  Int32GetDatum(0),                                        \
-										  Int32GetDatum(0),                                        \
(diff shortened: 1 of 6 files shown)


tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -46,15 +46,15 @@
 										  Float8GetDatum(0)))
 /* Right now, there is an infinite number of retries for drop_chunks jobs */
 #define DEFAULT_MAX_RETRIES -1
-/* Default retry period for drop_chunks_jobs is currently 12 hours */
+/* Default retry period for drop_chunks_jobs is currently 5 minutes */
 #define DEFAULT_RETRY_PERIOD                                                                       \
 	DatumGetIntervalP(DirectFunctionCall7(make_interval,                                           \
```
svenklemm: i think you should use interval_in instead of make_interval

davidkohn88: Done, and fixed elsewhere as well.

svenklemm: much better

PR #1481, opened 2019-10-11
Add last_successful_finish to bgw_job_stats

cevian:
This allows people to better monitor the bgw job health. It
indicates when the last time the job made progress was.

### DIFF
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -180,6 +180,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_internal.bgw_job_stat (
     last_start              TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     last_finish             TIMESTAMPTZ NOT NULL,
     next_start              TIMESTAMPTZ NOT NULL,
+    last_successful_finish  TIMESTAMPTZ NOT NULL,
     last_run_success        BOOL        NOT NULL,
     total_runs              BIGINT      NOT NULL,
     total_duration          INTERVAL    NOT NULL,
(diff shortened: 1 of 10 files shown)


sql/pre_install/tables.sql
```diff
@@ -180,6 +180,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_internal.bgw_job_stat (
     last_start              TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     last_finish             TIMESTAMPTZ NOT NULL,
     next_start              TIMESTAMPTZ NOT NULL,
+    last_successful_finish  TIMESTAMPTZ NOT NULL,
```
svenklemm: OK this is defined NOT NULL but the output below shows this without value

cevian: The output below is from a view with a left join on this table

test/expected/bgw_db_scheduler.out
```diff
@@ -1135,9 +1141,9 @@ SELECT wait_for_job_1_to_run(2);
...
-   1026 | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.25 1999 PST | t                |          2 | @ 0            |               2 |              0 |             0 |                    0 |                   0
+ job_id |           last_start            |           last_finish           |           next_start            |     last_successful_finish      | last_run_success | total_runs | total_duration | total_successes | total_failures | total_crashes | consecutive_failures | consecutive_crashes 
+--------+---------------------------------+---------------------------------+---------------------------------+---------------------------------+------------------+------------+----------------+-----------------+----------------+---------------+----------------------+---------------------
+   1026 | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.25 1999 PST | Fri Dec 31 16:00:00.15 1999 PST | t                |          2 | @ 0            |               2 |              0 |             0 |                    0 |                   0
```
svenklemm: This will fail in slower test environments leading to flaky tests, you should either cast to date or run through time_bucket

cevian: I don't think so since its the same as last_finish, which hasn't been flaky because it relies on our mock test timer and not clock time.

PR #1484, opened 2019-10-12
Bump the postgres version PR tests run against

svenklemm:
Changes the PR tests to run against 9.6.15, 10.10 and 11.5


RobAtticus:
Can I ask why we're bumping those and whether we should also bump our docs recommended versions? I think those two things should stay in sync, otherwise if there's a gap (docs recommend 11.4 currently), we may not be aware that we've stopped working on 11.4 because its not tested.

PR #1487, opened 2019-10-15
Only use local clang-format 7 or 8

svenklemm:
clang-format > 8 has slightly different oppinion about source code
formatting than earlier versions, unfortunately some of the new
options cannot be set in a backwardscompatible way so we fall back
to docker for newer versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -277,8 +277,8 @@ if (CLANG_FORMAT)
     message(FATAL_ERROR "Could not parse clang-format version ${CLANG_FORMAT_VERSION_OUTPUT}")
   endif ()
 
-  if(${CMAKE_MATCH_1} LESS "7")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -277,7 +277,7 @@ if (CLANG_FORMAT)
     message(FATAL_ERROR "Could not parse clang-format version ${CLANG_FORMAT_VERSION_OUTPUT}")
   endif ()
 
-  if(${CMAKE_MATCH_1} LESS "7")
+  if((${CMAKE_MATCH_1} LESS "7") OR (${CMAKE_MATCH_1} GREATER "8"))
     message(WARNING "clang-format is out of date (at least version 7 required)")
```
cevian: can we change the warning message? It's wrong for > 8.

PR #1514, opened 2019-11-04
Run 32 bit tests on i386 instead of ARM emulation

svenklemm:
The main reason to run ARM tests was not to identify issues with
ARM but to identify 32 bit issues e.g. int64 as pointer instead
type by value. Those issues don't need ARM emulation but can be
tested with i386 which is much faster.


k-rus:
@svenklemm Am I correct that Travis build failed on flaky tests?

PR #1525, opened 2019-11-07
Update to Postgres 12

JLockerman:
Currently all of the easy stuff (different function signatures and the like) is done, what remains is the more invasive changes, such as adding a compatibility layer to bridge the new TupleTableSlot-based and old tuple-based code.


src/chunk_adaptive.c
```diff
@@ -135,7 +135,7 @@ typedef enum MinMaxResult
 static MinMaxResult
 minmax_heapscan(Relation rel, Oid atttype, AttrNumber attnum, Datum minmax[2])
 {
-	HeapScanDesc scan;
+	TableScanDesc scan;
```
erimatnor: Should this maybe be `RelationScanDesc` given `heap_open` -> `relation_open`

src/bgw/job.c
```diff
@@ -738,7 +738,7 @@ ts_bgw_job_insert_relation(Name application_name, Name job_type, Interval *sched
 	CatalogSecurityContext sec_ctx;
 	bool nulls[Natts_bgw_job] = { false };
 
-	rel = heap_open(catalog_get_table_id(catalog, BGW_JOB), RowExclusiveLock);
+	rel = relation_open(catalog_get_table_id(catalog, BGW_JOB), RowExclusiveLock);
```
erimatnor: I assume the PG12 gets rid of `heap_open` and that's why we're making this change? Otherwise `heap_open` provided some protections against accidentally opening an index instead of a heap/table.

I would also have expected changing a lot of `heap_close` to `relation_close`, but I don't see that.

src/chunk_adaptive.c
```diff
@@ -146,7 +146,7 @@ minmax_heapscan(Relation rel, Oid atttype, AttrNumber attnum, Datum minmax[2])
 	if (NULL == tce || !OidIsValid(tce->cmp_proc))
 		elog(ERROR, "no comparison function for type %u", atttype);
 
-	scan = heap_beginscan(rel, GetTransactionSnapshot(), 0, NULL);
+	scan = table_beginscan(rel, GetTransactionSnapshot(), 0, NULL);
```
erimatnor: Same as above: maybe this should be called `relation_beginscan`?

src/chunk_append/planner.c
```diff
@@ -17,15 +17,20 @@
 #include <optimizer/prep.h>
 #include <optimizer/subselect.h>
 #include <optimizer/tlist.h>
-#include <optimizer/var.h>
 #include <parser/parsetree.h>
 
+#include "compat.h"
+#if PG12_LT
+#include <optimizer/var.h> /* f09346a */
```
erimatnor: What does the hash number mean here (commit ID)? Please clarify.

src/chunk_append/exec.h
```diff
@@ -9,7 +9,14 @@
 #include <postgres.h>
 #include <nodes/bitmapset.h>
 #include <nodes/extensible.h>
+
+#include "compat.h"
+
+#if PG12_LT /* nodes/relation.h renamed in fa2cf16 */
```
erimatnor: Does the hash-ID here refer to a commit? Is it a commit in PG or in our code? Please clarify. 

Also, maybe better with a link in that case.

src/compat.h
```diff
@@ -443,6 +545,31 @@ get_attname_compat(Oid relid, AttrNumber attnum, bool missing_ok)
...
 
+/* index_get_next
+ *
+ * index_get no longer exists as of c2fe139. As tables are now generalized to
+ * cover non-heap-based things, it is incorrect to return a HeapTuple in the
+ * general case. To allow us to migrate to the new APIs more incrementally, we
+ * provide our own, temporary, implementation of index_get_next, which should be
+ * removed at some point when all our usages are replaced with index_getnext_slot
+ */
+// #if PG12
```
erimatnor: What's this commented code? Also C++ style comments.

src/compat.h
```diff
@@ -578,6 +718,19 @@ get_attname_compat(Oid relid, AttrNumber attnum, bool missing_ok)
 #define PreventInTransactionBlock PreventTransactionChain
 #endif
 
+/*
+ * table_beginscan
+ * PG12 generalizes table scans to those no directly dependant on the heap.
```
erimatnor:
```suggestion
 * PG12 generalizes table scans to those not directly dependent on the heap.
```

Not sure I understand this formulation either.

src/compat.h
```diff
@@ -578,6 +718,19 @@ get_attname_compat(Oid relid, AttrNumber attnum, bool missing_ok)
...
+ * These are the functions we should generally use for version after 12. For
```
erimatnor:
```suggestion
 * These are the functions we should generally use for PG12 and later versions. For
```

src/tablespace.c
```diff
@@ -310,7 +310,7 @@ tablespace_insert(int32 hypertable_id, const char *tspcname)
 	Relation rel;
 	int32 id;
 
-	rel = heap_open(catalog_get_table_id(catalog, TABLESPACE), RowExclusiveLock);
+	rel = relation_open(catalog_get_table_id(catalog, TABLESPACE), RowExclusiveLock);
 	id = tablespace_insert_relation(rel, hypertable_id, tspcname);
 	heap_close(rel, RowExclusiveLock);
```
erimatnor: should this also change to `relation_close`?

src/bgw_policy/chunk_stats.c
```diff
@@ -114,7 +116,7 @@ ts_bgw_policy_chunk_stats_insert(BgwPolicyChunkStats *chunk_stats)
 {
 	Catalog *catalog = ts_catalog_get();
 	Relation rel =
-		heap_open(catalog_get_table_id(catalog, BGW_POLICY_CHUNK_STATS), RowExclusiveLock);
+		relation_open(catalog_get_table_id(catalog, BGW_POLICY_CHUNK_STATS), RowExclusiveLock);
 
 	ts_bgw_policy_chunk_stats_insert_with_relation(rel, chunk_stats);
 	heap_close(rel, RowExclusiveLock);
```
erimatnor: should this be `relation_close`? Otherwise mismatch with `heap_open`

src/chunk_adaptive.c
JLockerman: `TableScanDesc` is the name postgres chose ¯\_(ツ)_/¯

src/bgw/job.c
JLockerman: yup, the `heap_*` functions were removed. I'm also surprised I haven't seen errors on `heap_close` considering it's also removed, probably, just haven't gotten it to compile enough yet.

src/chunk_adaptive.c
JLockerman: also a Postgres name

src/compat.h
JLockerman: These are TODOs, I'm not sure yet if the code is right. It's deliberately mis-styled as preventions from going in.

src/chunk_adaptive.c
erimatnor: Ahh, of course. Didn't realize this was a PostgreSQL type

src/bgw/job.c
fvannee: Why not use table_open (and index_open for indices) instead? That still contains the check

fvannee: Ah I see that has been changed in the latest commit already

src/planner.c
```diff
@@ -422,6 +435,949 @@ timescaledb_set_rel_pathlist_query(PlannerInfo *root, RelOptInfo *rel, Index rti
 	return;
 }
 
+#if PG12
+
+static void
+set_rel_pathlist(PlannerInfo *root, RelOptInfo *rel, Index rti, RangeTblEntry *rte);
+
+
+/* copied from allpaths.c */
```
k-rus: Will it be good idea to move all copied code into a separate file? The rational is that the file is too big already and it will be easier to distinguish between the code origin.

src/planner.c
```diff
@@ -422,6 +435,949 @@ timescaledb_set_rel_pathlist_query(PlannerInfo *root, RelOptInfo *rel, Index rti
...
+					/* Sampled relation */
+					set_tablesample_rel_size(root, rel, rte);
+				}
+				else
+				{
+					/* Plain relation */
+					set_plain_rel_size(root, rel, rte);
+				}
+				break;
+			case RTE_SUBQUERY:
```
k-rus: Was this modification necessary? I see that the code between 11.4 and 12.1 is the same or very similar here.

src/planner.c
```diff
@@ -422,6 +435,949 @@ timescaledb_set_rel_pathlist_query(PlannerInfo *root, RelOptInfo *rel, Index rti
...
+		if(is_rte_hypertable(in_rte) && !in_rte->inh)
+		{
+			RelOptInfo *in_rel = root->simple_rel_array[i];
+			expand_hypertable_inheritance(root, in_rte->relid, in_rte->inh, in_rel);
+			//TODO move this back into ts_plan_expand_hypertable_chunks
+			in_rte->inh = true;
+			reenabled_inheritance = true;
+			//FIXME redo set_rel_consider_parallel
+			if (in_rel != NULL && in_rel->reloptkind == RELOPT_BASEREL)
+				set_rel_size(root, in_rel, i, in_rte);
```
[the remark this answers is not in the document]
JLockerman: Unfortunately not, IIRC the original function is marked `static`

src/planner.c
JLockerman: It's not really necessary, it's just that these kinds should never appear as chunks. (the reason this fuction is copied in isn't because we need to make these modifications, they're just sanity-checking anyway, but because the original postgres function is marked `static`, and only called at a point where hypertables aren't expanded, providing the wrong info)

k-rus:
@JLockerman Can you clarify why you don't use `get_relation_info_hook` to enable inheritance as it is done for PG11? Where is PG12 differ from PG11 in terms of calling `get_relation_info`, which calls the hook?

JLockerman:
```

**Comment by JLockerman** [comment]
```
Before PG12,  planning was organized something like as follows:

 1. construct add RelOptInfo for base and appendrels
 2. add restrict info, joins, etc.
 3. perform the actual planning with make_one_rel

For our optimizations we would expand hypertables in the middle of
step 1; since nothing in the query planner before make_one_rel cared
about the inheritance children, we didn’t have to be too precise about
where we were doing it.

However, with PG12, and the optimizations around declarative
partitioning, postgres now does care about when the children are
expanded, since it wants as much information as possible to perform
partition-pruning. Now planning is organized like:

 1. construct add RelOptInfo for base rels only
 2. add restrict info, joins, etc.
 3. expand appendrels, removing irrelevant declarative partitions
 4. perform the actual planning with make_one_rel

Step 3 always expands appendrels, so when we also expand them during
step 1, the hypertable gets expanded twice, and things in the planner
break.

This commit attempts to solve this problem by keeping the hypertable
root marked as a non-inheritance table until make_one_rel is called,
and only then revealing to postgres that it does in fact have
inheritance children. While this Strategy entails the least code change
on our end, the fact that the first hook we can use to re-enable
inheritance is set_rel_pathlist_hook it does entail a number of
annoyances:

 1. this hook is called after the sizes of tables are calculated, so we
    must recalculate the sizes of all hypertables, as they will not
    have taken the chunk sizes into account
 2. the table upon which the hook is called will have its paths planned
    under the assumption it has no inheritance children, so if it's a
    hypertable we have to replan it's paths

Unfortunately, the code for doing these is static, so we need to copy
them into our own codebase, instead of just using postgres's.
```

File: src/chunk_dispatch_state.c [hunk 14]
```diff
@@ -107,20 +147,32 @@ chunk_dispatch_exec(CustomScanState *node)
...
+			}
+
+			ExecSetSlotDescriptor(state->slot, chunk_desc);
+#endif
 		}
-#if defined(USE_ASSERT_CHECKING) && PG11_GE
+#if defined(USE_ASSERT_CHECKING) && PG11
 		if (state->parent->mt_conflproj != NULL)
 		{
 			TupleTableSlot *slot = get_projection_info_slot_compat(
```
k-rus: `slot` is already declared at the beginning of the function.

src/chunk_dispatch_state.c
```diff
@@ -143,7 +221,47 @@ chunk_dispatch_exec(CustomScanState *node)
 		MemoryContextSwitchTo(old);
 
 		/* Convert the tuple to the chunk's rowtype, if necessary */
-		tuple = ts_chunk_insert_state_convert_tuple(cis, tuple, &slot);
+#if PG12_LT
+		tuple = ts_chunk_insert_state_convert_tuple(cis, tuple, NULL);
```
k-rus: Is it on purpose address to `slot` is not provided, so it is not set? Seems to be different to the original code.

tsl/test/sql/continuous_aggs_query-11.sql
```diff
@@ -0,0 +1,189 @@
+-- This file and its contents are licensed under the Timescale License.
```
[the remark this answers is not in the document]
JLockerman: nope, needs to be added to the `.gitignore`

src/compat.h
```diff
@@ -287,11 +322,125 @@ MakeTupleTableSlotCompat(TupleDesc tupdesc)
...
+#define GetLockConflictsCompat(locktag, lockmode, countp) GetLockConflicts(locktag, lockmode, countp)
+#endif
+
+/*
+ * In PG12 OID columns were removed changing all OID columns in the catalog to
+ * be regular columns. This necessitates passing in the attnum of said column to
+ * any function that wishes to access these columns. In earlier versions, this
+ * parameter can be safely ignored.
+ */
+#if PG96 || PG10 || P11
```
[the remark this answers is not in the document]
JLockerman: 🤦‍♂

src/chunk_append/chunk_append.c
```diff
@@ -198,13 +203,20 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht,
...
+				{
+					Assert(lfirst_oid(lc_oid) ==
+						root->simple_rte_array[((Path *) lfirst(flat))->parent->relid]->relid);
+					merge_childs = lappend(merge_childs, lfirst(flat));
+					flat = lnext(flat);
+				}
 			}
 
-			if (list_length(merge_childs) > 1)
+			if (merge_childs == NIL)
```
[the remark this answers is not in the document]
JLockerman: yeah, so we could write this as
```C
if (list_length(merge_childs) < 1)
{
/*nop*/
}
else if (list_length(merge_childs) == 1 )
{
...
}
else
{
...
}
```
I held off because it was a larger change.

src/chunk_dispatch_state.c
JLockerman: Since this seems to predate my changes, I'm going to hold off editting this unless it breaks PG11

src/chunk_dispatch_state.c
JLockerman: I'll need to double check with PG11, I think I originally had the logic so that `slot` was always `state->slot`, so the store would be redundant, but that may have been on 12.

src/chunk_append/chunk_append.c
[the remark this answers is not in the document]
k-rus: Thanks. Now I see. It wasn't obvious by looking to PR in GH.
May be :
```c
if (list_length(merge_childs) > 1)
{
...
}
else if (list_length(merge_childs) == 1)
{
}
```

It will be less changes in the PR and more obvious to grasp. I guess further refactoring is for a separate issue.

JLockerman:
@k-rus ok, it works on PG11 now.

src/copy.c
```diff
@@ -335,13 +368,28 @@ timescaledb_CopyFrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht)
 
 				/* OK, store the tuple and create index entries for it */
 				heap_insert(resultRelInfo->ri_RelationDesc, tuple, mycid, hi_options, bistate);
```
erimatnor: I believe this needs to use `table_tuple_insert` or otherwise we won't invoke the right insert method for the table access method.

src/chunk_dispatch_state.c
```diff
@@ -83,9 +112,23 @@ chunk_dispatch_exec(CustomScanState *node)
 		dispatch->returning_index = state->parent->mt_whichplan;
 
 		/* Find or create the insert state matching the point */
-		cis = ts_chunk_dispatch_get_chunk_insert_state(dispatch, point, &cis_changed);
-		if (cis_changed)
+
+		cis = ts_chunk_dispatch_get_chunk_insert_state(dispatch, point, &cis_changed,
+#if PG12
+				state->slot->tts_ops
```
erimatnor: Can't we get the `tts_ops` from the chunk relation instead of passing it in?

src/chunk.c
```diff
@@ -158,7 +158,7 @@ ts_chunk_insert_lock(Chunk *chunk, LOCKMODE lock)
 	Catalog *catalog = ts_catalog_get();
 	Relation rel;
 
-	rel = heap_open(catalog_get_table_id(catalog, CHUNK), lock);
+	rel = relation_open(catalog_get_table_id(catalog, CHUNK), lock);
```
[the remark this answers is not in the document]
k-rus: Fixed

src/chunk_append/planner.c
k-rus: It's commit hash for Postgre GitHub repo and goes to https://github.com/postgres/postgres/commit/f09346a9c6218dd239fdf3a79a729716c0d305bd
I can check how it will look like in GH if the link is used.

k-rus: Or may be to write just `postgres/postgres@f09346a`in the code instead of `https://github.com/postgres/postgres/commit/f09346a9c6218dd239fdf3a79a729716c0d305bd`

k-rus:
The final work, which started from this PR, is submitted as https://github.com/timescale/timescaledb/pull/1807

PR #1526, opened 2019-11-07
Fix hypertable model

cevian:


src/hypertable.c
```diff
@@ -120,55 +120,130 @@ ts_hypertable_permissions_check_by_id(int32 hypertable_id)
 	ts_hypertable_permissions_check(table_relid, GetUserId());
 }
 
+static Oid
```
svenklemm: why not LookupFuncName? you want to limit datatypes otherwise you won't handle multiple functions with the same name but different argument types correctly

cevian: Fixed

PR #1542, opened 2019-11-21
Make continuous agg relative to now()

cevian:
Previously, refresh_lag in continuous aggs was calculated
relative to the maximum timestamp in the table. Change the
semantics so that it is relative to now(). This is more
intuitive.

Requires an integer_now function applied to hypertables
with integer-based time dimensions.

### DIFF
--- a/tsl/src/continuous_aggs/materialize.c
+++ b/tsl/src/continuous_aggs/materialize.c
@@ -14,8 +14,6 @@
 #include <executor/spi.h>
 #include <fmgr.h>
 #include <lib/stringinfo.h>
-#include <miscadmin.h>
-#include <utils/hsearch.h>
 #include <storage/lmgr.h>
 #include <utils/builtins.h>
 #include <utils/rel.h>
@@ -25,13 +23,13 @@
 
 #include <scanner.h>
 #include <compat.h>
+#include <interval.h>
 
 #include "chunk.h"
 #include "continuous_agg.h"
 #include "dimension.h"
 #include "hypertable.h"
 #include "hypertable_cache.h"
-#include "export.h"
 #include "partitioning.h"
 
 #include "utils.h"
@@ -277,9 +275,7 @@ get_continuous_agg(int32 mat_hypertable_id)
 	return cagg;
 }
 
-static bool hypertable_get_min_and_max(SchemaAndName hypertable, Name time_column,
-									   int64 search_start, Oid time_type, int64 *min_out,
-									   int64 *max_out);
+static int64 hypertable_get_min(SchemaAndName hypertable, Name time_column, Oid time_type);
 
 static int64
 get_materialization_end_point_for_table(int32 raw_hypertable_id, int32 materialization_id,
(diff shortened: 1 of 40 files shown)


JLockerman:
@cevian does this break continuous aggregates for existing tables with an integer time type? Can we add a fallback integer_now-function to such tables that uses the maximum time-value as `now()`?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -298,31 +292,19 @@ get_materialization_end_point_for_table(int32 raw_hypertable_id, int32 materiali
...
-			elog(INFO,
-				 "new materialization range not found for %s.%s (time column %s): no new data",
-				 NameStr(*hypertable.schema),
-				 NameStr(*hypertable.name),
-				 NameStr(time_column_name));
-		*materializing_new_range = false;
-		return old_completed_threshold;
+		/* If there is no completion threshold yet set, find the minimum value stored in the
+		 * hypertable */
+		start_time = hypertable_get_min(hypertable, &time_column_name, time_column_type);
```
JLockerman: this function call seems to be unconditional, unlike the comment?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -298,31 +292,19 @@ get_materialization_end_point_for_table(int32 raw_hypertable_id, int32 materiali
 	Dimension *time_column = hyperspace_get_open_dimension(raw_table->space, 0);
 	NameData time_column_name = time_column->fd.column_name;
 	Oid time_column_type = ts_dimension_get_partition_type(time_column);
-	bool found_new_tuples = false;
+	int64 now_time = ts_get_now_internal(time_column);
```
JLockerman: this function could be set to find `max` when there's an integer time-type and no integer `now` function, which would prevent this PR from breaking current users.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -392,78 +374,29 @@ get_materialization_end_point_for_table(int32 raw_hypertable_id, int32 materiali
...
-						 "SELECT max(%s), min(%s) FROM %s.%s WHERE %s >= %s",
-						 quote_identifier(NameStr(*time_column)),
-						 quote_identifier(NameStr(*time_column)),
-						 quote_identifier(NameStr(*hypertable.schema)),
-						 quote_identifier(NameStr(*hypertable.name)),
-						 quote_identifier(NameStr(*time_column)),
-						 quote_literal_cstr(search_start_str));
-	}
+	appendStringInfo(command,
+					 "SELECT min(%s) FROM %s.%s",
```
JLockerman: how come the removal of `WHERE time >= old_completed_threshold`?

tsl/src/continuous_aggs/materialize.c
cevian: huh? line 202 is `if (start_time == PG_INT64_MIN)`

tsl/src/continuous_aggs/materialize.c
cevian: discussed offline

tsl/src/continuous_aggs/materialize.c
cevian: this is only called if old_completed_threshold is not set

PR #1544, opened 2019-11-23
Fix crash when partializing agg with HAVING

erimatnor:
This change fixes an assertion-based crash that happened when using
the `partialize_agg` function together with HAVING clauses. For instance,

```
SELECT time_bucket('1 day', time), device,
__timescaledb_internal.partialize_agg(avg(temp))
GROUP BY 1, 2
HAVING avg(temp) > 3;
```

would crash because the HAVING clause's aggregate didn't have its
`Aggref` node set to partial aggregate mode.

Regular partial aggregations executed by the planner (i.e., those not
induced by the `partialize_agg` function) have their HAVING aggs
transparently moved to the target list during planning so that the
finalize node can use it when applying the final filter on the
resulting groups. However, it doesn't make much sense to transparently
do that when partializing with `partialize_agg` since it would be odd
to return more columns than requested by the user. Therefore, the
caller would have to do that manually. This, in fact, is also done
when materializing continuous aggregates.

For this reason, HAVING clauses with `partialize_agg` are blocked,
except in cases where the planner transparently reduces the HAVING
expression to a simple filter (e.g., `HAVING device > 3`).

Apart from fixing this issue, this change also refectors some of the
related code and adds tests for some error cases.

### DIFF
--- a/src/plan_partialize.c
+++ b/src/plan_partialize.c
@@ -6,7 +6,13 @@
 #include <postgres.h>
 #include <catalog/pg_type.h>
 #include <nodes/nodeFuncs.h>
+#include <nodes/nodes.h>
+#include <nodes/pg_list.h>
+#include <nodes/relation.h>
 #include <optimizer/planner.h>
+#include <optimizer/cost.h>
+#include <optimizer/var.h>
+#include <optimizer/tlist.h>
 #include <parser/parse_func.h>
 #include <utils/lsyscache.h>
 
@@ -15,15 +21,30 @@
 #include "utils.h"
 
 #define TS_PARTIALFN "partialize_agg"
+
 typedef struct PartializeWalkerState
 {
 	bool found_partialize;
+	bool found_non_partial_agg;
 	bool looking_for_agg;
 	Oid fnoid;
 } PartializeWalkerState;
 
+/*
+ * Look for the partialize function in a target list and mark the wrapped
+ * aggregate as a partial aggregate.
+ *
+ * The partialize function is an expression of the form:
+ *
+ * _timescaledb_internal.partialize_agg(avg(temp))
+ *
+ * where avg(temp) can be replaced by any aggregate that can be partialized.
+ *
+ * When such an expression is found, this function will mark the Aggref node
+ * for the aggregate as partial.
+ */
 static bool
-partialize_function_call_walker(Node *node, PartializeWalkerState *state)
+check_for_partialize_function_call(Node *node, PartializeWalkerState *state)
 {
 	if (node == NULL)
 		return false;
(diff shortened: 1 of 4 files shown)


tsl/test/sql/partialize_finalize.sql
```diff
@@ -182,8 +182,24 @@ select length(_timescaledb_internal.partialize_agg(min(a+1))) from foo;
...
+from foo
+group by 1;
 \set ON_ERROR_STOP 1
 
+--partializing works with HAVING when the planner can effectively
+--reduce it. In this case to a simple filter.
+select time_bucket('1 hour', b) as b, toastval, _timescaledb_internal.partialize_agg(avg(a))
+from foo
+group by b, toastval
+having toastval LIKE 'does not exist';
```
gayyappan: comment doesn't reflect the test. shouldn't this be  `HAVING  ...partialize_agg(avg(a))  > ... ` ?

gayyappan: Could you add 2 more tests:
1.  select ... , partialize_agg( AGGREGATE1), partialize_agg( AGGREGATE2)
 ....
HAVING partialize_agg(AGGREGATE3) ---> all 3 aggregates are different

2. select partialize_agg(AGGREGATE1), partialize_gg(AGGREAGTE2)
...
HAVING partialize_agg(AGGREGATE2)  --->same as 1 in target list

where AGGREGATE1, AGGREGATE2, AGGREGATE3 are different partializable aggregates.

erimatnor: > comment doesn't reflect the test. shouldn't this be `HAVING ...partialize_agg(avg(a)) > ... ` ?

Not sure what you mean. The point of this test is to show that HAVING quals that are filters can be accepted since the planner reduces them to filters on the base rel, thus internally removing the HAVING qual. If there are any remaining HAVING quals, the query will fail since this PR makes it so that HAVING quals aren't accepted together with partializaion.

erimatnor: > Could you add 2 more tests:
> 
>     1. select ... , partialize_agg( AGGREGATE1), partialize_agg( AGGREGATE2)
>        ....
>        HAVING partialize_agg(AGGREGATE3) ---> all 3 aggregates are different
> 
>     2. select partialize_agg(AGGREGATE1), partialize_gg(AGGREAGTE2)
>        ...
>        HAVING partialize_agg(AGGREGATE2)  --->same as 1 in target list
> 
> 
> where AGGREGATE1, AGGREGATE2, AGGREGATE3 are different partializable aggregates.

FWIW, the above expression/queries are non-sensible since HAVING must be a qual (boolean expression). Thus `partialize_agg(avg(temp) > 3)` doesn't works since that returns `bytea` and not a boolean, and, `partialize_agg(avg(temp)) > 3` doesn't work either for similar reasons.

gayyappan: I wanted to have a test similar to the cont. aggregate views. But see that cont. aggregates use finalize_agg (not partialize_agg) in the final query. My bad.

erimatnor: No worries, I am happy to add other tests if you come up with something.

src/plan_partialize.c
```diff
@@ -32,86 +53,147 @@ partialize_function_call_walker(Node *node, PartializeWalkerState *state)
...
+		if (NULL != parse->havingQual)
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("cannot partialize aggregate with HAVING clause"),
+					 errhint(
+						 "Any aggregates in a HAVING clause need to be partialized in the output "
+						 "target list.")));
+
+		partialize_agg_paths(input_rel);
+		partialize_agg_paths(output_rel);
```
svenklemm: this should only be called on either input_rel or output_rel

erimatnor: I was wondering about that, since it is reasonable that only the `output_rel` would actually have any `AggRefs` given that this is the rel that is doing the aggregation. But I am just following what was there before (although I moved this code into a function to minimize duplication). Would be good to get some input from the person that wrote this code (looks like @cevian or @gayyappan according to git blame).

erimatnor: Anyway, I changed this to only act on the `output_rel` and also put a guard in planner.c to only call this in case of `UPPERREL_GROUP_AGG`.

gayyappan: @JLockerman could you comment on this?

JLockerman: @gayyappan I'd defer to sven on this.

gayyappan: Thanks for taking a look @JLockerman.  looped you in as you wrote the code.

erimatnor: I am going to try and clarify the reasoning behind this change according to how I understand this code. 

The `input_rel` is the rel that you'd like to compute something _from_ and to `output_rel` is the _result_. Since we want to partialize aggregates, we should only be interested in the stage that applies aggregations to an `input_rel` to produce an  aggregate `output_rel`, i.e., the  `UPPERREL_GROUP_AGG` stage. In this case, the `input_rel` should be a base or join rel (no aggregates) and the `output_rel` the result of applying an aggregate on top of the base rel in `input_rel`.

In any later stage (after aggregates), the `input_rel` would already have aggregates applied (which we partialized in a previous call to this function), and no aggregates are computed in those stages. Therefore, there should never be a case where we'd have to partialize an `input_rel` (unless we somehow missed to do it on the `output_rel` in a previous step).

Please let me know if I am missing something.

PR #1547, opened 2019-11-25
Documentation revision (first paragraph)

joshuafitzmorris:
I believe the revision clarifies the usefulness of the library more quickly, potentially avoiding confusion for incoming users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,11 @@
 
 ## TimescaleDB
 
-TimescaleDB is an open-source database designed to make SQL scalable for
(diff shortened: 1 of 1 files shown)


mfreed:
Hi @joshuafitzmorris Really appreciate the desire to help, but there's some framing about TimescaleDB that we prefer.  We find that the framing of TimescaleDB as purely an "extension" to those not intimately familiar with PostgreSQL's extension framework to actually lead to confusion, as most extensions are much more minimal.  This is particularly the case as Timescale now offers a distributed, scale-out version.

PR #1551, opened 2019-11-26
Apply max_interval_per_job to invalidations

cevian:
This change the continuous aggregate materialization logic so that
the max_interval_per_job applies to invalidation entries as well
as new ranges in the materialization. The new logic is that the
MIPJ setting limits the sum of work done by the invalidations
and new ranges. Invalidations take precedence so new ranges
are only processed if there is time left over in the MIPJ
budget after all invalidations are done.

This forces us to calculate the invalidation range during the first
transaction. We still delete and/or cut the invalidation entries
in the second transaction. This change also more neatly separates concerns:
all decisions on work to be done happens in the first txn while only
execution happens in the second. Further refactoring could make
this more clear by passing a list of InternalRanges to represent the
work. But this PR is big enough, so that's left to a future refactor.

Note: There is remaining work to be done in breaking up invalidation
entries as created during inserts to constrain the length of the entries.
But that's a separate issue to be addressed in the future.

### DIFF
--- a/tsl/src/continuous_aggs/materialize.c
+++ b/tsl/src/continuous_aggs/materialize.c
@@ -24,6 +24,7 @@
 #include <scanner.h>
 #include <compat.h>
 #include <interval.h>
+#include <scan_iterator.h>
 
 #include "chunk.h"
 #include "continuous_agg.h"
@@ -38,6 +39,35 @@
 #include "continuous_aggs/materialize.h"
 
 /***********************
+ * Time ranges
+ ***********************/
+
+typedef struct InternalTimeRange
+{
+	Oid type;
+	int64 start; /* inclusive */
+	int64 end;   /* exclusive */
+} InternalTimeRange;
+
+typedef struct TimeRange
+{
+	Oid type;
+	Datum start;
+	Datum end;
+} TimeRange;
+
+static bool ranges_overlap(InternalTimeRange invalidation_range,
+						   InternalTimeRange new_materialization_range);
+static TimeRange internal_time_range_to_time_range(InternalTimeRange internal);
+static bool range_expand_right(InternalTimeRange *range, const InternalTimeRange to_cover,
+							   int64 max_length, int64 bucket_width);
+static void range_check(const InternalTimeRange range, int64 bucket_width);
+static int64 range_length(const InternalTimeRange range);
+static bool range_cut_invalidation_entry_left(const InternalTimeRange invalidation_range_knife,
+											  int64 *lowest_modified_value,
+											  int64 *greatest_modified_value,
+											  bool *delete_entry_out);
+/***********************
  * materialization job *
  ***********************/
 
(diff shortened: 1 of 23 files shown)


tsl/src/continuous_aggs/materialize.c
```diff
@@ -950,48 +1069,77 @@ update_materializations(SchemaAndName partial_view, SchemaAndName materializatio
...
+range_expand_right(InternalTimeRange *range, const InternalTimeRange to_cover, int64 max_length)
+{
+	int64 max_end = int64_saturating_add(range->start, max_length);
 
-	if (range->start > range->end)
-		range->start = range->end;
+	Assert(range->start <= to_cover.start);
+	if (max_end > to_cover.start)
+	{
+		range->end = int64_min(to_cover.end, max_end);
```
JLockerman: This does not seem to time_bucket the end correctly?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -521,32 +573,142 @@ drain_invalidation_log(int32 raw_hypertable_id, List **invalidations_out)
...
+												  &invalidation_form->greatest_modified_value,
+												  &delete_entry);
+			if (!delete_entry && invalidation_form->greatest_modified_value >= completed_threshold)
+			{
+				invalidation_form->greatest_modified_value =
+					int64_saturating_sub(completed_threshold, 1);
+				modify_entry = true;
+				delete_entry = (invalidation_form->lowest_modified_value >
+								invalidation_form->greatest_modified_value);
+			}
```
JLockerman: what does this branch do?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -521,32 +573,142 @@ drain_invalidation_log(int32 raw_hypertable_id, List **invalidations_out)
...
-						,
-						scankey /*=scankey*/,
-						1 /*=num_keys*/,
-						scan_del_materialization_invalidation_tuple /*=tuple_found*/,
-						RowExclusiveLock /*=lockmode*/,
-						&scan_state /*=data*/);
+		/* if we don't need to modify this entry we wont need to modify further entries
+		 * because the scan is ordered by lowest_modify_entry ASC*/
+		if (!modify_entry)
+			break;
```
JLockerman: Is this always true? What if you have values above the `completion_threshold` later on? (doesn't affect correctness, just performance)

tsl/src/continuous_aggs/materialize.c
cevian: /*
				 * modify the entry if it is higher than the completion threshold
				 * to only go up to the completion threshold. Nothing above completion
				 * has yet been materialized an thus does not need invalidation
				 */

tsl/src/continuous_aggs/materialize.c
cevian: you are right, changed to continue

tsl/src/continuous_aggs/materialize.c
cevian: good catch fixed

tsl/src/continuous_aggs/materialize.c
```diff
@@ -521,32 +573,142 @@ drain_invalidation_log(int32 raw_hypertable_id, List **invalidations_out)
...
+			entry_range.end = completed_threshold;
+
+		if (!found)
+		{
+			found = true;
+			invalidation_range = entry_range;
+		}
+		else
+		{
+			range_expand_right(&invalidation_range, entry_range, max_interval_per_job);
```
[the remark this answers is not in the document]
cevian: I think having invalidation length + new_length < max_refresh is the more principled rule. We may consider not combining non-overlapping  invalidations and keeping them separate though (then the sum of all invalidations+ne_length < max refresh). I think thats a different pr tho

PR #1552, opened 2019-11-27
Add MacOS PR check using GitHub Actions

erimatnor:
Since Travis doesn't support MacOS X builds, this adds a check that
builds and tests both Debug and Release builds on MacOS X using GitHub
Actions.

Note that GitHub Actions supports build matrixes and can build for
Linux, Windows, and MacOS X. Thus, we could potentially run all our
checks on Actions in the future.

Currently, however, this only builds against a brew-installed
PostgreSQL version and only for PG11 and MacOS X (thus, possibly
without assertions, etc.). However, this is currently covered by other
platforms and we can consider changing this in the future (although
this will slow down builds).

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,0 +1,27 @@
+name: Test
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
+    name: ${{ matrix.build_type }} [${{ matrix.os }}] [PG${{ matrix.pg }}]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        build_type: [ Debug, Release ]
+        os: [ macOS-latest ]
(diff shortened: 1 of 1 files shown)


.github/workflows/ccpp.yml
```diff
@@ -0,0 +1,27 @@
+name: Test
+on:
+  pull_request:
+    branches:
+      - multinode
```
k-rus: Branch `multinode` doesn't exist in this repo. Should this PR be against public master instead?

erimatnor: yes, forgot to update. Thanks for catching.

erimatnor:
Hmm, for some reason the Actions checks aren't showing up. Wondering if I need to resubmit as another PR because I filtered on the wrong branch initially?

PR #1553, opened 2019-11-27
Add MacOS PR check using GitHub Actions

erimatnor:
Since Travis doesn't support MacOS X builds, this adds a check that
builds and tests both Debug and Release builds on MacOS X using GitHub
Actions.

Note that GitHub Actions supports build matrixes and can build for
Linux, Windows, and MacOS X. Thus, we could potentially run all our
checks on Actions in the future.

Currently, however, this only builds against a brew-installed
PostgreSQL version and only for PG11 and MacOS X (thus, possibly
without assertions, etc.). However, this is currently covered by other
platforms and we can consider changing this in the future (although
this will slow down builds).

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,0 +1,28 @@
+# GitHub Actions onfiguration for TimescaleDB
+name: Test
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


erimatnor:
Not sure why GitHub actions aren't running. I suspect it has something to do with the PR coming from a forked repo, but according to the docs there are limitations only with PRs coming from _private_ forked repos.

erimatnor:
Closing for now since GitHub Actions seem to still have issues with running jobs from forks/private branches.

PR #1554, opened 2019-11-28
Notify internal Slack an Travis CRON build fail

k-rus:
Adds notification from Travis CI to an internal Slack channel when a
CRON build fails.

The notification goes to [travis-ci-org channel](https://iobeam.slack.com/archives/CR2GTBLP2). The channel contains an example of a notification.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,17 @@ os:
   - linux
 services:
   - docker
+notifications:
+  slack:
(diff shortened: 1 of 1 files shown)


k-rus:
@svenklemm @erimatnor I updated the RP to notify for CRON jobs executed on master branch only. This will allow to distinguish between multimode CRON job and mast CRON job after rebasing master branch on public, which will include this PR.

PR #1568, opened 2019-12-04
Add ignore_invalidation_older_than to continuous aggs

cevian:
We added a timescaledb.ignore_invalidation_older_than parameter for
continuous aggregatess. This parameter accept a time-interval (e.g. 1
month). if set, it limits the amount of time for which to process
invalidation. Thus, if
	timescaledb.ignore_invalidation_older_than = '1 month'
then any modifications for data older than 1 month from the current
timestamp at insert time will not cause updates to the continuous
aggregate. This limits the amount of work that a backfill can trigger.
This parameter must be >= 0. A value of 0 means that invalidations are
never processed.

When recording invalidations for the hypertable at insert time, we use
the maximum ignore_invalidation_older_than of any continuous agg attached
to the hypertable as a cutoff for whether to record the invalidation
at all. When materializing a particular continuous agg, we use that
aggs  ignore_invalidation_older_than cutoff. However we have to apply
that cutoff relative to the insert time not the materialization
time to make it easier for users to reason about. Therefore,
we record the insert time as part of the invalidation entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ value in the table. This change requires that an integer_now func be set on
 hypertables with integer-based time columns to use continuous aggregates on
 this table.
 
+We added a timescaledb.ignore_invalidation_older_than parameter for continuous 
+aggregatess. This parameter accept a time-interval (e.g. 1 month). if set, 
+it limits the amount of time for which to process invalidation. Thus, if
+timescaledb.ignore_invalidation_older_than = '1 month', then any modifications
+for data older than 1 month from the current timestamp at modification time may 
+not cause continuous aggregate to be updated. This limits the amount of work
+that a backfill can trigger. By default, all invalidations are processed.
+
 ## 1.5.1 (2019-11-12)
 
 This maintenance release contains bugfixes since the 1.5.0 release. We deem it low
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -250,9 +250,14 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_agg (
     direct_view_schema NAME NOT NULL,
     direct_view_name NAME NOT NULL,
     max_interval_per_job BIGINT NOT NULL,
+    ignore_invalidation_older_than BIGINT NOT NULL DEFAULT BIGINT '9223372036854775807',
     UNIQUE(user_view_schema, user_view_name),
     UNIQUE(partial_view_schema, partial_view_name)
 );
+
+CREATE INDEX IF NOT EXISTS continuous_agg_raw_hypertable_id_idx
+    ON _timescaledb_catalog.continuous_agg(raw_hypertable_id);
+
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_agg', '');
 
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_invalidation_threshold(
@@ -273,6 +278,7 @@ SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs
 -- table are performance critical
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_hypertable_invalidation_log(
     hypertable_id INTEGER NOT NULL,
+    modification_time BIGINT NOT NULL, --now time for txn when the raw table was modified
     lowest_modified_value BIGINT NOT NULL,
     greatest_modified_value BIGINT NOT NULL
 );
@@ -286,6 +292,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_materialization_
     materialization_id INTEGER
         REFERENCES _timescaledb_catalog.continuous_agg(mat_hypertable_id)
         ON DELETE CASCADE,
+    modification_time BIGINT NOT NULL, --now time for txn when the raw table was modified
     lowest_modified_value BIGINT NOT NULL,
     greatest_modified_value BIGINT NOT NULL
 );
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -22,3 +22,59 @@ BEGIN
     END IF;
 END
 $BODY$;
+
+
+ALTER TABLE  _timescaledb_catalog.continuous_agg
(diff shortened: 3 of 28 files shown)


sql/pre_install/tables.sql
```diff
@@ -273,6 +278,7 @@ SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs
 -- table are performance critical
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_hypertable_invalidation_log(
     hypertable_id INTEGER NOT NULL,
+    modification_time_internal BIGINT NOT NULL, --time at which the raw table was modified
```
JLockerman: I don't think we need the `internal`, all of this is pretty internal already.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -747,12 +749,25 @@ insert_materialization_invalidation_logs(List *caggs, List *invalidations,
 	{
 		ContinuousAgg *ca = lfirst(lc);
 		int32 cagg_id = ca->data.mat_hypertable_id;
+		int64 ignore_invalidation_older_than = ca->data.ignore_invalidation_older_than;
 		foreach (lc2, invalidations)
 		{
 			Invalidation *entry = (Invalidation *) lfirst(lc2);
+			int64 minimum_invalidation_time =
```
gayyappan: could you add an example at the top of this function to describe/clarify how min invalidation time is used to ignore some invalidations.

tsl/test/expected/continuous_aggs_bgw.out
```diff
@@ -69,8 +69,8 @@ SELECT * FROM timescaledb_information.policy_stats;
 (0 rows)
 
 SELECT * FROM _timescaledb_catalog.continuous_agg;
- mat_hypertable_id | raw_hypertable_id | user_view_schema | user_view_name | partial_view_schema | partial_view_name | bucket_width | job_id | refresh_lag | direct_view_schema | direct_view_name | max_interval_per_job 
--------------------+-------------------+------------------+----------------+---------------------+-------------------+--------------+--------+-------------+--------------------+------------------+----------------------
+ mat_hypertable_id | raw_hypertable_id | user_view_schema | user_view_name | partial_view_schema | partial_view_name | bucket_width | job_id | refresh_lag | direct_view_schema | direct_view_name | max_interval_per_job | ignore_invalidation_older_than
```
gayyappan: do we have a test with timescaledb_information.continuous_aggregates that displays the new parameter?

tsl/test/expected/continuous_aggs_multi.out
```diff
@@ -386,10 +386,126 @@ select * from cagg_1 where timed = 18 ;
...
+INFO:  new materialization range not found for public.continuous_agg_test_ignore_invalidation_older_than (time column timeval): not enough new data past completion threshold (14)
+INFO:  materializing continuous aggregate public.cagg_iia3: nothing to invalidate, no new range
+INFO:  materializing continuous aggregate public.cagg_iia3: no new range to materialize or invalidations found, exiting early
+select * from cagg_iia3 order by 1;
+ timed | maxval 
+-------+--------
+    10 |      5
+    12 |      7
+(2 rows)
+
```
gayyappan: would be good to add select from continuous_aggs_hypertable_invalidation_log here to show the modification_time before/after the new row is picked up.

tsl/test/expected/continuous_aggs_multi.out
```diff
@@ -386,10 +386,126 @@ select * from cagg_1 where timed = 18 ;
...
+refresh materialized view cagg_iia3;
+INFO:  new materialization range not found for public.continuous_agg_test_ignore_invalidation_older_than (time column timeval): not enough new data past completion threshold (14)
+INFO:  materializing continuous aggregate public.cagg_iia3: processing invalidations, no new range
+select * from cagg_iia3 order by 1;
+ timed | maxval 
+-------+--------
+    10 |     20
+    12 |      7
+(2 rows)
+
```
gayyappan: 1. also need invalidation test with update and delete. 2. And test with table that uses timestamp column for partitioning (these tests have integer column).

CHANGELOG.md
```diff
@@ -12,6 +12,14 @@ value in the table. This change requires that an integer_now func be set on
 hypertables with integer-based time columns to use continuous aggregates on
 this table.
 
+We added a timescaledb.ignore_invalidation_older_than parameter for continuous 
+aggregatess. This parameter accept a time-interval (e.g. 1 month). if set, 
+it limits the amount of time for which to process invalidation. Thus, if
+timescaledb.ignore_invalidation_older_than = '1 month', then any modifications
+for data older than 1 month from the current timestamp at insert time will
```
gayyappan: isn't this  - insert/update/delete time?

sql/updates/latest-dev.sql
```diff
@@ -22,3 +22,59 @@ BEGIN
     END IF;
 END
 $BODY$;
+
+
+ALTER TABLE  _timescaledb_catalog.continuous_agg
+    ADD COLUMN  ignore_invalidation_older_than BIGINT NOT NULL DEFAULT BIGINT '9223372036854775807';
```
gayyappan: is this value valid for all systems i.e.32 and 64 bit?

sql/updates/latest-dev.sql
```diff
@@ -22,3 +22,59 @@ BEGIN
...
+    materialization_id         INTEGER
+        REFERENCES _timescaledb_catalog.continuous_agg (mat_hypertable_id)
+            ON DELETE CASCADE,
+    modification_time BIGINT NOT NULL, --time at which the raw table was modified
+    lowest_modified_value      BIGINT NOT NULL,
+    greatest_modified_value    BIGINT NOT NULL
+);
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_materialization_invalidation_log', '');
+--modification_time == INT_MIN to cause these invalidations to be processed
+INSERT INTO _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
```
gayyappan: we need to test this explicitly with some pending invalidations at the time of upgrade and check if these are processed.

tsl/test/expected/continuous_aggs_bgw.out
cevian: added

sql/updates/latest-dev.sql
cevian: yes. The max values for BIGINT doesn't change

PR #1575, opened 2019-12-05
Reorder group-by clause for continuous aggregates

gayyappan:
Modify group by clause on continuous aggr views
    
    Reorder the group by clause to match the ordering of the
    ORDER BY by clause. SELECTs on continuous aggregate queries
    often have an order by clause on the time_bucket column.
    If the order-by does not match the grouping clause,
    the planner inserts an additional Sort node after
    the view evaluation (Aggregate node).
    preprocess_groupclause does this reordering if the ORDER BY clause
    is part of the current subquery being processed. But when we have
    a continuous aggregate view, the order by needs to be derived
    from the outer query. This PR allows the order by from
    outer query to propagate to group clause of continuous aggr. view.
    The rewrite is :
    select * from (select a, b, max(c), min(d) from ...group by a, b)
          order by b;
    is transformed as
     SELECT * from (select a, b, max(c), min(d) from .. group by b, a
      ) order by b;

### DIFF
--- a/src/guc.c
+++ b/src/guc.c
@@ -43,6 +43,7 @@ bool ts_guc_enable_chunk_append = true;
 bool ts_guc_enable_parallel_chunk_append = true;
 bool ts_guc_enable_runtime_exclusion = true;
 bool ts_guc_enable_constraint_exclusion = true;
+bool ts_guc_enable_cagg_reorder_groupby = true;
 TSDLLEXPORT bool ts_guc_enable_transparent_decompression = true;
 int ts_guc_max_open_chunks_per_insert = 10;
 int ts_guc_max_cached_chunks_per_hypertable = 10;
@@ -173,6 +174,17 @@ _guc_init(void)
 							 "Enable transparent decompression",
 							 "Enable transparent decompression when querying hypertable",
 							 &ts_guc_enable_transparent_decompression,
+							 true,
+							 PGC_USERSET,
+							 0,
+							 NULL,
+							 NULL,
+							 NULL);
+
+	DefineCustomBoolVariable("timescaledb.enable_cagg_reorder_groupby",
+							 "Enable group by reordering",
+							 "Enable group by clause reordering for continuous aggregates",
+							 &ts_guc_enable_cagg_reorder_groupby,
 							 true,
 							 PGC_USERSET,
 							 0,
--- a/src/planner.c
+++ b/src/planner.c
@@ -27,6 +27,7 @@
 #include <tcop/tcopprot.h>
 #include <optimizer/plancat.h>
 #include <nodes/nodeFuncs.h>
+#include <parser/analyze.h>
 
 #include <catalog/pg_constraint.h>
 #include "compat.h"
@@ -60,11 +61,19 @@
 void _planner_init(void);
 void _planner_fini(void);
 
+typedef struct TimescaledbWalkerState
+{
+	CmdType cmdtype;
+	Cache *hc;
+} TimescaledbWalkerState;
+
 static planner_hook_type prev_planner_hook;
 static set_rel_pathlist_hook_type prev_set_rel_pathlist_hook;
 static get_relation_info_hook_type prev_get_relation_info_hook;
 static create_upper_paths_hook_type prev_create_upper_paths_hook;
 static bool contain_param(Node *node);
+static void cagg_reorder_groupby_clause(RangeTblEntry *subq_rte, int rtno, List *outer_sortcl,
+										List *outer_tlist);
 
 #define CTE_NAME_HYPERTABLES "hypertable_parent"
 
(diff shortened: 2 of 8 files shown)


src/guc.c
```diff
@@ -43,6 +43,7 @@ bool ts_guc_enable_chunk_append = true;
 bool ts_guc_enable_parallel_chunk_append = true;
 bool ts_guc_enable_runtime_exclusion = true;
 bool ts_guc_enable_constraint_exclusion = true;
+bool ts_guc_enable_contagg_pushdown = true;
```
svenklemm: this lacks proper GUC definition, should probably also add a test that the GUC works

src/guc.c
```diff
@@ -180,6 +181,17 @@ _guc_init(void)
 							 NULL,
 							 NULL);
 
+	DefineCustomBoolVariable("timescaledb.enable_contagg_reorder_groupby",
```
svenklemm: maybe this should be enable_cagg_reorder_groupby, isn't that the abbreviation we usually use for continuous aggs

src/planner.c
```diff
@@ -65,6 +66,7 @@ static set_rel_pathlist_hook_type prev_set_rel_pathlist_hook;
 static get_relation_info_hook_type prev_get_relation_info_hook;
 static create_upper_paths_hook_type prev_create_upper_paths_hook;
 static bool contain_param(Node *node);
+static void pushdown_orderby_to_groupclause(Query *query);
```
svenklemm: i think the naming should be consistent, so this should be reorder_groupby as well

src/planner.c
```diff
@@ -737,6 +754,126 @@ contain_param(Node *node)
...
+			if (outer_sortcl && rte->subquery->groupClause && rte->subquery->sortClause == NIL &&
+				check_cagg_view_rte(rte))
+			{
+				List *new_groupclause = NIL;
+				foreach (lc, outer_sortcl)
+				{
+					SortGroupClause *outer_sc = (SortGroupClause *) lfirst(lc);
+					TargetEntry *outer_tle = get_sortgroupclause_tle(outer_sc, outer_tlist);
+					not_found = true;
+					if (IsA(outer_tle->expr, Var) && (((Var *) outer_tle->expr)->varno == rtno))
```
cevian: nit: to reduce the nesting level and increase readabilitiy. can we change from if(condition) { if (condition) ... not_found = false} to if(!condition) break; ... if(!condition) break ... not_found = false;

cevian: or break out into a separate function

src/planner.c
```diff
@@ -737,6 +754,126 @@ contain_param(Node *node)
...
+						int outer_attno = ((Var *) outer_tle->expr)->varattno;
+						Query *subq = rte->subquery;
+						TargetEntry *subq_tle = list_nth(subq->targetList, outer_attno - 1);
+						if (subq_tle->ressortgroupref > 0)
+						{
+							// get group clause corresponding to this.
+							SortGroupClause *subq_gclause =
+								get_sortgroupref_clause(subq_tle->ressortgroupref,
+														subq->groupClause);
+							if (subq_gclause->sortop != outer_sc->sortop)
```
cevian: need to check null_first in here too.

src/planner.c
```diff
@@ -737,6 +754,126 @@ contain_param(Node *node)
...
+							{
+								subq_gclause->sortop = outer_sc->sortop;
```
cevian: any reason to avoid making these changes unless we are actually doing the reorder?

src/planner.c
cevian: actually do we need this if at all?

gayyappan: is that possible: sortop is the same but nulls_first different?

> Some minorish changes... mostly looks good. Outstanding question about using the walker

checked the query_tree_walker -> this calls range_walker which steps inside the individual rtes and passes the appropriate node to the reorder_groupby function. so we lose context for view checking.

src/planner.c
gayyappan: changed the code to work on a copy and modify only if we are replacing.

src/planner.c
```diff
@@ -117,33 +135,59 @@ turn_off_inheritance_walker(Node *node, Cache *hc)
...
+		}
+		if (node_arg)
+		{
+			Cache *hc = ts_hypertable_cache_pin();
+			cxt.hc = hc;
+			/*
+			 * turn of inheritance on hypertables we will expand ourselves in
+			 * timescaledb_get_relation_info_hook
+			 */
+			timescaledb_query_walker((Node *) parse, &cxt);
```
cevian: should this be node_arg or parse?

src/planner.c
```diff
@@ -737,6 +781,124 @@ contain_param(Node *node)
...
+			if (IsA(outer_tle->expr, Var) && (((Var *) outer_tle->expr)->varno == rtno))
+			{
+				int outer_attno = ((Var *) outer_tle->expr)->varattno;
+				TargetEntry *subq_tle = list_nth(subq->targetList, outer_attno - 1);
+				if (subq_tle->ressortgroupref > 0)
+				{
+					// get group clause corresponding to this.
+					SortGroupClause *subq_gclause =
+						get_sortgroupref_clause(subq_tle->ressortgroupref, subq_groupclause_copy);
+					if (subq_gclause->sortop != outer_sc->sortop)
```
cevian: again, I see no reason why the sortops and nulls_first are aligned, both need to be fixed up. You can do ORDER BY X DESC NULLS FIRST or ORDER BY X DESC NULLS LAST

src/planner.c
gayyappan: changed to node_arg.

src/planner.c
gayyappan: modified and added test case for this.

PR #1576, opened 2019-12-06
Refactor chunk handling to separate out stub

cevian:
Previously, the Chunk struct was used to represent both a full
chunk and the stub used for joins. The stub used for joins
only contained valid values for some chunk fields and not others.
After the join determined that a Chunk was complete, it filled
in the rest of the chunk field. The fact that a chunk could have
only some fields filled out and not others at different times,
made the code hard to follow and error prone.

So we separate out the stub state of the chunk into a separate
struct that doesn't contain the not-filled-out fields inside
of it.  This leverages the type system to prevent errors that
try to access invalid fields during the join phase and makes
the code easier to follow.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -71,15 +71,16 @@ typedef enum ChunkResult
 	CHUNK_PROCESSED
 } ChunkResult;
 
-typedef ChunkResult (*on_chunk_func)(ChunkScanCtx *ctx, Chunk *chunk);
+typedef ChunkResult (*on_chunk_stub_func)(ChunkScanCtx *ctx, ChunkStub *stub);
 static void chunk_scan_ctx_init(ChunkScanCtx *ctx, Hyperspace *hs, Point *p);
 static void chunk_scan_ctx_destroy(ChunkScanCtx *ctx);
 static void chunk_collision_scan(ChunkScanCtx *scanctx, Hypercube *cube);
-static int chunk_scan_ctx_foreach_chunk(ChunkScanCtx *ctx, on_chunk_func on_chunk, uint16 limit);
-static Chunk **chunk_get_chunks_in_time_range(Oid table_relid, Datum older_than_datum,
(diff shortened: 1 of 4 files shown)


src/chunk.c
```diff
@@ -1016,7 +1034,7 @@ chunk_scan_ctx_get_chunk(ChunkScanCtx *ctx)
  * 2) For each dimension slice:
  *	  - Find all chunk constraints matching the dimension slice
  * 3) For each matching chunk constraint
- *	  - Insert a (stub) chunk in a hash table and add the constraint to the chunk
+ *	  - Insert a (chunk) chunk in a hash table and add the constraint to the chunk
```
JLockerman: huh?

src/chunk.c
```diff
@@ -1210,15 +1230,16 @@ append_chunk_oid(ChunkScanCtx *scanctx, Chunk *chunk)
 }
 
 static ChunkResult
-append_chunk(ChunkScanCtx *scanctx, Chunk *chunk)
+append_chunk(ChunkScanCtx *scanctx, ChunkStub *stub)
```
JLockerman:
```suggestion
append_chunk_stub(ChunkScanCtx *scanctx, ChunkStub *stub)
```

src/chunk.h
```diff
@@ -46,6 +46,13 @@ typedef struct Chunk
 	ChunkConstraints *constraints;
 } Chunk;
 
+typedef struct ChunkStub
```
JLockerman: I think this needs a comment explaining what it is conceptually. To my eye it seems like it identifies a subspace of a hyperspace which _could_ be covered by a single chunk.

src/chunk.c
cevian: good catch stupid find/replace

src/chunk.c
cevian: No, this function appends the chunk, not the stub to the data field

PR #1589, opened 2019-12-16
Allow drop_chunks on raw data while keeping continuous aggregate data

cevian:
Detailed description in the commit messages.

### DIFF
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -213,7 +213,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_drop_chunks (
     hypertable_id   		    INTEGER     UNIQUE      NOT NULL REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
     older_than	    _timescaledb_catalog.ts_interval    NOT NULL,
 	cascade					    BOOLEAN                 NOT NULL,
-    cascade_to_materializations BOOLEAN                 NOT NULL,
+    cascade_to_materializations BOOLEAN,
     CONSTRAINT valid_older_than CHECK(_timescaledb_internal.valid_ts_interval(older_than))
 );
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_policy_drop_chunks', '');
@@ -270,7 +270,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_completed_thresh
     materialization_id INTEGER PRIMARY KEY
         REFERENCES _timescaledb_catalog.continuous_agg(mat_hypertable_id)
         ON DELETE CASCADE,
-    watermark BIGINT NOT NULL
+    watermark BIGINT NOT NULL --exclusive (everything up to but not including watermark is done)
 );
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_completed_threshold', '');
 
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -124,6 +124,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.chunk (
     schema_name     NAME    NOT NULL,
     table_name      NAME    NOT NULL,
     compressed_chunk_id   INTEGER  REFERENCES _timescaledb_catalog.chunk(id),
+    dropped         BOOLEAN NOT NULL DEFAULT false,
     UNIQUE (schema_name, table_name)
 );
 CREATE INDEX IF NOT EXISTS chunk_hypertable_id_idx
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -78,3 +78,8 @@ DROP TABLE _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
 CREATE INDEX continuous_aggs_materialization_invalidation_log_idx
     ON _timescaledb_catalog.continuous_aggs_materialization_invalidation_log (materialization_id, lowest_modified_value ASC);
 GRANT SELECT ON  _timescaledb_catalog.continuous_aggs_materialization_invalidation_log TO PUBLIC;
+
+ALTER TABLE _timescaledb_config.bgw_policy_drop_chunks ALTER COLUMN cascade_to_materializations DROP NOT NULL;
+
+UPDATE _timescaledb_config.bgw_policy_drop_chunks SET cascade_to_materializations = NULL WHERE cascade_to_materializations = false;
+
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -83,3 +83,9 @@ ALTER TABLE _timescaledb_config.bgw_policy_drop_chunks ALTER COLUMN cascade_to_m
 
 UPDATE _timescaledb_config.bgw_policy_drop_chunks SET cascade_to_materializations = NULL WHERE cascade_to_materializations = false;
 
+ALTER TABLE  _timescaledb_catalog.chunk ADD COLUMN dropped BOOLEAN DEFAULT false;
+UPDATE _timescaledb_catalog.chunk SET dropped = false;
+ALTER TABLE _timescaledb_catalog.chunk ALTER COLUMN dropped SET NOT NULL;
+
+CLUSTER  _timescaledb_catalog.chunk USING chunk_pkey;
+ALTER TABLE _timescaledb_catalog.chunk SET WITHOUT CLUSTER;
--- a/src/bgw_policy/drop_chunks.h
+++ b/src/bgw_policy/drop_chunks.h
@@ -9,10 +9,15 @@
 
 #include "catalog.h"
 #include "export.h"
+#include "chunk.h"
 
 typedef struct BgwPolicyDropChunks
 {
-	FormData_bgw_policy_drop_chunks fd;
+	int32 job_id;
+	int32 hypertable_id;
+	FormData_ts_interval older_than;
+	bool cascade;
+	CascadeToMaterializationOption cascade_to_materializations;
 } BgwPolicyDropChunks;
 
 extern TSDLLEXPORT BgwPolicyDropChunks *ts_bgw_policy_drop_chunks_find_by_job(int32 job_id);
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -80,7 +80,8 @@ static int chunk_scan_ctx_foreach_chunk_stub(ChunkScanCtx *ctx, on_chunk_stub_fu
 static Chunk *chunk_get_chunks_in_time_range(Oid table_relid, Datum older_than_datum,
 											 Datum newer_than_datum, Oid older_than_type,
 											 Oid newer_than_type, char *caller_name,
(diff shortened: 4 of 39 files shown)


cevian:
@gayyappan When we move the completion_threshold up (materialize forward in time) we have to use multiple transaction so as to not block inserts for too long because we have to move the invalidation_threshold up. That's why we use multiple transactions to begin with. Thankfully here we only need to process invalidations, not move forward in time and so we can use transactional semantics. If you look at the code we only allow using transactional semantics in this case.

src/bgw_policy/drop_chunks.h
```diff
@@ -9,10 +9,15 @@
...
+#include "chunk.h"
 
 typedef struct BgwPolicyDropChunks
 {
-	FormData_bgw_policy_drop_chunks fd;
+	int32 job_id;
+	int32 hypertable_id;
+	FormData_ts_interval older_than;
+	bool cascade;
+	CascadeToMaterializationOption cascade_to_materializations;
```
gayyappan: I don't see why we need a 3-state cascade_to_materialization. The default should be false (like before).  And the user can over-ride it to true. Don't see why it is useful to have a null/unknown state.

src/chunk.c
```diff
@@ -716,10 +742,35 @@ ts_chunk_create(Hypertable *ht, Point *p, const char *schema, const char *prefix
...
 
 	/* Recheck if someone else created the chunk before we got the table lock */
-	chunk = ts_chunk_find(ht->space, p);
+	chunk = ts_chunk_find(ht->space, p, true);
+
+	if (chunk != NULL && chunk->fd.dropped)
+	{
+		/* mark as not dropped and recreate the pg objects */
+		ScanIterator iterator =
+			ts_scan_iterator_create(CHUNK, RowExclusiveLock, CurrentMemoryContext);
```
gayyappan: when is the lock on the row released? we should be holding it till the pg object creation is complete.

gayyappan: should we add an assert that the chunk is not a compressed chunk? Or should it be applicable to compressed chunks as well?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -620,12 +649,33 @@ materialization_invalidation_log_get_range(int32 materialization_id, Oid type, i
...
+		/* if invalidate_prior_to_time is set that means invalidations should process buckets cover
+		 * all points up to but excluding invalidate_prior_to_time. The last bucket should cover
+		 * those points and is allowed to go over. */
+		if (invalidate_prior_to_time != PG_INT64_MAX)
+		{
+			int64 prior_to_time_bucketed =
+				ts_time_bucket_by_type(bucket_width, invalidate_prior_to_time, type);
+			/* the last bucket should cover prior_to_time */
+			if (prior_to_time_bucketed != invalidate_prior_to_time)
+				prior_to_time_bucketed = int64_saturating_add(prior_to_time_bucketed, bucket_width);
```
JLockerman: do we validate the `invalidate_prior_to_time < completed_threshold`?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -620,12 +649,33 @@ materialization_invalidation_log_get_range(int32 materialization_id, Oid type, i
 		if (entry_range.start >= completed_threshold)
```
JLockerman: should we also have `if (entry_range.start >= invalidate_prior_to_time) break` ?

src/chunk.c
gayyappan: we should test this scenario:  1) drop chunk on table 2) change hypertable's chunk interval 3) create some new chunks 4) now try to recreate the old chunk and insert/update/delete/select from the old chunk that was recreated with the old settings.

src/chunk.c
```diff
@@ -1720,36 +1805,50 @@ ts_chunk_exists_relid(Oid relid)
...
-	ts_chunk_index_delete_by_chunk_id(form.id, true);
-	ts_compression_chunk_size_delete(form.id);
 
-	/* Check for dimension slices that are orphaned by the chunk deletion */
-	for (i = 0; i < ccs->num_constraints; i++)
+	/* if only marking as deleted, keep the constraints and dimension info */
+	if (!preserve_chunk_catalog_row)
 	{
-		ChunkConstraint *cc = &ccs->constraints[i];
+		ts_chunk_constraint_delete_by_chunk_id(form.id, ccs);
```
gayyappan: are we skipping over these chunks during query planning? Does it happen automatically because we use the postgres tables and not our catalogs for expansion?

src/chunk.c
```diff
@@ -1720,36 +1805,50 @@ ts_chunk_exists_relid(Oid relid)
 	return ts_chunk_get_by_relid(relid, 0, false) != NULL;
 }
 
-static ScanTupleResult
-chunk_tuple_delete(TupleInfo *ti, void *data)
+/* Delete the chunk tuple.
+ *
+ * mark_as_dropped_only - instead of deleting the row, mark it as dropped.
```
gayyappan: mark_as_dropped_only --> is this the old parmeter name for preserve_catalog_row

src/chunk.c
```diff
@@ -2151,23 +2266,143 @@ ts_chunk_drop(Chunk *chunk, DropBehavior behavior, int32 log_level)
 			 chunk->fd.table_name.data);
 
 	/* Remove the chunk from the hypertable table */
-	ts_chunk_delete_by_relid(chunk->table_id, behavior);
+	ts_chunk_delete_by_relid(chunk->table_id, behavior, preserve_catalog_row);
```
gayyappan: need a concurrent test where 1) 1 txn drops a chunk 2) the other txn is trying to recreate the chunk. this is to check that we are holding all locks till the first txn completes.

src/chunk.c
```diff
@@ -2151,23 +2266,143 @@ ts_chunk_drop(Chunk *chunk, DropBehavior behavior, int32 log_level)
...
+						cagg.user_view_schema.data,
+						cagg.user_view_name.data)));
+
+	/* error for now, maybe better as a warning and ignoring the chunks? */
+	/* We cannot move a completion threshold up transactionally without taking locks
+	 * that would block the system. So, just bail. The completion threshold
+	 * should be much higher than this anyway */
+	lowest_completion_time = ts_continuous_aggs_min_completed_threshold(ht->fd.id, &cagg);
+	if (lowest_completion_time < older_than_time)
+		ereport(ERROR,
```
gayyappan: do we have a test case for this?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -1163,6 +1205,29 @@ range_expand_right(InternalTimeRange *range, const InternalTimeRange to_cover, i
...
+static void
+range_clamp_end(InternalTimeRange *range, int64 end_value, int64 bucket_width)
+{
+	if (end_value == PG_INT64_MAX)
+		return;
+
+	/* the end value has to be bucketed, we don't bucket here since sometimes you want to round down
+	 * or round up and the caller should decide. */
+	Assert(end_value == ts_time_bucket_by_type(bucket_width, end_value, range->type));
+	if (end_value < range->start)
```
JLockerman: I think this should be an Assertion failure, it violates the invariant of the range type

JLockerman: whoops, misread the condition. Considering what this actually does I think the body should be
```
range->start = end;
range->end = end;
```
but not set on it

src/chunk.c
cevian: yes

sql/pre_install/tables.sql
```diff
@@ -213,7 +214,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_config.bgw_policy_drop_chunks (
     hypertable_id   		    INTEGER     UNIQUE      NOT NULL REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
     older_than	    _timescaledb_catalog.ts_interval    NOT NULL,
 	cascade					    BOOLEAN                 NOT NULL,
-    cascade_to_materializations BOOLEAN                 NOT NULL,
+    cascade_to_materializations BOOLEAN,
```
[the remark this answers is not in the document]
cevian: we want NULL to be the default and error if there are continuous aggs

sql/updates/latest-dev.sql
```diff
@@ -78,3 +78,14 @@ DROP TABLE _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
 CREATE INDEX continuous_aggs_materialization_invalidation_log_idx
     ON _timescaledb_catalog.continuous_aggs_materialization_invalidation_log (materialization_id, lowest_modified_value ASC);
 GRANT SELECT ON  _timescaledb_catalog.continuous_aggs_materialization_invalidation_log TO PUBLIC;
+
+ALTER TABLE _timescaledb_config.bgw_policy_drop_chunks ALTER COLUMN cascade_to_materializations DROP NOT NULL;
+
+UPDATE _timescaledb_config.bgw_policy_drop_chunks SET cascade_to_materializations = NULL WHERE cascade_to_materializations = false;
```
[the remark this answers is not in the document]
cevian: unfortunately previous calls to the policy with an argument of NULL were recorded as false, but the user didn't mean to explicitly make a decision. so we change it to NULL to be safe.

src/bgw_policy/drop_chunks.h
cevian: the default  should be NULL

src/chunk.c
cevian: yes. This is skipped in planning and tested

src/chunk.c
cevian: yes in tsl/test/expected/continuous_aggs_ddl-11.out (around line 622)

src/chunk.c
cevian: The lock on line 716 is not released until end of txn. (see comment above 716). This should work on compressed chunks too. I have a test for this but will add the chunk_interval change.

cevian: oh wait no it shouldn't be allowed on compressed chunks. I'll fix.

cevian: Oh the compressed chunk thing is checked during drop.

gayyappan: referring to the lock on line 751. RowExclusiveLock on CHUNK. when does that get released?

cevian: The lock is on line 742 `LockRelationOid(ht->main_table_relid, ShareUpdateExclusiveLock);`

src/chunk.c
cevian: The lock on line 742 `LockRelationOid(ht->main_table_relid, ShareUpdateExclusiveLock);` prevents this and hasn't changed with this PR so I don't think we need more tests here.

src/process_utility.c
```diff
@@ -810,7 +810,7 @@ process_drop_chunk(ProcessUtilityArgs *args, DropStmt *stmt)
 					ts_chunk_get_by_id(chunk->fd.compressed_chunk_id, 0, false);
 				/* The chunk may have been delete by a CASCADE */
 				if (compressed_chunk != NULL)
-					ts_chunk_drop(compressed_chunk, stmt->behavior, DEBUG1);
+					ts_chunk_drop(compressed_chunk, stmt->behavior, DEBUG1, false);
```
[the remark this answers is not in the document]
cevian: great idea, changed!

PR #1592, opened 2019-12-18
Improve continuous agg user messages

cevian:
Switch from using internal timestamps to more user-friendly
timestamps in our log messages and clean up some messages.

### DIFF
--- a/tsl/src/continuous_aggs/materialize.c
+++ b/tsl/src/continuous_aggs/materialize.c
@@ -257,8 +257,9 @@ continuous_agg_materialize(int32 materialization_id, ContinuousAggMatOptions *op
 
 		if (materializing_new_range)
 			appendStringInfo(msg,
-							 "new range up to " INT64_FORMAT,
-							 materialization_invalidation_threshold);
+							 "new range up to %s",
+							 ts_internal_to_time_string(materialization_invalidation_threshold,
+														time_column_type));
 		else
 			appendStringInfo(msg, "no new range");
 		elog(INFO, "%s", msg->data);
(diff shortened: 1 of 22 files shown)


tsl/src/continuous_aggs/materialize.c
```diff
@@ -421,11 +421,11 @@ get_materialization_end_point_for_table(int32 raw_hypertable_id, int32 materiali
 		if (verbose)
 			elog(INFO,
 				 "new materialization range not found for %s.%s (time column %s): "
-				 "not enough new data past completion threshold (" INT64_FORMAT ")",
+				 "not enough new data past completion threshold as of %s",
```
gayyappan: this looks a little strange with integer messages like : not enough new data past completion threshold as of 12. I think the older formatting "not enough new data past completion threshold(12)" looks better.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -396,11 +396,11 @@ get_materialization_end_point_for_table(int32 raw_hypertable_id, int32 materiali
 		if (verbose)
 			elog(INFO,
 				 "new materialization range not found for %s.%s (time column %s): not enough data "
-				 "in table (" INT64_FORMAT ")",
+				 "past minimum time value as of %s",
```
gayyappan: this should be reworded. suggest we use some terms from the docs for the messages. suggestion "past materialization threshold as of %s".

tsl/src/continuous_aggs/materialize.c
cevian: Fixed

tsl/src/continuous_aggs/materialize.c
cevian: This is actually the correct message since it deals with underflow not any of the thresholds

gayyappan: How about rewording it as: " new materialization range not found ....: not enough data that **satisfies refresh lag criterion** as of ..."? ( ***minimum time value*** does not make much sense to end user?)

PR #1594, opened 2019-12-19
Fix error in compression constraint check

### LINKED ISSUE #1590: Drop column on hypertable makes compression not activable
 - PostgreSQL version : 11.5
 - TimescaleDB version : 1.5.1
 - Installation method: docker with TimescaleDB's official 1.5.1-pg11 image

**Describe the bug**
For the full history, the bug has been described on Timescale's slack channel : https://timescaledb.slack.com/archives/C4GT3N90X/p1576592466079500
With the Timescale 1.5.1-pg11 version there is an issue for enabling the compression feature.
I noticed that I couldn't enable compression on my existing hypertable but if I created a new hypertable with the same fields and constraints it works well.

**To Reproduce**
```
CREATE TABLE public.mes (
	tstamp timestamptz NOT NULL,
	vunit_id int8 NOT NULL,
	identifierreading_id int8 NOT NULL,
	value float8 NULL,	
	CONSTRAINT mes_identifierreading_id_fk FOREIGN KEY (identifierreading_id) REFERENCES identifierreadingoverride(identifierreading_id),
	CONSTRAINT mes_vunit_id_fk FOREIGN KEY (vunit_id) REFERENCES vunit(vunit_id)
);
-- transform into hypertable
SELECT create_hypertable('mes', 'tstamp', chunk_time_interval => interval '7 day');
-- drop column no longer used
ALTER TABLE mes	DROP COLUMN vunit_id;
-- Enable compression segmented by identifierreading_id and order by tstamp desc 
ALTER TABLE mes SET (timescaledb.compress,timescaledb.compress_segmentby = 'identifierreading_id');	
-- -> ERROR: constraint mes_identifierreading_id_fk requires column value to be a segment_by column for compression
```
vunit_id and identifierreading_id are FKs to tables describing the devices and type of measures that are stored. As you can see I removed the vunit_id column, and the hypertable mes is now a simple (tstamp, device_id, value) hypertable.
And when I try to activate the compression feature I get the message :

> constraint mes_identifierreading_id_fk requires column value to be a segment_by column for compression 

Note that when I create a new hypertable like this, the compression feature is enabled without any issues :
```
CREATE TABLE public.mes2 (
	tstamp timestamptz NOT NULL,	
	identifierreading_id int8 NOT NULL,
	value float8 NULL,	
	CONSTRAINT mes_identifierreading_id_fk FOREIGN KEY (identifierreading_id) REFERENCES identifierreadingoverride(identifierreading_id)	
);
-- transform into hypertable
SELECT create_hypertable('mes2', 'tstamp', chunk_time_interval => interval '7 day');
-- Enable compression segmented by identifierreading_id and order by tstamp desc 
ALTER TABLE mes2 SET (timescaledb.compress,timescaledb.compress_segmentby = 'identifierreading_id');	
-- -> Everything OK
```

**Expected behavior**
I should be able to activate the compression feature on a hypertable even when I dropped a column that has a fk constraint on this hypertable beforehand.

**Actual behavior**
See error message above.

cevian:
The constraint check previously assumed that the col_meta
offset for a column was equal to that columns attribute
offset. This is incorrect in the presence of dropped columns.

Fixed to match on column names.

Fixes #1590

### DIFF
--- a/tsl/src/compression/create.c
+++ b/tsl/src/compression/create.c
@@ -676,6 +676,20 @@ add_time_to_order_by_if_not_included(List *orderby_cols, List *segmentby_cols, H
 	return orderby_cols;
 }
 
+static FormData_hypertable_compression *
+get_col_info_for_attnum(Hypertable *ht, CompressColInfo *colinfo, AttrNumber attno)
+{
+	char *attr_name = get_attname_compat(ht->main_table_relid, attno, false);
+	int colno;
+
+	for (colno = 0; colno < colinfo->numcols; colno++)
+	{
+		if (namestrcmp(&colinfo->col_meta[colno].attname, attr_name) == 0)
+			return &colinfo->col_meta[colno];
+	}
+	return NULL;
+}
+
 /* returns list of constraints that need to be cloned on the compressed hypertable
  * This is limited to foreign key constraints now
  */
@@ -738,18 +752,23 @@ validate_existing_constraints(Hypertable *ht, CompressColInfo *colinfo)
 			attnums = (int16 *) ARR_DATA_PTR(arr);
 			for (j = 0; j < numkeys; j++)
 			{
-				int16 colno = attnums[j] - 1;
+				FormData_hypertable_compression *col_def =
+					get_col_info_for_attnum(ht, colinfo, attnums[j]);
+
+				if (col_def == NULL)
+					elog(ERROR, "missing column definition for constraint");
+
 				if (form->contype == CONSTRAINT_FOREIGN)
 				{
 					/* is this a segment-by column */
-					if (colinfo->col_meta[colno].segmentby_column_index < 1)
+					if (col_def->segmentby_column_index < 1)
 					{
 						ereport(ERROR,
 								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 								 errmsg("constraint %s requires column %s to be a segment_by "
 										"column for compression",
 										NameStr(form->conname),
-										NameStr(colinfo->col_meta[colno].attname)),
+										NameStr(col_def->attname)),
 								 errhint("Only segment by columns can be used in foreign key"
 										 "constraints on hypertables that are compressed.")));
 					}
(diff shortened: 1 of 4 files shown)


tsl/src/compression/create.c
```diff
@@ -674,6 +674,23 @@ add_time_to_order_by_if_not_included(List *orderby_cols, List *segmentby_cols, H
 	return orderby_cols;
 }
 
+static FormData_hypertable_compression *
+get_col_info_for_attnum(Hypertable *ht, CompressColInfo *colinfo, AttrNumber attno)
+{
+	Relation rel = relation_open(ht->main_table_relid, AccessShareLock);
+	Form_pg_attribute attr = TupleDescAttr(RelationGetDescr(rel), AttrNumberGetAttrOffset(attno));
```
erimatnor: Consider using `get_attname` here instead.

tsl/src/compression/create.c
```diff
@@ -674,6 +674,23 @@ add_time_to_order_by_if_not_included(List *orderby_cols, List *segmentby_cols, H
...
+	int colno;
+	FormData_hypertable_compression *res = NULL;
+
+	for (colno = 0; colno < colinfo->numcols; colno++)
+	{
+		if (namestrcmp(&colinfo->col_meta[colno].attname, NameStr(attr->attname)) == 0)
```
erimatnor: Shouldn't there be a check for `attisdropped` here or at least an Assert?

tsl/src/compression/create.c
```diff
@@ -736,18 +753,23 @@ validate_existing_constraints(Hypertable *ht, CompressColInfo *colinfo)
 			attnums = (int16 *) ARR_DATA_PTR(arr);
 			for (j = 0; j < numkeys; j++)
 			{
-				int16 colno = attnums[j] - 1;
+				FormData_hypertable_compression *col_def =
```
gayyappan: if the column was dropped, why do we still see the constraint on the table? Is the constraint marked as not valid?

cevian: we don't but the index to col_meta will not match the attnum because col_meta does not include dropped columns

tsl/src/compression/create.c
cevian: good point. changed

tsl/src/compression/create.c
cevian: changed to using get_attname here so don't have that info, but double checked and that can't happen.

PR #1601, opened 2019-12-27
Add T-Digest functionality

spolcyn:
This PR builds on PipelineDB's implementation of the T-Digest approximation data structure to bring T-Digest to TimescaleDB. With minimal resource use, the T-Digest data structure enables rapid approximation of arbitrary percentile and CDF values from a dataset.


.gitignore
```diff
@@ -21,3 +21,5 @@ typedef.list
 /.vs
 /compile_commands.json
 /.DS_Store
+.*.*.swp
```
JLockerman: this ignore looks odd, what's it for?

src/tdigest_api.h
```diff
@@ -0,0 +1,60 @@
...
+ * For future extensibility and memory management reasons, we use a struct enclosing a TDigest,
+ * rather than a raw TDigest
+ */
+typedef struct TDigestTState
+{
+	TDigest *td;
+} TDigestTState;
+
+/* Transition support functions */
+extern Datum ts_tdigest_print_api(PG_FUNCTION_ARGS);
```
JLockerman: The naming of these functions is a bit odd, as it looks like we're e.g. printing a T-Digest API. If we want to have API in the I'd reccommend it be as a prefix, such as `ts_api_tdigest_print` or `ts_tdigest_api_print`. An alternative would be to use `sql` as the suffix.

src/tdigest.h
```diff
@@ -0,0 +1,104 @@
...
+	uint32 size;		// max number of centroids that can be stored in the t-digest
+
+	int64 total_weight; // total num of merged samples represented by the t-digest
+	float8 min;			// lowest centroid mean in the t-digest
+	float8 max;			// highest centroid mean in the t-digest
+
+	List *unmerged_centroids; // input buffer for new centroids
+	uint32 num_centroids;	 // total number of currently stored merged centroids (does not include
+							  // input buffer)
+	Centroid centroids[1];	// storage for merged centroids
```
JLockerman:
```suggestion
	Centroid centroids[FLEXIBLE_ARRAY_MEMBER];	// storage for merged centroids
```

src/tdigest.h
```diff
@@ -0,0 +1,104 @@
...
+
+	float8 compression; // increasing compressing lowers memory usage and decreases accuracy
+	uint32 threshold;   // maximum number of centroids to store before automatically compressing
+	uint32 size;		// max number of centroids that can be stored in the t-digest
+
+	int64 total_weight; // total num of merged samples represented by the t-digest
+	float8 min;			// lowest centroid mean in the t-digest
+	float8 max;			// highest centroid mean in the t-digest
+
+	List *unmerged_centroids; // input buffer for new centroids
```
JLockerman: The combination of this `List` along with the `vl_len_` is very suspicious; either it cannot be stored on disk, and thus is an `internal` type and does not need the header, or it can be, in which case there's a weird invariant that `unmerged_centroids` should always be read/written as `NIL`. If it's the latter, then the `List` does not belong in this `struct`; it should either be a seperate struct, as in an "expanded" view, or it should be a array laid out directly after this header struct in memory.

.gitignore
spolcyn: Vim swap files

test/sql/tdigest.sql
```diff
@@ -0,0 +1,139 @@
...
+GROUP BY time_bucket('1w', time);
+
+REFRESH MATERIALIZED VIEW cagg_view;
+
+--- Test querying from cont agg/view/table
+SELECT count(percentile_approx(tdigest, .5)) FROM cagg_view;
+SELECT count(cdf_approx(tdigest, 42)) FROM cagg_view;
+SELECT cdf_approx(tdigest, 42) FROM cagg_view LIMIT 5 ;
+SELECT percentile_approx(tdigest, .5) FROM cagg_view LIMIT 5;
+
```
gayyappan: need tests that verify t-digest is being updated correctly.
for each of these cases: inserts/updates and deletes to "test_table"  .
Then Refresh the cont. agg view and verify the tdigest gets updated by querying cont agg.

sql/tdigest.sql
```diff
@@ -0,0 +1,204 @@
...
+    procedure = _timescaledb_internal.tdigest_equal,
+    commutator = =,
+    negator = !=
+);
+
+CREATE OPERATOR >= (
+    LEFTARG = @extschema@.tdigest,
+    RIGHTARG = @extschema@.tdigest,
+    PROCEDURE = _timescaledb_internal.tdigest_ge,
+    commutator = <=,
```
gayyappan: why are the commutators of the  1) >= operator is <= and 2) for <= operator is >= ?

src/tdigest.c
```diff
@@ -0,0 +1,512 @@
...
+
+	left = b->mean - (a->mean + right);
+	a = b;
+	right = t->max - a->mean;
+
+	if (x < a->mean + right)
+		return (weight_so_far + a->weight * interpolate(x, a->mean - left, a->mean + right)) /
+			   t->total_weight;
+
+	return 1;
```
k-rus:
```suggestion
	return 1.0;
```

test/expected/tdigest.out
```diff
@@ -0,0 +1,259 @@
...
+
+SELECT (SELECT count(*) FROM test_table) = (SELECT tdigest_count((SELECT * FROM tdigest_temp))) AS count_works;
+ count_works 
+-------------
+ t
+(1 row)
+
+-- Make sure we can't pass empty parameter list
+\set ON_ERROR_STOP 0
+SELECT tdigest_create() FROM test_table;
```
k-rus: Can you also test to provide NULL values instead in the arguments?

sql/tdigest.sql
```diff
@@ -0,0 +1,204 @@
...
+    LEFTARG = @extschema@.tdigest,
+    RIGHTARG = @extschema@.tdigest,
+    PROCEDURE = _timescaledb_internal.tdigest_gt,
+    commutator = <,
+    NEGATOR = <= 
+);
+
+--- Implement complete operator class to support use of DISTINCT and ORDER BY
+--- Metric for total order is total number of data points added to a TDigest
+CREATE OPERATOR CLASS @extschema@.tdigest_ops
```
gayyappan: do we have a test for this?

src/tdigest.c
```diff
@@ -0,0 +1,512 @@
...
+
+	/* palloc0 means no manual nil'ing/setting to 0 of values necessary */
+	t = palloc0(sizeof(TDigest) + size * sizeof(Centroid));
+
+	t->compression = 1.0 * compression;
+	t->threshold = estimate_compression_threshold(compression);
+	t->size = size;
+	t->min = INFINITY;
+
+	SET_VARSIZE(t, ts_tdigest_size(t));
```
gayyappan: I looked at the code for tdigest_recv and send. Looks like we are assuming we can have int64 bytes alloced for a t-digest. Instead we need to add some memory checks to the T-digest alloc code (@jlockerman are there any gucs /functions used in PG to limit memory usage by a process that can be used here). This will avoid 32-bit porting and OOM issues.

gayyappan: v1_len can be used to store the alloced bytes and then reused by tdigest_send/recv instead of recomputing the size. I don't see v1_len actually being used. Is it?

src/tdigest.c
```diff
@@ -0,0 +1,512 @@
...
+ */
+TDigest *
+ts_tdigest_compress(TDigest *t, List *unmerged_centroids_list)
+{
+	int num_unmerged = list_length(unmerged_centroids_list);
+	Centroid *unmerged_centroids;
+	uint64_t unmerged_weight = 0;
+	ListCell *lc;
+	int i, j;
+	mergeArgs *args;
```
gayyappan: can we rename mergeArgs as tdigest_mergeArgs?

src/tdigest.h
```diff
@@ -0,0 +1,102 @@
...
+} Centroid;
+
+/*
+ * A t-digest structure holds all the centroids that represent a given
+ * dataset, as well as a few extra values for use in maintaining the data
+ * structure effciently.
+ */
+typedef struct TDigest
+{
+	uint32 vl_len_; // for postgres base type definition
```
gayyappan: not clear what this comment is trying to say? Could you elaborate?

src/tdigest_api.c
```diff
@@ -0,0 +1,693 @@
...
+		PG_RETURN_NULL();
+
+	old = MemoryContextSwitchTo(context);
+
+	if (PG_ARGISNULL(0)) // compress unmerged and reallocate in correct memory context
+	{
+		incoming_tts = (TDigestTState *) PG_GETARG_VARLENA_P(1);
+		ts_tdigest_trans_compress(incoming_tts);
+
+		receiver_tts = ts_tdigest_startup(0);
```
gayyappan: ts_tdigest_startup allocs and inits td. We are overwriting that here. Saw this in a few other places too. Need a version of startup like ts_tdigest_alloc_and_set( td) that takes in the passed in td and uses it.

sql/tdigest.sql
gayyappan: is it possible to create a table with "tdigest" type as column and an index on it. That will probably exercise this code.

src/tdigest_api.c
```diff
@@ -0,0 +1,693 @@
...
+}
+
+/*
+ * Merges two TDigest aggregates.
+ * Args: TDigestTState, TDigestTState
+ */
+TS_FUNCTION_INFO_V1(ts_tdigest_combine_sql);
+Datum
+ts_tdigest_combine_sql(PG_FUNCTION_ARGS)
+{
```
gayyappan: should check if code coverage hits this function.

src/tdigest_api.c
```diff
@@ -0,0 +1,693 @@
...
+ */
+static bytea *
+tdigest_send(StringInfo buf, TDigest *t)
+{
+	int i;
+
+	pq_begintypsend(buf);
+
+	/* send the tdigest size */
+	pq_sendint64(buf, ts_tdigest_size(t));
```
gayyappan: see my comment about size of tdigest (tdigest.c). once we resolve that, this should either read directly from VARSIZE(t) or use pq_sendint32.

src/tdigest.c
```diff
@@ -0,0 +1,512 @@
...
+{
+	int i;
+	float8 left, right, index;
+	uint64 weight_so_far;
+	Centroid *a, *b, tmp;
+
+	if (q < 0 || q > 1)
+		elog(ERROR, "percentile should be in [0, 1.0], got %f", q);
+
+	if (t->num_centroids == 0)
```
gayyappan: do we have a test for NaN?

src/tdigest_api.c
```diff
@@ -0,0 +1,693 @@
...
+	// configure fields
+	tts->td = t;
+	tts->unmerged_centroids = NIL;
+
+	return tts;
+}
+
+/*
+ * Compresses buffered data in a TDigest aggregate into the TDigest
+ * This is the only API function that should directly call TDigest's compress function
```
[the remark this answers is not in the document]
spolcyn: Doing so would require bringing one of the `tdigest` methods into the API file; rationale in keeping them separate is to keep the SQL-facing functions and the actual T-Digest logic decoupled.

src/tdigest_api.c
```diff
@@ -0,0 +1,693 @@
...
+	t = tts->td;
+
+	PG_RETURN_TEXT_P(CStringGetTextDatum(ts_tdigest_print(t)));
+}
+
+/*
+ * Sets up a TDigest aggregate
+ */
+static TDigestTState *
+ts_tdigest_startup(uint32_t k)
```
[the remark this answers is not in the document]
k-rus: I understood later that it is the same as in the article. I forgot to remove this comment.

src/tdigest_api.c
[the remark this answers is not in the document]
k-rus: I thought to move this function into T-Digist file, but it is not a good idea, since it will bring T-Digist-State there. So good point and it is better to keep as now.

src/tdigest.c
k-rus: I would prefer to avoid mixing Camel and low case notations and suggest `tdigest_merge_args`. Camel/Pascal notation is de facto reserved for PG functions and types.

sql/tdigest.sql
spolcyn: Done using ORDER BY in LIMIT statements in both T-Digest test and cont_agg_dump test

src/tdigest.c
```diff
@@ -0,0 +1,512 @@
...
+			args->index--;
+
+		t->num_centroids = args->index + 1;
+	}
+
+	/*
+	 * Adjust memory allocation for t-digest if we have more centroids than we started with
+	 * Necessary to ensure integrity of variable-size array hack used for centroid array in t-digest
+	 */
+	if (t->num_centroids > num_centroids)
```
gayyappan: Why are we repalloc-ing here? the alloc is for "size" * Centroid when we create the struct. (Not clear to me what the comment is trying to say either).

gayyappan: should rename "size" as max_centroids or something more descriptive. (I mistook size for the size of the alloced struct which it is not).

src/tdigest.c
gayyappan: Yes, TDigestMergeArgs  sounds good.

src/tdigest_api.c
```diff
@@ -0,0 +1,693 @@
...
+ * Avoids user needing to call count() on a dataset they've already put into a TDigest
+ */
+TS_FUNCTION_INFO_V1(ts_tdigest_count_sql);
+Datum
+ts_tdigest_count_sql(PG_FUNCTION_ARGS)
+{
+	TDigest *t;
+
+	if (PG_ARGISNULL(0))
+		PG_RETURN_INT32(0);
```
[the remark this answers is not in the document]
spolcyn: This function really should be strict (asking how many data points a NULL T-Digest has process is somewhat meaningless); NULL check and return 0 removed in favor of keeping STRICT.

src/tdigest_api.c
spolcyn: Called extensively during tests with the continuous agg when merging T-Digests

PR #1606, opened 2020-01-02
Fix constify params during runtime exclusion

### LINKED ISSUE #1598: Segfault on hyper table select query with NOT IN and date filter
Postgres will segfault when a select query with a filter on the date column and a NOT IN clause.  This will crash in timescale version 1.4.0+ in all Postgres versions (9.6, 10, and 11).  Using our replication steps we have found that the crash does not exist in timescale version 1.3.1.

Below are the replication steps:

1. Run timescale using the following command `docker run timescale/timescaledb:<timescale version>-pg<postgres version>`
2. Setup a timescale table in a psql terminal
```
CREATE TABLE public.test_table (
    uid text NOT NULL,
    dt date NOT NULL
);

select create_hypertable('public.test_table', 'dt', chunk_time_interval => interval '1 day');

COPY public.test_table (uid, dt) FROM stdin;
hi  2019-12-24
\.
```
3. Run the select query to trigger the segfault
```
WITH run_date (rd) AS (
  VALUES ('2019-12-24' at time zone 'UTC')
)
SELECT * FROM public.test_table
WHERE dt = (SELECT rd FROM run_date)
AND uid NOT IN
  (SELECT uid FROM public.test_table);
```

cevian:
Descending into subplans during constification of params
seems unsafe and has led to bugs. Turning this off seems
to be safe and not regress any tested optimizations.

In the future we may want to find a way to optimize this
case as well.

Fixes #1598.

### DIFF
--- a/src/chunk_append/exec.c
+++ b/src/chunk_append/exec.c
@@ -723,6 +723,11 @@ constify_param_mutator(Node *node, void *context)
 	if (node == NULL)
 		return NULL;
 
+	/* Don't descend into subplans to constify their parameters, because they may not be valid yet
+	 */
+	if (IsA(node, SubPlan))
+		return node;
+
 	if (IsA(node, Param))
 	{
 		Param *param = castNode(Param, node);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,13 @@ that a backfill can trigger. By default, all invalidations are processed.
 **Bugfixes**
 * #1591 Fix locf treat_null_as_missing option
 * #1594 Fix error in compression constraint check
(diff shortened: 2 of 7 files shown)


src/chunk_append/exec.c
```diff
@@ -723,6 +723,10 @@ constify_param_mutator(Node *node, void *context)
 	if (node == NULL)
 		return NULL;
 
+	/* Don't descend into subplans to constify their parameters */
```
JLockerman:
```suggestion
	/* Don't descend into subplans to constify their parameters, because they may nit be valid yet */
```

and any info on why would also be helpful

PR #1632, opened 2020-01-15
Check for NULL before dereferencing variable

svenklemm:
Move the NULL-check in continuous_agg_execute_materialization before
the first dereferncing of the variable.


k-rus:
Is there a test, which covers the error condition (I don't find any)? Can such test be provided?

PR #1647, opened 2020-01-23
Increase wait time for bgw tests

svenklemm:
TEST_SPINWAIT_ITERS is used as maximum number of iterations to wait
for background worker results. Each iteration has a 0.1 second sleep
so a setting of 10 makes the test wait 1 second for the desired result.
This PR changes the iterations to 100 to wait at most 10 seconds
which should help with some flaky bgw tests.


test/runner.sh
```diff
@@ -29,7 +29,7 @@ EXT_VERSION=${VERSION##version = }
 #docker doesn't set user
 USER=${USER:-`whoami`}
 
-TEST_SPINWAIT_ITERS=${TEST_SPINWAIT_ITERS:-10}
+TEST_SPINWAIT_ITERS=${TEST_SPINWAIT_ITERS:-100}
```
erimatnor: Does it always wait for this amount of time, or does it complete sooner if the test succeeds?

svenklemm: for successful tests this should not increase the test duration since it exits early if it succeeds this is just the maximum number of checks it does

PR #1648, opened 2020-01-23
Drop chunks for materialized hypertable

### LINKED ISSUE #1644: hypertables wşith continuous aggregates do not drop_chunks() properly
Timescaledb version 1.6

We've been evaluating timescale db recently - the continuous aggs are of particular interest to us. We've come across a big problem:

If we have a hypertable with raw data and create a continuous aggregate on it, when we run drop_chunks() on the raw hypertable data it doesn't cascade properly to the view.

we see that timescale is running a DELETE FROM on any continuous aggregates linked to the raw table. This degrades performance to the point that the database can lock up completely.

**To Reproduce**
create a simple hypertable with columsn such as:

time,
symbol,
measure

create a continuous aggregate on the hypertable to aggregate to 1 minute intervals and generate first,high,low,last

load some random data to the hypertable  - we tested about 12M rows spanning 12 hours

let the cagg get up to date

try to drop_chunks() from the hypertable - it may fail, it may take a long time and suceed, you may get bored and cancel it. (When cancelling we see a message showing that timescaledb is attempting to do a DELETE FROM to the materialized view hypertable)

if you drop the cagg and then attempt drop_chunks() then the data is dropped really fast.

**Expected behavior**
Would have thought the cascaded drops to the materialized view would be handled with a drop_chunks() command and not a DELETE FROM ....  

This bug makes continuous aggregates unusable for us right now.

In 1.6 you introduced the option to persist caggs after dropping raw data. If there was a way to drop data from the cagg independently of the raw data then that would provide a viable workaround but unfortunately that doesnt exist right now (we logged an issue for that also)

Thanks, Rich.

gayyappan:
When drop_chunks is called with cascade_to_materialization = true,
 the materialized data is deleted from the materialization
hypertable, but the chunks are not dropped. This fix drops chunks
if possible and deletes the data only if the materialized chunk
cannot be dropped (which is the case if the materialzied chunk
contains data from multiple raw chunks and some of the raw chunks
are not dropped).

Fixes #1644

### DIFF
--- a/tsl/src/continuous_aggs/drop.c
+++ b/tsl/src/continuous_aggs/drop.c
@@ -15,9 +15,38 @@
 
 #include "create.h"
 
+/* drop chunks from the materialization hypertable that fall within the time
+ * range.
+ */
+static void
+cagg_drop_mat_chunks(Oid mattable_relid, Datum older_than_datum, Datum newer_than_datum,
+					 Oid older_than_type, Oid newer_than_type, bool cascade, int32 log_level)
+{
+	uint64 i = 0, num_chunks = 0;
+	Chunk *matchunks;
+
+	matchunks = ts_chunk_get_chunks_in_time_range(mattable_relid,
+												  older_than_datum,
+												  newer_than_datum,
+												  older_than_type,
+												  newer_than_type,
+												  "drop chunks for materialized hypertable",
+												  CurrentMemoryContext,
+												  &num_chunks,
+												  false);
+	for (; i < num_chunks; i++)
+	{
+		ts_chunk_drop(&matchunks[i], cascade, log_level);
+	}
+}
+
 void
 ts_continuous_agg_drop_chunks_by_chunk_id(int32 raw_hypertable_id, Chunk **chunks_ptr,
-										  Size num_chunks)
+										  Size num_chunks,
+
+										  Datum older_than_datum, Datum newer_than_datum,
+										  Oid older_than_type, Oid newer_than_type, bool cascade,
+										  int32 log_level)
 {
 	ListCell *lc;
 	Oid arg_type = INT4OID;
@@ -37,7 +66,18 @@ ts_continuous_agg_drop_chunks_by_chunk_id(int32 raw_hypertable_id, Chunk **chunk
 		SPIPlanPtr delete_plan;
 		ContinuousAgg *agg = lfirst(lc);
 		Hypertable *mat_table = ts_hypertable_get_by_id(agg->data.mat_hypertable_id);
-
+		cagg_drop_mat_chunks(mat_table->main_table_relid,
+							 older_than_datum,
+							 newer_than_datum,
+							 older_than_type,
+							 newer_than_type,
+							 cascade,
+							 log_level);
+		/* we might still have materialization chunks that have data that refer
+		 * to the dropped chunks from the hypertable. This is because the
+		 * chunk interval on the mat. hypertable is NOT the same as the
+		 * chunk interval on the raw hypertable.
+		 */
 		resetStringInfo(command);
 
 		appendStringInfo(command,
--- a/tsl/test/expected/continuous_aggs_ddl-10.out
+++ b/tsl/test/expected/continuous_aggs_ddl-10.out
@@ -272,6 +272,7 @@ SELECT drop_chunks(
 INFO:  dropping chunk _timescaledb_internal._hyper_5_1_chunk
 INFO:  dropping chunk _timescaledb_internal._hyper_5_2_chunk
 INFO:  dropping chunk _timescaledb_internal._hyper_5_3_chunk
+INFO:  dropping chunk _timescaledb_internal._hyper_6_4_chunk
 ERROR:  cannot drop_chunks on a continuous aggregate materialization table
 \set ON_ERROR_STOP 1
 SELECT count(c) FROM show_chunks('drop_chunks_table') AS c;
@@ -576,14 +577,9 @@ DROP TABLE IF EXISTS drop_chunks_table_u CASCADE;
 NOTICE:  drop cascades to 2 other objects
 NOTICE:  drop cascades to table _timescaledb_internal._hyper_8_9_chunk
 CREATE TABLE drop_chunks_table(time BIGINT, data INTEGER);
-SELECT hypertable_id AS drop_chunks_table_u_id
-    FROM create_hypertable('drop_chunks_table', 'time', chunk_time_interval => 10);
+SELECT hypertable_id AS drop_chunks_table_nid
+    FROM create_hypertable('drop_chunks_table', 'time', chunk_time_interval => 10) \gset
 NOTICE:  adding not-null constraint to column "time"
- drop_chunks_table_u_id 
-------------------------
-                     12
-(1 row)
-
 CREATE OR REPLACE FUNCTION integer_now_test2() returns bigint LANGUAGE SQL STABLE as $$ SELECT coalesce(max(time), bigint '0') FROM drop_chunks_table $$;
 SELECT set_integer_now_func('drop_chunks_table', 'integer_now_test2');
  set_integer_now_func 
@@ -598,11 +594,11 @@ AS SELECT time_bucket('5', time), max(data)
 INSERT INTO drop_chunks_table SELECT i, i FROM generate_series(0, 20) AS i;
 \set ON_ERROR_STOP 0
 SELECT drop_chunks(table_name => 'drop_chunks_table', older_than => 13, cascade_to_materializations => false);
-ERROR:  older_than must be greater than the ignore_invalidation_older_than parameter of public.drop_chunks_view
+ERROR:  older_than must be greater than the timescaledb.ignore_invalidation_older_than parameter of public.drop_chunks_view
 ALTER VIEW drop_chunks_view SET (timescaledb.ignore_invalidation_older_than = 9);
 -- 9 is too small (less than timescaledb.ignore_invalidation_older_than)
 SELECT drop_chunks(table_name => 'drop_chunks_table', older_than => (integer_now_test2()-8), cascade_to_materializations => false);
-ERROR:  older_than must be greater than the ignore_invalidation_older_than parameter of public.drop_chunks_view
+ERROR:  older_than must be greater than the timescaledb.ignore_invalidation_older_than parameter of public.drop_chunks_view
 -- 10 works but we don't have the completion threshold far enough along
 SELECT drop_chunks(table_name => 'drop_chunks_table', older_than => (integer_now_test2()-9), cascade_to_materializations => false);
 ERROR:  the continuous aggregate public.drop_chunks_view is too far behind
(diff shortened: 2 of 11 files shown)


tsl/src/continuous_aggs/drop.c
```diff
@@ -15,9 +15,38 @@
 
 #include "create.h"
 
+/* drop chunks from the materialization hypertable that fall within the time
+ * range.
+ */
+static void
+cagg_drop_mat_chunks(Oid mattable_relid, Datum older_than_datum, Datum newer_than_datum,
```
cevian: I wonder why we don't simply call `ts_chunk_do_drop_chunks` here. Right now that doesn't support calls on continous agg mat tables, but that should actually also be fixed. And making this call that function would unify all the checks.

tsl/src/continuous_aggs/drop.c
```diff
@@ -15,9 +15,38 @@
...
+												  older_than_type,
+												  newer_than_type,
+												  "cagg_drop_mat_chunks",
+												  CurrentMemoryContext,
+												  &num_chunks,
+												  false);
+	// TODO any reason why this should be anything other than DROP_CASCADE
+	for (; i < num_chunks; i++)
+	{
+		ts_chunk_drop(matchunks + i, DROP_CASCADE, log_level);
```
cevian: DROP CASCADE is the more dangerous option so my question is can this be DROP_RESTRICT? but really can't we pass the same option that was part of the statement that caused this in the first place?

tsl/src/continuous_aggs/drop.c
```diff
@@ -15,9 +15,38 @@
...
+{
+	uint64 i = 0, num_chunks = 0;
+	Chunk *matchunks;
+
+	matchunks = ts_chunk_get_chunks_in_time_range(mattable_relid,
+												  older_than_datum,
+												  newer_than_datum,
+												  older_than_type,
+												  newer_than_type,
+												  "cagg_drop_mat_chunks",
```
cevian: this string can appear in error messages so should be more user friendly.

tsl/src/continuous_aggs/drop.c
erimatnor: I would actually go further and suggest we avoid `CASCADE` even when there are dependent objects that should also be removed. I think it is better to be explicit and first remove the dependent objects in that case---then one knows exactly what is going on and what the result is. It also protects against changes that might build new dependencies where `CASCADE` is not suitable anymore.

Minor nit:

I think  `&matchunks[i]` is more readable.

tsl/test/expected/continuous_aggs_ddl-10.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+ 
+(1 row)
+
+SELECT set_chunk_time_interval(:'drop_chunks_mat_tablen', 20);
+ set_chunk_time_interval 
+-------------------------
+ 
+(1 row)
+
+INSERT INTO drop_chunks_table select generate_series(1,35), 100;
```
erimatnor:
```suggestion
INSERT INTO drop_chunks_table SELECT generate_series(1,35), 100;
```

Nit: suggest consistent use of lower/upper case for improved readability. (Also below)

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+ _timescaledb_internal._hyper_12_15_chunk
+(2 rows)
+
+SELECT count(c) FROM show_chunks('drop_chunks_table') AS c;
+ count 
+-------
+     0
+(1 row)
+
+SELECT count(c) FROM show_chunks(:'drop_chunks_mat_tablen') AS c;
```
erimatnor: Would be useful to see the chunks on the materialization table before calling `drop_chunks`

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
  _timescaledb_internal._hyper_12_19_chunk | {"[50,60)"}
 (2 rows)
 
+--test drop_chunks with cascade_to_materialization set to true (github 1644)
+-- This checks if chunks from mat. hypertable are actually dropped 
+-- and deletes data from chunks that cannot be dropped from that mat. hypertable.
+SELECT format('%s.%s', schema_name, table_name) AS drop_chunks_mat_tablen
+    FROM _timescaledb_catalog.hypertable, _timescaledb_catalog.continuous_agg
```
erimatnor: Suggest using table aliases here to make WHERE clause more readable.

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+     0
+(1 row)
+
+SELECT count(c) FROM show_chunks(:'drop_chunks_mat_tablen') AS c;
+ count 
+-------
+     0
+(1 row)
+
+SELECT set_chunk_time_interval('drop_chunks_table', 10);
```
erimatnor: Add comment on what this is intended to achieve. Otherwise non-obvious

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+ set_chunk_time_interval 
+-------------------------
+ 
+(1 row)
+
+SELECT set_chunk_time_interval(:'drop_chunks_mat_tablen', 20);
```
erimatnor: I assume it is not common to set the chunk_time_interval on the materialization table? Again, would be good to get some hints on what you're trying to achieve here?

I guess it has something to do with trying to setup misaligned chunks between the raw hypertable and mat. hypertable?

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+ set_chunk_time_interval 
+-------------------------
+ 
+(1 row)
+
+INSERT INTO drop_chunks_table select generate_series(1,35), 100;
+update drop_chunks_table set data = 250 where time = 25;
+update drop_chunks_table set data = 290 where time = 29;
+REFRESH materialized view drop_chunks_view;
+INFO:  new materialization range not found for public.drop_chunks_table (time column time): not enough new data past completion threshold of 60 as of 35
```
erimatnor: Honestly, I have no clue what these INFO messages mean. Looks more like DEBUG output to me. Realize it is probably not an issue with this PR, but I am still pointing this out because this must be quite confusing for users. I suggest creating an issue to clean up these messages.

As a user-facing message, I think this should be something more akin to:

`continuous aggregate for "public.drop_chunks_table" was not expanded due to lack of new data`

Perhaps more deeper technical information is suitable for a `errdetail` message.

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+INFO:  materializing continuous aggregate public.drop_chunks_view: processing invalidations, no new range
+REFRESH materialized view drop_chunks_view;
+INFO:  new materialization range not found for public.drop_chunks_table (time column time): not enough new data past completion threshold of 60 as of 35
+INFO:  materializing continuous aggregate public.drop_chunks_view: processing invalidations, no new range
+--now we have chunks in both mat and raw hypertables
+select * from drop_chunks_view order by 1;
```
erimatnor:
```suggestion
SELECT * FROM drop_chunks_view ORDER BY 1;
```

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+
+SELECT drop_chunks(table_name => 'drop_chunks_table', older_than=>integer_now_test2() - 4 , cascade_to_materializations => true);
+               drop_chunks                
+------------------------------------------
+ _timescaledb_internal._hyper_12_13_chunk
+ _timescaledb_internal._hyper_12_14_chunk
+ _timescaledb_internal._hyper_12_20_chunk
+(3 rows)
+
+select * from drop_chunks_view order by 1;
```
erimatnor:
```suggestion
SELECT * FROM drop_chunks_view ORDER BY 1;
```

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+ 
+(1 row)
+
+SELECT set_chunk_time_interval(:'drop_chunks_mat_tablen', 20);
+ set_chunk_time_interval 
+-------------------------
+ 
+(1 row)
+
+INSERT INTO drop_chunks_table select generate_series(1,35), 100;
```
erimatnor: Would be good with some hints on what to expect here. Now I am trying to guess what this is supposed to show.

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+           5 | 100
+          10 | 100
+          15 | 100
+          20 | 100
+          25 | 290
+          30 | 100
+          35 | 100
+(8 rows)
+
+SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
```
erimatnor: Add an ORDER BY on the range here.

tsl/test/expected/continuous_aggs_ddl-11.out
```diff
@@ -943,3 +939,107 @@ SELECT chunk_table, ranges FROM chunk_relation_size('drop_chunks_table');
...
+               chunk_table                |   ranges    
+------------------------------------------+-------------
+ _timescaledb_internal._hyper_12_13_chunk | {"[0,10)"}
+ _timescaledb_internal._hyper_12_14_chunk | {"[10,20)"}
+ _timescaledb_internal._hyper_12_17_chunk | {"[30,40)"}
+ _timescaledb_internal._hyper_12_20_chunk | {"[20,30)"}
+(4 rows)
+
+SELECT chunk_table, ranges FROM chunk_relation_size(:'drop_chunks_mat_tablen');
```
erimatnor: Add ORDER BY here too for consistency.

tsl/src/continuous_aggs/drop.c
cevian: Well there may be user-defined dependencies here so I think if cascade was used in the original statement it's wise to pass it down and use it here. Otherwise, all internal dependencies (I don't think there are any) should be processed explicitly and restrict used here.

tsl/src/continuous_aggs/drop.c
gayyappan: The table_name parameter is optional for drop_chunks. The variant which can be called without a table_name can lead to inadvertent errors if we allow dropping materialization tables.
`SELECT drop_chunks(schema_name => 
    table_name => ,
    newer_than => -20,
    verbose => true);
`
with this variant everything works fine when a)  the user calls it explicitly on a materialization table OR b) calls it explicitly on a raw hypertable (cascade_to_materialization parameter decides what to do with the materialization hypertable).

Now consider this variant,
`
SELECT drop_chunks(
    newer_than => -20,
    verbose => true);
`
Here a user can *inadvertently drop materialization tables* . cascade_to_materialization parameter has no effect here i(if we allow drop chunks directly on the mat hypertable) and so the user is not forced to make a choice. Easy to go wrong with this call.
 
I think this variant of drop_chunks should actually be removed. (can this be done for 1.7 ?) The table_name parameter should be mandatory. This would be a cleaner interface and also maps to our drop_chunks policy.
We should enable drop_chunks on materialization hypertable only using the modified interface (where table_name is mandatory) to avoid unexpected issues with dropping chunks from materialized hypertables.

tsl/src/continuous_aggs/drop.c
```diff
@@ -15,9 +15,38 @@
...
+												  CurrentMemoryContext,
+												  &num_chunks,
+												  false);
+	// TODO any reason why this should be anything other than DROP_CASCADE
```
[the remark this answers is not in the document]
gayyappan: I use this as a note about an open item for reviewers to comment on.

tsl/src/continuous_aggs/drop.c
gayyappan: modified to pass on the cascade behavior.

tsl/test/expected/continuous_aggs_ddl-10.out
gayyappan: fixed

tsl/test/expected/continuous_aggs_ddl-11.out
gayyappan: Added a comment before the drop_chunks call.
The default mat. hypertable chunk size is 10 times the raw hypertable's chunk. I am trying to make sure atleast 1 chunk from the mat. hypertable will get dropped without having to insert too much data. Also want data from the other chunk to get deleted.

tsl/test/expected/continuous_aggs_ddl-11.out
gayyappan: This command is more of a debugging command and used in our test cases as we don't want the bgw jobs to do the refresh.
It has been helpful to have these messages in the pg log files (for bgw job runs) for helping users who are unfamiliar with cont. aggregates or figuring out issues with bgw jobs.

erimatnor: INFO is user-facing (sent to client) so I still find this odd even if we recommend bgw doring the refresh over a manual `REFRESH`.

PR #1658, opened 2020-01-27
Unify error message if hypertable not found

k-rus:
Refactors multiple implementations of finding hypertables in cache
and failing with different error messages if not found. The
implementations are replaced with calling functions, which encapsulate
a single error message. This provides the unified error message and
removes need for copy-paste.

PR notes:
Closes https://github.com/timescale/timescaledb-private/issues/539
I am not sure if I gave good names to functions. Let me know if you have better suggestions.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -1462,12 +1462,7 @@ ts_chunk_get_chunks_in_time_range(Oid table_relid, Datum older_than_datum, Datum
 	}
 	else
 	{
-		ht = ts_hypertable_cache_get_entry(hypertable_cache, table_relid);
-		if (!ht)
-			ereport(ERROR,
-					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-					 errmsg("table \"%s\" does not exist or is not a hypertable",
-							get_rel_name(table_relid))));
+		ht = ts_hypertable_cache_get_entry(hypertable_cache, table_relid, false);
 		hypertables = list_make1(ht);
 	}
 
@@ -2321,11 +2316,7 @@ ts_chunk_drop_process_materialization(Oid hypertable_relid,
 				 errmsg("must use older_than parameter to drop_chunks with "
 						"cascade_to_materializations")));
 
-	hcache = ts_hypertable_cache_pin();
-	ht = ts_hypertable_cache_get_entry(hcache, hypertable_relid);
-	if (!ht)
-		elog(ERROR, "can only call drop_chunks on hypertables");
-
+	ht = ts_hypertable_cache_get_cache_and_entry(hypertable_relid, false, &hcache);
 	time_dimension = hyperspace_get_open_dimension(ht->space, 0);
 	older_than_time = get_internal_time_from_endpoint_specifiers(ht->main_table_relid,
 																 time_dimension,
--- a/src/hypertable_cache.c
+++ b/src/hypertable_cache.c
@@ -9,6 +9,7 @@
 #include <utils/lsyscache.h>
 #include <utils/builtins.h>
 
+#include "errors.h"
 #include "hypertable_cache.h"
 #include "hypertable.h"
 #include "catalog.h"
@@ -129,24 +130,57 @@ ts_hypertable_cache_invalidate_callback(void)
 
 /* Get hypertable cache entry. If the entry is not in the cache, add it. */
 TSDLLEXPORT Hypertable *
-ts_hypertable_cache_get_entry(Cache *cache, Oid relid)
+ts_hypertable_cache_get_entry(Cache *const cache, const Oid relid, const bool missing_ok)
 {
-	if (!OidIsValid(relid))
-		return NULL;
+	Hypertable *ht;
 
-	return ts_hypertable_cache_get_entry_with_table(cache, relid, NULL, NULL);
+	if (!OidIsValid(relid))
+	{
+		if (missing_ok)
+			return NULL;
+		else
+			ereport(ERROR, (errcode(ERRCODE_UNDEFINED_OBJECT), errmsg("invalid Oid")));
+	}
+	else
+		ht = ts_hypertable_cache_get_entry_with_table(cache, relid, NULL, NULL);
+
+	if (!missing_ok && ht == NULL)
+	{
+		const char *const rel_name = get_rel_name(relid);
+		if (rel_name == NULL)
+			ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_TABLE),
+					 errmsg("OID %u does not refer to a table", relid)));
+		else
+			ereport(ERROR,
+					(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+					 errmsg("table \"%s\" is not a hypertable", rel_name)));
+	}
+
+	return ht;
+}
+
+/*
+ * Returns cache into the argument and hypertable as the function result.
+ * If hypertable is not found, fails with an error.
+ */
+Hypertable *
+ts_hypertable_cache_get_cache_and_entry(const Oid relid, const bool missing_ok, Cache **const cache)
+{
+	*cache = ts_hypertable_cache_pin();
+	return ts_hypertable_cache_get_entry(*cache, relid, missing_ok);
 }
 
 Hypertable *
 ts_hypertable_cache_get_entry_rv(Cache *cache, RangeVar *rv)
 {
-	return ts_hypertable_cache_get_entry(cache, RangeVarGetRelid(rv, NoLock, true));
+	return ts_hypertable_cache_get_entry(cache, RangeVarGetRelid(rv, NoLock, true), true);
 }
 
 TSDLLEXPORT Hypertable *
 ts_hypertable_cache_get_entry_by_id(Cache *cache, int32 hypertable_id)
 {
-	return ts_hypertable_cache_get_entry(cache, ts_hypertable_id_to_relid(hypertable_id));
+	return ts_hypertable_cache_get_entry(cache, ts_hypertable_id_to_relid(hypertable_id), true);
 }
 
 Hypertable *
--- a/src/hypertable_cache.h
+++ b/src/hypertable_cache.h
@@ -12,7 +12,10 @@
 #include "cache.h"
 #include "hypertable.h"
 
-extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, Oid relid);
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, Oid relid,
+															 bool missing_ok);
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_cache_and_entry(Oid relid, bool missing_ok,
+																	   Cache **cache);
 extern Hypertable *ts_hypertable_cache_get_entry_rv(Cache *cache, RangeVar *rv);
 extern Hypertable *ts_hypertable_cache_get_entry_with_table(Cache *cache, Oid relid,
 															const char *schema, const char *table);
--- a/src/planner.c
+++ b/src/planner.c
@@ -126,7 +126,7 @@ timescaledb_query_walker(Node *node, TimescaledbWalkerState *cxt)
 			{
 				/* turn off inheritance on hypertables for inserts and selects*/
 				Cache *hc = cxt->hc;
-				Hypertable *ht = ts_hypertable_cache_get_entry(hc, rte->relid);
+				Hypertable *ht = ts_hypertable_cache_get_entry(hc, rte->relid, true);
 
 				if (NULL != ht && ts_plan_expand_hypertable_valid_hypertable(ht, query, rti, rte))
 				{
@@ -372,10 +372,7 @@ is_hypertable_chunk_dml(PlannerInfo *root, RelOptInfo *rel, Index rti, RangeTblE
 		parent_oid = appinfo->parent_reloid;
 		if (parent_oid != InvalidOid && rte->relid != parent_oid)
 		{
-			Cache *hcache = ts_hypertable_cache_pin();
-			Hypertable *parent_ht = ts_hypertable_cache_get_entry(hcache, parent_oid);
-			ts_cache_release(hcache);
-			if (parent_ht)
+			if (ts_is_hypertable(parent_oid))
 				return true;
 		}
 	}
@@ -490,8 +487,6 @@ timescaledb_set_rel_pathlist(PlannerInfo *root, RelOptInfo *rel, Index rti, Rang
 		!(is_append_parent(rel, rte) || is_append_child(rel, rte) || is_htdml))
 		return;
 
-	hcache = ts_hypertable_cache_pin();
-
 	/*
 	 * if this is an append child or DML we use the parent relid to
 	 * check if its a hypertable
@@ -499,7 +494,7 @@ timescaledb_set_rel_pathlist(PlannerInfo *root, RelOptInfo *rel, Index rti, Rang
 	if (is_append_child(rel, rte) || is_htdml)
 		ht_reloid = get_parentoid(root, rti);
 
-	ht = ts_hypertable_cache_get_entry(hcache, ht_reloid);
+	ht = ts_hypertable_cache_get_cache_and_entry(ht_reloid, true, &hcache);
 
 	if (!is_htdml)
 	{
@@ -544,10 +539,8 @@ timescaledb_get_relation_info_hook(PlannerInfo *root, Oid relation_objectid, boo
 	 */
 	if (!rte->inh && is_rte_hypertable(rte))
 	{
-		Cache *hcache = ts_hypertable_cache_pin();
-		Hypertable *ht = ts_hypertable_cache_get_entry(hcache, rte->relid);
-
-		Assert(ht != NULL);
+		Cache *hcache;
+		Hypertable *ht = ts_hypertable_cache_get_cache_and_entry(rte->relid, false, &hcache);
 
 		Assert(rel->fdw_private == NULL);
 		rel->fdw_private = palloc0(sizeof(TimescaleDBPrivate));
@@ -560,8 +553,8 @@ timescaledb_get_relation_info_hook(PlannerInfo *root, Oid relation_objectid, boo
 	if (ts_guc_enable_transparent_decompression && is_append_child(rel, rte))
 	{
 		Oid ht_oid = get_parentoid(root, rel->relid);
-		Cache *hcache = ts_hypertable_cache_pin();
-		Hypertable *ht = ts_hypertable_cache_get_entry(hcache, ht_oid);
+		Cache *hcache;
+		Hypertable *ht = ts_hypertable_cache_get_cache_and_entry(ht_oid, true, &hcache);
 
 		if (ht != NULL && ht_oid != rte->relid && TS_HYPERTABLE_HAS_COMPRESSION(ht))
 		{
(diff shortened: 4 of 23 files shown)


src/chunk.c
```diff
@@ -1467,12 +1467,7 @@ chunk_get_chunks_in_time_range(Oid table_relid, Datum older_than_datum, Datum ne
 	}
 	else
 	{
-		ht = ts_hypertable_cache_get_entry(hypertable_cache, table_relid);
-		if (!ht)
-			ereport(ERROR,
-					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-					 errmsg("table \"%s\" does not exist or is not a hypertable",
-							get_rel_name(table_relid))));
+		ht = ts_hypertable_cache_get_entry_always(table_relid, hypertable_cache);
```
erimatnor: Find `_always` a bit ambiguous. I suggest `_or_fail` or similar.

Otherwise, a common pattern in PostgreSQL and also our own code is to have a boolean parameter called `missing_ok` or similar, which determines the behavior of the function. I guess this could be wrapped by a macro or other function that take only two arguments.

src/hypertable_cache.c
```diff
@@ -137,6 +138,34 @@ ts_hypertable_cache_get_entry(Cache *cache, Oid relid)
...
 }
 
+/*
+ * Returns hypertable by checking the given cache first.
+ * If hypertable is not found, fails with an error.
+ */
+Hypertable *
+ts_hypertable_cache_get_entry_always(const Oid relid, Cache *const cache)
+{
+	Hypertable *const ht= ts_hypertable_cache_get_entry(cache, relid);
```
erimatnor: missing whitespace

src/hypertable_cache.c
```diff
@@ -137,6 +138,34 @@ ts_hypertable_cache_get_entry(Cache *cache, Oid relid)
...
+
+	if (ht == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
```
mkindahl: Strictly speaking, there is a difference between "table does not exist" and "table is wrong type". The two error conditions belong to different error categories and I think we should be careful and make sure to not mix them up.

src/hypertable_cache.c
```diff
@@ -137,6 +138,34 @@ ts_hypertable_cache_get_entry(Cache *cache, Oid relid)
 	return ts_hypertable_cache_get_entry_with_table(cache, relid, NULL, NULL);
 }
 
+/*
+ * Returns hypertable by checking the given cache first.
+ * If hypertable is not found, fails with an error.
+ */
+Hypertable *
+ts_hypertable_cache_get_entry_always(const Oid relid, Cache *const cache)
```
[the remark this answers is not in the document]
erimatnor: Referring to my other comment on this naming and parameter choice.

src/hypertable_cache.c
```diff
@@ -137,6 +138,34 @@ ts_hypertable_cache_get_entry(Cache *cache, Oid relid)
...
+
+	return ht;
+}
+
+/*
+ * Returns cache into the argument and hypertable as the function result.
+ * If hypertable is not found, fails with an error.
+ */
+Hypertable *
+ts_hypertable_get_cache_always(const Oid relid, Cache **const cache)
```
erimatnor: Same here w.r.t. naming.

Might also want to allow passing a NULL `cache` argument here and not get the cache (without failure). The cache is automatically released at the end of the transaction, although it is good practice to release it early, if possible. Still, I think having the option to not manage the cache would be good.

src/hypertable_cache.c
k-rus: I am not sure what you are out for. I believe the error code was created exactly for this case that hypertable is not found for the given OID. Which means that the database object is not a hypertable. I believe this discussion is outside of this PR scope, since it doesn't introduce new error code, but uses existing one..

src/hypertable_cache.c
k-rus: @erimatnor I am not sure how long the cache is needed. Looking to all the code, which uses hypertable, it keeps cache pinned while hypertable is used and releases the cache only at the end of functions.
So I don't see possible use cases where cache will not be needed if hypertable is returned.

k-rus:
> I agree with @erimatnor that it might makes sense to support case where we pass `NULL` as a cache. I believe there are a lot of place where we have this pattern and getting cache only to access a hypertable.

There is no case where the cache is released immediately after obtaining a hypertable. Is it safe to release the cache so early? cc @pmwkaa @erimatnor

k-rus:
@mkindahl @erimatnor @pmwkaa Thank you for the naming suggestions. I improved the names, added `missing_ok` parameter for the basic function and created wrapper functions for the common cases.

src/hypertable_cache.h
```diff
@@ -12,7 +12,10 @@
 #include "cache.h"
 #include "hypertable.h"
 
-extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, Oid relid);
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, Oid relid,
+															 bool missing_ok);
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry_missing_ok(Cache *cache, Oid relid);
```
svenklemm: hmm i dont like those function names and I would prefer to have the missing_ok be a bool argument to the function call

k-rus: > hmm i dont like those function names and I would prefer to have the missing_ok be a bool argument to the function call

It's one of the most common case. I wanted to use the argument first, but then I realised that it is too many cases (renaming is automatic :). So I feel that having the wrapper is better. I am not sure what the function name will be better. Let me know if you have ideas for the name.

src/hypertable_cache.c
erimatnor: I agree with @mkindahl that we might want to distinguish between these cases:

* The relid is invalid (InvalidOid)  (ERRCODE_UNDEFINED_OBJECT)
* The relid is valid but is not a table (ERRCODE_WRONG_OBJECT_TYPE)
* The relid is a valid table relation, but is not a hypertable (TS_HYPERTABLE_DOES_NOT_EXIST), 
* The relid is a valid hypertable (no error)

We might choose to not distinguish between the first two errors, however. But since this is now all in one place, it is easy to be specific.

mkindahl: Some cases returned `ERRCODE_INVALID_PARAMETER_VALUE` and some returned `ERRCODE_TS_HYPERTABLE_NOT_EXIST`. Now they all return `ERRCODE_TS_HYPERTABLE_NOT_EXIST`.

k-rus: @erimatnor Thank you for clarifying. I see the point. Distinguishing between option 1 and 2 is easy as they are separate code paths.

src/hypertable_cache.c
```diff
@@ -129,24 +130,51 @@ ts_hypertable_cache_invalidate_callback(void)
...
 	if (!OidIsValid(relid))
-		return NULL;
+		ht = NULL;
+	else
+		ht = ts_hypertable_cache_get_entry_with_table(cache, relid, NULL, NULL);
+
+	if (!missing_ok && ht == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+				 errmsg("table \"%s\" is not a hypertable", get_rel_name(relid))));
```
erimatnor: Note that if relid is InvalidOid, this error might be strange (e.g., `"table "NULL" is not a hypertable`

src/hypertable_cache.c
k-rus: > Some cases returned `ERRCODE_INVALID_PARAMETER_VALUE` and some returned `ERRCODE_TS_HYPERTABLE_NOT_EXIST`. Now they all return `ERRCODE_TS_HYPERTABLE_NOT_EXIST`.

Looking to the actual cases it seems that the choice between the error codes was arbitrary and subjective. However, I agree with Erik's point.

erimatnor: Note that I am not sure we'd like to use `ERRCODE_INVALID_PARAMETER_VALUE` here since we don't have the context in this function to know whether the call came due to a user-invoked function. For instance, it would be a bit weird to raise this error on an insert or non-function operation.

tsl/src/continuous_aggs/create.c
```diff
@@ -355,7 +355,7 @@ cagg_add_trigger_hypertable(Oid relid, char *trigarg)
 	if (check_trigger_exists_hypertable(relid, CAGGINVAL_TRIGGER_NAME))
 		return;
 	hcache = ts_hypertable_cache_pin();
-	ht = ts_hypertable_cache_get_entry(hcache, relid);
+	ht = ts_hypertable_cache_get_entry_missing_ok(hcache, relid);
```
erimatnor: I think this should be `_or_error`

tsl/src/continuous_aggs/create.c
```diff
@@ -476,7 +476,7 @@ mattablecolumninfo_create_materialization_table(MatTableColumnInfo *matcolinfo,
 	cagg_create_hypertable(hypertable_id, mat_relid, matpartcolname, matpartcol_interval);
 	/* retrieve the hypertable id from the cache */
 	hcache = ts_hypertable_cache_pin();
-	ht = ts_hypertable_cache_get_entry(hcache, mat_relid);
+	ht = ts_hypertable_cache_get_entry_missing_ok(hcache, mat_relid);
```
erimatnor: This should definitely be `_or_error` given next line.

k-rus:
Thanks everyone for the feedback. I have suggestion for two improvements:
1. Return different error codes based on the type of error. See https://github.com/timescale/timescaledb/pull/1658#discussion_r371245110
2. Use `missing_ok` always and have only two functions: one takes pinned cache and another takes pointer to cache pointer to fill with the pinned cache.

The suggestion to implement version with releasing the cache is unclear to me, since there is no use case in the code.

@svenklemm @erimatnor @mkindahl @pmwkaa Do you support this?

tsl/test/src/test_ddl_hook.c
```diff
@@ -57,7 +57,7 @@ test_ddl_command_start(ProcessUtilityArgs *args)
 	{
 		Oid relid = lfirst_oid(cell);
 
-		ht = ts_hypertable_cache_get_entry(hcache, relid);
+		ht = ts_hypertable_cache_get_entry_missing_ok(hcache, relid);
```
erimatnor: `_or_error` and remove Assert

tsl/test/src/test_ddl_hook.c
```diff
@@ -91,7 +91,7 @@ test_ddl_command_end(EventTriggerData *command)
...
```
erimatnor: `_or_error` and remove Assert

tsl/test/src/test_continuous_aggs.c
```diff
@@ -46,8 +46,9 @@ ts_run_continuous_agg_materialization(PG_FUNCTION_ARGS)
 	int64 invalidation_threshold;
 	int64 completed_threshold;
 	Cache *hcache = ts_hypertable_cache_pin();
-	int32 hypertable_id = ts_hypertable_cache_get_entry(hcache, hypertable)->fd.id;
-	int32 materialization_id = ts_hypertable_cache_get_entry(hcache, materialization_table)->fd.id;
+	int32 hypertable_id = ts_hypertable_cache_get_entry_missing_ok(hcache, hypertable)->fd.id;
```
erimatnor: `_or_error` given dereference in both this and next line

src/planner.c
```diff
@@ -545,7 +545,7 @@ timescaledb_get_relation_info_hook(PlannerInfo *root, Oid relation_objectid, boo
 	if (!rte->inh && is_rte_hypertable(rte))
 	{
 		Cache *hcache = ts_hypertable_cache_pin();
-		Hypertable *ht = ts_hypertable_cache_get_entry(hcache, rte->relid);
+		Hypertable *ht = ts_hypertable_cache_get_entry_missing_ok(hcache, rte->relid);
```
erimatnor: `_or_error` and remove Assert

src/planner.c
```diff
@@ -373,7 +373,7 @@ is_hypertable_chunk_dml(PlannerInfo *root, RelOptInfo *rel, Index rti, RangeTblE
 		if (parent_oid != InvalidOid && rte->relid != parent_oid)
 		{
 			Cache *hcache = ts_hypertable_cache_pin();
-			Hypertable *parent_ht = ts_hypertable_cache_get_entry(hcache, parent_oid);
+			Hypertable *parent_ht = ts_hypertable_cache_get_entry_missing_ok(hcache, parent_oid);
```
erimatnor: I think this and surrounding lines should be replaced with `ts_is_hypertable()`

src/hypertable_cache.c
```diff
@@ -129,24 +130,51 @@ ts_hypertable_cache_invalidate_callback(void)
...
 
-	return ts_hypertable_cache_get_entry_with_table(cache, relid, NULL, NULL);
+/*
+ * Returns cache into the argument and hypertable as the function result.
+ * If hypertable is not found, fails with an error.
+ */
+Hypertable *
+ts_hypertable_cache_get_entry_or_error(const Oid relid, Cache **const cache)
+{
+	*cache = ts_hypertable_cache_pin();
```
erimatnor: Always forcing cache pinning here makes this function useless across multiple calls using the same cache.

I think you should either 

* skip this cache pinning functionality, or
* check if the cache pointer is already set and not pin again (although this is odd, because it forces always passing in a valid cache in that case), or
* provide a complementary function that uses an existing cache so that this function can be called multiple times for the same cache.

k-rus: > * provide a complementary function that uses an existing cache so that this function can be called multiple times for the same cache.

The function exists as in used. It's called `ts_hypertable_cache_get_entry`.

`ts_hypertable_cache_get_entry_or_error` is only used in the cases where `ts_hypertable_cache_pin` was present. My changes don't change any semantics and if you substitute the function with the body, you will get the original code.

src/planner.c
k-rus: I am not sure if such change is within the scope of the PR.

src/hypertable_cache.c
k-rus: Implemented suggestion from @erimatnor

src/hypertable_cache.h
k-rus: I went for argument `missing_ok` to be always required.

src/hypertable_cache.c
k-rus: Good point. Fixed.

tsl/src/continuous_aggs/create.c
k-rus: Agree. `missing_ok = false`.

src/planner.c
k-rus: Replaced with the suggestion.

k-rus:
> 1. rent error codes based on the type of error. See [#1658 (comment)](https://github.com/timescale/timescaledb/pull/1658#discussion_r371245110)
> 2. Use argument `missing_ok` always and have only two functions: one takes pinned cache and another takes pointer to cache pointer to fill with the pinned cache.

Implemented both

Will squash the second commit before merge or earlier.

src/hypertable_cache.c
```diff
@@ -129,24 +130,54 @@ ts_hypertable_cache_invalidate_callback(void)
...
+ts_hypertable_cache_get_entry(Cache *const cache, const Oid relid, const bool missing_ok)
 {
+	Hypertable *ht;
+
 	if (!OidIsValid(relid))
-		return NULL;
+		if (missing_ok)
+			return NULL;
+		else
+			ereport(ERROR, (errcode(ERRCODE_UNDEFINED_OBJECT), errmsg("invalid Oid")));
```
erimatnor: Nit, just realized there's an `ERRCODE_UNDEFINED_TABLE`, which is probably more suitable here. A common PG error message in these cases is also `relation with OID %u does not exist`

k-rus: I believe this comment is suitable for https://github.com/timescale/timescaledb/pull/1658/files#diff-3086b70871dfb929dfe95db4f60ef7b5R150 and I updated it with:
```
(errcode(ERRCODE_UNDEFINED_TABLE), errmsg("OID %u does not refer to a table", relid))
```

The line, which you target, is about OID value, which is invalid, i.e., `0`. The error code is correct for this case. I haven't found a common error message in PG.

src/hypertable_cache.c
mkindahl: > > Some cases returned `ERRCODE_INVALID_PARAMETER_VALUE` and some returned `ERRCODE_TS_HYPERTABLE_NOT_EXIST`. Now they all return `ERRCODE_TS_HYPERTABLE_NOT_EXIST`.
> 
> Looking to the actual cases it seems that the choice between the error codes was arbitrary and subjective. However, I agree with Erik's point.

Maybe, but the key point is that there are different reasons why the hypertable could not be found, which is what I tried to highlight. @erimatnor did a better job here. :)

src/hypertable_cache.h
```diff
@@ -12,7 +12,10 @@
 #include "cache.h"
 #include "hypertable.h"
 
-extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, Oid relid);
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, Oid relid,
+															 bool missing_ok);
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_cache_and_entry(Oid relid, bool missing_ok,
+																	   Cache **cache);
```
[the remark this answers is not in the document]
k-rus: @mkindahl Your preference is opposite to what was requested by others. `missing_ok` is widely used and it makes less implementations and less naming discussions.

src/hypertable_cache.c
```diff
@@ -129,24 +130,55 @@ ts_hypertable_cache_invalidate_callback(void)
...
+	else
+		ht = ts_hypertable_cache_get_entry_with_table(cache, relid, NULL, NULL);
```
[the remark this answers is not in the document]
k-rus: I don't understand why it is blocking request, since it is a personal taste and is not followed in our code. And I see your code, which doesn't use braces in multiline code.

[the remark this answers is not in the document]
k-rus: OK, the comment is about the then-statement to cover `if(missing_ok)`. I will fix it.

src/planner.c
```diff
@@ -371,13 +371,8 @@ is_hypertable_chunk_dml(PlannerInfo *root, RelOptInfo *rel, Index rti, RangeTblE
...
 		parent_oid = appinfo->parent_reloid;
 		if (parent_oid != InvalidOid && rte->relid != parent_oid)
-		{
-			Cache *hcache = ts_hypertable_cache_pin();
-			Hypertable *parent_ht = ts_hypertable_cache_get_entry(hcache, parent_oid);
-			ts_cache_release(hcache);
-			if (parent_ht)
+			if (ts_is_hypertable(parent_oid))
 				return true;
-		}
```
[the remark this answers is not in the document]
k-rus: Fixed.

src/hypertable_cache.c
```diff
@@ -129,24 +130,55 @@ ts_hypertable_cache_invalidate_callback(void)
...
 
-	return ts_hypertable_cache_get_entry_with_table(cache, relid, NULL, NULL);
+/*
+ * Returns cache into the argument and hypertable as the function result.
+ * If hypertable is not found, fails with an error.
+ */
+Hypertable *
+ts_hypertable_cache_get_cache_and_entry(const Oid relid, const bool missing_ok, Cache **const cache)
+{
+	*cache = ts_hypertable_cache_pin();
```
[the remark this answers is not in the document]
k-rus: All caches are pinned in our code (in most of the cases for sure). I don't think it is worth to update here and all other places.

src/hypertable_cache.c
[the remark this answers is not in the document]
k-rus: Fixed

PR #1666, opened 2020-01-30
Drop chunks from materialized hypertables

### LINKED ISSUE #1639: Continuous Aggregates Need options to drop data
Ubuntu 16.04
TimescaleDB version - 1.6
Docker
--------------------
In version 1.6 you've added additional functionality for continuous aggregates that allows us to ignore invalidation and therefore we can keep data in the cagg even if we drop the raw data. This is great but only if we have options to limit the amount of data we keep in the continuous aggregate.

For example - if we use a cagg to rollup to a 1 minute interval. If we have a lot of items we store data for we could quickly start hitting memory limits from having too much 1 minute data.

It would be nice to be able to use drop_chunks() on a continuous aggregate so that we can manage the data retention for rollups of different widths.

An example - if we have 100,000 stock symbols and we collect raw tick data for each into a hypertable we can then rollup to 1m intervals using a continuous agg. We can then drop the raw data after,say,  48 hours.  The problem is that after a couple of weeks we'll have too much 1 minute data

2 weeks * 7 days * 1440 minutes * 100,000 symbols > 2 Billion rows in our continuous agg view.

gayyappan:
Add support for dropping chunks from materialized
hypertables. drop_chunks_policy can now be set up
for materialized hypertables.
Fixes #1639

Note to Reviewers:
This PR supports:
A. add_drop_chunks_policy( '<schema_name>.<view_name>' , ... )
cannot pass in materialized hypertable name directly here.
Policies are added by specifying the cont. aggregate view name.
If the cont. aggregate is based on "integer-type" time table, then the integer_now function is inherited by the cont. aggregate as well.
B. drop_chunks support for materialized hypertables.
1) drop_chunks( 'schema_name', 'view_name' , ... ) supported
2) drop_chunks ('_timescaledb_internal', ' materialized_hypertable_name', ... ) is supported.
3) drop_chunks( older_than => ..)
The table_name/view_name is not specifically passed in.
This variant errors out if any materialization hypertable satisfies the criterion for older_than.
(see continuous_aggs_ddl.sql.in for the variants )

If the schema_name is not passed in, the "search_path" is used to determine the schemas that will be searched for tables matching the criterion.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -2392,11 +2392,14 @@ ts_chunk_drop_process_materialization(Oid hypertable_relid,
 	ts_cache_release(hcache);
 }
 
+/* Continuous agg materialization hypertables can be dropped
+ * only if a user explicitly specifies the table name
+ */
 List *
 ts_chunk_do_drop_chunks(Oid table_relid, Datum older_than_datum, Datum newer_than_datum,
 						Oid older_than_type, Oid newer_than_type, bool cascade,
 						CascadeToMaterializationOption cascades_to_materializations,
-						int32 log_level)
+						int32 log_level, bool user_supplied_table_name)
 {
 	uint64 i = 0;
 	uint64 num_chunks = 0;
@@ -2413,9 +2416,18 @@ ts_chunk_do_drop_chunks(Oid table_relid, Datum older_than_datum, Datum newer_tha
 	switch (ts_continuous_agg_hypertable_status(hypertable_id))
 	{
 		case HypertableIsMaterialization:
+			if (user_supplied_table_name == false)
+			{
+				elog(ERROR, "cannot drop chunks on a continuous aggregate materialization table");
+			}
+			has_continuous_aggs = false;
+			break;
 		case HypertableIsMaterializationAndRaw:
-			elog(ERROR, "cannot drop_chunks on a continuous aggregate materialization table");
-			pg_unreachable();
+			if (user_supplied_table_name == false)
+			{
+				elog(ERROR, "cannot drop chunks on a continuous aggregate materialization table");
+			}
+			has_continuous_aggs = true;
 			break;
 		case HypertableIsRawTable:
 			if (cascades_to_materializations == CASCADE_TO_MATERIALIZATION_UNKNOWN)
@@ -2485,7 +2497,8 @@ ts_chunk_do_drop_chunks(Oid table_relid, Datum older_than_datum, Datum newer_tha
 																older_than_type,
 																newer_than_type,
 																cascade,
-																log_level);
+																log_level,
+																user_supplied_table_name);
 	}
 	return dropped_chunk_names;
 }
@@ -2550,6 +2563,7 @@ ts_chunk_drop_chunks(PG_FUNCTION_ARGS)
 	bool cascade, verbose;
 	CascadeToMaterializationOption cascades_to_materializations;
 	int elevel;
+	bool user_supplied_table_name = true;
 
 	/*
 	 * When past the first call of the SRF, dropping has already been completed,
@@ -2585,10 +2599,27 @@ ts_chunk_drop_chunks(PG_FUNCTION_ARGS)
 	if (table_name != NULL)
 	{
 		if (ht_oids == NIL)
-			ereport(ERROR,
-					(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
-					 errmsg("hypertable \"%s\" does not exist", NameStr(*table_name))));
+		{
+			ContinuousAgg *ca = NULL;
+			ca = ts_continuous_agg_find_userview_name(schema_name ? NameStr(*schema_name) : NULL,
+													  NameStr(*table_name));
+			if (ca == NULL)
+				ereport(ERROR,
+						(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+						 errmsg("\"%s\" is not a hypertable or a continuous aggregate view",
+								NameStr(*table_name)),
+						 errhint("It is only possible to drop chunks from a hypertable or "
+								 "continuous aggregate view")));
+			else
+			{
+				int32 matid = ca->data.mat_hypertable_id;
+				Hypertable *mat_ht = ts_hypertable_get_by_id(matid);
+				ht_oids = lappend_oid(ht_oids, mat_ht->main_table_relid);
+			}
+		}
 	}
+	else
+		user_supplied_table_name = false;
 
 	/* Initial multi function call setup */
 	funcctx = SRF_FIRSTCALL_INIT();
@@ -2657,7 +2688,8 @@ ts_chunk_drop_chunks(PG_FUNCTION_ARGS)
 										  newer_than_type,
 										  cascade,
 										  cascades_to_materializations,
-										  elevel);
+										  elevel,
+										  user_supplied_table_name);
 		dc_names = list_concat(dc_names, dc_temp);
 
 		MemoryContextSwitchTo(oldcontext);
--- a/src/continuous_agg.c
+++ b/src/continuous_agg.c
@@ -402,6 +402,41 @@ ts_continuous_agg_find_by_view_name(const char *schema, const char *name)
 }
 
 ContinuousAgg *
+ts_continuous_agg_find_userview_name(const char *schema, const char *name)
+{
+	ScanIterator iterator =
+		ts_scan_iterator_create(CONTINUOUS_AGG, AccessShareLock, CurrentMemoryContext);
+	ContinuousAgg *ca = NULL;
+	int count = 0;
+	const char *chkschema = schema;
+
+	ts_scanner_foreach(&iterator)
+	{
+		ContinuousAggViewType vtyp;
+		FormData_continuous_agg *data =
+			(FormData_continuous_agg *) GETSTRUCT(ts_scan_iterator_tuple(&iterator));
+		if (schema == NULL)
+		{
+			/* only user visible views will be returned */
+			Oid relid = RelnameGetRelid(NameStr(data->user_view_name));
+			if (relid == InvalidOid)
+				continue;
+			chkschema = NameStr(data->user_view_schema);
+		}
+
+		vtyp = ts_continuous_agg_view_type(data, chkschema, name);
+		if (vtyp == ContinuousAggUserView)
+		{
+			ca = palloc0(sizeof(*ca));
+			continuous_agg_init(ca, data);
+			count++;
+		}
+	}
+	Assert(count <= 1);
+	return ca;
+}
+
+ContinuousAgg *
 ts_continuous_agg_find_by_job_id(int32 job_id)
 {
 	ScanIterator iterator =
@@ -782,4 +817,52 @@ ts_continuous_agg_get_user_view_oid(ContinuousAgg *agg)
 	if (!OidIsValid(view_relid))
 		elog(ERROR, "could not find user view for continuous agg");
 	return view_relid;
+}
+
+static int32
+find_raw_hypertable_for_materialization(int32 mat_hypertable_id)
+{
+	short count = 0;
+	int32 htid = INVALID_HYPERTABLE_ID;
+	ScanIterator iterator =
+		ts_scan_iterator_create(CONTINUOUS_AGG, RowExclusiveLock, CurrentMemoryContext);
+
+	init_scan_by_mat_hypertable_id(&iterator, mat_hypertable_id);
+	ts_scanner_foreach(&iterator)
+	{
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		HeapTuple tuple = ti->tuple;
+		Form_continuous_agg form = (Form_continuous_agg) GETSTRUCT(tuple);
+		htid = form->raw_hypertable_id;
+		count++;
+	}
+	Assert(count <= 1);
+	ts_scan_iterator_close(&iterator);
+	return htid;
+}
+
+/* Continuous aggregate materialization hypertables inherit integer_now func
+ * from the raw hypertable (unless it was explictly reset for cont. aggregate.
+ * Walk the materialzation hyperatable ->raw hypertable tree till
+ * we find a hypertable that has integer_now_func set.
+ */
+TSDLLEXPORT Dimension *
+ts_continous_agg_find_integer_now_func_by_materialization_id(int32 mat_htid)
+{
+	int32 raw_htid = mat_htid;
+	Dimension *par_dim = NULL;
+	while (raw_htid != INVALID_HYPERTABLE_ID)
+	{
+		Hypertable *raw_ht = ts_hypertable_get_by_id(raw_htid);
+		Dimension *open_dim = hyperspace_get_open_dimension(raw_ht->space, 0);
+		if (strlen(NameStr(open_dim->fd.integer_now_func)) != 0 &&
+			strlen(NameStr(open_dim->fd.integer_now_func_schema)) != 0)
+		{
+			par_dim = open_dim;
+			break;
+		}
+		mat_htid = raw_htid;
+		raw_htid = find_raw_hypertable_for_materialization(mat_htid);
+	}
+	return par_dim;
 }
--- a/src/interval.c
+++ b/src/interval.c
@@ -95,7 +95,7 @@ ts_interval_from_sql_input(Oid relid, Datum interval, Oid interval_type, const c
 {
 	Hypertable *hypertable;
 	Cache *hcache;
-	FormData_ts_interval *invl = palloc0(sizeof(FormData_ts_interval));
+	FormData_ts_interval *invl;
 	Oid partitioning_type;
 	Dimension *open_dim;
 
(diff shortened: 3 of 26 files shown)


src/chunk.c
```diff
@@ -2392,11 +2392,17 @@ ts_chunk_drop_process_materialization(Oid hypertable_relid,
 	ts_cache_release(hcache);
 }
 
+/* Continuous agg materialization hypertables can be dropped
+ * only if a user explicitly specifies the table name
+ * see discussion related to drop_chunks API:
+ * https://github.com/timescale/timescaledb/pull/1648
+ * API needs to be deprecated. this is the short term fix until then.
```
erimatnor: I am not 100% convinced this API needs deprecation, as discussed before. Let's remove this part of the comment for now and continue the discussion.

src/chunk.c
```diff
@@ -2413,9 +2419,18 @@ ts_chunk_do_drop_chunks(Oid table_relid, Datum older_than_datum, Datum newer_tha
 	switch (ts_continuous_agg_hypertable_status(hypertable_id))
 	{
 		case HypertableIsMaterialization:
+			if (user_supplied_table_name == false)
+			{
+				elog(ERROR, "cannot drop_chunks on a continuous aggregate materialization table");
```
erimatnor:
```suggestion
				elog(ERROR, "cannot drop chunks on a continuous aggregate materialization table");
```

src/chunk.c
```diff
@@ -2413,9 +2419,18 @@ ts_chunk_do_drop_chunks(Oid table_relid, Datum older_than_datum, Datum newer_tha
...
+			}
+			has_continuous_aggs = false;
+			break;
 		case HypertableIsMaterializationAndRaw:
-			elog(ERROR, "cannot drop_chunks on a continuous aggregate materialization table");
-			pg_unreachable();
+			if (user_supplied_table_name == false)
+			{
+				elog(ERROR, "cannot drop_chunks on a continuous aggregate materialization table");
```
erimatnor:
```suggestion
				elog(ERROR, "cannot drop chunks on a continuous aggregate materialization table");
```

src/chunk.c
```diff
@@ -2392,11 +2392,17 @@ ts_chunk_drop_process_materialization(Oid hypertable_relid,
...
+ * see discussion related to drop_chunks API:
+ * https://github.com/timescale/timescaledb/pull/1648
+ * API needs to be deprecated. this is the short term fix until then.
+ */
 List *
 ts_chunk_do_drop_chunks(Oid table_relid, Datum older_than_datum, Datum newer_than_datum,
 						Oid older_than_type, Oid newer_than_type, bool cascade,
 						CascadeToMaterializationOption cascades_to_materializations,
-						int32 log_level)
+						int32 log_level, bool user_supplied_table_name)
```
erimatnor: It seems to me that this function could just be kept as is if we don't recurse to materialization tables in the user-visible call to `drop_chunks`, unless the user explicitly told us to:

```
drop_chunks(older_than => '1 week');   # <--- only drop chunks on user-created hypertables
drop_chunks(older_than => '1 week', cascade_to_materializations => true);   # <---  drop chunks on user-created hypertables and associated materializations
```

src/chunk.c
```diff
@@ -2580,15 +2594,34 @@ ts_chunk_drop_chunks(PG_FUNCTION_ARGS)
...
+		/* default schema is public */
+		schema_name = (NameData *) palloc0(NAMEDATALEN);
+		namestrcpy(schema_name, "public");
+	}
 	ht_oids = ts_hypertable_get_all_by_name(schema_name, table_name, CurrentMemoryContext);
 
 	if (table_name != NULL)
 	{
+		ContinuousAgg *ca = NULL;
 		if (ht_oids == NIL)
```
erimatnor: Why isn't `if (ht_oids == NIL)` the outer check here? Seems more reasonable to do this check immediately following `ts_hypertable_get_all_by_name` since it returns this list.

src/chunk.c
```diff
@@ -2580,15 +2594,34 @@ ts_chunk_drop_chunks(PG_FUNCTION_ARGS)
...
-			ereport(ERROR,
-					(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
-					 errmsg("hypertable \"%s\" does not exist", NameStr(*table_name))));
+		{
+			ca = ts_continuous_agg_find_by_view_name(NameStr(*schema_name), NameStr(*table_name));
+			if (ca == NULL)
+				ereport(ERROR,
+						(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+						 errmsg("hypertable \"%s\" does not exist", NameStr(*table_name))));
```
erimatnor: I guess this error is a bit confusing in case one passed in the view name. I suggest updating the error to say something like: `errmsg("%s is not a hypertable or continuous aggregate view")` `errhint("It is only possible to drop chunks on a user-created hypertable or continuous aggregate view.")`

src/chunk.c
```diff
@@ -2580,15 +2594,34 @@ ts_chunk_drop_chunks(PG_FUNCTION_ARGS)
...
+	{
+		/* default schema is public */
+		schema_name = (NameData *) palloc0(NAMEDATALEN);
+		namestrcpy(schema_name, "public");
+	}
 	ht_oids = ts_hypertable_get_all_by_name(schema_name, table_name, CurrentMemoryContext);
 
 	if (table_name != NULL)
 	{
+		ContinuousAgg *ca = NULL;
```
erimatnor: Move declaration inside the local context where it is used.

src/continuous_agg.c
```diff
@@ -783,3 +783,51 @@ ts_continuous_agg_get_user_view_oid(ContinuousAgg *agg)
 		elog(ERROR, "could not find user view for continuous agg");
 	return view_relid;
 }
+static int32
```
erimatnor: Add newline

src/continuous_agg.c
```diff
@@ -783,3 +783,51 @@ ts_continuous_agg_get_user_view_oid(ContinuousAgg *agg)
...
+find_raw_hypertable_for_materialization(int32 mat_hypertable_id)
+{
+	short count = 0;
```
erimatnor: why a `short` and not `int`?

src/continuous_agg.c
```diff
@@ -783,3 +783,51 @@ ts_continuous_agg_get_user_view_oid(ContinuousAgg *agg)
...
+ */
+TSDLLEXPORT Dimension *
+ts_continous_agg_find_integer_now_func_by_materialization_id(int32 mat_htid)
+{
+	int32 raw_htid = mat_htid;
+	Dimension *par_dim = NULL;
+	while (raw_htid != INVALID_HYPERTABLE_ID)
+	{
+		Hypertable *raw_ht = ts_hypertable_get_by_id(raw_htid);
+
```
erimatnor: move newline to below next declaration

src/interval.c
```diff
@@ -109,8 +109,33 @@ ts_interval_from_sql_input(Oid relid, Datum interval, Oid interval_type, const c
 		elog(ERROR, "internal error: no open dimension found while parsing interval");
 
 	partitioning_type = ts_dimension_get_partition_type(open_dim);
+	if (IS_INTEGER_TYPE(partitioning_type))
+	{
+		if (strlen(NameStr(open_dim->fd.integer_now_func)) == 0 ||
+			strlen(NameStr(open_dim->fd.integer_now_func_schema)) == 0)
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("integer_now_func not set on hypertable %s", get_rel_name(relid))));
```
erimatnor:
```suggestion
					 errmsg("integer_now_func not set on hypertable \"%s\"", get_rel_name(relid))));
```

src/interval.c
```diff
@@ -109,8 +109,33 @@ ts_interval_from_sql_input(Oid relid, Datum interval, Oid interval_type, const c
 		elog(ERROR, "internal error: no open dimension found while parsing interval");
 
 	partitioning_type = ts_dimension_get_partition_type(open_dim);
+	if (IS_INTEGER_TYPE(partitioning_type))
+	{
+		if (strlen(NameStr(open_dim->fd.integer_now_func)) == 0 ||
+			strlen(NameStr(open_dim->fd.integer_now_func_schema)) == 0)
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
```
erimatnor: I suggest a different error code, since this one is reserved for invalid parameter values when calling (user-facing) SQL functions. For instance, user calls `SELECT set_port(port => 134000);` then the value for `port` is invalid (outside the port range). 

 Also, in my understanding, having a NULL integer now func is valid.

I suggest INTERNAL_ERROR or just a simple `elog`.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -37,6 +38,89 @@
...
+
+static void
+validate_drop_chunks_hypertable(Cache *hcache, Oid user_htoid, Oid older_than_type,
+								Datum older_than_datum, DropChunksMeta *meta)
+{
+	FormData_ts_interval *older_than;
+	ContinuousAggHypertableStatus status;
+
+	meta->ht = NULL;
+	meta->ht_oid = user_htoid;
```
erimatnor: I would move initialization of this struct outside this function. It seems that now you just need to pass in the struct and the parameters to fill in instead of just passing in the struct. Instead, the caller can do:

```
DropChunksMeta meta = {
   .ht_oid = user_htoid,
   .hcache = ts_hypertable_cache_pin(),
};
```

and avoid passing in the extra parameters.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -37,6 +38,89 @@
...
+	meta->is_matht = false;
+	meta->ht = ts_hypertable_cache_get_entry(hcache, meta->ht_oid, true /* missing_ok */);
+	if (meta->ht != NULL)
+	{
+		if (meta->ht->fd.compressed)
+		{
+			ts_cache_release(hcache);
```
erimatnor: Since the cache is pinned outside this function, I think it should also be released in that same context. Otherwise, it is hard to read the code and know where release happens. 

Note that the cache is auto-released on errors.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -37,6 +38,89 @@
...
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("cannot add drop chunks policy to hypertable \"%s\" which contains "
```
erimatnor: Suggestion: `"cannot add drop chunks policy to compressed hypertable \"%s\""`

(This also makes it consistent with the error below)

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -37,6 +38,89 @@
...
+							get_rel_name(user_htoid)),
+					 errhint("Please add the policy to the corresponding uncompressed hypertable "
+							 "instead.")));
+		}
+		status = ts_continuous_agg_hypertable_status(meta->ht->fd.id);
+		if ((status == HypertableIsMaterialization || status == HypertableIsMaterializationAndRaw))
+		{
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("cannot add drop chunks policy to materialized hypertables \"%s\" ",
```
erimatnor:
```suggestion
					 errmsg("cannot add drop chunks policy to materialized hypertable \"%s\" ",
```

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -37,6 +38,89 @@
...
+					(errcode(ERRCODE_TS_HYPERTABLE_NOT_EXIST),
+					 errmsg("\"%s\" is not a hypertable or a continuous aggregate view",
+							view_name)));
+		mat_id = ca->data.mat_hypertable_id;
+		meta->ht = ts_hypertable_get_by_id(mat_id);
+		Assert(meta->ht != NULL);
+		open_dim = hyperspace_get_open_dimension(meta->ht->space, 0);
+		partitioning_type = ts_dimension_get_partition_type(open_dim);
+		if (IS_INTEGER_TYPE(partitioning_type))
+		{
```
erimatnor: elide braces

tsl/src/bgw_policy/job.c
```diff
@@ -169,6 +169,30 @@ execute_reorder_policy(BgwJob *job, reorder_func reorder, bool fast_continue)
...
+	if (IS_INTEGER_TYPE(partitioning_type))
+	{
+		/* if this a materialization hypertable related to cont agg
+		 * then need to get the right dimension which has
+		 * integer_now function
+		 */
+
+		open_dim = ts_continous_agg_find_integer_now_func_by_materialization_id(mat_id);
+		if (open_dim == NULL)
+		{
```
erimatnor: elide braces

tsl/test/expected/continuous_aggs_bgw_drop_chunks.out
```diff
@@ -0,0 +1,123 @@
...
+ORDER BY ranges;
+               chunk_table               |   ranges    
+-----------------------------------------+-------------
+ _timescaledb_internal._hyper_2_40_chunk | {"[0,10)"}
+ _timescaledb_internal._hyper_2_41_chunk | {"[10,20)"}
+ _timescaledb_internal._hyper_2_42_chunk | {"[20,30)"}
+ _timescaledb_internal._hyper_2_43_chunk | {"[30,40)"}
+(4 rows)
+
+SELECT add_drop_chunks_policy( 'drop_chunks_view1', older_than=> 10, cascade_to_materializations=>false ) as drop_chunks_job_id1 \gset
```
erimatnor: What would `cascade_to_materializations=true` mean if this is executed on a view? Does it just silently do nothing or does it error out?

tsl/test/expected/continuous_aggs_ddl-10.out
```diff
@@ -1048,3 +1045,36 @@ ORDER BY ranges;
  _timescaledb_internal._hyper_13_22_chunk | {"[20,40)"}
 (1 row)
 
+-- TEST drop chunks from continuous aggregates by specifying view name
+SELECT drop_chunks(
+    table_name => 'drop_chunks_view',
```
erimatnor: Same here as above. What is the behavior of `cascade_to_materializations = true` when `drop_chunks` is called on the view?

Do we need to test?

tsl/test/expected/continuous_aggs_bgw_drop_chunks.out
gayyappan: it silently does nothing now.

tsl/test/expected/continuous_aggs_ddl-10.out
gayyappan: there are no dependencies on this view/materialized hypertable. But I will add a contrived case and test. This would help  when we extend functionality to create views based on existing views (like  1 day view based on 1 hour view).

src/chunk.c
gayyappan: restricting dropping  chunks from cagg view only if  the view name is explicitly specified.

src/continuous_agg.c
gayyappan: doesn't really matter here. This should be <= 1 after the search.

src/interval.c
gayyappan: I believe this code path is exercised by drop_chunks_policy. So this is a user facing error.

erimatnor: My point is that `integer_now_func` and `integer_now_func_schema` aren't, AFAIK, set as part of those functions, like in my `set_port` example. Further, the error is not about a function parameter but about some values of internal metadata being NULL.

So, I am not sure it makes sense to raise `ERRCODE_INVALID_PARAMETER_VALUE`. The error should tell you that the system is in the wrong state for the given operation.

gayyappan: These are errors for the "older_than" parameter for the policy. There is a specific action needed on the user's part to fix this issue. older_than parameter on integer-time based  hypertables doesn't make sense without specifying a now() function. INTERNAL_ERROR to me typically signifies something unexpected happening in the code which is probably a bug.

tsl/test/expected/continuous_aggs_ddl-10.out
gayyappan: Update: contrived case doesn't work as we disable cont. aggregates based on existing cont. aggregates.

src/chunk.c
```diff
@@ -2580,15 +2594,37 @@ ts_chunk_drop_chunks(PG_FUNCTION_ARGS)
 				 errmsg("older_than and newer_than timestamps provided to drop_chunks cannot both "
 						"be NULL")));
 
+	if (schema_name == NULL)
+	{
+		/* default schema is public */
+		schema_name = (NameData *) palloc0(NAMEDATALEN);
+		namestrcpy(schema_name, "public");
```
svenklemm: the schema shouldnt be hardcoded like that and you should call current_schema to get default schema instead

gayyappan: The behavior described in the docs is that we will only look for "public" schema if it is not specified explicitly. We were missing this check earlier.

tsl/src/bgw_policy/drop_chunks_api.c
```diff
@@ -57,27 +137,18 @@ drop_chunks_add_policy(PG_FUNCTION_ARGS)
...
-						get_rel_name(ht_oid)),
-				 errhint("Please add the policy to the corresponding uncompressed hypertable "
-						 "instead.")));
-
-	older_than = ts_interval_from_sql_input(ht_oid,
-											older_than_datum,
-											older_than_type,
-											"older_than",
-											"add_drop_chunks_policy");
+	hcache = ts_hypertable_cache_pin();
```
pmwkaa: Do we need to unpin the cache after usage in this function?

gayyappan: not needed as cache gets unpinned automatically at end of txn.

svenklemm:
Can you add a test where a hypertable with the same name exists in schema a and schema b which are both in search path

tsl/src/bgw_policy/drop_chunks_api.c
erimatnor: Seems to be released already. 

Generally, though, probably good practice to have explicit unpins/releases to avoid holding on to memory longer than necessary.

PR #1668, opened 2020-01-31
Cannot add dimension if table has empty chunks

### LINKED ISSUE #1623: Missing constraint after add_dimension

 - TimescaleDB version: 1.5.1

**Describe the bug**
If `add_dimension` is used on an empty hypertable, although chunks exist (i.e., from having previously inserted data and deleted), the existing chunks will not be updated to with constraints for the new dimension.

**To Reproduce**

```
postgres=# create table foo (time timestamptz, device int, temp float);
CREATE TABLE
postgres=# select create_hypertable('foo', 'time');
DETAIL:  Time dimensions cannot have NULL values
 create_hypertable 
-------------------
 (3,public,foo,t)
(1 row)

postgres=# insert into foo values (now()-interval '3 months', 1, 1.0);
INSERT 0 1
postgres=# delete from foo where device=1;
DELETE 1
postgres=# select add_dimension('foo', 'device', 2);
      add_dimension      
-------------------------
 (6,public,foo,device,t)
(1 row)


postgres=# \d+ _timescaledb_internal._hyper_3_4_chunk
                              Table "_timescaledb_internal._hyper_3_4_chunk"
 Column |           Type           | Collation | Nullable | Default | Storage | Stats target | Description 
--------+--------------------------+-----------+----------+---------+---------+--------------+-------------
 time   | timestamp with time zone |           | not null |         | plain   |              | 
 device | integer                  |           |          |         | plain   |              | 
 temp   | double precision         |           |          |         | plain   |              | 
Indexes:
    "_hyper_3_4_chunk_foo_time_idx" btree ("time" DESC)
Check constraints:
    "constraint_5" CHECK ("time" >= '2019-10-10 02:00:00+02'::timestamp with time zone AND "time" < '2019-10-17 02:00:00+02'::timestamp with time zone)
Inherits: foo

```

Note that the chunk only has the time-dimension constraint.

**Expected behavior**

Should probably see one of these behaviors:

1. The existing chunks should be updated with the new constraints, or 
2. The command should fail with an error saying that old chunks need to be removed, or
3. The empty chunks should be transparently removed when calling `add_dimension`.

**Actual behavior**
Existing (empty) chunks will miss important constraints.

gayyappan:
add_dimension should fail when table has no
data but still has empty chunks.
Fixes #1623

### DIFF
--- a/src/dimension.c
+++ b/src/dimension.c
@@ -1355,11 +1355,13 @@ ts_dimension_add(PG_FUNCTION_ARGS)
 
 	if (!info.skip)
 	{
-		if (ts_hypertable_has_tuples(info.table_relid, AccessShareLock))
+		if (ts_hypertable_has_chunks(info.table_relid, AccessShareLock))
 			ereport(ERROR,
 					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-					 errmsg("hypertable \"%s\" is not empty", get_rel_name(info.table_relid)),
-					 errdetail("It is not possible to add dimensions to a non-empty hypertable")));
+					 errmsg("hypertable \"%s\" has tuples or empty chunks",
+							get_rel_name(info.table_relid)),
+					 errdetail("It is not possible to add dimensions to a hypertable that has "
+							   "chunks. Please truncate the table.")));
 
 		/*
 		 * Note that space->num_dimensions reflects the actual number of
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -1250,6 +1250,16 @@ ts_hypertable_has_tuples(Oid table_relid, LOCKMODE lockmode)
 	return false;
 }
 
+bool
+ts_hypertable_has_chunks(Oid table_relid, LOCKMODE lockmode)
+{
+	List *chunks = find_inheritance_children(table_relid, lockmode);
+	if (chunks != NIL)
+		return true;
+	else
+		return false;
+}
+
 static void
 hypertable_create_schema(const char *schema_name)
 {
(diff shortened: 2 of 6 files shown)


src/hypertable.c
```diff
@@ -1250,6 +1250,17 @@ ts_hypertable_has_tuples(Oid table_relid, LOCKMODE lockmode)
 	return false;
 }
 
+bool
+ts_hypertable_has_chunks(Oid table_relid, LOCKMODE lockmode)
+{
+	ListCell *lc;
```
erimatnor: Unused variable.

src/hypertable.c
```diff
@@ -1250,6 +1250,17 @@ ts_hypertable_has_tuples(Oid table_relid, LOCKMODE lockmode)
...
+	List *chunks = find_inheritance_children(table_relid, lockmode);
```
erimatnor:
```suggestion
	return find_inheritance_children(table_relid, lockmode) != NIL;
```

Can be simplified.

src/dimension.c
```diff
@@ -1355,11 +1355,12 @@ ts_dimension_add(PG_FUNCTION_ARGS)
 
 	if (!info.skip)
 	{
-		if (ts_hypertable_has_tuples(info.table_relid, AccessShareLock))
+		if (ts_hypertable_has_chunks(info.table_relid, AccessShareLock))
 			ereport(ERROR,
 					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 					 errmsg("hypertable \"%s\" is not empty", get_rel_name(info.table_relid)),
```
erimatnor: "...is not empty" could be a confusing error message in case users deleted all data from the table already (but there are empty chunks). I suggest reformulating or keep this "has tuples" check in addition to the "has chunks" check.

src/hypertable.c
gayyappan: removed.

src/dimension.c
gayyappan: The errdetail actually gives more info about the error message. Also modified the error message .

PR #1674, opened 2020-02-04
Fix GapFill with ReScan

JLockerman:
The GapFill node was not fully reset on a ReScan, so if there was a
GapFill within a NestedLoop, only the first iteration would return
results. This commit fixes this issues.

**NOTE**: main reviewers are Cevian and Sven, others were added automatically


JLockerman:
@cevian see, I thought our conversation contained you explicitly telling me not to bother adding those in the name of getting it in faster ¯\_(ツ)_/¯
Creating one now.

PR #1715, opened 2020-02-28
Fix incorrect chunk exclusion and broken test 

erimatnor:
This change fixes the `plan_expand_hypertable` test, which was broken
and never ran the output comparison due to prematurely ending in an
uncaught error. The test appeared to succeeded, however, since also
the broken test "expected" files were committed to the repo.

Fixing the test revealed that the query output with our optimizations
enabled is incorrect for outer joins (i.e., the output from the query
differed from regular PostgreSQL). Restriction clauses were too
aggressively pushed down to outer relations, leading to chunk
exclusion when it shouldn't happen. This incorrect behavior has also
been fixed.

### DIFF
--- a/test/expected/plan_expand_hypertable-9.6.out
+++ b/test/expected/plan_expand_hypertable-9.6.out
@@ -88,12 +88,23 @@ psql:include/plan_expand_hypertable_load.sql:66: NOTICE:  adding not-null constr
 INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 1;
 INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 2;
 INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 3;
+--create a second table to test joins with
+CREATE TABLE metrics_timestamptz_2 (LIKE metrics_timestamptz);
+SELECT create_hypertable('metrics_timestamptz_2','time');
+         create_hypertable          
+------------------------------------
+ (7,public,metrics_timestamptz_2,t)
+(1 row)
+
+INSERT INTO metrics_timestamptz_2
+SELECT * FROM metrics_timestamptz;
+INSERT INTO metrics_timestamptz_2 VALUES ('2000-12-01'::timestamptz, 3);
 CREATE TABLE metrics_date(time date);
 SELECT create_hypertable('metrics_date','time');
-psql:include/plan_expand_hypertable_load.sql:72: NOTICE:  adding not-null constraint to column "time"
+psql:include/plan_expand_hypertable_load.sql:79: NOTICE:  adding not-null constraint to column "time"
      create_hypertable     
 ---------------------------
- (7,public,metrics_date,t)
+ (8,public,metrics_date,t)
 (1 row)
 
 INSERT INTO metrics_date SELECT generate_series('2000-01-01'::date,'2000-02-01'::date,'1d'::interval);
@@ -944,9 +955,9 @@ joins
 test constraint exclusion for constraints in ON clause of JOINs
 \qecho should exclude chunks on m1 and propagate qual to m2 because of INNER JOIN
 should exclude chunks on m1 and propagate qual to m2 because of INNER JOIN
-:PREFIX SELECT m1.time,m2.time FROM metrics_timestamptz m1 INNER JOIN metrics_timestamptz m2 ON m1.time = m2.time AND m1.time < '2000-01-10' ORDER BY m1.time;
-                                                         QUERY PLAN                                                          
------------------------------------------------------------------------------------------------------------------------------
+:PREFIX SELECT m1.time,m2.time FROM metrics_timestamptz m1 INNER JOIN metrics_timestamptz_2 m2 ON m1.time = m2.time AND m1.time < '2000-01-10' ORDER BY m1.time;
+                                                          QUERY PLAN                                                           
+-------------------------------------------------------------------------------------------------------------------------------
  Merge Join
    Merge Cond: (m1."time" = m2."time")
    ->  Custom Scan (ChunkAppend) on metrics_timestamptz m1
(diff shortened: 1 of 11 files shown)


test/expected/plan_expand_hypertable-9.6.out
```diff
@@ -998,6 +1086,9 @@ must not exclude on m1
          Order: m1."time"
          ->  Index Only Scan Backward using _hyper_6_160_chunk_metrics_timestamptz_time_idx on _hyper_6_160_chunk m1_1
          ->  Index Only Scan Backward using _hyper_6_161_chunk_metrics_timestamptz_time_idx on _hyper_6_161_chunk m1_2
+         ->  Index Only Scan Backward using _hyper_6_162_chunk_metrics_timestamptz_time_idx on _hyper_6_162_chunk m1_3
```
erimatnor: These are chunks that were previously excluded, leading to wrong output.

test/expected/plan_expand_hypertable-9.6.out
```diff
@@ -1070,6 +1244,9 @@ must not exclude chunks on m2
                Order: m2."time"
                ->  Index Only Scan Backward using _hyper_6_160_chunk_metrics_timestamptz_time_idx on _hyper_6_160_chunk m2_1
                ->  Index Only Scan Backward using _hyper_6_161_chunk_metrics_timestamptz_time_idx on _hyper_6_161_chunk m2_2
+               ->  Index Only Scan Backward using _hyper_6_162_chunk_metrics_timestamptz_time_idx on _hyper_6_162_chunk m2_3
```
erimatnor: Another example of chunks that were excluded that shouldn't be.

test/sql/include/plan_expand_hypertable_query.sql
```diff
@@ -141,19 +141,51 @@ SELECT * FROM cte ORDER BY value;
 
 \qecho test constraint exclusion for constraints in ON clause of JOINs
 
+-- Turn on chunk exclusion so that we can compare with how
+-- PostgreSQL does exclusion
+SET constraint_exclusion=on;
+SHOW constraint_exclusion;
+
 \qecho should exclude chunks on m1 and propagate qual to m2 because of INNER JOIN
+SET timescaledb.enable_constraint_exclusion = off;
```
gayyappan: I know these are existing test cases. But it would help to have different tables for JOIN , esp. RIGHT JOIN and LEFT JOIN cases. That helps easily identify when the chunks are being excluded and when they are not.

test/sql/include/plan_expand_hypertable_query.sql
```diff
@@ -141,19 +141,51 @@ SELECT * FROM cte ORDER BY value;
...
+:PREFIX SELECT m1.time,m2.time FROM metrics_timestamptz m1 INNER JOIN metrics_timestamptz m2 ON m1.time = m2.time AND m2.time < '2000-01-10' ORDER BY m1.time;
+SET timescaledb.enable_constraint_exclusion = on;
 :PREFIX SELECT m1.time,m2.time FROM metrics_timestamptz m1 INNER JOIN metrics_timestamptz m2 ON m1.time = m2.time AND m2.time < '2000-01-10' ORDER BY m1.time;
+
 \qecho must not exclude on m1
+SET timescaledb.enable_constraint_exclusion = off;
+:PREFIX SELECT m1.time,m2.time FROM metrics_timestamptz m1 LEFT JOIN metrics_timestamptz m2 ON m1.time = m2.time AND m1.time < '2000-01-10' ORDER BY m1.time;
+SET timescaledb.enable_constraint_exclusion = on;
 :PREFIX SELECT m1.time,m2.time FROM metrics_timestamptz m1 LEFT JOIN metrics_timestamptz m2 ON m1.time = m2.time AND m1.time < '2000-01-10' ORDER BY m1.time;
```
gayyappan: CREATE TABLE metrics_timestamp_2 as select * from metrics_timestamp;
SELECT create_hypertable(...);

Then use this new table for LEFT and RIGHT JOIN test cases

```suggestion
:PREFIX SELECT m1.time,m2.time FROM metrics_timestamptz m1 LEFT JOIN metrics_timestamptz_2 m2 ON m1.time = m2.time AND m1.time < '2000-01-10' ORDER BY m1.time;
```

erimatnor:
@svenklemm I can remove the explains, but I found it useful to see the actual difference between our exclusion and PGs. I would expect a reviewer to feel the same thing.

The fact that this was reviewed and merged previously without running the diff tests illustrates the weakness of relying only on _absence_ of output. Probably something we should fix because otherwise this will surely happen again.

test/sql/include/plan_expand_hypertable_query.sql
erimatnor: I thought about it, but I don't think it makes a practical difference. From the planners point of view, these are different rels (RelOptInfos) that just happens to point to the same underlying table. So, even if the table is the same, it can exclude differently depending on the planned rel, as shown in the examples (i.e., exclusion can happen on m1 but node m2, even if same table).

gayyappan: Yes, of course, it works the same way. It is easy to identify the tables in the explain plan output when they have different chunk names.

erimatnor: Ok, let me see what I can do

gayyappan:
> @svenklemm I can remove the explains, but I found it useful to see the actual difference between our exclusion and PGs. I would expect a reviewer to feel the same thing.
> 
> The fact that this was reviewed and merged previously without running the diff tests illustrates the weakness of relying only on _absence_ of output. Probably something we should fix because otherwise this will surely happen again.

We should keep the explain output because these test cases *will not* produce different output for LEFT and RIGHT joins. They are equivalent to the INNER JOIN here, I believe. It would be good to actually have the tests produce different output for the queries and then remove the explains.This is unfortunately a problem if the data for the tests are not carefully targetted for that particular query.

test/sql/include/plan_expand_hypertable_query.sql
erimatnor: Ok, this is now implemented.

test/sql/include/plan_expand_hypertable_load.sql
```diff
@@ -68,6 +68,12 @@ INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz
 INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 2;
 INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz,'2000-02-01'::timestamptz,'1d'::interval), 3;
```
gayyappan:
```suggestion
INSERT INTO metrics_timestamptz values( '2000-03-01'::timestamptz, 3);
```

test/sql/include/plan_expand_hypertable_load.sql
```diff
@@ -68,6 +68,12 @@ INSERT INTO metrics_timestamptz SELECT generate_series('2000-01-01'::timestamptz
...
 
+--create a second table to test joins with
+CREATE TABLE metrics_timestamptz_2 (LIKE metrics_timestamptz);
+SELECT create_hypertable('metrics_timestamptz_2','time');
+INSERT INTO metrics_timestamptz_2
+SELECT * FROM metrics_timestamptz;
+
```
gayyappan: adding a  row to metric_timestamptz and metric_timestamptz_2 which are different for both tables so that RIGHT/LEFT joins give different output.

```suggestion
INSERT INTO metrics_timestamptz_2 values( '2000-12-01'::timestamptz, 3);
```

test/sql/include/plan_expand_hypertable_load.sql
gayyappan: this might change the plan output for other plans as well. But the other one should be easy to add without too much diff noise.

erimatnor:
> I dont mind the EXPLAIN being included but it should not be in that file and that file should not modify optimization settings since the parent of that file modifies those and runs the file with different settings.

I guess you didn't actually re-review the code as I already addressed your concerns. Or is there something else you are referring to?

PR #1719, opened 2020-03-04
Fix restoring/scheduler entrypoint to avoid BGW death

### LINKED ISSUE #1712: pg_dump/pg_restore with continious aggregates fails
 - OS: Ubuntu 18.04
 - PostgreSQL version 11.7
 - TimescaleDB version 1.6.0

**Describe the bug**
A DB backup/restore procedure is executed as explained [in the TimescaleDB docu](https://docs.timescale.com/latest/using-timescaledb/backup).
However the `pg_restore` step is executed in a separate shell.
Running then `SELECT timescaledb_post_restore();` results into a PG ERROR log
`ERROR:  tried calling catalog_get when extension isn't loaded` and `LOG:  background worker "TimescaleDB Background Worker Scheduler" (PID 31298) exited with exit code 1` from TimescaleDB.
The hypertables and materialized views are populated, however, the Background worker is dead and does no longer update the materialized view (continuous aggregate).

**Expected behavior**
The continuous aggregate should work fine, after a dump/restore.

**Actual behavior**
The continuous aggregate Background worker are never scheduled.

**Note**
Restarting postgres fixes the issue, i.e. the continuous aggregate is populated again, if new data are inserted/updated.

davidkohn88:
There was a race condition between the post_restore function
restarting the background worker and the setting of the
restoring flag to "off". If the worker started before the
change to the restoring flag had been committed, it would not
see the change and then die because the worker should exit
when the db is in a restoring state. This modifies the
post_restore function to use a restart instead of a start
so that it waits on the commit to start up. It also adds
logic to the entrypoint to reload config changes caused
by an `ALTER DATABASE SET` command. These changes are
normally only seen at connection startup but we have to
wait until after our lock on the modifying transaction is
released to know whether we should adopt them.

Fixes #1712 

Changes the loader so would require a restart on upgrade.

### DIFF
--- a/sql/restoring.sql
+++ b/sql/restoring.sql
@@ -27,7 +27,7 @@ BEGIN
     SELECT current_database() INTO db;
     EXECUTE format($$ALTER DATABASE %I SET timescaledb.restoring ='off'$$, db);
     SET SESSION timescaledb.restoring='off';
-    PERFORM _timescaledb_internal.start_background_workers();
+    PERFORM _timescaledb_internal.restart_background_workers();
 
     --try to restore the backed up uuid, if the restore did not set one
     INSERT INTO _timescaledb_catalog.metadata
--- a/src/loader/bgw_launcher.c
+++ b/src/loader/bgw_launcher.c
@@ -40,6 +40,9 @@
 /* for allocating the htab storage */
 #include <utils/memutils.h>
 
+/* for getting settings correct before loading the versioned scheduler */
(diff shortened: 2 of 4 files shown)


davidkohn88:
There is the possibility of moving some of these changes to the versioned extension so we don't need any changes to the loader/require a restart on upgrade, but it seems significantly less intuitive code-wise to do it there to me.

PR #1720, opened 2020-03-04
Scheduler cache invalidations

gayyappan:
Make scheduler more resilient to failures
    
    Invalidate cache state when jobs are detected as
    potentially missing. This will update the jobs list
    used by the scheduler.
    Restrict job start times to a finite interval when
    there are consecutive job failures.
    Additional logging.

### DIFF
--- a/src/bgw/job_stat.c
+++ b/src/bgw/job_stat.c
@@ -19,6 +19,7 @@
 #include <utils/fmgrprotos.h>
 #endif
 #define MAX_INTERVALS_BACKOFF 5
+#define MAX_FAILURES_MULTIPLIER 20
 #define MIN_WAIT_AFTER_CRASH_MS (5 * 60 * 1000)
 
 static bool
@@ -186,17 +187,22 @@ calculate_jitter_percent()
 	return ldexp((double) (16 - (int) (percent % 32)), -7);
 }
 
-/* For failures we have standard exponential backoff based on consecutive failures
- * along with a ceiling at schedule_interval * MAX_INTERVALS_BACKOFF */
+/* For failures we have additive backoff based on consecutive failures
+ * along with a ceiling at schedule_interval * MAX_INTERVALS_BACKOFF
+ * We also limit the additive backoff in case of consecutive failures as we don't
+ * want to pass in input that leads to out of range timestamps and don't want to
+ * put off the next start time for the job indefinitely
+ */
 static TimestampTz
 calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failures, BgwJob *job)
 {
 	float8 jitter = calculate_jitter_percent();
 	/* consecutive failures includes this failure */
-	float8 multiplier = 1 << (consecutive_failures - 1);
 	TimestampTz res;
 	volatile bool res_set = false;
 	TimestampTz last_finish = finish_time;
+	float8 multiplier = (consecutive_failures > MAX_FAILURES_MULTIPLIER ? MAX_FAILURES_MULTIPLIER :
+																		  consecutive_failures);
 	MemoryContext oldctx;
 	if (!IS_VALID_TIMESTAMP(finish_time))
 	{
@@ -207,7 +213,7 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
 	BeginInternalSubTransaction("next start on failure");
 	PG_TRY();
 	{
-		/* ival = retry_period * 2^(consecutive_failures - 1)  */
+		/* ival = retry_period * (consecutive_failures)  */
 		Datum ival = DirectFunctionCall2(interval_mul,
 										 IntervalPGetDatum(&job->fd.retry_period),
 										 Float8GetDatum(multiplier));
@@ -231,8 +237,13 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
 	}
 	PG_CATCH();
 	{
-		RollbackAndReleaseCurrentSubTransaction();
+		ErrorData *errdata = CopyErrorData();
+		ereport(LOG,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("calculate_next_start_on_failure ran into an error, resetting value"),
+				 errdetail("Error: %s", errdata->message)));
 		FlushErrorState();
+		RollbackAndReleaseCurrentSubTransaction();
 	}
 	PG_END_TRY();
 	MemoryContextSwitchTo(oldctx);
--- a/src/bgw/scheduler.c
+++ b/src/bgw/scheduler.c
@@ -173,6 +173,7 @@ worker_state_cleanup(ScheduledBgwJob *sjob)
 			elog(WARNING,
 				 "scheduler detected that job %d was deleted after job quit",
 				 sjob->job.fd.id);
+			ts_bgw_job_cache_invalidate_callback();
 			sjob->may_need_mark_end = false;
 			return;
 		}
@@ -188,6 +189,7 @@ worker_state_cleanup(ScheduledBgwJob *sjob)
 			 * a signal (cancel or terminate), it won't be able to so we
 			 * should.
 			 */
+			elog(LOG, "job %d failed", sjob->job.fd.id);
 			mark_job_as_ended(sjob, JOB_FAILURE);
 			/* reload updated value */
 			job_stat = ts_bgw_job_stat_find(sjob->job.fd.id);
@@ -226,12 +228,6 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
 
 			job_stat = ts_bgw_job_stat_find(sjob->job.fd.id);
 
-			if (!ts_bgw_job_stat_should_execute(job_stat, &sjob->job))
-			{
-				scheduled_bgw_job_transition_state_to(sjob, JOB_STATE_DISABLED);
-				return;
-			}
-
 			Assert(!sjob->reserved_worker);
 			sjob->next_start = ts_bgw_job_stat_next_start(job_stat, &sjob->job);
 			break;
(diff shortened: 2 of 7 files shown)


src/bgw/scheduler.c
```diff
@@ -247,6 +249,7 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
 				elog(WARNING,
 					 "scheduler detected that job %d was deleted when starting job",
 					 sjob->job.fd.id);
+				ts_bgw_job_cache_invalidate_callback();
```
erimatnor: Why aren't we using `ts_catalog_invalidate_cache()` so that this invalidation is propagated to all backends and processes? Also, it seems to me that this shouldn't be necessary since we already have a callback to invalidate and call this function whenever a DELETE is issued on the `bgw_job` table.

src/bgw/scheduler.c
```diff
@@ -247,6 +249,7 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
...
 				scheduled_bgw_job_transition_state_to(sjob, JOB_STATE_SCHEDULED);
```
erimatnor: I know this is not a change in this PR, but why do we recursively try to transition to `JOB_STATE_SCHEDULED` when we've already asserted above that the current state is already  `JOB_STATE_SCHEDULED`?

This seems to create an infinite loop if the same error happens again (I assume the job won't magically appear in the `bgw_job` table, so the same issue is likely to happen in the recursive call).

src/bgw/job_stat.c
```diff
@@ -231,8 +235,13 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
 	}
 	PG_CATCH();
 	{
-		RollbackAndReleaseCurrentSubTransaction();
+		ErrorData *errdata = CopyErrorData();
+		ereport(LOG,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("calculating next start on failure "),
```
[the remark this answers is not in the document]
gayyappan: @erimatnor This was feedback from the last PR about adding logging failures related to recalculation instead of ignoring them. 
This is just failure in recalculating start time and doesn't give a reason why the job failed.

src/bgw/scheduler.c
gayyappan: @erminator This invalidates the bgw job cache since we detected a potential missing job. (This ideally should not be necessary since we use the ts_catalog_invalidate_cache callback for deletes as you said. But we appear to be missing some cache invalidations and try to schedule jobs that have been deleted in some cases. This code is trying to be defensive and refresh the job cache when we detect a potential problem).

src/bgw/scheduler.c
gayyappan: This will go into scheduled state and set up the next_start time for the job. The bgw scheduler makes 1 pass over all the jobs and then updates its cache if needed (ts_bgw_...cache_invalidate...callback will set it as needed). So in the next pass , it will not find this job if it has indeed been deleted.

gayyappan: By the way, the  recursive loop is what we have been running into. so we missed the ts_catalog_cache_invalidate callback somehow.

src/bgw/job_stat.c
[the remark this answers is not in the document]
erimatnor: I understand why we want to print this error. I am just think we should make clear in the error what actually failed. Was it the original failure of the log or the failure to recompute. The message "calculating next start on failure" is ambiguous and doesn't explain what error occurred; it just gives the name of this function more or less (without the underscores).

src/bgw/scheduler.c
erimatnor: I guess I don't understand why we at all try again if the job was deleted. Don't we trust that this is the case the first time around?

FWIW, `ts_bgw_job_get_share_lock` seems to scan the jobs table so it _should_ be trustworthy w.r.t. whether this job exists or not so it doesn't make sense to me to try to schedule it again just to fail. The fact that the cache was not invalidated is another issue.

gayyappan: we are not trying again here. The bgw scheduler runs in a continuous loop (ts_bgw_scheduler_process). This happens during 1 pass of the scheduler loop. For the next pass of the loop,  the scheduler should update the list of cached jobs (and remove this job from its cache) and so we should not run into it the next time around.

src/bgw/scheduler.c
erimatnor: I am worried that we are just brushing the real problem under the rug by being defensive. 

I don't get the "potential missing job"... Either it is missing or it is not. There must be a source of truth and I don't understand why we are trying to reschedule it here if it is indeed missing. The fact that the cache was stale just seems to be a follow-on problem to not doing the right thing here.

`ts_bgw_job_get_share_lock` seems to scan the jobs table so either that code is broken and returns the wrong information. I guess that code actually conflates two things:

- Unable to grab lock
- Job missing

I suggest returning a status enum from `ts_bgw_job_get_share_lock` that tells us which case we are dealing with and only reschedule in case the job existed but we couldn't grab the lock.

src/bgw/scheduler.c
erimatnor: We are calling `scheduled_bgw_job_transition_state_to(sjob, JOB_STATE_SCHEDULED)` on a job that doesn't exist and is already in  `JOB_STATE_SCHEDULED`. Isn't that trying again? In any case, the action taken here is really strange.

This is what is happening here, if I am not mistaken:

1. This function was called with sjob in state `JOB_STATE_SCHEDULED` (asserted) and trying to transition to `JOB_STATE_STARTED`.
2. The job was not found (or lock could not be taken), so we got here
3. The sjob is still in state `JOB_STATE_SCHEDULED` since we haven't transitioned yet and now we are trying to transition to the same state we are in, i.e., `JOB_STATE_SCHEDULED` to `JOB_STATE_SCHEDULED`.
4. Presumably, the steps above will eventually lead to a transition to `JOB_STATE_STARTED` again, which means we are back to step (1). Or, alternatively, the job doesn't exist and will fail somehow when rescheduled, but we knew that already in step (2).

src/bgw/scheduler.c
```diff
@@ -241,13 +253,24 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
...
-				elog(WARNING,
-					 "scheduler detected that job %d was deleted when starting job",
-					 sjob->job.fd.id);
-				scheduled_bgw_job_transition_state_to(sjob, JOB_STATE_SCHEDULED);
+				if (res == JOB_NOT_FOUND)
+				{
+					elog(WARNING,
+						 "scheduler detected that job %d was deleted when starting job",
+						 sjob->job.fd.id);
+					ts_bgw_job_cache_invalidate_callback();
```
erimatnor: Since the pattern now is to invalidate the cache if the lock function returns `JOB_NOT_FOUND`, maybe the invalidation should move into the locking function where it returns that value?

src/bgw/scheduler.c
```diff
@@ -216,6 +227,7 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
 	switch (new_state)
 	{
 		case JOB_STATE_DISABLED:
+			/* we don't have a way of getting to this state now */
```
erimatnor: This comment is a bit confusing given the transition to this state on line 243. If this is indeed true, maybe the transition on line 243 should be removed.

src/bgw/scheduler.c
```diff
@@ -241,13 +253,24 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
 			Assert(!sjob->reserved_worker);
 
 			StartTransactionCommand();
+			BgwJobLockResult res = ts_bgw_job_get_share_lock(sjob->job.fd.id, CurrentMemoryContext);
```
erimatnor: mixed declaration and code. But I guess the previous team decided this was now allowed?

src/bgw/job_stat.c
```diff
@@ -19,6 +19,7 @@
 #include <utils/fmgrprotos.h>
 #endif
 #define MAX_INTERVALS_BACKOFF 5
+#define MAX_FAILURES_MULTIPLIER 20
```
cevian: I would prefer to have a MAX_INTERVAL instead. Then we could use interval_div(max, multiplier) and if that answer is less than the original interval, use the max_interval instead.

src/bgw/scheduler.c
cevian: I agree that rescheduling was not the right thing. Gayathri removed that. This code is blocking so `unable to grab lock` is not a possibility. Gayathri fixed what I think was the cache problem in a previous PR. I think it's still worth being defensive here as this is a huge problem for customers on cloud if it does occur as there is no way to restart the server to recover.

src/bgw/scheduler.c
```diff
@@ -311,9 +307,12 @@ on_failure_to_start_job(ScheduledBgwJob *sjob)
 {
 	StartTransactionCommand();
 	if (!ts_bgw_job_get_share_lock(sjob->job.fd.id, CurrentMemoryContext))
+	{
 		elog(WARNING,
 			 "scheduler detected that job %d was deleted while failing to start",
 			 sjob->job.fd.id);
+		ts_bgw_job_cache_invalidate_callback();
```
erimatnor: Nit: Still think it makes sense to move cache invalidation into `ts_bgw_job_get_share_lock()` to avoid this repetition (and risk forgetting to invalidate the cache in case of a missing job when this function is called in other places)

src/bgw/job_stat.c
gayyappan: @cevian the suggestion is not clear to me. Are you saying limit next start by MAX_INTERVAL instead? Could you elaborate?

src/bgw/scheduler.c
svenklemm: would that not happen when you have lock_timeout set to a non-zero value

cevian: I believe that that timeout throws an error if lock timeout is reached

PR #1725, opened 2020-03-06
Stop background worker in tests

svenklemm:
To make tests more stable and to remove some repeated code in the
tests this PR changes the test runner to stop background workers.
Individual tests that need background workers can still start them
and this PR will only stop background workers for the initial database
for the test, behaviour for additional databases created during the
tests will not change.

### DIFF
--- a/test/runner.sh
+++ b/test/runner.sh
@@ -53,7 +53,7 @@ cd ${EXE_DIR}/sql
 
 # create database and install timescaledb
 ${PSQL} $@ -U $TEST_ROLE_SUPERUSER -d postgres -v ECHO=none -c "CREATE DATABASE \"${TEST_DBNAME}\";"
-${PSQL} $@ -U $TEST_ROLE_SUPERUSER -d ${TEST_DBNAME} -v ECHO=none -c "set client_min_messages=error; CREATE EXTENSION timescaledb;"
+${PSQL} $@ -U $TEST_ROLE_SUPERUSER -d ${TEST_DBNAME} -v ECHO=none -c "SET client_min_messages=error; CREATE EXTENSION timescaledb;"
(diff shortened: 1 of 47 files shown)


test/runner.sh
```diff
@@ -53,7 +53,8 @@ cd ${EXE_DIR}/sql
 
 # create database and install timescaledb
 ${PSQL} $@ -U $TEST_ROLE_SUPERUSER -d postgres -v ECHO=none -c "CREATE DATABASE \"${TEST_DBNAME}\";"
-${PSQL} $@ -U $TEST_ROLE_SUPERUSER -d ${TEST_DBNAME} -v ECHO=none -c "set client_min_messages=error; CREATE EXTENSION timescaledb;"
+${PSQL} $@ -U $TEST_ROLE_SUPERUSER -d ${TEST_DBNAME} -v ECHO=none -c "SET client_min_messages=error; CREATE EXTENSION timescaledb;"
+${PSQL} $@ -U $TEST_ROLE_SUPERUSER -d ${TEST_DBNAME} -v ECHO=none -c "SELECT _timescaledb_internal.stop_background_workers();" > /dev/null
```
erimatnor: Can we just add this to `testsupport.sql` instead, which gets included below? Might even make sense for line above that creates the extension.

k-rus:
@svenklemm I noticed that number of tests are failing. I guess some fixes are still needed.

test/runner.sh
svenklemm: i decided not to do this in case there were other users of testsupport.sql but since test runner is only user i moved it there

svenklemm:
> LGTM
> I am wonder if the cleanup from `bgw_job`s should be done by default? I noticed in some cases, but not all:
> 
> ```sql
> DELETE FROM _timescaledb_config.bgw_job WHERE TRUE;
>  TRUNCATE _timescaledb_internal.bgw_job_stat;
> ```

I dont think this should be done for all tests and only bgw specific tests interact with this

PR #1737, opened 2020-03-09
Fix flaky gapfill test

svenklemm:
The gapfill test assumed 1 day is always <= 24 hours which is not
true during DST switch leading to a failing test when run in that
time. This PR fixes the test to have reproducable output even
when run during DST switch.

### DIFF
--- a/tsl/test/expected/gapfill.out
+++ b/tsl/test/expected/gapfill.out
@@ -2026,10 +2026,10 @@ GROUP BY 1;
 
 -- time_bucket_gapfill with now() as start
 SELECT
- time_bucket_gapfill('1d'::interval,t)
+ time_bucket_gapfill('1h'::interval,t)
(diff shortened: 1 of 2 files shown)


tsl/test/expected/gapfill.out
```diff
@@ -2029,7 +2029,7 @@ SELECT
  time_bucket_gapfill('1d'::interval,t)
 FROM (VALUES (now()),(now())) v(t)
 WHERE
- t >= now() AND t < now() - '1d'::interval
+ t >= now() AND t < now() - '2d'::interval
```
erimatnor: Nit: isn't it safer to just use an interval which is not affected by timezone stuff? Like `24h`?

svenklemm: ok switched it all to 1h

PR #1740, opened 2020-03-09
Fix copy of continuous agg invalidation entries

gayyappan:
When we copy the invalidation logs for individual continuous
aggregates, the lowest-value was globally overwritten. Fix this
so that the change is specific to a continuous aggregate.
This bug could result in missing invalidations.


svenklemm:
Dont forget changelog entry

PR #1743, opened 2020-03-11
Fix continuous aggregate materialization timezone handling

### LINKED ISSUE #1695: refresh_lag timezone problem
 - OS: cent os 11
 - PostgreSQL version : 11.2
 - TimescaleDB version : 1.6
 - Installation method: yum

**Describe the bug**
I have a continuous aggregate with refresh_lag = - 1 hour and bucket size = 1 hour and runs every 10 minutes. In version 1.5 it had no problem. Every 10 minutes data become updated. However after upgrading from 1.5 to 1.6, data always updated to be 3 hours before the current time. Timezone is +3 gmt and i tried changing refresh_lag to be -4 hour and my problem was solved. 

**To Reproduce**
1. Set timezone to +3 gmt
2. Create view:
CREATE VIEW anonymous_log.service_usage_info_hourly_summary( date, application_type, province, service_code, usage_count )
WITH ( timescaledb.continuous,
    timescaledb.refresh_lag = '-1h',
  	timescaledb.max_interval_per_job = '1 week',
  	timescaledb.refresh_interval = '10m',
  	timescaledb.ignore_invalidation_older_than = '60 day'
    )
AS
SELECT time_bucket('1 hour', log_timestamp) as date, application_type, province, service_code, count(*)
FROM anonymous_log.service_usage_info
GROUP BY date, application_type, province, service_code;

3. Check data after 1 hour nothing happens, after 2 hour nothing happens. Check data after 3 - 4 hour. You will see table contains data from 3 hour earlier.

Note: I also tried with refresh_lag = '-1d' same problem occurs. refresh_lag with -1m and so can be given for testing purposes.

**Expected behavior**
Every 10 minutes data to become live.

**Actual behavior**
Table contains data from 3 hours earlier.

svenklemm:
For hypertables with a TIMESTAMP column materialization would not
honor the local timezone when determining the range for materialization
and instead treat times as UTC.

Fixes #1695

### DIFF
--- a/src/interval.c
+++ b/src/interval.c
@@ -374,6 +374,14 @@ ts_get_now_internal(Dimension *open_dim)
 #else
 		Datum now_datum = TimestampTzGetDatum(GetCurrentTimestamp());
 #endif
+
+		/*
+		 * If the type of the partitioning column is TIMESTAMP or DATE
+		 * we need to adjust the return value for the local timezone.
+		 */
+		if (dim_post_part_type == TIMESTAMPOID || dim_post_part_type == DATEOID)
+			now_datum = DirectFunctionCall1(timestamptz_timestamp, now_datum);
+
 		return ts_time_value_to_internal(now_datum, TIMESTAMPTZOID);
 	}
 }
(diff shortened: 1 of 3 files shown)


src/interval.c
```diff
@@ -374,6 +374,19 @@ ts_get_now_internal(Dimension *open_dim)
 #else
 		Datum now_datum = TimestampTzGetDatum(GetCurrentTimestamp());
 #endif
+
```
gayyappan: could you explain why this is a breaking change if the fix is global? We we could defer the fix until we can do this correctly. Any reason not to do that?

svenklemm: It's planned to fix this globally for 2.0, but changing the function invalidates already persisted data.

PR #1747, opened 2020-03-12
Fix returning error on chunk get and exporting

k-rus:
Fixes to return error in ts_chunk_get_by_relid if fail_if_not_found is
true. Removes TSDLLEXPORT macro from definitions in few C files, which
helps VS Code to work properly and find function definitions.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -442,7 +442,7 @@ chunk_collision_resolve(Hyperspace *hs, Hypercube *cube, Point *p)
 	chunk_scan_ctx_destroy(&scanctx);
 }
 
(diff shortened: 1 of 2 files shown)


k-rus:
> TSDLLEXPORT is required for windows builds (see export.h). Not clear why you are removing them here.

@gayyappan They are present in header files. I am removing the duplication of them in C files. Currently, the state is inconsistent between different exported functions, since some of them have macro only in header file and others have in both header and c files.
Windows build succeeded. So the PR removes unnecessary headers, which make hard to use VS Code.

PR #1753, opened 2020-03-14
Fix size utility functions for compressed hypertables

### LINKED ISSUE #1741: Compression breaks hypertable_relation_size, and hypertable_approximate_row_count
On timescaledb 1.6, when one uses compression the function hypertable_relation_size, hypertable_relation_size_pretty, and hypertable_approximate_row_count do not report any information on the compressed chunks.
Meaning if all the chunks are compressed the row count will be 0 and the table size 32k.

If not all the chunks are compressed we have the information on the not compressed chunks.

Information on the compressed chunks is available through timescaledb_information.compressed_chunk_stats, but it would be nice to still have it through the functions.
I think there was a request as well (#1683 ) to have the size in bytes rather than as pretty string, so it would fix this problem as well.

svenklemm:
This patch adjusts hypertable_approximate_row_count to return
correct row estimates for hypertables that include compressed
chunks. It also fixes a bug that hypertables without chunks
would not be included in the result and fixes the schema handling
for that function.

This patch also adjusts hypertable_relation_size and
hypertable_relation_size_pretty to include information about
compressed chunks.

Fixes #1741

### DIFF
--- a/sql/size_utils.sql
+++ b/sql/size_utils.sql
@@ -46,22 +46,37 @@ BEGIN
                FROM (
                SELECT *, total_bytes-index_bytes-COALESCE(toast_bytes,0) AS table_bytes FROM (
                       SELECT
-                      sum(pg_total_relation_size(format('%%I.%%I', c.schema_name, c.table_name)))::bigint as total_bytes,
-                      sum(pg_indexes_size(format('%%I.%%I', c.schema_name, c.table_name)))::bigint AS index_bytes,
-                      sum(pg_total_relation_size(reltoastrelid))::bigint AS toast_bytes
+                      sum(
+                        pg_total_relation_size(format('%%I.%%I', c.schema_name, c.table_name)) +
+                        CASE WHEN c_comp.id IS NOT NULL
+                          THEN pg_total_relation_size(format('%%I.%%I',c_comp.schema_name, c_comp.table_name))
+                          ELSE 0
+                        END
+                      )::bigint as total_bytes,
+                      sum(
+                        pg_indexes_size(format('%%I.%%I', c.schema_name, c.table_name)) +
+                        CASE WHEN c_comp.id IS NOT NULL
+                          THEN pg_indexes_size(format('%%I.%%I', c_comp.schema_name, c_comp.table_name))
+                          ELSE 0
+                        END
+                      )::bigint AS index_bytes,
+                      sum(
+                        pg_total_relation_size(reltoastrelid) +
+                        CASE WHEN c_comp.id IS NOT NULL
+                          THEN pg_total_relation_size(reltoastrelid)
+                          ELSE 0
+                        END
+                      )::bigint AS toast_bytes
                       FROM
-                      _timescaledb_catalog.hypertable h,
-                      _timescaledb_catalog.chunk c,
-                      pg_class pgc,
-                      pg_namespace pns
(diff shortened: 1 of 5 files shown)


sql/size_utils.sql
```diff
@@ -453,20 +483,24 @@ BEGIN
...
--- Thanks to @fvannee on Github for providing the initial draft
--- of this query
         RETURN QUERY
         SELECT h.schema_name,
             h.table_name,
-            row_estimate.row_estimate
+            coalesce((
+              SELECT
+                sum(
+                  CASE WHEN c.compressed_chunk_id IS NULL
```
gayyappan: also need to check for dropped chunk case?

tsl/test/expected/size_utils.out
```diff
@@ -0,0 +1,71 @@
...
+-- row count should be 32
+SELECT * FROM hypertable_approximate_row_count('compressed');
+ schema_name | table_name | row_estimate 
+-------------+------------+--------------
+ public      | compressed |           32
+(1 row)
+
+SELECT * FROM hypertable_relation_size('compressed');
+ table_bytes | index_bytes | toast_bytes | total_bytes 
+-------------+-------------+-------------+-------------
```
gayyappan: why is the toast bytes null after compression? Oh I see, you don't have any data in the table. So unexpected toast size.

tsl/test/sql/size_utils.sql
```diff
@@ -0,0 +1,36 @@
...
+INSERT INTO compressed SELECT generate_series('2000-01-01'::timestamptz, '2000-02-01'::timestamptz, '1d'::interval);
+
+VACUUM compressed;
+
+-- row count should be 32
+SELECT * FROM hypertable_approximate_row_count('compressed');
+
+SELECT * FROM hypertable_relation_size('compressed');
+SELECT * FROM hypertable_relation_size_pretty('compressed');
+
```
gayyappan: please add a query to verify we have 5 chunks here.

tsl/test/expected/size_utils.out
```diff
@@ -0,0 +1,71 @@
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+CREATE TABLE compressed(time timestamptz NOT NULL);
```
gayyappan: could you name the table as compressed_tbl or something similar to indicate it is a table.

sql/size_utils.sql
```diff
@@ -46,22 +46,37 @@ BEGIN
                FROM (
                SELECT *, total_bytes-index_bytes-COALESCE(toast_bytes,0) AS table_bytes FROM (
                       SELECT
-                      sum(pg_total_relation_size(format('%%I.%%I', c.schema_name, c.table_name)))::bigint as total_bytes,
-                      sum(pg_indexes_size(format('%%I.%%I', c.schema_name, c.table_name)))::bigint AS index_bytes,
-                      sum(pg_total_relation_size(reltoastrelid))::bigint AS toast_bytes
+                      sum(
```
gayyappan: I think this was intended to kept as is and not modified to expose compressed table information.  All compression related information is exposed only via the compression specific APIs. (Perhaps the compression usage related docs has some discussion related to it). We need a broader discussion on what these APIs should expose before merging this.

PR #1761, opened 2020-03-18
Modify isolation test makefile rules

gayyappan:
PG12 needs version specific output for some
isolation tests. Modify the makefile to make it
consistent with use of TEST_TEMPLATES in other
regression suites and add version specific output
files.


gayyappan:
> You're missing the PG12 golden files here.

This is on master - no PG12 support.

tsl/test/isolation/specs/CMakeLists.txt
```diff
@@ -1,14 +1,15 @@
...
 
 set(TEST_TEMPLATES
- reorder_deadlock.spec.in
- reorder_vs_insert_other_chunk.spec.in
+  continuous_aggs_insert.spec.in
+  continuous_aggs_multi.spec.in
 )
 
-set(TEST_TEMPLATES_DEBUG
+set(TEST_FILES_NEEDGEN_DEBUG
```
svenklemm: those are templates so the name should reflect that

gayyappan: These are not the same as TEST_TEMPLATES used for generating version specific .sql files. They have to be processed differently so that TS_MODULE_NAME is replaced by the correct path. We are not generating version specific files here.

erimatnor: I am also a bit confused by this.

It sounds like the files here are actually more like templates since there's actually a replacement going in the files, as opposed to the version-specific files that only require a file name change.

gayyappan: The general practice for our test files is to use TEMPLATE_FILES for version specific sql files. I am using TEMPLATE_FILES in this Makefile for the same purpose i.e. generate version specific .spec files. But the other files also need preprocessing (different from TEMPLATE_FILES) and cannot be added under TEST_FILES. You will find the Makefile has an additional block that does the preprocessing for TEST_FILES_NEEDGEN

erimatnor: I understand that need to generate files in different ways. I am just confused by the variable names because they don't seem accurate. FWIW, I don't think there's a general practice for how to do this, only historical reasons that may or may not be applicable anymore as the usage has grown. The need for version-specific tests was the initial reason for generating output files and we happened to call it TEMPLATE_FILES. 

Still, this is not a blocker and can be cleaned up later.

PR #1771, opened 2020-03-24
Run regression test on latest PG versions

k-rus:
Docker images are build with latest versions of PostgreSQL, thus
updating regression tests to run on the latest PG versions. Fixing 
authentication for PG docker images, since it is required after a 
recent change in docker-library/postgres@42ce743.

Running isolation tests on new versions of PG produces additional 
output:
- Notifications are not missed to print postgres/postgres@ebd4992
- Notifications are prefixed with session name 
postgres/postgres@a28e10e
- Timeout cancellation is printed in isolation tests 
postgres/postgres@b578404

The expected outputs are modified to succeed on latest PG versions, 
while the affected isolation tests are disabled for earlier versions 
of PG.


k-rus:
> Could you explain the change in behavior - why are ROLLBACKS replaced by COMMITS ?

@gayyappan Otherwise `compression_ddl` fails with timeout one of the session instead of to succeed. And it happens on all new PG versions. Actually PG isolation tests are using COMMIT to unlock, bot ROLLBACK. So I guess it is some implicit limitation.

k-rus:
New printout in the isolation test `compression_ddl` is due to https://github.com/postgres/postgres/commit/99351a8b5a3882ee5d131bd37d03a67ab07b4fea

k-rus:
Printing session name in notifications were added in https://github.com/postgres/postgres/commit/a28e10e82e540d45e448bc404af76df8a54008f1
which is available in all 12 versions.

It was cherry picked into previous PG version, e.g., through https://github.com/postgres/postgres/commit/b5784042ae7a8c15dc44a7d0112b575dfcc101dc
Covered versions: 11.7, 10.12, 9.6.17, 11.6, 10.11, 9.6.16

tsl/test/isolation/specs/CMakeLists.txt
```diff
@@ -4,22 +4,23 @@ set(TEST_TEMPLATES_MODULE
...
-  continuous_aggs_multi.spec.in
-)
-
 set(TEST_TEMPLATES_MODULE_DEBUG
   reorder_vs_insert.spec.in
   reorder_vs_select.spec.in
 )
 
+list(APPEND TEST_FILES
+  continuous_aggs_insert.spec
```
erimatnor: shouldn't we exclude isolation tests here on PG versions that do not support the new output?

k-rus: Yes, this is what I started to do already, but haven't finished and pushed. I will continue with this path. Thank you!

k-rus:
I re-requested reviews from everyone, since the PR has been changed significantly due to changes in new versions of PG, which broke isolation test outputs. Furthermore, all PG versions were changes to latest.

tsl/test/isolation/specs/compression_ddl.spec
```diff
@@ -29,7 +29,7 @@ step "Sc" {COMMIT;}
 
 session "LC"
 step "LockChunk1"   { BEGIN; select lock_chunktable (q.chname) from (select ch.schema_name || '.' || ch.table_name as chname from _timescaledb_catalog.hypertable ht, _timescaledb_catalog.chunk ch where ch.hypertable_id = ht.id and ht.table_name like 'ts_device_table' order by ch.id limit 1 ) q; }
-step "UnlockChunk" {ROLLBACK;}
+step "UnlockChunk" { ROLLBACK;}
```
gayyappan: looks like this change is not required.

k-rus: Since I will need to rebase, I will remove this change.

k-rus:
> Hmm I dont like the idea of minor version specific tests. I would prefer if those tests would use IGNORES environment variable so the tests are still run but dont make the test suite fail.

@svenklemm Why is it going to be useful to run and ignore? The test results will be ignored by majority, so the resources of running the test will be wasted. Furthermore, running all tests locally will fail, which will bring confusion and large time loose. So can you bring the motivation to your point?

PR #1787, opened 2020-03-31
Adds compressed_chunk_name to compressed_chunk_stats view

tylerfontaine:
This allows someone to much more easily see the related compressed chunk, without having to make a separate query to find it.

This change DOES result in uncompressed chunks no longer appearing in the view, which I don't think is a great loss. This is specifically stats for compressed chunks, so displaying uncompressed chunks here seems weird to me anyway. BUT I wanted to note it, because it does represent a change in behavior, in case that influences how/when to release it.

This also requires a docs PR to update the view schema.

I think we could also remove the CASE clause, since the Join I added makes it so that there are no uncompressed views, but I wasn't totally sure.

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -192,6 +192,7 @@ WITH mapq as
  FROM _timescaledb_catalog.compression_chunk_size map )
   SELECT format('%1$I.%2$I', srcht.schema_name, srcht.table_name)::regclass as hypertable_name,
   format('%1$I.%2$I', srcch.schema_name, srcch.table_name)::regclass as chunk_name,
+  format('%1$I.%2$I', srcch.schema_name, srccch.table_name)::regclass as compressed_chunk_name,
   CASE WHEN srcch.compressed_chunk_id IS NULL THEN 'Uncompressed'::TEXT ELSE 'Compressed'::TEXT END as compression_status,
   mapq.uncompressed_heap_bytes,
   mapq.uncompressed_index_bytes,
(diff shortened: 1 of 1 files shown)


gayyappan:
@tylerfontaine  We are reviewing the views and plan to modify them consistently. Compression internals are not exposed via the views.

tylerfontaine:
> @tylerfontaine We are reviewing the views and plan to modify them consistently. Compression internals are not exposed via the views.

I understand that they are not currently, but being unable to easily find these things is a usability issue. If we're going to start routing things appropriately if, say, an ANALYZE is run on the hypertable chunk (rather than the compressed hypertable chunk) then great, but until then, we need better visibility.

k-rus:
@tylerfontaine 
I support @gayyappan that compression chunks should not be exposed through the information views. If we want to do this, I believe it will be necessary to block user operations on compressed chunks, since otherwise users can break the database. However, I am not sure it is possible to block without destroying current functionality.
I think Postgres doesn't allow to replace table names for query plans. I am not sure how it is feasible.
So it feels that it is good to create an issue, since number of investigations is needed.

Can you provide context why it is emergent? Was there some support issue? I haven't noticed any related issues in public slack support channel.

Regarding the PR:
1. The failed builds show that number of tests fail as the view is different now.
2. The commit message doesn't comply with [the requirements](https://chris.beams.io/posts/git-commit/#seven-rules) and you should get a warning message when you committed it in your command line. 

This is just comments for future, since it is good to decide how to go with the proposal.

tylerfontaine:
Thanks for the comments! Sure - I can understand the trepidation in exposing this information.
No warnings on commit, BUT thanks for the link. I'll be sure to do that better. :)

This came from troubleshooting an issue, and for diagnostic purposes, needing to run an ANALYZE on a compressed vs uncompressed chunk. Since the compressed chunks are separate chunks altogether, this was confusing for the customer, and required a separate query to determine the correlative compressed chunk. This felt like a bug to me, but if there are compelling reasons not to make it easy to get to, great, but we should probably also make those things easier to do in that case. :)

In that case, we can close this PR, and move it to an issue, since there's some discussion about how we might improve experience here.

erimatnor:
@tylerfontaine thanks for bringing this usability issue to our attention. FWIW, we are aware of usability issues related to compression. As pointed out by others, we are reviewing how we deal with this post 1.7 and I think we'd like to get the basic abstraction of compressed chunks right rather than exposing more of the internals (which might actually further confuse people). 

Let's take your feedback into consideration for post 1.7 when we take a look at these issues.

erimatnor:
@tylerfontaine I am closing this since we will revisit the informational views for 2.0. Could reopen then if it makes sense.

PR #1795, opened 2020-04-01
Fix use-after-free in cagg union view creation

svenklemm:


.travis.yml
```diff
@@ -404,7 +404,7 @@ jobs:
         - bash ./scripts/docker-run-memory-test.sh
 
     # Sanitizer tests
-    - if: (type = cron) OR (branch = prerelease_test)
+    - if: (type = pull_request) OR (branch = prerelease_test)
```
gayyappan, line 407: I thought we are running this only for cron jobs. Is this temporary?

svenklemm: it was only for testing

PR #1804, opened 2020-04-08
Make compiler warnings into errors

erimatnor:
This change makes compiler warnings into errors by default on all
builds. This setting used to be applied on Travis builds, but was
removed (assumingly by accident) in a previous commit. The new
behavior can be disabled by including `-DWARNINGS_AS_ERRORS=OFF`
when running cmake.


appveyor.yml
```diff
@@ -1,7 +1,7 @@
 version: 1.0.{build}
 image: Windows Server 2019
 services:
-  - postgresql101
+  - postgresql111
```
k-rus, line 4: This is incorrect to my knowledge. Should be:
```suggestion
  - postgresql11

```

k-rus: But rebasing from the latest will bring pg 12, so it will not be an issue.

PR #1808, opened 2020-04-10
Fix ts_hypertable_get_all for compressed tables

### LINKED ISSUE #1535: show_chunks() and drop_chunks() throws a compressed hypertable error when no tables are specified.
 - OS: Ubuntu 18.04
 - PostgreSQL version (output of `postgres --version`): 11.4
 - TimescaleDB version (output of `\dx` in `psql`):1.5.1

I enabled compression on one of the hypertables but now I'm unable to see how many chunks I have even though I didn't compress any chuncks yet. This is how I created the table:

```
CREATE TABLE rows_store
(
    row_cnt bigint NOT NULL,
    kernel_id int,
    pass_nr int,
    acq_id bigint,
    ch_meta json,
    ch_spectra real[],
    ch_image_path text,
    time_stamp bigint,
    UNIQUE(row_cnt)
);

CREATE INDEX idx_rows_store_acq_id 
ON rows_store(acq_id);

SELECT create_hypertable('rows_store', 'row_cnt', chunk_time_interval => 1000000);

ALTER TABLE rows_store SET (
  timescaledb.compress, 
  timescaledb.compress_segmentby = 'acq_id'
);
```

And here is the error:
```
# SELECT show_chunks(newer_than => 0);
ERROR:  cannot call chunk_get_chunks_in_time_range on a compressed hypertable
# SELECT drop_chunks(newer_than =>0);
ERROR:  cannot call chunk_get_chunks_in_time_range on a compressed hypertable
```

Is this the entended behaviour? In your [documentation](https://docs.timescale.com/latest/using-timescaledb/compression#advanced-usage) it looks like you can call `SELECT show_chunks('conditions', older_than => interval '3 days');` and then compress without any other quiries or errors.

Olernov:
When calling show_chunks or drop_chunks without specifying
a particular hypertable TimescaleDB iterates through all
existing hypertables and builds a list. While doing this
it adds the internal '_compressed_hypertable_*' tables
which leads to incorrect behaviour of
ts_chunk_get_chunks_in_time_range function. This fix
filters out the internal compressed tables while scanning
at ts_hypertable_get_all function.

Fixes #1535

### DIFF
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -338,10 +338,26 @@ chunk_store_entry_free(void *cse)
 	MemoryContextDelete(((ChunkStoreEntry *) cse)->mcxt);
 }
 
+static bool
+hypertable_is_compressed_or_materialization(Hypertable *ht)
+{
+	ContinuousAggHypertableStatus status = ts_continuous_agg_hypertable_status(ht->fd.id);
+	return (ht->fd.compressed || status == HypertableIsMaterialization);
+}
+
+static ScanFilterResult
+hypertable_filter_exclude_compressed_and_materialization(TupleInfo *ti, void *data)
+{
+	Hypertable *ht = ts_hypertable_from_tupleinfo(ti);
+
+	return hypertable_is_compressed_or_materialization(ht) ? SCAN_EXCLUDE : SCAN_INCLUDE;
+}
+
 static int
 hypertable_scan_limit_internal(ScanKeyData *scankey, int num_scankeys, int indexid,
 							   tuple_found_func on_tuple_found, void *scandata, int limit,
-							   LOCKMODE lock, bool tuplock, MemoryContext mctx)
+							   LOCKMODE lock, bool tuplock, MemoryContext mctx,
+							   tuple_filter_func filter)
 {
 	Catalog *catalog = ts_catalog_get();
 	ScannerCtx	scanctx = {
@@ -353,6 +369,7 @@ hypertable_scan_limit_internal(ScanKeyData *scankey, int num_scankeys, int index
 		.limit = limit,
 		.tuple_found = on_tuple_found,
 		.lockmode = lock,
+        .filter = filter,
 		.scandirection = ForwardScanDirection,
 		.result_mctx = mctx,
 		.tuplock = {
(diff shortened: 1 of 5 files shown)


erimatnor:
Thanks @Olernov, this looks great and is very timely. We're about to look into multiple similar issues for our next release and we'll certainly consider this change as part of that work.

src/hypertable.c
```diff
@@ -337,10 +337,32 @@ chunk_store_entry_free(void *cse)
...
+	else
+		return SCAN_INCLUDE;
+}
+
 static int
 hypertable_scan_limit_internal(ScanKeyData *scankey, int num_scankeys, int indexid,
 							   tuple_found_func on_tuple_found, void *scandata, int limit,
-							   LOCKMODE lock, bool tuplock, MemoryContext mctx)
+							   LOCKMODE lock, bool tuplock, MemoryContext mctx,
+							   ScanFilterResult (*filter)(TupleInfo *ti, void *data))
```
erimatnor, line 365:
```suggestion
							   tuple_filter_func filter)
```

test/expected/chunk_utils_compression.out
```diff
@@ -0,0 +1,138 @@
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+\c :TEST_DBNAME :ROLE_SUPERUSER
```
erimatnor, line 4: We recommend running tests as a low-privileged user and only switch to super-user when necessary.

src/hypertable.c
```diff
@@ -337,10 +337,32 @@ chunk_store_entry_free(void *cse)
...
 }
 
+static bool
+table_is_compressed_or_materialization(Hypertable *ht)
+{
+	ContinuousAggHypertableStatus status = ts_continuous_agg_hypertable_status(ht->fd.id);
+	if (ht->fd.compressed || status == HypertableIsMaterialization)
+		return true;
+	else
+		return false;
```
erimatnor, line 347:
```suggestion
	return (ht->fd.compressed || status == HypertableIsMaterialization);
```

src/hypertable.c
```diff
@@ -337,10 +337,32 @@ chunk_store_entry_free(void *cse)
...
+}
+
+static ScanFilterResult
+filter_exclude_compressed_and_materialization(TupleInfo *ti, void *data)
+{
+	Hypertable *ht = ts_hypertable_from_tupleinfo(ti);
+
+	if (table_is_compressed_or_materialization(ht))
+		return SCAN_EXCLUDE;
+	else
```
pmwkaa, line 357: Just nit, I would suggest just `return SCAN_INCLUDE` at the end of function and omit this else statement. Sometimes different compiler versions warns about missing return at the end of a function

PR #1812, opened 2020-04-13
Record cagg view state in catalog

svenklemm:
Record materialized_only state of continuous aggregate view in
catalog and show state in timescaledb_information.continuous_aggregates.

### DIFF
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -252,6 +252,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_agg (
     direct_view_name NAME NOT NULL,
     max_interval_per_job BIGINT NOT NULL,
     ignore_invalidation_older_than BIGINT NOT NULL DEFAULT BIGINT '9223372036854775807',
+    materialized_only BOOL NOT NULL DEFAULT false,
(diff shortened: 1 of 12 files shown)


sql/pre_install/tables.sql
```diff
@@ -252,6 +252,7 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_agg (
     direct_view_name NAME NOT NULL,
     max_interval_per_job BIGINT NOT NULL,
     ignore_invalidation_older_than BIGINT NOT NULL DEFAULT BIGINT '9223372036854775807',
+    materialized_only BOOL NOT NULL DEFAULT true,
```
gayyappan, line 255: won't this cause upgrade failures?

svenklemm: the default in the catalog has to be true because all caggs created before the flag/feature got added are materialized only view

gayyappan: ok.

svenklemm: actually it doesnt have to be true but then the upgrade script needs to set it to true for all existing entries, maybe i should change it to that then the default would match the default behaviour when creating new cagg

PR #1830, opened 2020-04-17
Add 1.7.0 and PG12 to update test scripts

svenklemm:


svenklemm:
> Looks good, but don't we have any post-upgrade tests that check behavior after an upgrade?

This is the configuration for those tests, it diffs upgraded versions against newly installed version

PR #1831, opened 2020-04-17
Improve cmake PostgreSQL version check

pmwkaa:
This change improves cmake check for PostgreSQL minor/patch versions
and makes it consistent with src/compat.h version check

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,9 +200,14 @@ endif ()
 
 message(STATUS "Compiling against PostgreSQL version ${PG_VERSION}")
 
-if ((${PG_VERSION} VERSION_LESS "9.6") OR (${PG_VERSION_MAJOR} GREATER "12"))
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -200,9 +200,13 @@ endif ()
 
 message(STATUS "Compiling against PostgreSQL version ${PG_VERSION}")
 
-if ((${PG_VERSION} VERSION_LESS "9.6") OR (${PG_VERSION_MAJOR} GREATER "12"))
-  message(FATAL_ERROR "TimescaleDB only supports PostgreSQL 9.6, 10, 11 or 12")
-endif ()
+# Ensure that PostgreSQL version is supported and consistent
+# with src/compat.h version check
+if ((${PG_VERSION_MAJOR} EQUAL "9" AND ${PG_VERSION} VERSION_LESS "9.6.3") OR
```
svenklemm, line 205: should add a check for less then 9

PR #1843, opened 2020-04-22
Improve handling of "dropped" chunks 

erimatnor:
The internal chunk API is updated to avoid returning `Chunk` objects
that are marked `dropped=true` along with some refactoring, hardening,
and cleanup of the internal chunk APIs. In particular, chunks could
previously be returned in a partial state (without all fields set,
partial constraints, etc.) and this is no longer allowed. Further,
lock handling was unclear when joining chunk metadata from different
catalog tables. This is more clear now by having chunks built within
the scan loop so that proper locks are held when reading other
metadata (such as constraints).

This change also fixes issues with dropped chunks that caused chunk
metadata to be processed many times instead of just once, leading to
potential bugs or bad performance.

In particular, since the introduction of the “dropped” flag, chunk
metadata can exist in two states: 1. `dropped=false`
2. `dropped=true`. When dropping chunks (e.g., via `drop_chunks`,
`DROP TABLE <chunk>`, or `DROP TABLE <hypertable>`) there are also two
modes of dropping: 1. DELETE row and 2. UPDATE row and SET
dropped=true.

The deletion mode and the current state of chunk lead to a
cross-product resulting in 4 cases when dropping/deleting a chunk:

1. DELETE row when dropped=false
2. DELETE row when dropped=true
3. UPDATE row when dropped=false
4. UPDATE row when dropped=true

Unfortunately, the code didn't distinguish between these cases. In
particular, case (4) should not be able to happen, but since it did it
lead to a recursing loop where an UPDATE created a new tuple that then
is recursed to in the same loop, and so on.

To fix this recursing loop and make the code for dropping chunks less
error prone, a number of assertions have been added, including some
new light-weight scan functions to access chunk information without
building a full-blown chunk.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -79,6 +79,31 @@ typedef enum ChunkResult
 	CHUNK_PROCESSED
 } ChunkResult;
 
+/*
+ * Context for scanning and building a chunk from a stub.
+ *
+ * If found, the chunk will be created and the chunk pointer member is set in
+ * the result. Optionally, a caller can pre-allocate the chunk member's memory,
+ * which is useful if one, e.g., wants to fill in an memory-aligned array of
+ * chunks.
+ *
+ * If the chunk is a tombstone (dropped flag set), then the Chunk will not be
+ * created and instead is_dropped will be TRUE.
+ */
+typedef struct ChunkStubScanCtx
+{
+	ChunkStub *stub;
+	Chunk *chunk;
+	bool is_dropped;
+} ChunkStubScanCtx;
+
+static bool
+chunk_stub_is_valid(const ChunkStub *stub, unsigned int expected_slices)
+{
+	return stub && stub->id > 0 && stub->constraints && expected_slices == stub->cube->num_slices &&
+		   stub->cube->num_slices == stub->constraints->num_dimension_constraints;
+}
+
 typedef ChunkResult (*on_chunk_stub_func)(ChunkScanCtx *ctx, ChunkStub *stub);
 static void chunk_scan_ctx_init(ChunkScanCtx *ctx, Hyperspace *hs, Point *p);
 static void chunk_scan_ctx_destroy(ChunkScanCtx *ctx);
@@ -87,6 +112,19 @@ static int chunk_scan_ctx_foreach_chunk_stub(ChunkScanCtx *ctx, on_chunk_stub_fu
 											 uint16 limit);
 static Datum chunks_return_srf(FunctionCallInfo fcinfo);
 static int chunk_cmp(const void *ch1, const void *ch2);
+static Chunk *chunk_find(Hypertable *ht, Point *p, bool resurrect);
+static void init_scan_by_qualified_table_name(ScanIterator *iterator, const char *schema_name,
+											  const char *table_name);
+
+#define IS_VALID_CHUNK(chunk)                                                                      \
+	((chunk) && !(chunk)->fd.dropped && (chunk)->fd.id > 0 && (chunk)->fd.hypertable_id > 0 &&     \
+	 OidIsValid((chunk)->table_id) && OidIsValid((chunk)->hypertable_relid) &&                     \
+	 (chunk)->constraints && (chunk)->cube &&                                                      \
+	 (chunk)->cube->num_slices == (chunk)->constraints->num_dimension_constraints)
+
+#define ASSERT_IS_VALID_CHUNK(chunk) Assert(IS_VALID_CHUNK(chunk))
+
+#define ASSERT_IS_NULL_OR_VALID_CHUNK(chunk) Assert(chunk == NULL || IS_VALID_CHUNK(chunk))
 
 static HeapTuple
 chunk_formdata_make_tuple(const FormData_chunk *fd, TupleDesc desc)
@@ -170,16 +208,6 @@ ts_chunk_insert_lock(Chunk *chunk, LOCKMODE lock)
 	rel = table_open(catalog_get_table_id(catalog, CHUNK), lock);
 	chunk_insert_relation(rel, chunk);
 	table_close(rel, lock);
-}
-
-static void
-chunk_fill(Chunk *chunk, HeapTuple tuple, TupleDesc desc)
-{
-	chunk_formdata_fill(&chunk->fd, tuple, desc);
-
-	chunk->table_id = get_relname_relid(chunk->fd.table_name.data,
-										get_namespace_oid(chunk->fd.schema_name.data, true));
-	chunk->hypertable_relid = ts_inheritance_parent_relid(chunk->table_id);
 }
 
 /*-
@@ -612,23 +640,26 @@ get_am_name_for_rel(Oid relid)
  * instead needs the proper permissions on the database to create the schema.
  */
 Oid
-ts_chunk_create_table(Chunk *chunk, Hypertable *ht, char *tablespacename)
+ts_chunk_create_table(Chunk *chunk, Hypertable *ht, const char *tablespacename)
 {
 	Relation rel;
 	ObjectAddress objaddr;
 	int sec_ctx;
+	char *namespace = NameStr(ht->fd.schema_name);
+	char *hyper_name = NameStr(ht->fd.table_name);
 	CreateStmt stmt = {
 		.type = T_CreateStmt,
 		.relation = makeRangeVar(NameStr(chunk->fd.schema_name), NameStr(chunk->fd.table_name), 0),
-		.inhRelations =
-			list_make1(makeRangeVar(NameStr(ht->fd.schema_name), NameStr(ht->fd.table_name), 0)),
-		.tablespacename = tablespacename,
+		.inhRelations = list_make1(makeRangeVar(namespace, hyper_name, 0)),
+		.tablespacename = tablespacename ? pstrdup(tablespacename) : NULL,
 		.options = get_reloptions(ht->main_table_relid),
 #if PG12_GE
 		.accessMethod = get_am_name_for_rel(ht->main_table_relid),
 #endif
 	};
 	Oid uid, saved_uid;
+
+	Assert(chunk->hypertable_relid == ht->main_table_relid);
 
 	rel = table_open(ht->main_table_relid, AccessShareLock);
 
@@ -673,8 +704,7 @@ ts_chunk_create_table(Chunk *chunk, Hypertable *ht, char *tablespacename)
 }
 
 static Chunk *
-chunk_create_catalog_tables_after_lock(Hypertable *ht, Point *p, const char *schema,
-									   const char *prefix)
+chunk_create_metadata_after_lock(Hypertable *ht, Point *p, const char *schema, const char *prefix)
 {
 	Hyperspace *hs = ht->space;
 	Catalog *catalog = ts_catalog_get();
@@ -719,15 +749,13 @@ chunk_create_catalog_tables_after_lock(Hypertable *ht, Point *p, const char *sch
 	return chunk;
 }
 
-static void
-chunk_create_postgres_objects_after_lock(Hypertable *ht, Chunk *chunk)
+static Oid
+chunk_create_table_after_lock(Chunk *chunk, Hypertable *ht)
 {
 	/* Create the actual table relation for the chunk */
-	chunk->table_id =
-		ts_chunk_create_table(chunk, ht, ts_hypertable_select_tablespace_name(ht, chunk));
-	CommandCounterIncrement();
-	chunk->hypertable_relid = ts_inheritance_parent_relid(chunk->table_id);
-	Assert(ht->main_table_relid == chunk->hypertable_relid);
+	const char *tablespace = ts_hypertable_select_tablespace_name(ht, chunk);
+
+	chunk->table_id = ts_chunk_create_table(chunk, ht, tablespace);
 
 	if (!OidIsValid(chunk->table_id))
 		elog(ERROR, "could not create chunk table");
@@ -739,12 +767,14 @@ chunk_create_postgres_objects_after_lock(Hypertable *ht, Chunk *chunk)
 								chunk->hypertable_relid,
 								chunk->fd.hypertable_id);
 
-	ts_trigger_create_all_on_chunk(ht, chunk);
+	ts_trigger_create_all_on_chunk(chunk);
 
 	ts_chunk_index_create_all(chunk->fd.hypertable_id,
 							  chunk->hypertable_relid,
 							  chunk->fd.id,
 							  chunk->table_id);
+
+	return chunk->table_id;
 }
 
 static void
@@ -772,37 +802,15 @@ ts_chunk_create(Hypertable *ht, Point *p, const char *schema, const char *prefix
 	LockRelationOid(ht->main_table_relid, ShareUpdateExclusiveLock);
 
 	/* Recheck if someone else created the chunk before we got the table lock */
-	chunk = ts_chunk_find(ht->space, p, true);
+	chunk = chunk_find(ht, p, true);
 
-	if (chunk != NULL && chunk->fd.dropped)
+	if (NULL == chunk)
 	{
-		/* mark as not dropped and recreate the pg objects */
-		ScanIterator iterator =
-			ts_scan_iterator_create(CHUNK, RowExclusiveLock, CurrentMemoryContext);
-		init_scan_by_chunk_id(&iterator, chunk->fd.id);
-
-		ts_scanner_foreach(&iterator)
-		{
-			TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
-			HeapTuple new_tuple;
-			FormData_chunk form;
-
-			chunk_formdata_fill(&form, ti->tuple, ti->desc);
-			form.dropped = false;
-			new_tuple = chunk_formdata_make_tuple(&form, ti->desc);
-			ts_catalog_update_tid(ti->scanrel, &ti->tuple->t_self, new_tuple);
-			heap_freetuple(new_tuple);
-		}
-
-		chunk_create_postgres_objects_after_lock(ht, chunk);
-	}
-	else if (NULL == chunk)
-	{
-		chunk = chunk_create_catalog_tables_after_lock(ht, p, schema, prefix);
-		chunk_create_postgres_objects_after_lock(ht, chunk);
+		chunk = chunk_create_metadata_after_lock(ht, p, schema, prefix);
+		chunk_create_table_after_lock(chunk, ht);
 	}
 
-	Assert(chunk != NULL);
+	ASSERT_IS_VALID_CHUNK(chunk);
 
 	return chunk;
 }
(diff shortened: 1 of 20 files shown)


src/chunk.c
```diff
@@ -177,9 +177,15 @@ chunk_fill(Chunk *chunk, HeapTuple tuple, TupleDesc desc)
...
-	chunk->table_id = get_relname_relid(chunk->fd.table_name.data,
-										get_namespace_oid(chunk->fd.schema_name.data, true));
-	chunk->hypertable_relid = ts_inheritance_parent_relid(chunk->table_id);
+	if (!chunk->fd.dropped)
+	{
+		chunk->table_id = get_relname_relid(chunk->fd.table_name.data,
+											get_namespace_oid(chunk->fd.schema_name.data, true));
+		chunk->hypertable_relid = ts_inheritance_parent_relid(chunk->table_id);
+
+		Assert(OidIsValid(chunk->table_id));
```
erimatnor, line 186: FWIW, this assert exists in multinode and was hit because we chunks that are marked dropped=true obviously have no real data tables. But this wasn't checked/handled in master.

src/chunk.c
```diff
@@ -1845,17 +1968,29 @@ ts_chunk_exists_relid(Oid relid)
...
 	FormData_chunk form;
 	CatalogSecurityContext sec_ctx;
 	ChunkConstraints *ccs = ts_chunk_constraints_alloc(2, ti->mctx);
+	ChunkDeleteResult res;
 	int i;
 
 	chunk_formdata_fill(&form, ti->tuple, ti->desc);
 
+	if (preserve_chunk_catalog_row && form.dropped)
+		return CHUNK_ALREADY_MARKED_DROPPED;
```
gayyappan, line 1992: why is this not an error condition? How do we run into this case where we try to mark the same chunk as dropped more than once?

src/chunk_constraint.c
```diff
@@ -651,17 +651,23 @@ chunk_constraint_delete_metadata(TupleInfo *ti)
 		heap_getattr(ti->tuple, Anum_chunk_constraint_constraint_name, ti->desc, &isnull);
 	int32 chunk_id =
 		DatumGetInt32(heap_getattr(ti->tuple, Anum_chunk_constraint_chunk_id, ti->desc, &isnull));
-	Chunk *chunk = ts_chunk_get_by_id(chunk_id, 0, true);
-	Oid index_relid = get_constraint_index(
-		get_relation_constraint_oid(chunk->table_id, NameStr(*DatumGetName(constrname)), true));
+	/* Get the chunk relid. Note that, at this point, the chunk table can be
+	 * deleted already */
+	Oid chunk_relid = ts_chunk_get_relid(chunk_id, true);
```
gayyappan, line 656: I don't understand the reason for the code change in this file : use ts_chunk_get_relid versus ts_chunk_get_by_id. Could you elaborate why the code is modified here?  It looks like ts_chunk_get_by_id will throw an error if we try to do this on a  chunk that is marked dropped. Isn't that the desired behavior?

src/chunk.c
```diff
@@ -1831,12 +1840,126 @@ ts_chunk_get_by_id(int32 id, int16 num_constraints, bool fail_if_not_found)
...
+									chunk_form_tuple_found,
+									form,
+									0,
+									ForwardScanDirection,
+									AccessShareLock,
+									CurrentMemoryContext);
+
+	found = (num_found == 1 && !form->dropped);
+
+	Assert(num_found == 0 || found);
```
WireBaron, line 1877: It looks like we'll always hit this assertion if there's a dropped chunk, is that the intention?

src/chunk.c
erimatnor: The metadata table is scanned using the UNIQUE <schema, name> index, which doesn't include the dropped column. Therefore, there is no filter on dropped chunks and this acts as the filter instead. 

Also not that there is a case when we'd like to DELETE also chunk entries that are "dropped", so we cannot always filter rows with dropped=true.

src/chunk_constraint.c
erimatnor: There are a couple of reasons. The main one is that I added some Asserts for hardening that is now hit without this function. For example, there's now and Assert that checks that a Chunk cannot have an invalid `table_id`, i.e., you cannot have a Chunk object without the table actually existing. When you DROP a hypertable and it cascades to chunks, the data table is actually removed before we remove the metadata, so this assert would be hit. Thus, we'd need a way to read chunk metadata rows without building complete chunks.

`ts_chunk_get_by_id` allocates a "complete" chunk, and joins in other metadata like constraints, relids for the data tables, etc. When dropping a chunk, this function is called and in various places over-and-over again when the drop cascades to other metadata, like `chunk_constraint_delete_metadata` and `chunk_constraint_drop_constraint`. In contrast, `ts_chunk_get_relid` only reads the minimal data

Ideally, we'd never be able to build `Chunk` objects when they don't really exists (e.g., we should never have a `Chunk` object that has `dropped=true`) but this requires a major refactoring of the code.
This is a lot of unnecessary work that's not needed in many places. In contrast,  `ts_chunk_get_relid` translates the internal ID to a relid by reading _only_ the chunk row metadata.

src/chunk.c
erimatnor: No, I guess not. Let me take another look.

src/chunk.c
erimatnor: I couldn't agree more, but that's a major refactoring (which I guess should have been done when this feature was introduced). 

At this point, however, I wanted to do minimal changes to continue with our rebase of the multinode branch. More refactoring, while worthwhile, will make the rebase harder. I could do it if we agree this is important (and I think it needs to happen at some point because the handling of dropped chunk is quite broken at the moment).

gayyappan: Yes, agree that refactoring can be done after code rebase.

erimatnor:
@gayyappan @WireBaron I updated the PR with further refactoring and assertions based on feedback. Now it is asserted that every public function in `chunk.{c,h}` that returns `Chunk` objects actually returns "valid" chunks (where, among other things, dropped=false).

src/chunk.c
```diff
@@ -1181,17 +1275,26 @@ ts_chunk_find(Hyperspace *hs, Point *p, bool include_chunks_marked_as_dropped)
...
 	}
 
 	return chunk;
 }
 
+Chunk *
+ts_chunk_find(Hypertable *ht, Point *p)
+{
+	Chunk *chunk = chunk_find(ht, p, false);
+	ASSERT_IS_NULL_OR_VALID_CHUNK(chunk);
```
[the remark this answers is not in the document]
erimatnor: I don't think so since it shouldn't happen. A chunk should be valid or not found.

src/chunk.c
```diff
@@ -739,7 +770,7 @@ chunk_create_postgres_objects_after_lock(Hypertable *ht, Chunk *chunk)
 								chunk->hypertable_relid,
 								chunk->fd.hypertable_id);
 
-	ts_trigger_create_all_on_chunk(ht, chunk);
+	ts_trigger_create_all_on_chunk(chunk, ht->space->main_table_relid);
```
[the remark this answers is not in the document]
erimatnor: I guess it is actually sufficient (and better) for `ts_trigger_create_all_on_chunk` to only take the `Chunk`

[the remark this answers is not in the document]
gayyappan: yes, don't need to pass hypertable_relid.

src/chunk.c
```diff
@@ -1123,6 +1155,58 @@ chunk_scan_ctx_get_chunk(ChunkScanCtx *ctx)
...
+		heap_freetuple(new_tuple);
+		count++;
+		/* Assume there's only one match */
+		break;
+	}
+
+	Assert(count == 0 || count == 1);
+
+	if (count == 0)
+		elog(ERROR, "cannot resurrect a chunk without a tombstone");
```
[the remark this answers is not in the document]
erimatnor: I realized this probably shouldn't be an error at all. I guess you must either succeed to resurrect a chunk if you already found a stub (i.e., there existed metadata). If you fail, then the only reason would be that someone removed the metadata before you had a chance to resurrect. And, in that case the resurrection should not succeed.

I updated the code to reflect this (also ensuring that resurrection happens while holding the lock on the chunk metadata row).

src/chunk.c
```diff
@@ -1123,6 +1153,62 @@ chunk_scan_ctx_get_chunk(ChunkScanCtx *ctx)
...
+		chunk->fd.dropped = false;
+		new_tuple = chunk_formdata_make_tuple(&chunk->fd, ti->desc);
+		ts_catalog_update_tid(ti->scanrel, &ti->tuple->t_self, new_tuple);
+		heap_freetuple(new_tuple);
+
+		/* Fill in the stub information and create data table and related
+		 * objects while we hold lock on the metadata row */
+		chunk->constraints = stub->constraints;
+		chunk->cube = stub->cube;
+		chunk_create_table_after_lock(ht, chunk);
```
gayyappan, line 1195: where is this functionality used - i.e. when do we need to recreate the table versus just using the metadata for a tombstoned chunk?

src/chunk.c
```diff
@@ -1123,6 +1153,62 @@ chunk_scan_ctx_get_chunk(ChunkScanCtx *ctx)
...
+
+		/* Assume there's only one match. (We break early to avoid scanning
+		 * also the updated tuple.) */
+		break;
+	}
+
+	ts_scan_iterator_close(&iterator);
+
+	Assert(count == 0 || count == 1);
+
```
gayyappan, line 1206: this functions is called when we see the state as dropped. Something is wrong if we try to reconstruct the metadata and it has been deleted from under us. This should error out and not silently ignore.
I am not clear about the changes to this function. Should discuss this further.

erimatnor: Can you explain why it is wrong?

This is my reasoning: with read committed isolation, the current transaction would be able to see changes that another transaction has made as soon as the other transaction is committed. If that other transaction deleted the chunk metadata (e.g., because the continuous aggregate was dropped), then the chunk metadata would be gone by the time we scan it in this function unless we held a lock on the chunk row from the point we created the ChunkStub that is an input parameter to this function.

In other words, how can we guarantee that the the metadata we used to create the ChunkStub remains until we scan it again in this function?

We actually have similar logic when creating the chunk initially to protect against a case when another transaction races to create the chunk before us. We first do a cheap lookup for an existing chunk without grabbing a heavy lock, then if the chunk is not found, we grab the heavy lock (on hypertable) followed by another lookup to ensure that the chunk hasn't been created by someone else before we grabbed the lock.

Thus, the logic is similar here: when we have determined we need to resurrect a chunk, we need to grab a heavy lock (RowExclusiveLock) on the row, but if the row is not found it must mean someone else deleted it before we got the lock. Or, am I missing something?

Note that we still hold the hypertable lock here, but I don't think it protects us here since the path to delete is different from the path to create.

src/chunk.c
```diff
@@ -884,27 +905,31 @@ chunk_fill_from_stub(Chunk *result, ChunkStub *chunk_stub, bool tuplock)
...
-		result->cube = ts_hypercube_from_constraints(result->constraints, CurrentMemoryContext);
+	chunk = result->chunk;
+	Assert(NULL != chunk);
+	chunk->fd.id = chunk_stub->id;
+	chunk->constraints = chunk_stub->constraints;
+	chunk->cube = chunk_stub->cube;
+
+	if (NULL == chunk->cube)
+		chunk->cube = ts_hypercube_from_constraints(chunk->constraints, CurrentMemoryContext);
 	else
```
WireBaron, line 924: I know that we don't typically use braces for single lines of code, but given the comment and I think they would improve readability here.  At the very least can you remove the empty line.

src/chunk.c
```diff
@@ -836,39 +842,54 @@ ts_chunk_create_base(int32 id, int16 num_constraints)
...
-	result->cube = chunk_stub->cube;
-	ctx = (ScannerCtx) {
+	Chunk *chunk;
+	ScannerCtx ctx = {
 		.table = catalog_get_table_id(catalog, CHUNK),
 		.index = catalog_get_index(catalog, CHUNK, CHUNK_ID_INDEX),
 		.nkeys = 1,
 		.scankey = scankey,
 		.data = result,
+		.filter = chunk_tuple_dropped_filter,
```
WireBaron, line 887: How does this filter work?  Looking at the function, I'd expect that this would result in the scan not returning a dropped chunk.  However, you also added code below to handle the case where a dropped chunk will return NULL instead of hitting an error.

erimatnor: Dropped chunks are not returned, but the is_dropped boolean is set in the ChunkStubScanCtx to indicate to the caller that there was metadata marked as dropped.

Not sure this answers your question?

src/chunk.c
erimatnor: There are two cases when we create tables: 1. When the chunk is first created (no tombstone) and 2. when it is resurrected (tombstone exists).

src/chunk.c
erimatnor: @gayyappan is this resolved now?

src/chunk.c
```diff
@@ -1790,53 +1977,195 @@ ts_chunk_get_by_relid(Oid relid, int16 num_constraints, bool fail_if_not_found)
...
+			const char *schema = get_namespace_name(nspid);
+
+			found = chunk_simple_scan_by_name(schema, table, form, missing_ok);
+		}
+	}
+
+	if (!found && !missing_ok)
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("chunk with relid %u not found", relid)));
```
k-rus, line 2060: What's about when `relid` is not valid? In such case the error can be improved.

src/chunk.c
```diff
@@ -1790,53 +1977,195 @@ ts_chunk_get_by_relid(Oid relid, int16 num_constraints, bool fail_if_not_found)
...
+		Oid schemaid = get_namespace_oid(NameStr(form.schema_name), missing_ok);
+
+		if (OidIsValid(schemaid))
+			relid = get_relname_relid(NameStr(form.table_name), schemaid);
+	}
+
+	if (!OidIsValid(relid) && !missing_ok)
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_SCHEMA),
+				 errmsg("chunk with relid %u not found", relid)));
```
k-rus, line 2126: Why is the error given in terms of `relid`, not `chunk_id? Should it be:
```suggestion
				 errmsg("chunk with id %u not found", chunk_id)));
```

src/chunk.h
```diff
@@ -101,31 +101,31 @@ typedef enum CascadeToMaterializationOption
...
 extern bool ts_chunk_exists(const char *schema_name, const char *table_name);
+extern Oid ts_chunk_get_relid(int32 chunk_id, bool missing_ok);
+extern Oid ts_chunk_get_schema_id(int32 chunk_id, bool missing_ok);
 extern bool ts_chunk_exists_relid(Oid relid);
 
 extern TSDLLEXPORT bool ts_chunk_exists_with_compression(int32 hypertable_id);
 extern void ts_chunk_recreate_all_constraints_for_dimension(Hyperspace *hs, int32 dimension_id);
-extern TSDLLEXPORT void ts_chunk_drop_fks(Chunk *chunk);
-extern TSDLLEXPORT void ts_chunk_create_fks(Chunk *chunk);
+extern TSDLLEXPORT void ts_chunk_drop_fks(Chunk *const chunk);
```
k-rus, line 127: You expose an implementation detail, which only affects the usage of the variable inside the function implementation, and has no sense for the function user. I suggest to remove this and below changes.

src/chunk_constraint.c
```diff
@@ -651,17 +652,23 @@ chunk_constraint_delete_metadata(TupleInfo *ti)
...
 
-	/*
-	 * If this is an index constraint, we need to cleanup the index
-	 * metadata. Don't drop the index though, since that will happend when
-	 * the constraint is dropped.
-	 */
-	if (OidIsValid(index_relid))
-		ts_chunk_index_delete(chunk, index_relid, false);
+	if (OidIsValid(chunk_relid))
```
k-rus, line 659: I believe it cannot be the other case, when it is not valid. I suggest to either remove the condition or replace with assert.

src/chunk_constraint.c
```diff
@@ -674,15 +681,21 @@ chunk_constraint_drop_constraint(TupleInfo *ti)
...
-		.objectId =
-			get_relation_constraint_oid(chunk->table_id, NameStr(*DatumGetName(constrname)), true),
-	};
+	/* Get the chunk relid. Note that, at this point, the chunk table can be
+	 * deleted already. */
+	Oid chunk_relid = ts_chunk_get_relid(chunk_id, true);
 
-	if (OidIsValid(constrobj.objectId))
-		performDeletion(&constrobj, DROP_RESTRICT, 0);
+	if (OidIsValid(chunk_relid))
```
k-rus, line 688: The other condition should not happen. Either remove the condition or replace with assert.

k-rus:
@erimatnor I forgot to add to my review that the case of deleting dropped chunk is not tested according the code coverage. It is good to have such test.

erimatnor:
> @erimatnor I forgot to add to my review that the case of deleting dropped chunk is not tested according the code coverage. It is good to have such test.

I agree. In fact there are many cases with dropping chunks that I think can be better tested. But this PR doesn't change existing functionality so in the interete

erimatnor:
Accidentally closed this PR, so reopening.

src/chunk.h
erimatnor: The microsoft compiler warns against having different declarations, so if we are to enable warnings as errors, we need to keep declaration and definition consistent. I also think that it is good rule in general to keep consistency.

k-rus: @erimatnor Good to know. What version do you use? Is it through command line or Visual Studio?

src/chunk_constraint.c
erimatnor: I am not sure I follow. What other condition are you referring to?

k-rus: I mean the `else` case.

k-rus: I.e., `OidIsValid(chunk_relid)`is always true.

k-rus: Actually I am wrong the function call specified `missing_ok = true`.

k-rus: It was `false` before.
I guess the change supports dropped case when no metadata should be deleted and no error should be produced.

src/chunk_constraint.c
erimatnor: It can be invalid, per the comment above.

Basically, there are multiple orders in which events can happen. 

1) Either the chunk and/or constraint is deleted before our own metadata
2) The chunk and/or constraint is deleted after our metadata

This code handles both cases.

src/chunk.h
erimatnor: I believe I used Visual Studio 2017, both via the command-line and Visual Studio.

src/chunk_constraint.c
k-rus: I understood that the condition is correct, since it specifies `missing_ok = true`, while it was `false` before and will not work correctly with the described case.

src/chunk.h
k-rus: I haven't noticed these warnings in 2019, but I didn't pay too much attention. I will double check how it works there.

src/chunk.c
```diff
@@ -836,40 +844,109 @@ ts_chunk_create_base(int32 id, int16 num_constraints)
...
+static Chunk *
+chunk_build_from_tuple_and_stub(Chunk **chunkptr, TupleInfo *ti, const ChunkStub *stub)
+{
+	Chunk *chunk = NULL;
+	int num_constraints_hint = stub ? stub->constraints->num_constraints : 2;
+
+	if (NULL == chunkptr)
+		chunkptr = &chunk;
+
+	if (NULL == *chunkptr)
```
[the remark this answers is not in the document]
erimatnor: The memory for the chunk can be pre-allocated by the caller. For instance, there's one path in the code that allocates on array of chunks (not array of Chunk pointers) and fills them in. In other cases, the memory is generally not pre-allocated.

The questions whether the stub is valid is an orthogonal matter and is checked by `chunk_stub_is_valid`.

src/chunk.c
```diff
@@ -880,31 +957,24 @@ chunk_fill_from_stub(Chunk *result, ChunkStub *chunk_stub, bool tuplock)
...
+		Assert(num_found == 0);
+		return NULL;
+	}
 
-		/*
-		 * The hypercube slices were filled in during the scan. Now we need to
-		 * sort them in dimension order.
-		 */
-		ts_hypercube_slice_sort(result->cube);
+	if (num_found != 1)
```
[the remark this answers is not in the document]
erimatnor: Not sure what is wrong with the current condition given that the only possible value here should be 0 or 1. So, if it is not 1, the chunk is not found.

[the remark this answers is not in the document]
k-rus: It's confusing due to the assert above. According to other places with similar condition, `num_found == 0` gives one error message, while `num_found > 1` gives another error message.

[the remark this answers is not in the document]
erimatnor: Sorry, but I am not following how this is confusing. It is asserted that `num_found` can be either 1 or 0, so I don't understand why anything greater than 1 is important here?

In any case, I can change, but I think what you propose because the only viable value one returning from this function below is num_count==1, which is exactly what this error condition is saying (even without asserts).

src/chunk.c
```diff
@@ -1406,13 +1563,36 @@ chunk_find_all(Hyperspace *hs, List *dimension_vecs, on_chunk_stub_func on_chunk
 Chunk **
 ts_chunk_find_all(Hyperspace *hs, List *dimension_vecs, LOCKMODE lockmode, unsigned int *num_chunks)
 {
-	return chunk_find_all(hs, dimension_vecs, append_chunk, lockmode, num_chunks);
+	Chunk **chunks = chunk_find_all(hs, dimension_vecs, append_chunk, lockmode, num_chunks);
+
+#ifdef USE_ASSERT_CHECKING
```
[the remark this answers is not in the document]
erimatnor: AFAIK, it depends on how you configured your PostgreSQL build (`--enable-cassert`), which is true for all our `Assert`s. Unless I am mistaken, CodeCov indicates that this path is executed. Are you seeing something else?

src/chunk.c
erimatnor: This is a rare case and it would actually say that relid is 0. IMO, it is not worth adding a custom error for this.

erimatnor:
@k-rus I looked into the testing of deleting dropped chunks. It turns out that CodeCov indicates that the path is hit when deleting chunks by <schema,name>, but not by <hypertable_id>. I actually refactored the code to handle these two delete cases with common code, since they shared a lot of commonality and now I believe this one case should be hit.

erimatnor:
@k-rus I looked into the testing of deleting dropped chunks. It turns out that CodeCov indicates that the path is hit when deleting chunks by <schema,name>, but not by <hypertable_id>. I actually refactored the code to handle these two delete cases with common code, since they shared a lot of commonality and now I believe this one case should be hit.

src/chunk.c
gayyappan: yes.

src/chunk.c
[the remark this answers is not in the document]
k-rus: I see in CodeCov that it doesn't show as executed, but it doesn't show opposite either. I confirm that it should be configured for majority of the tests. So I believe I am wrong. It would be nice to understand why it is not shown in CodeCov, but it is not part of this issue.

src/chunk.c
k-rus: You add the error, so I don't understand why it should not print a meaningful message. I think it is fine to keep the same error code.

erimatnor: I guess I misunderstood what you are asking. Can you explain why the current error is not meaningful and how you'd like it to be changed?

k-rus: In my last comment I was thinking about the other case, where you applied the changes already. Sorry it is my fault.

Actually your first reply answers my question.

src/chunk.c
```diff
@@ -612,24 +640,27 @@ get_am_name_for_rel(Oid relid)
  * instead needs the proper permissions on the database to create the schema.
  */
 Oid
-ts_chunk_create_table(Chunk *chunk, Hypertable *ht, char *tablespacename)
+ts_chunk_create_table(Chunk *chunk, Hypertable *ht, const char *tablespacename)
```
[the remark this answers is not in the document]
erimatnor: Unfortunately, this doesn't work because none of the PostgreSQL functions or data types use const. We'd have to do a copy or cast in that case (for instance, the `makeRangeVar` below doesn't accept const).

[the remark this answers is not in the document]
k-rus: I missed that there were calls to PG functions in this function. I noticed in other similar place and didn't comment there.

src/chunk.c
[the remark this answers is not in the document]
k-rus: The error message is only suitable for `num_found == 0`. Thus I find it confusing. In any case I don't find it is so important, since the result of execution is correct.

PR #1854, opened 2020-04-26
Fix reorder policy job to skip compressed chunks

### LINKED ISSUE #1810: Reorder job fails because of already compressed chunks

 PostgreSQL 11.6 (Debian 11.6-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
 - TimescaleDB version: 1.6.0
 - Deployed using TimescaleDB Helm charts

**Describe the bug**

We use the following compression policy:

```
ALTER TABLE timeseries.quantity SET (timescaledb.compress, timescaledb.compress_orderby = 'event_time, sequence', timescaledb.compress_segmentby = 'metric_ref');
```

for this table

```
create table quantity
(
	metric_ref integer not null,
	event_time timestamp not null,
	sequence integer not null,
	value double precision
);

create unique index quantity_pk
	on quantity (metric_ref asc, event_time desc, sequence desc);
```

Now we have added a 

```
SELECT add_reorder_policy('timeseries.quantity', 'quantity_pk');
```

but the job fails with

```
2020-04-10 16:47:56 UTC | 29646 | 5e90a33c.73ce-2 | @ | 22023 | ERROR:  "_hyper_3_5237_chunk" is a compressed chunk
2020-04-10 16:47:56 UTC | 64 | 5e70f65c.40-103 | @ | 00000 | LOG:  background worker "Reorder Background Job" (PID 29646) exited with exit code 1
```

**Expected behavior**

We expected the job to finish by skipping already compressed chunks (as they cannot be changed and are already ordered).

Olernov:
Reorder policy does not skip compressed chunks when selecting
next chunk to apply reordering. This causes an error in job
execution since it's not possible to reorder compressed chunks.
With this fix the job excludes compressed chunks from selection.

Fixes #1810

### DIFF
--- a/src/dimension_slice.c
+++ b/src/dimension_slice.c
@@ -732,6 +732,8 @@ typedef struct ChunkStatInfo
 	int32 job_id;
 } ChunkStatInfo;
 
+/* Check that a a) job has not already been executed for the chunk and b) chunk is not compressed
+ * (a compressed chunk should not be reordered).*/
 static ScanTupleResult
 dimension_slice_check_chunk_stats_tuple_found(TupleInfo *ti, void *data)
 {
@@ -744,13 +746,16 @@ dimension_slice_check_chunk_stats_tuple_found(TupleInfo *ti, void *data)
 
 	foreach (lc, chunk_ids)
 	{
-		BgwPolicyChunkStats *chunk_stat =
-			ts_bgw_policy_chunk_stats_find(info->job_id, lfirst_int(lc));
+		/* Look for a chunk that a) doesn't have a job stat (reorder ) and b) is not compressed
+		 * (should not reorder a compressed chunk) */
+		int chunk_id = lfirst_int(lc);
+		BgwPolicyChunkStats *chunk_stat = ts_bgw_policy_chunk_stats_find(info->job_id, chunk_id);
 
-		if (chunk_stat == NULL || chunk_stat->fd.num_times_job_run == 0)
+		if ((chunk_stat == NULL || chunk_stat->fd.num_times_job_run == 0) &&
+			ts_chunk_can_be_compressed(chunk_id))
 		{
 			/* Save the chunk_id */
-			info->chunk_id = lfirst_int(lc);
+			info->chunk_id = chunk_id;
 			return SCAN_DONE;
 		}
 	}
@@ -759,10 +764,9 @@ dimension_slice_check_chunk_stats_tuple_found(TupleInfo *ti, void *data)
 }
 
 int
-ts_dimension_slice_oldest_chunk_without_executed_job(int32 job_id, int32 dimension_id,
-													 StrategyNumber start_strategy,
-													 int64 start_value, StrategyNumber end_strategy,
-													 int64 end_value)
+ts_dimension_slice_oldest_valid_chunk_for_reorder(int32 job_id, int32 dimension_id,
+												  StrategyNumber start_strategy, int64 start_value,
+												  StrategyNumber end_strategy, int64 end_value)
 {
 	ChunkStatInfo info = {
 		.job_id = job_id,
--- a/tsl/test/sql/CMakeLists.txt
+++ b/tsl/test/sql/CMakeLists.txt
@@ -58,7 +58,7 @@ if (${PG_VERSION_MAJOR} GREATER "9")
     compression_hypertable.sql
     compression_segment_meta.sql
     compression_bgw.sql
-    compress_bgw_drop_chunks.sql
+    compress_bgw_reorder_drop_chunks.sql
     transparent_decompression_queries.sql
   )
 
--- a/tsl/test/sql/compress_bgw_reorder_drop_chunks.sql
+++ b/tsl/test/sql/compress_bgw_reorder_drop_chunks.sql
@@ -96,3 +96,75 @@ FROM _timescaledb_catalog.chunk chunk
 INNER JOIN _timescaledb_catalog.hypertable comp_hyper ON (chunk.hypertable_id = comp_hyper.id)
 INNER JOIN _timescaledb_catalog.hypertable uncomp_hyper ON (comp_hyper.id = uncomp_hyper.compressed_hypertable_id)
 WHERE uncomp_hyper.table_name like 'test_drop_chunks_table';
+
+------------------------------
+-- Test reorder policy runs on compressed tables. Reorder policy job must skip compressed chunks 
+-- (see issue https://github.com/timescale/timescaledb/issues/1810).
+-- More tests for reorder policy can be found at bgw_reorder_drop_chunks.sql
+------------------------------
+
+CREATE TABLE test_reorder_chunks_table(time int not null, chunk_id int);
+CREATE INDEX test_reorder_chunks_table_time_idx ON test_reorder_chunks_table(time);
+SELECT create_hypertable('test_reorder_chunks_table', 'time', chunk_time_interval => 1);
+
+-- These inserts should create 6 different chunks
+INSERT INTO test_reorder_chunks_table VALUES (1, 1);
+INSERT INTO test_reorder_chunks_table VALUES (2, 2);
+INSERT INTO test_reorder_chunks_table VALUES (3, 3);
+INSERT INTO test_reorder_chunks_table VALUES (4, 4);
+INSERT INTO test_reorder_chunks_table VALUES (5, 5);
+INSERT INTO test_reorder_chunks_table VALUES (6, 6);
+
+-- Enable compression
+ALTER TABLE test_reorder_chunks_table set (timescaledb.compress, timescaledb.compress_orderby = 'time DESC');
+
+-- Compress 2 chunks:
+SELECT compress_chunk(show_chunks('test_reorder_chunks_table', newer_than => 2, older_than => 4));
+
+-- make sure we have total of 6 chunks:
+SELECT count(*) as count_chunks_uncompressed
+FROM _timescaledb_catalog.chunk chunk
+INNER JOIN _timescaledb_catalog.hypertable hypertable ON (chunk.hypertable_id = hypertable.id)
+WHERE hypertable.table_name like 'test_reorder_chunks_table';
+
+-- and 2 compressed ones:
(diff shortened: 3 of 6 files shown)


Olernov:
I can't figure out why my new test fails in Travis. What I did is I renamed the existing test tsl/test/sql/compress_bgw_drop_chunks.sql to compress_bgw_reorder_drop_chunks.sql added new logic and replaced expected/*.out file correspondingly. I ran the test suite against PG 11.7 and 12.2 and they succeeded on my machine, but Travis displays that test _compress_bgw_reorder_drop_chunks_ fails.

tsl/test/sql/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ if (${PG_VERSION_MAJOR} GREATER "9")
     compression_hypertable.sql
     compression_segment_meta.sql
     compression_bgw.sql
-    compress_bgw_drop_chunks.sql
+    compress_bgw_reorder_drop_chunks.sql
```
k-rus, line 61: The added test doesn't fit well into this test file, since the file is about compression background worker, while the test is about reorder background worker.

k-rus: Will it be more natural to add the test to `bgw_reorder_drop_chunks.sql` and rename it to be for compressed and drop chunks?

k-rus:
> After the fix there still might be a problem when compress bgw and reorder bgw, which will try to compress and reorder the same chunk simultaneously. However, I think it is a separate issue and PR.

After this fix it might be more likely that the collision will happen, while before the fix the reorder job fails and doesn't run. May be a better fix will need to take in account compress jobs.

tsl/test/sql/CMakeLists.txt
Olernov: But `bgw_reorder_drop_chunks.sql` runs for all versions of PG while `compress_bgw_reorder_drop_chunks.sql` runs only for version higher than 9. If I add compression tests to `bgw_reorder_drop_chunks.sql` it will fail for PG 9.6. So I added them to `compress_bgw_reorder_drop_chunks.sql`.

k-rus: You are right.

k-rus: I can think to improve the name of this test: `bgw_compress_reorder_drop_chunks.sql`. However, I don't know how others will like this idea :)

Olernov:
> > After the fix there still might be a problem when compress bgw and reorder bgw, which will try to compress and reorder the same chunk simultaneously. However, I think it is a separate issue and PR.
> 
> After this fix it might be more likely that the collision will happen, while before the fix the reorder job fails and doesn't run. May be a better fix will need to take in account compress jobs.

Won't StartTransactionCommand() at job.c help?
```
bool
execute_reorder_policy(BgwJob *job, reorder_func reorder, bool fast_continue)
{
	...
	if (!IsTransactionOrTransactionBlock())
	{
		started = true;
		StartTransactionCommand();
	}
```

I think you talk about a more general problem than bgw jobs. Not only jobs, but users may also call compress_chunk() and reorder_chunk() on a same chunk simultaneously. I believe some kind of exclusive locking must be implemented already.

tsl/test/sql/CMakeLists.txt
Olernov: No problem, if you prefer another name or a separated file for this test - let me know.

src/dimension_slice.c
```diff
@@ -793,7 +795,7 @@ dimension_slice_check_is_chunk_uncompressed_tuple_found(TupleInfo *ti, void *dat
 	foreach (lc, chunk_ids)
 	{
 		int32 chunk_id = lfirst_int(lc);
-		if (ts_chunk_can_be_compressed(chunk_id))
+		if (ts_chunk_can_be_compressed_or_reordered(chunk_id))
```
gayyappan, line 798: Shouldn't combine the 2 different policies together. This is only for compress chunk job. Need functionality specifically for reorder chunks. (similar to ts_dimension_slice_get_chunkid_to_compress )

Olernov: Actually it's not a combination of policies, it's just a check shared among both policies. Both policies require the chunk not to be compressed or dropped. I don't think it makes sense to clone `ts_chunk_can_be_compressed `into `ts_chunk_can_be_reordered` 'cause the code will be duplicated. Maybe rename this check to `ts_chunk_is_not_compressed_or_dropped`?

src/chunk.c
```diff
@@ -2812,10 +2812,10 @@ ts_chunks_in(PG_FUNCTION_ARGS)
 	pg_unreachable();
 }
 
-/* Check if this chunk can be compressed, that it is not dropped and has not
- * already been compressed. */
+/* Check if this chunk can be compressed or reordered, that it is not dropped
+ * and has not already been compressed. */
 bool
-ts_chunk_can_be_compressed(int32 chunk_id)
+ts_chunk_can_be_compressed_or_reordered(int32 chunk_id)
```
gayyappan, line 2818: could you revert this change ?

src/dimension_slice.c
```diff
@@ -744,13 +744,14 @@ dimension_slice_check_chunk_stats_tuple_found(TupleInfo *ti, void *data)
 
 	foreach (lc, chunk_ids)
 	{
-		BgwPolicyChunkStats *chunk_stat =
-			ts_bgw_policy_chunk_stats_find(info->job_id, lfirst_int(lc));
+		int chunk_id = lfirst_int(lc);
+		BgwPolicyChunkStats *chunk_stat = ts_bgw_policy_chunk_stats_find(info->job_id, chunk_id);
```
gayyappan, line 749: Could you add a comment here
```suggestion
Look for a chunk that a) doesn't have a job stat (reorder ) and b) is not compressed (should not reorder compressed chunk)
```

src/dimension_slice.c
```diff
@@ -759,10 +760,11 @@ dimension_slice_check_chunk_stats_tuple_found(TupleInfo *ti, void *data)
 }
 
 int
-ts_dimension_slice_oldest_chunk_without_executed_job(int32 job_id, int32 dimension_id,
-													 StrategyNumber start_strategy,
-													 int64 start_value, StrategyNumber end_strategy,
-													 int64 end_value)
+ts_dimension_slice_oldest_valid_chunk_for_job_execution(int32 job_id, int32 dimension_id,
```
gayyappan, line 763: This is the reorder specific function. Lets rename this.
```suggestion
ts_dimension_slice_oldest_valid_chunk_for_reorder(int32 job_id, int32 dimension_id,
```

src/dimension_slice.c
gayyappan: Could you revert this change?

tsl/test/sql/compress_bgw_reorder_drop_chunks.sql
```diff
@@ -96,3 +96,87 @@ FROM _timescaledb_catalog.chunk chunk
...
+SELECT job_id, last_run_success, total_runs, total_successes, total_failures, total_crashes
+    FROM _timescaledb_internal.bgw_job_stat
+    where job_id=:reorder_job_id;
+
+-- two chunks clustered
+SELECT indexrelid::regclass, indisclustered
+    FROM pg_index
+    WHERE indisclustered = true ORDER BY 1;
+
+-- third call to scheduler: nothing to reorder since an eligible chunk
```
gayyappan, line 169: Don't think we need these additional tests. Can be removed.

tsl/test/sql/compress_bgw_reorder_drop_chunks.sql
```diff
@@ -96,3 +96,87 @@ FROM _timescaledb_catalog.chunk chunk
...
 WHERE uncomp_hyper.table_name like 'test_drop_chunks_table';
+
+------------------------------
+-- Test reorder policy runs on compressed tables. Reorder policy job must skip compressed chunks 
+-- (see issue https://github.com/timescale/timescaledb/issues/1810).
+-- More tests for reorder policy can be found at bgw_reorder_drop_chunks.sql
+------------------------------
+
+CREATE TABLE test_reorder_chunks_table(time int not null, chunk_id int);
+CREATE INDEX ON test_reorder_chunks_table(time);
```
gayyappan, line 107: could you add the name of the index here.

tsl/test/sql/compress_bgw_reorder_drop_chunks.sql
```diff
@@ -96,3 +96,87 @@ FROM _timescaledb_catalog.chunk chunk
...
+    WHERE indisclustered = true ORDER BY 1;
+
+-- second call to scheduler
+SELECT ts_bgw_db_scheduler_test_run_and_wait_for_scheduler_finish(25);
+
+SELECT job_id, last_run_success, total_runs, total_successes, total_failures, total_crashes
+    FROM _timescaledb_internal.bgw_job_stat
+    where job_id=:reorder_job_id;
+
+-- two chunks clustered
```
gayyappan, line 164:
```suggestion
-- two chunks clustered , skips the compressed chunks
```

src/dimension_slice.c
Olernov: Is it OK to call `ts_chunk_can_be_compressed` from `dimension_slice_check_chunk_stats_tuple_found` which is used only for reorder policy? Or I should create a separate function `ts_chunk_can_be_reordered`?

gayyappan: I don't think we need another function since this is just a check for uncompressed chunk. (The other option is to rename ts_chunk_can_be_compressed as you did. But would prefer to keep the changes minimal.)

src/chunk.c
```diff
@@ -3195,8 +3195,8 @@ ts_chunks_in(PG_FUNCTION_ARGS)
 	pg_unreachable();
 }
 
-/* Check if this chunk can be compressed, that it is not dropped and has not
- * already been compressed. */
+/* Check if this chunk can be compressed or reordered, that it is not dropped
```
gayyappan, line 3198: could you revert the comment change as well? We are only checking if it can be compressed

PR #1856, opened 2020-04-27
Silence various compiler warnings

erimatnor:
This change fixes various compiler warnings that show up on different
compilers and platforms. In particular, MSVC is sensitive to functions
that do not return a value after throwing an error since it doesn't
realize that the code path is not reachable.

### DIFF
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -369,7 +369,7 @@ hypertable_scan_limit_internal(ScanKeyData *scankey, int num_scankeys, int index
 		.limit = limit,
 		.tuple_found = on_tuple_found,
 		.lockmode = lock,
-        .filter = filter,
+		.filter = filter,
 		.scandirection = ForwardScanDirection,
 		.result_mctx = mctx,
 		.tuplock = {
@@ -823,20 +823,26 @@ ts_hypertable_lock_tuple_simple(Oid table_relid)
 					 errmsg("hypertable \"%s\" has already been updated by another transaction",
 							get_rel_name(table_relid)),
 					 errhint("Retry the operation again")));
+			pg_unreachable();
+			return false;
 		case TM_BeingModified:
 			ereport(ERROR,
 					(errcode(ERRCODE_LOCK_NOT_AVAILABLE),
 					 errmsg("hypertable \"%s\" is being updated by another transaction",
 							get_rel_name(table_relid)),
 					 errhint("Retry the operation again")));
+			pg_unreachable();
+			return false;
 		case TM_WouldBlock:
 			/* Locking would block. Let caller decide what to do */
 			return false;
 		case TM_Invisible:
 			elog(ERROR, "attempted to lock invisible tuple");
+			pg_unreachable();
 			return false;
 		default:
 			elog(ERROR, "unexpected tuple lock status");
+			pg_unreachable();
 			return false;
 	}
 }
--- a/src/hypertable_cache.h
+++ b/src/hypertable_cache.h
@@ -12,10 +12,11 @@
 #include "cache.h"
 #include "hypertable.h"
 
-extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, const Oid relid,
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *const cache, const Oid relid,
 															 const unsigned int flags);
-extern TSDLLEXPORT Hypertable *
-ts_hypertable_cache_get_cache_and_entry(const Oid relid, const unsigned int flags, Cache **cache);
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_cache_and_entry(const Oid relid,
+																	   const unsigned int flags,
+																	   Cache **const cache);
 extern Hypertable *ts_hypertable_cache_get_entry_rv(Cache *cache, const RangeVar *rv);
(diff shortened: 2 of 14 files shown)


erimatnor:
@mkindahl The PR actually lacked some changes from another commit that I have now squashed in. So, please take another look.

tsl/src/compression/create.c
```diff
@@ -583,13 +584,20 @@ create_compress_chunk_table(Hypertable *compress_ht, Chunk *src_chunk)
...
-	;
+	/* Fail if we overflow the name limit */
+	namelen = snprintf(NameStr(compress_chunk->fd.table_name),
+					   NAMEDATALEN,
+					   "compress%s_%d_chunk",
+					   NameStr(compress_ht->fd.associated_table_prefix),
+					   compress_chunk->fd.id);
+
+	if (namelen >= NAMEDATALEN)
+		ereport(ERROR,
```
mkindahl, line 596:
```suggestion
	if (namelen >= NAMEDATALEN)
	{
		ereport(ERROR,
```

tsl/src/compression/create.c
```diff
@@ -583,13 +584,20 @@ create_compress_chunk_table(Hypertable *compress_ht, Chunk *src_chunk)
...
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("invalid name \"%s\" for compressed chunk",
+						NameStr(compress_chunk->fd.table_name)),
+				 errdetail("The associated table prefix is too long.")));
```
mkindahl, line 600:
```suggestion
				 errdetail("The associated table prefix is too long.")));
				 pg_unreachable();
		}
```

src/hypertable.c
```diff
@@ -823,12 +823,16 @@ ts_hypertable_lock_tuple_simple(Oid table_relid)
...
+			pg_unreachable();
+			return false;
 		case TM_BeingModified:
 			ereport(ERROR,
 					(errcode(ERRCODE_LOCK_NOT_AVAILABLE),
 					 errmsg("hypertable \"%s\" is being updated by another transaction",
 							get_rel_name(table_relid)),
 					 errhint("Retry the operation again")));
+			pg_unreachable();
+			return false;
```
pmwkaa, line 835: There are also `TM_Invisible` and default cases below which looks identical. Maybe move `pg_unreachable()` with `return false;` at the end of the function instead and call `break` here instead?

src/hypertable.c
```diff
@@ -823,12 +823,16 @@ ts_hypertable_lock_tuple_simple(Oid table_relid)
 					 errmsg("hypertable \"%s\" has already been updated by another transaction",
 							get_rel_name(table_relid)),
 					 errhint("Retry the operation again")));
+			pg_unreachable();
+			return false;
```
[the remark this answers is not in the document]
erimatnor: Yes, I realize this is the case. But I think this is similar to many other cases where the error prevents a (default) return value, e.g.,

```
if (some_condition)
    elog(ERROR, "some error");

return foo;
```

We don't require that "foo" makes sense when `some_condition` is true or use `pg_unreachable()` in this case.

src/hypertable_cache.h
```diff
@@ -12,10 +12,11 @@
 #include "cache.h"
 #include "hypertable.h"
 
-extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *cache, const Oid relid,
+extern TSDLLEXPORT Hypertable *ts_hypertable_cache_get_entry(Cache *const cache, const Oid relid,
```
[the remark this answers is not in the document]
erimatnor: This is to silence warnings by `MSVC`, which otherwise complains about declaration and definition being different. I do think there's a point to keeping these in sync too, even if what you point out is strictly true.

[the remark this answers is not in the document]
mkindahl: Sounds like an error in MSVC, but yeah, we have to handle these as well so in that case you have to keep it.

src/process_utility.c
```diff
@@ -2697,6 +2697,7 @@ process_altertable_end_subcmd(Hypertable *ht, Node *parsetree, ObjectAddress *ob
 			ereport(ERROR,
 					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 					 errmsg("hypertables do not support logical replication")));
+			break;
```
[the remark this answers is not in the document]
erimatnor: I am worried that this will lead down a path where we require a `pg_uncreachable` after every error statement. There are lots of places where we have breaks and returns after ereport(ERROR) and don't have `pg_uncreachable`.

tsl/src/compression/create.c
erimatnor: I don't understand this suggestion. Why a single brace here?

mkindahl: It was for adding the `pg_unreachable` below. Maybe could have been done better.

tsl/src/compression/create.c
erimatnor: Again, this is moving us down a path where we put `pg_uncreachable` after every error. In that case, we should try to make a macro or something that automatically  inserts `pg_uncreachable` in case of error.

tsl/src/compression/create.c
erimatnor: Got it. FWIW, you can do multi-line suggestions in GitHub now.

PR #1875, opened 2020-05-05
Fix hypertable detection in subqueries

### LINKED ISSUE #1832: Regression on pruning chunks 1.7.0
- OS: CentOS
 - PostgreSQL version (output of `postgres --version`): 11.2
 - Installation method: Source

**Describe the bug**
When a new session is started, ConstraintAwareAppend/ChunkAppend is somehow not considered when calling a SQL query in an inlineable function. Executing any other query (without functions) that trigger one of Timescales own nodes, fixes this strangely.

**To Reproduce**
Setup:
```
create table test (a int, b timestamptz);
select create_hypertable('public.test', 'b');
insert into test select i, current_date-10-i from generate_series(1,20) i;

create or replace function test_f(_ts timestamptz)
returns setof test language sql stable parallel safe
as $f$
   select distinct on (a) * from test where b >= _ts order by a, b desc
$f$;
```

Disconnect and connect again now!

```
explain
select * from test_f(now());

Unique  (cost=279.34..292.94 rows=200 width=12)
  ->  Sort  (cost=279.34..286.14 rows=2721 width=12)
        Sort Key: test.a, test.b DESC
        ->  Append  (cost=0.00..124.11 rows=2721 width=12)
              ->  Seq Scan on test  (cost=0.00..0.00 rows=1 width=12)
                    Filter: (b >= now())
              ->  Bitmap Heap Scan on _hyper_353_51958_chunk  (cost=7.42..27.62 rows=680 width=12)
                    Recheck Cond: (b >= now())
                    ->  Bitmap Index Scan on _hyper_353_51958_chunk_test_b_idx  (cost=0.00..7.25 rows=680 width=0)
                          Index Cond: (b >= now())
              ->  Bitmap Heap Scan on _hyper_353_51959_chunk  (cost=7.42..27.62 rows=680 width=12)
                    Recheck Cond: (b >= now())
                    ->  Bitmap Index Scan on _hyper_353_51959_chunk_test_b_idx  (cost=0.00..7.25 rows=680 width=0)
                          Index Cond: (b >= now())
              ->  Bitmap Heap Scan on _hyper_353_51960_chunk  (cost=7.42..27.62 rows=680 width=12)
                    Recheck Cond: (b >= now())
                    ->  Bitmap Index Scan on _hyper_353_51960_chunk_test_b_idx  (cost=0.00..7.25 rows=680 width=0)
                          Index Cond: (b >= now())
              ->  Bitmap Heap Scan on _hyper_353_51961_chunk  (cost=7.42..27.62 rows=680 width=12)
                    Recheck Cond: (b >= now())
                    ->  Bitmap Index Scan on _hyper_353_51961_chunk_test_b_idx  (cost=0.00..7.25 rows=680 width=0)
                          Index Cond: (b >= now())
```

Now run the same query but outside of the function:
```
explain
select distinct on (a) * from test where b >= now() order by a, b desc;

Unique  (cost=196.61..210.21 rows=200 width=12)
  ->  Sort  (cost=196.61..203.41 rows=2720 width=12)
        Sort Key: test.a, test.b DESC
        ->  Custom Scan (ChunkAppend) on test  (cost=7.42..41.44 rows=2720 width=12)
              Chunks excluded during startup: 4
```

And now run the function again. This time we see the expected plan:

```
explain
select * from test_f(now());

Unique  (cost=188.38..201.99 rows=200 width=12)
  ->  Sort  (cost=188.38..195.18 rows=2721 width=12)
        Sort Key: test.a, test.b DESC
        ->  Custom Scan (ChunkAppend) on test  (cost=0.00..33.15 rows=2721 width=12)
              Chunks excluded during startup: 4
              ->  Seq Scan on test test_1  (cost=0.00..0.00 rows=1 width=12)
                    Filter: (b >= now())
```

**Additional context**
Reproduces on v11. On v12 it also doesn't choose a ConstraintAwareAppend, but that's not a problem, because v12 does its own pruning instead. So the plan at least looks like expected there.

svenklemm:
When classify_relation is called for relations of subqueries it
would not be able to correctly classify the relation unless it
was already in cache. This patch changes classify_relation to
call get_hypertable without CACHE_FLAG_NOCREATE when the RangeTblEntry
has the inheritance flag set.

Fixes #1832

### DIFF
--- a/test/expected/plan_hypertable_cache.out
+++ b/test/expected/plan_hypertable_cache.out
@@ -1,0 +1,46 @@
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+-- test hypertable classification when hypertable is not in cache
+-- https://github.com/timescale/timescaledb/issues/1832
+\set PREFIX 'EXPLAIN (costs off)'
+CREATE TABLE test (a int, time timestamptz NOT NULL);
+SELECT create_hypertable('public.test', 'time');
+ create_hypertable 
+-------------------
+ (1,public,test,t)
+(1 row)
+
+INSERT INTO test SELECT i, current_date-10-i from generate_series(1,20) i;
+CREATE OR REPLACE FUNCTION test_f(_ts timestamptz)
+RETURNS SETOF test LANGUAGE SQL STABLE PARALLEL SAFE
+AS $f$
+   SELECT DISTINCT ON (a) * FROM test WHERE time >= _ts ORDER BY a, time DESC
+$f$;
+:PREFIX SELECT * FROM test_f(now());
+                   QUERY PLAN                    
+-------------------------------------------------
+ Unique
+   ->  Sort
+         Sort Key: test.a, test."time" DESC
+         ->  Custom Scan (ChunkAppend) on test
+               Chunks excluded during startup: 4
+               ->  Seq Scan on test test_1
+                     Filter: ("time" >= now())
+(7 rows)
+
+-- create new session
+\c
+-- plan output should be identical to previous session
+:PREFIX SELECT * FROM test_f(now());
+                   QUERY PLAN                    
+-------------------------------------------------
+ Unique
+   ->  Sort
+         Sort Key: test.a, test."time" DESC
+         ->  Custom Scan (ChunkAppend) on test
+               Chunks excluded during startup: 4
+               ->  Seq Scan on test test_1
+                     Filter: ("time" >= now())
+(7 rows)
+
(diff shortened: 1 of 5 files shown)


test/expected/plan_hypertable_cache.out
```diff
@@ -0,0 +1,44 @@
...
+                     Filter: (b >= now())
+(7 rows)
+
+-- create new session
+\c
+-- plan output should be identical to previous session
+explain select * from test_f(now());
+                                      QUERY PLAN                                      
+--------------------------------------------------------------------------------------
+ Unique  (cost=247.45..261.06 rows=200 width=12)
```
WireBaron, line 36: Plan output shouldn't include the cost.

PR #1881, opened 2020-05-09
Fix flaky hypertable cache plan test

erimatnor:
The `plan_hypertable_cache` test used `current_date` to generate data,
which is inherently flaky since it can create a different number of
chunks depending on which date you start at. When the number of chunks
differ, the test output changes too.


erimatnor:
Travis tests were failing because of this.

PR #1890, opened 2020-05-12
Remove unused static declaration of a function

k-rus:
This removes warning of a function declared static but never defined.


k-rus:
The warning was likely introduced in https://github.com/timescale/timescaledb/pull/1886

PR #1906, opened 2020-05-15
Add real-time aggregation test to update test

svenklemm:
This patch adds a continuous aggregate with real-time aggregation
enabled to the update test suite since we rebuild view definition
for real time aggregation during extension update.
The continuous aggregate is in its own schema because all view
definitions in the public schema are dumped and those view
definitions will change between versions.

### DIFF
--- a/test/sql/updates/setup.continuous_aggs.v2.sql
+++ b/test/sql/updates/setup.continuous_aggs.v2.sql
@@ -70,4 +70,55 @@ AS
   GROUP BY bucket, location
   HAVING min(location) >= 'NYC' and avg(temperature) > 2;
 
-  REFRESH MATERIALIZED VIEW mat_before;
+REFRESH MATERIALIZED VIEW mat_before;
+
+-- we create separate schema for realtime agg since we dump all view definitions in public schema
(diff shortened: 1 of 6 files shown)


test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -70,4 +70,55 @@ AS
...
 
-  REFRESH MATERIALIZED VIEW mat_before;
+REFRESH MATERIALIZED VIEW mat_before;
+
+-- we create separate schema for realtime agg since we dump all view definitions in public schema
+-- but realtime agg view definition is not stable across versions
+CREATE SCHEMA cagg;
+
+CREATE VIEW cagg.realtime_mat
+WITH ( timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.refresh_lag='-30 day', timescaledb.max_interval_per_job ='1000 day')
```
gayyappan, line 80: shouldn't this be materialized_only = false?

svenklemm: good catch. fixed

PR #1910, opened 2020-05-18
Update tests to run on newest PG

k-rus:
PostgreSQL released 12.3, 11.8, 10.13 and 9.6.18. So tests, which uses
latest PostgreSQL, are updated to these versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,9 +95,9 @@ jobs:
 
     - if: (type = cron) OR (branch = prerelease_test)
       stage: test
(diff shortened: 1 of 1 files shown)


erimatnor:
Nit on PR commit title: probably sufficient to say just "latest PG" instead of "new latest PG".

k-rus:
> Nit on PR commit title: probably sufficient to say just "latest PG" instead of "new latest PG".

@erimatnor I thought about it two times already. What I want to say that only tests, which run on latest PG, are updated to new version of PG.

PR #1918, opened 2020-05-19
Unify chunk index creation

### LINKED ISSUE #1866: Table references in expression indices
I'm trying to create an expression index on a hypertable. The expression is a function call with the whole row as an argument:
`
CREATE INDEX ON "test"."data" (
	"timestamp" ASC, 
	my_function("data".*) ASC
);
`
This gives an error code 42P01 about missing the table "data" in the FROM clause. 

I think this is caused when the index is replicated to the chunk tables and the table name reference is not adjusted.

Maybe the following function needs a special case for table references?

https://github.com/timescale/timescaledb/blob/28e9a443b3a17ab889fd7dbd874e1cc6aee667bb/src/chunk_index.c#L94

svenklemm:
This patch changes chunk index creation to use the same functions
for creating index in one transaction and using multiple transactions.
The single transaction index creation used to adjust the original
stmt and adjusted it for the chunk which lead to problems with table
references not being adjusted properly for the chunk.

Fixes #1866

### DIFF
--- a/test/sql/index.sql
+++ b/test/sql/index.sql
@@ -229,3 +229,16 @@ CREATE INDEX idx_predicate_test_b1 ON idx_predicate_test(b1) WHERE b1=true;
 INSERT INTO idx_predicate_test VALUES ('2000-01-01',true);
 DROP TABLE idx_predicate_test;
 
+-- test index with table references
+CREATE TABLE idx_tableref_test(time timestamptz);
+SELECT table_name FROM create_hypertable('idx_tableref_test', 'time');
+-- we use security definer to prevent function inlining
+CREATE OR REPLACE FUNCTION tableref_func(t idx_tableref_test) RETURNS timestamptz LANGUAGE SQL IMMUTABLE SECURITY DEFINER AS $f$ SELECT t.time; $f$;
+-- try creating index with no existing chunks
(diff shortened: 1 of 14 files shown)


test/sql/index.sql
```diff
@@ -229,3 +229,17 @@ CREATE INDEX idx_predicate_test_b1 ON idx_predicate_test(b1) WHERE b1=true;
...
+-- test index with table references
+CREATE TABLE idx_tableref_test(time timestamptz);
+SELECT table_name FROM create_hypertable('idx_tableref_test', 'time');
+-- we use security definer to prevent function inlining
+CREATE OR REPLACE FUNCTION tableref_func(t idx_tableref_test) RETURNS timestamptz LANGUAGE SQL IMMUTABLE SECURITY DEFINER AS $f$ SELECT t.time; $f$;
+CREATE INDEX tableref_idx ON idx_tableref_test(tableref_func(idx_tableref_test));
+INSERT INTO idx_tableref_test SELECT '2000-01-01';
+DROP INDEX tableref_idx;
+CREATE INDEX tableref_idx ON idx_tableref_test(tableref_func(idx_tableref_test));
+
```
WireBaron, line 241: Few extra newlines added here.

PR #1922, opened 2020-05-19
Cleanup GUC names

svenklemm:
Change our GUC names to use enable-prefix for all boolean GUCs
similar to postgres GUC names.

This patch renames disable_optimizations to enable_optimizations and
constraint_aware_append to enable_constraint_aware_append.


src/guc.c
```diff
@@ -34,10 +34,10 @@ static const struct config_enum_entry telemetry_level_options[] = {
 	{ "off", TELEMETRY_OFF, false }, { "basic", TELEMETRY_BASIC, false }, { NULL, 0, false }
 };
 
-bool ts_guc_disable_optimizations = false;
+bool ts_guc_enable_optimizations = false;
```
k-rus, line 37: Should it be set to `true` here? It will correspond to before renaming.

test/expected/join.out
```diff
@@ -192,10 +192,10 @@ INSERT INTO tenk2 SELECT * FROM tenk1;
...
 +++ Optimized results
 @@ -1,6 +1,6 @@
-               setting              | value 
- -----------------------------------+-------
-- timescaledb.disable_optimizations | on
-+ timescaledb.disable_optimizations | off
+              setting              | value 
+ ----------------------------------+-------
+- timescaledb.enable_optimizations | on
++ timescaledb.enable_optimizations | off
```
k-rus, line 198: Shouldn't the values flip here?

src/guc.c
svenklemm: fixed

svenklemm:
> There's also a GUC `timescaledb_telemetry.cloud` which isn't namespaced like our other GUCs. I guess this one should be `timescaledb.telemetry_cloud`

I'll leave this one out of this PR as this GUC is definitely used externally and i think other teams had some requests regarding telemetry GUCs so we should probably get feedback if we modify it.

PR #1929, opened 2020-05-21
Only enable codecoverage on specific travis jobs

svenklemm:
Because add_link_options is not supported on older cmake versions
available in older postgresql docker images we only enable
code coverage on the codecov jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ script:
   # Run only telemetry test to see that it works without OpenSSL
   - ${RETRY_PREFIX} docker exec -u postgres -it pgbuild /bin/sh -c "make -C /build/debug-nossl installcheck TESTS=telemetry PG_REGRESS_OPTS='--temp-instance=/tmp/pgdata'"
   # Now build with OpenSSL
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -272,7 +272,8 @@ jobs:
     - if: (type = pull_request) OR (type = cron) OR NOT (branch = master)
       stage: test
       name: "Regression 9.6.17"
-      env: PG_VERSION=9.6.17
+      env:
+        - PG_VERSION=9.6.17
```
k-rus, line 276: I don't follow why this change is needed. It makes it inconsistent with the jobs below.

svenklemm: i'll remove i had multiple envs there for a while so this is leftover

svenklemm:
> Since `-DCODECOVERAGE=ON` is removed from common script, which I guess is good to do. Should it be added to the CodeCov jobs?
> 
> Also minor comment about unnecessary change.

It is in OTHER_CMAKE_FLAGS for the codecov jobs already

PR #1932, opened 2020-05-24
Change compression locking order

### LINKED ISSUE #1614: Compression job locking table referenced by the compressed hypertable
 - OS: Ubuntu 18.04.3
 - PostgreSQL version (output of `postgres --version`): 11.6
 - TimescaleDB version (output of `\dx` in `psql`): 1.5.1

**Describe the bug**
We have a compressed hypertable `quotes` with a foreign key to a regular table `markets` (which we segment by). The `compress_chunk` function on `quotes`, both when run by the background worker and when run manually, takes out a `SHARE ROW EXCLUSIVE` lock on `markets`, and holds that lock for a long time. This causes a disruption to other services that need to perform inserts or updates on `markets`.

**Expected behavior**
I expect compression to require an exclusive lock on the chunk being compressed, and a somewhat exclusive lock on the hypertable that chunk belongs to. Knowing that timescale turns foreign keys on the hypertable into foreign keys on chunks, and that compression seems to basically "create new chunks", I also understand it taking out a `SHARE ROW EXCLUSIVE` lock _briefly_ when adding that new reference, but if so I expect the lock to last a very short period of time. I'm also a bit surprised that the foreign key isn't added `NOT VALID`, given that the foreign key already exists on the uncompressed chunk, but I haven't thought through it enough to be sure whether that's safe.

**Actual behavior**
It looked like the compression job was taking out this lock repeatedly, even though there was only one chunk that needed to be compressed. The lock was also being held for several minutes, making me think that it was holding the lock for the entire duration of the compression job. The compression job never actually finished; I was killing processes that were blocked on the lock and didn't have a sensible `lock_timeout`, so it's possible that my intervention was to blame for that. Eventually I just disabled the compression policy until we figure this out.

**Additional context**
The hypertable in question has ~100 million rows, across 94 chunks (3 day interval), 91 of which are currently compressed.

spolcyn:
This patch changes the order in which locks are taken during
compression to avoid taking strong locks for long periods on referenced
tables.

Previously, constraints from the uncompressed chunk were copied to the
compressed chunk before compressing the data. When the uncompressed
chunk had foreign key constraints, this resulted in a
ShareRowExclusiveLock being held on the referenced table for the
remainder of the transaction, which includes the (potentially long)
period while the data is compressed, and prevented any
INSERTs/UPDATEs/DELETEs on the referenced table during the remainder of
the time it took the compression transaction to complete.

Copying constraints after completing the actual data compression does
not pose safety issues (as any updates to referenced keys are caught by
the FK constraint on the uncompressed chunk), and it enables the
compression job to minimize the time during which strong locks are held
on referenced tables.

Fixes #1614.

### DIFF
--- a/tsl/src/compression/create.c
+++ b/tsl/src/compression/create.c
@@ -564,6 +564,10 @@ create_compression_table(Oid owner, CompressColInfo *compress_cols)
 	return compress_hypertable_id;
 }
 
+/*
+ * Constraints and triggers are not created on the PG chunk table.
+ * Caller is expected to do this explicitly.
+ */
 Chunk *
 create_compress_chunk_table(Hypertable *compress_ht, Chunk *src_chunk)
 {
@@ -622,15 +626,6 @@ create_compress_chunk_table(Hypertable *compress_ht, Chunk *src_chunk)
 
 	if (!OidIsValid(compress_chunk->table_id))
 		elog(ERROR, "could not create compressed chunk table");
-
-	/* Create the chunk's constraints*/
-	ts_chunk_constraints_create(compress_chunk->constraints,
-								compress_chunk->table_id,
-								compress_chunk->fd.id,
-								compress_chunk->hypertable_relid,
-								compress_chunk->fd.hypertable_id);
-
-	ts_trigger_create_all_on_chunk(compress_chunk);
 
 	ts_chunk_index_create_all(compress_chunk->fd.hypertable_id,
 							  compress_chunk->hypertable_relid,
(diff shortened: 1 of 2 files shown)


tsl/src/compression/create.c
```diff
@@ -621,15 +621,6 @@ create_compress_chunk_table(Hypertable *compress_ht, Chunk *src_chunk)
 	if (!OidIsValid(compress_chunk->table_id))
 		elog(ERROR, "could not create compressed chunk table");
```
gayyappan, line 623:
```suggestion
Constraints and triggers are not created on the PG chunk table. Caller is expected to do this explicitly.
```
would be good to add a comment here or as a comment to the function.

PR #1941, opened 2020-05-28
Change default prefix for distributed tables

WireBaron:
This change modifies the default 'associated_table_prefix' string
for compressed hypertables from "_hyper_n" to "_dist_hyper_n" where
n is the hypertable id of the table.  This change makes it so that
when the backend hypertable for a distributed hypertable is created
on a data node, it won't have an 'associated_table_prefix' that may
conflict with locally created tables.

As part of this change, chunk names for distributed tables no longer
append the string "dist" as this is now part of the prefix for those
chunk names (i.e. instead of "_hyper_1_1_dist_chunk" we will now
have "_dist_hyper_1_1_chunk").


src/hypertable.c
```diff
@@ -964,10 +965,17 @@ hypertable_insert(int32 hypertable_id, Name schema_name, Name table_name,
 	{
 		NameData default_associated_table_prefix;
 		memset(NameStr(default_associated_table_prefix), '\0', NAMEDATALEN);
-		snprintf(NameStr(default_associated_table_prefix),
-				 NAMEDATALEN,
-				 DEFAULT_ASSOCIATED_TABLE_PREFIX_FORMAT,
-				 fd.id);
+		Assert(replication_factor >= 0);
+		if (replication_factor == 0)
+			snprintf(NameStr(default_associated_table_prefix),
```
erimatnor, line 970: I suggest we add a check for return value/length here and raise an error in case the name was truncated, like we do for the chunk table name above.

pmwkaa: Also nit, maybe call `snprintf` just once in that case but place naming change logic under if here? Likely will require an additional variable

PR #1949, opened 2020-05-28
Make travis coverity run not depend on other tests

svenklemm:
This patch changes the travis configuration to make the coverity run
not depend on the result of other test runs and runs it on the
coverity_scan branch only since there is a quota on the number of
coverity runs we can do and we dont want to exceed it by multiple
pushes to prerelease_test.


svenklemm:
> With this change, I suggest to put coverity tests of PG 11 and PG 12 into 2 separate branches, e.g., `coverity_scan_11` and `converity_scan_12`. Since after this change, only coverity on PG 11 will be accepted and PG 12 will not be run due to quota. WDYT?

According to the coverity documentation we should have 3 builds per day
```Up to 21 builds per week, with a maximum of 3 builds per day, for projects with 100K to 500K lines of code```

No objection from me, splitting the two makes it easier to stay under the quota but a single run will not push us over quota.

k-rus:
> According to the coverity documentation we should have 3 builds per day
> `Up to 21 builds per week, with a maximum of 3 builds per day, for projects with 100K to 500K lines of code`

Coverity has never authorized the second job to run even in current setup, when the second job is started after the first is completed. [This job](https://travis-ci.org/github/timescale/timescaledb/jobs/689782115) wasn't authorized, while it was within quota.

> No objection from me, splitting the two makes it easier to stay under the quota but a single run will not push us over quota.

I hope Coverity doesn't care which branch is submitted for analysis.

> I guess we can have both a coverity_scan that does both and specific branches for 11 and 12 if you want to rerun specific versions.

@svenklemm Since running two jobs were never accepted, I suggest to move PG 12 the first, if they are in the same branch.

svenklemm:
@k-rus Coverity has accepted both submissions before. See for example https://travis-ci.org/github/timescale/timescaledb/jobs/687546779

k-rus:
> @k-rus Coverity has accepted both submissions before. See for example https://travis-ci.org/github/timescale/timescaledb/jobs/687546779

Yes, you are right! Thank you for the link.

PR #1954, opened 2020-05-30
Remove PG 9.6 and PG10 test output

svenklemm:
This patch removes PG 9.6 and PG 10 test output files, removing
code support for these versions will be in a subsequent patch.


svenklemm:
> For many tests removing 9.6 and 10 allows to unify expected outputs for 11 and 12 into one single file. For example, `diff test/expected/alternate_users-11.out test/expected/alternate_users-12.out` gives empty result.
> I suggest to do it in this PR.

I've merged the files were 11 output didnt differ from 12 output.

PR #1955, opened 2020-05-30
Remove support for PG9.6 and PG10

svenklemm:
This patch removes code support PG9.6 and PG10. In addition to
removing PG96 and PG10 macros the following changes are done:
- remove HAVE_INT64_TIMESTAMP since this is always true on PG10+
- remove PG_VERSION_SUPPORTS_MULTINODE


src/chunk.c
```diff
@@ -2859,11 +2845,11 @@ ts_chunk_create_fks(Chunk *const chunk)
 	ASSERT_IS_VALID_CHUNK(chunk);
 
 	rel = table_open(chunk->hypertable_relid, AccessShareLock);
-	fks = copy_fk_list_from_cache(RelationGetFKeyListCompat(rel));
+	fks = RelationGetFKeyList(rel);
```
k-rus, line 2848: The result needs be copied. This is why the copy function was present.

svenklemm: good catch, expanded the wrong macro there

k-rus: It happened that I remember this code :)
I haven't reviewed much of other code as it is still in draft, but this one unlikely to be caught by our tests.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -196,10 +196,10 @@ adjust_data_node_rel_attrs(PlannerInfo *root, RelOptInfo *data_node_rel, RelOptI
 
 	data_node_rel->baserestrictinfo = nodequals;
 	data_node_rel->joininfo =
-		(List *) adjust_appendrel_attrs_compat(root, (Node *) hyper_rel->joininfo, appinfo);
+		(List *) adjust_appendrel_attrs(root, (Node *) hyper_rel->joininfo, 1, &appinfo);
```
erimatnor, line 199:
```suggestion
		castNode(List, adjust_appendrel_attrs(root, (Node *) hyper_rel->joininfo, 1, &appinfo));
```

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -196,10 +196,10 @@ adjust_data_node_rel_attrs(PlannerInfo *root, RelOptInfo *data_node_rel, RelOptI
...
 
 	data_node_rel->reltarget->exprs =
-		(List *) adjust_appendrel_attrs_compat(root, (Node *) hyper_rel->reltarget->exprs, appinfo);
+		(List *) adjust_appendrel_attrs(root, (Node *) hyper_rel->reltarget->exprs, 1, &appinfo);
```
erimatnor, line 202:
```suggestion
		castNode(List, adjust_appendrel_attrs(root, (Node *) hyper_rel->reltarget->exprs, 1, &appinfo));
```

tsl/src/init.c
```diff
@@ -372,15 +245,11 @@ _PG_init(void)
...
-#if PG_VERSION_SUPPORTS_MULTINODE
 	_remote_connection_init();
-#endif
 }
 
 PGDLLEXPORT void
 _PG_fini(void)
 {
-#if PG_VERSION_SUPPORTS_MULTINODE
 	_remote_connection_fini();
```
erimatnor, line 254: I realize this isn't within the scope of this PR and that module finilization/cleanup isn't really supported, but it is odd to only cleanup connections here. Maybe connection cleanup should move to module_shutdown() and have that shudown function called from here?

(Optional to fix in this PR. Just mental note).

src/plan_expand_hypertable.c
```diff
@@ -11,27 +11,21 @@
...
-#if PG11_GE
-#include <partitioning/partbounds.h>
-#include <optimizer/cost.h>
-#endif
 
-#if PG12_LT
+#if PG11
 #include <optimizer/clauses.h>
 #include <optimizer/var.h>
 #elif PG12_GE
```
WireBaron, line 31: Shouldn't this just be `#else`?

src/utils.c
```diff
@@ -55,17 +52,7 @@ ts_pg_timestamp_to_unix_microseconds(PG_FUNCTION_ARGS)
 		ereport(ERROR,
 				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE), errmsg("timestamp out of range")));
 
-#ifdef HAVE_INT64_TIMESTAMP
 	microseconds = timestamp + TS_EPOCH_DIFF_MICROSECONDS;
```
WireBaron, line 55: I don't know that we need this local variable (just directly return this calculation).

tsl/src/debug.c
```diff
@@ -62,25 +60,16 @@ static const char *reloptkind_name[] = {
...
 static const char *upperrel_stage_name[] = {
-	[UPPERREL_SETOP] = "SETOP",
-#if PG11_GE
-	[UPPERREL_PARTIAL_GROUP_AGG] = "PARTIAL_GROUP_AGG",
-#endif
-	[UPPERREL_GROUP_AGG] = "GROUP_AGG",
-	[UPPERREL_WINDOW] = "WINDOW",
-	[UPPERREL_DISTINCT] = "DISTINCT",
-	[UPPERREL_ORDERED] = "ORDERED",
+	[UPPERREL_SETOP] = "SETOP",			[UPPERREL_PARTIAL_GROUP_AGG] = "PARTIAL_GROUP_AGG",
```
WireBaron, line 70: Formatting here looks odd, is this after `make format`?

svenklemm:
> There should also be a bunch of test output files of the form `*-{9.6|10}.out` that can be removed from test/expected and tsl/test/expected.
> 
> One additional task I'd like to see done is to look at our test templates for tests where the pg11 and pg12 output is identical (so only a template due to differences in earlier versions) and convert them back to regular sql tests. This can probably just be filed as a separate issue if you don't feel like doing it now.

This has already been done in #1954

svenklemm:
> Also it is good to use `EXECUTE FUNCTION` instead of `EXECUTE PROCEDURE`, since `PROCEDURE` is deprecated since PG10. I believe it is used only in `ddl_trigger.sql`.

EXECUTE PROCEDURE is deprecated in PG11+ but pg_get_triggerdef in PG11 still uses EXECUTE PROCEDURE which means we won't be able to get completely rid of it. Will do it in separate PR.

tsl/src/debug.c
svenklemm: make format messes this one up for some reason and doesnt respect the formatting

tsl/src/init.c
```diff
@@ -6,56 +6,50 @@
...
-#include "telemetry.h"
+#include "bgw_policy/compress_chunks_api.h"
+#include "bgw_policy/drop_chunks_api.h"
 #include "bgw_policy/job.h"
 #include "bgw_policy/reorder_api.h"
-#include "bgw_policy/drop_chunks_api.h"
-#include "bgw_policy/compress_chunks_api.h"
+#include "chunk_api.h"
+#include "chunk.h"
+#include "compat.h"
```
k-rus, line 15: Can it be removed?

PR #1959, opened 2020-06-01
Move misc checks to github action

svenklemm:
Move code formatting, license and git commit hook checks to GitHub
action.


svenklemm:
> High-level comment:
> 
> I think we need to figure out and weigh the pros and cons of having tests in separate workflows vs as jobs in a single workflow. The trade-offs aren't obviously clear to me.
> 
> Now we're going to have three workflows and I guess there will be more as additional tests are moved to GitHub. Do you imagine that we'll add to this workflow as we move over more tests or create new ones?
> 
> It seems, though, that we should aim to reduce near identical configurations across workflows. I'd like us to avoid the copy-paste approach we've had in Travis.
> 
> It would be good if we figured out a strategy around this to avoid messy tests in the future.

The trigger rules are per workflow. This workflow has all the checks that only need to run once and are independent of platform or build configuration so doesnt require a build matrix.

.github/workflows/misc_checks.yaml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v2
+    - name: Check license
+      run: ./scripts/check_license_all.sh
+    - name: Check git commit hooks
+      run: |
+        python2 ./scripts/githooks/commit_msg_tests.py
+        python3 ./scripts/githooks/commit_msg_tests.py
```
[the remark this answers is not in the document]
svenklemm: to make sure it works with python2 and python3

.travis.yml
```diff
@@ -59,36 +59,6 @@ after_script:
...
-      install:
-        - docker exec pgbuild /bin/bash -c "mkdir /tsdb_build && chown postgres /tsdb_build"
-        - docker exec --env ASAN_OPTIONS="${ASAN_OPTIONS}" pgbuild /bin/bash -c "cd /tsdb_build && cmake /timescaledb -DCMAKE_BUILD_TYPE=Debug -DPG_SOURCE_DIR=/usr/src/postgresql/ -DUSE_DEFAULT_VISIBILITY=1 && make"
-      after_failure:
-      after_success:
-      script:
-        - docker run -it -v ${TRAVIS_BUILD_DIR}:/timescaledb timescaledev/postgres-dev-clang:clang7-pg${PG_VERSION} /timescaledb/scripts/check_license_all.sh
-        - docker run -it -v ${TRAVIS_BUILD_DIR}:/timescaledb timescaledev/postgres-dev-clang:clang7-pg${PG_VERSION} /timescaledb/scripts/clang_format_all.sh
-        - git diff --exit-code
-        - docker exec pgbuild /bin/bash -c "/tsdb_build/scripts/export_prefix_check.sh"
```
erimatnor, line 79: It looks like the export prefix check is removed here but not added to the GitHub actions.

.github/workflows/misc_checks.yaml
```diff
@@ -0,0 +1,24 @@
+name: Misc. Checks
```
erimatnor, line 1: Can we have something more descriptive here? What about simply "Code style" or "Code style and compliance"? Those examples seem to broadly cover what's done.

.travis.yml
svenklemm: This is intentional since the export prefix check requires a build, so it will be part of the main regression workflow

erimatnor: Got it

PR #1966, opened 2020-06-04
Remove noise from git changes on Windows

k-rus:
Fixes null device on Windows to work in both CMD and PowerShell.
Ignores scratch files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ typedef.list
 
 /CMakeSettings.json
 /out/*
+
+scratch_file*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -25,3 +25,5 @@ typedef.list
 
 /CMakeSettings.json
 /out/*
+
+scratch_file*
```
svenklemm, line 29: you sure this is right path, arent those created in the sql folder

k-rus: You are right that they are created in `sql` folder. However, I guess the same usage of scratch files might be implemented in other places in future, thus I suggest to ignore them everywhere.

k-rus: This notation is enough for `gitignore` to ignore the files in any folder.

k-rus:
> i think scratch_file should be cleaned up instead of being ignored

cc @gayyappan as the committer of the original change.

PR #1975, opened 2020-06-08
Add slack notification to i386 tests

svenklemm:
This patch adds a slack notification on failed i386 tests.
Unfortunately most slack actions don't work with the 32bit
image so our options for better styling/costumisation are
somewhat limited.


svenklemm:
> I cannot judge if the action mentioned is the right one, but I assume that you have checked.

Well i guess there is no right one, there are multiple available slack actions but unfortunately some do not work with the 32bit container

.github/workflows/alpine-32bit-build-and-test.yaml
```diff
@@ -74,3 +74,10 @@ jobs:
         name: PostgreSQL log
         path: postmaster.log
 
+    - name: Slack Notification
+      if: failure()
+      env:
+        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
```
erimatnor, line 80: Should we specify channel here?

.github/workflows/alpine-32bit-build-and-test.yaml
```diff
@@ -74,3 +74,10 @@ jobs:
...
+        SLACK_TITLE: Regression 32-bit PG${{ matrix.pg }} ${{ job.status }}
```
erimatnor, line 81: Maybe the title should mention that this is a failure?

Also, shouldn't we have a message that contains, e.g., a link to the build job that failes, git commit, author, etc.? Or is this somehow auto-generated/default? (The current info seems somewhat barebones.)

svenklemm: image

svenklemm: job.status will be failure

.github/workflows/alpine-32bit-build-and-test.yaml
svenklemm: the channel is configured in the webhook configuration, but it could be overwritten here

PR #1976, opened 2020-06-08
Run linux regression tests as GitHub action

svenklemm:


.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+      matrix:
+        pg: [ "11.0", "11.8", "12.0", "12.3" ]
+        type: [ Debug, Release, ApacheOnly ]
+
+        include:
+          - type: Debug
+            on_pullrequest: true
+            build_type: debug
+            pg_build_args: --enable-debug --enable-cassert
```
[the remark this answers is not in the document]
svenklemm: llvm and openssl are always enabled in the build postgres step, debug and cassert are only enabled for debug builds

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,144 @@
...
+
+    - name: coverage
+      if: (matrix.on_pullrequest || github.event_name != 'pull_request') && matrix.coverage
+      run: make -k -C build coverage
+
+    - name: Upload coverage report
+      if: (matrix.on_pullrequest || github.event_name != 'pull_request') && matrix.coverage
+      uses: codecov/codecov-action@v1
+      with:
+        file: ./build/codecov/timescaledb-codecov.info
```
erimatnor, line 121: While I don't mind saving this file, I am wondering if we are dropping usage of codecov.io?

FWIW, if we generate the HTML output, which now works, then we could upload it to, e.g., s3 bucket or similar for viewing.

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,144 @@
...
+      if: matrix.on_pullrequest || github.event_name != 'pull_request'
+      run: |
+        sudo make -C $PG_SRC_DIR install
+        sudo make -C $PG_SRC_DIR/contrib/postgres_fdw install
+
+    - name: Checkout TimescaleDB
+      if: matrix.on_pullrequest || github.event_name != 'pull_request'
+      uses: actions/checkout@v2
+
+    - name: Test telemetry without OpenSSL
```
erimatnor, line 89: Since we rarely touch telemetry, maybe we should move it to CRON only?

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+        cd $PG_SRC_DIR
+        ./configure --prefix=/usr/local ${{ matrix.pg_build_args }} --with-openssl --with-llvm --without-readline --without-zlib
+        make
+        make -C src/test/isolation
+        make -C contrib/postgres_fdw
+
+    - name: Install PostgreSQL ${{ matrix.pg }}
+      if: matrix.on_pullrequest || github.event_name != 'pull_request'
+      run: |
+        sudo make -C $PG_SRC_DIR install
```
erimatnor, line 82: I suggest using an install location that doesn't require sudo

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+        key: ${{ runner.os }}-postgresql-${{ matrix.pg }}-${{ matrix.build_type }}
+
+    - name: Build PostgreSQL ${{ matrix.pg }}
+      if: ( matrix.on_pullrequest || github.event_name != 'pull_request' ) && steps.cache-postgresql.outputs.cache-hit != 'true'
+      run: |
+        wget -q -O postgresql.tar.bz2 https://ftp.postgresql.org/pub/source/v${{ matrix.pg }}/postgresql-${{ matrix.pg }}.tar.bz2
+        mkdir -p $PG_SRC_DIR
+        tar --extract --file postgresql.tar.bz2 --directory $PG_SRC_DIR --strip-components 1
+        cd $PG_SRC_DIR
+        ./configure --prefix=/usr/local ${{ matrix.pg_build_args }} --with-openssl --with-llvm --without-readline --without-zlib
```
erimatnor, line 74: Why put this in `/usr/local` instead of a subdir of workdir? I guess this would unnecessarily require sudo for a bunch of install steps below

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+    - name: Install Dependencies
+      if: matrix.on_pullrequest || github.event_name != 'pull_request'
+      run: sudo apt install lcov
+
+    - name: Cache PostgreSQL
+      if: matrix.on_pullrequest || github.event_name != 'pull_request'
+      id: cache-postgresql
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.PG_SRC_DIR }}
```
erimatnor, line 64: Is this correct? Why are we caching (only) the source dir rather than the install dir that has the binaries? This makes the cache a lot bigger since you are also caching all the build artifacts (objects files, etc.). Then you also need to repeat the install step in every build. This seems to somewhat negate the purpose of the cache.

If you need both the binaries and the source, then I would create a directory in the workdir with the following structure:

```
mkdir $WORKDIR/postgresql
mkdir $WORKDIR/postgresql/source
mkdir $WORKDIR/postgresql/pgsql
```
Then unpack the source in `source`, configure with `--prefix=$WORKDIR/postgresql`, `make install`, then run `make clean` and cache `$WORKDIR/postgresql`

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+      if: (matrix.on_pullrequest || github.event_name != 'pull_request') && matrix.coverage
+      run: make -k -C build coverage
+
+    - name: Upload coverage report
+      if: (matrix.on_pullrequest || github.event_name != 'pull_request') && matrix.coverage
+      uses: codecov/codecov-action@v1
+      with:
+        file: ./build/codecov/timescaledb-codecov.info
+
+    - name: Regression diffs
```
erimatnor, line 123:
```suggestion
    - name: Show regression diffs
```

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+      if: failure()
+      run: |
+        find build -name regression.diffs -exec cat {} + > regression.diffs
+        find build -name postmaster.log -exec cat {} + > postmaster.log
+        cat regression.diffs
```
erimatnor, line 128: maybe we don't need to cat the diff if we upload the file?

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+    # run daily 20:00 on master branch
+    - cron: '0 20 * * *'
+  push:
+    branches:
+      - prerelease_test
+  pull_request:
+jobs:
+  regress_linux:
+    name: PG${{ matrix.pg }} [${{ matrix.type }}]
+    runs-on: ubuntu-18.04
```
erimatnor, line 13: I had hoped that this would cover also Mac. My original config for this (which I believe you were going to base this off of) supported Mac. This seems like a regression and not the improvement we are hoping for over Travis.

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+# matrix entries, the matrix has all versions but
+# for pull request we only run the configurations
+# where on_pullrequest is true
+
+      matrix:
+        pg: [ "11.0", "11.8", "12.0", "12.3" ]
+        type: [ Debug, Release, ApacheOnly ]
+
+        include:
+          - type: Debug
```
erimatnor, line 27: It would be good to have some builds run on GCC and others on Clang. For instance, if we'd support Mac we could run those builds with Clang and Linux ones with GCC.

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+    - name: Build PostgreSQL ${{ matrix.pg }}
+      if: ( matrix.on_pullrequest || github.event_name != 'pull_request' ) && steps.cache-postgresql.outputs.cache-hit != 'true'
+      run: |
+        wget -q -O postgresql.tar.bz2 https://ftp.postgresql.org/pub/source/v${{ matrix.pg }}/postgresql-${{ matrix.pg }}.tar.bz2
+        mkdir -p $PG_SRC_DIR
+        tar --extract --file postgresql.tar.bz2 --directory $PG_SRC_DIR --strip-components 1
+        cd $PG_SRC_DIR
+        ./configure --prefix=/usr/local ${{ matrix.pg_build_args }} --with-openssl --with-llvm --without-readline --without-zlib
+        make
```
erimatnor, line 75: I suggest adding `-j [num_process]` to speed up build.

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+        BUILD_DIR=nossl ./bootstrap -DCMAKE_BUILD_TYPE=Debug -DPG_SOURCE_DIR=$PG_SRC_DIR -DUSE_OPENSSL=false
+        make -C nossl
+        sudo make -C nossl install
+        make -C nossl installcheck TESTS=telemetry
+
+    - name: Build TimescaleDB
+      if: matrix.on_pullrequest || github.event_name != 'pull_request'
+      run: |
+        ./bootstrap -DCMAKE_BUILD_TYPE="${{ matrix.build_type }}" -DPG_SOURCE_DIR=$PG_SRC_DIR ${{ matrix.tsdb_build_args }}
+        make -C build
```
erimatnor, line 101: Also add `-j [num_process]` to speed up builds in each step here and elsewhere.

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+      run: ./build/scripts/export_prefix_check.sh
+
+    - name: make installcheck
+      if: matrix.on_pullrequest || github.event_name != 'pull_request'
+      id: installcheck
+      run: make -k -C build installcheck
+
+    - name: coverage
+      if: (matrix.on_pullrequest || github.event_name != 'pull_request') && matrix.coverage
+      run: make -k -C build coverage
```
erimatnor, line 115: If we use codecov.io, this step wouldn't be necessary.  (Only needed to build local coverage file). We can do both codecov.io and local coverage report, but right now it seems we are no longer doing the former at all. Just want to point out the options here.

.github/workflows/linux-build-and-test.yaml
svenklemm: This is intentional, make install is very fast and if we cache the actual place it installs to then we would have to clean it up after every run because timescaledb installs files in there as well and we wouldnt have a clean slate.

.github/workflows/linux-build-and-test.yaml
svenklemm: we are still doing codecov.io

svenklemm: Also this step is necessary because we need to generate the info file that we upload to codecov in the next step

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
+name: Regression Linux
+on:
+  schedule:
+    # run daily 20:00 on master branch
+    - cron: '0 20 * * *'
```
k-rus, line 5: Should it be different time than other scheduled workflows, so they don't wait for resources?

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,143 @@
...
+# for pull request we only run the configurations
+# where on_pullrequest is true
+
+      matrix:
+        pg: [ "11.0", "11.8", "12.0", "12.3" ]
+        type: [ Debug, Release, ApacheOnly ]
+
+        include:
+          - type: Debug
+            on_pullrequest: true
```
k-rus, line 28: This name `on_pullrequest` is confusing, since it happens implicitly on pull request. Will it be correct to say `run_always`?

.github/workflows/linux-build-and-test.yaml
svenklemm: this step uploads to codecov.io

.github/workflows/linux-build-and-test.yaml
k-rus: IMHO it is easier to take a look into a diff if it is printed. Otherwise, it is necessary to download the file every time a failure happens.

svenklemm: yes this is just for convenience so you dont have to download if you just want to take quick peek

.github/workflows/linux-build-and-test.yaml
svenklemm: i think mac should be a separate workflow, otherwise we need platform specific steps to support certain features like getting core dumps, having it separate would also prevent bloating the matrix further

.github/workflows/linux-build-and-test.yaml
svenklemm: sure no objections from me, dont think this test has high value

.github/workflows/linux-build-and-test.yaml
erimatnor: This would lead to a lot of duplication of this configuration because there _should_ essentially be no difference between running these steps on a Mac vs Linux. I actually spent a fair amount of time in making my original config work across both OSes, so I think it would be a shame to abandon multi-OS support now. Maybe we should have further discussion about this? 

Core dumps is a whole different issue, and we don't support it right now. I'd rather split into two workflows only when necessary.

.github/workflows/linux-build-and-test.yaml
erimatnor: I see your point about TimescaleDB install files being added to the cache. But you could get around that by copying the install dir and caching only the copy.

I do worry about having a much bigger cache that will slow things down, but maybe it is not a huge issue? Still, there seems to be easy steps to get around the "clean slate" issue by using a copy so why not keep it as efficient as possible?

In any case, please add comments describing why your're doing this because it is non-obvious and I am sure the question will come up again (or someone will change the config without realizing the functionality).

.github/workflows/linux-build-and-test.yaml
erimatnor: This wasn't answered...

.github/workflows/linux-build-and-test.yaml
erimatnor: FWIW, this was also one of the reasons I built PG from source as it makes the steps OS agnostic compared to custom install steps.

.github/workflows/linux-build-and-test.yaml
svenklemm: main reason was so the binaries will end up in PATH but this might actually not be needed i'll check

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,144 @@
...
+
+# This is the main workflow for linux regression tests
+# Since GitHub Actions don't currently allow dynamic
+# matrix entries, the matrix has all versions but
+# for pull request we only run the configurations
+# where on_pullrequest is true
+
+      matrix:
+        pg: [ "11.0", "11.8", "12.0", "12.3" ]
+        type: [ Debug, Release, ApacheOnly ]
```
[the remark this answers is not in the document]
svenklemm: i put it there to keep the number of dimensions in the matrix low

.github/workflows/linux-build-and-test.yaml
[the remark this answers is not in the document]
pmwkaa: Is it also true for PG11?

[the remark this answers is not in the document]
svenklemm: yes

.github/workflows/linux-build-and-test.yaml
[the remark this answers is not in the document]
k-rus: Are there more specific technical issues, which will come from increasing the dimensions? The current approach seems to have less clarity and thus to be more error-prone.

Will it work to specify:
```yml
exclude:
  - source: ApacheOnly
    type: Debug
```

.github/workflows/linux-build-and-test.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+        include:
+          - type: Debug
+            on_pullrequest: true
+            build_type: Debug
```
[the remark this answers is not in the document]
svenklemm: type is a toplevel name which is one of Debug,Release,ApacheOnly
build_type is cmake build type of timescaledb

.github/workflows/linux-build-and-test.yaml
svenklemm: i think there is value in having them finish roughly at the same time, since this is trivial to change we can change later if it turns out as problem

PR #1977, opened 2020-06-08
Improve connection handling during transactions

erimatnor:
This change refactors how connections are handled during remote
transactions. In particular, the connection cache now stays consistent
during transactions, even during rollbacks. Previously, the connection
cache was replaced on every rollback, even if the rollback was
intentional (i.e, not due to an error). This made it hard to debug
connections since the cache became completely empty.

Connections could also be left in the cache in a bad state after
failed transactions. This has been fixed by moving connection checks
to the cache and tying transaction state changes to each
connection. This ensures that such checks are done in one canonical
place instead of being spread out throughout the code.

Given how tightly coupled a remote transaction is with its connection,
it might make sense to remove the separate remote transaction store
and instead put this information in each connection. This is left to a
future change, however.

In addition to the above changes, this commit includes:

* Showing transaction depth and invalidation in the transaction store
* Invalidation on individual connections instead of replacing the
  whole cache
* Closing of connections to a local database that is being dropped to
  prevent "in use" errors.
* Ability to add callbacks to async requests that are executed when a
  response is received. This is used by remote transactions to mark
  connections as having successfully completed a transaction. Thus, on
  errors, it is easy to detect connections that are in bad states.
* Error checks on each connection instead of having global error
  tracking for each remote transaction. This change removes the global
  error state for distributed transactions.

Note that this is a multi-commit PR. In particular, the first commit just adds the ability to print the cache and this commit can be viewed by itself to see some of the oddness that occurs when printing the connection cache without any additional changes in the other commits.

### DIFF
--- a/tsl/src/remote/async.c
+++ b/tsl/src/remote/async.c
@@ -87,6 +87,12 @@ typedef struct AsyncResponseCommunicationError
 	AsyncRequest *request;
 } AsyncResponseCommunicationError;
 
+typedef struct AsyncResponseError
+{
+	AsyncResponse base;
+	const char *errmsg;
+} AsyncResponseError;
+
 typedef struct AsyncRequestSet
 {
 	List *requests;
@@ -152,7 +158,8 @@ static AsyncRequest *
 async_request_send_internal(AsyncRequest *req, int elevel)
 {
 	if (req->state != DEFERRED)
-		elog(ERROR, "can't send async request in state \"%d\"", req->state);
+		elog(elevel, "can't send async request in state \"%d\"", req->state);
+
 	if (remote_connection_is_processing(req->conn))
 		return req;
 
@@ -319,6 +326,19 @@ async_response_timeout_create()
 	return ares;
 }
 
+static AsyncResponse *
+async_response_error_create(const char *errmsg)
+{
+	AsyncResponseError *ares = palloc0(sizeof(AsyncResponseError));
+
+	*ares = (AsyncResponseError){
+		.base = { .type = RESPONSE_ERROR },
+		.errmsg = pstrdup(errmsg),
+	};
+
+	return &ares->base;
+}
+
 void
 async_response_result_close(AsyncResponseResult *res)
 {
@@ -333,6 +353,7 @@ async_response_close(AsyncResponse *res)
 	switch (res->type)
 	{
 		case RESPONSE_RESULT:
+		case RESPONSE_ROW:
 			async_response_result_close((AsyncResponseResult *) res);
 			break;
 		default:
@@ -385,15 +406,53 @@ async_response_report_error(AsyncResponse *res, int elevel)
 	{
 		case RESPONSE_RESULT:
 		case RESPONSE_ROW:
-			remote_result_elog(((AsyncResponseResult *) res)->result, elevel);
+		{
+			AsyncResponseResult *aresult = (AsyncResponseResult *) res;
+			ExecStatusType status = PQresultStatus(aresult->result);
+
+			switch (status)
+			{
+				case PGRES_COMMAND_OK:
+				case PGRES_TUPLES_OK:
+				case PGRES_SINGLE_TUPLE:
+					break;
+				case PGRES_NONFATAL_ERROR:
+				case PGRES_FATAL_ERROR:
+					/* result is closed by remote_result_elog in case it throws
+					 * error */
+					remote_result_elog(aresult->result, elevel);
+					break;
+				default:
+					PG_TRY();
+					{
+						elog(elevel, "unexpected response status %u", status);
+					}
+					PG_CATCH();
+					{
+						async_response_close(res);
+						PG_RE_THROW();
+					}
+					PG_END_TRY();
+			}
 			break;
+		}
 		case RESPONSE_COMMUNICATION_ERROR:
 			remote_connection_elog(((AsyncResponseCommunicationError *) res)->request->conn,
 								   elevel);
 			break;
+		case RESPONSE_ERROR:
+			elog(elevel, "%s", ((AsyncResponseError *) res)->errmsg);
+			break;
 		case RESPONSE_TIMEOUT:
 			elog(elevel, "async operation timed out");
 	}
+}
+
+void
+async_response_report_error_or_close(AsyncResponse *res, int elevel)
+{
+	async_response_report_error(res, elevel);
+	async_response_close(res);
 }
 
 /*
@@ -407,7 +466,6 @@ async_request_wait_any_result(AsyncRequest *req)
 	AsyncResponseResult *result;
 
 	async_request_set_add(&set, req);
-
 	result = async_request_set_wait_any_result(&set);
 
 	/* Should expect exactly one response */
@@ -415,8 +473,24 @@ async_request_wait_any_result(AsyncRequest *req)
 		elog(ERROR, "remote request failed");
 
 	/* Make sure to drain the connection only if we've retrieved complete result set */
-	if (result->base.type == RESPONSE_RESULT && async_request_set_wait_any_result(&set) != NULL)
-		elog(ERROR, "request must be for one sql statement");
+	if (result->base.type == RESPONSE_RESULT)
+	{
+		AsyncResponseResult *extra;
+		bool got_extra = false;
+
+		/* Must drain any remaining result until NULL */
+		while ((extra = async_request_set_wait_any_result(&set)))
+		{
+			async_response_result_close(extra);
+			got_extra = true;
+		}
+
+		if (got_extra)
+		{
+			async_response_result_close(result);
+			elog(ERROR, "request must be for one sql statement");
+		}
+	}
 
 	return result;
 }
@@ -488,16 +562,18 @@ get_single_response_nonblocking(AsyncRequestSet *set)
 		PGconn *pg_conn = remote_connection_get_pg_conn(req->conn);
 
 		if (remote_connection_is_processing(req->conn) && req->state == DEFERRED)
-			elog(ERROR,
-				 "can't get a response because previous request hasn't completed. Aborted SQL "
-				 "statement: %s",
-				 req->sql);
+			return async_response_error_create("request already in process");
 
 		if (req->state == DEFERRED)
-			async_request_send_internal(req, ERROR);
+		{
+			req = async_request_send_internal(req, WARNING);
+
+			if (req == NULL)
+				return async_response_error_create("failed to send deferred request");
+		}
 
 		if (req->state != EXECUTING)
-			elog(ERROR, "can't get a response for a request that's not currently executing");
+			return async_response_error_create("no request currently executing");
 
 		if (!PQisBusy(pg_conn))
 		{
@@ -528,7 +604,7 @@ get_single_response_nonblocking(AsyncRequestSet *set)
  * Returns NULL on success or an "error" AsyncResponse
  */
 static AsyncResponse *
-wait_to_consume_data(AsyncRequestSet *set, int elevel, TimestampTz end_time)
+wait_to_consume_data(AsyncRequestSet *set, TimestampTz end_time)
 {
 	/*
 	 * Looks like there is no good way to modify a WaitEventSet so we have to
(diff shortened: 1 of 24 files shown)


tsl/src/remote/connection_cache.c
```diff
@@ -62,16 +70,50 @@ connection_cache_get_key(CacheQuery *query)
 }
 
 /*
- * Verify that a connection is still valid.
+ * Check if a connection needs to be remade.
+ *
+ * A connection can be in a bad state, in which case we need to
+ * fail. Otherwise, the connection could be invalidated and needs to be remade
+ * to apply new options. But a connection can onlyl be remade if we are not
```
pmwkaa, line 77: nit: typo `onlyl`

tsl/src/remote/connection_cache.c
```diff
@@ -156,49 +203,226 @@ connection_cache_create()
...
- * mark the cache as invalid.
+ * mark the cache entry as invalid.
+ */
+void
+remote_connection_cache_invalidate_callback(Datum arg, int cacheid, uint32 hashvalue)
+{
+	HASH_SEQ_STATUS scan;
+	ConnectionCacheEntry *entry;
+
+	if (cacheid != FOREIGNSERVEROID)
```
pmwkaa, line 233: Assert?

tsl/src/remote/connection_cache.c
```diff
@@ -156,49 +203,226 @@ connection_cache_create()
...
+ * It currently doesn't capture being connected on a local network interface
+ * address. That would require a platform-independent way to lookup local
+ * interface addresses (on UNIX one could use getifaddrs, for instance).
+ */
+static bool
+is_local_connection(const PGconn *conn)
+{
+	const char *host = PQhost(conn);
+
+	return (strcmp("localhost", host) == 0 || strcmp("127.0.0.1", host) ||
```
pmwkaa, line 260: This likely should include at least IPv6 `::1` host as well, but I am also not sure that it is enough as you said.

We probably should set this flag at the moment when host address being resolved first time instead checking it here, we likely will have more context about it from name resolver.

tsl/src/remote/connection_cache.c
```diff
@@ -156,49 +203,226 @@ connection_cache_create()
...
+
+	hash_seq_init(&scan, connection_cache->htab);
+
+	while ((entry = hash_seq_search(&scan)) != NULL)
+	{
+		PGconn *pgconn = remote_connection_get_pg_conn(entry->conn);
+
+		/* Remove the connection if is local and connects to a DB with the
+		 * same name as the one being dropped. */
+		if (strcmp(dbname, PQdb(pgconn)) == 0 && is_local_connection(pgconn))
```
pmwkaa, line 289: Maybe we should check port as well? Not sure if this case is possible here, what if we connect to some other local instance on different port but having the same db name

tsl/src/remote/txn.c
```diff
@@ -444,112 +510,162 @@ remote_txn_write_persistent_record(RemoteTxn *entry)
...
-		return false;
+	if (in_error_recursion_trouble() && remote_connection_xact_is_transitioning(entry->conn))
+		remote_connection_xact_transition_begin(entry->conn);
 
-	if (PQtransactionStatus(pg_conn) != PQTRANS_INTRANS &&
-		PQtransactionStatus(pg_conn) != PQTRANS_INERROR)
-		return false;
-
-	entry->have_subtxn_error = true;
+	if (!remote_connection_xact_is_transitioning(entry->conn))
```
pmwkaa, line 589: Do we have a test case for it?

tsl/src/remote/connection_cache.c
erimatnor: Good point. I will add a check in is_local_connection

tsl/src/remote/txn.c
erimatnor: This should be tested as part of the remote_txn test

tsl/src/remote/connection_cache.c
erimatnor: It's actually asserted in cache_invalidate_syscache_callback already so another assertion here wouldn't have any effect. Figured I'd leave this check here for clarity.

tsl/src/remote/connection_cache.c
svenklemm: dont you want to check for instance here and dont really care if its local or remote

erimatnor: @svenklemm Not sure what you mean. Can you reformulate?

FWIW, this check is for dropping a database on the same instance as the connection connects to. If the connection is to a remote instance, then we don't care.

svenklemm: Yes but the remote/local distinction is not important you only care if its the same instance even if its local

erimatnor: @svenklemm do you have a suggestion how to check for "same instance"?

svenklemm: could check it against information in MyPort https://doxygen.postgresql.org/structPort.html

erimatnor: Still don't understand how that helps (code examples welcome).  
Conceptually, the connections look like this:

```
client -> backend_A -> backend_B
```
As backend_A, we need to identify if the connection from backend_A to backend_B is to the local instance, i.e., backend_A and backend_B were forked from the same postmaster

I can access `MyProcPort`, but that only gives the (current) backend A's connection info (i.e., the client). I guess if was possible to get access to backend B's Port struct via shared memory I could somehow compare that against the connection info.

svenklemm: laddr has information about the server, i guess the most resilient way would be to check our own uuid which we could cache on connection setup, checking only against laddr might still fail in case postmaster is listening on multiple interfaces but it would be huge improvement over checking against "localhost" and "127.0.0.1"

tsl/src/remote/txn.c
```diff
@@ -444,112 +506,161 @@ remote_txn_write_persistent_record(RemoteTxn *entry)
...
-	if (in_error_recursion_trouble())
-		return false;
+	Assert(remote_connection_xact_depth_get(entry->conn) == curlevel);
+	Assert(remote_connection_xact_depth_get(entry->conn) > 1);
 
-	if (PQtransactionStatus(pg_conn) != PQTRANS_INTRANS &&
-		PQtransactionStatus(pg_conn) != PQTRANS_INERROR)
-		return false;
+	if (in_error_recursion_trouble() && remote_connection_xact_is_transitioning(entry->conn))
+		remote_connection_xact_transition_begin(entry->conn);
```
k-rus, line 582: Something is wrong here. The condition is satisfied if transitioning already started. Should it be:
```suggestion
	if (in_error_recursion_trouble() && !remote_connection_xact_is_transitioning(entry->conn))
		remote_connection_xact_transition_begin(entry->conn);
```

tsl/src/remote/txn.c
```diff
@@ -444,112 +506,161 @@ remote_txn_write_persistent_record(RemoteTxn *entry)
...
 
-	entry->have_subtxn_error = true;
+	if (!remote_connection_xact_is_transitioning(entry->conn))
```
k-rus, line 584: In the else case, is it correct that `success` is `true`? Should it be `false`? If so, then it is better to do it above, close to check for troubles with recursion, since it will make easier to follow.

k-rus: It seems to me if the success is returned, the transitioning should be ended, righ?

k-rus:
@erimatnor Have you looked to the [code coverage result](https://codecov.io/gh/timescale/timescaledb/compare/5bbab82eca4f65d5686f8622dc44059083b7ce3d...d298d0f73b1d0c959017e288e7d556728313c92a/diff)? Does it show correctly that many cases are not tested?

tsl/src/remote/async.c
```diff
@@ -615,17 +630,52 @@ async_request_set_wait_any_response_deadline(AsyncRequestSet *set, int elevel, T
...
+			case RESPONSE_TIMEOUT:
+				requests = set->requests;
+				break;
+		}
+
+		foreach (lc, requests)
+		{
+			AsyncRequest *req = lfirst(lc);
+
+			if (NULL != req->response_cb)
```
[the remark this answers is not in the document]
k-rus: Sorry, I missed it. You are right.

erimatnor:
@k-rus the code coverage report looked very strange last time I looked at it. I will take a closer look, but given that I haven't really added new functionality I didn't expand testing either.

tsl/test/expected/remote_txn.out
```diff
@@ -553,14 +735,40 @@ NOTICE:  [loopback]:  PREPARE prep_1 AS SELECT 1
  
 (1 row)
 
+        --connection in transaction state
```
cevian, line 738: In this test case I would add some data modification commands before `SAVEPOINT save_1;` and make sure those are visible after the commit. That's a much more surefire correctness check.

tsl/src/remote/async.c
```diff
@@ -473,16 +562,18 @@ get_single_response_nonblocking(AsyncRequestSet *set)
 		PGconn *pg_conn = remote_connection_get_pg_conn(req->conn);
 
 		if (remote_connection_is_processing(req->conn) && req->state == DEFERRED)
-			elog(ERROR,
-				 "can't get a response because previous request hasn't completed. Aborted SQL "
-				 "statement: %s",
-				 req->sql);
+			return async_response_error_create("request already in process");
```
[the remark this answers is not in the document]
erimatnor: I didn't feel like the previous message followed PG error style ("keeping the primary message short helps keep it to the point") and this really shouldn't happen so seems like a corner-case. I think the most important thing is ability to find the place in the code where it happens, if it would ever occur. I think it is unique enough for that.

tsl/src/remote/connection_cache.c
```diff
@@ -156,49 +201,245 @@ connection_cache_create()
...
+	if (cacheid != FOREIGNSERVEROID)
+		return;
+
+	hash_seq_init(&scan, connection_cache->htab);
+
+	while ((entry = hash_seq_search(&scan)) != NULL)
+	{
+		/* hashvalue == 0 means cache reset, so invalidate entire cache */
+		if (hashvalue == 0 || entry->hashvalue == hashvalue)
+			entry->invalidated = true;
```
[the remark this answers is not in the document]
erimatnor: Yes, the upside is that we can easily examine the cache at all times and we don't have a number of dangling caches.

tsl/src/remote/connection_cache.c
```diff
@@ -156,49 +201,245 @@ connection_cache_create()
...
+ * This function is called when a database is dropped on the local instance
+ * and is needed to prevent errors when the database serves as a
+ * "same-instance" data node (common in tests). When a data node exists on the
+ * local instance (i.e., in a separate local database), we need to close all
+ * connections to the data node database in order to drop it. Otherwise, the
+ * drop database will fail since the database is being used by the connections
+ * in the cache.
+ */
+void
+remote_connection_cache_dropped_db_callback(const char *dbname)
```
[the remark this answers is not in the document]
erimatnor: I thought about that, but figured it is unnecessarily hard-handed. Since this isn't critical, a best effort seems reasonable.

tsl/src/remote/async.c
```diff
@@ -570,8 +659,6 @@ wait_to_consume_data(AsyncRequestSet *set, int elevel, TimestampTz end_time)
 
 		if (rc == 0)
```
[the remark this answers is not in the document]
erimatnor: Yes, this is a conscious decision, and, actually, in most cases it previously passed in elevel=WARNING so it didn't throw previously either.

I want to get away from the inconsistent error handling, sometimes returning error, sometimes throwing, sometimes one can pass in elevel (although event that wasn't adhered to in all places so passing in WARNING didn't guarantee not throwing).

I will take an extra  look to ensure that all places handles this properly.

tsl/src/remote/txn.c
```diff
@@ -91,19 +94,24 @@ remote_txn_begin(RemoteTxn *entry, int curlevel)
 			sql = "START TRANSACTION ISOLATION LEVEL SERIALIZABLE";
 		else
 			sql = "START TRANSACTION ISOLATION LEVEL REPEATABLE READ";
+
+		remote_connection_xact_transition_begin(entry->conn);
```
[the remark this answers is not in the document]
erimatnor: The functionality is not equivalent. Looking at the original postgres_fdw, on which our code is base, it seems functionality that was originally there was removed.

The deferred error thing only covered errors in subtransactions (actually only used in one single place, AFAIK). The guards (re-)introduced in this PR actually marks whether every command that involves a transaction state change has fully completed. If that's the case at the end of the transaction, the connection is OK to keep in the cache, otherwise it has to be closed.

This means that connections can be kept, as long as errors don't happen in any transaction state change command (around which the guards exist). The old code simply threw away the complete cache on every rollback irrespective of whether this was a user-initiated rollback or a rollback as a result of an error (e.g., unique constraint violation).

PR #1978, opened 2020-06-08
Cleanup remote_txn on data node delete

pmwkaa:
Remove any 2PC transaction records associated with the data node
from _timescaledb_catalog.remote_txn on delete_data_node() call.


tsl/test/sql/remote_txn.sql
```diff
@@ -325,3 +325,13 @@ SELECT count(*) FROM pg_prepared_xacts;
...
+
+-- test remote_txn cleanup on data node delete
+--
+SELECT count(*) FROM _timescaledb_catalog.remote_txn;
+
+SELECT count(*) FROM _timescaledb_catalog.remote_txn WHERE data_node_name = 'loopback';
+SELECT * FROM delete_data_node('loopback');
+
+SELECT * FROM delete_data_node('loopback2');
+SELECT count(*) FROM _timescaledb_catalog.remote_txn WHERE data_node_name = 'loopback2';
```
WireBaron, line 337: You should use the same query as above (`loopback` instead of `loopback2`) to demonstrate that something actually changed.

pmwkaa: Not sure I am following. I've updated the test, I guess it should be better now?

PR #1990, opened 2020-06-16
Remove flakiness of cagg materializaion test

k-rus:
Replaces now() with mocked timestamp value in
continuous_aggs_materialize test, so the test doesn't produce
unexpected result when now() is close to a chunk border.

### DIFF
--- a/tsl/test/expected/continuous_aggs_materialize.out
+++ b/tsl/test/expected/continuous_aggs_materialize.out
@@ -1436,9 +1436,9 @@ SELECT view_name, refresh_lag, max_interval_per_job
 (4 rows)
 
 -- test timezone is respected when materializing cagg with TIMESTAMP time column
-RESET timescaledb.current_timestamp_mock;
 RESET client_min_messages;
 SET SESSION timezone TO 'GMT+5';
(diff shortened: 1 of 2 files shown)


tsl/test/expected/continuous_aggs_materialize.out
```diff
@@ -1436,17 +1436,20 @@ SELECT view_name, refresh_lag, max_interval_per_job
...
+\set NOW_TEST 'timestamptz \'2019-09-09 01:01:01 UTC\''
 CREATE TABLE timezone_test(time timestamp NOT NULL);
 SELECT table_name FROM create_hypertable('timezone_test','time');
   table_name   
 ---------------
  timezone_test
 (1 row)
 
-INSERT INTO timezone_test VALUES (now() - '30m'::interval), (now()), (now() + '30m'::interval);
+INSERT INTO timezone_test VALUES
```
gayyappan, line 1449: cc: @svenklemm I think you had a concern that this approach might break time zone testing.

k-rus:
Fixes timescale/timescaledb-private#778

PR #1991, opened 2020-06-16
Fix paths to PG binaries in GH actions

k-rus:
Uses Windows notation for paths to PostgreSQL executables, so they can
be found during GitHub Actions workflow execution.

### DIFF
--- a/.github/workflows/windows-build-and-test.yaml
+++ b/.github/workflows/windows-build-and-test.yaml
@@ -48,6 +48,7 @@ jobs:
       run: cmake --build ${{ matrix.build_type }} --config ${{ matrix.build_type }}
     - name: Install [${{ matrix.build_type }}]
       run: cmake --install ${{ matrix.build_type }} --config ${{ matrix.build_type }}
+    - run: ls $HOME\PostgreSQL
     - name: Create DB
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-build-and-test.yaml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - prerelease_test
   pull_request:
+defaults:
+  run:
+    shell: cmd
```
erimatnor, line 10: Do we really want to use `cmd` instead of Powershell? It seems to have many drawbacks, like no good fail-fast mode.

k-rus: I don't have strong opinion of which shell it is better to use. I don't see we use anything special for PowerShell, but I might miss it. I haven't found what is the default in the docs, so  I am experimenting to see how to go ever the failure.

k-rus:
I have created another draft PR, thus this one is closed. But there is still nothing to review.

.github/workflows/windows-build-and-test.yaml
k-rus: I found that the default is powershell.

PR #1994, opened 2020-06-17
Force installing PostgreSQL in GH Actions on Windows

k-rus:
PostgreSQL 12 is preinstalled, while 11 is not. To unify different 
paths of PG 11 and 12 binaries, this commit implements workaround by 
forcing installation of PostgreSQL 12, so it is in the same path as 
PostgreSQL 11.


k-rus:
I believe it's fine to have this workaround, since the current GH Actions workflow for Windows is temporal. Replacing the actual test in Appveyor should also fix the current issues with GH Actions on Windows.

PR #1995, opened 2020-06-17
Fix ANALYZE on replicated distributed hypertable

erimatnor:
With replicated chunks, the function to import column stats would
experience errors when updating `pg_statistics`, since it tried to
write identical stats from several replica chunks.

This change fixes this issue by filtering duplicate stats rows
received from data nodes.

In the future, this could be improved by only requesting stats from
"primary" chunks on each data node, thus avoiding duplicates without
having to filter the result. However, this would complicate the
function interface as it would require sending a list of chunks
instead of just getting the stats for all chunks in a hypertable.

### DIFF
--- a/tsl/src/chunk_api.c
+++ b/tsl/src/chunk_api.c
@@ -958,8 +958,8 @@ chunk_update_colstats(Chunk *chunk, int16 attnum, float nullfract, int32 width, 
 	if (HeapTupleIsValid(oldtup))
 	{
 		stup = heap_modify_tuple(oldtup, RelationGetDescr(sd), values, nulls, replaces);
+		CatalogTupleUpdate(sd, &oldtup->t_self, stup);
 		ReleaseSysCache(oldtup);
-		CatalogTupleUpdate(sd, &stup->t_self, stup);
 	}
 	else
 	{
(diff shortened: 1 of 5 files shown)


erimatnor:
@pmwkaa Interesting thought. Still, I am not sure how to generally make dist. commands chunk aware? I guess at this point there is no notion of chunks when executing commands on remote nodes, so it is a bit unclear how this should work for the general case.

erimatnor:
@mkindahl A chunk always has a "primary" data node, which is the one we use when executing a regular query (and thus the server set in the foreign table). Therefore, this notion of "primary" already exists. Still, which data node we pick for a certain chunk is not really important (i.e., it need to be the "primary") as long as we pick only one per chunk.

tsl/src/chunk_api.c
```diff
@@ -1187,9 +1260,12 @@ fetch_remote_chunk_stats(Hypertable *ht, FunctionCallInfo fcinfo, bool col_stats
 						"that cannot accept type record")));
 
 	cmdres = ts_dist_cmd_invoke_func_call_on_all_data_nodes(fcinfo);
+
 	/* Expect TEXT response format since dist command API currently defaults
 	 * to requesting TEXT */
 	tf = tuplefactory_create_for_tupdesc(tupdesc, true);
+	num_rows = ts_dist_cmd_total_row_count(cmdres);
+	stats_process_context_init(&statsctx, num_rows / ht->fd.replication_factor);
```
gayyappan, line 1268: There is a potential rounding error here due to division. You should use ceil(num_row/ht->fd_replication_factor). Or optimize by computing ceil(m/n) as (m+n-1)/n
The other option is we know exactly how many rows are expected. For relstats: 1 for each chunk, so this is equal to num_chunks . For colstats, it is num_chunks * num_cols?Is there a reason why we are not initializing using this approach?

erimatnor: Not sure rounding errors matters much and this is just to set the initial size of the hash table. It will automatically expand if necessary. If anything, I'd make the initial size a bit larger than necessary to avoid expanding the hash table.

I'll look into your second suggestion.

gayyappan: Didn't know the table expanded if there are more elements. It is fine in that case.

PR #1996, opened 2020-06-17
Add mac os regression test to GitHub Actions

svenklemm:
Some checks are commented out atm to make debugging quicker will reenable those before pushing

### DIFF
--- a/.github/workflows/linux-build-and-test.yaml
+++ b/.github/workflows/linux-build-and-test.yaml
@@ -1,4 +1,4 @@
-name: Regression Linux
+name: Regression
 on:
   schedule:
     # run daily 20:00 on master branch
@@ -19,8 +19,8 @@ jobs:
     - id: set-matrix
       run: python scripts/gh_matrix_builder.py ${{ github.event_name }}
 
-  regress_linux:
-    name: PG${{ matrix.pg }} ${{ matrix.name }}
+  regress:
+    name: PG${{ matrix.pg }} ${{ matrix.name }} ${{ matrix.os }}
     needs: matrixbuilder
     runs-on: ${{ matrix.os }}
     strategy:
@@ -32,9 +32,11 @@ jobs:
       MAKE_JOBS: 6
       CLANG: ${{ matrix.clang }}
       CC: ${{ matrix.cc }}
+      CXX: ${{ matrix.cxx }}
 
     steps:
     - name: Install Dependencies
+      if: runner.os == 'Linux'
       run: sudo apt install flex bison lcov systemd-coredump gdb ${{ matrix.extra_packages }}
 
     # we cache the build directory instead of the install directory here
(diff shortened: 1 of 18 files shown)


.github/workflows/linux-build-and-test.yaml
```diff
@@ -29,12 +29,15 @@ jobs:
     env:
       PG_SRC_DIR: pgbuild
       PG_INSTALL_DIR: postgresql
+      OPENSSL_ROOT_DIR: /usr/local/opt/openssl
```
pmwkaa, line 32: Is FindOpenSSL.cmake could not find this path by default?

svenklemm: yes cmake won't find openssl without this

erimatnor: Is it possible to add hints or something to the FindOpenSSL module?

scripts/gh_matrix_builder.py
```diff
@@ -72,8 +78,22 @@ def build_apache_config(overrides):
...
 
 # always test debug build on latest pg11 and latest pg12
-m["include"].append(build_debug_config({"pg":"11.8"}))
-m["include"].append(build_debug_config({"pg":"12.3"}))
+m["include"].append(build_debug_config({"pg":PG11_LATEST}))
+m["include"].append(build_debug_config({"pg":PG12_LATEST}))
+
+macos_config = {
+  "pg": PG12_LATEST,
+  "os": "macos-latest",
```
erimatnor, line 86: Maybe use explicit version here, like ubuntu? Otherwise things might break for us when they bump latest.

test/isolation/specs/deadlock_dropchunks_select.spec
```diff
@@ -16,12 +16,12 @@ teardown
 { DROP TABLE DT; DROP table ST; DROP table SL; }
 
 session "s1"
-setup	{ BEGIN; SET TRANSACTION ISOLATION LEVEL READ COMMITTED; SET LOCAL lock_timeout = '50ms'; SET LOCAL deadlock_timeout = '10ms'; }
+setup	{ BEGIN; SET TRANSACTION ISOLATION LEVEL READ COMMITTED; SET LOCAL lock_timeout = '500ms'; SET LOCAL deadlock_timeout = '10ms'; }
```
erimatnor, line 19: FWIW, this timeout seems like something we'd like to set globally, per test, or  maybe even in postgresql.conf. 

Not strictly necessary to fix in this PR, but if you're touching all these files, why not just make it a variable or global default instead?

.github/workflows/linux-build-and-test.yaml
svenklemm: OPENSSL_ROOT_DIR is the way to pass hints but i moved it from env to argument

test/isolation/specs/deadlock_dropchunks_select.spec
svenklemm: this is specific to isolation tests so it certainly cannot be a global also since the isolation tests dont use psql they dont have variable support so that would require changing all the tests to template files and let cmake generate them, so definitely a task for a separate PR

Issue #1912, opened 2020-05-18
drop_chunks works even with default_transaction_read_only=on


tylerfontaine:
**Relevant system information:**
 - PostgreSQL 12.2 on x86_64-pc-linux-gnu, compiled by gcc, a 67f83fccf p a91f14b60b, 64-bit
 - TimescaleDB version (output of `\dx` in `psql`): 1.7.0
 - Installation method: [e.g., "using Docker", "apt install", "source"]: Yum

**Describe the bug**
calling drop_chunks() on a hypertable successfully drops chunks even with transactions set to read_only.
It's worth noting it fails as expected when `cascade_to_materializations` is set to true, and there are materializations to cascade to.

**To Reproduce**
1. create a hypertable and add some data.
2. Create a continuous agg on that hypertable
2. set transactions to read only `set default_transaction_read_only=on;`
3. Call drop_chunks:
```
SELECT count(*)
FROM drop_chunks(INTERVAL '4 days', 'some_hypertable', cascade_to_materializations => TRUE);
```
It will fail: 
```
ERROR:  cannot execute DELETE in a read-only transaction
CONTEXT:  SQL statement "DELETE FROM _timescaledb_internal._materialized_hypertable_5 AS D WHERE D.chunk_id = $1"
```
Set cascade_to_materializations to FALSE
```
SELECT count(*)
FROM drop_chunks(INTERVAL '4 days', 'some_hypertable', cascade_to_materializations => FALSE);
```
and it will succeed.

This ALSO succeeds regardless of the cascade setting if the hypertable does not have any materializations attached to it.

**Expected behavior**
I would expect read_only transactions to cause drop_chunks to fail.

**Actual behavior**
Chunks were actually dropped if no CASCADE action happened.

k-rus:
@pmwkaa Have you fixed it?

svenklemm:
This was fixed by #1997

PR #1997, opened 2020-06-17
Add checks for read-only transactions

pmwkaa:
This change ensures that API functions which modify data respects read-only transaction state set by
`default_transaction_read_only` option.

Related to https://github.com/timescale/timescaledb/issues/1912

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -266,7 +266,7 @@ block_on_foreign_server(const char *const server_name)
 	return false;
 }
 
-static void
+static bool
 process_create_foreign_server_start(ProcessUtilityArgs *args)
 {
 	CreateForeignServerStmt *stmt = (CreateForeignServerStmt *) args->parsetree;
@@ -277,6 +277,8 @@ process_create_foreign_server_start(ProcessUtilityArgs *args)
 				 errmsg("operation not supported for a TimescaleDB data node"),
 				 errhint("Use add_data_node() to add data nodes to a "
 						 "TimescaleDB distributed database.")));
+
+	return false;
 }
 
 static void
@@ -298,7 +300,7 @@ process_drop_foreign_server_start(DropStmt *stmt)
 	}
 }
 
-static void
+static bool
 process_create_foreign_table_start(ProcessUtilityArgs *args)
 {
 	CreateForeignTableStmt *stmt = (CreateForeignTableStmt *) args->parsetree;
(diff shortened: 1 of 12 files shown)


tsl/src/data_node.c
```diff
@@ -638,6 +638,8 @@ data_node_add_internal(PG_FUNCTION_ARGS, bool set_distid)
 	bool PG_USED_FOR_ASSERTS_ONLY result;
 	DbInfo database;
 
+	PreventCommandIfReadOnly("add_data_node()");
```
erimatnor, line 641: Isn't it better to put these blocks for TSL in the outer-wrapper functions, (i.e., in `cross_module_fn.c`)? That would make it easier to verify that all mutation functions have this block. We could also move the read_only test to non-TSL in that case.

pmwkaa: I had this idea, but thought it would be odd that apache part would do those checks explicitly. Plus I did not like the idea to separate this logic from actual implementation, because there could be more of them, like preventing running inside transaction block, etc. Otherwise we should separate all of our API functions into separate file and to those checks consistently there

erimatnor: Ok, makes sense. I am on the fence myself, but figured it was good to raise the idea to get some pros and cons.

Happy to leave as is.

pmwkaa:
List of changes:

* Added support for DDL read-only operations checks with tests
* Slightly refactored process utility functions in order to make them have consistent definition and be able to distinct operations which needs read-only checks

PostgreSQL has function `check_xact_readonly(Node *parsetree)`, unfortunately it is not exported and cannot be used directly.

src/process_utility.c
```diff
@@ -277,6 +277,8 @@ process_create_foreign_server_start(ProcessUtilityArgs *args)
 				 errmsg("operation not supported for a TimescaleDB data node"),
 				 errhint("Use add_data_node() to add data nodes to a "
 						 "TimescaleDB distributed database.")));
+
+	return false;
```
erimatnor, line 281: I am confused by the change in return value here and in many functions below. Most or all of them just seem to always return `false` instead of `void` (nothing), and it is unclear what `false` semantically means.

I would at suggest to at least create an `enum` for the return value (e.g., DDL_DONE, DDL_CONTINUE) or similar. This would make it a lot more clear what the intention of the return value is.

src/process_utility.h
```diff
@@ -27,6 +27,8 @@ typedef struct ProcessUtilityArgs
 	char *completion_tag;
 } ProcessUtilityArgs;
 
+typedef bool (*ts_process_utility_handler_t)(ProcessUtilityArgs *args);
```
erimatnor, line 30:
```suggestion
typedef enum DDLResult
{
   DDL_CONTINUE,
   DDL_DONE
} DDLResult;

typedef DDLResult (*ts_process_utility_handler_t)(ProcessUtilityArgs *args);
```

src/process_utility.c
pmwkaa: Agree, makes sense. I am also a bit confused, but since we have most functions defined this way I expected to make them equal in definition. I would be happy to change it in another PR if it is ok

PR #1998, opened 2020-06-18
Preserve pg_stats on chunks before compression

WireBaron:
This change will ensure that the pg_statistics on a chunk are
updated immediately prior to compression. It also ensures that
these stats are not overwritten as part of a global or hypertable
targetted ANALYZE.

This addresses the issue that chunk will no longer generate valid
statistics durings an ANALYZE once the data's been moved to the
compressed table. Unfortunately any compressed rows will not be
captured in the parent hypertable's pg_statistics as there is no way
to change how PostGresQL samples child tables in PG11.

This approach assumes that the compressed table remains static, which
is mostly correct in the current implementation (though it is
possible to remove compressed segments). Once we start allowing more
operations on compressed chunks this solution will need to be
revisited. Note that in PG12 an approach leveraging table access
methods will not have a problem analyzing compressed tables.


tsl/src/compression/compress_utils.c
```diff
@@ -218,6 +250,9 @@ compress_chunk_impl(Oid hypertable_relid, Oid chunk_relid)
 	hcache = ts_hypertable_cache_pin();
 	compresschunkcxt_init(&cxt, hcache, hypertable_relid, chunk_relid);
 
+	/* Perform an analyze on the chunk to get up-to-date stats before compressing */
+	preserve_uncompressed_chunk_stats(chunk_relid);
```
gayyappan, line 254: Why do we need to preserve old stats? How is this useful? We need the compressed chunk stats anyway to make the right decisions for planning. (cc @svenklemm  Your thoughts?)

tsl/src/compression/compress_utils.c
```diff
@@ -203,6 +204,37 @@ compresschunkcxt_init(CompressChunkCxt *cxt, Cache *hcache, Oid hypertable_relid
...
+#else
+	vs.options = VACOPT_ANALYZE;
+	ExecVacuum(&vs, true);
+#endif
+
+	/* Now make sure that the chunk doesn't get accidentally analyzed */
+	at_cmd = makeNode(AlterTableCmd);
+	at_cmd->subtype = AT_SetRelOptions;
+	at_cmd->def =
+		(Node *) list_make1(makeDefElem("autovacuum_enabled", (Node *) makeString("false"), -1));
```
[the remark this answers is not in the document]
WireBaron: While users can theoretically toggle this on individual chunks, it seems like such a bizarre corner case that I'm reluctant to say we should go to any lengths to support this.  I do think that on decompression we should toggle this back to whatever the setting for the hypertable is though.

tsl/src/compression/compress_utils.c
```diff
@@ -203,6 +204,37 @@ compresschunkcxt_init(CompressChunkCxt *cxt, Cache *hcache, Oid hypertable_relid
...
+	AlterTableInternal(chunk_relid, list_make1(at_cmd), false);
```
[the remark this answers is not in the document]
WireBaron: Nice, looks much tighter.

tsl/src/compression/compress_utils.c
WireBaron: The pg_statistics give us things like most common value and histograms which aren't available through the compressed chunk stats.  While perhaps not as useful as stats such as live_tuples, these are still useful for choosing optimal plans in some cases.

WireBaron:
> Needs test for pg_stat_tables stuff especially n_live_tuples

Unfortunately these stat tables are updated asynchronously by the stats collector, and the update seems to lag the inserts by over a minute.  I don't see any way to force this to update more quickly, and this seems far to long to have a test wait.

tsl/src/compression/compress_utils.c
```diff
@@ -208,6 +209,66 @@ compresschunkcxt_init(CompressChunkCxt *cxt, Cache *hcache, Oid hypertable_relid
...
+#if PG12_GE
+	ExecVacuum(NULL, &vs, true);
+#else
+	ExecVacuum(&vs, true);
+#endif
+	AlterTableInternal(chunk_relid, list_make1(&at_cmd), false);
+}
+
+/* This function is intended to undo the disabling of autovacuum done when we compressed a chunk.
+ * Note that we do not cache the previous value for this (as we don't expect users to toggle this
```
erimatnor, line 247: I guess this is OK, but just want to point out the this will potentially revert changes that someone has done to the vacuum setting on an individual chunk.

tsl/test/isolation/expected/compression_ddl.out
```diff
@@ -132,7 +132,7 @@ step A2: COMMIT;
 step C1: <... completed>
 compress_chunk 
 
-_timescaledb_internal._hyper_7_13_chunk
+_timescaledb_internal._hyper_7_12_chunk
```
svenklemm, line 135: why did chunk numbering in the isolation tests change?

WireBaron: I believe this is a result of changing the locking the source chunk in compress_chunk_impl from a AccessShareLock to a ShareLock.  This causes it to block before it creates the chunk to compress into in the previous permutation, where the compress ends up being rolled back due to the conflicting `ALTER`.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -362,10 +363,10 @@ ts_decompress_chunk_generate_paths(PlannerInfo *root, RelOptInfo *chunk_rel, Hyp
 	/* translate chunk_rel->baserestrictinfo */
 	pushdown_quals(root, chunk_rel, compressed_rel, info->hypertable_compression_info);
 	set_baserel_size_estimates(root, compressed_rel);
-	new_row_estimate = compressed_rel->rows * DECOMPRESS_CHUNK_BATCH_SIZE;
-	/* adjust the parent's estimate by the diff of new and old estimate */
-	hypertable_rel->rows += (new_row_estimate - chunk_rel->rows);
-	chunk_rel->rows = new_row_estimate;
+
+	/* use the cost estimate from the chunk_rel, as that has better stats */
+	compressed_rel->baserestrictcost = chunk_rel->baserestrictcost;
```
svenklemm, line 368: i dont think thats correct as those expressions can differ quite a bit

tsl/test/shared/expected/ordered_append_join.out
```diff
@@ -2549,32 +2549,31 @@ FROM :TEST_TABLE o1
     AND o1.time = o2.time
   ORDER BY o1.time
   LIMIT 100;
```
svenklemm, line 2551: hmm this plan is wrong and doesnt match the expected output

WireBaron:
I've removed the second commit which modified the estimates on the DecompressChunk node as there seems to be cases where the new estimates are worse than the old ones.  I've created a new issue (#2239) solely focused on improving these estimates which can be worked on once the rest of this PR is merged.

svenklemm (approved):

PR #2007, opened 2020-06-22
Run regresscheck-shared tests in parallel

svenklemm:
This patch changes regresscheck-shared to no longer run individual
tests serially but runs them in parallel instead.


tsl/test/shared/sql/CMakeLists.txt
```diff
@@ -25,7 +25,19 @@ if (TEST_FILES_SHARED)
...
+
+set(GROUP_MEMBERS 0)
 foreach(TEST_FILE ${TEST_FILES_SHARED})
   string(REGEX REPLACE "(.+)\.sql" "\\1" TESTS_TO_RUN ${TEST_FILE})
-  file(APPEND ${TEST_SCHEDULE_SHARED} "test: ${TESTS_TO_RUN}\n")
+  if(GROUP_MEMBERS EQUAL 0)
+    file(APPEND ${TEST_SCHEDULE_SHARED} "\ntest: ")
+  endif()
+  file(APPEND ${TEST_SCHEDULE_SHARED} "${TESTS_TO_RUN} ")
+  MATH(EXPR GROUP_MEMBERS "(${GROUP_MEMBERS}+1)%${PARALLEL_GROUP_SIZE}")
```
gayyappan, line 41: Could you explain why we need to set GROUP_MEMBERS? How is this used?

svenklemm: this is just an internal counter for the current group, the actual parallel stuff is not implemented here but inside pg_regress, we just generate a schedule that makes pg_regress run tests in parallel

PR #2019, opened 2020-06-24
Run ARM64 test as github action

svenklemm:
This workflow could also be used to run tests i386, ARM32 and PPC but
unfortunately the docker image name can't be made a variable so
it has to be changed manually when other architectures should be
tested.


.github/workflows/docker-arm-build-and-test.yaml
```diff
@@ -0,0 +1,74 @@
...
+            yes | /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh &&
+            DEBIAN_FRONTEND=noninteractive apt install -y postgresql-${{ matrix.pg }} postgresql-server-dev-${{ matrix.pg }} cmake sudo libkrb5-dev
+            ./bootstrap -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DREGRESS_CHECKS=OFF &&
+            make -j $MAKE_JOBS -C build &&
+            make -C build install &&
+            chown -R postgres:postgres . &&
+            sudo -u postgres make -C build -k installcheck
+          "
+
+    - name: Show regression diffs
```
pmwkaa, line 45: I guess we are not going to show backtraces here as well?

svenklemm: i did get an error with coredumpctl so i didnt include it in initial PR but will investigate if its easy to add

PR #2020, opened 2020-06-24
Move update test to GitHub actions

svenklemm:


.github/workflows/update-test.yaml
```diff
@@ -0,0 +1,41 @@
...
+            pg_major: 12
+      fail-fast: false
+    env:
+      PG_VERSION: ${{ matrix.pg }}
+    steps:
+    - name: Checkout TimescaleDB
+      uses: actions/checkout@v2
+
+    - name: Update tests ${{ matrix.pg }}
+      run: sudo ./scripts/test_updates_pg${{ matrix.pg_major}}.sh
```
erimatnor, line 30: Why do we need to run these as `root` now? I can't see that Travis did this and I have never had to do it when running locally.

I think, generally, that we should avoid running tests with elevated privileges because it might mask issues that might otherwise occur.

.github/workflows/update-test.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Slack Notification
+      if: failure() && github.event_name != 'pull_request'
+      env:
+        SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+        SLACK_COLOR: '#ff0000'
+        SLACK_USERNAME: GitHub Action
+        SLACK_TITLE: Update test PG${{ matrix.pg }} ${{ job.status }}
+        SLACK_MESSAGE: ${{ github.event.head_commit.message }}
+      uses: rtCamp/action-slack-notify@v2.0.2
+
```
erimatnor, line 41: Don't we have some test log output that we can upload as an artifact in case of failure?

PR #2024, opened 2020-06-25
Add status badge for Linux i386 build

k-rus:


k-rus:
> Seems to have syntax error cause it doesnt render properly

Thank you.  I noticed it and working on this.

k-rus:
> Seems to have syntax error cause it doesnt render properly

Fixed it.

PR #2025, opened 2020-06-25
Disable fsync in test config files

pmwkaa:
fsync is enabled by default in PostgreSQL, there is no good reason to keep it enabled for our tests since it usually increases execution time

### DIFF
--- a/test/postgresql.conf.in
+++ b/test/postgresql.conf.in
@@ -17,3 +17,4 @@ log_line_prefix='%u [%p] %d '
 extra_float_digits=0
 timescaledb.passfile='@TEST_PASSFILE@'
 hba_file='@TEST_PG_HBA_FILE@'
+fsync=off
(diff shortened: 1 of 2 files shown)


pmwkaa:
> Don't think we should disable fsync, the preferred/recommended tunable for this situation is `synchronous_commit`

I think this is mostly true for production environments

pmwkaa:
Apparently PostgreSQL regression test tool already sets `fsync=off` by default, thats why I haven't seen much improvement on my machine.

```diff
> +SHOW fsync;
> + fsync 
> +-------
> + off
> +(1 row)
> +
> +SHOW synchronous_commit;
> + synchronous_commit 
> +--------------------
> + on
> +(1 row)
> +
```

Should we set `synchronous_commit=off` as well?

gayyappan:
> Should we set `synchronous_commit=off` as well?
 I don't see a good reason to turn this off.
The default would be the common setting and we should use that code path.

svenklemm:
also seems to make no difference performance wise on my local system, probably no gain if fsync is already off

erimatnor:
@pmwkaa what's the status of this PR. Seems it doesn't make much of a difference and fsync was already off. Should we close this?

pmwkaa:
Closing it, because the PostgreSQL regression test suite already runs with disabled fsync.

Issue #2001, opened 2020-06-18
Fix: Do not forward relcache invalidation forwarded to syscache invalidation


erimatnor:
In `src/cache_invalidate.c`, a general relcache invalidation (`relid == InvalidOid`) triggers a full cache invalidation, including invalidation of TSL syscache callback.

It's wrong to forward relcache invalidations as syscache invalidations, in particular cacheid=0, which doesn't seem like a possible thing for the syscache.

For instance, a relcache invalidation for an `InvalidOid` relid, triggers an invalidation of the connection cache with cacheid=0, because the connection cache listens to invalidation events for `pg_foreign_server`. But such an event with cacheid=0 is not possible; for instance, the `postgres_fdw` code Asserts that `cacheid==FOREIGNSERVEROID`, which would crash if cacheid=0 was a possible event.

erimatnor:
Closed by https://github.com/timescale/timescaledb/pull/2026

PR #2026, opened 2020-06-25
Do not forward relcache invalidation as syscache invalidation

pmwkaa:
It is incorrect to forward relcache invalidations as syscache invalidations, like `cacheid = InvalidOid` which is not possible condition at the moment. Allow syscache invalidations only for `FOREIGNSERVEROID`s.

Related to: http://github.com/timescale/timescaledb/issues/2001

### DIFF
--- a/src/cache_invalidate.c
+++ b/src/cache_invalidate.c
@@ -58,19 +58,6 @@ cache_invalidate_relcache_all(void)
 	ts_bgw_job_cache_invalidate_callback();
 }
 
-static inline void
-cache_invalidate_syscache_all(void)
-{
-	ts_cm_functions->cache_syscache_invalidate(PointerGetDatum(NULL), 0, 0);
-}
-
-static inline void
-cache_invalidate_all()
-{
-	cache_invalidate_relcache_all();
-	cache_invalidate_syscache_all();
-}
-
 /*
  * This function is called when any relcache is invalidated.
  * Should route the invalidation to the correct cache.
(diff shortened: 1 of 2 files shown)


src/cache_invalidate.c
```diff
@@ -58,19 +58,6 @@ cache_invalidate_relcache_all(void)
 	ts_bgw_job_cache_invalidate_callback();
 }
 
-static inline void
-cache_invalidate_syscache_all(void)
-{
-	ts_cm_functions->cache_syscache_invalidate(PointerGetDatum(NULL), 0, 0);
```
gayyappan, line 64: Do we need  ts_cm_functions-> cache_syscache_invalidate anymore?

pmwkaa: Yes, it is still used by the `cache_invalidate_syscache_callback`

src/cache_invalidate.c
```diff
@@ -142,7 +129,7 @@ cache_invalidate_xact_end(XactEvent event, void *arg)
 			 * change since the transaction hasn't been committed and other
 			 * backends cannot have the invalid state.
 			 */
-			cache_invalidate_all();
+			cache_invalidate_relcache_all();
```
erimatnor, line 132: One question is what happens when you, in one transaction, add a foreign server and open a connection to it (thus adding it to the cache), followed by an abort of the transaction. Will the connection remain in the cache or will it receive a syscache invalidation event?

svenklemm: Might be good to have isolation test for this

pmwkaa: This situation is not possible, since we are not allow to execute `add_data_node()` as part of a multi-statement transaction.

PR #2027, opened 2020-06-25
Execute regresscheck-t tests in parallel

svenklemm:
This patch adds parallel execution capabilities to regresscheck-t.


svenklemm:
Quite a few of the blacklisted tests can probably be made to run in parallel with minimal adjustments but thats outside of the scope of this PR

PR #2032, opened 2020-06-28
Add testing for released apt packages

svenklemm:
This workflow will install our apt package and then try to enable
timescaledb in the database and also check the version installed
from the package against the expected version.

### DIFF
--- a/.github/workflows/apt-packages.yaml
+++ b/.github/workflows/apt-packages.yaml
@@ -1,0 +1,89 @@
+name: APT packages
+on:
+  schedule:
+    # run daily 0:00 on master branch
+    - cron: '0 0 * * *'
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-test.yaml
```diff
@@ -0,0 +1,77 @@
+name: Release-Test
```
erimatnor, line 1: Can we make the name a bit less generic? E.g., `Debian packages` or something? (I guess "Test" is optional since obvious).

This applies also to the workflow file name

svenklemm: oh i was gonna add rpm package tests to the same workflow and wanted it to trigger when pushing to release_test and when tagging branches

PR #2033, opened 2020-06-28
Remove unused functions

svenklemm:
Remove functions that are no longer used due to refactorings.
Removes the following functions:
- hypertable_tuple_match_name
- ts_hypertable_get_all_by_name
- ts_hypertable_has_tuples

### DIFF
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -1381,24 +1381,6 @@ table_has_replica_identity(Relation rel)
 static bool inline table_has_rules(Relation rel) { return rel->rd_rules != NULL; }
 
 bool
-ts_hypertable_has_tuples(Oid table_relid, LOCKMODE lockmode)
-{
-	ListCell *lc;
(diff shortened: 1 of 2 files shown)


tsl/src/chunk.c
```diff
@@ -175,92 +175,3 @@ chunk_set_default_data_node(PG_FUNCTION_ARGS)
 
 	PG_RETURN_BOOL(chunk_set_foreign_server(chunk, server));
 }
-
-/* Should match definition in ddl_api.sql */
```
gayyappan, line 179: How is drop_chunk executed on data node? I don't see references to this code, but do we need it for that case?

erimatnor: Just talked to @mkindahl and we agreed that we'd like to keep this function and submit a PR to reenable use of it.

svenklemm: ok i'll not remove it in this PR then

PR #2036, opened 2020-06-29
Upload coredumps as artifact

svenklemm:
This patch adds collecting coredumps to the regression workflow
in addition to the binary all shared libraries used by the coredump
are collected as well.

### DIFF
--- a/.github/workflows/alpine-32bit-build-and-test.yaml
+++ b/.github/workflows/alpine-32bit-build-and-test.yaml
@@ -52,7 +52,9 @@ jobs:
 
     - name: make installcheck
       id: installcheck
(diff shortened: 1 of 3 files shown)


src/time_bucket.c
```diff
@@ -180,6 +180,8 @@ ts_timestamptz_bucket(PG_FUNCTION_ARGS)
 	Interval *interval = PG_GETARG_INTERVAL_P(0);
 	TimestampTz timestamp = PG_GETARG_TIMESTAMPTZ(1);
 
+  abort();
```
gayyappan, line 183: Need to undo this before merge.

PR #159, opened 2017-08-15
Handle ALTER TABLE commands on hypertables

erimatnor:
Commands that alter hypertable partitioning columns need to update
associated hypertable metadata (such as dimension column name and
type) as well as related constraints on chunks. This change adds
support for this, previously missing or broken, ensuring that
metadata is in sync. In particular, dimension CHECK constraints
are now named in the chunk_constraint table, so that the
corresponding constraints in pg_constraint can be found and kept
in sync.

This change also blocks certain commands, like altering partitioning
dimension constraints on chunks.

PR #2046, opened 2020-07-01
Use index tablespace during chunk creation

### LINKED ISSUE #903: Tablespace for primary key constraint is not been updated by SET TABLESPACE
 - OS: Windows Server 2012 x64
 - PostgreSQL version (output of `postgres --version`): 9.6.9
 - TimescaleDB version (output of `\dx` in `psql`): 0.9.2
 - Installation method: Windows installer

**Describe the bug**
The hypertable was created with the wrong PK contraint tablespace, after realizing that, we fixed the tablespaces for all existing chunk indexes and on the hypertable, but new chunks are been created on the old tablespace.

**To Reproduce**
- Create a table with the primary key index on tablespace TABLESPACE_1
- Insert some data so that the initial chunks are created
- Alter the index tablespace for PK constraint using ALTER INDEX ... SET TABLESPACE TABLESPACE_2
- Insert more data in a way that new chunks got created
- New chunks are still been created on TABLESPACE_1 not on current one that is TABLESPACE_2

**Expected behavior**
After updating the tablespace for an Index, new chunks will be created in the current tablespace.

**Actual behavior**
After updating the tablespace for an Index, new chunks are created in the original tablespace.

k-rus:
If a tablespace is provided for an index on a hypertable, it will be
also used for the index on new chunks.

Fixes #903

### DIFF
--- a/sql/chunk_constraint.sql
+++ b/sql/chunk_constraint.sql
@@ -14,6 +14,8 @@ DECLARE
     constraint_oid OID;
     check_sql TEXT;
     def TEXT;
+    indx_tablespace NAME;
+    tablespace_def TEXT;
 BEGIN
     SELECT * INTO STRICT chunk_row FROM _timescaledb_catalog.chunk c WHERE c.id = chunk_constraint_row.chunk_id;
     SELECT * INTO STRICT hypertable_row FROM _timescaledb_catalog.hypertable h WHERE h.id = chunk_row.hypertable_id;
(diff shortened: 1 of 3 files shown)


sql/chunk_constraint.sql
```diff
@@ -29,7 +31,14 @@ BEGIN
         SELECT oid INTO STRICT constraint_oid FROM pg_constraint
         WHERE conname=chunk_constraint_row.hypertable_constraint_name AND
               conrelid = format('%I.%I', hypertable_row.schema_name, hypertable_row.table_name)::regclass::oid;
-        def := pg_get_constraintdef(constraint_oid);
+        SELECT tablespace INTO indx_tablespace FROM pg_indexes
+        WHERE indexname = chunk_constraint_row.hypertable_constraint_name;
```
erimatnor, line 35: I worry that assuming that the index and constraint names are the same isn't strictly correct and only works incidentally. I think the correct approach is to get the `pg_constraint` row for the constraint and get the index OID (`pg_constraint.coninid`) from there.

k-rus: I believe that constraint name and index name are always the same, but they are stored in different catalog tables. So might be difficult to guarantee.

So implemented the suggestion.

svenklemm:
This is probably a good candidate for refactoring later, not in this PR of course

k-rus:
> This is probably a good candidate for refactoring later, not in this PR of course

@svenklemm I agree. It is two totally different paths for copying indexes from hypertable to chunks. One is in SQL, which is affected in this PR, and another in C, which already takes care of table spaces. Do you mean to unify the two paths?

svenklemm:
yes i would drop the sql path

erimatnor:
> yes i would drop the sql path

FWIW, this upcall to SQL has been a long-standing annoyance for me. For a bit of historical context: I once did a PR for handling constraints in C code but got shot down since there were arguments being made that PL/pgSQL was a better language to write the extension in. This was way back when we had a C vs. PL/pgSQL debate and there were still arguments being made for writing most of the extension in PL/pgSQL and use only C when absolutely necessary. 

This is, I believe, the last remaining "upcall" since I got rid of all the other ones. The rejected PR with code that may still be relevant: https://github.com/timescale/timescaledb/pull/159

PR #2050, opened 2020-07-01
Add testing for RPM packages

svenklemm:
This workflow will install our rpm packages and then try to enable
timescaledb in the database and also check the version installed
from the package against the expected version.


svenklemm:
Test run: https://github.com/svenklemm/timescaledb/actions/runs/154606544

PR #2051, opened 2020-07-01
Fix tuple concurrently deleted error with multiple continuous aggregates

### LINKED ISSUE #1940: Multiple continuous aggregations for same hypertable crash with: ERROR:  tuple concurrently deleted
 - OS: Debian 4.9.144-3.1 (2019-02-19) x86_64 GNU/Linux
 - PostgreSQL: 12.3
 - TimescaleDB version:  1.7.1

**Describe the bug**
I have 5 continuous aggregates all using the same hypertable. Each continuous aggregate groups by a different parameter.
```
select * from timescaledb_information.continuous_aggregates;

-[ RECORD 2 ]------------------+-----------------------------------------------------------------------------------------------------------------------------
view_name                      | sessions_summary_daily
view_owner                     | postgres
refresh_lag                    | 00:00:00
refresh_interval               | 06:00:00
max_interval_per_job           | 20 days
ignore_invalidation_older_than |
materialized_only              | f
materialization_hypertable     | _timescaledb_internal._materialized_hypertable_4
view_definition                |  SELECT sessions.property_id,                                                                                               +
                               |     time_bucket('1 day'::interval, sessions."timestamp") AS bucket,                                                         +
                               |     count(*) AS visits                                                                                                      +
                               |    FROM sessions                                                                                                            +
                               |   GROUP BY sessions.property_id, (time_bucket('1 day'::interval, sessions."timestamp"));
-[ RECORD 3 ]------------------+-----------------------------------------------------------------------------------------------------------------------------
view_name                      | sessions_summary_daily_by_channel
view_owner                     | postgres
refresh_lag                    | 00:00:00
refresh_interval               | 06:00:00
max_interval_per_job           | 20 days
ignore_invalidation_older_than |
materialized_only              | f
materialization_hypertable     | _timescaledb_internal._materialized_hypertable_6
view_definition                |  SELECT sessions.property_id,                                                                                               +
                               |     sessions.channel,                                                                                                       +
                               |     time_bucket('1 day'::interval, sessions."timestamp") AS bucket,                                                         +
                               |     count(*) AS visits                                                                                                      +
                               |    FROM sessions                                                                                                            +
                               |   GROUP BY sessions.property_id, sessions.channel, (time_bucket('1 day'::interval, sessions."timestamp"));

etc...
```

These continuous aggregates are all stuck in a failing state, repeatedly failing and crashing and unable to complete.
```
select * from timescaledb_information.continuous_aggregate_stats;

-[ RECORD 2 ]----------+--------------------------------------
view_name              | sessions_summary_daily
completed_threshold    | 2020-05-27 00:00:00+00
invalidation_threshold | 2020-05-27 00:00:00+00
job_id                 | 1000
last_run_started_at    | 2020-05-27 07:22:30.271622+00
last_successful_finish | 2020-05-27 01:22:30.271376+00
last_run_status        |
job_status             | Running
last_run_duration      |
next_scheduled_run     | -infinity
total_runs             | 12
total_successes        | 11
total_failures         | 0
total_crashes          | 1
-[ RECORD 3 ]----------+--------------------------------------
view_name              | sessions_summary_daily_by_channel
completed_threshold    | 2020-05-26 00:00:00+00
invalidation_threshold | 2020-05-27 00:00:00+00
job_id                 | 1002
last_run_started_at    | 2020-05-27 07:16:51.688871+00
last_successful_finish | 2020-05-26 19:22:26.391064+00
last_run_status        |
job_status             | Running
last_run_duration      |
next_scheduled_run     | -infinity
total_runs             | 12
total_successes        | 10
total_failures         | 1
total_crashes          | 1
```

In the postgresql log file i see the following log lines during each crash
```
2020-05-27 01:22:29.176 UTC [24702] LOG:  materializing continuous aggregate public.sessions_summary_daily: processing invalidations, new range up to 2020-05-27 00:00:00+00
2020-05-27 01:22:29.182 UTC [24742] LOG:  job 1002 threw an error
2020-05-27 01:22:29.182 UTC [24742] ERROR:  tuple concurrently deleted
2020-05-27 01:22:29.184 UTC [8844] LOG:  background worker "Continuous Aggregate Background Job" (PID 24742) exited with exit code 1
2020-05-27 01:22:31.710 UTC [24745] LOG:  materializing continuous aggregate public.sessions_summary_daily_by_document: processing invalidations, new range up to 2020-05-27 00:00:00+00
2020-05-27 01:22:35.227 UTC [24746] LOG:  materializing continuous aggregate public.sessions_summary_daily_by_integration: processing invalidations, new range up to 2020-05-27 00:00:00+00
2020-05-27 01:22:37.392 UTC [24748] LOG:  materializing continuous aggregate public.sessions_summary_daily_by_segment: processing invalidations, new range up to 2020-05-27 00:00:00+00
2020-05-27 07:17:16.386 UTC [16767] LOG:  checkpoints are occurring too frequently (8 seconds apart)
2020-05-27 07:17:16.386 UTC [16767] HINT:  Consider increasing the configuration parameter "max_wal_size".
2020-05-27 07:18:05.758 UTC [8844] LOG:  background worker "Continuous Aggregate Background Job" (PID 14467) was terminated by signal 9: Killed
2020-05-27 07:18:05.835 UTC [8844] LOG:  terminating any other active server processes
```

It looks like the most relevant error message in this log snippet is
```
ERROR:  tuple concurrently deleted
```

**To Reproduce**
This is occurring on a production server and I haven't tried to reproduce it locally yet. Reproduction steps would likely be
1.) Set up a hypertable with several million rows of data
2.) Set up 5 continuous aggregates for that hypertable, each grouping on a different column.
3.) Schedule these jobs all to run at the same time.

**Expected behavior**
Continuous aggregations run without error.

**Actual behavior**
Continuous aggregations fail with the error message `ERROR:  tuple concurrently deleted`

**Screenshots**

**Additional context**
A workaround was suggested to alter the job schedules so that the continuous aggregations run in a staggered schedule. I will try that workaround and see if it alleviates the issue and report back.

gayyappan:
When we have multiple continuous aggregates defined on
    the same hypertable, they could try to delete the hypertable
    invalidation logs concurrently. Resolve this by serializing
    invalidation log deletes by hypertable id.

Fixes #1940

### DIFF
--- a/tsl/src/continuous_aggs/materialize.c
+++ b/tsl/src/continuous_aggs/materialize.c
@@ -78,6 +78,7 @@ static int64 get_materialization_end_point_for_table(int32 raw_hypertable_id,
 													 int64 completed_threshold,
 													 bool *materializing_new_range,
 													 bool *truncated_materialization, bool verbose);
+static void lock_invalidation_threshold_hypertable_row(int32 raw_hypertable_id);
 static void drain_invalidation_log(int32 raw_hypertable_id, List **invalidations_out);
 static void insert_materialization_invalidation_logs(List *caggs, List *invalidations,
 													 Relation rel);
@@ -197,8 +198,18 @@ continuous_agg_materialize(int32 materialization_id, ContinuousAggMatOptions *op
 
 	catalog = ts_catalog_get();
 
-	/* copy over all the materializations from the raw hypertable to all the continuous aggs */
+	/* copy over all the materializations from the raw hypertable to all the continuous aggs
+	 */
 	caggs = ts_continuous_aggs_find_by_raw_table_id(cagg_data.raw_hypertable_id);
+
+	/* REFRESH cagg_1; REFRESH cagg_2 can try to read+delete the
+	 * invalidation logs for the same hypertable.
+	 * If both REFRESH jobs try to delete the
+	 * same set of rows, we run into "tuple concurrently deleted error".
+	 * (github issue 1940)
+	 * Prevent this by serializing on the raw hypertable row
+	 */
+	lock_invalidation_threshold_hypertable_row(cagg_data.raw_hypertable_id);
 	drain_invalidation_log(cagg_data.raw_hypertable_id, &invalidations);
 	materialization_invalidation_log_table_relation =
 		table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_MATERIALIZATION_INVALIDATION_LOG),
@@ -660,6 +671,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
 	return SCAN_CONTINUE;
 }
 
+static ScanTupleResult
+invalidation_threshold_htid_found(TupleInfo *tinfo, void *data)
+{
+	if (tinfo->lockresult != TM_Ok)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("could not acquire lock for invalidation threshold row %d",
+						tinfo->lockresult),
+				 errhint("Retry the operation again")));
+	}
+	return SCAN_DONE;
+}
+
+/* lock row corresponding to hypertable id in
+ * continuous_aggs_invalidation_threshold table in AccessExclusive mode,
+ * block till lock is acquired.
+ */
+static void
+lock_invalidation_threshold_hypertable_row(int32 raw_hypertable_id)
+{
+	ScanTupLock scantuplock = {
+		.waitpolicy = LockWaitBlock,
+		.lockmode = LockTupleExclusive,
+	};
+	Catalog *catalog = ts_catalog_get();
+	ScanKeyData scankey[1];
+	int retcnt = 0;
+	ScannerCtx scanctx;
+
+	ScanKeyInit(&scankey[0],
+				Anum_continuous_aggs_invalidation_threshold_pkey_hypertable_id,
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(raw_hypertable_id));
+
+	/* lock table in AccessShare mode and the row with AccessExclusive */
+	scanctx = (ScannerCtx){ .table = catalog_get_table_id(catalog,
+														  CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
+							.index = catalog_get_index(catalog,
+													   CONTINUOUS_AGGS_INVALIDATION_THRESHOLD,
+													   CONTINUOUS_AGGS_INVALIDATION_THRESHOLD_PKEY),
+							.nkeys = 1,
+							.scankey = scankey,
+							.limit = 1,
+							.tuple_found = invalidation_threshold_htid_found,
+							.lockmode = AccessShareLock,
+							.scandirection = ForwardScanDirection,
+							.result_mctx = CurrentMemoryContext,
+							.tuplock = &scantuplock };
+	retcnt = ts_scanner_scan(&scanctx);
+	if (retcnt > 1)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("found multiple invalidation rows for hypertable %d", raw_hypertable_id)));
+	}
+}
+
+/* Read invalidation log table, copy entries and delete them from this table */
 static void
 drain_invalidation_log(int32 raw_hypertable_id, List **invalidations_out)
 {
@@ -793,6 +864,12 @@ materialization_invalidation_log_get_range(int32 materialization_id, Oid type, i
 	return invalidation_range;
 }
 
+/* We already locked the materialization table in
+ * ShareRowExclusiveLock mode for the Session in the first txn.
+ * This implies that 2 concurrent refreshes on the same cont aggregate
+ * i.e. REFRESH cagg_1; REFRESH cagg_1
+ * are serialized. So 2 concurrent refresh on cagg_1  will not get here.
+ */
 static void
 materialization_invalidation_log_delete_or_cut(int32 cagg_id, InternalTimeRange invalidation_range,
 											   int64 completed_threshold)
--- a/tsl/test/isolation/expected/continuous_aggs_multi.out
+++ b/tsl/test/isolation/expected/continuous_aggs_multi.out
@@ -1,4 +1,4 @@
-Parsed test spec with 9 sessions
+Parsed test spec with 11 sessions
 
 starting permutation: Setup2 LockCompleted LockMat1 Refresh1 Refresh2 UnlockCompleted UnlockMat1
 step Setup2: 
@@ -130,3 +130,53 @@ bkt            maxl
 
 0              4              
 40             1000           
+
+starting permutation: Setup2 Refresh1 Refresh2 Refresh1_sel Refresh2_sel U1 U2 LInvRow Refresh1 Refresh2 UnlockInvRow Refresh1_sel Refresh2_sel
+step Setup2: 
+    CREATE VIEW continuous_view_1( bkt, cnt)
+        WITH ( timescaledb.continuous, timescaledb.refresh_lag = '-5', timescaledb.refresh_interval='72 hours', timescaledb.materialized_only = true)
+        AS SELECT time_bucket('5', time), COUNT(val)
+            FROM ts_continuous_test
+            GROUP BY 1;
+    CREATE VIEW continuous_view_2(bkt, maxl)
+        WITH ( timescaledb.continuous,timescaledb.refresh_lag='-10',  timescaledb.refresh_interval='72 hours', timescaledb.materialized_only = true)
+        AS SELECT time_bucket('5', time), max(val)
+            FROM ts_continuous_test
+            GROUP BY 1;
+    CREATE FUNCTION lock_mattable( name text) RETURNS void AS $$
+    BEGIN EXECUTE format( 'lock table %s', name);
+    END; $$ LANGUAGE plpgsql;
+
+R1: LOG:  materializing continuous aggregate public.continuous_view_1: nothing to invalidate, new range up to 30
+step Refresh1: REFRESH MATERIALIZED VIEW continuous_view_1;
+R2: LOG:  materializing continuous aggregate public.continuous_view_2: nothing to invalidate, new range up to 30
+step Refresh2: REFRESH MATERIALIZED VIEW continuous_view_2;
+step Refresh1_sel: select * from continuous_view_1 where bkt = 0 or bkt > 30
+bkt            cnt            
+
+0              5              
+step Refresh2_sel: select * from continuous_view_2 where bkt = 0 or bkt > 30 order by bkt;
+bkt            maxl           
+
+0              4              
+step U1: update ts_continuous_test SET val = 5555 where time < 10;
+step U2: update ts_continuous_test SET val = 5 where time > 15 and time < 25;
+step LInvRow: BEGIN; update _timescaledb_catalog.continuous_aggs_invalidation_threshold set watermark = 20 where hypertable_id in ( select raw_hypertable_id from _timescaledb_catalog.continuous_agg where user_view_name like 'continuous_view_1' ); 
+
+step Refresh1: REFRESH MATERIALIZED VIEW continuous_view_1; <waiting ...>
+step Refresh2: REFRESH MATERIALIZED VIEW continuous_view_2; <waiting ...>
+step UnlockInvRow: ROLLBACK;
+R1: LOG:  new materialization range not found for public.ts_continuous_test (time column time): not enough new data past completion threshold of 30 as of 29
+R1: LOG:  materializing continuous aggregate public.continuous_view_1: processing invalidations, no new range
+step Refresh1: <... completed>
+R2: LOG:  new materialization range not found for public.ts_continuous_test (time column time): not enough new data past completion threshold of 30 as of 29
+R2: LOG:  materializing continuous aggregate public.continuous_view_2: processing invalidations, no new range
+step Refresh2: <... completed>
+step Refresh1_sel: select * from continuous_view_1 where bkt = 0 or bkt > 30
+bkt            cnt            
+
+0              5              
+step Refresh2_sel: select * from continuous_view_2 where bkt = 0 or bkt > 30 order by bkt;
+bkt            maxl           
+
+0              5555           
--- a/tsl/test/isolation/specs/continuous_aggs_multi.spec
+++ b/tsl/test/isolation/specs/continuous_aggs_multi.spec
@@ -68,6 +68,19 @@ step "UnlockMat1" { ROLLBACK; }
 #alter the refresh_lag for continuous_view_1
 session "CVddl"
 step "AlterLag1" { alter view continuous_view_1 set (timescaledb.refresh_lag = 10); }
+
+#update the hypertable
+session "Upd"
+step "U1" { update ts_continuous_test SET val = 5555 where time < 10; } 
+step "U2" { update ts_continuous_test SET val = 5 where time > 15 and time < 25; } 
+
+#simulate an update to the invalidation threshold table that would lock the hypertable row
+#this would block refresh that needs to get a row lock for the hypertable
+session "LInv"
+step "LInvRow" { BEGIN; update _timescaledb_catalog.continuous_aggs_invalidation_threshold set watermark = 20 where hypertable_id in ( select raw_hypertable_id from _timescaledb_catalog.continuous_agg where user_view_name like 'continuous_view_1' ); 
+}
+step "UnlockInvRow" { ROLLBACK; }
+
  
 #refresh1, refresh2 can run concurrently
 permutation "Setup2" "LockCompleted" "LockMat1" "Refresh1" "Refresh2" "UnlockCompleted" "UnlockMat1"
@@ -75,6 +88,11 @@ permutation "Setup2" "LockCompleted" "LockMat1" "Refresh1" "Refresh2" "UnlockCom
 #refresh1 and refresh2 run concurrently and see the correct invalidation
 #test1 - both see the same invalidation
 permutation "Setup2" "Refresh1" "Refresh2" "LockCompleted" "LockMat1" "I1" "Refresh1" "Refresh2" "UnlockCompleted" "UnlockMat1" "Refresh1_sel" "Refresh2_sel"
+
 ##test2 - continuous_view_2 should see results from insert but not the other one.
 ## Refresh2 will complete first due to LockMat1 and write the invalidation logs out. 
 permutation "Setup2" "AlterLag1" "Refresh1" "Refresh2" "Refresh1_sel" "Refresh2_sel" "LockCompleted" "LockMat1" "I2" "Refresh1" "Refresh2" "UnlockCompleted" "UnlockMat1" "Refresh1_sel" "Refresh2_sel"
+
+#test3 - both see the updates i.e. the invalidations
+##Refresh1 and Refresh2 are blocked by LockInvRow, when that is unlocked, they should complete serially
+permutation "Setup2" "Refresh1" "Refresh2" "Refresh1_sel" "Refresh2_sel" "U1" "U2" "LInvRow" "Refresh1" "Refresh2" "UnlockInvRow" "Refresh1_sel" "Refresh2_sel"   


tsl/src/continuous_aggs/materialize.c
```diff
@@ -793,6 +806,11 @@ materialization_invalidation_log_get_range(int32 materialization_id, Oid type, i
 	return invalidation_range;
 }
 
+/* We already locked the materialization_invalidation_log table in
+ * ShareRowExclusiveLock mode for the Session in the first txn.
```
JLockerman, line 810: where do we take a session lock for the `materialization_invalidation_log`? The only ones I see we take, and the only ones mentioned in the comment on line 156 are: the materialization table, the partial view and the raw hypertable...

gayyappan: fixed the comment. It should be session lock for materialization table.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+						   .lockmode = RowExclusiveLock,
+						   .scandirection = ForwardScanDirection,
+						   .result_mctx = CurrentMemoryContext,
+						   .tuplock = {
+							   .waitpolicy = LockWaitBlock,
+							   .lockmode = LockTupleExclusive,
+							   .enabled = true,
+						   } };
+	/* acquire lock explicitly.
+	 * the catalog scanning code , releases it at end of the call
```
cevian, line 726: is the row lock also released at by catalog scanning code, or no?

gayyappan: tuple locks are released at the end of the txn.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+	ScanKeyData scankey[1];
+
+	ScanKeyInit(&scankey[0],
+				Anum_continuous_aggs_invalidation_threshold_pkey_hypertable_id,
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(raw_hypertable_id));
+
+	/* lock table in RowExclusive mode and the row with AccessExclusive */
+	ScannerCtx scanctx = { .table = catalog_get_table_id(catalog,
```
erimatnor, line 712: mixed declaration and code. I suggest you move this above the `ScanKeyInit`. It should be safe to initialize the scankey after this declaration even if the scankey is referenced by the scanctx.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+						   .scandirection = ForwardScanDirection,
+						   .result_mctx = CurrentMemoryContext,
+						   .tuplock = &scantuplock };
+	/* acquire lock explicitly.
+	 * the catalog scanning code , releases it at end of the call
+	 */
+	invalidation_threshold_table_relation =
+		table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
+				   RowExclusiveLock);
+	ts_scanner_scan(&scanctx);
```
erimatnor, line 731: I suggest you check the return value here (num tuples found) and raise an error in case of not 1.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -197,8 +198,21 @@ continuous_agg_materialize(int32 materialization_id, ContinuousAggMatOptions *op
 
 	catalog = ts_catalog_get();
 
-	/* copy over all the materializations from the raw hypertable to all the continuous aggs */
+	/* copy over all the materializations from the raw hypertable to all the continuous aggs
+	 */
 	caggs = ts_continuous_aggs_find_by_raw_table_id(cagg_data.raw_hypertable_id);
+	if (list_length(caggs) > 1)
```
k-rus, line 204: Do I understand correctly that the lock is only obtained when there are more than one cagg is defined, so there is a risk of concurrency? What's about a corner case when second cagg is created after materialising the first one passed this line? Can it be a race condition if meterializing the first one is slow or interrupted?
Or is this condition to deal with a deadlock?

erimatnor, line 204: Isn't it more safe to always lock here? Could there be a case when cagg_2 is created in another transaction, but not yet visible to this transaction? If cagg_2 is immediately refreshed, I imagine that you won't lock in that case, although you still have concurrent refreshes.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+						   .nkeys = 1,
+						   .scankey = scankey,
+						   .limit = 1,
+						   .tuple_found = invalidation_threshold_htid_found,
+						   .lockmode = RowExclusiveLock,
+						   .scandirection = ForwardScanDirection,
+						   .result_mctx = CurrentMemoryContext,
+						   .tuplock = &scantuplock };
+	/* acquire lock explicitly.
+	 * the catalog scanning code , releases it at end of the call
```
k-rus, line 726: Since they are complete sentences:
```suggestion
	/* Acquire lock explicitly.
	 * The catalog scanning code , releases it at end of the call.
```

tsl/src/continuous_aggs/materialize.c
```diff
@@ -197,8 +198,21 @@ continuous_agg_materialize(int32 materialization_id, ContinuousAggMatOptions *op
...
-	/* copy over all the materializations from the raw hypertable to all the continuous aggs */
+	/* copy over all the materializations from the raw hypertable to all the continuous aggs
+	 */
 	caggs = ts_continuous_aggs_find_by_raw_table_id(cagg_data.raw_hypertable_id);
+	if (list_length(caggs) > 1)
+	{
+		/* REFRESH cagg_1; REFRESH cagg_2 can try to read+delete the
+		 * invalidation logs for the same hypertable.
+		 * If both REFRESH jobs try to delete the
+		 * same set of rows, we run into "tuple concurrently deleted error".
```
erimatnor, line 209: I guess we could in theory handle this without locking. A concurrently deleted tuple might not strictly be an error  since it doesn't really matter which job deleted the tuple. If the job that experience the error aimed to delete the tuple itself, then I guess it is fine for that job to proceed as if it actually succeeded. 

This can, of course, be left as a future optimization. But I still figured it is worth pointing this out. Maybe leave a comment.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+						   .result_mctx = CurrentMemoryContext,
+						   .tuplock = &scantuplock };
+	/* acquire lock explicitly.
+	 * the catalog scanning code , releases it at end of the call
+	 */
+	invalidation_threshold_table_relation =
+		table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
+				   RowExclusiveLock);
+	ts_scanner_scan(&scanctx);
+	table_close(invalidation_threshold_table_relation, NoLock); /*release lock at end
```
erimatnor, line 732: I don't understand why we keep the table lock here (in `RowExclusiveLock`) mode. That seems to negate the purpose of taking the tuple lock since you are now serializing around the invalidation threshold table instead of the tuple. So, in effect, this is a much more heavy-handed (global) serialization that blocks refreshes across all caggs for all hypertables.

Or, am I missing something here?

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+						   .lockmode = RowExclusiveLock,
+						   .scandirection = ForwardScanDirection,
+						   .result_mctx = CurrentMemoryContext,
+						   .tuplock = &scantuplock };
+	/* acquire lock explicitly.
+	 * the catalog scanning code , releases it at end of the call
+	 */
+	invalidation_threshold_table_relation =
+		table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
+				   RowExclusiveLock);
```
erimatnor, line 730: Why is this lock `RowExclusiveLock`? Are we modifying this table in addition to taking the tuple lock? I suggest adding a comment on why this is the appropriate lock mode since it is non-obvious.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+
+	ScanKeyInit(&scankey[0],
+				Anum_continuous_aggs_invalidation_threshold_pkey_hypertable_id,
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(raw_hypertable_id));
+
+	/* lock table in RowExclusive mode and the row with AccessExclusive */
+	ScannerCtx scanctx = { .table = catalog_get_table_id(catalog,
+														 CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
```
k-rus, line 713: Will it be more efficient, but still correct to call `catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD)` once, store in a variable, and use in both places, here and below?

tsl/src/continuous_aggs/materialize.c
k-rus, line 712: This declaration is placed after calling `ScanKeyInit`. Would it be better to do all declarations before the call?

tsl/src/continuous_aggs/materialize.c
k-rus, line 732: I don't follow why a lock needs to be obtained if the entire table is locked until the end of the transaction.

k-rus: I think I understand now that the tuple is locked for the session, which is longer than the transaction, right? Where is the lock on the tuple released?

tsl/test/isolation/expected/continuous_aggs_multi.out
```diff
@@ -130,3 +130,53 @@ bkt            maxl
...
+
+step Refresh1: REFRESH MATERIALIZED VIEW continuous_view_1; <waiting ...>
+step Refresh2: REFRESH MATERIALIZED VIEW continuous_view_2; <waiting ...>
+step UnlockInvRow: ROLLBACK;
+R1: LOG:  new materialization range not found for public.ts_continuous_test (time column time): not enough new data past completion threshold of 30 as of 29
+R1: LOG:  materializing continuous aggregate public.continuous_view_1: processing invalidations, no new range
+step Refresh1: <... completed>
+R2: LOG:  new materialization range not found for public.ts_continuous_test (time column time): not enough new data past completion threshold of 30 as of 29
+R2: LOG:  materializing continuous aggregate public.continuous_view_2: processing invalidations, no new range
+step Refresh2: <... completed>
```
k-rus, line 174: Is it deterministic that `Refresh1` will be executed first after the lock is released and then `Refresh2`?

k-rus:
@gayyappan Is this PR related to any public issue?

tsl/src/continuous_aggs/materialize.c
erimatnor: What is a session lock? I guess it is confusing because it sounds like the lock is held for the duration of the session. Is it even possible to hold locks across multiple transactions unless it is an advisory lock? The documentation says: "Once acquired, a lock is normally held until the end of the transaction."

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
 	return SCAN_CONTINUE;
 }
 
+static ScanTupleResult
+invalidation_threshold_htid_found(TupleInfo *tinfo, void *data)
+{
+	if (tinfo->lockresult != TM_Ok)
+	{
+		ereport(ERROR,
```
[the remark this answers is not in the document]
gayyappan: Anything other than TM_Ok is not valid here.  Knowing the specific failure doesn't seem to be helpful since we have to abort the transaction anyway.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +674,66 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("unexpected lockresult %d in invalidation_threshold_htid_found",
+						tinfo->lockresult)));
+	}
+	return SCAN_DONE;
+}
+
+/* lock row corresponding to hypertable id in
+ * continuous_aggs_invalidation_threshold table in AccessExclusive mode
+ * block till lock is acquired.
```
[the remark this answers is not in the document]
gayyappan: The process is waiting on the lock till it is acquired.

tsl/src/continuous_aggs/materialize.c
gayyappan: tuple locks are released at the end of the transaction.

tsl/src/continuous_aggs/materialize.c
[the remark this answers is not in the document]
k-rus: Thank you for clarifying. I suggest to fix the grammar, so it is easier to read later. I guess comma or `and` is missing.

tsl/src/continuous_aggs/materialize.c
k-rus: So why is it necessary to obtain tuple lock if the table is already locked for the same duration?

gayyappan: Yes, since a concurrent transaction will not conflict on the table lock. The tuple lock helps serialize the transactions. This begs the question is the tuple lock sufficient? :  I believe that you need a table lock before you acquire the tuple lock. (I am going to change the table lock from RowExclusive to AccessShare)

tsl/src/continuous_aggs/materialize.c
gayyappan: IIRC we said that is okay to mix declarations and code. We are on C99.

tsl/src/continuous_aggs/materialize.c
gayyappan: @k-rus @erimatnor Yes, the corner case is possible. But not fully convinced that we should deal with the error for the corner case. The locks have a concurrency cost. So acquiring a lock only if it is necessary to do so seems to be a good trade off to me. ( There is an additional optimization possible here : a first time cagg does not have to deal with any invalidations since it has to fully materialize all the data. But we don't optimize for that case. Then this code might actually not be exercised in this scenario).

tsl/src/continuous_aggs/materialize.c
gayyappan: the session lock is for the duration of the materialization.

tsl/src/continuous_aggs/materialize.c
gayyappan: changed to AccessShare  based on our discssion.

tsl/test/isolation/expected/continuous_aggs_multi.out
gayyappan: No, cannot be guaranteed. But if Refresh1 starts before Refresh2, it  would acquire the released lock first. The sessions are generally started in the order in which they appear in the spec file.

tsl/src/continuous_aggs/materialize.c
erimatnor: Do we know how big this cost is and do we throw a reasonable error when we hit the corner case? If not, I think we should cover the corner-case too and care about "correctness" prior to optimizing. We can always optimize further at a later time.

tsl/src/continuous_aggs/materialize.c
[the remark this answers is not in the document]
erimatnor: But the user might see this error and an update doesn't seem like it would be "unexpected". Maybe not common, but not unexpected either.

tsl/src/continuous_aggs/materialize.c
k-rus: I would like to add: which PG processes will be affected by this cost? Will insertion to hypertable or querying hypertable be affected by this concurrency overhead?

tsl/test/isolation/expected/continuous_aggs_multi.out
k-rus: Does this test reproduce the bug before the fix?

gayyappan: All this test proves is the refresh gets serialized on the hypertable row lock. This is true for all the concurrency tests. They do not necessarily "catch" all bugs with the code.

tsl/src/continuous_aggs/materialize.c
erimatnor: More of a suggestion. But FWIW, we haven't used mixed-declaration-and-code in most of our other code.  Not the appropriate place to discuss decisions. But  this is suggested by two reviewers and seems like an easy fix.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +671,75 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+	 */
+	invalidation_threshold_table_relation =
+		table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
+				   AccessShareLock);
+	retcnt = ts_scanner_scan(&scanctx);
+	if (retcnt > 1)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("unexpected return value %d in lock_invalidation_threhold_hypertable_row",
```
erimatnor, line 735: Using the function name in error messages is redundant as elog already tracks this and the function and line number can be shown if the right log configuration is set. I think it is better to formulate a good error message to describe the error, e.g.: "duplicate catalog entry for hypertable \"%s\"". As a last resort, use the `__func__` macro as that also handles function name changes.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -660,6 +671,75 @@ scan_take_invalidation_tuple(TupleInfo *ti, void *data)
...
+						   .result_mctx = CurrentMemoryContext,
+						   .tuplock = &scantuplock };
+	/* Acquire lock explicitly.
+	 * the catalog scanning code , releases it at end of the call
+	 */
+	invalidation_threshold_table_relation =
+		table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
+				   AccessShareLock);
+	retcnt = ts_scanner_scan(&scanctx);
+	if (retcnt > 1)
```
erimatnor, line 731: Is anything but 1 expected here? Isn't 0 (or less) also errors?

tsl/src/continuous_aggs/materialize.c
erimatnor: Lock still held here. I thought that the table lock was going to be released here since tuple lock is held and that is the actual serialization point.

tsl/src/continuous_aggs/materialize.c
gayyappan: could be 0 if there are no entries in the table yet for that hypertable.

tsl/src/continuous_aggs/materialize.c
gayyappan: I think the struct initialization is more readable here.  scanctx depends on scankey. That's why there is mixed declaration in this block

k-rus: `get_lowest_invalidated_time_for_hypertable` contains similar code with variable declaration first:
```c
	ScanKeyData scankey[1];
	ScannerCtx scanctx;

	ScanKeyInit(&scankey[0],
				Anum_continuous_aggs_invalidation_threshold_pkey_hypertable_id,
				BTEqualStrategyNumber,
				F_INT4EQ,
				Int32GetDatum(ts_hypertable_relid_to_id(hypertable_relid)));
	scanctx = (ScannerCtx){
		.table = catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
		.index = catalog_get_index(catalog,
								   CONTINUOUS_AGGS_INVALIDATION_THRESHOLD,
								   CONTINUOUS_AGGS_INVALIDATION_THRESHOLD_PKEY),
		.nkeys = 1,
		.scankey = scankey,
		.tuple_found = &invalidation_tuple_found,
		.filter = NULL,
		.data = &min_val,
		.lockmode = AccessShareLock,
		.scandirection = ForwardScanDirection,
		.result_mctx = NULL,
	};
```
I don't have strong opinion, but there might be a test, which will fail declaration after code. However, I might be mistaking.

tsl/src/continuous_aggs/materialize.c
gayyappan: I misunderstood that - modified so that we only have the tuple lock.

tsl/src/continuous_aggs/materialize.c
gayyappan: Thanks for the pointer.

PR #2054, opened 2020-07-02
Add regresscheck-shared sql files to license check

svenklemm:
This adds the directories used by regresscheck-shared to the license
check and adds a license where it is currently missing.


k-rus:
> I guess the expected output files should get the license comments into it.

Or the outputs of the shared tests are hidden by the calling tests.

svenklemm:
Haven't rebuild the output files, will do before merging

PR #2059, opened 2020-07-04
Improve infering start and stop arguments from gapfill query

svenklemm:
Infering start and stop parameter used to only work for top
level constraints. However even when constraints are at the
toplevel in the query they might not end up at the top-level
of the jointree depending on the plan shape. This patch
changes the gapfill code to traverse the jointree to find
valid start and stop arguments.

### DIFF
--- a/tsl/src/nodes/gapfill/exec.c
+++ b/tsl/src/nodes/gapfill/exec.c
@@ -374,6 +374,99 @@ get_boundary_expr_value(GapFillState *state, GapFillBoundary boundary, Expr *exp
 	return gapfill_datum_get_internal(arg_value, state->gapfill_typid);
 }
 
+typedef struct CollectBoundaryContext
+{
+	List *quals;
+	Var *ts_var;
+} CollectBoundaryContext;
+
+/*
+ * expression references our gapfill time column and could be
+ * a boundary expression, more thorough check is in
+ * infer_gapfill_boundary
+ */
+static bool
+is_boundary_expr(Node *node, CollectBoundaryContext *context)
+{
+	OpExpr *op;
+	Node *left, *right;
+
+	if (!IsA(node, OpExpr))
+		return false;
+
+	op = castNode(OpExpr, node);
+
+	if (op->args->length != 2)
+		return false;
+
+	left = linitial(op->args);
+	right = llast(op->args);
+
+	/* Var OP Var is not useful here because we are not yet at a point
+	 * where we could evaluate them */
+	if (IsA(left, Var) && IsA(right, Var))
+		return false;
+
+	if (IsA(left, Var) && var_equal(castNode(Var, left), context->ts_var))
+		return true;
+
+	if (IsA(right, Var) && var_equal(castNode(Var, right), context->ts_var))
+		return true;
+
+	return false;
+}
+
+static bool
+collect_boundary_walker(Node *node, CollectBoundaryContext *context)
+{
+	Node *quals = NULL;
+
+	if (node == NULL)
+		return false;
+
+	if (IsA(node, FromExpr))
+	{
+		quals = castNode(FromExpr, node)->quals;
+	}
+	else if (IsA(node, JoinExpr))
+	{
+		JoinExpr *j = castNode(JoinExpr, node);
+
+		/* don't descend into outer join */
+		if (IS_OUTER_JOIN(j->jointype))
+			return false;
+
+		quals = j->quals;
+	}
+
+	if (quals)
+	{
+		ListCell *lc;
+
+		foreach (lc, castNode(List, quals))
+		{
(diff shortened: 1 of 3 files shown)


tsl/src/nodes/gapfill/exec.c
```diff
@@ -374,6 +374,99 @@ get_boundary_expr_value(GapFillState *state, GapFillBoundary boundary, Expr *exp
 	return gapfill_datum_get_internal(arg_value, state->gapfill_typid);
 }
 
+typedef struct collect_boundary_context
```
erimatnor, line 377: Nit: our code style says types/structs should use camelcase like PostgreSQL, i.e., `CollectBoundaryContext`

tsl/src/nodes/gapfill/exec.c
```diff
@@ -374,6 +374,99 @@ get_boundary_expr_value(GapFillState *state, GapFillBoundary boundary, Expr *exp
...
+static bool
+collect_boundary_walker(Node *node, collect_boundary_context *context)
+{
+	Node *quals = NULL;
+
+	if (node == NULL)
+		return false;
+
+	if (IsA(node, FromExpr))
+	{
```
erimatnor, line 428: Nit: elide braces, like above.

tsl/src/nodes/gapfill/exec.c
```diff
@@ -403,18 +497,17 @@ infer_gapfill_boundary(GapFillState *state, GapFillBoundary boundary)
...
 		Var *var;
 		Expr *expr;
 		Oid op;
 		int64 value;
 
-		if (!IsA(lfirst(lc), OpExpr))
-			continue;
-
-		opexpr = lfirst(lc);
+		opexpr = lfirst_node(OpExpr, lc);
```
erimatnor, line 510: Nit: move this up to declaration

tsl/test/expected/gapfill.out
```diff
@@ -1843,14 +1843,39 @@ FROM metrics_int m1 INNER JOIN metrics_int m2 ON m1.time=m2.time
 WHERE m1.time >=0 AND m1.time < 2
 GROUP BY 1;
 ERROR:  invalid time_bucket_gapfill argument: could not infer start boundary from WHERE clause
+-- test outer join with constraints in join condition
+-- not usable as start/stop
+SELECT
+  time_bucket_gapfill(1,m1.time)
+FROM metrics_int m1 LEFT OUTER JOIN metrics_int m2 ON m1.time=m2.time AND m1.time >=0 AND m1.time < 2
+GROUP BY 1;
+ERROR:  invalid time_bucket_gapfill argument: could not infer start boundary from WHERE clause
```
erimatnor, line 1852: This is an odd error message and it doesn't follow the error style guide. Besides, I cannot figure out what it means. Why is it saying invalid gapfill argument and then complaining about the WHERE clause, which neither is an argument to gapfill, nor exists in the query?

Maybe it should just say: errmsg: `could not infer start boundary for gapfill`. errhint: `A gapfill query requires a start boundary on "time" in the WHERE clause.`

svenklemm: A gapfill query requires a start and stop arguments, in the first version you had to supply those as arguments, we later added code to infer start and stop from the WHERE-clause however it can only be infered when additional constraints are satisifed (expression needs to be immutable). So the error means we couldnt infer start from the WHERE clause and it needs to be supplied as argument to the time_bucket_gapfill call (or WHERE clause needs to be rewritten so it can be infered)

tsl/src/nodes/gapfill/exec.c
```diff
@@ -374,6 +374,99 @@ get_boundary_expr_value(GapFillState *state, GapFillBoundary boundary, Expr *exp
...
+
+	if (IsA(node, FromExpr))
+	{
+		quals = castNode(FromExpr, node)->quals;
+	}
+	else if (IsA(node, JoinExpr))
+	{
+		JoinExpr *j = castNode(JoinExpr, node);
+
+		/* don't descdend into outer join */
```
pmwkaa, line 435: descend? :)

tsl/src/nodes/gapfill/exec.c
```diff
@@ -374,6 +374,99 @@ get_boundary_expr_value(GapFillState *state, GapFillBoundary boundary, Expr *exp
 	return gapfill_datum_get_internal(arg_value, state->gapfill_typid);
 }
 
+typedef struct collect_boundary_context
+{
+	List *quals;
+	Var *ts_var;
+} collect_boundary_context;
```
pmwkaa, line 381: nit: I guess you dont' have to keep the same name for type and struct, probably better to omit struct name in that case completely.

PR #2064, opened 2020-07-06
Use macros to define crossmodule wrapper functions

svenklemm:
This patch adds 2 helper macros to define cross module
wrapper functions to reduce code repetition.


src/cross_module_fn.c
```diff
@@ -13,372 +13,96 @@
...
-Datum
-ts_remote_connection_cache_show(PG_FUNCTION_ARGS)
-{
-	PG_RETURN_DATUM(ts_cm_functions->remote_connection_cache_show(fcinfo));
-}
+#define TSL_FORWARD_CALL(func)                                                                     \
+	TS_FUNCTION_INFO_V1(ts_##func);                                                                \
+	Datum ts_##func(PG_FUNCTION_ARGS) { return ts_cm_functions->func(fcinfo); }
+
+#define TSL_FORWARD_CALL2(func, cmfunc)                                                            \
```
WireBaron, line 20: Not a fan of of this macro name.  Maybe `TSL_FORWARD_REDIRECT` or `TSL_FORWARD_TO_FUNCTION`.

src/cross_module_fn.c
```diff
@@ -13,372 +13,96 @@
...
-}
-
-Datum
-ts_remote_connection_cache_show(PG_FUNCTION_ARGS)
-{
-	PG_RETURN_DATUM(ts_cm_functions->remote_connection_cache_show(fcinfo));
-}
+#define TSL_FORWARD_CALL(func)                                                                     \
+	TS_FUNCTION_INFO_V1(ts_##func);                                                                \
+	Datum ts_##func(PG_FUNCTION_ARGS) { return ts_cm_functions->func(fcinfo); }
```
mkindahl, line 18: `PG_RETURN_DATUM` does not do much, I know, but this is at least makes it clear that the function is expected to return a `Datum`
```suggestion
	Datum ts_##func(PG_FUNCTION_ARGS) { PG_RETURN_DATUM(ts_cm_functions->func(fcinfo)); }
```

src/cross_module_fn.c
svenklemm: how about CROSSMODULE_WRAPPER

mkindahl: Sounds fine.

src/cross_module_fn.c
```diff
@@ -13,372 +13,96 @@
...
 	PG_RETURN_BOOL(ts_cm_functions->set_distributed_id(PG_GETARG_DATUM(0)));
 }
 
+TS_FUNCTION_INFO_V1(ts_dist_set_peer_id);
 Datum
 ts_dist_set_peer_id(PG_FUNCTION_ARGS)
 {
 	ts_cm_functions->set_distributed_peer_id(PG_GETARG_DATUM(0));
 	PG_RETURN_VOID();
 }
```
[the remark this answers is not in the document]
mkindahl: Yeah, it's a `PG_RETURN_VOID`, but you might want to consider introducing a version for that as well.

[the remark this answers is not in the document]
svenklemm: the signature of those 3 functions is different but i will adjust those functions so the macro can be used  for those as well

src/cross_module_fn.c
mkindahl: You could actually do something along these lines, if you want to be fancy and make something more generic:

```c
#define WRAPPER(FUNC, RETURNER) \
  TS_FUNCTION_INFO_V1(ts_##FUNC); \
  Datum ts_##FUNC (PG_FUNCTION_ARGS) { RETUNER(ts_cm_functions->func(fcinfo);
```
```

File: src/cross_module_fn.c (hunk 3 above)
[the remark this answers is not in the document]
Comment (svenklemm) [reply src/cross_module_fn.c:73]: its not about the return but the argument that are forwarded, all the other functions take fcinfo

File: src/cross_module_fn.c (hunk 1 above)
Comment (erimatnor) [reply src/cross_module_fn.c:20]: I would consider coming up with a standard for naming functions that doesn't require two different macros and mandates that the public function has a similar name to the function pointer in the cross-module struct.

Our [code style recommends namespacing](https://github.com/timescale/timescaledb/blob/master/docs/StyleGuide.md#modular-code-and-namespacing) for public functions, but a lot of functions for compression, bgw, and caggs, don't seem to follow that principle. Thus to avoid multiple macros, I suggest:

1. Mandate that function pointers in the cross-module struct have similar name as public function
2. Make sure public functions follow code style (i.e., use module prefix)

I guess (1) makes sense for this PR, while (2) isn't strictly necessary to get rid of the extra macro.

Comment (svenklemm) [reply src/cross_module_fn.c]: i removed 2nd macro, making all the function names consistent is definitely outside the scope of this PR and could also be done per subsystem

**Comment by svenklemm** [comment]
Jumping to the implementation to these is not useful at all and to see what they actually call you need to look it up in tsl/src/init.c to see what gets assigned to the function pointer.
I dont understand your 2nd point all of those are crossmodule cause they are getting called from non-tsl called since all our sql functions are defined in non-tsl code

File: src/cross_module_fn.c [hunk 4]
```diff
@@ -13,372 +13,95 @@
...
 	ts_cm_functions->validate_as_data_node();
 	PG_RETURN_VOID();
 }
 
-Datum
-ts_distributed_exec(PG_FUNCTION_ARGS)
-{
-	ts_cm_functions->distributed_exec(fcinfo);
-	PG_RETURN_VOID();
-}
```
mkindahl, line 258: This is different from the others (it uses `PG_RETURN_VOID`) but I agree that it should actually just forward the return value from `distributed_exec`.

PR #2066, opened 2020-07-07
Move cron tests to GitHub actions

svenklemm:
This patch moves the sanitizer, abi breakage and memory leak
tests from travis to github actions.


.github/workflows/cron-tests.yaml
```diff
@@ -0,0 +1,57 @@
...
+  abi_breakage:
+    name: ABI Breakage PG${{ matrix.pg }}
+    runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        pg: [11, 12]
+    env:
+      PG_MAJOR: ${{ matrix.pg }}
+      PG_MINOR_COMPILE: 0
```
WireBaron, line 19: If I'm reading this correctly, this will run on PG 11.0 and 12.0. If that's correct, should we try to run this on more current version (12.3 and 11.8)?

svenklemm: actually what this does is build the extension on the specified version and then spins up the latest postgres version and tests that everything is still working so it builds on 11.0 and then tests against 11.8

PR #2082, opened 2020-07-14
Adopt table scan methods for Scanner

### LINKED ISSUE #2075: Refactor Scanner to use table scan interface
Since PG12, tables can have different access methods (storage). However, our Scanner code still uses heap access, which means it can only scan heap tables. 

We should adopt the table scan API to make  Scanner agnostic to the underlying storage of a table.

erimatnor:
This change makes the Scanner code agnostic to the underlying storage
implementation of the tables it scans. This also fixes a bug that made
it impossible to use non-heap table access methods on a
hypertable. The bug existed because a check is made for existing data
before a table is made into a hypertable. And, since this check reads
data from the table using the Scanner, it must be able to read the
data irrespective of the underlying storage.

As a result of the more generic scan interface, resource management is
also improved by delivering tuples in reference-counted tuple table
slots. A backwards-compatibility layer is used for PG11, which maps
all table access functions to the heap equivalents.

A potential memory corruption issue is also fixed when scanning chunk constraints (separate commit).

Fixes #2075

### DIFF
--- a/src/bgw/job.c
+++ b/src/bgw/job.c
@@ -185,9 +185,11 @@ get_job_type_from_name(Name job_type_name)
 }
 
 static BgwJob *
-bgw_job_from_tuple(TupleInfo *ti, size_t alloc_size)
+bgw_job_from_tupleinfo(TupleInfo *ti, size_t alloc_size)
 {
 	BgwJob *job;
+	bool should_free;
+	HeapTuple tuple;
 	bool isnull;
 	MemoryContext old_ctx;
 	Datum value;
@@ -197,17 +199,21 @@ bgw_job_from_tuple(TupleInfo *ti, size_t alloc_size)
 	 * the STRUCT_FROM_TUPLE macro
 	 */
 	Assert(alloc_size >= sizeof(BgwJob));
-	job = (BgwJob *) MemoryContextAllocZero(ti->mctx, alloc_size);
-	memcpy(job, GETSTRUCT(ti->tuple), sizeof(FormData_bgw_job));
+	job = MemoryContextAllocZero(ti->mctx, alloc_size);
+	tuple = ts_scanner_fetch_heap_tuple(ti, false, &should_free);
+	memcpy(job, GETSTRUCT(tuple), sizeof(FormData_bgw_job));
+
+	if (should_free)
+		heap_freetuple(tuple);
 
 	/*
 	 * GETSTRUCT does not work with variable length types and NULLs so we have
 	 * to do special handling for hypertable_id and the jsonb column
 	 */
-	value = heap_getattr(ti->tuple, Anum_bgw_job_hypertable_id, ti->desc, &isnull);
+	value = slot_getattr(ti->slot, Anum_bgw_job_hypertable_id, &isnull);
 	job->fd.hypertable_id = isnull ? 0 : DatumGetInt32(value);
 
-	value = heap_getattr(ti->tuple, Anum_bgw_job_config, ti->desc, &isnull);
+	value = slot_getattr(ti->slot, Anum_bgw_job_config, &isnull);
 	old_ctx = MemoryContextSwitchTo(ti->mctx);
 	job->fd.config = isnull ? NULL : DatumGetJsonbP(value);
 	MemoryContextSwitchTo(old_ctx);
@@ -227,7 +233,7 @@ static ScanTupleResult
 bgw_job_accum_tuple_found(TupleInfo *ti, void *data)
 {
 	AccumData *list_data = data;
-	BgwJob *job = bgw_job_from_tuple(ti, list_data->alloc_size);
+	BgwJob *job = bgw_job_from_tupleinfo(ti, list_data->alloc_size);
 	MemoryContext orig = MemoryContextSwitchTo(ti->mctx);
 
 	list_data->list = lappend(list_data->list, job);
@@ -240,7 +246,7 @@ static ScanFilterResult
 bgw_job_filter_scheduled(TupleInfo *ti, void *data)
 {
 	bool isnull;
-	Datum scheduled = heap_getattr(ti->tuple, Anum_bgw_job_scheduled, ti->desc, &isnull);
+	Datum scheduled = slot_getattr(ti->slot, Anum_bgw_job_scheduled, &isnull);
 	Assert(!isnull);
 
 	return DatumGetBool(scheduled);
@@ -443,8 +449,8 @@ ts_bgw_job_find_with_lock(int32 bgw_job_id, MemoryContext mctx, LOCKMODE tuple_l
 
 	ts_scanner_foreach(&iterator)
 	{
-		job = bgw_job_from_tuple(ts_scan_iterator_tuple_info(&iterator), sizeof(BgwJob));
-
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		job = bgw_job_from_tupleinfo(ti, sizeof(BgwJob));
 		Assert(num_found == 0);
 		num_found++;
 	}
@@ -498,7 +504,7 @@ ts_bgw_job_find(int32 bgw_job_id, MemoryContext mctx, bool fail_if_not_found)
 	ts_scanner_foreach(&iterator)
 	{
 		Assert(num_found == 0);
-		job = bgw_job_from_tuple(ts_scan_iterator_tuple_info(&iterator), sizeof(BgwJob));
+		job = bgw_job_from_tupleinfo(ts_scan_iterator_tuple_info(&iterator), sizeof(BgwJob));
 		num_found++;
 	}
 
@@ -556,7 +562,11 @@ static ScanTupleResult
 bgw_job_tuple_delete(TupleInfo *ti, void *data)
 {
 	CatalogSecurityContext sec_ctx;
-	int32 job_id = ((FormData_bgw_job *) GETSTRUCT(ti->tuple))->id;
+	bool isnull;
+	Datum datum = slot_getattr(ti->slot, Anum_bgw_job_id, &isnull);
+	int32 job_id = DatumGetInt32(datum);
+
+	Assert(!isnull);
 
 	/* Also delete the bgw_stat entry */
 	ts_bgw_job_stat_delete(job_id);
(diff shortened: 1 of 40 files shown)


src/chunk_index.c
```diff
@@ -863,8 +891,11 @@ chunk_index_tuple_rename(TupleInfo *ti, void *data)
 	else
 		namestrcpy(&chunk_index->index_name, info->newname);
 
-	ts_catalog_update(ti->scanrel, tuple);
-	heap_freetuple(tuple);
+	ts_catalog_update(ti->scanrel, new_tuple);
+	heap_freetuple(new_tuple);
+
+	if (should_free)
+		heap_freetuple(tuple);
```
k-rus, line 898: Can it be freed directly after copying to `new_tuple` or use `materialise = true` instead of `new_tuple`?

src/planner.c
```diff
@@ -33,8 +33,8 @@
 #include <parser/analyze.h>
 
 #include <catalog/pg_constraint.h>
-#include "compat.h"
 #include "compat-msvc-exit.h"
+#include "compat.h"
```
k-rus, line 37: Is there any particular reason for this reordering? I doubt `compat.h` depends on `compat-msvc-exit.h`. If not, then it breaks lexicographical order and is unnecessary change for the PR.

src/scan_iterator.h
```diff
@@ -37,16 +37,27 @@ ts_scan_iterator_tuple_info(ScanIterator *iterator)
...
+	return iterator->tinfo->slot;
+}
+
 static inline HeapTuple
 ts_scan_iterator_tuple(ScanIterator *iterator)
 {
-	return iterator->tinfo->tuple;
+	return ExecFetchSlotHeapTuple(iterator->tinfo->slot, false, NULL);
+}
+static inline HeapTuple
```
k-rus, line 51: New line is missing:
```suggestion

static inline HeapTuple
```

src/scan_iterator.h
```diff
@@ -37,16 +37,27 @@ ts_scan_iterator_tuple_info(ScanIterator *iterator)
...
+ts_scan_iterator_slot(ScanIterator *iterator)
+{
+	return iterator->tinfo->slot;
+}
+
 static inline HeapTuple
 ts_scan_iterator_tuple(ScanIterator *iterator)
 {
-	return iterator->tinfo->tuple;
+	return ExecFetchSlotHeapTuple(iterator->tinfo->slot, false, NULL);
```
k-rus, line 49: Since `ts_scan_iterator_slot` is defined, shouldn't it be used already here?
```suggestion
	return ExecFetchSlotHeapTuple(ts_scan_iterator_slot(iterator), false, NULL);
```

src/scan_iterator.h
```diff
@@ -37,16 +37,27 @@ ts_scan_iterator_tuple_info(ScanIterator *iterator)
...
+}
+static inline HeapTuple
+ts_scan_iterator_materialized_tuple(ScanIterator *iterator)
+{
+	return ExecFetchSlotHeapTuple(iterator->tinfo->slot, true, NULL);
```
k-rus, line 54:
```suggestion
	return ExecFetchSlotHeapTuple(ts_scan_iterator_slot(iterator), true, NULL);
```

src/scan_iterator.h
```diff
@@ -37,16 +37,27 @@ ts_scan_iterator_tuple_info(ScanIterator *iterator)
 	return iterator->tinfo;
 }
 
+static inline TupleTableSlot *
+ts_scan_iterator_slot(ScanIterator *iterator)
+{
+	return iterator->tinfo->slot;
+}
```
k-rus, line 44: I don't see this function is used anywhere. Should it be dropped or used?

src/scanner.c
```diff
@@ -306,3 +337,13 @@ ts_scanner_scan_one(ScannerCtx *ctx, bool fail_if_not_found, char *item_type)
 			return false;
 	}
 }
+
+TSDLLEXPORT ItemPointer
```
k-rus, line 341: We usually don't include exporting details such as this macro into C files, but this file seems to be a complete exception :)

tsl/src/continuous_aggs/options.c
```diff
@@ -173,17 +180,25 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
 		bool nulls[Natts_continuous_agg];
 		Datum values[Natts_continuous_agg];
 		bool repl[Natts_continuous_agg] = { false };
-		HeapTuple new;
+		bool should_free;
+		HeapTuple tuple = ExecFetchSlotHeapTuple(ti->slot, false, &should_free);
+		HeapTuple new_tuple;
+		TupleDesc tupdesc = ts_scan_iterator_tupledesc(&iterator);
```
k-rus, line 186: I've noticed that other code files use the code of `ts_scan_iterator_tupledesc` to get tuple description. It would be nice homogenise that, e.g., all relevant cases use `ts_scan_iterator_tupledesc`.

src/utils.h
```diff
@@ -96,6 +98,8 @@ extern TSDLLEXPORT List *ts_get_reloptions(Oid relid);
 
 #define STRUCT_FROM_TUPLE(tuple, mctx, to_type, form_type)                                         \
 	(to_type *) ts_create_struct_from_tuple(tuple, mctx, sizeof(to_type), sizeof(form_type));
```
k-rus, line 100: I don't think it used anywhere as it is replaced with `STRUCT_FROM_SLOT`. Can it be removed?

src/bgw_policy/chunk_stats.c
```diff
@@ -154,17 +157,19 @@ static ScanTupleResult
...
 {
 	TimestampTz *updated_last_time_job_run = data;
-	HeapTuple tuple = heap_copytuple(ti->tuple);
-	BgwPolicyChunkStats *chunk_stats = STRUCT_FROM_TUPLE(ti->tuple,
-														 ti->mctx,
-														 BgwPolicyChunkStats,
-														 FormData_bgw_policy_chunk_stats);
+	bool should_free;
+	HeapTuple tuple = ExecFetchSlotHeapTuple(ti->slot, false, &should_free);
+	HeapTuple new_tuple = heap_copytuple(tuple);
```
gayyappan, line 162: I am puzzled by this change: What do we gain by introducing a call to ExecFetchSlotHeapTuple at line 161.? We make a copy of the tuple here.

src/bgw/job.c
```diff
@@ -184,9 +184,11 @@ get_job_type_from_name(Name job_type_name)
 }
 
 static BgwJob *
-bgw_job_from_tuple(HeapTuple tuple, size_t alloc_size, MemoryContext mctx)
+bgw_job_from_tupleinfo(TupleInfo *ti, size_t alloc_size, MemoryContext mctx)
 {
 	BgwJob *job;
+	bool should_free;
+	HeapTuple tuple = ExecFetchSlotHeapTuple(ti->slot, false, &should_free);
```
gayyappan, line 191: How does the memory behave before/after this change? I see code where we copy the tuple contents into alloc'ed memory. Could you elaborate on how/why ExecFetchSlotHeapTuple helps here.

src/chunk.c
```diff
@@ -2603,7 +2608,7 @@ chunk_tuple_delete(TupleInfo *ti, DropBehavior behavior, bool preserve_chunk_cat
 
 	if (!preserve_chunk_catalog_row)
 	{
-		ts_catalog_delete(ti->scanrel, ti->tuple);
+		ts_catalog_delete_tid(ti->scanrel, ts_scanner_get_tuple_tid(ti));
```
gayyappan, line 2611: I see some places where ts_catalog_delete has been replaced by ts_catalog_delete_tid. What is the reasoning behind the change? Aren't they equivalent in this case?

erimatnor: The table API wraps tuples in tuple slots, so we don't have direct access to the tuple unless it is potentially re-materialized, which comes at a cast. FWIW, `ts_catalog_delete` takes a tuple, but internally just calls `ts_catalog_delete_tid`.

src/bgw/job_stat.c
```diff
@@ -117,8 +116,10 @@ ts_bgw_job_stat_delete(int32 bgw_job_id)
 static ScanTupleResult
 bgw_job_stat_tuple_mark_start(TupleInfo *ti, void *const data)
 {
-	HeapTuple tuple = heap_copytuple(ti->tuple);
-	FormData_bgw_job_stat *fd = (FormData_bgw_job_stat *) GETSTRUCT(tuple);
+	bool should_free;
+	HeapTuple tuple = ExecFetchSlotHeapTuple(ti->slot, false, &should_free);
+	HeapTuple new_tuple = heap_copytuple(tuple);
```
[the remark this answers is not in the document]
erimatnor: This was my first inclination as well, but the following comment on `ExecFetchSlotHeapTuple` made me opt out of this:

```
* NB: If materialize is true, modifications of the returned tuple are
 * allowed. But it depends on the type of the slot whether such modifications
 * will also affect the slot's contents. While that is not the nicest
 * behaviour, all such modifications are in the process of being removed.```

File: src/scanner.c (hunk 7 above)
Comment (erimatnor) [reply src/scanner.c:341]: Yes, this was also the reason I added it here. I figured a separate PR can clean this up if necessary

File: src/utils.h (hunk 9 above)
Comment (erimatnor) [reply src/utils.h:100]: I'll take a look

File: src/scanner.c [hunk 14]
```diff
@@ -23,45 +23,60 @@ enum ScannerType
...
 typedef struct Scanner
 {
-	Relation (*openheap)(InternalScannerCtx *ctx);
+	Relation (*openscan)(InternalScannerCtx *ctx);
 	ScanDesc (*beginscan)(InternalScannerCtx *ctx);
 	bool (*getnext)(InternalScannerCtx *ctx);
 	void (*endscan)(InternalScannerCtx *ctx);
-	void (*closeheap)(InternalScannerCtx *ctx);
+	void (*closescan)(InternalScannerCtx *ctx);
 } Scanner;
```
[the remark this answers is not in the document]
erimatnor: This matches the calls on the underlying scan API it wraps. 

endscan ends the scan, but doesn't close the relation. In other words, you can start another scan after endscan without having to reopen the relation.

src/bgw/job_stat.c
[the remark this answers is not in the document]
k-rus: So it wasn't only me who didn't understand :)

src/planner.c
k-rus: Thank you for clarifying!

PR #2084, opened 2020-07-14
Utility function and hypertables view

gayyappan:
Add hypertable_detailed_size and chunk_detailed_size functions.
Remove size functionality from timescaledb_information.hypertable view and rename the view
as timescaledb_information.hypertables.

To reviewers:
chunk_relation_size and variants will be removed in another PR. 
All internal views and functions have been added to _timescaledb_internal. External facing views are in timescaledb_information. External facing functions are not schema prefixed.

### DIFF
--- a/sql/size_utils.sql
+++ b/sql/size_utils.sql
@@ -5,7 +5,68 @@
 -- This file contains utility functions to get the relation size
 -- of hypertables, chunks, and indexes on hypertables.
 
-CREATE OR REPLACE FUNCTION _timescaledb_internal.hypertable_relation_local_size(
+CREATE OR REPLACE VIEW _timescaledb_internal.hypertable_chunk_local_size AS 
+SELECT
+   h.schema_name,
+   h.table_name,
+   h.id as hypertable_id,
+   c.id as chunk_id,
+   c.schema_name as chunk_schema,
+   c.table_name as chunk_name,
+   pg_total_relation_size(format('%I.%I', c.schema_name, c.table_name))::bigint AS total_bytes,
+   pg_indexes_size(format('%I.%I', c.schema_name, c.table_name))::bigint AS index_bytes,
+   pg_total_relation_size(reltoastrelid)::bigint AS toast_bytes,
+   map.compressed_heap_size,
+   map.compressed_index_size,
+   map.compressed_toast_size 
+FROM
+   _timescaledb_catalog.hypertable h 
+   INNER JOIN
+      _timescaledb_catalog.chunk c 
+      ON h.id = c.hypertable_id 
+      and c.dropped = false 
+   INNER JOIN
+      pg_class pgc 
+      ON pgc.relname = h.table_name 
+   INNER JOIN
+      pg_namespace pns 
+      ON pns.oid = pgc.relnamespace 
+      AND pns.nspname = h.schema_name 
+   LEFT OUTER JOIN
+      _timescaledb_catalog.compression_chunk_size map 
+      ON map.chunk_id = c.id 
+WHERE pgc.relkind = 'r';
+
+GRANT SELECT ON  _timescaledb_internal.hypertable_chunk_local_size TO PUBLIC;
+ 
+CREATE OR REPLACE FUNCTION _timescaledb_internal.data_node_hypertable_info(
+    node_name              NAME,
+    schema_name_in name,
+    table_name_in name
+)
+RETURNS TABLE (
+    table_bytes     bigint,
+    index_bytes     bigint,
+    toast_bytes     bigint,
+    total_bytes     bigint)
+AS '@MODULE_PATHNAME@', 'ts_dist_remote_hypertable_info' LANGUAGE C VOLATILE STRICT;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.data_node_chunk_info(
+    node_name              NAME,
+    schema_name_in name,
+    table_name_in name
+)
+RETURNS TABLE (
+    chunk_id        integer,
+    chunk_schema    name,
+    chunk_name      name,
+    table_bytes     bigint,
+    index_bytes     bigint,
+    toast_bytes     bigint,
+    total_bytes     bigint)
+AS '@MODULE_PATHNAME@', 'ts_dist_remote_chunk_info' LANGUAGE C VOLATILE STRICT;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.hypertable_local_size(
     schema_name_in name,
     table_name_in name)
 RETURNS TABLE (
@@ -16,48 +77,47 @@ RETURNS TABLE (
 LANGUAGE PLPGSQL STABLE STRICT AS
 $BODY$
 BEGIN
-    RETURN QUERY
-    SELECT
-        sub2.table_bytes,
-        sub2.index_bytes,
-        sub2.toast_bytes,
-        sub2.total_bytes
-    FROM (
-        SELECT
+   RETURN QUERY 
+   SELECT
+      (sub2.table_bytes + sub2.compressed_heap_bytes)::bigint as heap_bytes,
+      (sub2.index_bytes + sub2.compressed_index_bytes)::bigint as index_bytes,
+      (sub2.toast_bytes + sub2.compressed_toast_bytes)::bigint as toast_bytes,
+      (sub2.total_bytes + sub2.compressed_heap_bytes + sub2.compressed_index_bytes + sub2.compressed_toast_bytes)::bigint as total_bytes 
+   FROM
+      (
+         SELECT
             *,
-            sub1.total_bytes - sub1.index_bytes - COALESCE(sub1.toast_bytes, 0) AS table_bytes
-        FROM (
-            SELECT
-                sum(pg_total_relation_size(format('%I.%I', c.schema_name, c.table_name)))::bigint AS total_bytes,
-                sum(pg_indexes_size(format('%I.%I', c.schema_name, c.table_name)))::bigint AS index_bytes,
-                sum(pg_total_relation_size(reltoastrelid))::bigint AS toast_bytes
-            FROM
-                _timescaledb_catalog.hypertable h,
-                _timescaledb_catalog.chunk c,
-                pg_class pgc,
-                pg_namespace pns
-            WHERE
-                h.schema_name = schema_name_in
-                AND h.table_name = table_name_in
-                AND c.dropped = FALSE
-                AND c.hypertable_id = h.id
-                AND pgc.relname = h.table_name
-                AND pns.oid = pgc.relnamespace
-                AND pns.nspname = h.schema_name
-                AND relkind = 'r'
-             ) sub1
-         ) sub2;
+            sub1.total_bytes - sub1.index_bytes - sub1.toast_bytes AS table_bytes 
+         FROM
+            (
+               SELECT
+                  sum(ch.total_bytes) as total_bytes,
+                  COALESCE( sum(ch.index_bytes) , 0 ) as index_bytes,
+                  COALESCE( sum(ch.toast_bytes), 0 ) as toast_bytes,
+                  COALESCE( sum(ch.compressed_heap_size) , 0 ) as compressed_heap_bytes,
+                  COALESCE( sum(ch.compressed_index_size) , 0) as compressed_index_bytes,
+                  COALESCE( sum(ch.compressed_toast_size) , 0 ) as compressed_toast_bytes 
+               FROM
+                  _timescaledb_internal.hypertable_chunk_local_size ch 
+               WHERE
+                  schema_name = schema_name_in 
+                  AND table_name = table_name_in 
+               GROUP BY
+                  hypertable_id 
+            ) sub1 
+      ) sub2;
 END;
 $BODY$;
 
-CREATE OR REPLACE FUNCTION _timescaledb_internal.hypertable_relation_remote_size(
+CREATE OR REPLACE FUNCTION _timescaledb_internal.hypertable_remote_size(
     schema_name_in name,
     table_name_in name)
 RETURNS TABLE (
     table_bytes bigint,
     index_bytes bigint,
     toast_bytes bigint,
-    total_bytes bigint)
+    total_bytes bigint,
+    node_name   NAME)
 LANGUAGE PLPGSQL STABLE STRICT AS
 $BODY$
 BEGIN
@@ -66,11 +126,12 @@ BEGIN
         sum(entry.table_bytes)::bigint AS table_bytes,
         sum(entry.index_bytes)::bigint AS index_bytes,
         sum(entry.toast_bytes)::bigint AS toast_bytes,
-        sum(entry.total_bytes)::bigint AS total_bytes
+        sum(entry.total_bytes)::bigint AS total_bytes,
+        srv.node_name
     FROM (
         SELECT
             s.node_name,
-            _timescaledb_internal.ping_data_node (node_name) AS node_up
+            _timescaledb_internal.ping_data_node (s.node_name) AS node_up
         FROM
             _timescaledb_catalog.hypertable AS ht,
             _timescaledb_catalog.hypertable_data_node AS s
@@ -79,21 +140,18 @@ BEGIN
             AND ht.table_name = table_name_in
             AND s.hypertable_id = ht.id
          ) AS srv
-    LEFT OUTER JOIN LATERAL @extschema@.data_node_hypertable_info(
+    LEFT OUTER JOIN LATERAL _timescaledb_internal.data_node_hypertable_info(
     CASE WHEN srv.node_up THEN
         srv.node_name
     ELSE
         NULL
-    END) entry ON TRUE
-    WHERE
-        entry.table_schema = schema_name_in
-        AND entry.table_name = table_name_in;
+    END, schema_name_in, table_name_in) entry ON TRUE
+    GROUP BY srv.node_name;
 END;
 $BODY$;
 
 -- Get relation size of hypertable
 -- like pg_relation_size(hypertable)
--- (https://www.postgresql.org/docs/9.6/static/functions-admin.html#FUNCTIONS-ADMIN-DBSIZE)
 --
 -- main_table - hypertable to get size of
 --
@@ -103,13 +161,14 @@ $BODY$;
 -- toast_bytes        - Disk space of toast tables
 -- total_bytes        - Total disk space used by the specified table, including all indexes and TOAST data
 
-CREATE OR REPLACE FUNCTION hypertable_relation_size(
+CREATE OR REPLACE FUNCTION hypertable_detailed_size(
     main_table              REGCLASS
 )
 RETURNS TABLE (table_bytes BIGINT,
                index_bytes BIGINT,
                toast_bytes BIGINT,
-               total_bytes BIGINT
+               total_bytes BIGINT,
+               node_name   NAME
                ) LANGUAGE PLPGSQL STABLE STRICT
                AS
 $BODY$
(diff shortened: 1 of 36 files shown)


k-rus:
@gayyappan Do you plan to keep 4 commits? The commit messages don't follow our guidelines.
I see that the PR fails on `view` test. Can the failing test be resolved already now?

sql/size_utils.sql
```diff
@@ -126,151 +184,280 @@ BEGIN
...
+               FROM @extschema@.hypertable_detailed_size(main_table) hd;
 
+END;
+$BODY$;
 
-CREATE OR REPLACE FUNCTION _timescaledb_internal.partitioning_column_to_pretty(
-    d   _timescaledb_catalog.dimension
-)
-    RETURNS TEXT LANGUAGE PLPGSQL STABLE STRICT AS
+CREATE OR REPLACE FUNCTION _timescaledb_internal.chunk_local_size(
```
erimatnor, line 231: A couple of notes:

* it sounds like this takes a chunk as input, but I think it takes a hypertable. This is confusing.
* why is the input schema+name and not regclass?
* Why does it return chunk_id? I guess this is fine for an internal functions, but not clear why we need it.
* Why do we return only name and not schema (or regclass).

More accurate names would be `chunks_local_size`, `local_chunks_size` or `hypertable_local_chunks_size`

sql/size_utils.sql
```diff
@@ -126,151 +184,280 @@ BEGIN
...
+   _timescaledb_internal.hypertable_chunk_local_size ch 
+   WHERE
+      ch.schema_name = schema_name_in 
+      AND ch.table_name = table_name_in 
+; 
+END;
 $BODY$;
 
+---should return same information as chunk_local_size--
+CREATE OR REPLACE FUNCTION _timescaledb_internal.chunk_remote_size(
```
erimatnor, line 262: Same here as above. Why schema+name and not regclass as parameter. Intuitively, I would think this takes a chunk as input, but I don't think this is the case.

sql/size_utils.sql
gayyappan: Will rename this. Intended to be used by both distributed (remote calls this internally) and local versions. So we pass schema_name + table_name explicitly instead of regclass. 
Also, this relates to efficient access of the catalog tables. We get index access only if we use table_name. Oids are not present in the catalog tables.  Since this is an internal function, I would not worry about this now - can be modified at a later date if use an oid.

sql/size_utils.sql
```diff
@@ -98,18 +155,19 @@ $BODY$;
 -- main_table - hypertable to get size of
 --
 -- Returns:
--- table_bytes        - Disk space used by main_table (like pg_relation_size(main_table))
+-- heap_bytes        - Disk space used by main_table (like pg_relation_size(main_table))
```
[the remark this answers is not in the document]
gayyappan: There was no objection to heap_bytes earlier.  Makes sense now that PG is moving away from heap terminology. renamed as table_bytes.

sql/size_utils.sql
gayyappan: To be deprecated function chunk_relation_size returns chunk_id. This is useful for  ordering the output. chunk_id gives seq ordering of the chunks. chunk_name has textual ordering and that does not match.

sql/size_utils.sql
gayyappan: renamed as chunks_remote_size

sql/size_utils.sql
```diff
@@ -126,151 +184,280 @@ BEGIN
...
--- table_size         - Pretty output of table_bytes for the data node
--- index_bytes        - Pretty output of index_bytes for the data node
--- toast_bytes        - Pretty output of toast_bytes for the data node
--- total_size         - Pretty output of total_bytes for the data node
-
-CREATE OR REPLACE FUNCTION hypertable_data_node_relation_size(
+-- index_name           - index on hyper table
+-- total_bytes          - size of index on disk
+
+CREATE OR REPLACE FUNCTION indexes_relation_size(
```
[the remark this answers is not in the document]
gayyappan: We already have the aggregated index information in hypertable_detailed_size and chunks_detailed_size. The only reason I have retained it is because it breaks it down per index.

sql/size_utils.sql
```diff
@@ -469,79 +656,34 @@ BEGIN
...
---
--- main_table - hypertable to get index sizes of
---
--- Returns:
--- index_name           - index on hyper table
--- total_bytes          - size of index on disk
-
-CREATE OR REPLACE FUNCTION indexes_relation_size(
-    main_table              REGCLASS
+CREATE OR REPLACE FUNCTION _timescaledb_internal.range_value_to_pretty(
```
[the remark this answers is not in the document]
gayyappan: This will be removed by a later PR when chunk_relation_size functions are removed.

gayyappan:
> A quick summary of high-level points:
> 
> * My biggest gripe is with the chunk-related size functions. The names of most of these functions imply that they return the size of a given chunk, but they really take a hypertable as input. This is confusing. I think there are some naming issues to figure out here to make this more clear. (Also, I didn't get you comment about removing these functions later. I guess I don't recall what our decision was, but they're still listed in the design doc.)

Renamed the external facing function as chunks_detailed_size based on the team discussion.
Existing chunks_relation-sizexxx will be removed as part of another PR.

> * There's still confusion and inconsistencies in use of regclass vs. schema+name (in particular for function parameters). Sometimes we also return just the table name, but no schema, which is not useful.

All the external facing functions use regclass.  (I responded why it is better to use table_name for the internal functions  while addressing the individual comment.)

> * I lack the simple size functions (that return only bigint), but I guess the intention might be to add those later?

Yes, to be added in another PR.

sql/size_utils.sql
```diff
@@ -126,151 +184,280 @@ BEGIN
...
         ELSE
-            RETURN QUERY SELECT * FROM _timescaledb_internal.hypertable_relation_local_size(schema_name, table_name);
+            RETURN QUERY SELECT *, NULL::name FROM _timescaledb_internal.hypertable_local_size(schema_name, table_name);
         END CASE;
 END;
 $BODY$;
 
+-- Get relation size of hypertable
+-- like pg_relation_size(hypertable)
+-- (https://www.postgresql.org/docs/9.6/static/functions-admin.html#FUNCTIONS-ADMIN-DBSIZE)
```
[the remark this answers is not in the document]
gayyappan: fixed

sql/size_utils.sql
```diff
@@ -126,151 +184,280 @@ BEGIN
...
+    ELSE
+        NULL
+    END , schema_name_in, table_name_in) entry ON TRUE;
+END;
+$BODY$;
 
--- Get relation size of hypertable
--- like pg_relation_size(hypertable)
+-- Get relation size of the chunks of an hypertable
+-- like pg_relation_size
```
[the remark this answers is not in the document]
gayyappan: fixed

sql/size_utils.sql
```diff
@@ -126,151 +184,280 @@ BEGIN
...
+-- index_bytes                   - Disk space used by indexes
+-- toast_bytes                   - Disk space of toast tables
+-- total_bytes                   - Disk space used in total
 
-CREATE OR REPLACE FUNCTION hypertable_relation_size_pretty(
+CREATE OR REPLACE FUNCTION chunk_detailed_size(
     main_table              REGCLASS
 )
-RETURNS TABLE (table_size  TEXT,
+RETURNS TABLE (chunk_id INT,
```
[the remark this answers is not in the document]
gayyappan: I think the schema name is redundant here. We have that information in the "chunks" view. This is just supposed to give size information for chunks. We identify it by chunk_id and/or name.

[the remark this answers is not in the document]
erimatnor: I don't think the function should return the internal ID since it doesn't mean anything to the user. The canonical identifier is schema+name, which I think is what we should use when we don't use OIDs. (Also because this is what you need to resolve the OID.)

Another reason for not returning the internal ID is because it is inconsistent with other functions. For instance, neither `hypertable_detailed_size` nor `indexes_detailed_size` return such IDs.

sql/size_utils.sql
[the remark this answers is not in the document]
erimatnor: * Still see "relation" in the name. 
* Why is the index name returned as a TEXT field and not schema+name? This is yet a different way of returning this information and it is not consistent with neither `chunks_detailed_size` nor `hypertable_detailed_size`.

sql/size_utils.sql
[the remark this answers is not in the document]
erimatnor: Yeah, I don't mind that internal functions (under `_timescaledb_internal`) returns IDs if it is needed for some reason, but public functions should only deal with official identifiers such as schema+name or OIDs.

tsl/src/dist_util.c
```diff
@@ -183,65 +186,34 @@ Datum
...
-	Assert(!PG_ARGISNULL(0)); /* Strict function */
+	StringInfo query_str = makeStringInfo();
+	/* Strict function */
+	Assert(!PG_ARGISNULL(0) && !PG_ARGISNULL(1) && !PG_ARGISNULL(2));
+	Name schema_name = PG_GETARG_NAME(1);
+	Name table_name = PG_GETARG_NAME(2);
+	appendStringInfo(query_str,
+					 "SELECT * from _timescaledb_internal.hypertable_local_size( '%s', '%s'  );",
+					 quote_identifier(NameStr(*schema_name)),
+					 quote_identifier(NameStr(*table_name)));
```
mkindahl, line 197: Function `quote_identifier` cannot be used here. If the identifier contain a single quote (`'`) you will get the strange output `'"foo'bar"'`. Use `quote_literal_cstr` instead.
```suggestion
	appendStringInfo(query_str,
					 "SELECT * from _timescaledb_internal.hypertable_local_size(%s, %s);",
					 quote_literal_cstr(NameStr(*schema_name)),
					 quote_literal_cstr(NameStr(*table_name)));
```

tsl/src/dist_util.c
```diff
@@ -183,65 +186,34 @@ Datum
...
+	char *node_name;
+	StringInfo query_str = makeStringInfo();
+	/* Strict function */
+	Assert(!PG_ARGISNULL(0) && !PG_ARGISNULL(1) && !PG_ARGISNULL(2));
+	Name schema_name = PG_GETARG_NAME(1);
+	Name table_name = PG_GETARG_NAME(2);
+	appendStringInfo(query_str,
+					 "SELECT * from _timescaledb_internal.chunks_local_size( '%s', '%s'  );",
+					 quote_identifier(NameStr(*schema_name)),
+					 quote_identifier(NameStr(*table_name)));
```
mkindahl, line 214:
```suggestion
	appendStringInfo(query_str,
					 "SELECT * from _timescaledb_internal.chunks_local_size(%s, %s);",
					 quote_literal_cstr(NameStr(*schema_name)),
					 quote_literal_cstr(NameStr(*table_name)));
```

tsl/src/dist_util.c
```diff
@@ -183,65 +186,34 @@ Datum
...
+	node_name = PG_GETARG_NAME(0)->data;
```
mkindahl, line 215:
```suggestion
	node_name = NameStr(*PG_GETARG_NAME(0));
```

tsl/src/dist_util.c
```diff
@@ -183,65 +186,34 @@ Datum
...
+	return dist_util_remote_srf_query(fcinfo, node_name, query_str->data);
```
mkindahl, line 216:
```suggestion
	return dist_util_remote_srf_query(fcinfo, node_name, NameStr(*query_str));
```

tsl/src/dist_util.c
```diff
@@ -338,3 +310,66 @@ dist_util_is_compatible_version(const char *data_node_version, const char *acces
...
+		TupleDesc tupdesc;
+
+		funcctx = SRF_FIRSTCALL_INIT();
+		oldcontext = MemoryContextSwitchTo(funcctx->multi_call_memory_ctx);
+
+		if (get_call_result_type(fcinfo, NULL, &tupdesc) != TYPEFUNC_COMPOSITE)
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("function returning record called in context "
+							"that cannot accept type record")));
```
mkindahl, line 338: Nit: this is so common in our code now that we might create a function for it.

tsl/src/dist_util.c
```diff
@@ -338,3 +310,66 @@ dist_util_is_compatible_version(const char *data_node_version, const char *acces
...
+
+		funcctx->user_fctx =
+			ts_dist_cmd_invoke_on_data_nodes(sql_query, list_make1((void *) node_name), true);
```
mkindahl, line 341: Cast to `void*` not needed (and can hide errors).
```suggestion
			ts_dist_cmd_invoke_on_data_nodes(sql_query, list_make1(node_name), true);
```

tsl/test/sql/compression.sql
```diff
@@ -176,7 +176,13 @@ vacuum full conditions;
...
 -- completely reclaimed. Sets it at 8K (page size). So a chunk which has
 -- been compressed still incurs an overhead of n * 8KB (for every index + toast table) storage on the original uncompressed chunk.
-select * from timescaledb_information.hypertable
+select pg_size_pretty(table_bytes), pg_size_pretty(index_bytes),
+pg_size_pretty(toast_bytes), pg_size_pretty(total_bytes) 
+from hypertable_detailed_size('foo');
+select pg_size_pretty(table_bytes), pg_size_pretty(index_bytes),
+pg_size_pretty(toast_bytes), pg_size_pretty(total_bytes) 
+from hypertable_detailed_size('conditions');
+select * from timescaledb_information.hypertables
```
mkindahl, line 185: Nit: very hard to read. Can you indent it?

tsl/src/dist_util.c
gayyappan: query_str is StringInfo.

tsl/src/dist_util.c
gayyappan: modified.

sql/size_utils.sql
```diff
@@ -5,7 +5,66 @@
...
+   LEFT OUTER JOIN
+      _timescaledb_catalog.compression_chunk_size map 
+      ON map.chunk_id = c.id ;
+
+GRANT SELECT ON  _timescaledb_internal.hypertable_chunk_local_size TO PUBLIC;
+ 
+CREATE OR REPLACE FUNCTION _timescaledb_internal.data_node_hypertable_info(
+    node_name              NAME,
+    schema_name_in name,
+    table_name_in name
```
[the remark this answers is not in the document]
gayyappan: will run this through formatter in follow-up PR.

sql/size_utils.sql
```diff
@@ -5,7 +5,66 @@
...
+      ON h.id = c.hypertable_id 
+      and c.dropped = false 
+   INNER JOIN
+      pg_class pgc 
+      ON pgc.relname = h.table_name 
+   INNER JOIN
+      pg_namespace pns 
+      ON pns.oid = pgc.relnamespace 
+      AND pns.nspname = h.schema_name 
+      AND relkind = 'r'
```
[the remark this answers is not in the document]
gayyappan: modified

tsl/src/dist_util.c
mkindahl: Ok. Just assumed it was `Name` since it referenced data.

sql/size_utils.sql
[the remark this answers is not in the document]
gayyappan: This PR modifies only adds chunk_detailed_size and hypertable_detailed_size information. and needed dependencies.  The indexes function came up as part of the PR review.The changes being suggested are more extensive than simply removing a function.  So I am going to deal with changes to indexes_relation_size in a follow-up PR.

sql/size_utils.sql
```diff
@@ -66,11 +124,12 @@ BEGIN
...
         sum(entry.index_bytes)::bigint AS index_bytes,
         sum(entry.toast_bytes)::bigint AS toast_bytes,
-        sum(entry.total_bytes)::bigint AS total_bytes
+        sum(entry.total_bytes)::bigint AS total_bytes,
+        srv.node_name
     FROM (
         SELECT
             s.node_name,
-            _timescaledb_internal.ping_data_node (node_name) AS node_up
+            _timescaledb_internal.ping_data_node (s.node_name) AS node_up
```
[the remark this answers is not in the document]
gayyappan: We query the node only if the ping is true. Does that address your concern?

sql/size_utils.sql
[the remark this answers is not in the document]
gayyappan: @WireBaron Good catch on the chunk_id. Removing it from the external function.

gayyappan:
> There are mainly these things I am concerned about:
> 
> * We have a ping inside a view, which might seem convenient but it will block a query if the remote node is down. This is largely something that I believe users will find unexpected, so I think it might be better to return a cached result and instead have a timestamp with the last known time a result set was returned. We might consider this as out of scope for this PR, but we should at least have a discussion about if this is what we want.

Doesn't the ping return false if the node is down. The query is passed down to the node only if the node is up.

> * Chunk names are returned, but no schema name for the chunk. Chunks can reside in other schemas than the default, so we should make the schema name available as well whenever we make a chunk name available.
> * Chunk id's are used, which I believe we decided that we should consider internal.

This has been removed.

> * Some incorrect usage of `quote_identifier`. Use `quote_literal_cstr` instead.
Could you point out where this is a concern?

sql/size_utils.sql
```diff
@@ -126,151 +183,201 @@ BEGIN
...
 --
 -- Returns:
--- table_size         - Pretty output of table_bytes
--- index_bytes        - Pretty output of index_bytes
--- toast_bytes        - Pretty output of toast_bytes
--- total_size         - Pretty output of total_bytes
-
-CREATE OR REPLACE FUNCTION hypertable_relation_size_pretty(
+-- chunk_id                      - Timescaledb id of a chunk
+-- chunk_table                   - Table used for the chunk
```
WireBaron, line 289:
```suggestion
-- chunk_name                    - Table used for the chunk
```

tsl/src/dist_util.c
gayyappan: needed , otherwise compiler errors.

sql/size_utils.sql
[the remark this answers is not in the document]
mkindahl: Not really, sorry. The problem is the ping itself, which will block waiting for a timeout, so the question is not about correctness, it is about responsiveness.

Note that I do not necessarily think we can solve this problem in this PR, so I would not consider it a blocker, but I think we need to discuss it.

sql/size_utils.sql
```diff
@@ -126,151 +183,197 @@ BEGIN
...
-
-CREATE OR REPLACE FUNCTION _timescaledb_internal.partitioning_column_to_pretty(
-    d   _timescaledb_catalog.dimension
-)
-    RETURNS TEXT LANGUAGE PLPGSQL STABLE STRICT AS
+CREATE OR REPLACE FUNCTION _timescaledb_internal.chunks_local_size(
+    schema_name_in name,
+    table_name_in name)
+RETURNS TABLE (
+    chunk_id    integer,
```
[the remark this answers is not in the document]
gayyappan: chunk id removed from alle xternal functions and replaced by schema_name + table_name

sql/size_utils.sql
```diff
@@ -126,151 +183,197 @@ BEGIN
...
-               pg_size_pretty(toast_bytes) as toast,
-               pg_size_pretty(total_bytes) as total
-            FROM timescaledb_information.data_node s
-                LEFT OUTER JOIN LATERAL @extschema@.data_node_hypertable_info(s.node_name) size ON TRUE
-            WHERE size.table_schema = %L
-            AND size.table_name = %L;
-        $$,
-        local_schema_name, local_table_name);
+        RETURN QUERY
+        SELECT format('%I.%I', h.schema_name, ci.hypertable_index_name),
```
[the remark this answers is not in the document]
gayyappan: This is an existing function . Not intended to be modified by this PR>

tsl/src/dist_util.c
gayyappan: Good catch. fixed and added a test.

PR #2089, opened 2020-07-15
Add compression settings view

gayyappan:
Add informational view that lists the settings
used while enabling compression on a hypertable.

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -259,5 +259,28 @@ CREATE OR REPLACE VIEW timescaledb_information.data_node AS
     LEFT OUTER JOIN LATERAL @extschema@.data_node_hypertable_info(CASE WHEN s.node_up THEN s.node_name ELSE NULL END) size ON TRUE
   GROUP BY s.node_name, s.node_up, s.owner, s.options;
 
+---compression parameters information ---
+CREATE VIEW timescaledb_information.compression_settings
+AS
+SELECT
+    ht.schema_name,
+    ht.table_name,
+    segq.attname,
+    segq.segmentby_column_index,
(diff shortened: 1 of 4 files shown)


sql/views.sql
```diff
@@ -259,5 +259,40 @@ CREATE OR REPLACE VIEW timescaledb_information.data_node AS
...
+SELECT
+   ht.schema_name,
+   ht.table_name,
+   segq.attname,
+   segq.segmentby_column_index,
+   segq.orderby_column_index,
+   segq.orderby_asc,
+   segq.orderby_nullsfirst 
+FROM
+   (
```
svenklemm, line 274: should get rid of subselect and also run pg_format on the query

gayyappan: don't care for the sub-select myself. But is there any specific reason why it should not be used (didn't see any plan issues).

tsl/test/sql/compression.sql
```diff
@@ -29,6 +29,7 @@ SET timescaledb.enable_transparent_decompression to OFF;
 select id, schema_name, table_name, compressed, compressed_hypertable_id from
 _timescaledb_catalog.hypertable order by id;
 select * from _timescaledb_catalog.hypertable_compression order by hypertable_id, attname;
+select * from timescaledb_information.compression_settings ;
```
svenklemm, line 32: needs ORDER BY clause

gayyappan: The view already has ORDER BY.

gayyappan:
> Looks good once you add the ORDER BY requested by Sven.

The view definition already has the ORDER BY.

PR #2092, opened 2020-07-16
Fix delete on tables involving hypertables with compression

### LINKED ISSUE #2091: DELETE fails  when the query references compressed chunk on another table.
** To reproduce **
CREATE SCHEMA prom_data;
CREATE SCHEMA prom_data_SERIES;
drop TABLE prom_data_SERIES.pg_settings_max_sync_workers_per_subscription;
drop TABLE prom_data.pg_settings_max_sync_workers_per_subscription;

CREATE TABLE prom_data_SERIES.pg_settings_max_sync_workers_per_subscription(
time       timestamp with time zone,
 value      double precision   ,
 series_id  integer);

select create_hypertable('prom_data_SERIES.pg_settings_max_sync_workers_per_subscription', 'time', chunk_time_interval=>'1 day'::interval);

CREATE INDEX ON prom_data_SERIES.pg_settings_max_sync_workers_per_subscription(
time, series_id);

INSERT into prom_data_SERIES.pg_settings_max_sync_workers_per_subscription values
( '2020-04-20 01:01', 100, 1),
( '2020-05-20 01:01', 100, 1),
( '2020-04-20 01:01', 200, 2);

============================================================
Now create the table that will be compressed.
CREATE TABLE prom_data.pg_settings_max_sync_workers_per_subscription(
time       timestamp with time zone,
 value      double precision   ,
 series_id  integer);

select create_hypertable('prom_data.pg_settings_max_sync_workers_per_subscription', 'time', chunk_time_interval=>'1 day'::interval);

ALTER TABLE prom_data.pg_settings_max_sync_workers_per_subscription
SET(timescaledb.compress);

CREATE INDEX ON prom_data.pg_settings_max_sync_workers_per_subscription(
time, series_id);

INSERT into prom_data.pg_settings_max_sync_workers_per_subscription values
( '2020-04-20 01:01', 100, 1),
( '2020-05-20 01:01', 100, 1);

--compress chunks  in this table ------
select compress_chunk('_timescaledb_internal._hyper_28_49_chunk');

==========================================================

Now run this query:
test3=# With confirmed_drop_series AS (
                 SELECT series_id
                 FROM  prom_data.pg_settings_max_sync_workers_per_subscription  data_exists
                 WHERE time >= '2020-04-17 17:14:24.161989+00'
        )
          DELETE from prom_data_SERIES.pg_settings_max_sync_workers_per_subscription
          WHERE series_id IN (SELECT series_id FROM confirmed_drop_series);
ERROR:  cannot update/delete rows from chunk "_hyper_28_49_chunk" as it is compressed
@cevian reported on 1.7.2.

svenklemm:
The DML blocker to block INSERTs and UPDATEs on compressed hypertables
would trigger if the UPDATE or DELETE referenced any hypertable with
compressed chunks. This patch changes the logic to only block if the
target of the UPDATE or DELETE is a compressed hypertable.

Fixes #2091


mfreed:
@svenklemm Quick clarification:  Should the commit message say:

```
target of the UPDATE or DELETE is a compressed CHUNK WITHIN A hypertable
```

(Don't mean to be pedantic, just wasn't sure if this is exactly what you meant, i.e., does it only block if it would other DELETE/UPDATE a row that falls in a chunk currently compressed, or could it block under different scenarios as well.)

svenklemm:
> @svenklemm Quick clarification: Should the commit message say:
> 
> ```
> target of the UPDATE or DELETE is a compressed CHUNK WITHIN A hypertable
> ```
> 
> (Don't mean to be pedantic, just wasn't sure if this is exactly what you meant, i.e., does it only block if it would other DELETE/UPDATE a row that falls in a chunk currently compressed, or could it block under different scenarios as well.)

This part of the logic is not changed, so your suggested change is correct.

tsl/src/nodes/decompress_chunk/exec.c
```diff
@@ -188,8 +188,10 @@ constify_tableoid_walker(Node *node, ConstifyTableOidContext *ctx)
 		 * we doublecheck system columns here because projection will
 		 * segfault if any system columns get through
 		 */
-		if (var->varattno < 0)
-			elog(ERROR, "transparent decompression only supports tableoid system column");
+		if (var->varattno < SelfItemPointerAttributeNumber)
+			elog(ERROR,
+				 "transparent decompression only supports tableoid system column %d",
```
cevian, line 193: this message should probably be reverted or made more explcit that the %d attno is not supported

src/planner.c
```diff
@@ -726,6 +726,18 @@ valid_hook_call(void)
 	return ts_extension_is_loaded() && planner_hcache_exists();
 }
 
+static bool
+dml_involves_hypertable(PlannerInfo *root, Hypertable *ht, Index rti)
```
cevian, line 730: does this still throw an error if the hypertable delete is inside a CTE. ie `WITH cte as (DELETE FROM compressed_ht RETURNING *) SELECT * FROM CTE` I ask because we get the resultRelation from the root and I'm not sure if thats the root of the query or the cte. In any case I don't think we have a test for that, but should.

svenklemm: root will be the root of the cte when planning the subquery, i've added a test

PR #2094, opened 2020-07-17
Fetch collations from data nodes during ANALYZE

### LINKED ISSUE #2093: Distributed hypertables don't update statistic collations during ANALYZE in PG12
 - OS: all
 - TimescaleDB version (output of `\dx` in `psql`): 2.0-dev
 - Installation method: source

**Describe the bug**
When pulling in statistics from data nodes, the timescale extensions update the pg_stacollN fields which were introduced in Postgresql 12.

**To Reproduce**
1. Create a distributed hypertable with a text field.
2. Add some data to the table (a single row is sufficient)
3. Analyze the table.
4. Try querying with a filter on the text field (i.e. WHERE foo ='bar')

**Expected behavior**
Query runs normally and returns matching rows.

**Actual behavior**
Query fails with `ERROR:  cache lookup failed for collation NNNN`

WireBaron:
This change fixes the stats collecting code to also return the slot
collation fields for PG12. This fixes a bug (#2093) where running an
ANALYZE in PG12 would break queries on distributed tables.

Fixes #2093


test/runner.sh
```diff
@@ -13,9 +13,11 @@ TEST_OUTPUT_DIR=${TEST_OUTPUT_DIR:-${EXE_DIR}}
 TEST_SUPPORT_FILE=${CURRENT_DIR}/sql/utils/testsupport.sql
 
 # PGAPPNAME will be 'pg_regress/test' so we cut off the prefix
-# to get the name of the test
+# to get the name of the test. To ensure TEST_DBNAME is a valid
+# postgres identifier, replace hyphens with underscores (this is
+# particularly important for template tests)
```
svenklemm, line 18: This is not correct, `-` is a valid character in an identifier otherwise all the template tests wouldnt work. psql can quote variables for you like so: `:"TEST_DBNAME"`

PR #2095, opened 2020-07-17
Add test of drop_chunks on distributed hypertable

k-rus:
Testing that drop_chunks works correctly on a distributed hypertable.
Tests of different arguments are assumed to be done on a usual
hypertable previously.

### DIFF
--- a/tsl/test/expected/dist_api_calls.out
+++ b/tsl/test/expected/dist_api_calls.out
@@ -1,0 +1,207 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+-- Basic testing of API functions on distributed hypertable
+-- Need to be super user to create extension and add data nodes
+\c :TEST_DBNAME :ROLE_CLUSTER_SUPERUSER;
+-- Support for execute_sql_and_filter_server_name_on_error()
+\unset ECHO
+psql:include/remote_exec.sql:5: NOTICE:  schema "test" already exists, skipping
+-- Cleanup from other potential tests that created these databases
+SET client_min_messages TO ERROR;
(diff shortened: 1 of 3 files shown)


tsl/test/expected/dist_api_calls.out
```diff
@@ -0,0 +1,207 @@
...
+-------------+-----------+-------+-------------+--------------+------------------+-------------------
+ data_node_2 | localhost | 55432 | data_node_2 | t            | t                | t
+(1 row)
+
+SELECT * FROM add_data_node('data_node_3', host => 'localhost',
+                            database => 'data_node_3');
+  node_name  |   host    | port  |  database   | node_created | database_created | extension_created 
+-------------+-----------+-------+-------------+--------------+------------------+-------------------
+ data_node_3 | localhost | 55432 | data_node_3 | t            | t                | t
+(1 row)
```
svenklemm, line 36: can you do the add_data_node calls in a single query so the setup is a bit more compact like so:
```
SELECT node_name,
  host,
  database,
  node_created,
  database_created,
  extension_created
FROM ( VALUES ('data_node_1'), ('data_node_2'), ('data_node_3')) node(name),
  LATERAL (SELECT * FROM add_data_node (node.name, host => 'localhost', database => node.name)) call;
```
or with similar generator statement, also you should probably not include port as that might not be the same for all test environments

k-rus: The suggested compact query to setup data nodes is difficult to follow. In such case I believe we should move setup of distributed database into a separate SQL test include file. Even better to have test fuzzing, which will setup distributed database.
Currently, I follow the current practice, which I agree is not the best.

k-rus: I don't see an easy and nice way to improve here.

PR #2100, opened 2020-07-18
Analyze ignore_invalidation_older_than for default value

### LINKED ISSUE #1921: drop_chunks() Fails against hypertable with cagg defined with no ignore_invalidation_older set
 - OS: Centos 7
 - PostgreSQL version (output of `postgres --version`):  PostgreSQL 11.8 on x86_64-pc-linux-gnu
 - TimescaleDB version (output of `\dx` in `psql`): 1.7.1
 - Installation method: [e.g., "using Docker", "apt install", "source"] Yum

**Describe the bug**
Calling `SELECT count(*)
FROM drop_chunks(INTERVAL ‘3 days’, ‘some_hypertable’, cascade_to_materializations => FALSE);` against a hypertable that has a cagg defined against it with no `ignore_invalidation_older_than` set.

It works as expected if `cascade_to_materializations` is TRUE

This is problematic, because `ignore_invalidation_older_than` is not required to create a cagg, and it is non-intuitive that this should fail if not cascading to the materalization.

**To Reproduce**
1. Create a hypertable.
2. Create a cagg on that hypertable, do NOT set `ignore_invalidation_older_than` on the cagg.
3. Execute a drop_chunks against the hypertable, setting `cascade_to_materializations` to FALSE in the call : `drop_chunks(INTERVAL ‘3 days’, ‘some_hypertable’, cascade_to_materializations => FALSE);`
4. See error: `ERROR:  older_than must be greater than the timescaledb.ignore_invalidation_older_than parameter of public.some_cagg`

**Expected behavior**
The expected chunks would be dropped from the hypertable.

**Actual behavior**
The call fails with the following error: `ERROR:  older_than must be greater than the timescaledb.ignore_invalidation_older_than parameter of public.some_cagg`

Olernov:
Analyze ignore_invalidation_older_than parameter of a continuous
aggregate for default value (when this parameter is not set
during the continous aggregate creation). This allows to drop
chunks without having an error.

Fixes #1921

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -3192,7 +3192,8 @@ ts_chunk_drop_process_materialization(Oid hypertable_relid,
 														 ignore_invalidation_older_than);
 
 	/* minimum_invalidation_time is inclusive; older_than_time is exclusive */
-	if (minimum_invalidation_time < older_than_time)
+	if (ignore_invalidation_older_than != PG_INT64_MAX /* if this parameter is set for cagg*/
+		&& minimum_invalidation_time < older_than_time)
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 				 errmsg("older_than must be greater than the "
--- a/tsl/test/expected/continuous_aggs_ddl.out
+++ b/tsl/test/expected/continuous_aggs_ddl.out
@@ -573,7 +573,7 @@ AS SELECT time_bucket('5', time), max(data)
 INSERT INTO drop_chunks_table SELECT i, i FROM generate_series(0, 20) AS i;
 \set ON_ERROR_STOP 0
 SELECT drop_chunks('drop_chunks_table', older_than => 13, cascade_to_materializations => false);
-ERROR:  older_than must be greater than the timescaledb.ignore_invalidation_older_than parameter of public.drop_chunks_view
+ERROR:  the continuous aggregate public.drop_chunks_view is too far behind
 ALTER VIEW drop_chunks_view SET (timescaledb.ignore_invalidation_older_than = 9);
 -- 9 is too small (less than timescaledb.ignore_invalidation_older_than)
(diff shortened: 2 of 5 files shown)


Olernov:
For information: file tsl/test/sql/continuous_aggs_drop_chunks.sql was already present in the repo but it was missing in CMakeLists.txt so the tests did not start. I added this file to CMakeLists.txt, fixed it a little and added new tests to it.

gayyappan:
@Olernov The current behavior is correct and as expected.  Chunks cannot be dropped from hypertables if there are continuous aggregates that depend on them. The ignore_invalidation_older_than setting is used to find the  chunks from the hypertable  that are required by the continuous agg.

Olernov:
But #1921 is marked as "bug", isn't it? We are considering the case when "cascade_to_materializations => FALSE" so maybe it makes sense to allow dropping chunks when changes are not propagated to the continuous aggregates?

gayyappan:
@Olernov We are considering some changes to the API to address this instead : the confusion stems more from the API rather than the 

> But #1921 is marked as "bug", isn't it? We are considering the case when "cascade_to_materializations => FALSE" so maybe it makes sense to allow dropping chunks when changes are not propagated to the continuous aggregates?
We are considering addressing this via API changes : this is more related to confusion around expected behavior.

erimatnor:
@Olernov thanks for this contribution. As @gayyappan points out, the current behavior is by design, although we agree the UX isn't ideal. Therefore, we decided to change this behavior as part of a larger refactor of continuous aggregates that will be released in the 2.0 version. In fact, the `ignore_invalidation_older_than` parameter will no longer exist in this new version. 

Issue #1921 should probably not have been labeled as a "bug". I will close that issue, along with this PR, with a note on planned changes.

I hope this won't discourage you from submitting further contributions in the future. We will also try to be more clear about plans and how we classify issues.

Olernov:
No problem, @erimatnor, I completely understand this.

PR #2102, opened 2020-07-20
Use "cordon" terminology for blocking data nodes

erimatnor:
This change renames `block_new_chunks` and `allow_new_chunks` to
`cordon_data_node` and `uncordon_data_node`. It also cleans up some
error messages related to this functionality. The renaming is
motivated by the following:

1. It makes the naming similar to other data node functions, such as
   `add_data_node` and `attach_data_node`. Having "data node" in the
   function name makes it clear that this is an operation involving
   data nodes.
2. Kubernetes uses "cordon"/"uncordon" for similar operations on
   nodes. Thus, there is a precedence for this terminology that
   creates familiarity for users.
3. It opens up for more advanced semantics rather than only blocking
   new chunks. For instance, we should also divert reads away from
   any cordoned nodes.

Any additional functionality will be implementated in follow-up
changes.


sql/ddl_api.sql
```diff
@@ -192,15 +192,27 @@ CREATE OR REPLACE FUNCTION detach_data_node(
...
+CREATE OR REPLACE FUNCTION cordon_data_node(
+    node_name NAME,
+    hypertable REGCLASS = NULL,
+    force BOOLEAN = FALSE
+) RETURNS INTEGER
+AS '@MODULE_PATHNAME@', 'ts_data_node_cordon' LANGUAGE C VOLATILE;
+
+-- Uncordon a data node across all distributed hypertables or a
+-- specific one when a hypertable is specified. If the data node was
+-- previously cordened, it will now be used to place new chunks and it
```
pmwkaa, line 209: typo: cordoned

sql/ddl_api.sql
```diff
@@ -192,15 +192,27 @@ CREATE OR REPLACE FUNCTION detach_data_node(
...
-CREATE OR REPLACE FUNCTION block_new_chunks(data_node_name NAME, hypertable REGCLASS = NULL, force BOOLEAN = FALSE) RETURNS INTEGER
-AS '@MODULE_PATHNAME@', 'ts_data_node_block_new_chunks' LANGUAGE C VOLATILE;
-
--- Reallow chunk creations on a blocked data node for a distributed hypertable. NULL hypertable means it will
--- allow chunks for all distributed hypertables
-CREATE OR REPLACE FUNCTION allow_new_chunks(data_node_name NAME, hypertable REGCLASS = NULL) RETURNS INTEGER
-AS '@MODULE_PATHNAME@', 'ts_data_node_allow_new_chunks' LANGUAGE C VOLATILE;
+-- Cordon a data node across all distributed hypertables or a specific
+-- one if a hypertable is specified. A cordoned node will not be used
+-- for queries or have any new chunks placed on it. The force
```
pmwkaa, line 197: I think this is slightly misleading, I guess cordoned chunk still can be used for queries, right?  Only new chunks will not be created on it but old data should still be accessible

erimatnor:
@svenklemm FWIW, cordon is used in, e.g., Kubernetes for similar purposes, so there's precedence and familiarity there. I also think it is actually a very accurate term for what we want to do, i.e., "to cordon something off",  and it  doesn't seem uncommon to me.

sql/ddl_api.sql
erimatnor: It shouldn't be used for queries, but that will be implemented separately. I can remove this comment in the meantime if you think it is important.

pmwkaa: I am ok to keep it, but we should do it before the release I guess. Our docs don't mention that either

erimatnor:
> The word "cordon" might be associated with negative things, so I would suggest the word "fence" instead, which is commonly used in distributed systems.

Thanks for the suggestion. I have never heard of "fence" in this context. Not sure it works as a verb either, e.g.,:`fence_data_node` doesn't sound good to me. 

As stated earlier, I think it is difficult, if not impossible, to find a name that is entirely descriptive so I am aiming for something that is at least accurate and memorable (or colorful).

mkindahl:
> > The word "cordon" might be associated with negative things, so I would suggest the word "fence" instead, which is commonly used in distributed systems.
> 
> Thanks for the suggestion. I have never heard of "fence" in this context. Not sure it works as a verb either, e.g.,:`fence_data_node` doesn't sound good to me.
> 
> As stated earlier, I think it is difficult, if not impossible, to find a name that is entirely descriptive so I am aiming for something that is at least accurate and memorable (or colorful).

https://en.wikipedia.org/wiki/Fencing_(computing)

erimatnor:
> > > The word "cordon" might be associated with negative things, so I would suggest the word "fence" instead, which is commonly used in distributed systems.
> > 
> > 
> > Thanks for the suggestion. I have never heard of "fence" in this context. Not sure it works as a verb either, e.g.,:`fence_data_node` doesn't sound good to me.
> > As stated earlier, I think it is difficult, if not impossible, to find a name that is entirely descriptive so I am aiming for something that is at least accurate and memorable (or colorful).
> 
> https://en.wikipedia.org/wiki/Fencing_(computing)

Fencing has different semantics and doesn't match what this command does, so it would be confusing to use. AFAIK, fencing is about isolating a malfunctioning node to avoid data corruption in the rest of the system.

sql/ddl_api.sql
k-rus: +1

erimatnor:
> The name is OK and I don't have a better proposal. The name is not intuitive and requires to check manual when it is seen the first time. Since it is not commonly used operation, it is totally fine. What is important, the name is not misleading.

I think you make an important point here. Requiring to check the manual is OK; in fact, it's probably something that is required no matter which name we choose since there is no name that can capture the full semantics. That's also why I prefer a "colorful" as long as it is not misleading.

tsl/src/data_node.c
```diff
@@ -970,21 +970,22 @@ data_node_detach_validate(const char *node_name, Hypertable *ht, bool force, Ope
...
 						operation,
 						node_name,
-						NameStr(ht->fd.table_name)),
-				 errhint("Ensure the data node \"%s\" has no non-replicated data before %s it.",
-						 node_name,
+						get_rel_name(ht->main_table_relid)),
+				 errdetail("The data node has the only replica of some data"
+						   " and therefore cannot be %s",
+						   operation),
+				 errhint("Make sure all chunks are fully replicated before %s the data node.",
```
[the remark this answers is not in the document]
erimatnor: I'll remove "fully". I guess it was an attempt to emphasize that it needs N replicas, i.e., 2 of 3 replicas are not enough, unless you force the operation.

tsl/test/sql/data_node.sql
```diff
@@ -459,37 +459,37 @@ SELECT * FROM _timescaledb_catalog.chunk_data_node;
 -- some ERROR cases
 \set ON_ERROR_STOP 0
 -- Will error due to under-replication
-SELECT * FROM block_new_chunks('data_node_2');
--- can't block/allow non-existing data node
-SELECT * FROM block_new_chunks('data_node_12345', 'disttable');
-SELECT * FROM allow_new_chunks('data_node_12345', 'disttable');
+SELECT * FROM cordon_data_node('data_node_2');
+-- can't cordon/uncordon non-existing data node
```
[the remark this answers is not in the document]
erimatnor: It tests both cordon and uncordon, which is why it said both. Would it work to say "Can't cordon and uncordon...."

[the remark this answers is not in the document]
k-rus: I am wonder if it is possible to pass `cordon` or `uncordon` instead of `cordon/uncordon`. I cannot find the source, which takes care of the error message.

erimatnor:
FWIW, Kubernetes doc on "cordon node": https://kubernetes.io/docs/concepts/architecture/nodes/#manual-node-administration

erimatnor:
> I dont think this should be named "cordon" but a more common name, i had to look up what cordon actually means.

@svenklemm Rejecting won't move this forward, so can you perhaps suggest alternatives?

k-rus:
@erimatnor We had discussion around it and there are two suggestions:
1. Drop this implementation for now as not so important for 2.0 release and blocking new chunks without blocking writes might not help with disk space, which was one of the reason for the functionality.
2. Use notion of data node mode, which can be set, e.g., to be in the mode of not creating new chunks, be in read-only mode, etc.
@mkindahl might already described it in more details.

mkindahl:
> @erimatnor We had discussion around it and there are two suggestions:
> 
>     1. Drop this implementation for now as not so important for 2.0 release and blocking new chunks without blocking writes might not help with disk space, which was one of the reason for the functionality.
> 
>     2. Use notion of data node mode, which can be set, e.g., to be in the mode of not creating new chunks, be in read-only mode, etc.
>        @mkindahl might already described it in more details.

It is added as issue #2214

erimatnor:
@k-rus While I think is fair to punt on this for 2.0, I am not sure I understand the reasoning.
 

In particular, I find reason 1 quite unconvincing. I don't see a use case forcompletely blocking writes since that would fail all inserts to a chunk on that node. So the only way to make that work without failing many writes is to move all data off of the node, or allow the node to get out of synch. Both these operations are well beyond what this functionality is intended to do, or I think should do.

For the same reason I don't understand what it would mean to put a node in, e.g.,  read-only mode, as suggested in option 2. Again, that would mean failing a lot of writes or letting the node get out of sync. I don't see a practical use case for any of these modes.

The current semantics of this command has a useful practical application: you can avoid the majority of new writes giving you enough time to upgrade storage or fix the node without failing many writes (or any). I think we should aim to solve such practical use cases, not try to imitate file mode settings without having a clear use case.

The current command is not intended to be used when a disk is already full, because at that point you're already in a bad spot and there's not much any command would do to help apart from just failing until storage is expanded.

k-rus:
@erimatnor I am reporting the suggestion from the team. It might not the best formulation. I personally think that the original name is better than suggested by this PR after the meeting.

erimatnor:
@k-rus I understand that wasn't necessarily your personal opinion. I am just giving my point of view in response to what you reported. Thanks for keeping me up-to-date.

k-rus:
> The current semantics of this command has a useful practical application: you can avoid the majority of new writes giving you enough time to upgrade storage or fix the node without failing many writes (or any).

I don't think it is of any use, since it still allows to write a lot of data into existing chunks. I doubt it is common case to rush with this functionality.

PR #2105, opened 2020-07-20
Support moving compressed chunks

pmwkaa:
Allow move_chunk() to work with uncompressed chunk and automatically move associated compressed chunk to specified
tablespace. Block move_chunk() execution for compressed chunks.

This change executes alter set tablespace underneath for compressed chunk. I guess it we might also use index used for orderby compression and call chunk_reorder() with it, not really sure if that would work.

Related issue: https://github.com/timescale/timescaledb/issues/2067

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -2765,19 +2765,34 @@ init_scan_by_compressed_chunk_id(ScanIterator *iterator, int32 compressed_chunk_
 								   Int32GetDatum(compressed_chunk_id));
 }
 
-bool
-ts_chunk_contains_compressed_data(Chunk *chunk)
+Chunk *
+ts_chunk_get_compressed_chunk_parent(Chunk *chunk)
 {
 	ScanIterator iterator = ts_scan_iterator_create(CHUNK, AccessShareLock, CurrentMemoryContext);
-	bool found = false;
+	Oid parent_id = InvalidOid;
 
 	init_scan_by_compressed_chunk_id(&iterator, chunk->fd.id);
 	ts_scanner_foreach(&iterator)
 	{
-		Assert(!found);
-		found = true;
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		bool isnull;
+
+		Assert(!OidIsValid(parent_id));
+		parent_id = DatumGetObjectId(heap_getattr(ti->tuple, Anum_chunk_id, ti->desc, &isnull));
 	}
-	return found;
+
+	if (OidIsValid(parent_id))
+		return ts_chunk_get_by_id(DatumGetObjectId(parent_id), true);
+
+	return NULL;
+}
+
+bool
+ts_chunk_contains_compressed_data(Chunk *chunk)
+{
+	Chunk *parent_chunk = ts_chunk_get_compressed_chunk_parent(chunk);
+
+	return parent_chunk != NULL;
 }
 
 List *
--- a/tsl/src/reorder.c
+++ b/tsl/src/reorder.c
@@ -121,6 +121,7 @@ tsl_move_chunk(PG_FUNCTION_ARGS)
 		PG_ARGISNULL(2) ? InvalidOid : get_tablespace_oid(PG_GETARG_NAME(2)->data, false);
 	Oid index_id = PG_ARGISNULL(3) ? InvalidOid : PG_GETARG_OID(3);
 	bool verbose = PG_ARGISNULL(4) ? false : PG_GETARG_BOOL(4);
+	Chunk *chunk;
 
 	/* used for debugging purposes only see finish_heap_swaps */
 	Oid wait_id = PG_NARGS() < 6 || PG_ARGISNULL(5) ? InvalidOid : PG_GETARG_OID(5);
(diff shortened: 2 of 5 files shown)


tsl/src/reorder.c
```diff
@@ -148,12 +149,38 @@ tsl_move_chunk(PG_FUNCTION_ARGS)
...
+
+	if (ts_chunk_contains_compressed_data(chunk))
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("\"%s\" is a compressed chunk", get_rel_name(chunk_id)),
+				 errhint("Please use the corresponding chunk on the uncompressed hypertable "
+						 "instead.")));
+
 	reorder_chunk(chunk_id,
 				  index_id,
```
gayyappan, line 167: @JLockerman Does reorder_chunk work with compressed chunks?

tsl/src/reorder.c
```diff
@@ -148,12 +149,38 @@ tsl_move_chunk(PG_FUNCTION_ARGS)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("\"%s\" is not a chunk", get_rel_name(chunk_id))));
+
+	if (ts_chunk_contains_compressed_data(chunk))
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("\"%s\" is a compressed chunk", get_rel_name(chunk_id)),
+				 errhint("Please use the corresponding chunk on the uncompressed hypertable "
+						 "instead.")));
```
erimatnor, line 164: This can be a potentially confusing error. Since the user can call `compress_chunk('A')`, I guess the user considers `A` to be compressed. But this error refers to the internal chunk (let's call it `B`) associated with `A` as compressed. It is also a bit odd to refer to the main hypertable as being "uncompressed". I don't think that's how the user thinks of the table given that this is the table they enabled compression on.

I think we need to be clear in the error that the operation was executed on an internal chunk and refer to `A`. For instance:

```suggestion
		ereport(ERROR,
				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
				 errmsg("cannot directly move internal compression data"),
				 errdetail("Chunk \"%s\" contains compressed data for chunk \"%s\" and cannot be moved directly.", get_rel_name(chunk_B), get_rel_name(chunk_A)),
				 errhint("Moving chunk \"%s\" will also move the compressed data.",
				 get_rel_name(chunk_A))));
```

tsl/src/reorder.c
```diff
@@ -148,12 +149,38 @@ tsl_move_chunk(PG_FUNCTION_ARGS)
...
+
```
erimatnor, line 165: The reorder will not do much when issued on the "main" chunk. I guess it will still move the empty chunk to the new tablespace. But I think this should be made clear in a comment and we should also consider whether we want to reorder the internal compressed chunk when possible. Maybe not for this PR, but in that case leave a comment on what is going on and that this is intentionally left out.

pmwkaa:
List of changes:

* Instead of using reorder_chunk() for compressed table, do ALTER SET TABLESPACE for both chunks
* Updated chunk search function and error message as suggested by @erimatnor

src/chunk.c
```diff
@@ -2766,18 +2766,29 @@ init_scan_by_compressed_chunk_id(ScanIterator *iterator, int32 compressed_chunk_
 }
 
 bool
-ts_chunk_contains_compressed_data(Chunk *chunk)
+ts_chunk_contains_compressed_data(Chunk *chunk, Chunk **parent)
 {
 	ScanIterator iterator = ts_scan_iterator_create(CHUNK, AccessShareLock, CurrentMemoryContext);
-	bool found = false;
+	Datum parent_id = 0;
```
erimatnor, line 2772: Is this a reliable and supported way of indicating a non-set Datum? I mean, it seems 0 is a valid value  for ,e.g., int datums.

src/chunk.c
```diff
@@ -2766,18 +2766,29 @@ init_scan_by_compressed_chunk_id(ScanIterator *iterator, int32 compressed_chunk_
...
-		Assert(!found);
-		found = true;
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		Assert(!parent_id);
+		bool isnull;
+		parent_id = heap_getattr(ti->tuple, Anum_chunk_id, ti->desc, &isnull);
 	}
-	return found;
+
+	if (parent_id)
```
erimatnor, line 2783: I would avoid these kinds of checks on integers, since it is ambiguous what is a true expression here, especially for a Datum type which can also have NULL values via the separate` isnull` check. I would just move the `isnull` out of the loop and use that here instead.

tsl/src/reorder.c
```diff
@@ -148,12 +150,45 @@ tsl_move_chunk(PG_FUNCTION_ARGS)
...
+						   "moved directly.",
+						   get_rel_name(chunk_id),
+						   get_rel_name(chunk_parent->table_id)),
+				 errhint("Moving chunk \"%s\" will also move the compressed data.",
+						 get_rel_name(chunk_parent->table_id))));
+
+	/* If chunk is compressed move it by altering tablespace on both chunks */
+	if (OidIsValid(chunk->fd.compressed_chunk_id))
+	{
+		Chunk *compressed_chunk = ts_chunk_get_by_id(chunk->fd.compressed_chunk_id, true);
```
erimatnor, line 174: For this case, do we simply ignore the index parameter given by the user. Is that parameter optional or always given?

I am thinking that we might want to raise a NOTICE that the reordering could not happen because the chunk was compressed whenever the index parameter is given.

src/chunk.c
```diff
@@ -2766,18 +2766,29 @@ init_scan_by_compressed_chunk_id(ScanIterator *iterator, int32 compressed_chunk_
 }
 
 bool
-ts_chunk_contains_compressed_data(Chunk *chunk)
+ts_chunk_contains_compressed_data(Chunk *chunk, Chunk **parent)
```
svenklemm, line 2769: that function name is not what the function does, it seems to check if chunk is a compressed chunk and looks up the parent if it is

svenklemm: ts_chunk_is_compressed would make more sense

svenklemm: or ts_chunk_is_compressed_chunk

pmwkaa: Any suggestions on a better name? :)

pmwkaa: Problem is that it is not clear what compressed chunk actually means here. Is it a main chunk or chunk associated with compressed data. Also to note, that this function is used in other places as well

tsl/src/reorder.c
svenklemm: index is required on first call, if we decide to print a message here then it should be DEBUG1 or similar

tsl/src/reorder.c
```diff
@@ -148,12 +150,45 @@ tsl_move_chunk(PG_FUNCTION_ARGS)
...
+		AlterTableCmd cmd = { .type = T_AlterTableCmd,
+							  .subtype = AT_SetTableSpace,
+							  .name = get_tablespace_name(destination_tablespace) };
+
+		AlterTableInternal(chunk_id, list_make1(&cmd), false);
+		AlterTableInternal(compressed_chunk->table_id, list_make1(&cmd), false);
```
gayyappan, line 180: If the chunk data is compressed and a valid index  is passed, I think we should print a NOTICE saying that the index is ignored as the data is compressed.

src/chunk.c
gayyappan, line 2769: Why not  add a new function which returns the parent chunk instead of overloading this one? That would keep the semantics and usage separate.

pmwkaa: I can do it, just not a fan of doing the same scan twice

tsl/src/reorder.c
```diff
@@ -148,12 +149,53 @@ tsl_move_chunk(PG_FUNCTION_ARGS)
...
+
+	/* If chunk is compressed move it by altering tablespace on both chunks */
+	if (OidIsValid(chunk->fd.compressed_chunk_id))
+	{
+		Chunk *compressed_chunk = ts_chunk_get_by_id(chunk->fd.compressed_chunk_id, true);
+		AlterTableCmd cmd = { .type = T_AlterTableCmd,
+							  .subtype = AT_SetTableSpace,
+							  .name = get_tablespace_name(destination_tablespace) };
+
+		ereport(NOTICE,
```
svenklemm, line 182: not a fan of this being NOTICE, we are already way too spammy with NOTICES imo i think this should be DEBUG1

pmwkaa: @erimatnor, @gayyappan are you ok with it?

gayyappan: This should be conditional message.
Maybe reword it slightly as well.

if ( IsValidOid( index_id) ) { 
      elog( NOTICE, "Ignoring index parameter. Chunk will not be reordered as it has compressed data");
}

pmwkaa: @gayyappan updated

PR #2113, opened 2020-07-23
Make scheduler ignore jobs where scheduled = false

svenklemm:
This patch makes the scheduler ignore all jobs where scheduled = false.
The schedule flag can be used to disable jobs without deleting the job
entry.

This patch also changes the job retrieval to use an index scan to
satisfy the assumption of the scheduler to receive a list ordered
by job_id.

### DIFF
--- a/src/bgw/job.c
+++ b/src/bgw/job.c
@@ -272,6 +272,118 @@ ts_bgw_job_get_scheduled(size_t alloc_size, MemoryContext mctx)
 }
 
 static void
+init_scan_by_proc_name(ScanKeyData *scankey, const char *proc_name)
+{
+	ScanKeyInit(scankey,
+				Anum_bgw_job_proc_hypertable_id_idx_proc_name,
+				BTEqualStrategyNumber,
+				F_NAMEEQ,
+				CStringGetDatum(proc_name));
+}
+
+static void
+init_scan_by_proc_schema(ScanKeyData *scankey, const char *proc_schema)
+{
+	ScanKeyInit(scankey,
+				Anum_bgw_job_proc_hypertable_id_idx_proc_schema,
+				BTEqualStrategyNumber,
+				F_NAMEEQ,
+				CStringGetDatum(proc_schema));
+}
+
+static void
+init_scan_by_hypertable_id(ScanKeyData *scankey, int32 hypertable_id)
+{
+	ScanKeyInit(scankey,
+				Anum_bgw_job_proc_hypertable_id_idx_hypertable_id,
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(hypertable_id));
+}
+
+List *
+ts_bgw_job_find_by_proc_and_hypertable_id(const char *proc_name, const char *proc_schema,
+										  int32 hypertable_id)
+{
+	Catalog *catalog = ts_catalog_get();
+	ScanKeyData scankey[3];
+	AccumData list_data = {
+		.list = NIL,
+		.alloc_size = sizeof(BgwJob),
+	};
+	ScannerCtx scanctx = {
+		.table = catalog_get_table_id(catalog, BGW_JOB),
+		.index = catalog_get_index(ts_catalog_get(), BGW_JOB, BGW_JOB_PROC_HYPERTABLE_ID_IDX),
+		.data = &list_data,
+		.scankey = scankey,
+		.nkeys = sizeof(scankey) / sizeof(*scankey),
+		.tuple_found = bgw_job_accum_tuple_found,
+		.lockmode = AccessShareLock,
+		.scandirection = ForwardScanDirection,
+	};
+
+	init_scan_by_proc_name(&scankey[0], proc_name);
+	init_scan_by_proc_schema(&scankey[1], proc_schema);
+	init_scan_by_hypertable_id(&scankey[2], hypertable_id);
+
+	ts_scanner_scan(&scanctx);
+	return list_data.list;
+}
+
+List *
+ts_bgw_job_find_by_proc(const char *proc_name, const char *proc_schema)
+{
+	Catalog *catalog = ts_catalog_get();
+	ScanKeyData scankey[2];
+	AccumData list_data = {
+		.list = NIL,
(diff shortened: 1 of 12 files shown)


test/expected/bgw_db_scheduler.out
```diff
@@ -100,11 +100,45 @@ SELECT * FROM sorted_bgw_log;
...
+ 
+(1 row)
+
+-- empty
+SELECT * FROM _timescaledb_internal.bgw_job_stat;
+ job_id | last_start | last_finish | next_start | last_successful_finish | last_run_success | total_runs | total_duration | total_successes | total_failures | total_crashes | consecutive_failures | consecutive_crashes 
+--------+------------+-------------+------------+------------------------+------------------+------------+----------------+-----------------+----------------+---------------+----------------------+---------------------
+(0 rows)
+
+SELECT delete_job(1000);
```
erimatnor, line 130: Do we have a `run_job(1000)` function to execute unscheduled jobs?

src/bgw/job.c
```diff
@@ -246,9 +251,9 @@ ts_bgw_job_get_all(size_t alloc_size, MemoryContext mctx)
 	};
 	ScannerCtx scanctx = {
 		.table = catalog_get_table_id(catalog, BGW_JOB),
-		.index = InvalidOid,
+		.index = catalog_get_index(ts_catalog_get(), BGW_JOB, BGW_JOB_PKEY_IDX),
```
erimatnor, line 254: Do we want an index on (id,scheduled)?

src/bgw/job.c
```diff
@@ -246,9 +251,9 @@ ts_bgw_job_get_all(size_t alloc_size, MemoryContext mctx)
...
 		.data = &list_data,
-		.tuple_found = bgw_job_accum_tuple_found,
+		.tuple_found = bgw_job_accum_tuple_found_scheduled,
```
erimatnor, line 256: I suggest adding a filter function  that does heap_getattr on scheduled instead of overloading the tuple found func.

src/bgw/job.c
gayyappan, line 254: could you add a comment on the need to use the index : i.e. we are depending on the sorted order.

test/expected/bgw_db_scheduler.out
```diff
@@ -100,11 +100,45 @@ SELECT * FROM sorted_bgw_log;
...
+ delete_job 
+------------
+ 
+(1 row)
+
 --
```
gayyappan, line 136: also need a test that shows if the state is changed from scheduled = false to scheduled = true, the job gets picked up.
And vice-versa. You probably need some additional code for this to work.

svenklemm: this is already handled, the job list is updated on cache invalidations

gayyappan: You need an API in job.c to update the scheduled state . IIRC invalidation callbacks are triggered only in that case.

svenklemm: well i didnt wanna make api changes in this PR, i was gonna add the scheduled flag to alter_job_schedule in a different PR, do you think this should be part of this PR?

src/bgw/job.c
svenklemm: i think we dont need (id,scheduled)

test/expected/bgw_db_scheduler.out
svenklemm: not atm but should probably be part of the other PR that adds the add_job/delete_job API

test/expected/bgw_db_scheduler.out
gayyappan: I think it makes sense to add all the "scheduled" changes as part of this PR.

svenklemm: i disagree i think it makes sense to have all the api changes together especially since there are probably more changes to alter_job_schedule

test/expected/bgw_db_scheduler.out
```diff
@@ -100,11 +104,73 @@ SELECT * FROM sorted_bgw_log;
...
+ 
+(1 row)
+
+-- empty
+SELECT * FROM _timescaledb_internal.bgw_job_stat;
+ job_id | last_start | last_finish | next_start | last_successful_finish | last_run_success | total_runs | total_duration | total_successes | total_failures | total_crashes | consecutive_failures | consecutive_crashes 
+--------+------------+-------------+------------+------------------------+------------------+------------+----------------+-----------------+----------------+---------------+----------------------+---------------------
+(0 rows)
+
+SELECT test_toggle_scheduled(1000);
```
k-rus, line 135: How do you know that `job_id = 1000` for the targeted job?

test/sql/bgw_db_scheduler.sql
```diff
@@ -103,11 +108,30 @@ SELECT * FROM _timescaledb_internal.bgw_job_stat;
...
+--
+
+TRUNCATE bgw_log;
+SELECT ts_bgw_params_reset_time();
+SELECT insert_job('unscheduled', 'bgw_test_job_1', INTERVAL '100ms', INTERVAL '100s', 5, INTERVAL '1s',scheduled:= false);
+SELECT ts_bgw_db_scheduler_test_run_and_wait_for_scheduler_finish(50);
+-- empty
+SELECT * FROM _timescaledb_internal.bgw_job_stat;
+
+SELECT test_toggle_scheduled(1000);
```
k-rus, line 122: It toggles from `false` to `true`, while a test from `true` to `false` seems to be missing.

test/expected/bgw_db_scheduler.out
svenklemm: because thats how the sequence is initialized, the first job will always have id 1000

test/sql/bgw_db_scheduler.sql
svenklemm: the current testing is flawed in that it doesnt actually test that cache invalidations happen but fixing that and adding more tests is outside the scope of this PR

k-rus: Can you link the issue, which will resolve this issue?

PR #2125, opened 2020-07-24
Add refresh function for continuous aggregates

### LINKED ISSUE #2117: Add continuous aggregate refresh function
The new API for continuous aggregates separates policy from core functionality and exposes a new function to refresh the continuous aggregate:

```
SELECT refresh_continuous_aggregate('daily_temp', '2020-01-01', '2020-05-01');
```

The signature should be something like:
```
CREATE OR REPLACE FUNCTION refresh_continuous_aggregate(
    cagg                     REGCLASS,
    window_start             "any",
    window_end               "any"
) RETURNS VOID;
```

We might want to have default parameters for the window, for instance `NULL` to mean "infinite". But that might have to wait until the underlying implementation has the optimizations needed to make refreshing large windows efficient.

Related to https://github.com/timescale/timescaledb/issues/1428

erimatnor:
This change adds a new refresh function called
`refresh_continuous_aggregate` that allows refreshing a continuous
aggregate over a given window of data, called the "refresh window".

This is the first step in a larger overhaul of the continuous
aggregate feature with the goal of cleaning up the API and separating
policy from the core functionality.

Currently, the refresh function does a brute-force refresh of a window
and it bypasses the whole invalidation framework. Future updates
intend to integrate with this framework (with modifications) to
optimize refreshes. An exclusive lock is take on the continuous
aggregate's internal materialized hypertable in order to protect
against concurrent refreshing. However, as this serializes refreshes,
we might want to relax this locking in the future to allow, e.g.,
concurrent refreshes of non-overlapping windows.

The new refresh functionality includes basic tests for bad input and
refreshing across different windows. Unfortunately, a bug in the
optimization code for `time_bucket` causes timestamps to overflow the
allowed MAX time. Therefore, refresh windows that are close to the MAX
allowed size are not yet supported or tested.

Closes #2117

### DIFF
--- a/sql/ddl_api.sql
+++ b/sql/ddl_api.sql
@@ -211,5 +211,12 @@ AS '@MODULE_PATHNAME@', 'ts_distributed_exec' LANGUAGE C VOLATILE;
 CREATE OR REPLACE FUNCTION  set_replication_factor(
     hypertable              REGCLASS,
     replication_factor      INTEGER
-) RETURNS VOID 
+) RETURNS VOID
 AS '@MODULE_PATHNAME@', 'ts_hypertable_distributed_set_replication_factor' LANGUAGE C VOLATILE;
+
+-- Refresh a continuous aggregate across the given window.
+CREATE OR REPLACE FUNCTION refresh_continuous_aggregate(
+    cagg                     REGCLASS,
+    window_start             "any",
+    window_end               "any"
+) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_continuous_agg_refresh' LANGUAGE C VOLATILE;
--- a/tsl/src/continuous_aggs/refresh.c
+++ b/tsl/src/continuous_aggs/refresh.c
@@ -1,0 +1,371 @@
+/*
+ * This file and its contents are licensed under the Timescale License.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-TIMESCALE for a copy of the license.
+ */
+#include <postgres.h>
+#include <utils/lsyscache.h>
+#include <utils/fmgrprotos.h>
+#include <storage/lmgr.h>
+#include <miscadmin.h>
+#include <fmgr.h>
+
+#include <catalog.h>
+#include <continuous_agg.h>
+#include <dimension.h>
+#include <hypertable.h>
+#include <hypertable_cache.h>
+#include <time_bucket.h>
+#include <utils.h>
+
+#include "refresh.h"
+#include "materialize.h"
+
+typedef struct CaggRefreshState
+{
+	ContinuousAgg cagg;
+	Hypertable *cagg_ht;
+	InternalTimeRange refresh_window;
+	SchemaAndName partial_view;
+} CaggRefreshState;
+
+static Hypertable *
+cagg_get_hypertable_or_fail(int32 hypertable_id)
+{
+	Hypertable *ht = ts_hypertable_get_by_id(hypertable_id);
+
+	if (NULL == ht)
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("invalid continuous aggregate state"),
+				 errdetail("A continuous aggregate references a hypertable that does not exist.")));
+
+	return ht;
+}
+
+/*
+ * Get the largest window allowed given the time type of a continuous
+ * aggregate and other restrictions.
+ *
+ * The largest window allowed sets an upper bound on the range a refresh
+ * window can have given a specific time type. Note that this bound might be
+ * lower than supported by the time type itself. This is because we internally
+ * use UNIX epoch time for all non-integer types and must be able to convert
+ * back-and-forth between them.
+ *
+ * Note that each time type has its own range of possible time values. For
+ * instance, Date has a much bigger range than timestamp.
+ *
+ * This is further complicated by the fact that we internally convert
+ * timestamp and date types to UNIX time. The conversion ensures that all
+ * hypertables use the same epoch and ranges internally, although this adds
+ * further restrictions as we must be able to convert back-and-forth between
+ * types and time epochs.
+ */
+static InternalTimeRange
+get_max_window(Oid timetype)
+{
+	InternalTimeRange maxrange = {
+		.type = timetype,
+	};
+
+	/* PG checks for valid timestamps and dates in the range MIN <= time <
+	 * END. So, for those types we subtract 1 from the MAX to get into the
+	 * valid range. The MAX values account for ability to convert to UNIX time
+	 * without overflow. */
+	switch (timetype)
+	{
+		case DATEOID:
+			/* Since our internal conversion turns a date into a timestamp,
+			 * dates are governed by the same limits as timestamps. This is
+			 * probably a limitation we want to do away with, even though it
+			 * has little effect in practice. */
+			maxrange.start = DATETIME_MIN_JULIAN - POSTGRES_EPOCH_JDATE;
+			maxrange.end = TIMESTAMP_END_JULIAN - (2 * POSTGRES_EPOCH_JDATE) + UNIX_EPOCH_JDATE - 1;
+			break;
+		case TIMESTAMPTZOID:
+		case TIMESTAMPOID:
+			maxrange.start = MIN_TIMESTAMP;
+			maxrange.end = (END_TIMESTAMP - TS_EPOCH_DIFF_MICROSECONDS) - 1;
+			break;
+		case INT8OID:
+		case INT4OID:
+		case INT2OID:
+			/* There is no "date"/"time" range for integers so they can take
+			 * any value. */
+			maxrange.start = PG_INT64_MIN;
+			maxrange.end = PG_INT64_MAX;
+			break;
+		default:
+			elog(ERROR, "unrecognized time type %d", timetype);
+			break;
+	}
+
+	maxrange.start = ts_time_value_to_internal(Int64GetDatum(maxrange.start), timetype);
+	maxrange.end = ts_time_value_to_internal(Int64GetDatum(maxrange.end), timetype);
+
+	return maxrange;
+}
+
+/*
+ * Compute the largest possible bucketed window given the time type and
+ * internal restrictions.
+ *
+ * The largest bucketed window is governed by restrictions set by the type and
+ * internal, TimescaleDB-specific legacy details (see get_max_window above for
+ * further explanation).
+ */
+static InternalTimeRange
+get_largest_bucketed_window(Oid timetype, int64 bucket_width)
+{
+	InternalTimeRange maxwindow = get_max_window(timetype);
+	InternalTimeRange maxbuckets;
+
+	/* For the MIN value, the corresponding bucket either falls on the exact
+	 * MIN or it will be below it. Therefore, we add (bucket_width - 1) to
+	 * move to the next bucket to be within the allowed range. */
+	maxwindow.start = maxwindow.start + bucket_width - 1;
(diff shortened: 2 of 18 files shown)


sql/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ set(SOURCE_FILES
   restoring.sql
   timescaledb_fdw.sql
   remote_txn.sql
+  continuous_agg.sql
```
erimatnor, line 54: I guess we could put this in `ddl_api.sql` with all the other functions.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,295 @@
...
+		.end = PG_INT64_MAX,
+	};
+
+	PreventCommandIfReadOnly("refresh_continuous_aggregate()");
+
+	if (!OidIsValid(cagg_relid))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("invalid continuous aggregate")));
+
+	if (PG_ARGISNULL(1) && PG_ARGISNULL(2))
```
WireBaron, line 248:
```suggestion
	if (PG_ARGISNULL(1) || PG_ARGISNULL(2))
```

erimatnor: Should actually remove this altogether.

tsl/test/expected/continuous_aggs_refresh.out
```diff
@@ -0,0 +1,268 @@
...
+SELECT refresh_continuous_aggregate('daily_temp', 'xyz', '2020-05-05');
+ERROR:  invalid input syntax for type timestamp with time zone: "xyz"
+SELECT refresh_continuous_aggregate('daily_temp', '2020-05-03', 'xyz');
+ERROR:  invalid input syntax for type timestamp with time zone: "xyz"
+SELECT refresh_continuous_aggregate('daily_temp', '2020-05-03', '2020-05-01');
+ERROR:  invalid refresh window
+SELECT refresh_continuous_aggregate('daily_temp', '2020-05-03', '2020-05-03');
+ERROR:  invalid refresh window
+-- Bad time input
+SELECT refresh_continuous_aggregate('daily_temp', '2020-05-01'::text, '2020-05-03'::text);
```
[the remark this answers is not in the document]
erimatnor: this is what `get_time_value_from_arg` is for. TEXT is not a valid input type.

I'll expand on a comment to the function.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,291 @@
...
+	 * time without overflow. */
+	switch (timetype)
+	{
+		case DATEOID:
+			/* Since our internal conversion turns a date into a timestamp,
+			 * dates are governed by the same limits as timestamps. This is
+			 * probably a limitation we want to do away with, even though it
+			 * has little effect in practice. */
+			maxrange.start = DATETIME_MIN_JULIAN - POSTGRES_EPOCH_JDATE;
+			maxrange.end = TIMESTAMP_END_JULIAN - (2 * POSTGRES_EPOCH_JDATE) + UNIX_EPOCH_JDATE - 1;
```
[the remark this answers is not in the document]
erimatnor: PostgreSQL defines MIN and END times and verifies that: MIN <= value < END. Thus to be a valid time value, the end must be less than END or there will be an error.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,291 @@
...
+		case TIMESTAMPTZOID:
+		case TIMESTAMPOID:
+			maxrange.start = MIN_TIMESTAMP;
+			maxrange.end = (END_TIMESTAMP - TS_EPOCH_DIFF_MICROSECONDS) - 1;
+			break;
+		case INT8OID:
+		case INT4OID:
+		case INT2OID:
+			maxrange.start = PG_INT64_MIN;
+			maxrange.end = PG_INT64_MAX;
```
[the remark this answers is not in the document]
erimatnor: integer is not a date/timestamp so the entire range is valid. For DATE/TIMESTAMPs the integer range is not the same as the allowed MIN and END time values.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,291 @@
...
+continuous_agg_refresh_execute(CaggRefreshState *refresh)
+{
+	SchemaAndName cagg_hypertable_name = {
+		.schema = &refresh->cagg_ht->fd.schema_name,
+		.name = &refresh->cagg_ht->fd.table_name,
+	};
+	InternalTimeRange new_invalidation_range = {
+		.type = refresh->refresh_window.type,
+		.start = 0,
+		.end = 0,
```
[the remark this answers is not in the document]
erimatnor: The materialization function takes two ranges/windows. We use only one for refresh so this one is zero (not used).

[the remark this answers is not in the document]
k-rus: Can you give better name to the variable or give a comment?

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,291 @@
...
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("invalid continuous aggregate state"),
+				 errdetail("A continuous aggregate references a hypertable that does not exist.")));
+
+	return ht;
+}
+
+static InternalTimeRange
+get_max_window(Oid timetype)
```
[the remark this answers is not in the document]
erimatnor: The ugliness is unfortunately necessary until we can address the time epoch/conversion issues of internal time.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,291 @@
...
+	};
+	InternalTimeRange new_invalidation_range = {
+		.type = refresh->refresh_window.type,
+		.start = 0,
+		.end = 0,
+	};
+	Dimension *time_dim = hyperspace_get_open_dimension(refresh->cagg_ht->space, 0);
+
+	Assert(time_dim != NULL);
+	continuous_agg_update_materialization(refresh->partial_view,
```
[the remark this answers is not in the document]
erimatnor: Added a heavy lock on the materialized hypertable for now.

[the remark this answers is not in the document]
gayyappan: Two concurrent deletes do not necessarily acquire tuple locks in the same order like when we have overlaps: txn A deletes from (10,20) and txn B deletes from (15,30).

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,381 @@
...
+}
+
+/*
+ * Adjust the refresh window to align with buckets in an inclusive manner.
+ *
+ * It is OK to refresh more than the given refresh window, but not less. Since
+ * we can only refresh along bucket boundaries, we need to adjust the refresh
+ * window to be inclusive in both ends to be able to refresh the given
+ * region. For example, if the dotted region below is the original window, the
+ * adjusted refresh window includes all four buckets shown.
```
gayyappan, line 141:
```suggestion
 * adjusted refresh window includes all three buckets shown.
```
typo here

tsl/src/continuous_aggs/refresh.c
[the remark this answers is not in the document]
k-rus: It will be a better solution, so it is good to know that there is a good way forward.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,381 @@
...
+		result.start = ts_time_bucket_by_type(bucket_width, result.start, result.type);
+
+	if (result.end >= largest_bucketed_window.end)
+		result.end = largest_bucketed_window.end;
+	else
+	{
+		/* We get the time value for the start of the bucket, so need to add
+		 * bucket_width to get the end of it */
+		result.end = ts_time_bucket_by_type(bucket_width, result.end, result.type);
+		result.end = result.end + bucket_width;
```
gayyappan, line 164: if the value is at a bucket edge, you are moving up by a bucket here.
say bucket_width = 10, result.end = 20, then final result.end = 30

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,381 @@
...
+	 * concurrent refreshes that don't have overlapping refresh windows.
+	 *
+	 * Concurrent refreshes on the same continuous aggregate could be achieved
+	 * if we protect the aggregate with a UNIQUE constraint on the GROUP BY
+	 * columns. This would allow concurrent refreshes, but overlapping ones
+	 * might fail with, e.g., unique violation errors. Those could be
+	 * captured, however, and ignored when we know it means someone else just
+	 * did the same work.
+	 */
+	LockRelationOid(refresh->cagg_ht->main_table_relid, ExclusiveLock);
```
gayyappan, line 200: This is a heavy lock here. Should be revisited. This would block drop chunks policies from running as well. (Not sure what the rest of the design is going to look like or if thresholds will be updated at this point. The auxiliary tables like the thresholds table are some of the other tables/rows which could be locked).

tsl/src/continuous_aggs/refresh.c
```diff
@@ -0,0 +1,381 @@
...
+}
+
+/*
+ * Finalize a refresh and cleanup state, e.g., release locks.
+ */
+static void
+continuous_agg_refresh_finish(const CaggRefreshState *refresh)
+{
+	UnlockRelationOid(refresh->cagg_ht->main_table_relid, ExclusiveLock);
```
gayyappan, line 209: shouldn't this be released at the end of the transaction? Want to confirm that is the semantics here.

erimatnor: Yes, good catch. Fixed.

tsl/src/continuous_aggs/refresh.c
erimatnor: Sure, we can always revisit later. But I am not sure how you would allow drop_chunks to run simultaneously unless you allow one transaction to potentially fail.

tsl/src/continuous_aggs/refresh.c
erimatnor: The edge case was already checked for in the if clause, so we must be below the edge, which means that the bucket start must be at least one bucket down from the edge and this would just bring it at most up to the edge again, which is OK.

tsl/src/continuous_aggs/refresh.c
erimatnor: The illustration shows four buckets.

sql/ddl_api.sql
```diff
@@ -211,5 +211,12 @@ AS '@MODULE_PATHNAME@', 'ts_distributed_exec' LANGUAGE C VOLATILE;
 CREATE OR REPLACE FUNCTION  set_replication_factor(
     hypertable              REGCLASS,
     replication_factor      INTEGER
-) RETURNS VOID 
+) RETURNS VOID
 AS '@MODULE_PATHNAME@', 'ts_hypertable_distributed_set_replication_factor' LANGUAGE C VOLATILE;
+
+-- Refresh a continuous aggregate across the given window.
+CREATE OR REPLACE FUNCTION refresh_continuous_aggregate(
+    cagg                     REGCLASS,
```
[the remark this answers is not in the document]
k-rus: I agree with @svenklemm that `continuous_aggregate` will be better to use in the public API and not `cagg`, which is not in general known by users. Note that users don't type it usually.

PR #2141, opened 2020-07-27
Refactor reorder policy

svenklemm:
This patch changes the reorder policy to store it's configuration
in the bgw_job table and removes the bgw_policy_reorder table.


tsl/test/expected/tsl_tables.out
```diff
@@ -452,18 +434,12 @@ select count(*) from _timescaledb_config.bgw_policy_drop_chunks where job_id=:jo
      1
 (1 row)
 
-select count(*) from _timescaledb_config.bgw_policy_reorder where job_id=:reorder_job_id;
- count 
--------
-     1
-(1 row)
-
 select * from _timescaledb_config.bgw_job;
```
k-rus, line 437: Can you add `ORDER BY id`?

tsl/test/expected/tsl_tables.out
```diff
@@ -688,10 +658,10 @@ select job_id,chunk_id,num_times_job_run from _timescaledb_internal.bgw_policy_c
 (1 row)
 
 select * from _timescaledb_config.bgw_job where job_type in ('drop_chunks', 'reorder');
```
k-rus, line 660: Can you add `ORDER BY id`?

sql/views.sql
```diff
@@ -78,16 +78,15 @@ CREATE OR REPLACE VIEW timescaledb_information.drop_chunks_policies as
...
 
 CREATE OR REPLACE VIEW timescaledb_information.reorder_policies as
-  SELECT format('%1$I.%2$I', ht.schema_name, ht.table_name)::regclass as hypertable, p.hypertable_index_name, p.job_id, j.schedule_interval,
+  SELECT format('%1$I.%2$I', ht.schema_name, ht.table_name)::regclass as hypertable, jsonb_object_field_text(j.config,'index_name') AS hypertable_index_name, j.id AS job_id, j.schedule_interval,
     j.max_runtime, j.max_retries, j.retry_period
-  FROM _timescaledb_config.bgw_policy_reorder p
-    INNER JOIN _timescaledb_catalog.hypertable ht ON p.hypertable_id = ht.id
-    INNER JOIN _timescaledb_config.bgw_job j ON p.job_id = j.id;
+  FROM _timescaledb_config.bgw_job j
+    INNER JOIN _timescaledb_catalog.hypertable ht ON j.hypertable_id = ht.id;
```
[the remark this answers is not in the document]
k-rus: Can you add a test, which will prevent such miss in future, i.e., a test where `reorder_policies` is queried when `bgw_job` contains not only reorder policies?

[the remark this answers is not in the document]
svenklemm: this view is getting dropped by other planned changes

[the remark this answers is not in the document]
k-rus: Thank you for clarifying. I don't find where it is suggested.

sql/updates/latest-dev.sql
```diff
@@ -128,13 +134,29 @@ ALTER TABLE _timescaledb_config.bgw_job ADD COLUMN config JSONB;
 ALTER TABLE _timescaledb_config.bgw_job DROP CONSTRAINT valid_job_type;
 ALTER TABLE _timescaledb_config.bgw_job ADD CONSTRAINT valid_job_type CHECK (job_type IN ('telemetry_and_version_check_if_enabled', 'reorder', 'drop_chunks', 'continuous_aggregate', 'compress_chunks', 'custom'));
 
+-- migrate reorder jobs
+UPDATE
+  _timescaledb_config.bgw_job job
+SET proc_name = 'policy_reorder',
+  proc_schema = '_timescaledb_internal',
+  config = jsonb_build_object('hypertable_id', reorder.hypertable_id, 'index_name', reorder.hypertable_index_name),
```
erimatnor, line 142: Do we need the `hypertable_id` config if we also set the `hypertable_id` column in the jobs table? It seems to me that only the `index_name` is necessary.

tsl/src/bgw_policy/reorder_api.c
```diff
@@ -42,6 +41,35 @@
 #define DEFAULT_RETRY_PERIOD                                                                       \
 	DatumGetIntervalP(DirectFunctionCall3(interval_in, CStringGetDatum("5 min"), InvalidOid, -1))
 
+#define CONFIG_KEY_HYPERTABLE_ID "hypertable_id"
```
erimatnor, line 44: As pointed out in the update script, I guess this key and value is redundant with the information already in the jobs table column `hypertable_id`. 

If we want to keep hypertable info in the json, then maybe it makes more sense to have the canonical name (schema+name). That is less redundant and provides at least some human-readable info in the json, similar to `index_name`.

tsl/src/bgw_policy/reorder_api.c
```diff
@@ -42,6 +41,35 @@
...
+#define CONFIG_KEY_INDEX_NAME "index_name"
+
+#define POLICY_REORDER_PROC_NAME "policy_reorder"
+
+int32
+policy_reorder_get_hypertable_id(BgwJob *job)
```
erimatnor, line 50: This whole function seems unnecessary if we get the hypertable_id directly from the jobs table.

tsl/src/bgw_policy/reorder_api.c
svenklemm: storing names likes that in json has serious downsides because it would need adjustments on object renames which is circumvented by storing id

sql/updates/latest-dev.sql
svenklemm: the hypertable_id in the jobs table is only used for dependency tracking, it makes sense to have all configuration values in the config object because that is part of the function call

svenklemm: i had it not in the jsonb at first but then decided to add it because it makes the function call more generic

tsl/src/bgw_policy/reorder_api.c
erimatnor: Good point about name. Still not sure about the redundant info.

sql/updates/latest-dev.sql
gayyappan: I have the same comment: why do we need hypertable_id in the json? It is already part of the bgw_job table. I don't get what you mean by  "makes the function call generic " : what function are you referring to here?

svenklemm: the function call signature for scheduled functions is (job_id int32, config jsonb ), having the hypertable_id in the config makes the function implementation independent from job because all required values are in config. So it's a tradeoff between non-redundant internal storage and cleaner implementation of external API.

PR #2148, opened 2020-07-28
Rename add_drop_chunks_policy

### LINKED ISSUE #2119: Rename drop chunks policy function
The function to add a retention policy is called `add_drop_chunks_policy`. This name doesn't convey well what type of policy this is unless one knows what drop_chunks is and how it works, which is an implementation detail. It makes more sense to call this function `add_retention_policy`.

We might also want to reconsider the parameters of the function to reflect the name change. For instance, `older_than` is a direct mapping from `drop_chunks`, but it is a bit confusing since it is an INTERVAL in the policy but, typically, a TIMESTAMPTZ in `drop_chunks`. A better name for this parameter is probably something like `retention_window`, which is common terminology and would also make it similar refresh policies.

WireBaron:
This change replaces the add_drop_chunks_policy function with
add_retention_policy.  This also renames the older_than parameter
of that function as retention_window.

Fixes #2119


sql/bgw_scheduler.sql
```diff
@@ -21,13 +21,13 @@ INSERT INTO _timescaledb_config.bgw_job (id, application_name, job_type, schedul
 (1, 'Telemetry Reporter', 'telemetry_and_version_check_if_enabled', INTERVAL '24h', INTERVAL '100s', -1, INTERVAL '1h')
 ON CONFLICT (id) DO NOTHING;
 
-CREATE OR REPLACE FUNCTION add_drop_chunks_policy(
+CREATE OR REPLACE FUNCTION add_retention_policy(
```
erimatnor, line 24: Maybe we should take the opportunity to also add a description of this function:

```suggestion
-- Add a retention policy to a hypertable or continuous aggregate.
-- The retention_window (typically an INTERVAL) determines the 
-- window beyond which data is dropped at the time
-- of execution of the policy (e.g., '1 week'). Note that the retention 
-- window will always align with chunk boundaries, thus the window 
-- might be larger than the given one, but never smaller. In other
-- words, some data beyond the retention window 
-- might be kept, but data within the window will never be deleted.
CREATE OR REPLACE FUNCTION add_retention_policy(
```


Looking at this signature, it is unclear what the time epoch is from here the retention window starts. I assume it is wall clock (system clock), but I am not sure that always fits the data. We might need to add a separate parameter that allows the user to specify a function that gives the epoch, typically defaulting to `now()`, similar to what I proposed for the `add_refresh_policy` function. This is obviously for another PR, however.

WireBaron: I'm not sure I see the case where one wouldn't assume this was wall clock?

sql/updates/latest-dev.sql
```diff
@@ -1,6 +1,7 @@
 -- Add new function definitions, columns and tables for distributed hypertables
 DROP FUNCTION IF EXISTS create_hypertable(regclass,name,name,integer,name,name,anyelement,boolean,boolean,regproc,boolean,text,regproc,regproc);
 DROP FUNCTION IF EXISTS add_drop_chunks_policy;
+DROP FUNCTION IF EXISTS add_retention_policy;
```
svenklemm, line 4: thats wrong, you are adding this function, but you need to drop remove_drop_chunks_policy

k-rus:
When internal identifies should be renamed to follow the new name? An example: `JOB_TYPE_DROP_CHUNKS`
Will it be a separate issue/PR?

sql/bgw_scheduler.sql
erimatnor: It is probably not the common case, but if the data ingested is not aligned with the wall clock, then the retention policy won't work. Let's say, for instance, you are importing data in `TIMESTAMP WITHOUT TIMEZONE` and the database runs in another timezone than where the data originated.

Or, let's say your data uses some completely different notion of time, e.g., an integer that is an increasing counter, e.g., event number. The question is if the retention internally uses the `integer_now_function` in that case to get the current notion of time when the policy runs?

WireBaron:
> When internal identifies should be renamed to follow the new name? An example: `JOB_TYPE_DROP_CHUNKS`
> Will it be a separate issue/PR?

I'm not sure it makes sense to change all of the internal identifiers.  For example, `JOB_TYPE_DROP_CHUNKS` still probably makes sense for the retention policy.  There are probably a few places where it does make more sense, but this will probably require some careful thought and perhaps discussion, and will be much lower priority than the change of the public facing names.

sql/bgw_scheduler.sql
```diff
@@ -21,13 +21,21 @@ INSERT INTO _timescaledb_config.bgw_job (id, application_name, job_type, schedul
...
 
-CREATE OR REPLACE FUNCTION add_drop_chunks_policy(
+-- Add a retention policy to a hypertable or continuous aggregate.
+-- The retention_window (typically an INTERVAL) determines the 
+-- window beyond which data is dropped at the time
+-- of execution of the policy (e.g., '1 week'). Note that the retention 
+-- window will always align with chunk boundaries, thus the window 
+-- might be larger than the given one, but never smaller. In other
+-- words, some data beyond the retention window 
+-- might be kept, but data within the window will never be deleted.
```
erimatnor, line 31: Nit (on my own suggestion):

```suggestion
-- Add a retention policy to a hypertable or continuous aggregate.
-- The retention_window (typically an INTERVAL) determines the 
-- window (e.g., '1 week') beyond which data is dropped at the time
-- of execution of the policy . Note that the retention 
-- window will always align with chunk boundaries, thus the window 
-- might be larger than the given one, but never smaller. In other
-- words, some data beyond the retention window 
-- might be kept, but data within the window will never be deleted.
```

PR #2152, opened 2020-07-29
Refactor compress chunk policy

### LINKED ISSUE #2149: Rename compress chunks policy
We should rename `add_compress_chunk_policy` to `add_compression_policy`. This would be more consistent with the naming of other policies, giving us:

* `add_retention_policy`
* `add_refresh_policy`
* `add_reorder_policy`
* `add_compression_policy`

At the same time, we should reconsider the names of parameters to match the new terminology and be consistent across all the policy functions. For instance, a number of policies deal with windows/periods, e.g., refresh_window, retention_window, etc. Compression policies also deal with a "window" of data to leave in an uncompressed state. Maybe raw_data_window, uncompressed_window or similar makes sense?

svenklemm:
This patch changes the compression policy to store it's configuration
in the bgw_job table and removes the bgw_policy_compress_chunks table.

Fixes #2149


svenklemm:
There is some intermediate code that i can't fully clean up in this PR because it requires migrating drop_chunks policy.

sql/updates/latest-dev.sql
```diff
@@ -151,12 +151,31 @@ FROM _timescaledb_config.bgw_policy_reorder reorder
...
   AND job.id = reorder.job_id;
 
+-- migrate compression jobs
+UPDATE
+  _timescaledb_config.bgw_job job
+SET proc_name = 'policy_compression',
+  proc_schema = '_timescaledb_internal',
+  config = jsonb_build_object('hypertable_id', c.hypertable_id, 'older_than', CASE WHEN (older_than).is_time_interval THEN (older_than).time_interval::text ELSE (older_than).integer_interval::text END),
+  hypertable_id = c.hypertable_id,
+  OWNER = (
```
erimatnor, line 161: Why is owner in allcaps?

tsl/src/bgw_policy/compression_api.c
```diff
@@ -9,51 +9,108 @@
...
-										  Int32GetDatum(0),                                        \
-										  Float8GetDatum(0)))
+	DatumGetIntervalP(DirectFunctionCall3(interval_in, CStringGetDatum("1 hour"), InvalidOid, -1))
+
+#define POLICY_COMPRESS_PROC_NAME "policy_compression"
+#define CONFIG_KEY_HYPERTABLE_ID "hypertable_id"
+#define CONFIG_KEY_OLDER_THAN "older_than"
+
+int32
+policy_compression_get_hypertable_id(Jsonb *config)
```
erimatnor, line 44: I guess these functions should ideally be:

```suggestion
policy_compression_get_hypertable_id(const Jsonb *config)
```

to make clear that the json isn't modified.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -9,51 +9,108 @@
...
+	if (interval == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("could not find older_than in config for job")));
+
+	return interval;
+}
+
+static bool
+lag_equal(Oid partitioning_type, Jsonb *config, Oid lag_type, Datum lag_datum)
```
erimatnor, line 85: I know what you mean by lag because you mentioned earlier. But it seems way too generic a term to use by itself without context or any comment. Was confused at first what this meant.

I would suggest at least calling it `compression_lag_equal`

tsl/src/bgw_policy/compression_api.c
```diff
@@ -138,10 +192,38 @@ compress_chunks_add_policy(PG_FUNCTION_ARGS)
 	/* insert a new job into jobs table */
 	namestrcpy(&application_name, "Compress Chunks Background Job");
 	namestrcpy(&compress_chunks_name, "compress_chunks");
-	namestrcpy(&proc_name, "");
-	namestrcpy(&proc_schema, "");
+	namestrcpy(&proc_name, POLICY_COMPRESS_PROC_NAME);
+	namestrcpy(&proc_schema, INTERNAL_SCHEMA_NAME);
 	namestrcpy(&owner, GetUserNameFromId(owner_id, false));
 
+	JsonbParseState *parseState = NULL;
```
erimatnor, line 199: According to our code style:

```suggestion
	JsonbParseState *parse_state = NULL;
```

sql/updates/latest-dev.sql
svenklemm: that was pg_formats doing because its SQL keyword

k-rus:
> This patch changes the compression policy to store **it's** configuration
> in the bgw_job table and removes the bgw_policy_compress_chunks table.

@svenklemm typo in the commit message, should be
> its configuration

tsl/src/bgw_policy/job.c
```diff
@@ -89,14 +89,65 @@ get_chunk_id_to_reorder(int32 job_id, Hypertable *ht)
 															 -1);
 }
 
+static int64
+get_compression_window_end_value(Dimension *dim, Jsonb *config)
```
k-rus, line 93: If I noticed correctly, there are three different names are used for identifiers in the code for the same value: `older_than`, `lag`, `window_end`. If it is so, it is confusing and would be nice to use only one name.

tsl/src/bgw_policy/job.c
```diff
@@ -89,14 +89,65 @@ get_chunk_id_to_reorder(int32 job_id, Hypertable *ht)
...
+	if (IS_INTEGER_TYPE(partitioning_type))
+	{
+		int64 lag = policy_compression_get_older_than_int(config);
+		Oid now_func = ts_get_integer_now_func(dim);
+		ts_interval_now_func_validate(now_func, partitioning_type);
+
+		if (InvalidOid == now_func)
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("integer_now function must be set")));
```
k-rus, line 106: It's already validated in `ts_interval_now_func_validate`.

tsl/src/bgw_policy/job.c
```diff
@@ -89,14 +89,65 @@ get_chunk_id_to_reorder(int32 job_id, Hypertable *ht)
...
+static int64
+get_compression_window_end_value(Dimension *dim, Jsonb *config)
+{
+	Oid partitioning_type = ts_dimension_get_partition_type(dim);
+
+	if (IS_INTEGER_TYPE(partitioning_type))
+	{
+		int64 lag = policy_compression_get_older_than_int(config);
+		Oid now_func = ts_get_integer_now_func(dim);
+		ts_interval_now_func_validate(now_func, partitioning_type);
```
k-rus, line 101: To distinguish from variable declaration block:
```suggestion

		ts_interval_now_func_validate(now_func, partitioning_type);
```

tsl/src/bgw_policy/job.c
```diff
@@ -89,14 +89,65 @@ get_chunk_id_to_reorder(int32 job_id, Hypertable *ht)
...
-get_chunk_to_compress(Hypertable *ht, FormData_ts_interval *older_than)
+get_chunk_to_compress(Hypertable *ht, Jsonb *config)
 {
 	Dimension *open_dim = hyperspace_get_open_dimension(ht->space, 0);
 	StrategyNumber end_strategy = BTLessStrategyNumber;
-	Oid partitioning_type = ts_dimension_get_partition_type(open_dim);
-	int64 end_value = ts_time_value_to_internal(ts_interval_subtract_from_now(older_than, open_dim),
-												partitioning_type);
+
+	int64 end_value = get_compression_window_end_value(open_dim, config);
```
k-rus, line 150: I think new line should distinguish declaration block from return call, not between declarations (to improve readability):
```suggestion
	int64 end_value = get_compression_window_end_value(open_dim, config);
	
```

tsl/src/bgw_policy/job.c
svenklemm: older_than and lag are used interchangeably, i started using lag because i prefer it over older_than but what we name this should by in sync with external api where we havent decided on the new name yet.
window_end is not equal to those `window_end = now() - lag`

tsl/src/bgw_policy/compression_api.c
```diff
@@ -9,51 +9,108 @@
...
+			default:
+				return false;
+		}
+	}
+	else
+	{
+		if (lag_type != INTERVALOID)
+			return false;
+
+		Interval *config_value = policy_compression_get_older_than_interval(config);
```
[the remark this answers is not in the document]
svenklemm: we decided this is valid according to our code style

tsl/src/bgw_policy/job.c
k-rus: Thank you for clarifying about `window_end`. I was confused, since arguments to `ts_interval_from_now_func_get_datum` didn't fit the window during review on GH website.

tsl/src/bgw_policy/job.c
svenklemm: there is some intermediate code here that i want to fully clean up after migrating drop_chunk because that allows me to remove ts_interval and have some shared functions between drop chunk and compression policy

k-rus: Can you move this check before calling `ts_interval_now_func_validate`? Then it will be reasonable, since it produces new error message.

k-rus:
@svenklemm Updating compression setting with existing compression policy is not tested. I suggest either to add test, or enable it in another PR.

k-rus:
> @svenklemm Updating compression setting with existing compression policy is not tested. I suggest either to add test, or enable it in another PR.

The issue is https://github.com/timescale/timescaledb/issues/2158 I put it into 2.0.0 milestone.

svenklemm:
The code needs to handle those cases because it would be similar to direct compress_chunk calls, so the block makes no sense.

tsl/src/bgw_policy/job.c
svenklemm: removed the call, since it is indeed redundant as ts_get_integer_now_func already does the check

tsl/src/bgw_policy/compression_api.c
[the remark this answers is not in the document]
erimatnor: @svenklemm To be fair, the change in code style was a unilateral decision by a subset of people, and in the current team configuration there's actually a majority that wasn't part of the decision. AFAIK, there has been no update to the [Style Guide ](https://github.com/timescale/timescaledb/blob/master/docs/StyleGuide.md) to make this change clear either.

I've let this pass, given the push-back and controversy.

PR #2153, opened 2020-07-29
Add support for calling custom functions to bgw scheduler

svenklemm:
This patch adds internal support for calling user-defined
functions and procedures to the background worker scheduler.


tsl/src/bgw_policy/job.c
```diff
@@ -379,6 +381,92 @@ bgw_policy_job_check_enterprise_license(BgwJob *job)
...
+	name = list_make2(makeString(NameStr(job->fd.proc_schema)),
+					  makeString(NameStr(job->fd.proc_name)));
+	proc = LookupFuncName(name, 2, proc_args, false);
+
+	prokind = get_func_prokind(proc);
+
+	arg1 = makeConst(INT4OID, -1, InvalidOid, 4, Int32GetDatum(job->fd.id), false, true);
+	arg2 = makeConst(JSONBOID, -1, InvalidOid, -1, JsonbPGetDatum(job->fd.config), false, false);
+
+	FuncExpr *funcexpr = makeFuncExpr(proc,
```
erimatnor, line 440: I suggest moving declaration up to other declarations since most of them are already in the beginning of the function.

tsl/src/bgw_policy/job.c
```diff
@@ -379,6 +381,92 @@ bgw_policy_job_check_enterprise_license(BgwJob *job)
...
+	/* we don't need to create proper param list cause we pass in all arguments as Const */
+#ifdef PG11
+	ParamListInfo params = palloc0(offsetof(ParamListInfoData, params));
+#else
+	ParamListInfo params = makeParamList(0);
+#endif
+	ExecuteCallStmt(call, params, false, dest);
+}
+
+static bool
```
erimatnor, line 414: What's the reason to return `bool` here and what does it mean? It seems the result is always `true`.

I guess some kind of exit code from the user function would make sense, but that's not the case here.

tsl/src/bgw_policy/job.c
```diff
@@ -379,6 +381,92 @@ bgw_policy_job_check_enterprise_license(BgwJob *job)
...
+
+	FreeExprContext(econtext, true);
+	FreeExecutorState(estate);
+}
+
+static void
+job_execute_procedure(FuncExpr *funcexpr)
+{
+	CallStmt *call = makeNode(CallStmt);
+	call->funcexpr = funcexpr;
```
erimatnor, line 403: I suggest moving this down to before `ExecuteCallStmt`

tsl/src/bgw_policy/job.c
```diff
@@ -379,6 +381,92 @@ bgw_policy_job_check_enterprise_license(BgwJob *job)
...
+			break;
+	}
+
+	if (started)
+	{
+		PopActiveSnapshot();
+		CommitTransactionCommand();
+	}
+
+	return true;
```
erimatnor, line 467: Should it return `started`?

svenklemm: i was just adhering to the other execute_functions, it doesnt make much sense and should be removed once all the policies have been migrated, all the errors are handled by exceptions. It means the job run successfully.

tsl/src/bgw_policy/job.c
svenklemm: i was just adhering to the other execute_functions, it doesnt make much sense and should be removed once all the policies have been migrated, all the errors are handled by exceptions. It means the job run successfully.

This is all wrapped in try/catch in the scheduler so we dont need the return value

svenklemm: I guess the reason for the bool is if there is any error condition we detect without throwing error we could signal it that way

svenklemm:
This is taken from https://github.com/timescale/timescaledb/pull/2006 which has tests and adds the public API for the functionality

PR #2166, opened 2020-07-30
Add functions for compression stats

gayyappan:
Add chunk_compression_stats and hypertable_compression_stats
functions to get before/after compression sizes


tsl/src/dist_util.c
```diff
@@ -234,6 +234,23 @@ dist_util_remote_chunk_info(PG_FUNCTION_ARGS)
...
 
+Datum
+dist_util_remote_compressed_chunk_info(PG_FUNCTION_ARGS)
+{
+	char *node_name;
+	StringInfo query_str = makeStringInfo();
+	/* Strict function */
+	Name schema_name = PG_GETARG_NAME(1);
+	Name table_name = PG_GETARG_NAME(2);
+	Assert(!PG_ARGISNULL(0) && !PG_ARGISNULL(1) && !PG_ARGISNULL(2));
```
svenklemm, line 245: can you make this an actual check strict checks only apply when calling functions from sql so its better to have those checks in place and also PG_NARGS

gayyappan: Can you elaborate what you are asking for?

svenklemm: change the assert into if and check PG_NARGS == 3

gayyappan: I would like to understand the issue better. This is called via SQL/Pgsql and declared as strict. Are you saying that we are missing the strictness check via PG here? Isn't that the case only if we call via C?

erimatnor: I think what @svenklemm is saying is that if the function is marked `STRICT` it won't be called if any of the arguments are NULL, so `PG_NARGS` must be equal to 3.

tsl/src/dist_util.c
```diff
@@ -234,6 +234,23 @@ dist_util_remote_chunk_info(PG_FUNCTION_ARGS)
 	return dist_util_remote_srf_query(fcinfo, node_name, query_str->data);
 }
 
+Datum
+dist_util_remote_compressed_chunk_info(PG_FUNCTION_ARGS)
+{
+	char *node_name;
+	StringInfo query_str = makeStringInfo();
```
svenklemm, line 241:
```suggestion
    if (PG_NARGS() != 3 || PG_ARGISNULL(0) || !PG_ARGISNULL(1) || !PG_ARGISNULL(2))
       PG_RETURN_NULL();
	char *node_name;
	StringInfo query_str = makeStringInfo();
```

PR #2172, opened 2020-07-31
Test both debug and release on MacOS in CI

k-rus:
Switches MacOS test in PR to run release, so the build time is
reduced. Adds the test in Debug to scheduled job.


svenklemm:
can you push to prerelease_test to see if mac config works

k-rus:
> can you push to prerelease_test to see if mac config works

@svenklemm Pushed, and works correctly.

PR #2184, opened 2020-08-02
Fix background worker scheduler memory consumption

### LINKED ISSUE #2115: Huge memory usage from Background Worker Scheduler
 - OS: Ubuntu 20.04
 - TimescaleDB version (output of `\dx` in `psql`): 1.7.2

**Describe the bug**
I have a database where I do a lot of selects from my application that read a big number of rows (10k+) from different tables.

I noticed that a Background Worker Scheduler will consume a huge amount of memory and never release it, even if I close my application, as shown in this image:

image

My postgresql.conf is the default from the docker, but here are some of the possible relevant values:
```
shared_buffers = 2GB
work_mem = 1747kB
maintenance_work_mem = 1GB
max_wal_size = 1GB
effective_cache_size = 6GB
timescaledb.max_background_workers = 8
```

If I try to limit the memory available to the Docker container, The worker will be killed after some time probably due to OOM and the database will go to recovery mode:
```
my_postgres | 2020-07-24 00:10:42.114 UTC [1] LOG:  background worker "TimescaleDB Background Worker Scheduler" (PID 29) was terminated by signal 9: Killed
my_postgres | 2020-07-24 00:10:42.114 UTC [1] LOG:  terminating any other active server processes
my_postgres | 2020-07-24 00:10:42.114 UTC [5209] WARNING:  terminating connection because of crash of another server process
my_postgres | 2020-07-24 00:10:42.114 UTC [5209] DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
my_postgres | 2020-07-24 00:10:42.114 UTC [5209] HINT:  In a moment you should be able to reconnect to the database and repeat your command.
```

**To Reproduce**
I'm not so sure how to explain how to reproduce since this would need basically my application running, but I will try to pinpoint some important things.

My system needs to read a lot of rows (more than 10k+ lines normally) from each table from my database during a load procedure. This will be done in parallel 10 at a time.

So basically the system chooses 10 tables, read the rows, do some computation with it then choose another 10 tables and do the same until it read all the tables in the database.

This database has 12945 tables, maybe this has some influence on the worker memory usage?

**Expected behavior**
Background Worker Scheduler would use way less ram following the config parameters and then releasing it when not needed anymore.

**Actual behavior**
Background Worker Scheduler increases ram consumption infinitely until OOM, never releasing it.

svenklemm:
This patch changes the scheduler main loop to use a scratch
context during operation which gets reset after every scheduling
loop iteration.

Fixes #2115

### DIFF
--- a/src/bgw/scheduler.c
+++ b/src/bgw/scheduler.c
@@ -60,6 +60,9 @@ static bool jobs_list_needs_update;
 /* has to be global to shutdown jobs on exit */
 static List *scheduled_jobs = NIL;
 
+static MemoryContext scheduler_mctx;
+static MemoryContext scratch_mctx;
+
 /* See the README for a state transition diagram */
 typedef enum JobState
 {
@@ -107,6 +110,34 @@ typedef struct ScheduledBgwJob
 static void on_failure_to_start_job(ScheduledBgwJob *sjob);
 
 static volatile sig_atomic_t got_SIGHUP = false;
+
+BackgroundWorkerHandle *
+ts_bgw_start_worker(const char *function, const char *name, const char *extra)
+{
+	BackgroundWorker worker = {
+		.bgw_flags = BGWORKER_SHMEM_ACCESS | BGWORKER_BACKEND_DATABASE_CONNECTION,
+		.bgw_start_time = BgWorkerStart_RecoveryFinished,
+		.bgw_restart_time = BGW_NEVER_RESTART,
+		.bgw_notify_pid = MyProcPid,
+		.bgw_main_arg = ObjectIdGetDatum(MyDatabaseId),
+	};
+	BackgroundWorkerHandle *handle = NULL;
+
+	StrNCpy(worker.bgw_name, name, BGW_MAXLEN);
+	StrNCpy(worker.bgw_library_name, ts_extension_get_so_name(), BGW_MAXLEN);
+	StrNCpy(worker.bgw_function_name, function, BGW_MAXLEN);
+
+	Assert(strlen(extra) < BGW_EXTRALEN);
+	StrNCpy(worker.bgw_extra, extra, BGW_EXTRALEN);
+
+	/* handle needs to be allocated in long-lived memory context */
+	MemoryContextSwitchTo(scheduler_mctx);
+	if (!RegisterDynamicBackgroundWorker(&worker, &handle))
+		handle = NULL;
+	MemoryContextSwitchTo(scratch_mctx);
+
+	return handle;
+}
 
 #if USE_ASSERT_CHECKING
 static void
@@ -245,6 +276,7 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
 					 sjob->job.fd.id);
 				ts_bgw_job_cache_invalidate_callback();
 				CommitTransactionCommand();
+				MemoryContextSwitchTo(scratch_mctx);
 				return;
 			}
 
@@ -258,6 +290,7 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
 					 NameStr(sjob->job.fd.application_name));
 				scheduled_bgw_job_transition_state_to(sjob, JOB_STATE_SCHEDULED);
 				CommitTransactionCommand();
+				MemoryContextSwitchTo(scratch_mctx);
 				return;
 			}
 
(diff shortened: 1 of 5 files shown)


src/bgw/scheduler.c
```diff
@@ -60,6 +60,16 @@ static bool jobs_list_needs_update;
 /* has to be global to shutdown jobs on exit */
 static List *scheduled_jobs = NIL;
 
+static MemoryContext scratch_mctx;
+
+/*
+ * macro to make sure we stay in scratch memory context because CommitTransactionCommand will
+ * reset memory context to TopMemoryContext
+ */
+#define CommitTransactionCommand()                                                                 \
```
pmwkaa, line 69: I strongly suggest to make the name more explicit about what it does, without overriding the original command or not use this macro at all

src/bgw/scheduler.c
```diff
@@ -671,11 +682,12 @@ ts_bgw_scheduler_process(int32 run_for_interval_ms,
 {
 	TimestampTz start = ts_timer_get_current_timestamp();
 	TimestampTz quit_time = DT_NOEND;
-	MemoryContext scheduler_mctx = CurrentMemoryContext;
+	scratch_mctx = AllocSetContextCreate(CurrentMemoryContext, "Scratch", ALLOCSET_DEFAULT_SIZES);
```
pmwkaa, line 685: It might be unclear what scratch actually means here. If I am not mistaken, this names can be seen outside when debugging memory contexts. Maybe add a scheduler prefix, etc?

src/bgw/scheduler.c
```diff
@@ -720,12 +733,13 @@ ts_bgw_scheduler_process(int32 run_for_interval_ms,
 		if (jobs_list_needs_update)
 		{
 			StartTransactionCommand();
-			scheduled_jobs = ts_update_scheduled_jobs_list(scheduled_jobs, scheduler_mctx);
+			scheduled_jobs = ts_update_scheduled_jobs_list(scheduled_jobs, TopMemoryContext);
```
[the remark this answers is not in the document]
erimatnor: Btw, this main loop is really weird. Why don't we read the scheduled jobs list at the start of the loop, then run jobs, then sleep? The ordering now is really strange. This seems like a more reasonable loop:

```
while (quit_time > ts_timer_get_current_timestamp() && !ProcDiePending && !ts_shutdown_bgw)
	{
		TimestampTz next_wakeup = quit_time;
/*
		 * Process any cache invalidation message that indicates we need to
		 * update the jobs list
		 */
		AcceptInvalidationMessages();

		/* txn to read the list of jobs from the DB */
		if (jobs_list_needs_update)
		{
			StartTransactionCommand();
			scheduled_jobs = ts_update_scheduled_jobs_list(scheduled_jobs, scheduler_mctx);
			CommitTransactionCommand();
			jobs_list_needs_update = false;
		}

		check_for_stopped_and_timed_out_jobs();
		/* start jobs, and then check when to next wake up */
		start_scheduled_jobs(bgw_register);

		next_wakeup = least_timestamp(next_wakeup, earliest_wakeup_to_start_next_job());
		next_wakeup = least_timestamp(next_wakeup, earliest_job_timeout());

		ts_timer_wait(next_wakeup);

		CHECK_FOR_INTERRUPTS();

		if (got_SIGHUP)
		{
			got_SIGHUP = false;
			ProcessConfigFile(PGC_SIGHUP);
		}
	}
```

Then we can also skip the initial transaction outside the loop that reads the scheduled jobs list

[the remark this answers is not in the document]
svenklemm: I think we only care about stuff not getting allocated in TopMemoryContext uncontrolled so everything wrapped in Start/Commit will be allocated in CurTransactionContext which gets deleted by CommitTransactionCommand

src/bgw/scheduler.c
```diff
@@ -694,6 +706,7 @@ ts_bgw_scheduler_process(int32 run_for_interval_ms,
 	while (quit_time > ts_timer_get_current_timestamp() && !ProcDiePending && !ts_shutdown_bgw)
```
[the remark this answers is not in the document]
WireBaron: This does seem a little backwards to me as well.  I'd have expected you to switch to the necessary context here rather than asserting that the previous loop left you in a particular state.

src/bgw/scheduler.c
```diff
@@ -245,6 +276,7 @@ scheduled_bgw_job_transition_state_to(ScheduledBgwJob *sjob, JobState new_state)
 					 sjob->job.fd.id);
 				ts_bgw_job_cache_invalidate_callback();
 				CommitTransactionCommand();
+				MemoryContextSwitchTo(scratch_mctx);
```
erimatnor, line 279: Would be nice to have Assertions for the memory contexts we're on after `StartTransactionCommand` and `CommitTransactionCommand` just to make that clear.

svenklemm: i've added it in one place in the main loop

PR #2189, opened 2020-08-04
Add hypertable_index_size function

gayyappan:


test/sql/views.sql
```diff
@@ -53,6 +54,7 @@ SELECT * FROM timescaledb_information.hypertables WHERE owner = 'super_user' ORD
 ---Add integer table --
 CREATE TABLE test_table_int(time bigint, junk int);
 SELECT create_hypertable('test_table_int', 'time', chunk_time_interval => 10);
+SELECT attach_tablespace('tablespace1', 'test_table_int');
 CREATE OR REPLACE function table_int_now() returns BIGINT LANGUAGE SQL IMMUTABLE as  'SELECT 1::BIGINT';
```
mkindahl, line 58: These changes does not seem relevant for the change you're doing. Might be better to take them as a separate PR.

test/sql/size_utils.sql
```diff
@@ -5,7 +5,13 @@
 \ir include/insert_two_partitions.sql
 
 SELECT * FROM hypertable_detailed_size('"public"."two_Partitions"');
-SELECT * FROM indexes_relation_size('"public"."two_Partitions"');
+SELECT * FROM hypertable_index_relation_size('"public"."two_Partitions_device_id_timeCustom_idx"');
```
mkindahl, line 8: Nit: You don't need the quotes.
```suggestion
SELECT * FROM hypertable_index_relation_size('public.two_Partitions_device_id_timeCustom_idx');
```

test/sql/size_utils.sql
```diff
@@ -56,4 +62,4 @@ SELECT * FROM hypertable_approximate_row_count(NULL);
 
 SELECT * FROM chunks_detailed_size(NULL);
 SELECT * FROM hypertable_detailed_size(NULL);
-SELECT * FROM indexes_relation_size(NULL);
+SELECT * FROM hypertable_index_relation_size(NULL);
```
mkindahl, line 65: I don't see any negative tests, e.g., what happens if you pass in a relation name or an Oid that is incorrect?

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+<<main>>
+DECLARE
+        ht_index_name       NAME;
+        ht_schema_name      NAME;
+        ht_name      NAME;
+        is_distributed   BOOL;
+        ht_id INTEGER;
+BEGIN
+
+   --select c.relname, cind.indexrelid, cind.indrelid, cl.relname, nsp.nspname
```
mkindahl, line 737:
```suggestion
```

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+            ht.schema_name = schema_name_in
+            AND ht.table_name = table_name_in
+            AND s.hypertable_id = ht.id
+         ) AS srv
+    LEFT OUTER JOIN LATERAL _timescaledb_internal.data_node_index_size(
+    CASE WHEN srv.node_up THEN
+        srv.node_name
+    ELSE
+        NULL
+    END, index_name_in) entry ON TRUE
```
mkindahl, line 707: Wouldn't a sub-select be better and easier to follow?

sql/size_utils.sql
k-rus, line 737: Should it be removed?

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+   WHERE c.oid = cind.indexrelid and cind.indrelid = cl.oid
+         and cl.relnamespace = nsp.oid and c.oid = index_name;
+        
+   SELECT replication_factor > 0
+   INTO STRICT is_distributed
+   FROM _timescaledb_catalog.hypertable ht
+   WHERE ht.schema_name = ht_schema_name and ht.table_name = ht_name;
+
+   CASE WHEN is_distributed THEN
+         --RETURN QUERY SELECT NULL, NULL ;
```
k-rus, line 750: Should it be removed?

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+         and cl.relnamespace = nsp.oid and c.oid = index_name;
+        
+   SELECT replication_factor > 0
+   INTO STRICT is_distributed
+   FROM _timescaledb_catalog.hypertable ht
+   WHERE ht.schema_name = ht_schema_name and ht.table_name = ht_name;
+
+   CASE WHEN is_distributed THEN
+         --RETURN QUERY SELECT NULL, NULL ; 
+         RETURN QUERY SELECT *
```
k-rus, line 751: It will be more clear:
```suggestion
         RETURN QUERY SELECT ch.total_bytes, ch.node_name
```

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+   SELECT c.relname, cl.relname, nsp.nspname       
+   INTO STRICT ht_index_name, ht_name, ht_schema_name  
+   FROM pg_class c, pg_index cind, pg_class cl, pg_namespace nsp
+   WHERE c.oid = cind.indexrelid and cind.indrelid = cl.oid
+         and cl.relnamespace = nsp.oid and c.oid = index_name;
+        
+   SELECT replication_factor > 0
+   INTO STRICT is_distributed
+   FROM _timescaledb_catalog.hypertable ht
+   WHERE ht.schema_name = ht_schema_name and ht.table_name = ht_name;
```
k-rus, line 747:
```suggestion
   WHERE ht.schema_name = ht_schema_name AND ht.table_name = ht_name;
```

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+CREATE OR REPLACE FUNCTION  hypertable_index_relation_size(
+    index_name              REGCLASS
+)
+RETURNS TABLE (
+               total_bytes BIGINT,
+               node_name   NAME)
+LANGUAGE PLPGSQL STABLE STRICT
+AS
+$BODY$
+<<main>>
```
k-rus, line 728: What does this mean?

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+        is_distributed   BOOL;
+        ht_id INTEGER;
+BEGIN
+
+   --select c.relname, cind.indexrelid, cind.indrelid, cl.relname, nsp.nspname       
+   SELECT c.relname, cl.relname, nsp.nspname       
+   INTO STRICT ht_index_name, ht_name, ht_schema_name  
+   FROM pg_class c, pg_index cind, pg_class cl, pg_namespace nsp
+   WHERE c.oid = cind.indexrelid and cind.indrelid = cl.oid
+         and cl.relnamespace = nsp.oid and c.oid = index_name;
```
k-rus, line 742:
```suggestion
   WHERE c.oid = cind.indexrelid AND cind.indrelid = cl.oid
         AND cl.relnamespace = nsp.oid AND c.oid = index_name;
```

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+            _timescaledb_internal.ping_data_node (s.node_name) AS node_up
+        FROM
+            _timescaledb_catalog.hypertable AS ht,
+            _timescaledb_catalog.hypertable_data_node AS s
+        WHERE
+            ht.schema_name = schema_name_in
+            AND ht.table_name = table_name_in
+            AND s.hypertable_id = ht.id
+         ) AS srv
+    LEFT OUTER JOIN LATERAL _timescaledb_internal.data_node_index_size(
```
k-rus, line 702: Is there any sense to have outer joint, since it is not meaningful to ask for size on data node with NULL name? Can it be just?:
 ```suggestion
    JOIN LATERAL _timescaledb_internal.data_node_index_size(
```

File: test/expected/relocate_extension.out [hunk 12]
```diff
@@ -164,12 +164,17 @@ SELECT * FROM "testSchema0".hypertable_detailed_size('test_ts');
 (1 row)
 
 -- testing hypertable_detailed_size END
-SELECT * FROM "testSchema0".indexes_relation_size('test_ts') ORDER BY index_name;
-           index_name           | total_bytes 
---------------------------------+-------------
- public.test_ts_device_time_idx |       32768
- public.test_ts_time_idx        |       32768
-(2 rows)
+SELECT * FROM "testSchema0".hypertable_index_relation_size('test_ts_time_idx');
```
k-rus, line 167: IMHO it is going to be confusing for users on a single node, e.g., in Timescale Cloud, where multimode is not an option.

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
 END;
 $BODY$;
 
+-------------Get index size for hypertables -------
+-- hypertable_index_name      - index on hyper table
+CREATE OR REPLACE FUNCTION _timescaledb_internal.indexes_local_size(
+    index_name_in              NAME
+)
+RETURNS TABLE ( hypertable_id INTEGER,
```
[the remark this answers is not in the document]
k-rus: @mkindahl This function is internal, not for public use.

test/sql/views.sql
gayyappan: It is a separate commit.

sql/size_utils.sql
```diff
@@ -682,3 +635,126 @@ BEGIN
...
+$BODY$;
+
+-- Get sizes of indexes on a hypertable
+--
+-- index_name           - index on hyper table
+--
+-- Returns:
+-- total_bytes          - size of index on disk
+
+CREATE OR REPLACE FUNCTION  hypertable_index_relation_size(
```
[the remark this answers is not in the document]
erimatnor: I also thought we agreed to remove "relation" from the function names, unless the function is truly agnostic of the type of relation.

tsl/test/sql/dist_views.sql
```diff
@@ -8,17 +8,17 @@
 \c :TEST_DBNAME :ROLE_CLUSTER_SUPERUSER
 
 SET client_min_messages TO ERROR;
-DROP DATABASE IF EXISTS data_node_1;
-DROP DATABASE IF EXISTS data_node_2;
-DROP DATABASE IF EXISTS data_node_3;
```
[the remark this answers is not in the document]
gayyappan: The test is now moved to a different location in CMakeLists.txt. There are issues with running tests in parallel if multiple tests use the same data node name. The renaming prevents unexpected  CI failures.

sql/size_utils.sql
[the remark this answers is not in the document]
svenklemm: name is definitely not the right datatype cause it has 64 char limit but since both table name and schema can be 64 chars its too short, regclass would be more appropriate or text if we dont want oid but i would prefer regclass here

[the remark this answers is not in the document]
svenklemm: seems to be missing schema, i think regclass would be more appropriate here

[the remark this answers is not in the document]
gayyappan: added schema name. Needs to be used across data nodes, so cannot use regclass here.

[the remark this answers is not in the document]
mkindahl: Not sure I follow: why would regclass not work? The parameter to `drop_chunks` is regclass and it works.

sql/size_utils.sql
```diff
@@ -682,3 +635,125 @@ BEGIN
...
+    index_name_in              NAME
+)
+RETURNS TABLE ( hypertable_id INTEGER,
+                total_bytes BIGINT ) 
+               LANGUAGE PLPGSQL STABLE STRICT
+               AS
+$BODY$
+BEGIN
+        RETURN QUERY
+        SELECT ci.hypertable_id, sum(pg_relation_size(c.oid))::bigint
```
mkindahl, line 654: Since this is just a select, does it have to be `plpgsql`?

tsl/src/dist_util.c
```diff
@@ -254,6 +254,26 @@ dist_util_remote_compressed_chunk_info(PG_FUNCTION_ARGS)
...
 
+Datum
+dist_util_remote_hypertable_index_info(PG_FUNCTION_ARGS)
+{
+	char *node_name;
+	StringInfo query_str;
+	Name schema_name, index_name;
+	/* Strict function */
+	if (PG_NARGS() != 3 || PG_ARGISNULL(0) || PG_ARGISNULL(1) || PG_ARGISNULL(2))
+		PG_RETURN_NULL();
```
mkindahl, line 265: Nit: CodeCov complains about this. Not sure if it is really necessary to test, but if you want to silence CodeCov you can probably create a separate test where you define the function as non-strict and call it so that this condition is triggered.

tsl/test/sql/dist_views.sql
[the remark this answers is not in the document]
mkindahl: Yeah, I've stumbled over that as well, but an easy way around it is to make the test a solo-test. IMHO, we need to have something more advanced in place to allow execution of tests with data nodes to run in parallel.

test/sql/views.sql
gayyappan: removed this.

test/expected/views.out
```diff
@@ -32,6 +32,7 @@ SELECT * FROM timescaledb_information.hypertables ORDER BY table_schema, table_n
 (2 rows)
 
 \c :TEST_DBNAME :ROLE_SUPERUSER
+CREATE TABLESPACE tablespace1 OWNER :ROLE_DEFAULT_PERM_USER LOCATION :TEST_TABLESPACE1_PATH;
```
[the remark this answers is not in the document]
gayyappan: removed this from the current test. will be another PR.

sql/size_utils.sql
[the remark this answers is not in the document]
gayyappan: we can move to regclass when the catalogs use regclass. The regclass here needs to be translated into schema+name to effectively query our catalogs. We do not want to do the mapping twice: once in the external function and then again in the internal function.

test/sql/size_utils.sql
gayyappan: test aded

sql/size_utils.sql
gayyappan: does not make a difference anymore since the function signature changed and we are aggregating the result.

sql/size_utils.sql
gayyappan: No, it doesn't.  But don't see any special benefit from using plain SQL either.

sql/size_utils.sql
```diff
@@ -682,3 +635,125 @@ BEGIN
...
+-- index_name      - index on hyper table
+---note that the query matches against the hypertable's schema name as
+-- the input is on the hypertable index nd not the chunk index.
+CREATE OR REPLACE FUNCTION _timescaledb_internal.indexes_local_size(
+    schema_name_in             NAME,
+    index_name_in              NAME
+)
+RETURNS TABLE ( hypertable_id INTEGER,
+                total_bytes BIGINT ) 
+               LANGUAGE PLPGSQL STABLE STRICT
```
svenklemm, line 649: i think this should be volatile instead of strict

gayyappan: Could you elaborate why? this function does not mutate anything.

svenklemm: because the output is not constant for the same input within the same session

tsl/src/dist_util.c
gayyappan: Yes, this is become a bit of an annoyance : I can remove this check. Since this is a strict function , it will not get triggered if called via SQL. Or be defensive and add this code and then see a bunch of code coverage errors. I don't want to add addl. unnecessary tests here. It feels like spinning wheels just because the tool points out what could be potential issues.

sql/size_utils.sql
gayyappan: it only needs to be constant for the statement:  https://www.postgresql.org/docs/8.1/xfunc-volatility.html

svenklemm: yes but that is not guaranteed either

svenklemm:
```
dev=# \df+ pg_relation_size
                                                                                                                                List of functions
   Schema   |       Name       | Result data type | Argument data types | Type | Volatility | Parallel |  Owner   | Security | Access privileges | Language |                  Source code                   |                            Description

------------+------------------+------------------+---------------------+------+------------+----------+----------+----------+-------------------+----------+------------------------------------------------+-----------------------------------------------------------------
---
 pg_catalog | pg_relation_size | bigint           | regclass            | func | volatile   | safe     | postgres | invoker  |                   | sql      | select pg_catalog.pg_relation_size($1, 'main') | disk space usage for the main fork of the specified table or ind
ex
 pg_catalog | pg_relation_size | bigint           | regclass, text      | func | volatile   | safe     | postgres | invoker  |                   | internal | pg_relation_size                               | disk space usage for the specified fork of a table or index
```

gayyappan: Ah, I see that: select hypertable_index_size('idx'),  hypertable_index_size('idx) ? The 2 function calls could return different values.

gayyappan: ^ Looks like the messages crossed and we agree. Modified to volatile.

tsl/test/expected/dist_views.out
```diff
@@ -89,20 +89,20 @@ SELECT * FROM chunks_detailed_size('dist_table'::regclass)
...
- _timescaledb_internal | _dist_hyper_1_3_chunk |        8192 |       32768 |           0 |       40960 | data_node_3
+ _timescaledb_internal | _dist_hyper_1_1_chunk |        8192 |       32768 |        8192 |       49152 | view_node_1
+ _timescaledb_internal | _dist_hyper_1_1_chunk |        8192 |       32768 |        8192 |       49152 | view_node_2
+ _timescaledb_internal | _dist_hyper_1_2_chunk |        8192 |       32768 |           0 |       40960 | view_node_2
+ _timescaledb_internal | _dist_hyper_1_2_chunk |        8192 |       32768 |           0 |       40960 | view_node_3
+ _timescaledb_internal | _dist_hyper_1_3_chunk |        8192 |       32768 |           0 |       40960 | view_node_1
+ _timescaledb_internal | _dist_hyper_1_3_chunk |        8192 |       32768 |           0 |       40960 | view_node_3
 (6 rows)
 
 SELECT * FROM hypertable_detailed_size('dist_table'::regclass);
```
svenklemm, line 100: missing order by

gayyappan: This query always returns a single record.

svenklemm: it has 3 records in the test

gayyappan: sorry, I misspoke.

gayyappan: fixed.

sql/size_utils.sql
```diff
@@ -74,7 +74,7 @@ RETURNS TABLE (
     index_bytes bigint,
     toast_bytes bigint,
     total_bytes bigint)
-LANGUAGE PLPGSQL STABLE STRICT AS
+LANGUAGE PLPGSQL VOLATILE STRICT AS
```
svenklemm, line 77: they are all PARALLEL SAFE are they not?

tsl/src/dist_util.c
mkindahl: I don't think you should remove the check. This seems to be used elsewhere in our code for strict functions and I think we should be consistent. I think it's fine to have tests that only intend to make coverage good, just to get the noise down, but I don't feel this is very important so you can leave it as it is as far as I am concerned.

test/expected/size_utils.out
```diff
@@ -179,8 +209,14 @@ SELECT * FROM hypertable_detailed_size(NULL);
...
+ hypertable_index_size 
+-----------------------
+                      
+(1 row)
 
+-- tests with tables that are not hypertables
+CREATE TABLE regtab( a integer, b integer);
+CREATE INDEX regtab_idx ON regtab( a);
+SELECT * FROM hypertable_index_size('regtab_idx');
+ERROR:  query returned no rows
```
mkindahl, line 222: Strange that this is an error. "No rows" are usually not considered an error.

sql/size_utils.sql
gayyappan: I do not think these are parallel safe as we potentially access other data nodes to get the information.

test/expected/size_utils.out
gayyappan: That's because we don't have any explicit code that handles non hypertables.

PR #2192, opened 2020-08-05
Add dist compression policy test, fix policy name

k-rus:
Adds test of the compression policy on a distributed hypertable. 
Currently the policy cannot be added as the distributed hypertables 
are not marked with enabled compression.

Fixes error messages to use new name for compression policy.

**PR note**
Issue to support or block compression policy on distributed hypertable is https://github.com/timescale/timescaledb/issues/2193

### DIFF
--- a/tsl/test/expected/dist_compression.out
+++ b/tsl/test/expected/dist_compression.out
@@ -324,7 +324,7 @@ ORDER BY chunk_name, node_name;
  _timescaledb_internal | _dist_hyper_1_3_chunk |        8192 |       32768 |           0 |       40960 | data_node_3
 (6 rows)
 
-SELECT * FROM hypertable_detailed_size('compressed'::regclass) 
+SELECT * FROM hypertable_detailed_size('compressed'::regclass);
  table_bytes | index_bytes | toast_bytes | total_bytes |  node_name  
 -------------+-------------+-------------+-------------+-------------
        16384 |       65536 |           0 |       81920 | data_node_1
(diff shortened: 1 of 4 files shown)


k-rus:
> i think the ts_test_auto_compress_chunks and similar functions should be dropped and the new api be used instead

@svenklemm Can you point to the API functions to be used?

k-rus:
> i think the ts_test_auto_compress_chunks and similar functions should be dropped and the new api be used instead

The corresponding changes were removed from the PR.

tsl/test/expected/dist_compression.out
```diff
@@ -324,11 +324,20 @@ ORDER BY chunk_name, node_name;
...
        16384 |       65536 |           0 |       81920 | data_node_2
 (3 rows)
 
+------------------------------------------------------
+-- Test compression policy on a distributed hypertable
+------------------------------------------------------
+INSERT INTO compressed VALUES 
+(now()::TIMESTAMPTZ, 1, 0.1), (now()::TIMESTAMPTZ, 2, 0.2);
+\set ON_ERROR_STOP 0
+SELECT add_compression_policy('compressed', '60d'::interval);
```
gayyappan, line 341: compression is enabled. But this fails. I think it is better to add support for the compression policy in the same PR. Or not add the test here.

k-rus: @gayyappan How can this error happen if the distributed hypertable is marked with the compression enabled? I suspect that the alter statement, which enables compression, is just propagated to data nodes, but is not set on the distributed hypertable. What am I missing?

k-rus:
Closing in favour #2193
Error messages fix will be in new PR.

k-rus:
> Error messages fix will be in new PR.

See https://github.com/timescale/timescaledb/pull/2196

svenklemm:
> i think the ts_test_auto_compress_chunks and similar functions should be dropped and the new api be used instead

you can call the procedures directly they are defined in sql/bgw_scheduler.sql but it requires a bit effort so i'm fine with current approach and changing it once we have `run_job`

PR #2194, opened 2020-08-05
Set invalidation threshold during continuous aggregate refresh

### LINKED ISSUE #2177: Move the invalidation threshold on continuous aggregate refresh
The invalidation threshold is used by the insert trigger as the threshold beyond which it should not write any invalidations. This avoids some write amplification that would otherwise occur in regions that are hot for writes.

Currently, the existing materializer code moves the invalidation threshold in its own transaction, prior to and separate from the transaction that does the materialization. This is important because any INSERTS that occur during materialization, and which would invalidate the materialized region, must be recorded while the materialization happens.

To retain this functionality for `refresh_continuous_aggregate`, it must be converted to a procedure instead of a function, since the latter doesn't support multiple transactions in a single call.

Note that procedures are executed with `CALL` instead of `SELECT`.

erimatnor:
The invalidation threshold governs the window of data from the head of
a hypertable that shouldn't be subject to invalidations in order to
reduce write amplification during inserts on the hypertable.

When a continuous aggregate is refreshed, the invalidation threshold
must be moved forward (or initialized if it doesn't previously exist)
whenever the refresh window stretches beyond the current threshold.

This change implements the functionality that initializes or moves the
invalidation threshold. This happens in its own transaction during a
refresh of a continuous aggregate in order to "publish" the new
threshold before materialization. The extra transaction requires an
access exclusive lock in order to avoid losing invalidations.

Some care has been taken to not unnecessarily lock the threshold table
with an access exclusive lock when the threshold needs no update. This
also avoids the extra transaction that "publishes" the new threshold
during a refresh of a continuous aggregate.

Note that the invalidation threshold currently cannot be moved
back. It might make sense to reset the threshold to the max time value
of a hypertable (or lower) in case recent data is deleted. However,
functionally it won't matter if the threshold remains---it will only
mean more write amplification until the data once again reaches beyond
the threshold.

Tests for setting the invalidation threshold are also added, including new
isolation tests for concurrency.

Closes #2177

### DIFF
--- a/sql/ddl_api.sql
+++ b/sql/ddl_api.sql
@@ -214,8 +214,8 @@ CREATE OR REPLACE FUNCTION  set_replication_factor(
 AS '@MODULE_PATHNAME@', 'ts_hypertable_distributed_set_replication_factor' LANGUAGE C VOLATILE;
 
 -- Refresh a continuous aggregate across the given window.
-CREATE OR REPLACE FUNCTION refresh_continuous_aggregate(
+CREATE OR REPLACE PROCEDURE refresh_continuous_aggregate(
     cagg                     REGCLASS,
     window_start             "any",
     window_end               "any"
-) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_continuous_agg_refresh' LANGUAGE C VOLATILE;
+) LANGUAGE C AS '@MODULE_PATHNAME@', 'ts_continuous_agg_refresh';
--- a/tsl/src/continuous_aggs/invalidation_threshold.c
+++ b/tsl/src/continuous_aggs/invalidation_threshold.c
@@ -1,0 +1,261 @@
+/*
+ * This file and its contents are licensed under the Timescale License.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-TIMESCALE for a copy of the license.
+ */
+#include <postgres.h>
+#include <access/htup_details.h>
+#include <access/htup.h>
+#include <access/xact.h>
+#include <nodes/memnodes.h>
+#include <storage/lockdefs.h>
+#include <storage/lmgr.h>
+#include <utils/memutils.h>
+#include <utils/snapmgr.h>
+
+#include <catalog.h>
+#include <scanner.h>
+#include <scan_iterator.h>
+#include <compat.h>
+
+#include "continuous_agg.h"
+#include "continuous_aggs/materialize.h"
+#include "invalidation_threshold.h"
+
+/*
+ * Invalidation threshold.
+ *
+ * The invalidation threshold acts as a dampener on a hypertable to make sure
+ * that invalidations written during inserts won't cause too much write
+ * amplification in "hot" regions---typically the "head" of the table. The
+ * presumption is that most inserts happen at recent time intervals, and those
+ * intervals will be invalid until writes move out of them. Therefore, it
+ * isnt't worth writing invalidations in that region since it is presumed
+ * out-of-date anyway. Further, although it is possible to refresh a
+ * continuous aggregate in those "hot" regions, it will lead to partially
+ * filled buckets. Thus, refreshing those intervals is discouraged since the
+ * aggregate will be immediately out-of-date until the buckets are filled. The
+ * invalidation threshold is, in other words, used as a marker that lags
+ * behind the head of the hypertable, where invalidations are written before
+ * the threshold but not after it.
+ *
+ * The invalidation threshold is moved forward (and only forward) by refreshes
+ * on continuous aggregates when it covers a window that streches beyond the
+ * current threshold. The invalidation threshold needs to be moved in its own
+ * transaction, with exclusive access, before the refresh starts to
+ * materialize data. This is to avoid losing any invalidations that occur
+ * between the start of the transaction that moves the threshold and its end
+ * (when the new threshold becomes visible).
+ *
+ * ______________________________________________
+ * |_______________________________________|_____| recent data
+ *                                        ^
+ *      invalidations written here        |  no invalidations
+ *                                        |
+ *                               invalidation threshold
+ *
+ */
+
+typedef struct InvalidationThresholdData
+{
+	int64 threshold;
+	bool was_updated;
+} InvalidationThresholdData;
+
+static ScanTupleResult
+scan_update_invalidation_threshold(TupleInfo *ti, void *data)
+{
+	InvalidationThresholdData *invthresh = data;
+	bool should_free;
+	HeapTuple tuple = ts_scanner_fetch_heap_tuple(ti, false, &should_free);
+	Form_continuous_aggs_invalidation_threshold form =
+		(Form_continuous_aggs_invalidation_threshold) GETSTRUCT(tuple);
+
+	if (invthresh->threshold > form->watermark)
+	{
+		HeapTuple new_tuple = heap_copytuple(tuple);
+		form = (Form_continuous_aggs_invalidation_threshold) GETSTRUCT(new_tuple);
+
+		form->watermark = invthresh->threshold;
+		ts_catalog_update(ti->scanrel, new_tuple);
+		heap_freetuple(new_tuple);
+		invthresh->was_updated = true;
+	}
+	else
+	{
+		elog(DEBUG1,
+			 "hypertable %d existing  watermark >= new invalidation threshold " INT64_FORMAT
+			 " " INT64_FORMAT,
+			 form->hypertable_id,
+			 form->watermark,
+			 invthresh->threshold);
+	}
+
+	if (should_free)
+		heap_freetuple(tuple);
+
+	return SCAN_DONE;
+}
+
+/* every cont. agg calculates its invalidation_threshold point based on its
+ *refresh_lag etc. We update the raw hypertable's invalidation threshold
+ * only if this new value is greater than the existsing one.
+ */
+bool
+continuous_agg_invalidation_threshold_set(int32 raw_hypertable_id, int64 invalidation_threshold)
+{
+	bool threshold_found;
+	InvalidationThresholdData data = {
+		.threshold = invalidation_threshold,
+		.was_updated = false,
+	};
+	ScanKeyData scankey[1];
+
+	ScanKeyInit(&scankey[0],
+				Anum_continuous_aggs_invalidation_threshold_pkey_hypertable_id,
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(raw_hypertable_id));
+
+	/* NOTE: this function deliberately takes an AccessExclusiveLock when updating the invalidation
+	 * threshold, instead of the weaker RowExclusiveLock lock normally held for such operations: in
+	 * order to ensure we do not lose invalidations from concurrent mutations, we must ensure that
+	 * all transactions which read the invalidation threshold have either completed, or not yet read
+	 * the value; if we used a RowExclusiveLock we could race such a transaction and update the
+	 * threshold between the time it is read but before the other transaction commits. This would
+	 * cause us to lose the updates. The AccessExclusiveLock ensures no one else can possibly be
+	 * reading the threshold.
+	 */
+	threshold_found =
+		ts_catalog_scan_one(CONTINUOUS_AGGS_INVALIDATION_THRESHOLD /*=table*/,
+							CONTINUOUS_AGGS_INVALIDATION_THRESHOLD_PKEY /*=indexid*/,
+							scankey /*=scankey*/,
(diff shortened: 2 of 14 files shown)


tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -0,0 +1,272 @@
...
+			 new_threshold,
+			 hyper_id);
+
+		rel = table_open(threshold_relid, AccessExclusiveLock);
+
+		/* Check if another transaction inserted a threshold before we got
+		 * exclusive lock on the table. */
+		threshold = invalidation_threshold_read(hyper_id, &tid, AccessExclusiveLock);
+
+		if (ItemPointerIsValid(&tid))
```
erimatnor, line 240: FWIW, I have tests for these two cases in the concurrent refresh test. But they are currently commented because they are flaky since two concurrent refresh process race to get to this case, or the other one below. Still trying to figure out if there is a way to make them deterministic, but probably not possible.

tsl/test/isolation/specs/continuous_aggs_concurrent_refresh.spec
```diff
@@ -65,111 +74,220 @@ setup
...
+# inserts of new thresholds when one doesn't previously exist. However
+# the tests are flaky because the two refreshes block on a lock and
+# will then race to the a second lock when the first lock is released.
+#
+
+# Check that two refreshes (R1 then R3), which both try to insert a
+# new threshold, will not lead to a conflict. R1 and R3 should both
+# block on trying to insert a new threshold, but only one of them
+# should succeed and the other should do an update instead.
+#permutation "L1_lock_threshold_table" "R1_refresh" "R3_refresh" "L1_unlock_threshold_table" "L2_read_unlock_threshold_table" "L3_unlock_cagg_table" "S1_select"
```
erimatnor, line 266: The two commented permutations here tests two concurrent inserts of a new threshold, but they are flaky since they rely on two blocked refreshes that race to insert a threshold after the lock is released. Enabling them leads to better test coverage but will make tests flaky. Continuing investigation, but figured best to leave in here for now given the expected questions on test coverage for the concurrent threshold insert case.

tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -0,0 +1,272 @@
...
+ * lost.
+ *
+ * Since the invalidation threshold is currently stored in a shared table, all
+ * INSERTs (that read the threshold) and refreshes of continuous aggregates
+ * will be blocked. Therefore, we try to avoid heavy locking if possible by
+ * first reading the threshold with a read lock. If the threshold needs no
+ * update, we need not grab any further locks.
+ */
+bool
+continuous_agg_invalidation_threshold_set(int32 hyper_id, int64 new_threshold)
```
gayyappan, line 178: We already have code that sets thresholds and acquires the right locks on tables for  materialization in materialize.c. IMO we should be looking at integrating the refresh code that can work on arbitrary windows with existing code instead of rewriting everything from scratch. we are not taking take advantage of all the bug fixes and tests that work with the existing code.

erimatnor: This isn't rewriting everything from scratch. I've studied the existing code extensively and I am trying to keep as much as possible, while at the same time applying improvements, refactoring, and cleanup. But just saying we should reuse the existing code, while changing the core functionality and semantics, is easier said than done.

Obviously, any help in making sure we maintain the lessons learned  is appreciated. But, IMO, `materialize.c` does too much and conflates a lot of different functionality (setting thresholds, materializing, etc.). It is also difficult to understand, TBH. Separating the different functionality is an improvement. Thus, I am trying clean things up and break out different functionality into separate source files. I am trying to do this without also "blowing up the house" at the same time. If I would simply reuse and change the existing code, it would have ripple effects, and even affect existing tests---something I don't want to do at this point while things are still in flux. It would cause too many changes in a single PR and would make things harder to review. So, I am trying to keep things as stable as possible, introducing one distinct change at a time.

While doing the refactoring, I am also trying to fix a number of issues, for instance, the lack of memory context use. This code also improves on the previous function to set the threshold by not always grabbing an `AccessExclusiveLock` unless necessary.

erimatnor: @gayyappan I've added additional commits to integrate with the existing code and to show the history of refactoring. The PR is now a lot bigger, with more changes, but hopefully this better matches your expectations.

sql/ddl_api.sql
```diff
@@ -214,8 +214,8 @@ CREATE OR REPLACE FUNCTION  set_replication_factor(
 AS '@MODULE_PATHNAME@', 'ts_hypertable_distributed_set_replication_factor' LANGUAGE C VOLATILE;
 
 -- Refresh a continuous aggregate across the given window.
-CREATE OR REPLACE FUNCTION refresh_continuous_aggregate(
+CREATE OR REPLACE PROCEDURE refresh_continuous_aggregate(
```
svenklemm, line 217: i think this requires some discussion which parts we want as procedure and which parts we want as function, while a procedure allows transaction control it has some disadvantages and is gives less flexibility to the caller.

tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -56,149 +56,235 @@
...
+		{
+			elog(DEBUG2,
+				 "tried to set invalidation threshold to " INT64_FORMAT
+				 " but it was updated to " INT64_FORMAT " before getting update lock, skipping",
+				 new_threshold,
+				 threshold);
+		}
+	}
+	else
+	{
```
gayyappan, line 242: I am wary of adding the optimizations from the last commit  (avoiding accessexclusivelock unless necessary) without the other pieces of the code in place.   I don't  understand the whole locking picture (that will come with the other PRs)  and that's my reasoning behind the caution here.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -372,6 +386,19 @@ continuous_agg_refresh(PG_FUNCTION_ARGS)
 				 errmsg("invalid refresh window"),
 				 errhint("The start of the window must be before the end.")));
 
+	LockRelationOid(catalog_get_table_id(catalog, CONTINUOUS_AGGS_INVALIDATION_THRESHOLD),
+					AccessExclusiveLock);
+
+	if (continuous_agg_invalidation_threshold_set(cagg->data.raw_hypertable_id, refresh_window.end))
```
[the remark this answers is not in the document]
erimatnor: TBH, I don't understand the numerous locks taken in the materializer code. Would appreciate some explanation and guidance. In particular, I don't understand the session locks that last across the two transactions. There's a comment that says "SessionLock the objects we are using for this materialization to ensure they're not alter in a way which would invalidate our work." The objects locked are materialized and raw hypertables, the various views, etc. But I can't make sense of this since the first transaction only moves the invalidations from hyper to cagg log and (potentially) updates the invalidation threshold. Thus the work in the first transaction doesn't really depend on these objects (beyond the individual thresholds and invalidation entries). 

The only thing  that I can see that these session-locks protect against is dropping the objects between the first and the second transaction. But since the work done is to process invalidations for _all_ hypertables, then why aren't _all_ hypertables referenced in the log locked, including all their caggs and dependent objects?

Thus, the first transaction seems to be pure bookkeeping and I don't see the harm in simply failing the second transaction if the objects no longer exist, like you would normally do if, e.g., a table was dropped just as someone was about to query it. In any case, if we care about not having objects removed under our feet, I think it is fine to add these session locks later. They seem more to be there to avoid unexpected error for the user--not for correctness. But maybe I am missing something?

It seems also that many locks are taken just before a scan that then takes the same locks. If the concern is that the scan will release the lock before transaction end, it is now possible to specify that locks are kept after the scan. This means that many locks that are taken explicitly, then immediately taken again in the scan, are no longer needed since the scan takes the appropriate lock. I've never before seen this pattern of taking the same lock twice, neither in most of our code nor in PostgreSQL itself.

tsl/test/expected/continuous_aggs_invalidation.out
```diff
@@ -207,8 +253,7 @@ SELECT hypertable_id AS hyper_id,
  hyper_id | start | end 
 ----------+-------+-----
         2 |    20 |  20
-        2 |    30 |  80
-(2 rows)
+(1 row)
```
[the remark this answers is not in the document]
erimatnor: The  `continuous_aggs_materialization_invalidation_log` is shown in the next query below this one.

The hypertable invalidation log contains invalidations for _all_ hypertables, in this case ID 1 and 2. During a refresh, only the entries that are on hypertable 1, which the cagg exists on, will be moved over. Refreshing a cagg on the other hypertable would move those entries over.

There's a comment above that makes this clear.

But this PR is about moving the invalidation threshold, so moving invalidations between logs is not the subject of this PR. The only reason there's a change here is because now it was possible to replace the `REFRESH MATERIALIZED VIEW` command with a `continuous_aggregate_refresh` on a window (since it now moves the invalidation threshold so that new invalidation are generated).

sql/ddl_api.sql
erimatnor: This was discussed in design docs and if we are to retain the old functionality of how the invalidation threshold is set there is no real option.

tsl/src/continuous_aggs/invalidation_threshold.c
erimatnor: I reverted the optimizations and will submit this as a separate PR.

tsl/src/continuous_aggs/refresh.c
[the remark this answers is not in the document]
gayyappan: Yes, they are being locked to prevent objects from being removed from under us. If a DROP or ALTER is in progress , you would want the materialization to wait and fail. And vice-versa. 
Yes, the explicit lock + catalog lock is due to the fact that the catalog code also acquires the lock and releases it. Having the locks in one place helps reason about it as well.

tsl/test/expected/continuous_aggs_invalidation.out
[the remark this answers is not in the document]
gayyappan: Thanks for the clarification. Yes, I see the logs for the materialization_invalidation_log below this test case.  I thought the invalidations belonged to hypertable-id 2 and not to 1.

tsl/src/continuous_aggs/refresh.c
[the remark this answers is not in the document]
erimatnor: Since this PR is just about setting the invalidation threshold, does it sound reasonable that we review the need for the session locks and possible additions separately (for later PR)?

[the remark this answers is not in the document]
gayyappan: Yes, would be good to make a pass over the locking after all the functionality is in place.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -366,6 +387,19 @@ continuous_agg_refresh(PG_FUNCTION_ARGS)
...
+	{
+		/* Start a new transaction if the threshold was set. Note that this
```
k-rus, line 395:
```suggestion
		/* Start a new transaction if the threshold was updated. Note that this
```

tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	if (should_free)
+		heap_freetuple(tuple);
+
+	return SCAN_DONE;
+}
+
+/* every cont. agg calculates its invalidation_threshold point based on its
+ *refresh_lag etc. We update the raw hypertable's invalidation threshold
```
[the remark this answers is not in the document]
erimatnor: Since this code was just moved from another file, I'll defer fixing this until a separate PR. I did include a refactoring commit of this code previously, but it was suggested to not do it in this PR, which makes sense.

tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -0,0 +1,261 @@
...
+ * only if this new value is greater than the existsing one.
```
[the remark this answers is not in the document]
erimatnor: Same here, this code was moved as is from another file. A  later refactoring PR will change this.

tsl/src/continuous_aggs/invalidation_threshold.c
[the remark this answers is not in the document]
k-rus: You do similar fixes in other PRs even in the code, which is not touched. So I see no good reason to keep as it is. Also there is no way to track that it was copied code, but thank you for clarifying this, since I was sure it is new code.

PR #2201, opened 2020-08-06
Add check for distributed hypertable to continuous aggs

pmwkaa:
Show an error message in case if a distributed hypertable being used.

Issue: https://github.com/timescale/timescaledb-private/issues/835

### DIFF
--- a/tsl/src/continuous_aggs/create.c
+++ b/tsl/src/continuous_aggs/create.c
@@ -784,6 +784,11 @@ cagg_validate_query(Query *query)
 		Dimension *part_dimension = NULL;
 		ht = ts_hypertable_cache_get_cache_and_entry(rte->relid, CACHE_FLAG_NONE, &hcache);
 
+		if (hypertable_is_distributed(ht))
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("continuous aggregates are not supported on distributed hypertables")));
+
 		/* there can only be one continuous aggregate per table */
 		switch (ts_continuous_agg_hypertable_status(ht->fd.id))
 		{
--- a/tsl/test/expected/dist_ddl.out
+++ b/tsl/test/expected/dist_ddl.out
@@ -1820,6 +1820,36 @@ DROP INDEX disttable_device_idx;
 ERROR:  [data_node_1]: index "disttable_device_idx" does not exist
 \set ON_ERROR_STOP 1
 DROP TABLE disttable;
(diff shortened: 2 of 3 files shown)


tsl/src/continuous_aggs/create.c
```diff
@@ -784,6 +784,13 @@ cagg_validate_query(Query *query)
 		Dimension *part_dimension = NULL;
 		ht = ts_hypertable_cache_get_cache_and_entry(rte->relid, CACHE_FLAG_NONE, &hcache);
 
+		if (hypertable_is_distributed(ht))
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("hypertable is a distributed hypertable"),
+					 errhint("creating continuous aggregates for distributed hypertables is not "
```
erimatnor, line 791: I would use errdetail here since hint is usually about how to resolve/fix the error.  But with my suggested change to main error this detail seem a bit redundant.

tsl/src/continuous_aggs/create.c
```diff
@@ -784,6 +784,13 @@ cagg_validate_query(Query *query)
 		Dimension *part_dimension = NULL;
 		ht = ts_hypertable_cache_get_cache_and_entry(rte->relid, CACHE_FLAG_NONE, &hcache);
 
+		if (hypertable_is_distributed(ht))
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("hypertable is a distributed hypertable"),
```
erimatnor, line 790: I suggest: "continuous aggregates are not supported on distributed hypertables" to give more context.

pmwkaa:
> Please use a combined test file for simple stuff like this you are creating 4 databases just to check that caggs dont work on distributed hypertables

But the idea behind it was to continue with this file for distributed hypertable related stuff. We have similar dist_compression and there were no caggs related before. Any better suggestion? @svenklemm

tsl/test/sql/CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ set(TEST_FILES_DEBUG
   deparse.sql
   dist_commands.sql
   dist_compression.sql
+  dist_continuous_aggs.sql
```
gayyappan, line 43: Just so you know and don't have to spend time debugging if you run into this: I have seen failures on CI when multiple dist tests create data nodes with the same name. Renaming the data nodes to something unique to the test file usually resolves this.

svenklemm:
since you only want to see the error you could put it into dist_ddl

PR #2209, opened 2020-08-11
Add distributed hypertable to regresscheck-shared

svenklemm:
This patch sets up a distributed hypertable in the regresscheck-shared
environment to enable running distributed tests.


tsl/test/shared/sql/include/shared_setup.sql
```diff
@@ -72,3 +74,24 @@ ALTER TABLE metrics_space_compressed SET (timescaledb.compress, timescaledb.comp
 SELECT compress_chunk(c.schema_name|| '.' || c.table_name)
 FROM _timescaledb_catalog.chunk c, _timescaledb_catalog.hypertable ht where c.hypertable_id = ht.id and ht.table_name = 'metrics_space_compressed' and c.compressed_chunk_id IS NULL
 ORDER BY c.table_name DESC;
+
+RESET ROLE;
+-- Add data nodes
+SELECT * FROM add_data_node('data_node_1', host => 'localhost', database => 'data_node_1');
+SELECT * FROM add_data_node('data_node_2', host => 'localhost', database => 'data_node_2');
+SELECT * FROM add_data_node('data_node_3', host => 'localhost', database => 'data_node_3');
```
k-rus, line 82: This means that all tests, which will use the shared setup, need to be run sequentially, since creating the databases under the same names during `add_data_node` will fail. Or do I miss anything?

svenklemm:
Yes it will because all shared tests will use the same. You dont need setup datanodes in individual tests in regresscheck-shared

tsl/test/shared/sql/include/shared_setup.sql
svenklemm: this is the pre-setup script for the shared environment, shared tests always run in parallel and they will all use the same data nodes.

svenklemm:
correct

> LGTM
> If I understand correctly, the shared test setup is executed once before tests are run in parallel.

correct

Issue #1902, opened 2020-05-14
hypertable_approximate_row_count returns 0 on distributed hypertable


noctarius:
**Relevant system information:**
PostgreSQL version: 11
TimescaleDB version: 2.0.0-beta4
Installation method: Helm chart (with image changed to 2.0.0-beta4-pg11)

**Describe the bug**
When executing *hypertable_approximate_row_count* on a distributed hypertable, the result seems to always be 0, which certainly is wrong in my case.

**To Reproduce**
1. Create a distributed hypertable
2. Generate test data
3. Execute *hypertable_approximate_row_count*
4. See estimated row count is 0

**Expected behavior**
Push the estimation request down to the data nodes and sum up the result on the access node. It's an estimation anyways, therefore async is fine and transactions don't matter.

**Actual behavior**
Result is always 0. My guess the estimation happens on the access node's local hypertable which is empty.

pmwkaa:
@noctarius I guess this function should work out of the box after `ANALYZE` being executed on the access node

pmwkaa:
We also done some refactoring and introduced support for compression and regular tables: https://github.com/timescale/timescaledb/issues/2252

PR #2220, opened 2020-08-14
Add distributed hypertable support to hypertable_approximate_row_count()

pmwkaa:
Related to: https://github.com/timescale/timescaledb/issues/1902

### DIFF
--- a/tsl/test/expected/dist_api_calls.out
+++ b/tsl/test/expected/dist_api_calls.out
@@ -172,6 +172,17 @@ SELECT * FROM disttable ORDER BY time;
  Mon Jul 02 08:01:00 2018 PDT |     87 |        0
 (8 rows)
 
+SELECT * FROM show_chunks('disttable');
+                 show_chunks                 
+---------------------------------------------
+ _timescaledb_internal._dist_hyper_1_4_chunk
+ _timescaledb_internal._dist_hyper_1_5_chunk
+ _timescaledb_internal._dist_hyper_1_6_chunk
+ _timescaledb_internal._dist_hyper_1_7_chunk
+ _timescaledb_internal._dist_hyper_1_8_chunk
+ _timescaledb_internal._dist_hyper_1_9_chunk
+(6 rows)
+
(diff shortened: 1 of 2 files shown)


sql/size_utils.sql
```diff
@@ -365,6 +365,76 @@ BEGIN
...
+$BODY$
+DECLARE
+BEGIN
+	RETURN QUERY
+	SELECT h.schema_name,
+		h.table_name,
+		row_estimate.row_estimate
+	FROM _timescaledb_catalog.hypertable h
+	CROSS JOIN LATERAL (
+		SELECT sum(cl.reltuples)::BIGINT AS row_estimate
```
gayyappan, line 427: We have to take compressed chunks into account. The row count before compression is in the compression stats tables.
We have to account for duplicate chunks for the distributed case .

pmwkaa: We haven't done this before, maybe introduce compression support in separate PR in that case? 

Not really sure about the duplicate chunks case, can you specify? I am not sure we support native replication yet

svenklemm: actually with brians patch applied this will get correct stats for compressed and uncompressed chunks

pmwkaa:
Will try different approach after discussion with Erik

pmwkaa:
@gayyappan @svenklemm  

Apparently `hypertable_approximate_row_count()` should work out of the box right after `ANALYZE` being executed on distributed hypertable. We support `ANALYZE` in a way that it will fetch stats from remote nodes and apply them locally on the access node. There is even native replication support already, because access node does not create additional chunks, but data nodes keeps the duplicates. In case of replicated chunk last stats for the same chunk will be applied (stats are essentially should be the same on copies).

So instead of previous implementation, I've added tests cases instead.

k-rus (commented):
The PR adds test for the function. Thus the commit message should reflect what the commits do. Current message is misleading as there is only test code.

erimatnor (approved):

PR #2222, opened 2020-08-15
Fix negative bitmapset member not allowed in decompression

svenklemm:
Due to a typo in the code the relid of the relation instead of the
rangetable index was used when building the pathkey for the
DecompressChunk node. Since relids are unsigned but bitmapsets
use signed int32 this lead to 'negative bitmapset member not allowed'
being thrown as error when the relid is greater than INT32_MAX.
This patch also adds an assert to prevent this from happening again.


svenklemm:
While its possible to create a test i don't think it's reasonable as we would have to make postgres consume 2 billion oids to guarantee we get a relid that triggers the bug

PR #2223, opened 2020-08-15
Fix signature of cagg_watermark

### LINKED ISSUE #2216: Fix signature for _timescaledb_internal.cagg_watermark

svenklemm:
This patch changes the signature from cagg_watermark(oid) to
cagg_watermark(int). Since this is an API breaking change it couldn't
be done in an earlier release.

Fixes #2216

### DIFF
--- a/sql/updates/post-update.sql
+++ b/sql/updates/post-update.sql
@@ -16,3 +16,6 @@ BEGIN
 END
 $$;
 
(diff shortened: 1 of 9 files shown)


svenklemm:
This is cleanup for #1869 in 1.7.1 which we couldnt do back then because it would break API

Issue #2167, opened 2020-07-31
Internal now() returns system clock time instead of transaction time


erimatnor:
In short: `ts_get_now_internal` uses `GetCurrentTime()` as its standard time function. This returns system clock instead of the transaction timestamp, which means that the time will vary between each call to this function within the same transaction. Note that the standard `now()` function returns transaction time, so our internal time function departs from the standard. This can affect queries and other things that rely on getting a consistent time for a transaction.

PR #2235, opened 2020-08-20
Fix ts_get_now_internal to use transaction time

pmwkaa:
Related Issue: #2167

Not really sure how to test it, since we not exported this function. If necessary I guess I could make a test c function to wrap it up and then check that the result of it does not change when executed inside an active transaction more then once.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -759,14 +759,14 @@ ts_get_now_internal(const Dimension *open_dim)
 #ifdef TS_DEBUG
 	Datum now_datum;
 	if (ts_current_timestamp_mock == NULL || strlen(ts_current_timestamp_mock) == 0)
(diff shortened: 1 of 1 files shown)


svenklemm (changes requested):
i think the current behaviour is intentional and not a bug

mkindahl (approved):

erimatnor:
FWIW, this now-replacement is used in two places in the continuous aggregate code: first to set the modification time for invalidations in the insert trigger, then in the materializer to calculate the end of the range to materialize. IMO, transaction time makes sense in both of these cases.

pmwkaa:
@svenklemm any thoughts, suggestions?

svenklemm:
since caggs seems to be only user of this function i guess its no problem to change it to whatever makes most sense for current caggs

svenklemm (approved):

PR #2240, opened 2020-08-21
Fix segfault in alter_job

svenklemm:
When trying to alter a job with NULL config alter_job did not
set the isnull field for config and would segfault when trying
to build the resultset tuple.


mkindahl (approved):

k-rus (approved):

erimatnor (approved):
LGTM. On a related note (and maybe for a separate PR), there's a lack of NULL checks for  the job_id and if_exists parameters in `job_alter`. Maybe it is good to add those explicit checks as well, including tests for NULL input (e.g., `alter_job(NULL, ...)`?

PR #2245, opened 2020-08-22
Cleanup drop and show chunks

### LINKED ISSUE #2246: Make drop_chunks and show_chunks similar
For consistency, `drop_chunks` and `show_chunks` should have similar semantics. They used to behave the same (e.g., the hypertable was optional) but they have diverged since the hypertable became mandatory for `drop_chunks`,  while it is still optional for `show_chunks`.

There's also a bunch of code related to these functions that is no longer used (e.g., cascading-related code paths) and there's an opportunity to unify and simplify the shared code.

erimatnor:
This change removes, simplifies, and unifies code related to
`drop_chunks` and `show_chunks`. As a result of prior changes to
`drop_chunks`, e.g., making table relid mandatory and removing
cascading options, there's an opportunity to clean up and simplify the
rather complex code for dropping and showing chunks.

In particular, `show_chunks` is now consistent with `drop_chunks`; the
relid argument is mandatory, a continuous aggregate can be used in
place of a hypertable, and the input time ranges are checked and
handled in the same way.

Unused code is also removed, for instance, code that cascaded drop
chunks to continuous aggregates remained in the code base while the
option no longer exists.

Fixes #2246


test/sql/chunk_utils.sql
```diff
@@ -318,12 +309,12 @@ BEGIN;
...
-    SELECT show_chunks(hypertable => 'drop_chunk_test_tstz', older_than => now() - interval '1 minute', newer_than => now() - interval '6 minute');
-    SELECT show_chunks(hypertable => 'drop_chunk_test_tstz', newer_than => now() - interval '1 minute');
-    SELECT show_chunks(hypertable => 'drop_chunk_test_tstz', older_than => now() - interval '1 minute');
+    SELECT show_chunks('drop_chunk_test_tstz');
+    SELECT show_chunks('drop_chunk_test_tstz', older_than => now() - interval '1 minute', newer_than => now() - interval '6 minute');
+    SELECT show_chunks('drop_chunk_test_tstz', newer_than => now() - interval '1 minute');
+    SELECT show_chunks('drop_chunk_test_tstz', older_than => now() - interval '1 minute');
 
-    \set QUERY1 'SELECT show_chunks(older_than => interval \'1 minute\', hypertable => \'drop_chunk_test_tstz\')::NAME'
+    \set QUERY1 'SELECT show_chunks(older_than => interval \'1 minute\', hypertable_or_cagg => \'drop_chunk_test_tstz\')::NAME'
```
mkindahl, line 317: To match the `drop_chunks` call below.
```suggestion
    \set QUERY1 'SELECT show_chunks(\'drop_chunk_test_tstz\', interval \'1 minute\')::NAME'
```

test/sql/relocate_extension.sql
```diff
@@ -52,19 +52,19 @@ SELECT AVG(temp) AS avg_tmp, "testSchema0".time_bucket('1 day', time, INTERVAL '
 
 -- testing drop_chunks START
 -- show_chunks and drop_chunks output should be the same
-\set QUERY1 'SELECT "testSchema0".show_chunks(older_than => \'2017-03-01\'::timestamp, hypertable => \'test_ts\')::REGCLASS::TEXT'
+\set QUERY1 'SELECT "testSchema0".show_chunks(older_than => \'2017-03-01\'::timestamp, hypertable_or_cagg => \'test_ts\')::REGCLASS::TEXT'
```
mkindahl, line 55:
```suggestion
\set QUERY1 'SELECT "testSchema0".show_chunks(\'test_ts\', \'2017-03-01\'::timestamp)::REGCLASS::TEXT'
```

test/sql/relocate_extension.sql
```diff
@@ -52,19 +52,19 @@ SELECT AVG(temp) AS avg_tmp, "testSchema0".time_bucket('1 day', time, INTERVAL '
...
 \set QUERY2 'SELECT "testSchema0".drop_chunks(\'test_ts\', \'2017-03-01\'::timestamp)::TEXT'
 \set ECHO errors
 \ir  :QUERY_RESULT_TEST_EQUAL_RELPATH
 \set ECHO all
 SELECT * FROM test_ts ORDER BY time;
-\set QUERY1 'SELECT "testSchema0".show_chunks(older_than => interval \'1 minutes\', hypertable => \'test_tz\')::REGCLASS::TEXT'
+\set QUERY1 'SELECT "testSchema0".show_chunks(older_than => interval \'1 minutes\', hypertable_or_cagg => \'test_tz\')::REGCLASS::TEXT'
```
mkindahl, line 61:
```suggestion
\set QUERY1 'SELECT "testSchema0".show_chunks(\'test_tz\', interval \'1 minutes\')::REGCLASS::TEXT'
```

test/sql/relocate_extension.sql
```diff
@@ -52,19 +52,19 @@ SELECT AVG(temp) AS avg_tmp, "testSchema0".time_bucket('1 day', time, INTERVAL '
...
 \set QUERY2 'SELECT "testSchema0".drop_chunks(\'test_tz\', interval \'1 minutes\')::TEXT'
 \set ECHO errors
 \ir  :QUERY_RESULT_TEST_EQUAL_RELPATH
 \set ECHO all
 SELECT * FROM test_tz ORDER BY time;
-\set QUERY1 'SELECT "testSchema0".show_chunks(older_than => interval \'1 minutes\', hypertable => \'test_dt\')::REGCLASS::TEXT'
+\set QUERY1 'SELECT "testSchema0".show_chunks(older_than => interval \'1 minutes\', hypertable_or_cagg => \'test_dt\')::REGCLASS::TEXT'
```
mkindahl, line 67:
```suggestion
\set QUERY1 'SELECT "testSchema0".show_chunks(\'test_dt\', interval \'1 minutes\')::REGCLASS::TEXT'
```

tsl/src/continuous_aggs/refresh.c
```diff
@@ -236,70 +237,6 @@ continuous_agg_refresh_with_window(ContinuousAgg *cagg, const InternalTimeRange
...
-				break;
-			default:
-				ereport(ERROR,
-						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-						 errmsg("invalid window parameter"),
-						 errhint("The window parameter requires an explicit cast.")));
-		}
-	}
-
-	return ts_time_value_to_internal(arg, argtype);
```
mkindahl, line 300: It seems this function belongs to the `time_utils` module as well.

src/chunk.c
```diff
@@ -3419,21 +3227,18 @@ ts_chunk_do_drop_chunks(Hypertable *ht, Datum older_than_datum, Datum newer_than
...
-											   older_than_type,
-											   newer_than_type,
-											   "drop_chunks",
-											   CurrentMemoryContext,
-											   &num_chunks,
-											   &tuplock);
+	chunks = get_chunks_in_time_range(ht,
+									  older_than,
+									  newer_than,
+									  "drop_chunks",
```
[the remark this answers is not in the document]
k-rus: The macro is already defined in this file:
```
#define DROP_CHUNKS_FUNCNAME "drop_chunks"
``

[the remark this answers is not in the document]
Comment (mkindahl) [reply src/chunk.c:3233]: Even more reason to use the macro then.

File: sql/ddl_api.sql [hunk 7]
```diff
@@ -92,26 +92,19 @@ CREATE OR REPLACE FUNCTION  set_number_partitions(
...
--- if and only if all the hypertables in the database
--- have the same type as the given time constraint argument
+-- `hypertable` must be a valid hypertable or continuous aggregate.
 CREATE OR REPLACE FUNCTION show_chunks(
-    hypertable  REGCLASS = NULL,
-    older_than "any" = NULL,
-    newer_than "any" = NULL
+    hypertable_or_cagg     REGCLASS,
+    older_than             "any" = NULL,
+    newer_than             "any" = NULL
```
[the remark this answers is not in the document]
erimatnor: The description wasn't accurate. I'll add back a basic one.

src/chunk.c
```diff
@@ -3368,12 +3180,8 @@ lock_referenced_tables(Oid table_relid)
...
 }
 
-/* Continuous agg materialization hypertables can be dropped
- * only if a user explicitly specifies the table name
- */
 List *
-ts_chunk_do_drop_chunks(Hypertable *ht, Datum older_than_datum, Datum newer_than_datum,
-						Oid older_than_type, Oid newer_than_type, int32 log_level,
+ts_chunk_do_drop_chunks(Hypertable *ht, int64 older_than, int64 newer_than, int32 log_level,
 						List **affected_data_nodes)
```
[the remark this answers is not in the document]
erimatnor: Nothing has changed w.r.t. data nodes in this PR.

src/chunk.c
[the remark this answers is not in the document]
erimatnor: This was there prior to my changes, but I'll fix it.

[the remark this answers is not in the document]
k-rus: @erimatnor You are right it is in the file with the same name but in TSL. (This practice with repeating the name is sometimes/often confusing.)

src/time_utils.c
```diff
@@ -0,0 +1,127 @@
...
+		case TIMESTAMPTZOID:
+			return DirectFunctionCall2(timestamptz_mi_interval, res, IntervalPGetDatum(interval));
+		case DATEOID:
+			res = DirectFunctionCall1(timestamptz_timestamp, res);
+			res = DirectFunctionCall2(timestamp_mi_interval, res, IntervalPGetDatum(interval));
+			return DirectFunctionCall1(timestamp_date, res);
+		default:
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("unknown time type %s", format_type_be(timetype))));
```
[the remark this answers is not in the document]
erimatnor: It is not possible with the current code since the only other supported (possible) time types are integer ones and this case is checked prior to calling this function. Note that this code was there before, I just moved it to a separate file so that it can be reused.

test/sql/chunk_utils.sql
erimatnor: I think is is intentional to test different parameter order. In any case, this was there before.

test/sql/relocate_extension.sql
erimatnor: Testing different parameter order is intentional.

test/sql/chunk_utils.sql
mkindahl: I made changes here before but didn't update calls to `show_chunks` since it used different parameters. IIRC, the original code looked the same for `show_chunks` and `drop_chunks` and I also believe there is a comment about that somewhere.

tsl/src/continuous_aggs/refresh.c
erimatnor: If you are referring to `get_time_value_from_arg`; it is already in that file (and the reason the time_utils.c was created)

If you are referring to `ts_time_value_to_internal`, then that would be part of a larger refactor (there are lots of messay  time-related functions in `utils.c`) which is out-of-scope of this PR.

src/time_utils.c
```diff
@@ -0,0 +1,127 @@
...
+				arg = OidFunctionCall3(infuncid,
+									   arg,
+									   ObjectIdGetDatum(InvalidOid),
+									   Int32GetDatum(-1));
+				break;
+			default:
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg("invalid window parameter"),
+						 errhint("The window parameter requires an explicit cast.")));
```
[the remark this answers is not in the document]
erimatnor: This shouldn't be possible since we currently do not support timetypes that take anything but 1 or 3 input arguments. I guess we can make this an assertion, but I think this really needs to return an error since an assertion would not be in release code.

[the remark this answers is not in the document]
mkindahl: Ack.

test/sql/relocate_extension.sql
mkindahl: Not sure why this is necessary. This is something that PG handles, so there is no reason to test that different parameter orders work.

Changing the order to avoid the keywords will also avoid having to change the names of the parameters if we decide to use different names or correct typos.

tsl/src/continuous_aggs/refresh.c
mkindahl: I was referring to `ts_time_value_to_internal`. Ack on deferring to a larger refactor.

test/sql/chunk_utils.sql
erimatnor: Can you explain why the parameter order is important and why both calls need to look the same? The current order is still a valid way to call this function so I am not sure I see a problem with what's there now beyond maybe visual preference.

sql/ddl_api.sql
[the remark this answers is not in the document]
mkindahl: Well, it is missing some parameters in that case and the given parameter have the wrong name. (It's `hypertable_or_cagg`, not `hypertable`.)

test/sql/chunk_utils.sql
mkindahl: AFAICT, it is not important, but it seems it was to the original author.

src/chunk.c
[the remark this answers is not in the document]
erimatnor: @k-rus I agree and I moved the macro to the non-TSL header so that it can be referenced from both places.

test/sql/relocate_extension.sql
erimatnor: If the test needs a cleanup than I suggest we do that separately. I am just doing the changes necessary for this PR.

src/time_utils.c
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * API functions that take supported time types as arguments often use a
+ * pseudo-type paremeter to represent these. For instance, the "any"
+ * pseudo-type is often used to represent any of these supported types.
+ *
+ * The downside of "any", however, is that it lacks type information and often
+ * forces users to add explicit type casts. For instance, with the following
+ * API call
+ *
+ * drop_chunk('conditions', '2020-10-01');
```
svenklemm, line 54: drop_chunk => drop_chunks

sql/ddl_api.sql
```diff
@@ -92,26 +92,19 @@ CREATE OR REPLACE FUNCTION  set_number_partitions(
 -- Drop chunks older than the given timestamp for the specific
 -- hypertable or continuous aggregate.
 CREATE OR REPLACE FUNCTION drop_chunks(
-    hypertable_or_cagg  REGCLASS,
-    older_than "any" = NULL,
-    newer_than "any" = NULL,
-    verbose BOOLEAN = FALSE
+    hypertable_or_cagg     REGCLASS,
```
svenklemm, line 95: not a fan of this name cause this will get ugly if we add support for more object types, would prefer either just hypertable or a generic name like object or relation

mkindahl (approved):

sql/ddl_api.sql
erimatnor: I am OK with relation, but FWIW, this was there before and I am just making things consistent.

svenklemm (approved):

PR #2249, opened 2020-08-24
Fix segfault when executing jobs with NULL config

svenklemm:
This patch adds check for NULL config as makeConst will segfault
when trying to detoast datum when passed a NULL config.


svenklemm:
Tests for this are in job_type removal PR

mkindahl (approved):

k-rus (approved):

PR #2250, opened 2020-08-24
Ignore build and IDE folders

k-rus:
Updating gitignore to exclude default build folders and CLion folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ typedef.list
 
 /CMakeSettings.json
 /out/*
+/debug/
+/cmake-build-debug/
+/.idea/


mkindahl (commented):
These directories are specific to your setup and are typically put in your personal `~/.gitignore` (see https://git-scm.com/docs/gitignore for more information).

IMHO, we should just keep patterns here that matches files that are a part of the build system and that are not stored in the build directory (or that are common enough that we want to avoid accidentally adding them, such as `*.bak`, `*.old`, and `*.rej`). I know there are a lot more patterns here than what is motivated and I think we should start removing patterns and follow the rules in the link above.

svenklemm:
I agree with mats here these are your tool preferences so you they should be handled in your personal .gitignore

mkindahl (approved):

Issue #2252, opened 2020-08-24
Make hypertable_approximate_row_count just return row count


erimatnor:
The function `hypertable_approximate_row_count` currently returns count and the schema+name of the table that was given as input. This latter two return values are unnecessary and make the function quite different from `count` that it is trying to provide an alternative for.

Thus, the function should be changed to just return the count, which also makes it easier to use in queries:

```
SELECT approximate_row_count(*)
FROM mytable;
```

JH-D:
@erimatnor  @pmwkaa 
Hello, may I ask a question.
**What is the correct query method in 2.4.1?**

I make a distributed hypertable and insert billions rows, but the result is 0.
![图片](https://user-images.githubusercontent.com/7844108/132431714-97150e60-c9fe-4ac0-bfc8-36527a52cc95.png)
![图片](https://user-images.githubusercontent.com/7844108/132431849-bcad11e1-3410-481f-8c88-e888bd37fd5e.png)

## Relevant system information:
- Install method: docker, image is "timescale/timescaledb:latest-pg12"
- Postgresql version: PostgreSQL 12.8 on x86_64-pc-linux-musl, compiled by gcc (Alpine 10.3.1_git20210424) 10.3.1 20210424, 64-bit
- TimescaleDB version: 2.4.1
- Multi-node setup: 1 access node, 2 data node.

PR #2262, opened 2020-08-26
Make hypertable_approximate_row_count return row count only

pmwkaa:
This change renames function to `approximate_row_count()` and adds support for regular tables. Return a row count estimate for a table instead of a table list. Add test for compression.

Issue: https://github.com/timescale/timescaledb/issues/2252 and https://github.com/timescale/timescaledb/issues/1741

### DIFF
--- a/sql/size_utils.sql
+++ b/sql/size_utils.sql
@@ -367,52 +367,48 @@ $BODY$;
 
 -- Convenience function to return approximate row count
 --
--- main_table - hypertable to get approximate row count for; if NULL, get count
---              for all hypertables
+-- main_table - table or hypertable to get approximate row count for
 --
 -- Returns:
--- schema_name      - Schema name of the hypertable
--- table_name       - Table name of the hypertable
--- row_estimate     - Estimated number of rows according to catalog tables
-CREATE OR REPLACE FUNCTION hypertable_approximate_row_count(
-    main_table REGCLASS = NULL
-)
-    RETURNS TABLE (schema_name NAME,
-                   table_name NAME,
-                   row_estimate BIGINT
-                  ) LANGUAGE PLPGSQL VOLATILE
-    AS
+-- Estimated number of rows according to catalog tables
+CREATE OR REPLACE FUNCTION approximate_row_count(relation REGCLASS)
+RETURNS BIGINT
+LANGUAGE PLPGSQL VOLATILE STRICT AS
 $BODY$
-<<main>>
 DECLARE
-        table_name       NAME;
-        schema_name      NAME;
+	table_name       NAME;
+	schema_name      NAME;
+	row_count_parent BIGINT;
+	row_count        BIGINT;
 BEGIN
-        IF main_table IS NOT NULL THEN
-            SELECT relname, nspname
-            INTO STRICT table_name, schema_name
-            FROM pg_class c
-            INNER JOIN pg_namespace n ON (n.OID = c.relnamespace)
-            WHERE c.OID = main_table;
-        END IF;
+	SELECT relname, nspname, c.reltuples::bigint
+	INTO table_name, schema_name, row_count_parent
+	FROM pg_class c
+	INNER JOIN pg_namespace n ON (n.OID = c.relnamespace)
+	WHERE c.OID = relation;
 
--- Thanks to @fvannee on Github for providing the initial draft
--- of this query
-        RETURN QUERY
-        SELECT h.schema_name,
-            h.table_name,
-            row_estimate.row_estimate
-        FROM _timescaledb_catalog.hypertable h
-        CROSS JOIN LATERAL (
-            SELECT sum(cl.reltuples)::BIGINT AS row_estimate
-            FROM _timescaledb_catalog.chunk c
-            JOIN pg_class cl ON cl.relname = c.table_name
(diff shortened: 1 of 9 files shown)


test/expected/size_utils.out
```diff
@@ -153,6 +153,30 @@ SELECT * FROM chunks_detailed_size('toast_test');
...
+(1 row)
+
+ANALYZE approx_count;
+SELECT * FROM approximate_row_count('approx_count');
+ approximate_row_count 
+-----------------------
+                     7
+(1 row)
+
+DROP TABLE approx_count;
```
erimatnor, line 179: I guess you need not drop the table here, you can just do create_hypertable with migrate_data=>true

sql/size_utils.sql
```diff
@@ -367,52 +367,46 @@ $BODY$;
...
-        WHERE (main.table_name IS NULL OR h.table_name = main.table_name)
-        AND (main.schema_name IS NULL OR h.schema_name = main.schema_name)
-        ORDER BY h.schema_name, h.table_name;
+	SELECT relname, nspname, c.reltuples::BIGINT
+	INTO STRICT table_name, schema_name, row_count
+	FROM pg_class c
+	INNER JOIN pg_namespace n ON (n.OID = c.relnamespace)
+	WHERE c.OID = main_table;
+
+	SELECT row_estimate.row_estimate
```
erimatnor, line 391: Instead of splitting this into different queries for regular tables and hypertables, I think this should just use [pg_inherits](https://www.postgresql.org/docs/current/catalog-pg-inherits.html) to find the child tables. Although, admittedly, this might become complicated when there are multiple levels of inheritance (children-of-children, etc.)

erimatnor (commented):
While I generally approve of these changes, I think the query could be improved, although possibly for another PR (see inline comment). This also needs to fix the update script (old function needs to be removed).

test/expected/size_utils.out
pmwkaa: I did it because did not want to mess with saved statistics by previous ANALYZE call

test/expected/size_utils.out
```diff
@@ -153,6 +153,30 @@ SELECT * FROM chunks_detailed_size('toast_test');
...
+INSERT INTO approx_count VALUES('2004-01-01 10:00:06', 6);
+INSERT INTO approx_count VALUES('2004-01-01 10:00:07', 7);
+SELECT * FROM approximate_row_count('approx_count');
+ approximate_row_count 
+-----------------------
+                     0
+(1 row)
+
+ANALYZE approx_count;
+SELECT * FROM approximate_row_count('approx_count');
```
k-rus, line 173: Can you add test with `count` for comparison? I.e.,
```sql
SELECT count(8) FROM approx_count;
```
It is expected to give the same result.

k-rus (changes requested):
Since the function should work on PG tables, can you add tests with inheritance and with declarative partitioning?

test/expected/size_utils.out
pmwkaa: Sure, but note that it will not match in that case untill ANALYZE call

k-rus: Yes, I expect the result is the same after `ANALYZE` was run and no inserts after running it :). Thus my comment is after the line with `ANANLYZE`.

pmwkaa:
List of changes:

* Reworked query to support inherited tables with single logic for regular tables and hypertables
* Added more test cases

k-rus (changes requested):
Add test of table with declarative partitioning, https://www.postgresql.org/docs/12/ddl-partitioning.html

mfreed:
Can we clarify if this also fixes the issue with compression?  https://github.com/timescale/timescaledb/issues/1741

pmwkaa:
Added test for declarative partitioning with regular tables @k-rus

erimatnor (approved):

sql/updates/latest-dev.sql
```diff
@@ -10,6 +10,7 @@ DROP FUNCTION  indexes_relation_size;
 DROP FUNCTION _timescaledb_internal.partitioning_column_to_pretty;
 DROP FUNCTION _timescaledb_internal.range_value_to_pretty;
 -- end of do not reorder
+DROP FUNCTION hypertable_approximate_row_count;
```
svenklemm, line 13: need to use DROP FUNCTION IF EXISTS

sql/size_utils.sql
```diff
@@ -367,52 +367,48 @@ $BODY$;
...
-CREATE OR REPLACE FUNCTION hypertable_approximate_row_count(
-    main_table REGCLASS = NULL
-)
-    RETURNS TABLE (schema_name NAME,
-                   table_name NAME,
-                   row_estimate BIGINT
-                  ) LANGUAGE PLPGSQL VOLATILE
-    AS
+-- Estimated number of rows according to catalog tables
+CREATE OR REPLACE FUNCTION approximate_row_count(main_table REGCLASS)
```
svenklemm, line 374: the argument is not required to be the main table so should probably be named `relation`

svenklemm (approved):

k-rus (approved):
LGTM

PR #2263, opened 2020-08-26
Fix timestamp overflow in time_bucket optimization

### LINKED ISSUE #2127: Timestamp overflow in time_bucket optimization
This bug causes timestamp overflow, which means certain queries with restrictions (quals) that use timestamps close to the MAX allowed timestamp will fail.

During chunk expansion, quals are processed and certain optimizations applied to expression to make them more amenable to use by indexes on a hypertable. One of these optimizations involve `time_bucket` expressions where such an expression is replaced by as simple time column comparison instead:

```
 time_bucket(width, time) < time_value
/* Replace with: */
 time < time_value + width
```

However, this replacement is done without checking that the new timestamp value is within the allowable bounds of a timestamp (which is not the same as the integer bound). Further, the new time value is computed with simple integer arithmetic, so large widths could possibly lead to integer overflow as well.

The problem is further complicated by the fact that TimescaleDB internally uses UNIX epoch instead of PG epoch, which lowers the allowable MAX timestamp in order to facilitate conversion between the two epochs.

The fix for this bug is probably to use the built-in PG functions for adding a timestamp and an interval, or check for overflow manually. Extra checks might be needed to check for TimescaleDB-specific timestamp limits.

This bug prevents refreshing a continuous aggregate with a full refresh window.

erimatnor:
An optimization for `time_bucket` transforms expressions of the form
`time_bucket(10, time) < 100` to `time < 100 + 10` in order to do
chunk exclusion and make better use of indexes on the time
column. However, since one bucket is added to the timestamp when doing
this transformation, the timestamp can overflow.

While a check for such overflows already exists, it uses `+Infinity`
(INT64_MAX/DT_NOEND) as the upper bound instead of the actual end of
the valid timestamp range. A further complication arises because
TimescaleDB internally converts timestamps to UNIX epoch time, thus
losing a little bit of the valid timestamp range in the process. Dates
are further restricted by the fact that they are internally first
converted to timestamps (thus limited by the timestamp range) and then
converted to UNIX epoch.

This change fixes the overflow issue by only applying the
transformation if the resulting timestamps or dates stay within the
valid (TimescaleDB-specific) ranges.

A test has also been added to show the valid timestamp and date
ranges, both PostgreSQL and TimescaleDB-specific ones.

Fixes #2127


svenklemm (changes requested):
Please use an existing testcase like timestamp for this instead of adding a new test for just 2 sql commands

erimatnor:
> Please use an existing testcase like timestamp for this instead of adding a new test for just 2 sql commands

I did it this way as to avoid having to move the  timestamp test into debug builds only. If we are OK with moving basic timestamp tests into debug builds, then I can merge them.

pmwkaa (approved):
Looks good.

Just a thought: maybe place tests for this into more generic file like limits.sql instead?

erimatnor:
@svenklemm this is currently blocked by you. Is it OK for you to move all timestamp tests into DEBUG build in order to merge the two tests or is it OK to keep separate?

svenklemm:
Hmm i guess there is value in having those tests run for non-DEBUG as well

svenklemm (approved):

erimatnor:
@svenklemm if it's any consolation, the timestamp_limits test will probably go away once we no longer use UNIX time internally.

PR #2269, opened 2020-08-27
Handle AT_CookedColumnDefault in process_utility

svenklemm:
Postgres added a new alter table subcommand in a bugfix for
CREATE TABLE LIKE with inheritance which got backported to previous
versions. We check for presence of the subcommand with cmake so
we can support building on snapshot versions which also fixes the
ABI breakage test.

https://github.com/postgres/postgres/commit/50289819230d8ddad510879ee4793b04a05cf13b

### DIFF
--- a/src/build-defs.cmake
+++ b/src/build-defs.cmake
@@ -37,6 +37,15 @@ if(APACHE_ONLY)
   add_definitions(-DAPACHE_ONLY)
 endif()
 
+# AT_CookedColumnDefault was backported to PG12 but is not yet in a released version.
+# To be able to build against snapshots and REL_12_STABLE branch we check header file
+# for existance. Once PG 12.5 is out this can safely be turned into a version check.
+file(READ ${PG_INCLUDEDIR_SERVER}/nodes/parsenodes.h PG_PARSENODES_H)
+string(REGEX MATCH "AT_CookedColumnDefault" PG_HAS_COOKEDCOLUMNDEFAULT ${PG_CONFIG_H})
+if (PG_HAS_COOKEDCOLUMNDEFAULT)
+  add_definitions(-DPG_HAS_COOKEDCOLUMNDEFAULT)
+endif()
+
 include_directories(${PROJECT_SOURCE_DIR}/src ${PROJECT_BINARY_DIR}/src)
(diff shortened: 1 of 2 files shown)


src/build-defs.cmake
```diff
@@ -37,6 +37,12 @@ if(APACHE_ONLY)
   add_definitions(-DAPACHE_ONLY)
 endif()
 
+file(READ ${PG_INCLUDEDIR_SERVER}/nodes/parsenodes.h PG_PARSENODES_H)
+string(REGEX MATCH "AT_CookedColumnDefault" PG_HAS_COOKEDCOLUMNDEFAULT ${PG_CONFIG_H})
+if (PG_HAS_COOKEDCOLUMNDEFAULT)
+  add_definitions(-DPG_HAS_COOKEDCOLUMNDEFAULT)
```
erimatnor, line 43: A couple of questions/suggestions:

* Shouldn't this be tied to PostgreSQL version rather than parsing header?
* I think we should write this to `config.h` rather than relying on command-line/compile-time defines. It is better to be able to see the build configuration in a single file.

svenklemm: this code is not yet in a specific version thats why i check the header, but it's in the REL_12_STABLE branch and in snapshot builds which we use in our ABI tests, once 12.5 gets tagged we can turn this into a simple version check.

erimatnor: For the former (tying to PG version), I guess something like this would be sufficient in a header file or in `process_utility.c`

```
#if ((PG11 && PG_VERSION_NUM > XXXXXX) || (PG12 && PG_VERSION_NUM > YYYYYY) || PG12_GT)
#define PG_HAS_COOKEDCOLUMNDEFAULT 1
#endif
```

svenklemm: that code would not work, see my previous comment or commit message why i did it this way

erimatnor: Got it. I recommend adding a note/comment about this in the CMake file (and maybe the commit message) since it is non-obvious why it was done this way.

erimatnor (approved):

pmwkaa (approved):

PR #2280, opened 2020-08-29
Add TRUNCATE support for continuous aggregates

erimatnor:
This change adds the ability to truncate a continuous aggregate and
its source hypertable. Like other mutations (DELETE, UPDATE) on a
continuous aggregate or the underlying hypertable, an invalidation
needs to be added to ensure the aggregate can be refreshed again.

When a hypertable is truncated, an invalidation is added to the
hypertable invalidation log so that all its continuous aggregates will
be invalidated. When a specific continuous aggregate is truncated, the
invaliation is instead added to the continuous aggregate invalidation
log, so that only the one aggregate is invalidated.

### DIFF
--- a/tsl/src/continuous_aggs/invalidation.c
+++ b/tsl/src/continuous_aggs/invalidation.c
@@ -195,12 +195,95 @@ invalidation_cagg_log_add_entry(int32 cagg_hyper_id, int64 modtime, int64 start,
 	CatalogSecurityContext sec_ctx;
 	HeapTuple tuple;
 
+	Assert(start <= end);
 	tuple = create_invalidation_tup(RelationGetDescr(rel), cagg_hyper_id, modtime, start, end);
 	ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
 	ts_catalog_insert_only(rel, tuple);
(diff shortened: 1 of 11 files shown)


tsl/src/continuous_aggs/invalidation.c
```diff
@@ -203,6 +204,88 @@ invalidation_cagg_log_add_entry(int32 cagg_hyper_id, int64 modtime, int64 start,
...
+		}
+		case HypertableIsRawTable:
+			dim = hyperspace_get_open_dimension(ht->space, 0);
+			timeval = ts_get_now_internal(dim);
+			invalidation_hyper_log_add_entry(ht->fd.id, timeval, start, end);
+			break;
+		case HypertableIsMaterializationAndRaw:
+			break;
+		case HypertableIsNotContinuousAgg:
+			ereport(ERROR,
```
pmwkaa, line 280: Do we need a test for this case?

pmwkaa (approved):
Looks good overall

tsl/src/continuous_aggs/invalidation.c
erimatnor: This should not be possible. We could consider asserting, but figured it is safer to generate an error.

PR #2290, opened 2020-08-31
Fix issues relate to non-inclusive refresh window

erimatnor:
This change fixes some corner-case issues that could lead to a refresh
not actually refreshing when it should.

The issues arise because invalidations are inclusive in both ends
while the refresh window is exclusive in the end. In some case, this
wasn't correctly accounted for.

To fix the issues, the remainder after cutting invalidations has been
adjusted and we always add 1 to the end of the refresh window when it
is set from an invalidation. In addition, we add an extra check for
this case when computing the bucketed refresh window and the end of
the refresh window is at the start of the bucket (i.e., the window
size is 1).

The test suite has also been expanded to test for some of these corner
cases.

### DIFF
--- a/tsl/test/sql/continuous_aggs_invalidation.sql
+++ b/tsl/test/sql/continuous_aggs_invalidation.sql
@@ -51,7 +51,7 @@ CREATE MATERIALIZED VIEW cond_10
 WITH (timescaledb.continuous,
       timescaledb.materialized_only=true)
 AS
-SELECT time_bucket(BIGINT '10', time) AS day, device, avg(temp) AS avg_temp
+SELECT time_bucket(BIGINT '10', time) AS bucket, device, avg(temp) AS avg_temp
 FROM conditions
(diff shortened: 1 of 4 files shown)


pmwkaa (approved):

mkindahl (approved):

tsl/test/sql/continuous_aggs_invalidation.sql
```diff
@@ -302,3 +302,23 @@ SELECT materialization_id AS cagg_id,
        greatest_modified_value AS end
        FROM _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
        ORDER BY 1,2,3;
+
+-- Test invalidation of bucket start with minimal range. This
+-- generates and invalidation at the start of a bucket with minimal
```
WireBaron, line 307:
```suggestion
-- generates an invalidation at the start of a bucket with minimal
```

PR #2307, opened 2020-09-02
Overflow handling for refresh policy with integer based time

gayyappan:
When we compute the start and end intervals for the
refresh job execution, there is a potential to
run into overflow with integer computations. Handle 
overflow condition by setting the start/end boundaries to 
PG_INT64_MIN or PG_INT64_MAX.

### DIFF
--- a/tsl/src/bgw_policy/continuous_aggregate_api.c
+++ b/tsl/src/bgw_policy/continuous_aggregate_api.c
@@ -19,6 +19,7 @@
 #include "hypertable_cache.h"
 #include "time_utils.h"
 #include "policy_utils.h"
+#include "time_utils.h"
 
 #define POLICY_REFRESH_CAGG_PROC_NAME "policy_refresh_continuous_aggregate"
 #define CONFIG_KEY_MAT_HYPERTABLE_ID "mat_hypertable_id"
@@ -60,13 +61,12 @@ get_interval_from_config(const Dimension *dim, const Jsonb *config, const char *
 		if (!found)
 		{
 			*isnull = true;
-			return (Datum) 0;
+			return 0;
 		}
 		Oid now_func = ts_get_integer_now_func(dim);
 
 		Assert(now_func);
-
-		return subtract_integer_from_now(interval_val, partitioning_type, now_func);
+		return ts_subtract_integer_from_now_saturating(now_func, interval_val, partitioning_type);
 	}
 	else
 	{
@@ -83,15 +83,24 @@ get_interval_from_config(const Dimension *dim, const Jsonb *config, const char *
 }
 
 int64
-policy_refresh_cagg_get_refresh_start(const Dimension *dim, const Jsonb *config, bool *start_isnull)
+policy_refresh_cagg_get_refresh_start(const Dimension *dim, const Jsonb *config)
 {
-	return get_interval_from_config(dim, config, CONFIG_KEY_START_OFFSET, start_isnull);
+	bool start_isnull;
+	int64 res = get_interval_from_config(dim, config, CONFIG_KEY_START_OFFSET, &start_isnull);
+	/* interpret NULL as min value for that type */
+	if (start_isnull)
+		return ts_time_get_min(ts_dimension_get_partition_type(dim));
+	return res;
 }
 
 int64
-policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config, bool *end_isnull)
+policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config)
 {
-	return get_interval_from_config(dim, config, CONFIG_KEY_END_OFFSET, end_isnull);
+	bool end_isnull;
+	int64 res = get_interval_from_config(dim, config, CONFIG_KEY_END_OFFSET, &end_isnull);
+	if (end_isnull)
+		return ts_time_get_end_or_max(ts_dimension_get_partition_type(dim));
+	return res;
 }
 
 Datum
--- a/tsl/src/bgw_policy/policy_utils.h
+++ b/tsl/src/bgw_policy/policy_utils.h
@@ -11,5 +11,7 @@
 bool policy_config_check_hypertable_lag_equality(Jsonb *config, const char *json_label,
 												 Oid dim_type, Oid lag_type, Datum lag_datum);
 int64 subtract_integer_from_now(int64 interval, Oid time_dim_type, Oid now_func);
+int64 subtract_integer_from_now_internal(int64 interval, Oid time_dim_type, Oid now_func,
+										 bool *overflow);
 Datum subtract_interval_from_now(Interval *interval, Oid time_dim_type);
 #endif /* TIMESCALEDB_TSL_BGW_POLICY_UTILS_H */
--- a/tsl/test/expected/continuous_aggs_policy.out
+++ b/tsl/test/expected/continuous_aggs_policy.out
@@ -230,7 +230,7 @@ NOTICE:  adding not-null constraint to column "a"
  smallint_tab
 (1 row)
 
-CREATE OR REPLACE FUNCTION integer_now_smallint_tab() returns smallint LANGUAGE SQL STABLE as $$ SELECT 20::smallint $$;
+CREATE OR REPLACE FUNCTION integer_now_smallint_tab() returns smallint LANGUAGE SQL STABLE as $$ SELECT coalesce(max(a)::smallint, 0::smallint) FROM smallint_tab ; $$;
 SELECT set_integer_now_func('smallint_tab', 'integer_now_smallint_tab');
  set_integer_now_func 
 ----------------------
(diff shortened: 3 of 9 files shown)


tsl/src/bgw_policy/policy_utils.c
```diff
@@ -79,28 +79,46 @@ subtract_integer_from_now(int64 interval, Oid time_dim_type, Oid now_func)
 		case INT2OID:
 			res = DatumGetInt16(now) - interval;
 			if (res < PG_INT16_MIN || res > PG_INT16_MAX)
-				ereport(ERROR,
-						(errcode(ERRCODE_INTERVAL_FIELD_OVERFLOW), errmsg("ts_interval overflow")));
-			return Int16GetDatum(res);
+				*overflow = true;
```
svenklemm, line 82: probably makes sense to use pg_sub_s16_overflow and pg_sub_s32_overflow for int2/int4

gayyappan: Is there a reason why it wasn't implemented that way? Do you know? (This is refactored code from job.c)

svenklemm (changes requested):
I dont think this is  correct  as the range of valid values depends on the type of the dimension and you are always returning INT64_MIN/INT64_MAX

gayyappan:
> I dont think this is correct as the range of valid values depends on the type of the dimension and you are always returning INT64_MIN/INT64_MAX

I discussed this with @erimatnor . The range clamping is done by the refresh function:  It adjusts the range based on the data type of the dimension column. I don't see a reason to replicate that in the policy code as well.

erimatnor:
> > I dont think this is correct as the range of valid values depends on the type of the dimension and you are always returning INT64_MIN/INT64_MAX
> 
> I discussed this with @erimatnor . The range clamping is done by the refresh function: It adjusts the range based on the data type of the dimension column. I don't see a reason to replicate that in the policy code as well.

A recently merged PR has cleaned up a lot of things related to type-specific ranges. There are now utility functions to get the type-specific ranges, which are probably appropriate to use in this case.

tsl/src/bgw_policy/policy_utils.c
```diff
@@ -79,28 +79,46 @@ subtract_integer_from_now(int64 interval, Oid time_dim_type, Oid now_func)
...
+Datum
+subtract_integer_from_now(int64 interval, Oid time_dim_type, Oid now_func)
+{
+	AssertArg(IS_INTEGER_TYPE(time_dim_type));
+	bool overflow = false;
+	int64 res = subtract_integer_from_now_internal(interval, time_dim_type, now_func, &overflow);
+
+	if (overflow)
+	{
+		ereport(ERROR, (errcode(ERRCODE_INTERVAL_FIELD_OVERFLOW), errmsg("ts_interval overflow")));
```
erimatnor, line 109: I probably raised this before and don't recall the discussion. But do we really want to raise errors instead of just capping at boundaries? This would make the policy fail when it need not.

There are functions in `time_utils.c` now to handle saturating add and sub for time that clamp to valid time boundaries.

gayyappan: Yes, this PR needs some changes after the other utility functions were added. Not yet in.

gayyappan: I went back and looked at time_utils.c We have 2 versions of subtract_integer_from_now: this version throws the overflow error and used by many other policies. (Modifying other policies is outside the scope of this PR ). 

The internal version (called at line 105)  indicates the overflow and leaves it to the caller to handle as needed. This is the one used by refresh. This is needed for the refresh since the interval clamping is at the lower or upper bound  (i.e INT64_MAX or INT64_MIN ) based on whether this is the start/end interval of the refresh policy

tsl/src/bgw_policy/policy_utils.c
gayyappan: checked the pg_sub_s16_overflow and other variants. 
Makes this code a bit unwieldy to use the pg functions directly:  the result type  would  have to be int16, int32 , int64 for  sub_s16/sub_s32/sub_s64 respectively.  For example:
`       case INT2OID:
        {
            int16 res2 = 0;
            *overflow = pg_sub_s16_overflow(DatumGetInt16(now), interval, &res2);
            res = res2;
            break;
        }
`
I don't see a benefit in using pg_sub_sxx_overflow function for all the types since it is safe to do the subtraction using int64 types inline and just checking the range here.

tsl/test/expected/continuous_aggs_policy.out
```diff
@@ -219,13 +219,46 @@ INSERT INTO smallint_tab VALUES(5);
...
+SELECT hypertable_id as mat_id FROM _timescaledb_config.bgw_job where id = :job_id \gset
+\c :TEST_DBNAME :ROLE_SUPERUSER
+SELECT config FROM _timescaledb_config.bgw_job WHERE id = :job_id;
+                              config                               
+-------------------------------------------------------------------
+ {"end_interval": 0, "start_interval": 15, "mat_hypertable_id": 8}
+(1 row)
+
+UPDATE _timescaledb_config.bgw_job 
+SET config = jsonb_build_object('mat_hypertable_id', :mat_id,
```
erimatnor, line 241: Is this a valid policy? If time is 20 (according to the now function), then the refresh interval will be `[20-90000, 20-900000] => [ -89980, -899980 ]`. If I am not mistaken, this is  an invalid range because the start interval is smaller than the end interval (i.e., the start of the refresh is after the end). The values are also, obviously, out-of-range for smallint. Which of these error cases are you testing?

I am also wondering what the expected behavior for such a bad policy would be. I guess it shouldn't even be possible to add such a policy, although here you are updating metadata directly so some error/validation at runtime would be expected, right? 

However, this doesn't seem to be the case when the policy is run below. Instead, it seems to update the aggregate. Can you explain what is going on?

tsl/src/bgw_policy/policy_utils.c
gayyappan: @erimatnor  To clarify, The refresh policy *does not throw errors*. It clamps the range.
This function is ***not*** called by refresh policy.
This is a utility function invoked by other policies. This PR does not modify the behavior of other policies.

gayyappan: refresh policy calls subtract_integer_from_now_internal. This function does not throw an error.

tsl/src/bgw_policy/policy_utils.h
```diff
@@ -11,5 +11,7 @@
 bool policy_config_check_hypertable_lag_equality(Jsonb *config, const char *json_label,
 												 Oid dim_type, Oid lag_type, Datum lag_datum);
 Datum subtract_integer_from_now(int64 interval, Oid time_dim_type, Oid now_func);
+int64 subtract_integer_from_now_internal(int64 interval, Oid time_dim_type, Oid now_func,
+										 bool *overflow);
 Datum subtract_interval_from_now(Interval *interval, Oid time_dim_type);
```
gayyappan, line 16: time_utils.c now defines the same function (new changes in master). I will remove subtract_interval_from_now from this file. working on the change.

tsl/test/expected/continuous_aggs_policy.out
gayyappan: updated tests and modified  code based on our discussion

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -81,15 +90,23 @@ get_interval_from_config(const Dimension *dim, const Jsonb *config, const char *
...
 
 int64
-policy_refresh_cagg_get_refresh_start(const Dimension *dim, const Jsonb *config, bool *start_isnull)
+policy_refresh_cagg_get_refresh_start(const Dimension *dim, const Jsonb *config)
 {
-	return get_interval_from_config(dim, config, CONFIG_KEY_START_INTERVAL, start_isnull);
+	bool start_isnull;
+	int64 res = get_interval_from_config(dim, config, CONFIG_KEY_START_INTERVAL, &start_isnull);
+	if (start_isnull)
+		return PG_INT64_MIN;
```
svenklemm, line 98: i think you should get the boundaries from the added utility functions instead of hardcoding

gayyappan: done.

svenklemm (approved):

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -81,15 +90,24 @@ get_interval_from_config(const Dimension *dim, const Jsonb *config, const char *
...
 
 int64
-policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config, bool *end_isnull)
+policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config)
 {
-	return get_interval_from_config(dim, config, CONFIG_KEY_END_INTERVAL, end_isnull);
+	bool end_isnull;
+	int64 res = get_interval_from_config(dim, config, CONFIG_KEY_END_INTERVAL, &end_isnull);
+	if (end_isnull)
+		return ts_time_get_max(ts_dimension_get_partition_type(dim));
```
erimatnor, line 109: I think we need to revisit the "defaults" when we input NULL values. It makes more sense to me that NULL end_interval means "current time" (i.e., no offset) instead of infinite refresh.

I we keep this interpretation o NULL, we should return "end" if possible, otherwise we won't refresh the whole time range for timestamp types (given that the refresh window is exclusive). Also, would be good to get this line covered by a test.

```suggestion
		return ts_time_get_end_or_max(ts_dimension_get_partition_type(dim));
```

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -53,18 +54,26 @@ get_interval_from_config(const Dimension *dim, const Jsonb *config, const char *
...
 		Assert(now_func);
 
-		return subtract_integer_from_now(interval_val, partitioning_type, now_func);
+		res = subtract_integer_from_now_internal(interval_val,
+												 partitioning_type,
+												 now_func,
+												 &overflow);
+		/* overflowed, so return max for that data type */
+		if (overflow)
+			res = ts_time_get_max(partitioning_type);
```
erimatnor, line 75: If we are doing subtraction, shouldn't we return min instead of max? I guess the interval could be negative, in which case the subtraction effectively becomes an addition. It seems we need to check for overflow in each end of the valid time range.

Let's say you have a smallint table and do this:

```
SELECT add_continuous_aggregate_policy('cagg', 50000, -50000, '1 hour');
```

This should basically be turned into a refresh of [MIN, MAX], but, if I am not mistaken, this will become [ MAX, MAX] with the current code.

tsl/src/bgw_policy/policy_utils.c
```diff
@@ -64,43 +64,61 @@ policy_config_check_hypertable_lag_equality(Jsonb *config, const char *json_labe
...
-			return Int16GetDatum(res);
+				*overflow = true;
+			break;
 		case INT4OID:
 			res = DatumGetInt32(now) - interval;
 			if (res < PG_INT32_MIN || res > PG_INT32_MAX)
-				ereport(ERROR,
-						(errcode(ERRCODE_INTERVAL_FIELD_OVERFLOW), errmsg("ts_interval overflow")));
-			return Int32GetDatum(res);
+				*overflow = true;
```
erimatnor, line 87: Would be good to hit this in a test. Perhaps a C-based unit test is best to test this corner-cases?

erimatnor (approved):
I am approving, but with caveats and with an assumptions that the issues below will be addressed.

* It would be good to improve test coverage. At least two of the important corner cases are never hit in a test. 
* I am also suspicious of the overflow handling in `get_interval_from_config` since it always returns max in case of overflow. But it seems to me that an interval can overflow `now()` in either direction depending on whether the interval is positive or negative. 
* It is also odd that that `policy_refresh_cagg_get_refresh_start` would return max time in case of overflow while returning min time in case of NULL.
* I guess we decided to allow negative intervals, but see no tests for such cases.

tsl/src/bgw_policy/continuous_aggregate_api.c
gayyappan: I'll change this to use ts_time_get_end_or_max. We use start_interval=NULL to represent -infinity. So makes sense to use that for end_interval as well to represent +infinity, instead of arbitrarily clamping it to current time. (This means the window could include data from the "future").  Wouldn't this translate to the max time value in the data anyway since there is nothing else to process?

PR #2325, opened 2020-09-05
Fix rename constraint/rename index

### LINKED ISSUE #2101: [Bug] could not find arbiter index for hypertable
 - OS: Ubuntu 18.04.4
 - PostgreSQL version (output of `postgres --version`): 11.8
 - TimescaleDB version (output of `\dx` in `psql`): 1.7.2

**Describe the bug**
Rename a unique constraint on an hypertable leads to upsert operation failing

**To Reproduce**
```
CREATE EXTENSION IF NOT EXISTS timescaledb CASCADE;

CREATE TABLE t_hypertable (
  id_extern INTEGER NOT NULL,
  time      TIMESTAMPTZ NOT NULL,
  value     FLOAT NOT NULL CHECK (value > 0),
  UNIQUE(id_extern, time)
);

SELECT create_hypertable('t_hypertable', 
                         'time', 
                         'id_extern',
                         7,
                         chunk_time_interval => interval '2 hours');

INSERT INTO t_hypertable AS h
VALUES (
  1,
  '2020-01-01 00:00:00',
  3.2
)
ON CONFLICT (id_extern, time)
  DO UPDATE SET
    value = h.value + EXCLUDED.value;

ALTER TABLE t_hypertable RENAME CONSTRAINT t_hypertable_id_extern_time_key TO t_new_constraint;
```
Now a further upsert on the same chunk doesn't work anymore:

```
INSERT INTO t_hypertable AS h
VALUES (
  1,
  '2020-01-01 00:01:00',
  3.2
)
ON CONFLICT (id_extern, time)
  DO UPDATE SET
    value = h.value + EXCLUDED.value;
```

```
ERROR:  could not find arbiter index for hypertable index "t_new_constraint" on chunk "_hyper_1_1_chunk"
```

Important fact: upsert on new chunks works, upserts are not working only on old preexisting chunks at the moment of contraint renamed

Given the fact this is a index issue I did a few operation in the attempt to fix it

 - move_chunk didn't work
 - reorder_chunk didn't work
 - reindex table/chunk didn't work
 - rename constraint on chunk is not supported

The issue can be "fixed" this way:

```
// Unfortunately this can not be performed concurrently
CREATE UNIQUE INDEX idx_tmp ON t_hypertable(id_extern, time);                 

// This fixes the issue
ALTER TABLE t_hypertable DROP CONSTRAINT t_new_constraint;    
```

at this point the upsert works again, I tried to add back the constraint:

```
// not supported
ALTER TABLE t_hypertable ADD CONSTRAINT  new_constraint UNIQUE USING INDEX idx_tmp; 
```

As you can understand the fix as it is can not be applied on a live database, actually I tried with
```
CREATE INDEX ... WITH (timescaledb.transaction_per_chunk, ...);
```
but this can not be used for unique indexes :(

and the following is not supported eiter
```
ALTER TABLE t_hypertable ADD CONSTRAINT  new_constraint UNIQUE USING INDEX idx_tmp; 
```

svenklemm:
When a constraint is backed by an index like a unique constraint
or a primary key constraint the constraint can be renamed by either
ALTER TABLE RENAME CONSTRAINT or by ALTER INDEX RENAME. Depending
on the command used to rename different internal metadata tables
would be adjusted leading to corrupt metadata. This patch makes
ALTER TABLE RENAME CONSTRAINT and ALTER INDEX RENAME adjust the
same metadata tables.

Fixes #2101

### DIFF
--- a/src/chunk_constraint.c
+++ b/src/chunk_constraint.c
@@ -853,11 +853,63 @@ chunk_constraint_rename_hypertable_from_tuple(TupleInfo *ti, const char *newname
 										   NameStr(new_chunk_constraint_name));
 
 	new_tuple = heap_modify_tuple(tuple, tupdesc, values, nulls, repl);
+
+	ts_chunk_index_adjust_meta(chunk_id,
+							   newname,
+							   NameStr(*old_chunk_constraint_name),
+							   NameStr(new_chunk_constraint_name));
+
 	ts_catalog_update(ti->scanrel, new_tuple);
 	heap_freetuple(new_tuple);
 
 	if (should_free)
 		heap_freetuple(tuple);
+}
+
+/*
+ * Adjust internal metadata after index/constraint rename
+ */
+int
+ts_chunk_constraint_adjust_meta(int32 chunk_id, const char *ht_constraint_name, const char *oldname,
+								const char *newname)
+{
+	ScanIterator iterator =
+		ts_scan_iterator_create(CHUNK_CONSTRAINT, RowExclusiveLock, CurrentMemoryContext);
+	int count = 0;
+
+	init_scan_by_chunk_id_constraint_name(&iterator, chunk_id, oldname);
+
+	ts_scanner_foreach(&iterator)
+	{
+		bool nulls[Natts_chunk_constraint];
+		bool repl[Natts_chunk_constraint] = { false };
+		Datum values[Natts_chunk_constraint];
+		bool should_free;
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		HeapTuple tuple = ts_scanner_fetch_heap_tuple(ti, false, &should_free);
+		HeapTuple new_tuple;
+
+		heap_deform_tuple(tuple, ts_scanner_get_tupledesc(ti), values, nulls);
+
+		values[AttrNumberGetAttrOffset(Anum_chunk_constraint_hypertable_constraint_name)] =
+			CStringGetDatum(ht_constraint_name);
+		repl[AttrNumberGetAttrOffset(Anum_chunk_constraint_hypertable_constraint_name)] = true;
+		values[AttrNumberGetAttrOffset(Anum_chunk_constraint_constraint_name)] =
+			CStringGetDatum(newname);
+		repl[AttrNumberGetAttrOffset(Anum_chunk_constraint_constraint_name)] = true;
+
+		new_tuple = heap_modify_tuple(tuple, ts_scanner_get_tupledesc(ti), values, nulls, repl);
+
+		ts_catalog_update(ti->scanrel, new_tuple);
+		heap_freetuple(new_tuple);
+
+		if (should_free)
+			heap_freetuple(tuple);
+
+		count++;
+	}
+
+	return count;
 }
 
 int
--- a/src/chunk_index.c
+++ b/src/chunk_index.c
@@ -37,6 +37,7 @@
 #include "hypertable_cache.h"
 #include "catalog.h"
 #include "scanner.h"
+#include "scan_iterator.h"
 #include "chunk.h"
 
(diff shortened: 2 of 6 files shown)


mkindahl (approved):

src/chunk_index.c
```diff
@@ -907,6 +912,59 @@ chunk_index_tuple_rename(TupleInfo *ti, void *data)
...
+	heap_freetuple(new_tuple);
+
+	return SCAN_DONE;
+}
+
+/*
+ * Adjust internal metadata after index/constraint rename
+ */
+int
+ts_chunk_index_adjust_meta(int32 chunk_id, const char *ht_index_name, const char *oldname,
```
erimatnor, line 940: What is the difference between "adjust" and "rename"; especially given that this code is near identical to `ts_chunk_index_rename`?

I don't understand why we need both "adjust" and "rename", and at the very least they should share most of the code.

src/chunk_index.c
```diff
@@ -907,6 +912,59 @@ chunk_index_tuple_rename(TupleInfo *ti, void *data)
 	return SCAN_DONE;
 }
 
+static ScanTupleResult
+chunk_index_tuple_adjust_meta(TupleInfo *ti, void *data)
```
erimatnor, line 916: This function is very similar to `chunk_index_tuple_rename`. It is unclear why we need both. At the very least they should be able to share most code. See related comment below.

src/chunk_constraint.c
```diff
@@ -853,13 +853,65 @@ chunk_constraint_rename_hypertable_from_tuple(TupleInfo *ti, const char *newname
...
 
 	if (should_free)
 		heap_freetuple(tuple);
 }
 
+/*
+ * Adjust internal metadata after index/constraint rename
+ */
+int
+ts_chunk_constraint_adjust_meta(int32 chunk_id, const char *ht_constraint_name, const char *oldname,
```
erimatnor, line 873: There's already a function `ts_chunk_constraint_rename_hypertable_constraint`. Why do we need also this function and what is the difference? It seems that there's a lot of code and functionality overlap since this function is doing mostly the same thing.

erimatnor (commented):
These new "adjust" function have significant functionality overlap with existing renaming functionality for constraints and indexes. It is unclear why we need both and what the difference is.

src/chunk_index.c
svenklemm: adjust updates metadata only while rename has the ddl commands to actually rename the objects

src/chunk_constraint.c
svenklemm: The current code mixes DDL operations with metadata adjustments which doesnt work when multiple adjustments need to happen as those DDL operations are not idempotent. While it makes sense to refactor the code to separate DDL from metadata adjustment thats not the focus of this PR

src/chunk_index.c
erimatnor: There's still the issue of code duplication. If I am not mistaken, this function is literally identical to `ts_chunk_index_rename` save for two lines of code. I think it makes sense to reuse the existing code and just call another tuple handler, or have extra options in `ChunkIndexRenameInfo` to say, e.g., "do not rename objects".

src/chunk_constraint.c
erimatnor: You can, for instance, just add a boolean parameter to the existing function  "apply_ddl=false` that doesn't do DDL changes when set to false and you can reuse the existing code.

svenklemm: i think your suggesting changes actually decrease code quality, i initially had it all in a single function and decided against it. Also this is a bugfix PR that needs backporting into 1.7.x branch so it is beneficial to keep refactoring separate.

pmwkaa (approved):
Looks good but I agree with @erimatnor here, it would be nice to reduce number of functions doing similar changes if possible

PR #2335, opened 2020-09-07
Fix segfaults in tests in 32bit environment

svenklemm:


erimatnor (approved):
We really need those 32-bit tests on PRs. Otherwise this will keep happening because it is so easy to make a mistake.

k-rus (approved):

PR #2338, opened 2020-09-08
Cap invalidation threshold at last data bucket

### LINKED ISSUE #2333: Do not move invalidation threshold further than necessary
The invalidation threshold exists to reduce write amplification when invalidations are written on a hypertable. Invalidations that fall beyond the threshold (i.e., they are more recent than the threshold) are never written since everything beyond the threshold is assumed invalid anyway. 

Currently, the invalidation threshold is set by the refresh window, which means also that if the window is, eg., `-Infinity,+Infinity`, the invalidation threshold will move to the end of time. This means that the threshold no longer prohibits write amplification.

A more efficient way of dealing with this is to never increase the threshold beyond the max time value of the current data in the hypertable. Although that will not prohibit write amplification immediately after the refresh, once data fills beyond the threshold it will.

This issue is related to #2191

erimatnor:
When refreshing with an "infinite" refresh window going forward in
time, the invalidation threshold is also moved forward to the end of
the valid time range. This effectively renders the invalidation
threshold useless, leading to unnecessary write amplification.

To handle infinite refreshes better, this change caps the refresh
window at the end of the last bucket of data in the underlying
hypertable, as to not move the invalidation threshold further than
necessary. For instance, if the max time value in the hypertable is
11, a refresh command such as:

```
CALL refresh_continuous_aggregate(NULL, NULL);
```
would be turned into
```
CALL refresh_continuous_aggregate(NULL, 20);
```

assuming that a bucket starts at 10 and ends at 20 (exclusive). Thus
the invalidation threshold would at most move to 20, allowing the
threshold to still do its work once time again moves forward and
beyond it.

Note that one must never process invalidations beyond the invalidation
threshold without also moving it, as that would clear that area from
invalidations and thus prohibit refreshing that region once the
invalidation threshold is moved forward. Therefore, if we do not move
the threshold further than a certain point, we cannot refresh beyond
it either. An alternative, and perhaps safer, approach would be to
always invalidate the region over which the invalidation threshold is
moved (i.e., new_threshold - old_threshold). However, that is left for
a future change.

It would be possible to also cap non-infinite refreshes, e.g.,
refreshes that end at a higher time value than the max time value in
the hypertable. However, when an explicit end is specified, it might
be on purpose so optimizing this case is also left for the future.

Closes #2333

### DIFF
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -19,6 +19,7 @@
 #include <commands/tablecmds.h>
 #include <commands/tablespace.h>
 #include <commands/trigger.h>
+#include <executor/spi.h>
 #include <funcapi.h>
 #include <miscadmin.h>
 #include <nodes/makefuncs.h>
@@ -2680,4 +2681,52 @@ ts_hypertable_func_call_on_data_nodes(Hypertable *ht, FunctionCallInfo fcinfo)
 {
 	if (hypertable_is_distributed(ht))
 		ts_cm_functions->func_call_on_data_nodes(fcinfo, ts_hypertable_get_data_node_name_list(ht));
+}
+
+/*
+ * Get the max value of an open dimension.
+ */
+Datum
+ts_hypertable_get_open_dim_max_value(const Hypertable *ht, int dimension_index, bool *isnull)
+{
+	StringInfo command;
+	Dimension *dim;
+	int res;
+	bool max_isnull;
+	Datum maxdat;
+
+	dim = hyperspace_get_open_dimension(ht->space, dimension_index);
+
+	if (NULL == dim)
+		elog(ERROR, "invalid open dimension index %d", dimension_index);
+
+	/* Query for the last bucket in the materialized hypertable */
+	command = makeStringInfo();
+	appendStringInfo(command,
+					 "SELECT max(%s) FROM %s.%s",
+					 quote_identifier(NameStr(dim->fd.column_name)),
+					 quote_identifier(NameStr(ht->fd.schema_name)),
+					 quote_identifier(NameStr(ht->fd.table_name)));
+
+	if (SPI_connect() != SPI_OK_CONNECT)
+		elog(ERROR, "could not connect to SPI");
+
+	res = SPI_execute(command->data, true /* read_only */, 0 /*count*/);
+
+	if (res < 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 (errmsg("could not find the maximum time value for hypertable \"%s\"",
+						 get_rel_name(ht->main_table_relid)))));
+
+	Assert(SPI_gettypeid(SPI_tuptable->tupdesc, 1) == ts_dimension_get_partition_type(dim));
+	maxdat = SPI_getbinval(SPI_tuptable->vals[0], SPI_tuptable->tupdesc, 1, &max_isnull);
+
+	if (isnull)
+		*isnull = max_isnull;
+
+	res = SPI_finish();
+	Assert(res == SPI_OK_FINISH);
+
+	return maxdat;
 }
--- a/tsl/src/continuous_aggs/invalidation_threshold.c
+++ b/tsl/src/continuous_aggs/invalidation_threshold.c
@@ -10,6 +10,7 @@
 #include <nodes/memnodes.h>
 #include <storage/lockdefs.h>
 #include <storage/lmgr.h>
+#include <utils/builtins.h>
 #include <utils/memutils.h>
 #include <utils/snapmgr.h>
 
@@ -17,6 +18,8 @@
 #include <scanner.h>
 #include <scan_iterator.h>
 #include <compat.h>
+#include <time_utils.h>
+#include <time_bucket.h>
 
 #include "continuous_agg.h"
 #include "continuous_aggs/materialize.h"
@@ -84,11 +87,12 @@ scan_update_invalidation_threshold(TupleInfo *ti, void *data)
 	else
 	{
 		elog(DEBUG1,
-			 "hypertable %d existing  watermark >= new invalidation threshold " INT64_FORMAT
+			 "hypertable %d existing watermark >= new invalidation threshold " INT64_FORMAT
 			 " " INT64_FORMAT,
 			 form->hypertable_id,
 			 form->watermark,
 			 invthresh->threshold);
+		invthresh->threshold = form->watermark;
 	}
 
 	if (should_free)
@@ -97,12 +101,17 @@ scan_update_invalidation_threshold(TupleInfo *ti, void *data)
 	return SCAN_DONE;
 }
 
-/* every cont. agg calculates its invalidation_threshold point based on its
- *refresh_lag etc. We update the raw hypertable's invalidation threshold
- * only if this new value is greater than the existsing one.
+/*
+ * Set a new invalidation threshold.
+ *
+ * The threshold is only updated if the new threshold is greater than the old
+ * one.
+ *
+ * On success, the new threshold is returned, otherwise the existing threshold
+ * is returned instead.
  */
-bool
-continuous_agg_invalidation_threshold_set(int32 raw_hypertable_id, int64 invalidation_threshold)
+int64
+invalidation_threshold_set_or_get(int32 raw_hypertable_id, int64 invalidation_threshold)
 {
 	bool threshold_found;
 	InvalidationThresholdData data = {
@@ -156,10 +165,9 @@ continuous_agg_invalidation_threshold_set(int32 raw_hypertable_id, int64 invalid
 
 		ts_catalog_insert_values(rel, desc, values, nulls);
 		table_close(rel, NoLock);
-		data.was_updated = true;
 	}
 
-	return data.was_updated;
+	return data.threshold;
 }
 
 static ScanTupleResult
@@ -177,7 +185,7 @@ invalidation_threshold_tuple_found(TupleInfo *ti, void *data)
 }
 
 int64
-continuous_agg_invalidation_threshold_get(int32 hypertable_id)
+invalidation_threshold_get(int32 hypertable_id)
 {
 	int64 threshold = 0;
 	ScanKeyData scankey[1];
@@ -220,7 +228,7 @@ invalidation_threshold_htid_found(TupleInfo *tinfo, void *data)
  * block till lock is acquired.
  */
 void
-continuous_agg_invalidation_threshold_lock(int32 raw_hypertable_id)
+invalidation_threshold_lock(int32 raw_hypertable_id)
 {
 	ScanTupLock scantuplock = {
 		.waitpolicy = LockWaitBlock,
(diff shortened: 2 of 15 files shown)


src/hypertable.c
```diff
@@ -2681,3 +2682,54 @@ ts_hypertable_func_call_on_data_nodes(Hypertable *ht, FunctionCallInfo fcinfo)
...
+	res = SPI_execute_with_args(command->data,
+								0 /*=nargs*/,
+								NULL,
+								NULL,
+								NULL /*=Nulls*/,
+								true /*=read_only*/,
+								0 /*count*/);
+	if (res < 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
```
pmwkaa, line 2720: I wonder, if this case is possible and should we test it in that case

pmwkaa (approved):
Looks good

k-rus:
@erimatnor Is it applicable only in the case of refresh policy or with manual refresh too? From the commit message it sounds that both ways are covered.

erimatnor:
> @erimatnor Is it applicable only in the case of refresh policy or with manual refresh too? From the commit message it sounds that both ways are covered.

Both are covered. The policy uses the same "manual" refresh API underneath the hood.

tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -259,3 +267,42 @@ continuous_agg_invalidation_threshold_lock(int32 raw_hypertable_id)
 				 errmsg("found multiple invalidation rows for hypertable %d", raw_hypertable_id)));
 	}
 }
+
+/*
+ * Compute a new invalidation threshold.
+ *
+ * The new invalidation threshold returned is the end of the given refresh
+ * window, unless it ends at "infinity" in which case the threshold is capped
+ * at the end of the last bucket materialized.
```
WireBaron, line 276: Looks like this will return infinity when that's the end of the refresh window and there's no materialized data yet.  Is this correct?  Should you mention this case in the comment here?

WireBaron (approved):
Just some minor questions/comments.

src/hypertable.c
```diff
@@ -2681,3 +2682,54 @@ ts_hypertable_func_call_on_data_nodes(Hypertable *ht, FunctionCallInfo fcinfo)
...
+	appendStringInfo(command,
+					 "SELECT max(%s) FROM %s.%s",
+					 quote_identifier(NameStr(dim->fd.column_name)),
+					 quote_identifier(NameStr(ht->fd.schema_name)),
+					 quote_identifier(NameStr(ht->fd.table_name)));
+
+	if (SPI_connect() != SPI_OK_CONNECT)
+		elog(ERROR, "could not connect to SPI");
+
+	res = SPI_execute_with_args(command->data,
```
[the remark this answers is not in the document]
erimatnor: Yes, that's a good question. This was code that existed previously, which I moved here for code-reuse purposes. I will see if a simple SPI_execute works (it should).

src/hypertable.c
erimatnor: I don't know a good way to make SPI fail. Maybe inject a bad query; but it seems odd to deliberately do that. I think we have to trust that SPI works the way it should.

k-rus:
@erimatnor A suggestion to improve the commit message, the second paragraph:

> To handle infinite refreshes better, this change caps the invalidation
> forward to the end of the last
> bucket of data in the underlying hypertable. Subsequently the refresh window 
> is caped from +infinity to the end of the last
> bucket of data in the underlying hypertable. Caping the refresh window is necessary,
since  one cannot
> refresh and process invalidations beyond the invalidation threshold,
> as that would clear that area from invalidations and thus prohibit
> refreshing that region once the invalidation threshold is moved
> forward. An alternative, and perhaps safer, approach would be to
> always invalidate the region over which the invalidation threshold is
> moved (i.e., new_threshold - old_threshold). However, that is left for
> a future change.

tsl/src/continuous_aggs/invalidation_threshold.c
erimatnor: Good catch. Not exactly intended, not wrong either, but certainly non-optimal. I fixed this so that it returns the min time value in that case and we avoid moving the threshold unnecessarily. I added extra test cases to cover this corner-case.

src/hypertable.c
```diff
@@ -2681,3 +2682,54 @@ ts_hypertable_func_call_on_data_nodes(Hypertable *ht, FunctionCallInfo fcinfo)
...
+Datum
+ts_hypertable_get_open_dim_max_value(const Hypertable *ht, int dimension_index, bool *isnull)
+{
+	StringInfo command;
+	Dimension *dim;
+	int res;
+	bool max_isnull;
+	Datum maxdat;
+
+	dim = hyperspace_get_open_dimension(ht->space, dimension_index);
```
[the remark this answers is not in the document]
erimatnor: Check added.

erimatnor:
> @erimatnor A suggestion to improve the commit message, the second paragraph:
> 
> > To handle infinite refreshes better, this change caps the invalidation
> > forward to the end of the last
> > bucket of data in the underlying hypertable. Subsequently the refresh window
> > is caped backward from +infinity to the end of the last
> > bucket of data in the underlying hypertable. Caping the refresh window is necessary,
> > since  one cannot
> > refresh and process invalidations beyond the invalidation threshold,
> > as that would clear that area from invalidations and thus prohibit
> > refreshing that region once the invalidation threshold is moved
> > forward. An alternative, and perhaps safer, approach would be to
> > always invalidate the region over which the invalidation threshold is
> > moved (i.e., new_threshold - old_threshold). However, that is left for
> > a future change.

I changed to commit message, hopefully to your liking. I didn't really understand your suggested change, however, in particular "Subsequently the refresh window is caped backward from +infinity to the end of the last bucket...".

tsl/src/continuous_aggs/refresh.c
```diff
@@ -342,7 +357,37 @@ continuous_agg_refresh_internal(const ContinuousAgg *cagg,
...
+	 * existing threshold is returned. */
+	invalidation_threshold = invalidation_threshold_set_or_get(cagg->data.raw_hypertable_id,
+															   computed_invalidation_threshold);
+
+	/* We must also cap the refresh window at the invalidation threshold. If
+	 * we process invalidations after the threshold, the continuous aggregates
+	 * won't be refreshed when the threshold is moved forward in the
+	 * future. The invalidation threshold should already be aligned on bucket
+	 * boundary. */
+	if (invalidation_threshold <= refresh_window.end)
```
k-rus, line 377:
```suggestion
	if (refresh_window.end => invalidation_threshold)
```
then it is the same order as the next line and thus easier to follow.

k-rus, line 377: Why should it be inclusive?:
```suggestion
	if (invalidation_threshold < refresh_window.end)
```

tsl/src/continuous_aggs/refresh.c
```diff
@@ -342,7 +357,37 @@ continuous_agg_refresh_internal(const ContinuousAgg *cagg,
...
+	{
+		refresh_window.end = invalidation_threshold;
+
+		/* Capping the end might have made the window 0, or negative, so
+		 * nothing to refresh in that case */
+		if (refresh_window.start >= refresh_window.end)
+		{
+			emit_up_to_date_notice(cagg);
+			return;
+		}
```
k-rus, line 387: Will it be useful to check it anyway, i.e., outside the `if` statement?

k-rus (approved):
LGTM
Few nits.

k-rus:
> I changed to commit message, hopefully to your liking. I didn't really understand your suggested change, however, in particular "Subsequently the refresh window is caped backward from +infinity to the end of the last bucket...".

The new commit message doesn't improve the most confusing sentence. It says:
> this change caps the invalidation
threshold, and subsequently the refresh window, at the end of the last
bucket of data in the underlying hypertable.

The refreshed window is capped from `+infinity`, which is the request to user. While the invalidation threshold cannot be capped, but should be increased forward. The confusion comes from that the candidate invalidation threshold is capped, not the invalidation threshold itself. Will it be better?:

> this change caps the refresh window at the of the last bucket of data in underlying hypertable, and the invalidation threshold is move forward to the capped value.

tsl/src/continuous_aggs/refresh.c
erimatnor: It doesn't matter since the end will be at the invalidation threshold in either case. But I'll make it exclusive if that makes it more clear.

tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -97,12 +101,17 @@ scan_update_invalidation_threshold(TupleInfo *ti, void *data)
...
+ * The threshold is only updated if the new threshold is greater than the old
+ * one.
+ *
+ * On success, the new threshold is returned, otherwise the existing threshold
+ * is returned instead.
  */
-bool
-continuous_agg_invalidation_threshold_set(int32 raw_hypertable_id, int64 invalidation_threshold)
+int64
+invalidation_threshold_set_or_get(int32 raw_hypertable_id, int64 invalidation_threshold)
```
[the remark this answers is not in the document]
erimatnor: I wanted to point out that the existing threshold is returned, which is not clear if it is just called update. Update, like set, implies it is always updated, but it is only set if the new value is greater than the old one.

tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -259,3 +267,47 @@ continuous_agg_invalidation_threshold_lock(int32 raw_hypertable_id)
...
+invalidation_threshold_compute(const ContinuousAgg *cagg, const InternalTimeRange *refresh_window)
+{
+	bool max_refresh = false;
+	Hypertable *ht = ts_hypertable_get_by_id(cagg->data.raw_hypertable_id);
+
+	if (TS_TIME_IS_INTEGER_TIME(refresh_window->type))
+		max_refresh = TS_TIME_IS_MAX(refresh_window->end, refresh_window->type);
+	else
+		max_refresh = TS_TIME_IS_END(refresh_window->end, refresh_window->type) ||
+					  TS_TIME_IS_NOEND(refresh_window->end, refresh_window->type);
```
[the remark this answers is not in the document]
erimatnor: I think it makes sense to do that when, and if, the need arises.

src/hypertable.c
```diff
@@ -2681,3 +2682,52 @@ ts_hypertable_func_call_on_data_nodes(Hypertable *ht, FunctionCallInfo fcinfo)
...
+	if (SPI_connect() != SPI_OK_CONNECT)
+		elog(ERROR, "could not connect to SPI");
+
+	res = SPI_execute(command->data, true /* read_only */, 0 /*count*/);
+
+	if (res < 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 (errmsg("could not find the maximum time value for hypertable"),
+				  errdetail("SPI error when calculating continuous aggregate watermark: %d.",
```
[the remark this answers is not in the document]
erimatnor: Forgot to update this message when I refactored. I actually think it makes sense to remove the detail message now. This error shouldn't really happen unless the query is corrupted or something else really weird happens.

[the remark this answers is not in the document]
k-rus: I am totally fine with removing the detailed message.

tsl/src/continuous_aggs/refresh.c
erimatnor: Sorry, what is it that you would like to check outside the if statement?

k-rus: Can `refresh_window.start >= refresh_window.end` happen if the window is not capped? In particular `refresh_window.start == refresh_window.end`

erimatnor: Ohh, I think you want to check that start is not after end outside the larger if clause, or?

This is only an issue after capping since there is already a check above for the window given as arg to the function.

k-rus: I see. Since I suggested to use strict comparison between the window end and capped invalidation threshold, I wasn't sure if this nested if needs to be moved out or not.

PR #2362, opened 2020-09-10
Update policy_stats view with additional fields

gayyappan:
Update policy_stats view to make it more end user
friendly. These changes mirror the fields in
continuous_aggregate_stats.

### DIFF
--- a/tsl/test/expected/bgw_reorder_drop_chunks.out
+++ b/tsl/test/expected/bgw_reorder_drop_chunks.out
@@ -59,8 +59,8 @@ SELECT * FROM _timescaledb_config.bgw_job;
 (0 rows)
 
 SELECT * FROM timescaledb_information.policy_stats;
- hypertable | job_id | last_run_success | last_finish | last_successful_finish | last_start | next_start | total_runs | total_failures 
-------------+--------+------------------+-------------+------------------------+------------+------------+------------+----------------
+ hypertable | job_id | last_run_started_at | last_successful_finish | last_run_status | job_status | last_run_duration | next_scheduled_run | total_runs | total_successes | total_failures 
+------------+--------+---------------------+------------------------+-----------------+------------+-------------------+--------------------+------------+-----------------+----------------
 (0 rows)
 
 \c :TEST_DBNAME :ROLE_DEFAULT_PERM_USER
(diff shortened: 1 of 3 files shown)


tsl/test/expected/bgw_reorder_drop_chunks.out
```diff
@@ -292,9 +292,9 @@ SELECT indexrelid::regclass, indisclustered
 
 --check that views work correctly
 SELECT * FROM timescaledb_information.policy_stats;
-     hypertable     | job_id | last_run_success |           last_finish           |     last_successful_finish      |           last_start            |           next_start            | total_runs | total_failures 
---------------------+--------+------------------+---------------------------------+---------------------------------+---------------------------------+---------------------------------+------------+----------------
- test_reorder_table |   1000 | t                | Fri Dec 31 16:00:00.05 1999 PST | Fri Dec 31 16:00:00.05 1999 PST | Fri Dec 31 16:00:00.05 1999 PST | Tue Jan 04 16:00:00.05 2000 PST |          3 |              0
+     hypertable     | job_id |       last_run_started_at       |     last_successful_finish      | last_run_status | job_status | last_run_duration |       next_scheduled_run        | total_runs | total_successes | total_failures
```
erimatnor, line 295: I think there are some inconsistencies here.

For instance, we say `last_run_started_at` but not `last_successful_finish_at`.

Also, I think we talked about leaving out unnecessary columns that can easily be computed from other columns. Examples include:

* total_successes (which I guess is total_runs - total_failures).  Also, why do we need the "total_" prefix? Isn't it obvious that this is the total? Can it be anything else? 
* Similarly, I guess `last_run_duration=last_successful_finish - last_run_started_at` or?

What is the difference between `job_status` and `last_run_status`? Intuitively, I would think that `last_run_status` is the last status of the previous job run, which is the same status shown in `job_status` for the current run. The name implies that they are similar somehow (both show status), but it seems that these are different statuses. For instance, I guess it doesn't make sense for `last_run_status` to be `Scheduled` while job status obviously can be. Maybe `last_run_status` should be a boolean named `last_run_success`?

erimatnor (commented):
This changes look good, but I am wondering how it deviates, if at all, from the view redesign (or I guess we maybe punted on this).

I am wondering why these fields weren't in policy_stats before, and if it now means that many fields aren't relevant for many policies?

The test diffs only show one actual policy in the output, which makes me wonder what other policies look like?

Also, didn't we decide that we should leave out columns like `total_successes` since it can easily be computed from `total_runs` and `total_failures`?

tsl/test/expected/bgw_reorder_drop_chunks.out
gayyappan: @erimatnor The reason to include total_runs, total_successes and total_failures explicitly is due to possibility of crashes. If they don't add up , then there was a crash.
Yes, last_run_duration is a derived column and can be dropped.
last_run_status shows success/failure when the job was last run. (I don't really like the name last_run_success. User now has to interpret what this boolean means whereas Success/Failed is explicit).
job_status shows what is the current execution status : Executing or Scheduled.

erimatnor: Is it correct to say that `total_runs` means "number of jobs started", i.e., it could include jobs that crashed and never registered in `total_successes`. If this is the case, it is non-obvious. Total runs sound like it is total completed runs. Maybe better name sin that case are `num_starts`, `num_failures` and `num_successes`?

Anyway, it might make sense to take this to the design doc and let other people chime in.

erimatnor (approved):
Approving, given that this is an improvement from before, but let's revisit later to see if we need to tweak before 2.0. Could you add an issue for it?

WireBaron (approved):

svenklemm (approved):

PR #2372, opened 2020-09-11
Check start > end for refresh policy

gayyappan:
Add check to verify that start_interval >
end_interval specified by the refresh continuous
aggregate policy

### DIFF
--- a/tsl/src/bgw_policy/continuous_aggregate_api.c
+++ b/tsl/src/bgw_policy/continuous_aggregate_api.c
@@ -176,6 +176,47 @@ convert_interval_arg(Oid dim_type, Datum interval, Oid *interval_type, const cha
 	return ts_time_datum_convert_arg(interval, interval_type, convert_to);
 }
 
+static void
+check_valid_interval_values(Oid interval_type, Datum start_interval, Datum end_interval)
+{
+	bool valid = true;
+	if (IS_INTEGER_TYPE(interval_type))
+	{
+		switch (interval_type)
+		{
+			case INT2OID:
+			{
+				if (DatumGetInt16(start_interval) <= DatumGetInt16(end_interval))
+					valid = false;
+				break;
+			}
+			case INT4OID:
+			{
+				if (DatumGetInt32(start_interval) <= DatumGetInt32(end_interval))
+					valid = false;
+				break;
+			}
+			case INT8OID:
+			{
+				if (DatumGetInt64(start_interval) <= DatumGetInt64(end_interval))
+					valid = false;
+				break;
+			}
+		}
+	}
+	else
+	{
+		Assert(interval_type == INTERVALOID);
+		valid = DatumGetBool(DirectFunctionCall2(interval_gt, start_interval, end_interval));
+	}
+	if (!valid)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("start interval should be greater than end interval")));
+	}
+}
+
 Datum
 policy_refresh_cagg_add(PG_FUNCTION_ARGS)
 {
@@ -192,6 +233,7 @@ policy_refresh_cagg_add(PG_FUNCTION_ARGS)
 	Oid dim_type, start_interval_type, end_interval_type;
 	Oid cagg_oid, owner_id;
 	List *jobs;
+	JsonbParseState *parse_state = NULL;
 	bool if_not_exists, start_isnull, end_isnull;
 
 	/* Verify that the owner can create a background worker */
@@ -228,6 +270,12 @@ policy_refresh_cagg_add(PG_FUNCTION_ARGS)
 	if (!end_isnull)
 		end_interval =
 			convert_interval_arg(dim_type, end_interval, &end_interval_type, "end_interval");
+
+	if (!start_isnull && !end_isnull)
+	{
+		Assert(start_interval_type == end_interval_type);
+		check_valid_interval_values(start_interval_type, start_interval, end_interval);
+	}
 
 	if (PG_ARGISNULL(3))
 		ereport(ERROR,
@@ -285,7 +333,6 @@ policy_refresh_cagg_add(PG_FUNCTION_ARGS)
 	namestrcpy(&proc_schema, INTERNAL_SCHEMA_NAME);
 	namestrcpy(&owner, GetUserNameFromId(owner_id, false));
 
-	JsonbParseState *parse_state = NULL;
 	pushJsonbValue(&parse_state, WJB_BEGIN_OBJECT, NULL);
 	ts_jsonb_add_int32(parse_state, CONFIG_KEY_MAT_HYPERTABLE_ID, mat_htid);
 	if (!start_isnull)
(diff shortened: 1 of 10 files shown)


tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -266,11 +310,16 @@ policy_refresh_cagg_add(PG_FUNCTION_ARGS)
 	namestrcpy(&owner, GetUserNameFromId(owner_id, false));
 
 	if (!start_isnull)
-		check_valid_interval(dim_type, start_interval_type, "start_interval");
+		check_valid_interval_type(dim_type, start_interval_type, "start_interval");
 	if (!end_isnull)
-		check_valid_interval(dim_type, end_interval_type, "end_interval");
+		check_valid_interval_type(dim_type, end_interval_type, "end_interval");
+
+	if (!(start_isnull || end_isnull))
```
WireBaron, line 317: I think it's easier to follow like this:
```suggestion
	if (!start_isnull && !end_isnull)
```

WireBaron (approved):

svenklemm (changes requested):
this is not correct because an interval consists of 3 components and whether X - INTERVAL is smaller then X - INTERVAL2 can only be decided by knowing X and comparing the evaluated expressions
```
dev=# SELECT '2020-10-26'::timestamptz - '1d'::interval, '2020-10-26'::timestamptz - '24h'::interval, '1d'::interval < '24h'::interval, '2020-10-26'::timestamptz - '1d'::interval < '2020-10-26'::timestamptz - '24h'::interval;
        ?column?        |        ?column?        | ?column? | ?column?
------------------------+------------------------+----------+----------
 2020-10-25 00:00:00+02 | 2020-10-25 01:00:00+02 | f        | t
(1 row)
```

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -165,6 +165,49 @@ check_valid_interval(Oid dim_type, Oid interval_type, const char *str_msg)
...
+				if (DatumGetInt64(start_interval) <= DatumGetInt64(end_interval))
+					valid = false;
+				break;
+			}
+		}
+	}
+	else
+	{
+		Assert(interval_type == INTERVALOID);
+		valid = DatumGetBool(DirectFunctionCall2(interval_gt,
```
erimatnor, line 199: Interval can be much larger than the valid timestamp/date type on the table, so we should probably check that the interval is not bigger than the valid timestamp range too (e.g., TIMESTAMP_END - TIMESTAMP_MIN).

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -266,11 +310,16 @@ policy_refresh_cagg_add(PG_FUNCTION_ARGS)
...
 	if (!start_isnull)
-		check_valid_interval(dim_type, start_interval_type, "start_interval");
+		check_valid_interval_type(dim_type, start_interval_type, "start_interval");
 	if (!end_isnull)
-		check_valid_interval(dim_type, end_interval_type, "end_interval");
+		check_valid_interval_type(dim_type, end_interval_type, "end_interval");
+
+	if (!(start_isnull || end_isnull))
+	{
+		Assert(start_interval_type == end_interval_type);
```
pmwkaa, line 319: I am not really sure that is correct assumption, maybe we should error instead? This will likely fail if we pass start_interval::smallint and end_interval::int as arguments, unless I am mistaken

gayyappan: yes, it should. We check that the interval_type matches the dim_type just before this (line 313). Even with the PR change for automatically discovering the types , the interval types should be the same (and match the dimension type) before we compare their values.

tsl/test/sql/continuous_aggs_policy.sql
```diff
@@ -153,6 +158,28 @@ SELECT * FROM mat_smallint;
...
+\set ON_ERROR_STOP 0
+SELECT add_continuous_aggregate_policy('mat_bigint', 5::bigint, 10::bigint , '1 h'::interval) ;
+\set ON_ERROR_STOP 1
+SELECT add_continuous_aggregate_policy('mat_bigint', 15::bigint, 0::bigint , '1 h'::interval) as job_id \gset
+INSERT INTO bigint_tab VALUES(5);
+INSERT INTO bigint_tab VALUES(10);
+INSERT INTO bigint_tab VALUES(20);
+CALL run_job(:job_id);
+SELECT * FROM mat_bigint;
+
```
svenklemm, line 182: can you add a check with '10month' start and '299 day' end and show its accepted but throws runtime error

svenklemm: also can you add test with '1000000 year' which would always overflow timestamp and show this is handled properly

gayyappan: added

svenklemm (approved):

tsl/test/expected/continuous_aggs_policy.out
```diff
@@ -53,11 +53,14 @@ SELECT add_continuous_aggregate_policy('mat_m1', '1 day'::interval, 10 );
...
 ERROR:  invalid parameter value for end_interval
+--start_interval < end_interval
+SELECT add_continuous_aggregate_policy('mat_m1', 5, 10, '1h'::interval) as job_id \gset
+ERROR:  start interval should be greater than end interval
 SELECT add_continuous_aggregate_policy('mat_m1', 20, 10, '1h'::interval) as job_id \gset
 --adding again should warn/error
 SELECT add_continuous_aggregate_policy('mat_m1', 20, 10, '1h'::interval, if_not_exists=>false);
 ERROR:  refresh policy already exists for continuous aggregate "mat_m1"
-SELECT add_continuous_aggregate_policy('mat_m1', 10, 20, '1h'::interval, if_not_exists=>true);
+SELECT add_continuous_aggregate_policy('mat_m1', 20, 15, '1h'::interval, if_not_exists=>true);
```
k-rus, line 63: I don't see a point in this change. Should the error be different?

tsl/test/expected/continuous_aggs_policy.out
```diff
@@ -118,9 +121,10 @@ ERROR:  must be owner of continuous aggregate "mat_m1"
 SET ROLE :ROLE_DEFAULT_PERM_USER;
 DROP MATERIALIZED VIEW mat_m1;
 --- code coverage tests : add policy for timestamp and date based table ---
+--- to prevent NOTICES set message level to warning
+SET client_min_messages TO warning;
```
k-rus, line 125: Can it be done at the place of the problematic test and then restored?

k-rus (commented):
Two questions related to test changes.

tsl/test/expected/continuous_aggs_policy.out
gayyappan: This is to make sure we have a test case for this scenario (line 64). Added tests for start < end.

k-rus: Should it be added at another place, since the error here is that the policy already exists? Or do I miss anything?

tsl/test/expected/continuous_aggs_policy.out
gayyappan: done.

tsl/test/expected/continuous_aggs_policy.out
k-rus: Sorry, I misunderstood. I still often confused that the start interval should be bigger than the end interval. Thank you for explaining, now I follow.

PR #2376, opened 2020-09-13
Fix caggs width expression handling on int based hypertables

### LINKED ISSUE #1597: Can't create contiunous aggregate with bigint time column
 - OS: Docker - `timescale/timescaledb:1.5.1-pg11`
 - TimescaleDB version (output of `\dx` in `psql`): 1.5.1
 - Installation method: Docker

**Describe the bug**
Can't create continous aggregate using bigint as time column. Timescaledb complains that 
`[0A000] ERROR: first argument to time_bucket function should be a constant for continuous aggregate query`

**To Reproduce**
1. Create simple table:
```sql
create table test
(
    time  bigint,
    value float,
    primary key (time)
)
```
2. Make it a hypertable
```sql
select *
from create_hypertable('test', 'time', chunk_time_interval := 60 * 60 * 60)
```
3. Attempt to create continuos aggregate
```sql

create view test_rollup_5m
            WITH (timescaledb.continuous) --This flag is what makes the view continuous
as
select time_bucket(300::bigint, time) as bucket,
       sum(value)
from test
group by bucket;
```
4. Creation fails with `[0A000] ERROR: first argument to time_bucket function should be a constant for continuous aggregate query` error.

**Expected behavior**
TimescaleDB should create the continous aggregate view.

**Actual behavior**
TimescaleDB fails to create continuous aggregate view.

**Screenshots**
_None_

**Additional context**
_None_

svenklemm:
This patch changes the time_bucket validation to constify the width
argument so any immutable expression will be allowed. This change
allows continuous aggregates on hypertables with int8 time dimension
without requiring a workaround from a user to get the time_bucket
call accepted.

Fixes #1597

### DIFF
--- a/tsl/test/expected/continuous_aggs_ddl.out
+++ b/tsl/test/expected/continuous_aggs_ddl.out
@@ -940,5 +940,126 @@ DROP MATERIALIZED VIEW whatever_view_1;
 NOTICE:  drop cascades to table _timescaledb_internal._hyper_13_24_chunk
 DROP MATERIALIZED VIEW whatever_view_2;
 NOTICE:  drop cascades to table _timescaledb_internal._hyper_14_25_chunk
+-- test bucket width expressions on integer hypertables
+CREATE TABLE metrics_int2 (
+  time int2 NOT NULL,
+  device_id int,
+  v1 float,
+  v2 float
+);
+CREATE TABLE metrics_int4 (
+  time int4 NOT NULL,
+  device_id int,
+  v1 float,
+  v2 float
+);
+CREATE TABLE metrics_int8 (
+  time int8 NOT NULL,
+  device_id int,
+  v1 float,
+  v2 float
+);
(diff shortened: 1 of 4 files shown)


tsl/test/expected/continuous_aggs_ddl.out
```diff
@@ -940,5 +940,105 @@ DROP MATERIALIZED VIEW whatever_view_1;
 NOTICE:  drop cascades to table _timescaledb_internal._hyper_13_24_chunk
 DROP MATERIALIZED VIEW whatever_view_2;
 NOTICE:  drop cascades to table _timescaledb_internal._hyper_14_25_chunk
+-- test bucket width expressions
+CREATE TABLE metrics_int2 (
```
erimatnor, line 944: FWIW, tests on all types of tables already exists in other tests, e.g., `continuous_aggs_refresh`. I suggest repurposing/extending those tests instead of adding new ones that are basically the same.

gayyappan (approved):

pmwkaa (approved):

PR #2384, opened 2020-09-14
Filter materialized hypertables in view

### LINKED ISSUE #2383: Do not show materialized hypertables in the hypertables view
Materialized hypertables provide internal storage for continuous aggregates and are created as a side effect of creating a continuous aggregate. Therefore, they are not user-facing objects and it would clean up the hypertables view to only show hypertables that are directly created by the user.

Compression hypertables are similarly created as a side-effect of enabling compression and those internal hypertables are not shown in the hypertables view.

erimatnor:
This change filters materialized hypertables from the hypertables
view, similar to how internal compression hypertables are
filtered.

Materialized hypertables are internal objects created as a side effect
of creating a continuous aggregate, and these internal hypertables are
still listed in the continuous_aggregates view.

Fixes #2383

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -25,6 +25,8 @@ CREATE OR REPLACE VIEW timescaledb_information.hypertables AS
         INNER JOIN pg_tables t
         ON ht.table_name=t.tablename
            AND ht.schema_name=t.schemaname
+        LEFT OUTER JOIN _timescaledb_catalog.continuous_agg ca
+        ON ca.mat_hypertable_id=ht.id
         LEFT OUTER JOIN (
             SELECT hypertable_id,
             array_agg(tablespace_name ORDER BY id) as tablespace_list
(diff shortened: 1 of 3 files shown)


pmwkaa (approved):

pmwkaa:
@erimatnor to note: unless intentional, link in the description points to pull request.

sql/views.sql
```diff
@@ -38,7 +40,7 @@ CREATE OR REPLACE VIEW timescaledb_information.hypertables AS
                       GROUP BY hypertable_id) dn
     ON ht.id = dn.hypertable_id
     WHERE ht.compressed is false --> no internal compression tables
```
k-rus, line 42: It's not direct part of PR, but can you fix capitalisation of the keywords, since it is related and was overlooked in the correct PR review?

k-rus (approved):

k-rus:
@erimatnor I agree with @pmwkaa that the commit message should link to an issue, which it fixes, not to a PR.

erimatnor:
> @erimatnor to note: unless intentional, link in the description points to a pull request.

Fixed!

sql/views.sql
erimatnor: This issue is prevalent throughout the SQL code. I suggest we adopt a formatter and make sure this is fixed in all places.

k-rus: This is next to your line, so it was an obvious and easy fix, taking in account that you were fixing commit message and rebase in any case. Postponing is also fine (it just looked to me weird to have two lines next to each other in different style).

svenklemm (approved):

mkindahl (approved):

Issue #2286, opened 2020-08-31
Avoid having to cast time args in policies


erimatnor:
Policies such as retention and refresh requires casts on arguments that are of type "any". This is awkward for the user and we can do better.

The function refresh_continuous_aggregate already has a way to "guess" the type based on the time dimension's type. We should try to do the same for policies.

erimatnor:
Closing since fix was merged in #2387

PR #2387, opened 2020-09-14
Avoid having to cast time arg for cagg policy

pmwkaa:
This patch does a minor refactoring and adds a way to guess
interval argument type based on used cagg

Issue: #2286

### DIFF
--- a/src/time_utils.c
+++ b/src/time_utils.c
@@ -41,6 +41,45 @@ subtract_interval_from_now(Oid timetype, const Interval *interval)
 	return res;
 }
 
+Datum
+ts_time_datum_convert_arg(Datum arg, Oid *argtype, Oid timetype)
+{
+	Oid type = *argtype;
+
+	if (!OidIsValid(type) || type == UNKNOWNOID)
+	{
+		Oid infuncid = InvalidOid;
+		Oid typeioparam;
+
+		type = timetype;
+		getTypeInputInfo(type, &infuncid, &typeioparam);
+
+		switch (get_func_nargs(infuncid))
+		{
+			case 1:
+				/* Functions that take one input argument, e.g., the Date function */
+				arg = OidFunctionCall1(infuncid, arg);
+				break;
+			case 3:
+				/* Timestamp functions take three input arguments */
+				arg = OidFunctionCall3(infuncid,
+									   arg,
+									   ObjectIdGetDatum(InvalidOid),
+									   Int32GetDatum(-1));
+				break;
+			default:
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg("invalid time argument"),
+						 errhint("Time argument requires an explicit cast.")));
+		}
+
+		*argtype = type;
+	}
+
+	return arg;
+}
+
 /*
  * Get the internal time value from a pseudo-type function argument.
  *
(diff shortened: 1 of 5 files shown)


tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -143,26 +145,31 @@ json_add_dim_interval_value(JsonbParseState *parse_state, const char *json_label
 	}
 }
 
-static void
-check_valid_interval(Oid dim_type, Oid interval_type, const char *str_msg)
+static Datum
+convert_interval_arg(Oid dim_type, Datum interval, Oid *interval_type, const char *str_msg)
```
erimatnor, line 149: Isn't this conversion useful for other policies as well (e.g., retention)? I figured this would be in time_utils or policy_utils for use by all policies.

pmwkaa: Yes, I thought maybe to do that in another PR related to the retention policy.

tsl/test/sql/continuous_aggs_policy.sql
```diff
@@ -86,9 +86,9 @@ CREATE MATERIALIZED VIEW max_mat_view_date
         GROUP BY 1 WITH NO DATA;
 
 \set ON_ERROR_STOP 0
-SELECT add_continuous_aggregate_policy('max_mat_view_date', '2 days'::interval, 10 , '1 day'::interval);
+SELECT add_continuous_aggregate_policy('max_mat_view_date', '2 days', 10, '1 day'::interval);
 \set ON_ERROR_STOP 1
-SELECT add_continuous_aggregate_policy('max_mat_view_date', '2 day'::interval, '1 day'::interval , '1 day'::interval) as job_id \gset
+SELECT add_continuous_aggregate_policy('max_mat_view_date', '2 day', '1 day', '1 day'::interval) as job_id \gset
```
erimatnor, line 91: Nit: Would be nice to get rid of the cast on schedule_interval too. Perhaps also for a separate PR?

```suggestion
SELECT add_continuous_aggregate_policy('max_mat_view_date', '2 days', '1 day', '1 day'::interval) as job_id \gset
```

erimatnor (approved):

src/time_utils.c
```diff
@@ -41,6 +41,45 @@ subtract_interval_from_now(Oid timetype, const Interval *interval)
 	return res;
 }
 
+Datum
```
k-rus, line 44: I think it is worth do document this function as it is part of internal API, so it is clear what to expect from it and how to use it. Currently I cannot guess from the function name what to expect from it.

src/time_utils.c
```diff
@@ -41,6 +41,45 @@ subtract_interval_from_now(Oid timetype, const Interval *interval)
...
+		Oid typeioparam;
+
+		type = timetype;
+		getTypeInputInfo(type, &infuncid, &typeioparam);
+
+		switch (get_func_nargs(infuncid))
+		{
+			case 1:
+				/* Functions that take one input argument, e.g., the Date function */
+				arg = OidFunctionCall1(infuncid, arg);
```
k-rus, line 61: Am I correct that this is a reuse of variable `arg`, which first contains the original argument, and then assigned the converted argument? I suggest to use different variable and actually it will contain different object to my understanding.

src/time_utils.c
```diff
@@ -75,36 +114,9 @@ subtract_interval_from_now(Oid timetype, const Interval *interval)
...
-			default:
-				ereport(ERROR,
-						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-						 errmsg("invalid time argument"),
-						 errhint("Time argument requires an explicit cast.")));
-		}
-	}
+	/* If no explicit cast was done by the user, try to convert the argument
+	 * to the time type used by the continuous aggregate. */
+	arg = ts_time_datum_convert_arg(arg, &argtype, timetype);
```
k-rus, line 119: Same comment here, since `arg` will contain different object after the call. To avoid code changes, the original variable can be called `arg_in`.

src/time_utils.c
```diff
@@ -41,6 +41,45 @@ subtract_interval_from_now(Oid timetype, const Interval *interval)
...
+				arg = OidFunctionCall3(infuncid,
+									   arg,
+									   ObjectIdGetDatum(InvalidOid),
+									   Int32GetDatum(-1));
+				break;
+			default:
+				ereport(ERROR,
+						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+						 errmsg("invalid time argument"),
+						 errhint("Time argument requires an explicit cast.")));
```
k-rus, line 74: Is it possible to test this case? Codecov complains that it's not covered.

k-rus (approved):
LGTM
It might be useful to give short comment describing the function.
And I am not favour of reusing variable names, especially it is set to a different value.

src/time_utils.c
pmwkaa: I think most of those tested are already covered by continuous agg refresh function tests, which are reusing the same functionality.

k-rus: I guess Codecov would not complain if it is covered. I searched that such error is never expected in our tests.

pmwkaa: I don't think this case is actually possible right now, it was made more like a precaution in case new time functions will arrive in future versions I believe

PR #2391, opened 2020-09-14
Allow REFRESH MATERIALIZED VIEW to refresh caggs

erimatnor:
It is now possible to use `REFRESH MATERIALIZED VIEW` as an
alternative to `refresh_continuous_aggregate` when doing "infinite"
refreshes.

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -3330,8 +3330,7 @@ process_refresh_mat_view_start(ProcessUtilityArgs *args)
 	ScanIterator continuous_aggregate_iter;
 	NameData view_name;
 	NameData view_schema;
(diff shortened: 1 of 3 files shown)


pmwkaa (approved):

cevian (commented):
How does this interact with dropped data? It would be great if this refreshed everything /that hasn't been dropped/

erimatnor:
Note that this (temporarily) redirects `REFRESH MATERIALIZED VIEW` to the new refresh procedure instead of blocking. This is to avoid breaking existing update scripts while still having the command do something useful as opposed to calling the old materializer.

k-rus (approved):

PR #2392, opened 2020-09-14
Remove options from continuous aggregate

k-rus:
Removes options refresh_lag, max_interval_per_job and 
ignore_invalidation_older_than from continuous aggregate creation with
CREATE MATERIALIZED VIEW, since they are not related this statement 
any more. They are already replaced with the corresponding options in 
add_continuous_aggregate_policy.

This commit removes only options, while the options are still stored 
in the catalog and need to be removed from there in a separate PR.


erimatnor (commented):
Looks good. I think there are, however, several functions in, e.g., `src/continuous_agg.c` related to these options that can now be removed, like `ts_continuous_aggs_max_ignore_invalidation_older_than` and `ts_continuous_aggs_get_minimum_invalidation_time`.

src/continuous_agg.h
```diff
@@ -20,10 +20,7 @@
 typedef enum ContinuousAggViewOption
 {
 	ContinuousEnabled = 0,
-	ContinuousViewOptionRefreshLag,
-	ContinuousViewOptionMaxIntervalPerRun,
 	ContinuousViewOptionCreateGroupIndex,
-	ContinuousViewOptionIgnoreInvalidationOlderThan,
 	ContinuousViewOptionMaterializedOnly,
 } ContinuousAggViewOption;
```
WireBaron, line 25: We don't persist any values referring to these options in any of our catalog tables do we?  It might be safer to just rename these options like `ContinuousViewOptionRefreshLag_deprecated`.

WireBaron (approved):

src/continuous_agg.h
k-rus: @WireBaron The catalog tables still have the values, but they are not used anywhere. I don't understand the idea about renaming to `ContinuousViewOptionRefreshLag_deprecated`. Can you clarify? I think it is better not to provide the options at all, than to allow to set the options without any effect on anything.

k-rus:
> ooks good. I think there are, however, several functions in, e.g., `src/continuous_agg.c` related to these options that can now be removed, like `ts_continuous_aggs_max_ignore_invalidation_older_than` and `ts_continuous_aggs_get_minimum_invalidation_time`.

@erimatnor Removing these exported functions require some work and I would not be able to complete it for this beta.

erimatnor:
> > ooks good. I think there are, however, several functions in, e.g., `src/continuous_agg.c` related to these options that can now be removed, like `ts_continuous_aggs_max_ignore_invalidation_older_than` and `ts_continuous_aggs_get_minimum_invalidation_time`.
> 
> @erimatnor Removing these exported functions require some work and I would not be able to complete it for this beta.

I believe it is just a deleting the functions and places where called since options aren't used it should not make a difference to remove the code. But it is fine to leave for later.

erimatnor (approved):

k-rus:
> I believe it is just a deleting the functions and places where called since options aren't used it should not make a difference to remove the code. But it is fine to leave for later.

At least one function is indirectly used in the invalid cache initialisation. So I will need to be more careful than just removing.

PR #2399, opened 2020-09-15
Fix failing isolation test on PG11.0

k-rus:
Early version of PG11 produces slightly different output for isolation
tests. Thus the affected tests are ignored.
continuous_aggs_concurrent_refresh is one of them and this commit adds
it to the ignored test list.

PR comment:
[Link](https://github.com/timescale/timescaledb/runs/1114025453?check_suite_focus=true) to the failed test


pmwkaa (approved):
Wasn't it possible to update the output for PG11 instead?

k-rus:
> Wasn't it possible to update the output for PG11 instead?

@pmwkaa The output between running on PG11.0 and PG11.7 is different for this test. So we ignore such tests on PG11.0 but run on latest PG11. I don't think it is worth the effort to build infrastructure, which will work in both cases.

gayyappan (approved):

PR #2403, opened 2020-09-15
Add invalidations for incomplete aggregates

### LINKED ISSUE #2314: Add invalidation from completed_threshold and onward during upgrade
When upgrading a continuous aggregate, from pre 2.0, we need to add an invalidation `[ completed_threshold, +Infinity ]`. Otherwise, data that has not yet been refreshed will not be properly refreshed.

WireBaron:
As part of the 2.0 continous aggregate changes, we are removing the
continuous_aggs_completed_threshold table.  However, this may result
in currently running aggregates being considered complete even if
their completed threshold hadn't reached the invalidation threshold.
This change fixes this by adding an entry to the invalidation log
for any such aggregates.

Fixes #2314


erimatnor:
@WireBaron Is it possible to test this somehow? Perhaps in an update script test that compares refreshes in a new region on an updated instance vs a freshly installed one?

sql/updates/latest-dev.sql
```diff
@@ -303,6 +303,14 @@ SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_job', 'WHERE
 GRANT SELECT ON _timescaledb_config.bgw_job TO PUBLIC;
 GRANT SELECT ON _timescaledb_config.bgw_job_id_seq TO PUBLIC;
 
+-- capture invalidations for any caggs which are not yet complete
+INSERT INTO _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
+SELECT materialization_id, BIGINT '-9223372036854775808', complete.watermark, invalid.watermark
```
gayyappan, line 308: The new invalidated region is [complete.watermark, infinity]?

gayyappan: Also need a test for this case.

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -263,3 +263,51 @@ BEGIN
...
+  END IF;
+END $$;
+
+REFRESH MATERIALIZED VIEW mat_inval;
+
+INSERT INTO inval_test
+SELECT tm.tval, loc.lval, temp 
+FROM (SELECT generate_series(now() + '30 seconds'::interval, now() + '4 hours'::interval, '30 seconds') AS tval) tm 
+    CROSS JOIN (select 'loc_' || generate_series(1,3,1) AS lval) loc,
+    generate_series(70.0, 90.0, 1440) temp;
```
svenklemm, line 313: ok this query does not look right, the now() will guarantee that this test will be flaky and the `generate_series(70.0, 90.0, 1440)` just returns a single row so you could just replace it by `70.0`

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -263,3 +263,51 @@ BEGIN
...
+
+-- test proper invalidations after update
+CREATE TABLE inval_test (time TIMESTAMPTZ, location TEXT, temperature DOUBLE PRECISION);
+SELECT create_hypertable('inval_test', 'time', chunk_time_interval => INTERVAL '1 hour');
+ 
+INSERT INTO inval_test
+SELECT tm.tval, loc.lval, temp 
+FROM (SELECT generate_series(now() - '4 hours'::interval, now(), '30 seconds') AS tval) tm 
+    CROSS JOIN (select 'loc_' || generate_series(1,3,1) AS lval) loc,
+    generate_series(50.0, 70.0, 1443) temp;
```
svenklemm, line 275: ok this query does not look right, the now() will guarantee that this test will be flaky and the generate_series(50.0, 90.0, 1443) just returns a single row so you could just replace it by 50.0

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -263,3 +263,51 @@ BEGIN
...
+  ts_version TEXT;
+BEGIN
+  SELECT extversion INTO ts_version FROM pg_extension WHERE extname = 'timescaledb';
+
+  IF ts_version < '2.0.0' THEN
+    CREATE VIEW mat_inval
+    WITH ( timescaledb.continuous, timescaledb.materialized_only=true,
+           timescaledb.refresh_lag='-2 hour',
+           timescaledb.ignore_invalidation_older_than = '24 hour',
+           timescaledb.refresh_interval='20 minutes' )
```
gayyappan, line 288: would be better to set max_interval_per_job to a big value if the intent here is to get all the data processed by the REFRESH. e.g. max_interval_per_job of 72 hours etc.

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -263,3 +263,51 @@ BEGIN
...
+      FROM inval_test
+      GROUP BY bucket, location WITH NO DATA;
+
+    PERFORM add_continuous_aggregate_policy('mat_inval', '1 day'::interval, '-2 hours'::interval, '20 minutes');
+  END IF;
+END $$;
+
+REFRESH MATERIALIZED VIEW mat_inval;
+
+INSERT INTO inval_test
```
gayyappan, line 309: None of this data is in the original data set . So don't think you have any invalidations for the last completed threshold range. 
The inserts here should be of 2 kinds.
1. generate invalidations:  a subset of the range from the insert at line 271.
2. New data (if you are testing this case as well) for a new range like at line 311.

WireBaron: This is actually intended to all be new data.  I believe we already have other tests for existing invalidations, what this test is intended to verify is that we generate a new invalidation that covers new data when updating to 2.0.

test/sql/updates/setup.continuous_aggs.v2.sql
WireBaron: See my comment above about now() just being used to generate a set number of rows above the invalidation threshold.  If you still think this will be flaky, I can change this to an arbitrarily far in the future data.  Fixed the generate series.

test/sql/updates/setup.continuous_aggs.v2.sql
erimatnor: Is the refresh and query post upgrade done int test-rerun.v6.sql?

erimatnor (approved):
Approving.

The update test is a little bit hard to follow, but I assume you have tried the update test without adding the invalidation during upgrade with a failed result?

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -263,3 +263,49 @@ BEGIN
 END $$;
 
 REFRESH MATERIALIZED VIEW mat_ignoreinval;
+
+-- test proper invalidations after update
```
gayyappan, line 267: Could you modify the comment here? This test case is for inserts after a REFRESH. (We have been using the term invalidations to refer to changes to the raw data that would invalidate a previously computed materialization). These inserts (lines 309-311)  are not logged into the  invalidation_log catalog tables.

sql/updates/latest-dev.sql
```diff
@@ -303,6 +303,11 @@ SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_job', 'WHERE
 GRANT SELECT ON _timescaledb_config.bgw_job TO PUBLIC;
 GRANT SELECT ON _timescaledb_config.bgw_job_id_seq TO PUBLIC;
 
+-- set up future invalidation for current caggs
```
gayyappan, line 306:
```suggestion
-- Add entry to materialization invalidation log to indicate that [watermark, +infinity) is invalid
```

sql/updates/latest-dev.sql
```diff
@@ -303,6 +303,11 @@ SELECT pg_catalog.pg_extension_config_dump('_timescaledb_config.bgw_job', 'WHERE
...
+INSERT INTO _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
+SELECT materialization_id, BIGINT '-9223372036854775808', watermark, 9223372036854775807
```
gayyappan, line 308: If we have a new cagg that has been never refreshed , do we need an entry in the cont agg materialization log that says [-infinity, +infinity) is invalid? (cc @erimatnor )

svenklemm (approved):

sql/updates/latest-dev.sql
WireBaron: added

test/sql/updates/setup.continuous_aggs.v2.sql
WireBaron: Yes, I needed to move that from the post script to the rerun script, as the post script isn't called on the rerun container (the post was just run on the upgraded container and a dump/restore of the upgraded container, both of which had the wrong data and hence no diff).

test/sql/updates/setup.continuous_aggs.v2.sql
WireBaron: Right, it's confusing as the terminology has slightly changed in 2.0 (we now also have invalidations for regions that have never been refreshed).  I'm going with `test new data beyond the invalidation threshold is properly handled` here, let me know if that works.

sql/updates/latest-dev.sql
WireBaron: It seems like that's pretty tight window.  I believe the background job gets launched as soon as we create the materialized view in 1.7.

WireBaron:
I did add a new invalidation for (-inf, inf) when updating 1.7 caggs that have no completion_threshold.  Unfortunately I was unable to add a test to reliably demonstrate the correctness of this change in the update test.

erimatnor:
One question was whether we should add an invalidation for [-Inf, ignore_invalidation_older_than] to allow refreshing of the region where invalidations previously had been ignored. Need not be in this PR, though.

gayyappan (approved):

WireBaron:
> One question was whether we should add an invalidation for [-Inf, ignore_invalidation_older_than] to allow refreshing of the region where invalidations previously had been ignored. Need not be in this PR, though.

Given how long this has been in review, I'm going to go ahead and push this now that it's finally cleared to go.  I'll add a new issue for this older region though.

PR #2404, opened 2020-09-16
Fix alter_job not updating new job fields

### LINKED ISSUE #2402: alter_job fails to update job config
When using alter_job to update a job with a new configuration, the command succeeds, but makes no changes to the job tuple.

In the code, bgw_job_tuple_update_by_id doesn't appear to propagate updates to 'config' or 'scheduled' values to the bgw_jobs table.

svenklemm:
Fix bgw_job_tuple_update_by_id to also update scheduled and config
field of bgw_job.

Fixes #2402

### DIFF
--- a/src/bgw/job.c
+++ b/src/bgw/job.c
@@ -901,31 +901,35 @@ ts_bgw_job_insert_relation(Name application_name, Name job_type, Interval *sched
 	return values[AttrNumberGetAttrOffset(Anum_bgw_job_id)];
 }
 
+/* This function only updates the fields modifiable with alter_job. */
 static ScanTupleResult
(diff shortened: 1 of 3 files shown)


gayyappan (approved):

k-rus (commented):
LGTM, few nit comments
Can you add commit message body and clarify, which new job fields are fixed? Isn't just the config in JSON?
Can you add link to the issue being fixed?
Variable name `mod_tuple` is not easy to follow, so I suggest to fix it back to `new_tuple`, which is commonly used through the code.
It will be an advantage of having new line separating declaration block and the assert.

WireBaron (approved):

PR #2411, opened 2020-09-17
Fix ApacheOnly regression test

svenklemm:
The recently added test for hypertable detection used compression
which is not available in ApacheOnly tests so we use that test
to regresscheck-shared. Additionally we move the other test in
plan_hypertable_cache to plan_expand_hypertable to reduce the number
of tests.


k-rus:
`decompress_placeholdervar` fails on 32bit after the test was moved. There is some difference in a query plan.

erimatnor (approved):

gayyappan (approved):

PR #2415, opened 2020-09-18
Check owner when refreshing a continuous aggregate

### LINKED ISSUE #2408: Permissions checks for refreshing a continuous aggregate
Currently, permissions for refreshing a continuous aggregate is governed primarly by standard INSERT and DELETE perms on the materialized hypertable during materialization. We should probably do basic checks prior to materialization to ensure that we check for appropriate permissions early on.

Refreshing a regular materialized view requires being the owner. It is reasonable that we have the same requirement for refresh on a continuous aggregate.

erimatnor:
This change checks for ownership of the user view when refreshing a
continuous aggregate (it's assumed the other objects have the same
owner). This mimics the checks done when refreshing a regular
materialized view.

Fixes #2408

### DIFF
--- a/tsl/src/continuous_aggs/refresh.c
+++ b/tsl/src/continuous_aggs/refresh.c
@@ -358,6 +358,12 @@ continuous_agg_refresh_internal(const ContinuousAgg *cagg,
 	int64 computed_invalidation_threshold;
 	int64 invalidation_threshold;
 
+	/* Like regular materialized views, require owner to refresh. */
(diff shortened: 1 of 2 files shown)


mkindahl (approved):

k-rus (approved):
This change is definitely an improvement. And it seems to me more work is necessary related to the checks that there are several paths in the code, which do cagg refresh, or there should be refactoring of different paths in future.

PR #2417, opened 2020-09-18
Add timescaledb_information.jobs view

svenklemm:
This patch adds a convenience view to list all configured jobs.

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -69,6 +69,27 @@ CREATE OR REPLACE VIEW timescaledb_information.policy_stats as
     INNER JOIN _timescaledb_internal.bgw_job_stat js on j.id = js.job_id
   ORDER BY ht.schema_name, ht.table_name;
 
+-- view for background worker jobs
+CREATE OR REPLACE VIEW timescaledb_information.jobs AS
+SELECT j.id AS job_id,
+  j.application_name,
+  j.schedule_interval,
+  j.max_runtime,
+  j.max_retries,
+  j.retry_period,
+  j.proc_schema,
+  j.proc_name,
+  j.owner,
+  j.scheduled,
+  j.config,
+  js.next_start,
+  ht.schema_name AS hypertable_schema,
+  ht.table_name AS hypertable_name
+FROM _timescaledb_config.bgw_job j
+  LEFT JOIN _timescaledb_catalog.hypertable ht ON ht.id = j.hypertable_id
+  LEFT JOIN _timescaledb_internal.bgw_job_stat js ON js.job_id = j.id
+ORDER BY j.id;
+
 -- views for continuous aggregate queries ---
 CREATE OR REPLACE VIEW timescaledb_information.continuous_aggregates as
   SELECT format('%1$I.%2$I', cagg.user_view_schema, cagg.user_view_name)::regclass as view_name,
--- a/tsl/test/expected/bgw_custom.out
+++ b/tsl/test/expected/bgw_custom.out
@@ -56,6 +56,17 @@ SELECT add_job('custom_func_definer', '1h', config:='{"type":"function"}'::jsonb
     1004
 (1 row)
 
+SELECT * FROM timescaledb_information.jobs ORDER BY 1;
+ job_id |    application_name    | schedule_interval |   max_runtime   | max_retries | retry_period |      proc_schema      |      proc_name      |       owner       | scheduled |        config         | next_start | hypertable_schema | hypertable_name 
+--------+------------------------+-------------------+-----------------+-------------+--------------+-----------------------+---------------------+-------------------+-----------+-----------------------+------------+-------------------+-----------------
+      1 | Telemetry Reporter [1] | @ 24 hours        | @ 1 min 40 secs |          -1 | @ 1 hour     | _timescaledb_internal | policy_telemetry    | super_user        | t         |                       |            |                   | 
+   1000 | Custom Job [1000]      | @ 1 hour          | @ 0             |          -1 | @ 5 mins     | public                | custom_func         | default_perm_user | t         | {"type": "function"}  |            |                   | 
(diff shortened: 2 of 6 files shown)


sql/views.sql
```diff
@@ -69,6 +69,23 @@ CREATE OR REPLACE VIEW timescaledb_information.policy_stats as
...
+  j.schedule_interval,
+  j.max_runtime,
+  j.max_retries,
+  j.retry_period,
+  j.proc_schema,
+  j.proc_name,
+  j.owner,
+  j.scheduled,
+  j.config,
+  js.next_start
```
k-rus, line 85: This column is the same as `next_scheduled_run` in `policy_stats`. I guess it makes sense to have in both views, but then it is good to use the same name, and `next_scheduled_run` is more meaningful name.

tsl/test/expected/bgw_custom.out
```diff
@@ -108,6 +119,12 @@ SELECT * FROM _timescaledb_config.bgw_job WHERE id >= 1000;
 ----+------------------+-------------------+-------------+-------------+--------------+-------------+-----------+-------+-----------+---------------+--------
 (0 rows)
 
+SELECT count(*) FROM timescaledb_information.jobs WHERE job_id >= 1000;
```
k-rus, line 122: Isn't it better just replace the internal table with the view in the above select?

k-rus (commented):
Minor suggestions

sql/views.sql
svenklemm: this view has all the settings you can set with alter_job where it is named next_start

k-rus: Using different names is confusing. The other possibility is to change the name in `policy_stats` or use `next_scheduled_run` everywhere.

sql/views.sql
```diff
@@ -69,6 +69,23 @@ CREATE OR REPLACE VIEW timescaledb_information.policy_stats as
     INNER JOIN _timescaledb_internal.bgw_job_stat js on j.id = js.job_id
   ORDER BY ht.schema_name, ht.table_name;
 
+-- view for background worker jobs
+CREATE OR REPLACE VIEW timescaledb_information.jobs AS
```
gayyappan, line 73: Should this be named policies instead? We now have 2 views related to jobs: one is called jobs and the other is called policy stats. Since the primary use case is for setting up policies, I suggest we rename this as timescaledb_information.policies.

sql/views.sql
gayyappan: I agree , the name should be consistent. Makes sense to keep next_start here since that is the name of the setting for alter_job and rename  policy_stats view column to next_start.

sql/views.sql
svenklemm: the distinction i tried to introduce between job/policy is that job is something that is handled by the scheduler and it doesnt matter what it does, while policies are concrete implementations that have a specific purpose, so in that sense both should use job since the have nothing policy-specific in them

sql/views.sql
k-rus: The `alter_job` can be renamed too ;) @gayyappan was there any discussion around the name `next_scheduled_run`?
We can fix names later, so now it  can be `next_start` in both views as part of this PR. I will include both views in the discussion doc.

sql/views.sql
gayyappan: then policy_stats should be renamed to job_stats. Also better to get wider input on the name so that we don't have to back and revise it.

svenklemm: i would prefer job then it would also follow our catalog names, but keeping that job/policy distinction in mind i guess the new feature could also be called custom policies

sql/views.sql
gayyappan: @k-rus We finalized on next_start for alter_job when we discussed the job apis. We just need to use that consistently across all the views.

gayyappan (commented):
Also policy_stats now returns a regclass. The other views return schema_name + table_name. So queries need a cast
```
 SELECT *
  FROM timescaledb_information.policy_stats
  WHERE hypertable::text = 'test_table'; 
```
Could you please make this consistent with the other views as part of the policy related view changes?

sql/views.sql
```diff
@@ -69,6 +69,23 @@ CREATE OR REPLACE VIEW timescaledb_information.policy_stats as
...
+  j.application_name,
+  j.schedule_interval,
+  j.max_runtime,
+  j.max_retries,
+  j.retry_period,
+  j.proc_schema,
+  j.proc_name,
+  j.owner,
+  j.scheduled,
+  j.config,
```
WireBaron, line 84: I notice we're writing the entire jobs table except `hypertable_id`.  While I agree that field isn't going to be publicly useful, it might be useful to join in the hypertable table and include the table name and schema.

svenklemm: hypertable_id is completely internal and may point to internal hypertables

WireBaron: Hmm, I do still think including the table name would be useful, though definitely less so when it points to an internal table.  Right now if a user creates two policies with the same parameters, there is no way for them to tell from this new view which job is for which table.

svenklemm: in case of caggs this points to the materialization hypertable, i guess showing the name would be nice but what would you call the columns? for caggs we should resolve to the cagg view name which means we couldnt call the column hypertable_schema/name

WireBaron: I think it's actually okay to use the materialization hypertable for caggs as the continuous_aggregates information view also shows the materialization_hypertable, and can be used to match the job to the cagg view name.

svenklemm: ok added hypertable name and schema

tsl/test/expected/bgw_custom.out
k-rus: BUMP

WireBaron (approved):

gayyappan (approved):

Issue #2359, opened 2020-09-10
Move enterprise features to community (TSL)


erimatnor:
All enterprise features will be moved to community. The only function affected still should be `move_chunks`.

License key checks can also be removed.

PR #2435, opened 2020-09-23
Move enterprise features to community

pmwkaa:
This patch removes enterprise license support and moves `move_chunk()` function under community license (TSL).

Licensing validation code been reworked and simplified.
Previously used `timescaledb.license_key` guc renamed to `timescaledb.license`.

Currently supported license guc values are: `apache` and `tsl` (this is up for discussion @mfreed). License can be changed only if it is set from server configuration file or the server command line.

This change also makes testing code more strict against used license. Apache test suite now is run under apache license which is set in the configuration file. Because of that some test files been moved to the tsl directory. Had to move bgw_db_scheduler test to tsl/test/sql because it used some tsl only api. If this is an issue, we might want to split this test @svenklemm.

Issue: https://github.com/timescale/timescaledb/issues/2359


sql/edition.sql
```diff
@@ -2,36 +2,19 @@
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-APACHE for a copy of the license.
 
-CREATE OR REPLACE FUNCTION _timescaledb_internal.enterprise_enabled() RETURNS BOOLEAN
-AS '@MODULE_PATHNAME@', 'ts_enterprise_enabled' LANGUAGE C;
-
-CREATE OR REPLACE FUNCTION _timescaledb_internal.current_license_key() RETURNS TEXT
-AS '@MODULE_PATHNAME@', 'ts_current_license_key' LANGUAGE C;
-
 CREATE OR REPLACE FUNCTION _timescaledb_internal.tsl_loaded() RETURNS BOOLEAN
```
pmwkaa, line 5: I think we should also remove this function and function below or at least move them to some C function in the tests.

src/guc.c
```diff
@@ -360,15 +360,15 @@ _guc_init(void)
...
 							   /* long_dec= */ "Determines which features are enabled",
-							   /* valueAddr= */ &ts_guc_license_key,
-							   /* bootValue= */ TS_DEFAULT_LICENSE,
+							   /* valueAddr= */ &ts_guc_license,
+							   /* bootValue= */ TS_LICENSE_DEFAULT,
 							   /* context= */ PGC_SUSET,
-							   /* flags= */ GUC_SUPERUSER_ONLY,
-							   /* check_hook= */ ts_license_update_check,
-							   /* assign_hook= */ ts_license_on_assign,
+							   /* flags= */ 0,
```
erimatnor, line 369: Shouldn't this still be superuser only?

src/license_guc.c
```diff
@@ -107,269 +117,149 @@ ts_license_enable_module_loading(void)
 							   false);
 
 	if (result <= 0)
-		elog(ERROR, "invalid value for timescaledb.license_key '%s'", ts_guc_license_key);
+		elog(ERROR, "invalid value for timescaledb.license '%s'", ts_guc_license);
```
erimatnor, line 120: Either:

```suggestion
		elog(ERROR, "invalid value for timescaledb.license: \"%s\"", ts_guc_license);
```

or

```suggestion
		elog(ERROR, "invalid value \"%s\" for timescaledb.license", ts_guc_license);
```

test/expected/edition.out
```diff
@@ -14,40 +14,34 @@ SELECT _timescaledb_internal.tsl_loaded();
...
---------------------
- f
-(1 row)
-
 \unset ECHO
-SET timescaledb.license_key='ApacheOnly';
-ERROR:  invalid value for parameter "timescaledb.license_key": "ApacheOnly"
-DETAIL:  Cannot downgrade a running session to Apache Only.
+SET timescaledb.license='apache';
+ERROR:  invalid value for parameter "timescaledb.license": "apache"
```
erimatnor, line 19: misleading error since "apache" is a valid parameter value. Just not allowed to be set in this way. Is it possible to have a more generic error, e.g., "cannot set TimescaleDB license" and then let the detail message tell us whether the parameter values is not allowed or it is not possible to downgrade.

erimatnor (commented):
Did a quick first pass. Didn't yet look too much into the new logic. Will do that next.

test/expected/edition.out
pmwkaa: I think we can add them in the same way was done for other guc hints. Just to note, that guc hooks should not throw errors

src/guc.c
pmwkaa: I'll double check that, but I think it is still is because of the used context. If am not mistaken, GUC_SUPERUSER_ONLY means to make this guc visible to superuser only, which makes sense because previously it could have an enterprise license key, which we did not want to expose.

test/expected/edition.out
pmwkaa: Actually I am not sure we can change this. Seems like the best we can do here, is to show DETAIL and HINT, which we already did

sql/edition.sql
pmwkaa: Moved this stuff into test files.

src/telemetry/telemetry.c
```diff
@@ -59,6 +59,7 @@
 #define REQ_RELATED_EXTENSIONS "related_extensions"
 #define REQ_METADATA "db_metadata"
 #define REQ_LICENSE_EDITION_APACHE "apache_only"
```
pmwkaa, line 61: still keeping those names for telemetry

src/guc.c
erimatnor: Yeah, I think you are right. Looks like this is only to restrict who can see the GUC. Do we have a check, though, that no-one else but the super user can set the GUC? I am thinking of the "upgrade" case, where a superuser at runtime can "upgrade" to TSL (but not downgrade to Apache).

test/expected/edition.out
erimatnor: So, is "invalid value for parameter..." a PostgreSQL error? I guess we'll just have to live with this in that case.

src/guc.c
pmwkaa: Yes. We have a permission test case

test/expected/edition.out
pmwkaa: Yeah, unfortunately. At least we have additional details with it

src/license_guc.c
```diff
@@ -71,315 +43,218 @@ current_license_can_downgrade_to_apache(void)
...
 }
-#endif
 
-/*
- * `ts_license_on_assign`
- * Links the cross-module function pointers.
- */
-void
-ts_license_on_assign(const char *newval, void *extra)
+TSDLLEXPORT void
```
erimatnor, line 107: Why does this need to be exported? I see it called from TSL, which seems really strange given that loading must already be enabled when called from TSL, otherwise the library wouldn't be loaded. 

I realize this was there before, but the whole thing seems quite intuitive.

src/license_guc.c
```diff
@@ -71,315 +43,218 @@ current_license_can_downgrade_to_apache(void)
...
-		default:
-			elog(ERROR, "Invalid license key '%s'", ts_guc_license_key);
-			pg_unreachable();
+		case LICENSE_UNDEF:
+			return false;
+	}
+
+	if (!load_enabled)
+	{
+		load_source = source;
```
erimatnor, line 222: Shouldn't we use `extras` instead of global variables? At least the comment above seems to imply that we should limit side-effect to extras. The GUC docs says one can `malloc` data into extras (note, not `palloc`) and it will be passed on the the assign hook and freed automatically when not needed.
 
Also, why do we need to pass on the source? We've already validated it here, so we no longer need it. I guess that's why the assign hook doesn't even have it as a parameter.

src/license_guc.c
```diff
@@ -71,315 +43,218 @@ current_license_can_downgrade_to_apache(void)
...
 	{
-		case LICENSE_TYPE_APACHE_ONLY:
-			edition = TS_LICENSE_APACHE_ONLY_PRINT_TEXT;
-			break;
-		case LICENSE_TYPE_COMMUNITY:
-			edition = TS_LICENSE_COMMUNITY_PRINT_TEXT;
+		case LICENSE_APACHE:
+			/* Allow setting Apache license only if is is set from postgresql.conf
+			 * or the server command line */
+			if (source == PGC_S_FILE || source == PGC_S_ARGV)
```
erimatnor, line 205: I am wondering if it doesn't make sense for us to support other sources here as well. For instance, per-database or env variables seem reasonable. Less important, but maybe useful is per-user settings (e.g., someone in a multi-user system wants to restrict features on a per-user basis). 

FWIW, the docs for GUCs says that anything less than `PGC_S_INTERACTIVE` is a non-interactive setting. So, maybe we should just allow anything below that?

src/license_guc.c
```diff
@@ -71,315 +43,218 @@ current_license_can_downgrade_to_apache(void)
...
+	Assert(newval != NULL);
+	if (license_type_of(newval) == LICENSE_APACHE)
+	{
+		Assert(license_is_downgradable());
+		Assert(extra == NULL);
+
+		tsl_module_shutdown();
+		return;
+	}
+
```
erimatnor, line 258: Assert LICENSE_TSL here?

src/license_guc.c
```diff
@@ -71,315 +43,218 @@ current_license_can_downgrade_to_apache(void)
...
-			elog(ERROR, "Invalid license key '%s'", ts_guc_license_key);
-			pg_unreachable();
+		case LICENSE_UNDEF:
+			return false;
+	}
+
+	if (!load_enabled)
+	{
+		load_source = source;
+		return true;
```
erimatnor, line 223: Why do we return `true` here even though load is not enabled? Feels like it should be a failure, or is the loading deferred somehow?

I guess I am wondering what the difference is between returning `true` here and returning `true` below (at the end of the function).

test/sql/gapfill.sql
```diff
@@ -1,21 +0,0 @@
...
-\set ECHO queries
-\set VERBOSITY default
-
-\c :TEST_DBNAME :ROLE_SUPERUSER
-
-SELECT allow_downgrade_to_apache();
-SET timescaledb.license_key='ApacheOnly';
-
-\set ON_ERROR_STOP 0
-SELECT locf(1);
```
erimatnor, line 18: Why don't we want this test anymore? Was it moved to another file?

I guess it tests that gapfill doesn't work under Apache license?

test/expected/edition.out
```diff
@@ -14,63 +32,39 @@ SELECT _timescaledb_internal.tsl_loaded();
...
+ timescaledb.license 
+---------------------
+ tsl
 (1 row)
 
-SELECT * FROM  _timescaledb_internal.enterprise_enabled();
- enterprise_enabled 
---------------------
- f
+SHOW timescaledb.license;
```
erimatnor, line 64: Same statement as above. Remove?

test/expected/edition.out
```diff
@@ -14,63 +32,39 @@ SELECT _timescaledb_internal.tsl_loaded();
...
+ timescaledb.license 
+---------------------
+ apache
 (1 row)
 
-select * from timescaledb_information.license;
-  edition  | expired | expiration_time 
------------+---------+-----------------
- community | f       | infinity
+SELECT set_license('tsl');
```
erimatnor, line 52: What is this function testing, as opposed to just `SET timescaledb.license =`?

Looks like it is testing database-level settings and session-level settings at once, but I don't get what I should expect to happen. Seems it just sets the license and it works.

I guess a more interesting use case is to have only database-level setting so that is the source of the variable.

test/expected/edition.out
```diff
@@ -79,36 +73,52 @@ SELECT * FROM  _timescaledb_internal.tsl_loaded();
  t
 (1 row)
 
-SELECT * FROM  _timescaledb_internal.license_expiration_time();
- license_expiration_time 
--------------------------
- infinity
+SET timescaledb.license='tsl';
+ERROR:  permission denied to set parameter "timescaledb.license"
```
erimatnor, line 77: Why is this an error? I don't see a switch from superuser to unprivileged user, so why does it generate an error?

test/expected/edition.out
```diff
@@ -79,36 +73,52 @@ SELECT * FROM  _timescaledb_internal.tsl_loaded();
...
+SHOW timescaledb.license;
+ timescaledb.license 
+---------------------
+ tsl
 (1 row)
 
-SELECT * FROM  _timescaledb_internal.print_license_expiration_info();
- print_license_expiration_info 
--------------------------------
+SELECT _timescaledb_internal.tsl_loaded();
```
erimatnor, line 84: This was called just above, and nothing changed so why show?

test/expected/edition.out
```diff
@@ -79,36 +73,52 @@ SELECT * FROM  _timescaledb_internal.tsl_loaded();
...
+ timescaledb.license 
+---------------------
+ tsl
 (1 row)
 
-SELECT get_sqlstate($$SELECT _timescaledb_internal.current_db_set_license_key('ApacheOnly')$$);
- get_sqlstate 
---------------
- 42501
+SELECT _timescaledb_internal.tsl_loaded();
```
erimatnor, line 113: Maybe call this function when TSL is _not_ loaded above to see a false result?

erimatnor (commented):
Some things are quite hard to understand, so I mostly have questions and comments at this point.

test/sql/gapfill.sql
pmwkaa: We do, this moved to license test

src/license_guc.c
pmwkaa: I think this is to allow delayed loading to work. Basically it saves the GucSource value and then retries it later

src/license_guc.c
```diff
@@ -71,315 +43,218 @@ current_license_can_downgrade_to_apache(void)
...
-		return true;
-	}
+	return license_type_of(ts_guc_license) == LICENSE_APACHE;
+}
 
-	if (!try_to_load_tsl)
+static bool
+license_is_downgradable(void)
+{
+	if (downgrade_enabled)
```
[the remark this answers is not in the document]
pmwkaa: Removed downgrade completely

src/license_guc.c
```diff
@@ -71,315 +43,218 @@ current_license_can_downgrade_to_apache(void)
...
-	Assert(extra == NULL);
-	extra = revalidate_license(newval);
-#endif
-	Assert(extra != NULL);
-	ts_cm_functions->tsl_license_on_assign(newval, extra);
-}
+	load_enabled = true;
+
+	/* re-set the license to actually load the submodule if needed */
+	result = set_config_option("timescaledb.license",
```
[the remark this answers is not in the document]
pmwkaa: Yes, this is for delayed execution. I guess it is up to the hook to validate the value once again

src/license_guc.c
pmwkaa: This is for different purpose I believe. This gucsource will be used in delayed call again wit the same value. So it is not used pass it between hooks.

src/license_guc.c
pmwkaa: I guess we could, not sure if it is worth it. But I had to add S_DEFAULT otherwise it would not allow to set default value on the extension load.

erimatnor (approved):

src/cross_module_fn.c
```diff
@@ -127,21 +127,8 @@ error_no_default_fn_community(void)
...
-static void
-error_no_default_fn_enterprise(void)
-{
-	ereport(ERROR,
-			(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-			 errmsg("functionality not supported under the current license \"%s\", license",
-					ts_guc_license_key),
-			 errhint("Request a trial license to try this feature for free or contact us for more "
-					 "information at https://www.timescale.com/pricing")));
+			 errmsg("functionality not supported under the current \"%s\" license", ts_guc_license),
```
svenklemm, line 130: Hmm you are changing the text here and the current binary might not be upgradable when compiled with APACHE_ONLY, maybe switch text depending on whether compiled with APACHE_ONLY

src/license_guc.h
```diff
@@ -13,61 +13,22 @@
...
-{
-	LICENSE_TYPE_APACHE_ONLY = 'A',
-	LICENSE_TYPE_COMMUNITY = 'C',
-	LICENSE_TYPE_ENTERPRISE = 'E',
-} LicenseType;
-
-#define TS_APACHE_ONLY_LICENSE "ApacheOnly"
-#define TS_COMMUNITY_LICENSE "CommunityLicense"
+#define TS_LICENSE_APACHE "apache"
+#define TS_LICENSE_TSL "tsl"
```
svenklemm, line 17: why not use apache and community here that's we call them externally?

pmwkaa: I am fine with community. I think we just need to discuss and decide which name we would like to use

k-rus (commented):
Can you include into the commit message that it fixes #2359?

tsl/test/sql/license.sql
```diff
@@ -0,0 +1,8 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+
+\c :TEST_DBNAME :ROLE_CLUSTER_SUPERUSER;
+
+SHOW timescaledb.license;
+SELECT _timescaledb_internal.tsl_loaded();
```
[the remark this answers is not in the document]
pmwkaa: We have the same test there

test/expected/license.out
```diff
@@ -0,0 +1,36 @@
...
+SELECT _timescaledb_internal.tsl_loaded();
+ tsl_loaded 
+------------
+ f
+(1 row)
+
+SET timescaledb.license='apache';
+ERROR:  invalid value for parameter "timescaledb.license": "apache"
+DETAIL:  Cannot change a license in a running session.
+HINT:  Change the license in the configuration file or server command line.
```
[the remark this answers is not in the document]
pmwkaa: Unfortunately it is generated by PostgreSQL itself when check hook function returns false. At least we can provide details

src/cross_module_fn.c
pmwkaa: If it was compiled with APACHE_ONLY this will also change the DEFAULT_LICENSE to apache. This value will be represented in ts_guc_license as well

mkindahl (approved):

src/license_guc.c
```diff
@@ -107,279 +101,99 @@ ts_license_enable_module_loading(void)
...
+	/* Allow setting a license only if is is set from postgresql.conf
+	 * or the server command line */
+	switch (type)
 	{
-		Assert(tsl_startup_fn != NULL);
+		case LICENSE_APACHE:
+		case LICENSE_TSL:
+			if (source == PGC_S_FILE || source == PGC_S_ARGV || source == PGC_S_DEFAULT)
+				break;
+			GUC_check_errdetail("Cannot change a license in a running session.");
```
k-rus, line 160: Can you provide detail from which value to which the change requested?

src/license_guc.h
```diff
@@ -13,61 +13,22 @@
...
-
-/* Each of these functions takes a LicenseUpdateExtra for their extra param */
-extern bool ts_license_update_check(char **newval, void **extra, GucSource source);
-extern void ts_license_on_assign(const char *newval, void *extra);
+extern bool ts_license_guc_check_hook(char **newval, void **extra, GucSource source);
+extern void ts_license_guc_assign_hook(const char *newval, void *extra);
 
-extern void TSDLLEXPORT ts_license_enable_module_loading(void);
+extern TSDLLEXPORT void ts_license_enable_module_loading(void);
+extern TSDLLEXPORT bool ts_license_is_apache(void);
```
k-rus, line 32: Are they called from TSL module?

test/expected/license.out
```diff
@@ -0,0 +1,36 @@
...
+SET timescaledb.license='tsl';
+ERROR:  invalid value for parameter "timescaledb.license": "tsl"
+DETAIL:  Cannot change a license in a running session.
+HINT:  Change the license in the configuration file or server command line.
+SET timescaledb.license='something_else';
+ERROR:  invalid value for parameter "timescaledb.license": "something_else"
```
k-rus, line 27: I suggest to add a hint and print acceptable values, since it might be a typo or user tried to use Community.

tsl/src/README.module.md
```diff
@@ -0,0 +1,34 @@
+# Submodule Licensing and Initialization #
+
+## Loading and Activation ##
+
+We link module loading and activation to the license GUC itself.
+While keeping multiple GUCs, e.g., one for module loading and one licensing,
```
k-rus, line 6:
```suggestion
While keeping multiple GUCs, e.g., one for module loading and one for licensing,
```

tsl/src/README.module.md
```diff
@@ -0,0 +1,34 @@
...
+is tempting, it leads to hairy edge cases when the GUCs disagree.
```
k-rus, line 7: Is it necessary to have this sentence discussing what is not done? May be just focus what is available?

k-rus (commented):
I noticed that one error message is not tested. Can you add a test? I believe the error message will look weird, something like:
> the current "tsl" license

while manually written errors contain something like:
> license: TSL

Also `tsl license` is a tautology. So original order `license: tsl` might be better.

See other comments too.

src/license_guc.h
pmwkaa: Only this one ts_license_enable_module_loading(). Removed export from the apache check function

src/license_guc.c
pmwkaa: Not sure this is relevant here because it can't change any license runtime

k-rus: I agree that it is not relevant. Is there a test, which runs into this error? It will be more clear if it is already good enough by looking to the actual scenarios when it happens.

pmwkaa:
@mkindahl @erimatnor @k-rus I actually agree with @k-rus about the `TSL license` being a tautology. I used the new name `timescale` this time, which is much more cleaner now.

Basically license guc can be set either to `apache` or `timescale` now.

src/cross_module_fn.c
```diff
@@ -127,21 +127,8 @@ error_no_default_fn_community(void)
...
+			 errhint("Upgrade your license to TSL to use this free community feature.")));
```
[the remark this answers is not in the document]
pmwkaa: I actually renamed tsl to timescale

src/license_guc.c
pmwkaa: Yes, it is tested in test/sql/license.sql

k-rus:
> Basically license guc can be set either to `apache` or `timescale` now.

@pmwkaa does it look good?:
```sql
SET timescale.license = 'Timescale';
```
I don't know myself.

src/license_guc.c
k-rus: @pmwkaa Is there a test, which gets this error?

pmwkaa:
> > Basically license guc can be set either to `apache` or `timescale` now.
> 
> @pmwkaa does it look good?:
> 
> ```sql
> SET timescale.license = 'Timescale';
> ```
> 
> I don't know myself.

It does for me at least. We can use `community` instead but I think this not accurate because we have `Timescale License` `tsl/LICENSE-TIMESCALE`. I think `community` edition was relevant when there were `enterprise` edition as well.

src/license_guc.c
pmwkaa: https://github.com/timescale/timescaledb/pull/2435/files#diff-a1388d76a54b03a1aea1f9c9b4fe3749R20

k-rus: Thank you for the link! Looks good!

erimatnor:
> > > Basically license guc can be set either to `apache` or `timescale` now.
> > 
> > 
> > @pmwkaa does it look good?:
> > ```sql
> > SET timescale.license = 'Timescale';
> > ```
> > 
> > 
> > I don't know myself.
> 
> It does for me at least. We can use `community` instead but I think this not accurate because we have `Timescale License` `tsl/LICENSE-TIMESCALE`. I think `community` edition was relevant when there were `enterprise` edition as well.

I'm OK with "Timescale". I am not fond of "community" since that referred to the (now defunct) license edition, which made more sense when there was also an "enterprise" edition. In fact, we've never had a "community" license, so that was a bit confusing already prior to this change.

src/license_guc.c
```diff
@@ -15,90 +15,84 @@
...
+	return LICENSE_UNDEF;
+}
+
+bool
+ts_license_is_apache(void)
+{
+	return license_type_of(ts_guc_license) == LICENSE_APACHE;
+}
+
 TSDLLEXPORT void
```
k-rus, line 83: I suggest to remove `TSDLLEXPORT` for consistency with the rest of the code. Sorry for the late comment about this, I was waiting for the response to Erik's comment.
```suggestion
void
```

src/license_guc.c
```diff
@@ -107,279 +101,100 @@ ts_license_enable_module_loading(void)
...
-			break;
-		case LICENSE_TYPE_ENTERPRISE:
-			edition = TS_LICENSE_ENTERPRISE_PRINT_TEXT;
-			break;
-		default:
-			elog(ERROR, "Invalid license key '%s'", ts_guc_license_key);
-			pg_unreachable();
+		GUC_check_errdetail("Could not find TSL timescaledb module.");
+		GUC_check_errhint("Check that \"%s\" is available.", EXTENSION_TSL_SO);
+		return false;
```
k-rus, line 182: According to Codecov this error condition is not tested.

test/sql/license.sql
```diff
@@ -0,0 +1,42 @@
...
+DROP FUNCTION custom_func;
+
+CREATE TABLE metrics(time timestamptz NOT NULL, value float);
+SELECT create_hypertable('metrics', 'time');
+ALTER TABLE metrics SET (timescaledb.compress);
+DROP TABLE metrics;
+
+\set ON_ERROR_STOP 1
+
+
```
k-rus, line 42: Extra new line can be removed:
```suggestion

```

tsl/src/README.module.md
```diff
@@ -0,0 +1,32 @@
...
+To enable binaries which only contain Apache-Licensed code,
+we dynamically link in Timescale-Licensed code on license activation,
+and handle all function calls into the module via function pointers.
+
+The registry in `ts_cm_functions` of type `CrossModuleFunctions` (declared in
+[`cross_module_fn.h`](/src/cross_module_fn.h) and defined in
+[`cross_module_fn.c`](/src/cross_module_fn.c)) stores all of the cross-module
+functions. To add a new cross-module function you must:
+  - Add a struct member `CrossModuleFunctions.<function name>`.
+  - Add default function to `ts_cm_functions_default` that will be called from the Apache version, usually this function should just call `error_no_default_fn`. **NOTE** Due to function-pointer casting rules, the default function must have the exact same signature as the function pointer; you may _not_ cast another function pointer of another type.
```
k-rus, line 29: Can you split into multiple lines?

k-rus (changes requested):
I checked that `appveyor.yml` needs to be fixed accordantly. It still uses `timescaledb.license_key`

src/license_guc.c
pmwkaa: This function is used in TSL module, so keeping this export macro. I removed one from the function above

src/license_guc.c
pmwkaa: Not sure it is possible to test it now

pmwkaa:
> I checked that `appveyor.yml` needs to be fixed accordantly. It still uses `timescaledb.license_key`

No it does not

tsl/src/README.module.md
pmwkaa: Moved it a bit, but not sure it makes much sense since this is a markdown file which might be reinterpreted by parser differently.

k-rus:
> > I checked that `appveyor.yml` needs to be fixed accordantly. It still uses `timescaledb.license_key`
> 
> No it does not. I had to put test/sql/license into ignore for now, because windows build uses separate configuration file. All tests are tested under timescale license for now in order to pass in appveyor.

I missed that you already removed it 🤦 It was a mistake for me to request changes.

k-rus (approved):
LGTM
It would be nice to run `license` test instead of ignoring. E.g., by:
```
docker exec -it pgregress /bin/bash -c "cp C:\Program Files\postgresql\12\data\postgresql.conf C:\Program Files\postgresql\12\data\postgresql.conf.tsl"
Add-Content "C:\Program Files\postgresql\12\data\postgresql.conf" "timescaledb.license = 'apache'"
... # execute Apache tests
docker exec -it pgregress /bin/bash -c "mv C:\Program Files\postgresql\12\data\postgresql.conf.tsl C:\Program Files\postgresql\12\data\postgresql.conf"
```

PR #2439, opened 2020-09-24
Allow freezing a continuous aggregate

### LINKED ISSUE #2135: Make sure refreshing a continuous aggregate is governed by user privileges
Refreshing a continuous aggregate translates to either an INSERT, UPDATE, or DELETE or a mix thereof on the underlying materialized hypertable.

Thus, there's an opportunity to limit certain types of refreshes to specific users through the corresponding privileges to do those operations. For instance, we'd like certain users to be unable to refresh a continuous aggregate if that refresh translates to deleting certain areas of the aggregate data. Only a user with INSERT, UPDATE, and DELETE privileges should be able to refresh a continuous aggregate under any circumstance.

Thus, to grant permissions to such a user one would do:
```
GRANT INSERT, UPDATE, DELETE ON daily_temp TO cagg_admin;
```
and this would translate to setting the corresponding privileges on the internal materialized hypertable.

erimatnor:
Continuous aggregates can now be frozen across a given window, thus
protecting the aggregate from refreshes that overlap with the frozen
window. Similarly, a frozen continuous aggregate can be unfrozen to
allow refreshes once again.

Two new API procedures are introduced:

```
CALL freeze_continuous_aggregate('daily_temp', '2020-01-01', '2020-02-01');
CALL unfreeze_continuous_aggregate('daily_temp', '2020-01-01', '2020-02-01');
```

The new API calls are procedures for consistency with
`refresh_continuous_aggregate`, even though they don't need
transactional control. The new procedures are also created in the
internal namespace to signal that this isn't an official API yet.

Closes #2135


tsl/src/continuous_aggs/common.c
```diff
@@ -0,0 +1,143 @@
...
+
+/*
+ * Compute the largest possible bucketed window given the time type and
+ * internal restrictions.
+ *
+ * The largest bucketed window is governed by restrictions set by the type and
+ * internal, TimescaleDB-specific legacy details (see get_max_window above for
+ * further explanation).
+ */
+static InternalTimeRange
```
erimatnor, line 26: Code moved from `refresh.c`

tsl/src/continuous_aggs/invalidation.c
```diff
@@ -907,10 +967,133 @@ invalidation_process_cagg_log(const ContinuousAgg *cagg, const InternalTimeRange
...
+									mergedentry.greatest_modified_value,
+									mergedentry.frozen);
+	ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
+	ts_catalog_insert_only(state.cagg_log_rel, tuple);
+	ts_catalog_restore_user(&sec_ctx);
+	heap_freetuple(tuple);
+	invalidation_state_cleanup(&state);
+}
+
+void
```
erimatnor, line 1074: code moved

erimatnor:
Closing to revisit later.

PR #2441, opened 2020-09-24
Remove unused refresh arguments from invalidation

k-rus:
After recent refactoring the refresh window arguments are not used in
many invalidation functions. This fixes the code and removes unused
arguments to avoid future confusions.


erimatnor (approved):
Was just about to fix this myself :) Great that I don't have to.

pmwkaa (approved):

k-rus:
> Was just about to fix this myself :) Great that I don't have to.

It was confusing me, so I decided to bring it outside my PR. I might need to do another refactoring change, so it is good to have it as a basis.

mkindahl (approved):

PR #2445, opened 2020-09-24
Make update test consistent

svenklemm:
This patch changes the update test to use the same checks
between clean / updated install and dumped/restored install.
Previously only a small subset of the checks would be run against
the updated instance and most of the tests would only run against
the dumped and restored container.

Needs #2249 merged to pass tests


test/sql/updates/setup.continuous_aggs.sql
```diff
@@ -122,6 +122,12 @@ BEGIN
     PERFORM add_continuous_aggregate_policy('mat_before', NULL, '-30 days'::interval, '336 h');
 
   END IF;
+
+  IF ts_version >= '2.0.0' THEN
```
pmwkaa, line 126: Is the comparison actually works as intended?

pmwkaa (approved):

gayyappan (approved):

PR #2449, opened 2020-09-25
Set autovacuum_enabled to false for compressed chunks

svenklemm:
Commit 8e1e6036 changed chunk compression to disable autovacuum
on compressed chunks but did not apply the setting to chunks
compressed before that change. So this patch changes chunks
compressed with previous version to disable autovacuum as well.


WireBaron (approved):
I'm a little mixed on this change.  The reason we turned off autovacuum was to keep the pg_stats we save for the compressed chunk from being overwritten.  However, we never saved these stats for chunks written prior to this change, so there's really no point in disabling autovacuum for them.  However, I'm not sure there's any harm in doing this, and I can see it being necessary to get the update test working properly.

svenklemm:
I agree we need to also update stats for previously compressed chunks, otherwise our row estimate functions won't work for them but thats outside scope of this PR

erimatnor (approved):

pmwkaa (approved):

PR #2465, opened 2020-09-29
Create low end invalidation when updating caggs

### LINKED ISSUE #2450: Create invalidation from (-inf, ignore_invalidation) when upgrading caggs to 2.0
Prior to 2.0, we do not record invalidations prior to the lower bound (ignore_invalidations_older_than) of a continuous aggregate.  In 2.0 we assume that the invalidation log for an aggregate is a complete record of where the aggregate is up to date.  This means if a user requests a cagg update of a region where a write occurred behind a cagg in 1.7, we will not properly update the cagg.

WireBaron:
This change will add an invalidation to the
materialization_invalidation_log for any region earlier than the
ignore_invalidation_older_than parameter when updating a continuous
aggregate to 2.0. This is needed as we do not record invalidations
in this region prior to 2.0 and there is no way to ensure the
aggregate is up to date within this range.

Fixes #2450

### DIFF
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -215,28 +215,23 @@ SET
   proc_schema = '_timescaledb_internal',
   proc_name = 'policy_refresh_continuous_aggregate',
   job_type = 'custom',
-  config = jsonb_build_object('mat_hypertable_id', cagg.mat_hypertable_id, 'start_interval', 
-   CASE WHEN cagg.ignore_invalidation_older_than IS NULL 
-      THEN NULL 
-      WHEN cagg.ignore_invalidation_older_than = 9223372036854775807 
-         AND  
-            ts_tmp_get_time_type( cagg.raw_hypertable_id ) IN  ('TIMESTAMP'::regtype, 'DATE'::regtype, 'TIMESTAMPTZ'::regtype) 
-      THEN NULL 
-      ELSE
-         CASE WHEN 
-            ts_tmp_get_time_type( cagg.raw_hypertable_id ) IN  ('TIMESTAMP'::regtype, 'DATE'::regtype, 'TIMESTAMPTZ'::regtype) 
-            THEN ts_tmp_get_interval(cagg.ignore_invalidation_older_than)::TEXT 
-            ELSE cagg.ignore_invalidation_older_than::TEXT 
-         END
-   END,
-   'end_interval', 
-   CASE
-      WHEN 
-            ts_tmp_get_time_type( cagg.raw_hypertable_id ) IN  ('TIMESTAMP'::regtype, 'DATE'::regtype, 'TIMESTAMPTZ'::regtype) 
-      THEN ts_tmp_get_interval(cagg.refresh_lag)::TEXT 
-      ELSE cagg.refresh_lag::TEXT 
-   END
-),
+  config = 
+    CASE WHEN ts_tmp_get_time_type( cagg.raw_hypertable_id ) IN  ('TIMESTAMP'::regtype, 'DATE'::regtype, 'TIMESTAMPTZ'::regtype) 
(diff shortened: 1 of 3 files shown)


mkindahl (approved):

pmwkaa (approved):

sql/updates/latest-dev.sql
```diff
@@ -309,8 +309,19 @@ SELECT materialization_id, BIGINT '-9223372036854775808', watermark, 92233720368
...
 INSERT INTO _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
-SELECT (SELECT mat_hypertable_id FROM _timescaledb_catalog.continuous_agg EXCEPT SELECT materialization_id FROM _timescaledb_catalog.continuous_aggs_completed_threshold), 
--9223372036854775808, -9223372036854775808, 9223372036854775807;
+SELECT unrun_cagg.id, -9223372036854775808, -9223372036854775808, 9223372036854775807 FROM
+(SELECT mat_hypertable_id id FROM _timescaledb_catalog.continuous_agg EXCEPT SELECT materialization_id FROM _timescaledb_catalog.continuous_aggs_completed_threshold) as unrun_cagg;
+
+-- Also add an invalidation from [-infinity, ignore_invaliation_older_than] to cover any missed invalidations
+-- For NULL or inf ignore_invalidations_older_than, use julian 0 for consistency with 2.0
+INSERT INTO _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
+SELECT cagg.mat_hypertable_id, BIGINT '-9223372036854775808', BIGINT '-9223372036854775808',
```
gayyappan, line 318: I think this should be added only if ignore_invalidation_older_than is not the default value and not the other case.
Could you explain how we get the value at line 321?  ignore_invalidation_older_than could be set for an integer table. In this case what does it mean to subtract from now() ( line 322)?

test/sql/updates/post.continuous_aggs.v2.sql
```diff
@@ -16,6 +16,9 @@ SELECT view_name, schedule_interval, materialized_only, materialization_hypertab
 
 SELECT maxtemp FROM mat_ignoreinval ORDER BY 1;
 
+SELECT materialization_id FROM _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
```
svenklemm, line 19: should probably add an ORDER BY here

test/sql/updates/post.continuous_aggs.v2.sql
```diff
@@ -16,6 +16,10 @@ SELECT view_name, schedule_interval, materialized_only, materialization_hypertab
 
 SELECT maxtemp FROM mat_ignoreinval ORDER BY 1;
 
+-- Note that greatest_modified value for tables with an 'ignore_validation_other_than' won't match
```
gayyappan, line 19: Could you elaborate on this comment?

sql/updates/latest-dev.sql
WireBaron: This is the invalidation that 2.0 adds to every new cagg by default.  I'm not entirely sure what the reasoning behind this was, but I figured it would be better to match the 2.0 invalidation here.  I'm not sure what you're asking about the subtraction from now(), as that only happens in the other case.

test/sql/updates/post.continuous_aggs.v2.sql
WireBaron: Done

test/sql/updates/post.continuous_aggs.v2.sql
WireBaron: I was trying to improve this check a bit, but it turned out to not be possible, so that change and this comment have been removed.

gayyappan (approved):

PR #2467, opened 2020-09-29
Use postgres implementation of estimate_hashagg_tablesize

svenklemm:
Only use our own implementation of estimate_hashagg_tablesize on
versions where the postgres implementation is static.

### DIFF
--- a/src/compat.h
+++ b/src/compat.h
@@ -295,6 +295,14 @@
 #define create_append_path_compat create_append_path
 #endif
 
+/*
+ * estimate_hashagg_tablesize is a static function in PG11 and earlier, so we map
+ * to our own copy when it's not available.
+ */
+#if PG11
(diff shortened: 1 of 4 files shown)


WireBaron (approved):

src/compat.h
```diff
@@ -295,6 +295,14 @@
 #define create_append_path_compat create_append_path
 #endif
 
+/*
+ * estimate_hashagg_tablesize is a static function in PG11 and earlier so we map
+ * to our own copy when it's not available
```
k-rus, line 300:
```suggestion
 * estimate_hashagg_tablesize is a static function in PG11 and earlier, so we map
 * to our own copy when it's not available.
```

k-rus (approved):
LGTM

pmwkaa (approved):

PR #2470, opened 2020-09-30
Test Apache license on Windows

k-rus:
Fixes the regression test of Apache license on Windows.

### PR note
GH is confused by some reason and shows difference on number of unchanged lines.

#2435 removed the Apache license test on Windows. This PR fixes it back.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -122,6 +122,8 @@ build_script:
     Add-Content "C:\Program Files\postgresql\12\data\postgresql.conf" "timescaledb.ssl_dir='C:/projects/timescaledb/build/tsl/test'"
     Add-Content "C:\Program Files\postgresql\12\data\postgresql.conf" "timescaledb.passfile='C:/projects/timescaledb/build/tsl/test/pgpass.conf'"
 
+    Add-Content "C:\Program Files\postgresql\12\data\postgresql.conf" "timescaledb.license = 'apache'"
+
     # Add-Content "C:\Program Files\postgresql\12\data\postgresql.conf" "log_min_messages='debug5'"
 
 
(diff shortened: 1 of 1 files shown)


pmwkaa:
@k-rus Funny that I've tried exact same approach, did not worked for some reason, probably missed something. Hope it will now.

The changed lines might be to the reason that Windows text files are usually use different end of line chars (`\r\n` vs `\n`). Maybe they have changed accidentally?

erimatnor (approved):
There's a lot of lines that seem unrelated to the changes in this PR, which makes it harder to review. I actually only identified three lines that seem relevant. Would be good to keep formatting as a separate change.

k-rus:
> There's a lot of lines that seem unrelated to the changes in this PR, which makes it harder to review. I actually only identified three lines that seem relevant. Would be good to keep formatting as a separate change.

@erimatnor See my not in https://github.com/timescale/timescaledb/pull/2470#issue-495473572
I haven't done any changes to these lines. It might be due to git settings, I am checking it.

mkindahl (approved):
Looks good, but would be good to just have changed lines. There is no difference between some of the changed lines, so I suspect that it is a change in line ending as a result of the editor being helpful.

pmwkaa (approved):
Apparently license tests passed, so this change makes sense. Yet there are some odd issues with other tests, likely not related to this PR

k-rus:
@pmwkaa @erimatnor @mkindahl I made a comment in the OP of the PR regarding unrelated lines. I guess nobody reads it:
>GH is confused by some reason and shows difference on number of unchanged lines.

One explanation can be that git silently fixes the line endings. I have not set the configuration, so it is default behaviour.

k-rus:
> Apparently license tests passed, so this change makes sense. Yet there are some odd issues with other tests, likely not related to this PR

@pmwkaa You are right the failures are not related to the PR, so I created an issue for the deterministic failures.



> Funny that I've tried exact same approach, did not worked for some reason, probably missed something. Hope it will now.

You might not put in exactly the same place. It needs to be before the PG service is started or restarted.

PR #2471, opened 2020-09-30
Prevent run_job of procedure inside transaction block

svenklemm:


tsl/src/bgw_policy/job.c
```diff
@@ -399,6 +399,9 @@ job_execute(BgwJob *job)
 
 	prokind = get_func_prokind(proc);
 
+	if (prokind == PROKIND_PROCEDURE)
+		PreventInTransactionBlock(true, "run_job");
```
erimatnor, line 403: I don't understand this. 

- Why do we prevent this at execution time rather than when adding the job? It seems you can now add a job that will always fail.
- Is it generally the case that procedures cannot run in transaction blocks? My impression is that it is up to the implementation and that the procedure itself should prevent it rather than us preventing it across the board for all procedures.
- Why do we say `run_job` here when this function is `job_execute`?
- If this is specifically about `run_job`, shouldn't this prevention happen in that function?

svenklemm: No the job will only fail when you try to execute it with run_job inside a transaction outside of transaction it works without problem. Procedures that don't do their own transaction control can be executed inside transaction as well but procedures that do their own transaction handling will hit an assertion inside postgres when run with run_job and executed inside transaction block.

erimatnor: I still don't understand why we're preventing transaction blocks based on the executed function. The function that is executed by `run_job` should have its own `PreventInTransactionBlock` if it doesn't allow running in a block. 

If this is only about preventing `run_job` from executing in a transaction block, why is it conditional on the prokind of the function/procedure we execute?

k-rus (approved):
Can you explain in the commit message, what it is the issue, which is prevented by this commit?

PR #2475, opened 2020-10-01
Fix lost notice in cagg test

k-rus:
PR #2460 made changes to continuous aggregates refresh, which resulted
in a notice in continuous_aggs_policy test expected result. Later this
notice disappeared and this commit fixes it back.

### DIFF
--- a/tsl/test/expected/continuous_aggs_policy.out
+++ b/tsl/test/expected/continuous_aggs_policy.out
@@ -163,6 +163,7 @@ CREATE MATERIALIZED VIEW max_mat_view_timestamp
         GROUP BY 1 WITH NO DATA;
 SELECT add_continuous_aggregate_policy('max_mat_view_timestamp', '10 day', '1 h'::interval , '1 h'::interval) as job_id \gset
 CALL run_job(:job_id);
+NOTICE:  continuous aggregate "max_mat_view_timestamp" is already up-to-date
 SELECT config FROM _timescaledb_config.bgw_job
 WHERE id = :job_id;
                                        config                                        


mkindahl (approved):

tsl/test/expected/continuous_aggs_policy.out
```diff
@@ -163,6 +163,7 @@ CREATE MATERIALIZED VIEW max_mat_view_timestamp
         GROUP BY 1 WITH NO DATA;
 SELECT add_continuous_aggregate_policy('max_mat_view_timestamp', '10 day', '1 h'::interval , '1 h'::interval) as job_id \gset
 CALL run_job(:job_id);
+NOTICE:  continuous aggregate "max_mat_view_timestamp" is already up-to-date
```
erimatnor, line 166: How  could the previous PR have passed tests without this output?

k-rus: @erimatnor it was there, see https://github.com/timescale/timescaledb/pull/2460/files#diff-89a5e070f0ba4c1e9ee0a467d26819d7
I don't understand how suddenly it disappeared from `master` branch. I wasn't able to find out the answer.

erimatnor: No, that's a different line number: 149, while this is line 166

k-rus: @erimatnor Thank you! then it might be flaky test and it has this notice today, but will not be tomorrow, I guess. I need to investigate more.

erimatnor: I guess this test is inherently flaky since this is a timestamp table which does not have a test-specific integer-now function and instead uses system clock time internally to calculate the timestamp for refresh.

k-rus: @erimatnor Should we ignore this test for now, so it doesn't block PRs?

PR #2477, opened 2020-10-01
Ignore flaky test for now

k-rus:
continuous_aggs_policy is flaky and prevents to proceed with current
PRs. This commit ignores it until it is fixed.

##PR note:
GH has problem with apprveyor.yml and shows diff for no changes.

### DIFF
--- a/.github/workflows/alpine-32bit-build-and-test.yaml
+++ b/.github/workflows/alpine-32bit-build-and-test.yaml
@@ -22,9 +22,9 @@ jobs:
         build_type: [ Debug ]
         include:
           - pg: 11.9
-            ignores: append-11 chunk_adaptive-11 continuous_aggs_bgw_drop_chunks remote_txn transparent_decompression-11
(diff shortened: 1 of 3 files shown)


erimatnor:
My guess is that there's a windows vs. unix line ending diff in Appveyor.yml. It should be possible to fix in that case.

k-rus:
> My guess is that there's a windows vs. unix line ending diff in Appveyor.yml. It should be possible to fix in that case.

@erimatnor do you mean that the diff in the line endings confuses GH?

mkindahl (approved):
Not entirely sure that it's a good idea to ignore an error is such a critical test, but approving since it's hard to get PRs in otherwise.

pmwkaa (approved):

k-rus:
> My guess is that there's a windows vs. unix line ending diff in Appveyor.yml. It should be possible to fix in that case.

Actually, I think it was git who was fixing line endings.

PR #2482, opened 2020-10-02
Improve check for fallthrough support

svenklemm:
Change the check for fallthrough support to check for the attribute
instead of specific compiler versions.

### DIFF
--- a/src/annotations.h
+++ b/src/annotations.h
@@ -6,18 +6,14 @@
 #ifndef TIMESCALEDB_ANNOTATIONS_H
 #define TIMESCALEDB_ANNOTATIONS_H
 
-/* Fall-through annotation */
-#if defined(__clang__)
-#if (__clang_major__ >= 12) || (__clang_analyzer__)
+/* Supported since clang 12 and GCC 7 */
+#if defined __has_attribute
+#if __has_attribute(fallthrough)
 #define TS_FALLTHROUGH __attribute__((fallthrough))
 #else
 #define TS_FALLTHROUGH /* FALLTHROUGH */
-#endif				   /* __clang__ */
-#elif defined(__GNUC__)
-/* Supported since GCC 7 */
-#define TS_FALLTHROUGH __attribute__((fallthrough))
+#endif
 #else
-/*  MSVC and other compilers */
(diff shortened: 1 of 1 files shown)


src/annotations.h
```diff
@@ -6,18 +6,15 @@
 #ifndef TIMESCALEDB_ANNOTATIONS_H
 #define TIMESCALEDB_ANNOTATIONS_H
 
-/* Fall-through annotation */
-#if defined(__clang__)
-#if (__clang_major__ >= 12) || (__clang_analyzer__)
+
+/* Supported since clang 12 and GCC 7 */
```
pmwkaa, line 10: we are not supporting older compiler versions?

svenklemm: we are the fallthrough marking won't work on those but otherwise they are still supported

pmwkaa: We actually have some defines about it in the cmake file. For eg `CC_SUPPORTS_IMPLICIT_FALLTHROUGHT`, maybe rely on it if makes sense?

svenklemm: this tests whether the compiler supports -Wimplicit-fallthrough not necessarily whether it supports fallthrough attribute

k-rus (approved):

src/annotations.h
```diff
@@ -6,18 +6,15 @@
...
+#if defined __has_attribute
+#if __has_attribute (fallthrough)
```
erimatnor, line 12: Is this standard across both clang and gcc? I couldn't find good information about it so that's why I didn't use it.

svenklemm: https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html
https://clang.llvm.org/docs/LanguageExtensions.html#has-attribute

erimatnor: I was more concern at what version this was introduced and whether one can rely on the check being there for all versions that support the attribute. It seems as if `__attribute__((fallthrough))` was added to clang 12 (I couldn't find docs for previous versions), and I guess the macro check was added prior to that or in the same release so it should be ok.

erimatnor (approved):

PR #2483, opened 2020-10-02
Rename column in job_stats view

gayyappan:
Rename next_scheduled_run to next_start so that
jobs and job_stats views have the same column names


pmwkaa (approved):

svenklemm (approved):

k-rus (commented):
Can you update the following line in continuous_aggs_bgw.sql test?
```
--alter the refresh interval and check if next_scheduled_run is altered
```
You can just search for all occurrences of `next_scheduled_run`

WireBaron (approved):

PR #2489, opened 2020-10-04
Show unscheduled jobs as paused in job_stats

### LINKED ISSUE #2488: Job_stats view shows paused jobs as scheduled.
When jobs are altered to no longer be scheduled (`alter_job(<id>, scheduled => false)`), this change is not reflected in the job_stats view.

WireBaron:
This change updates the timescaledb_information.job_stats view to
check whether a job is currently scheduled in the bgw_config table.
If it is not, the `job_status` field will show `Paused` and the
`next_start` field will be NULL.

Fixes #2488

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -55,16 +55,19 @@ CREATE OR REPLACE VIEW timescaledb_information.job_stats as
                    WHEN js.last_run_success = 'f' THEN 'Failed'
               END
     END as last_run_status,
-    CASE WHEN js.last_finish < '4714-11-24 00:00:00+00 BC' THEN 'Running'
-         WHEN js.next_start IS NOT NULL THEN 'Scheduled'
+    CASE WHEN pgs.state = 'active' THEN 'Running'
+         WHEN j.scheduled = false THEN 'Paused'
+         ELSE 'Scheduled'
     END as job_status,
     CASE WHEN js.last_finish > js.last_start THEN (js.last_finish - js.last_start)
     END as last_run_duration,
-    js.next_start as next_start,
+    CASE WHEN j.scheduled THEN js.next_start 
+    END as next_start,
     js.total_runs, js.total_successes, js.total_failures
   FROM _timescaledb_config.bgw_job j
     INNER JOIN _timescaledb_catalog.hypertable ht ON j.hypertable_id = ht.id
     INNER JOIN _timescaledb_internal.bgw_job_stat js on j.id = js.job_id
+    LEFT JOIN pg_stat_activity pgs ON pgs.datname = current_database() AND pgs.application_name = j.application_name
   ORDER BY ht.schema_name, ht.table_name;
 
 -- view for background worker jobs
--- a/tsl/test/expected/tsl_tables.out
+++ b/tsl/test/expected/tsl_tables.out
@@ -765,6 +765,31 @@ select next_start = now() from alter_job(:job_id, next_start=>now());
  t
 (1 row)
 
+--test pausing/resuming via scheduled parameter
+select job_id from alter_job(:job_id, scheduled=>false);
+ job_id 
+--------
+   1014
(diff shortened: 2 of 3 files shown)


sql/views.sql
```diff
@@ -55,12 +55,14 @@ CREATE OR REPLACE VIEW timescaledb_information.job_stats as
                    WHEN js.last_run_success = 'f' THEN 'Failed'
               END
     END as last_run_status,
-    CASE WHEN js.last_finish < '4714-11-24 00:00:00+00 BC' THEN 'Running'
+    CASE WHEN j.scheduled = false THEN 'Paused'
+         WHEN js.last_finish < '4714-11-24 00:00:00+00 BC' THEN 'Running'
          WHEN js.next_start IS NOT NULL THEN 'Scheduled'
```
erimatnor, line 60: Shouldn't this be simply:

```suggestion
         WHEN j.scheduled = true THEN 'Scheduled'
```

otherwise, what is the meaning of `j.scheduled`?

sql/views.sql
```diff
@@ -55,12 +55,14 @@ CREATE OR REPLACE VIEW timescaledb_information.job_stats as
                    WHEN js.last_run_success = 'f' THEN 'Failed'
               END
     END as last_run_status,
-    CASE WHEN js.last_finish < '4714-11-24 00:00:00+00 BC' THEN 'Running'
+    CASE WHEN j.scheduled = false THEN 'Paused'
+         WHEN js.last_finish < '4714-11-24 00:00:00+00 BC' THEN 'Running'
```
svenklemm, line 59: i think you should do outer join against pg_stat_activity here to get running status

erimatnor, line 59: For clarity, maybe add (although not strictly necessary):
```suggestion
         WHEN j.scheduled AND js.last_finish < '4714-11-24 00:00:00+00 BC' THEN 'Running'
```

This also makes this line independent of other cases (e.g., in case other cases change).

sql/views.sql
gayyappan: There is a change in semantics in 2.0 with the new "scheduled" field in bgw_job. Earlier the only possible states for a job are Scheduled Or Running. (If it  "Scheduled" state , then next_start field is set).
Now we have a third state.
So the view should reflect this new possibility :  ``` if j.scheduled = false then 'Not Scheduled' and then add the other options for the scheduled case ```

sql/views.sql
gayyappan: use "Not scheduled" instead of "Paused"?

gayyappan: @svenklemm We should be able to infer the status based on ( scheduled= true and last_finish ) without resorting to an outer join.

WireBaron: "Not scheduled" seems like a poor semantic fit for `job_status`.  Maybe "Stopped" makes more sense here than "Paused"?

sql/views.sql
WireBaron: Sven actually pointed out that scheduled may be false even if the job is already running, so I moved that to the second case.

sql/views.sql
```diff
@@ -55,16 +55,19 @@ CREATE OR REPLACE VIEW timescaledb_information.job_stats as
                    WHEN js.last_run_success = 'f' THEN 'Failed'
               END
     END as last_run_status,
-    CASE WHEN js.last_finish < '4714-11-24 00:00:00+00 BC' THEN 'Running'
-         WHEN js.next_start IS NOT NULL THEN 'Scheduled'
+    CASE WHEN pgs.state = 'active' THEN 'Running'
```
erimatnor, line 58: Great suggestion by @svenklemm to get rid of that hard-coded timestamp.

erimatnor (approved):
Looks cleaner now.

svenklemm (approved):

PR #2495, opened 2020-10-05
Fix timescaledb_fdw sql script

svenklemm:
Since CREATE FOREIGN DATA WRAPPER is not idempotent it must not be
grouped with the normal sql scripts but has to be in the pre_install
group.


gayyappan (commented):
timescaledb_fdw.sql should be moved to pre_install directory as well?

svenklemm:
> timescaledb_fdw.sql should be moved to pre_install directory as well?

This PR moves the file so not sure what you are asking

WireBaron (approved):

erimatnor (approved):

PR #2497, opened 2020-10-05
Run pg_format on sql files

svenklemm:


erimatnor (approved):
Looks OK, but don't we have other SQL code that code benefit from pg_format?

svenklemm:
> Looks OK, but don't we have other SQL code that code benefit from pg_format?

I think eventually we want to run it on all/most sql files but that probably requires some tuning steps to find suitable configuration, with my current config it splits function definitions weirdly

sql/pre_install/tables.sql
```diff
@@ -40,302 +38,310 @@
...
-    ),
-    CHECK (
-        (integer_now_func_schema IS NULL AND integer_now_func IS NULL) OR
-        (integer_now_func_schema IS NOT NULL AND integer_now_func IS NOT NULL)
-    ),
-    UNIQUE (hypertable_id, column_name)
+  id serial NOT NULL PRIMARY KEY,
+  hypertable_id integer NOT NULL REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
+  column_name name NOT NULL,
+  column_type REGTYPE NOT NULL,
```
erimatnor, line 90: Is it intentional that pg_format leaves this type capitalized?

WireBaron (approved):
Not sure I agree with all of the changes, but I do like that we do have a uniform style now.

sql/views.sql
```diff
@@ -91,179 +106,190 @@ FROM _timescaledb_config.bgw_job j
...
-          and N.nspname = cagg.user_view_schema ) viewinfo, LATERAL
-        ( select pg_get_viewdef(C.oid) as viewdefinition
-          FROM pg_class C LEFT JOIN pg_namespace N on (N.oid = C.relnamespace)
-          where C.relkind = 'v' and C.relname = cagg.direct_view_name
-          and N.nspname = cagg.direct_view_schema ) directview
-  WHERE cagg.mat_hypertable_id = ht.id;
+CREATE OR REPLACE VIEW timescaledb_information.continuous_aggregates AS
+SELECT cagg.user_view_schema AS view_schema,
+  cagg.user_view_name AS view_name,
+  viewinfo.viewowner AS view_owner,
```
erimatnor, line 112: Non-formatting nit. I guess we used just `owner` for the `hypertables` and `data_nodes` views. Perhaps we should adopt the same here?

sql/views.sql
```diff
@@ -91,179 +106,190 @@ FROM _timescaledb_config.bgw_job j
...
+CREATE OR REPLACE VIEW timescaledb_information.chunks AS
+SELECT hypertable_schema,
+  hypertable_name,
+  schema_name AS chunk_schema,
+  chunk_name,
+  primary_dimension,
+  primary_dimension_type,
+  range_start,
+  range_end,
+  integer_range_start AS range_start_integer,
```
erimatnor, line 164: Also nit unrelated to formatting. A bit strange to first alias this to `integer_range_start` to then re-alias here to `range_start_integer`

erimatnor (commented):
Pointed out a couple of view inconsistencies, not necessary to fix now.

sql/pre_install/tables.sql
erimatnor: manually fix then?

PR #2526, opened 2020-10-11
Fix flaky remote_txn test

svenklemm:
Change expected test output to sqlstate only in places where the
error message might differ between platforms/environments.

### DIFF
--- a/tsl/test/expected/remote_txn.out
+++ b/tsl/test/expected/remote_txn.out
@@ -426,10 +426,11 @@ NOTICE:  [loopback]:  INSERT INTO "S 1"."T 1" VALUES (10002,1,'bleh', '2001-01-0
  
 (1 row)
 
-COMMIT;
-WARNING:  kill event: pre-prepare-transaction
-WARNING:  transaction rollback on data node "loopback" failed
-ERROR:  [loopback]: SSL connection has been closed unexpectedly
+-- since the error messages/warnings from this COMMIT varies between
+-- platforms/environments we remove it from test output and show SQLSTATE instead.
+-- SQLSTATE should be 08006 connection_failure
+\echo 'COMMIT SQLSTATE' :SQLSTATE
+COMMIT SQLSTATE 08006
 --the connection was killed, so should be cleared
 SELECT node_name, connection_status, transaction_status, transaction_depth, processing
 FROM _timescaledb_internal.show_connection_cache() ORDER BY 1,4;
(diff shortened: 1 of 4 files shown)


tsl/test/sql/remote_txn.sql
```diff
@@ -217,7 +217,11 @@ FROM _timescaledb_internal.show_connection_cache() ORDER BY 1,4;
 BEGIN;
     SELECT remote_node_killer_set_event('pre-prepare-transaction', 'loopback');
     SELECT test.remote_exec('{loopback}', $$ INSERT INTO "S 1"."T 1" VALUES (10002,1,'bleh', '2001-01-01', '2001-01-01', 'bleh') $$);
+-- since the error message might differ between environments we suppress normal output and print sqlstate
```
pmwkaa, line 220: I suggest to say in the comment about the COMMIT statement, since it is not obvious what is going in the output log.

tsl/test/sql/remote_txn.sql
```diff
@@ -217,7 +217,11 @@ FROM _timescaledb_internal.show_connection_cache() ORDER BY 1,4;
...
+--<exclude_from_test>
 COMMIT;
+--</exclude_from_test>
+select :SQLSTATE AS sqlstate;
```
pmwkaa, line 224: Where do we set the `SQLSTATE` in our code?

svenklemm: > Where do we set the `SQLSTATE` in our code?

It's psql feature: https://www.postgresql.org/docs/current/app-psql.html

svenklemm: There is also VERBOSITY sqlstate for psql but unfortunately thats PG12 only

pmwkaa: Ok, cool for some reason I wasn't able to query it as is psql app, so thought it is something else.

tsl/test/expected/remote_txn.out
```diff
@@ -426,10 +426,9 @@ NOTICE:  [loopback]:  INSERT INTO "S 1"."T 1" VALUES (10002,1,'bleh', '2001-01-0
  
 (1 row)
 
-COMMIT;
-WARNING:  kill event: pre-prepare-transaction
-WARNING:  transaction rollback on data node "loopback" failed
-ERROR:  [loopback]: SSL connection has been closed unexpectedly
+-- since the error message might differ between environments we suppress normal output and print sqlstate
+\echo :SQLSTATE
+08006
```
erimatnor, line 431: It would be good to know what the SQLSTATE means in this case. Now it is very opaque. Would it possible to use, e.g., `LAST_ERROR_SQLSTATE` or `LAST_ERROR_MESSAGE` instead?

svenklemm: LAST_ERROR_SQLSTATE does not get reset on successful queries so it would be wrong to use because we want to get the sqlstate of the commit and not any command that happened to trigger last error. LAST_ERROR_MESSAGE is platform dependent and is basically what this PR is trying to avoid using.
I'll add comment which state is expected. You can look up the states here: https://www.postgresql.org/docs/current/errcodes-appendix.html

pmwkaa (approved):
I would still change the comment "since the error message might differ between environments we suppress normal output and print sqlstate" and include information that it does COMMIT operation. This comment goes into the test output log and it is not obvious what is going on by reading it

PR #2527, opened 2020-10-12
Fix flaky issues test

svenklemm:
Change the distributed insert test to shared test so it can run in
parallel and doesn't require dedicated distributed setup.


mkindahl (approved):

pmwkaa (approved):
How moving this test helps not being flaky, does it conflict with others?

svenklemm:
> How moving this test helps not being flaky, does it conflict with others?

The flakyness was due to non-unique datanode names, which using shared test circumvents

pmwkaa:
> datanode

If it is related to conflicting database names @mkindahl is working on PR to make them separate based on test name. Maybe it should be fixed that way instead?

svenklemm:
> > datanode
> 
> If it is related to conflicting database names @mkindahl is working on PR to make them separate based on test name. Maybe it should be fixed that way instead?

Thats for a different test, and in general tests should preferably run as shared test as there is less overhead involved in those leading to faster test execution.

PR #2533, opened 2020-10-12
Fix timescaledb_fdw function handling in update script

svenklemm:
This patch splits the timescaledb_fdw sql file into two parts to
separate the idempotent parts from the non-idempotent ones so
the function definitions can be included in the regular update
script.

### DIFF
--- a/sql/CMakeLists.txt
+++ b/sql/CMakeLists.txt
@@ -9,6 +9,7 @@ set(PRE_INSTALL_SOURCE_FILES
   pre_install/tables.sql
   pre_install/insert_data.sql
   pre_install/bgw_scheduler_startup.sql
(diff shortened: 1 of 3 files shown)


pmwkaa (commented):
Why do we need this change? It feels odd to separate functions specified in the CREATE FOREIGN DATA WRAPPER statement into separate file

svenklemm:
> Why do we need this change? It feels odd to separate functions specified in the CREATE FOREIGN DATA WRAPPER statement into separate file

The functions need to be recreated as part of the update so they point to the correct version of the library. #2534 adds a check to the update test for this issue.

pmwkaa (approved):

erimatnor (approved):

PR #2539, opened 2020-10-12
Ensure reltuples are preserved during compression

### LINKED ISSUE #2524: Keep precompressed values for reltuples and relpages for compressed hypertables.
Currently when we compress a hypertable, we create a new table to store the compressed data and then truncate the existing table.  This ends up zeroing the reltuples, relpages, and relallvisible columns of the pg_class entry for the hypertable, which may result in poor planning of queries on that table.

Additionally, we use the reltuples value from pg_class for our approximate_row_count function.  So until this issue is fixed, this function will be wildly inaccurate for compressed hypertables.

WireBaron:
This change captures the reltuples and relpages (and relallvisible)
statistics from the pg_class table for chunks immediately before
truncating them during the compression code path.  It then restores
the values after truncating, as there is no way to keep postgresql
from clearing these values during this operation.  It also properly
uses these values properly during planning, working around some
postgresql code which substitutes in arbitrary sizing for tables
which don't see to hold data.

Fixes #2524

### DIFF
--- a/src/planner.c
+++ b/src/planner.c
@@ -872,15 +872,32 @@ timescaledb_get_relation_info_hook(PlannerInfo *root, Oid relation_objectid, boo
 
 				if (chunk->fd.compressed_chunk_id > 0)
 				{
+					Relation uncompressed_chunk = table_open(relation_objectid, NoLock);
+
 					ts_get_private_reloptinfo(rel)->compressed = true;
 
 					/* Planning indexes are expensive, and if this is a compressed chunk, we
-					 * know we'll never need to us indexes on the uncompressed version, since
+					 * know we'll never need to use indexes on the uncompressed version, since
 					 * all the data is in the compressed chunk anyway. Therefore, it is much
 					 * faster if we simply trash the indexlist here and never plan any useless
 					 * IndexPaths at all
 					 */
 					rel->indexlist = NIL;
+
+					/* Relation size estimates are messed up on compressed chunks due to there
+					 * being no actual pages for the table in the storage manager.
+					 */
+					rel->pages = (BlockNumber) uncompressed_chunk->rd_rel->relpages;
+					rel->tuples = (double) uncompressed_chunk->rd_rel->reltuples;
+					if (rel->pages == 0)
+						rel->allvisfrac = 0.0;
+					else if (uncompressed_chunk->rd_rel->relallvisible >= rel->pages)
+						rel->allvisfrac = 1.0;
+					else
+						rel->allvisfrac =
+							(double) uncompressed_chunk->rd_rel->relallvisible / rel->pages;
+
+					table_close(uncompressed_chunk, NoLock);
 				}
 			}
 			break;
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -157,6 +157,49 @@ static void row_compressor_finish(RowCompressor *row_compressor);
  ** compress_chunk **
  ********************/
 
+static void
+capture_pgclass_stats(Oid table_oid, int *out_pages, int *out_visible, float *out_tuples)
+{
+	Relation pg_class = table_open(RelationRelationId, RowExclusiveLock);
+	HeapTuple tuple = SearchSysCacheCopy1(RELOID, ObjectIdGetDatum(table_oid));
+	Form_pg_class classform;
+
+	if (!HeapTupleIsValid(tuple))
+		elog(ERROR, "could not find tuple for relation %u", table_oid);
+
+	classform = (Form_pg_class) GETSTRUCT(tuple);
+
+	*out_pages = classform->relpages;
+	*out_visible = classform->relallvisible;
+	*out_tuples = classform->reltuples;
+
+	heap_freetuple(tuple);
+	table_close(pg_class, RowExclusiveLock);
+}
+
+static void
+restore_pgclass_stats(Oid table_oid, int pages, int visible, float tuples)
+{
+	Relation pg_class;
+	HeapTuple tuple;
+	Form_pg_class classform;
+
+	pg_class = table_open(RelationRelationId, RowExclusiveLock);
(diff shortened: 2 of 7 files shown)


tsl/test/shared/expected/ordered_append_join.out
```diff
@@ -3671,14 +3670,14 @@ FROM :TEST_TABLE o1
     AND o1.time = o2.time
   ORDER BY o1.time
   LIMIT 100;
-                                                                                            QUERY PLAN                                                                                            
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+                                                                                               QUERY PLAN                                                                                               
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Limit (actual rows=100 loops=1)
```
svenklemm, line 3675: Hmm this plan looks worse now with this change

src/planner.c
```diff
@@ -872,15 +872,31 @@ timescaledb_get_relation_info_hook(PlannerInfo *root, Oid relation_objectid, boo
 
 				if (chunk->fd.compressed_chunk_id > 0)
 				{
+					Relation relation = table_open(relation_objectid, NoLock);
```
gayyappan, line 875: Could you rename this variable to something more informative? What does relation_objectid refer to : is this the chunk's oid or compressed chunk's oid?

WireBaron: In this case, it is the oid of the uncompressed chunk.  I've renamed `relation` to `uncompressed_chunk` to be clearer.

tsl/test/shared/expected/ordered_append_join.out
svenklemm: Hmm Limit -> Sort is usually not favorable plan as the Sort prevents further pushdown of the limit

WireBaron: I assume it looked at both paths and found there wasn't enough benefit to pushing down the limit.  We actually run this same query on four different tables and it still uses the merge join in two of the cases (in the other case, both before and after this change it uses a Limit->Sort).

svenklemm (approved):

tsl/src/compression/compression.c
```diff
@@ -157,6 +157,51 @@ static void row_compressor_finish(RowCompressor *row_compressor);
  ** compress_chunk **
  ********************/
 
+static void
+capture_pgclass_stats(Relation rel, int *out_pages, int *out_visible, float *out_tuples)
+{
+	Relation pg_class = table_open(RelationRelationId, RowExclusiveLock);
+	HeapTuple tuple = SearchSysCacheCopy1(RELOID, ObjectIdGetDatum(RelationGetRelid(rel)));
```
erimatnor, line 164: Nit: maybe pass in just relation OID instead of `Relation`.

tsl/src/compression/compression.c
```diff
@@ -157,6 +157,51 @@ static void row_compressor_finish(RowCompressor *row_compressor);
...
+
+static void
+restore_pgclass_stats(Relation rel, int pages, int visible, float tuples)
+{
+	Relation pg_class;
+	HeapTuple tuple;
+	Form_pg_class classform;
+
+	pg_class = table_open(RelationRelationId, RowExclusiveLock);
+	tuple = SearchSysCacheCopy1(RELOID, ObjectIdGetDatum(RelationGetRelid(rel)));
```
erimatnor, line 188: Nit: same here, consider passing in just relation OID.

tsl/src/compression/compression.c
```diff
@@ -157,6 +157,51 @@ static void row_compressor_finish(RowCompressor *row_compressor);
...
+	classform->relpages = pages;
+	classform->relallvisible = visible;
+	classform->reltuples = tuples;
+
+	CatalogTupleUpdate(pg_class, &tuple->t_self, tuple);
+
+	heap_freetuple(tuple);
+	table_close(pg_class, RowExclusiveLock);
+
+	CommandCounterIncrement();
```
erimatnor, line 202: Often it is better to do this outside the functions since the caller might want to make multiple changes visible in one single CommandCounterIncrement

erimatnor (approved):
Just a couple of nits.

src/planner.c
```diff
@@ -872,15 +872,32 @@ timescaledb_get_relation_info_hook(PlannerInfo *root, Oid relation_objectid, boo
...
 					 */
 					rel->indexlist = NIL;
+
+					/* Relation size estimates are messed up on compressed chunks due to there
+					 * being no actual pages for the table in the storage manager.
+					 */
+					rel->pages = (BlockNumber) uncompressed_chunk->rd_rel->relpages;
+					rel->tuples = (double) uncompressed_chunk->rd_rel->reltuples;
+					if (rel->pages == 0)
+						rel->allvisfrac = 0;
```
erimatnor, line 893:
```suggestion
						rel->allvisfrac = 0.0;
```

src/planner.c
```diff
@@ -872,15 +872,32 @@ timescaledb_get_relation_info_hook(PlannerInfo *root, Oid relation_objectid, boo
...
+					else if (uncompressed_chunk->rd_rel->relallvisible >= rel->pages)
+						rel->allvisfrac = 1;
```
erimatnor, line 895:
```suggestion
						rel->allvisfrac = 1.0;
```

tsl/src/compression/compression.c
WireBaron: fixed

tsl/src/compression/compression.c
WireBaron: fixed

tsl/src/compression/compression.c
WireBaron: fixed

src/planner.c
WireBaron: fixed

src/planner.c
WireBaron: fixed

PR #2553, opened 2020-10-15
Suspend retention policies with caggs conflicts

### LINKED ISSUE #2530: Disable failing retention policies when upgrading to 2.0
Prior to 2.0, it was possible to create a continuous aggregate on a hypertable that had a retention policy. This would make the retention policy fail, but this wasn't made clear. It was also possible to add retention policies after creating continuous aggregates, even if those policies wouldn't work due to conflicting `ignore_invalidation_older_than` settings.

Because of the above situation, it is not unlikely that a database that is upgraded to 2.0 will have retention policies running that are in conflict with continuous aggregates and would therefore always fail. If those retention policies are migrated to 2.0 they will start to work, with possibly unintended consequences. Therefore, we should check during extension update to 2.0 for such retention policies and disable them.

WireBaron:
When upgrading from 1.7, it's possible to have retention policies
which overlap with continuous aggregates.  These make use of the
cascade_to_materializations parameter to avoid invalidating the
aggregate.

In 2.0 there is no equivalent behavior to prevent the retention from
disrupting the aggregate.  So during the 2.0 upgrade, check for any
running retention policies that are dropping chunks still used by a
continuous aggregate and suspend them (scheduled=>false).  This will
also print a notice informing the user of what happened and how to
resume the retention policy if that's what they truly want.

Fixes #2530

### DIFF
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -63,3 +63,24 @@ CREATE INDEX continuous_aggs_materialization_invalidation_log_idx ON
     _timescaledb_catalog.continuous_aggs_materialization_invalidation_log (
         materialization_id, lowest_modified_value ASC);
 GRANT SELECT ON  _timescaledb_catalog.continuous_aggs_materialization_invalidation_log TO PUBLIC;
+
+-- Suspend any running retention policies that conflict with continuous aggs
+-- Note that this approach will work for both timestamp and integer time columns
+DO $$
+DECLARE
+  jobid INTEGER;
+BEGIN
+    FOR jobid IN
+        SELECT c.id
+        FROM _timescaledb_config.bgw_job a
+        LEFT JOIN _timescaledb_catalog.continuous_agg b ON a.hypertable_id = b.mat_hypertable_id
+        INNER JOIN _timescaledb_config.bgw_job c ON c.hypertable_id = b.raw_hypertable_id
+        WHERE a.proc_name = 'policy_refresh_continuous_aggregate' AND c.proc_name = 'policy_retention' AND c.scheduled
+            AND ((a.config->'start_offset') = NULL OR (a.config->'start_offset')::text::interval > (c.config->'drop_after')::text::interval)
+    LOOP
+        RAISE NOTICE 'suspending data retention policy with job id %.', jobid
+            USING DETAIL = 'The retention policy (formerly drop_chunks policy) will drop chunks while a continuous aggregate is still running on them. This will likely result in overwriting the aggregate with empty data.',
+            HINT = ('To restore the retention policy, with the possibility of updating aggregates with dropped data, run: SELECT alter_job(%, scheduled=>true);  Otherwise, please create a new rention_policy with a larger drop_after parameter and remove the old policy with: SELECT delete_job(%);', jobid, jobid);
+        UPDATE _timescaledb_config.bgw_job SET scheduled = false WHERE id = jobid;
+    END LOOP;
+END $$;
--- a/test/sql/updates/setup.continuous_aggs.v2.sql
+++ b/test/sql/updates/setup.continuous_aggs.v2.sql
@@ -398,3 +398,42 @@ REFRESH MATERIALIZED VIEW mat_inttime2;
 CALL refresh_continuous_aggregate('mat_inttime',NULL,NULL);
 CALL refresh_continuous_aggregate('mat_inttime2',NULL,NULL);
 \endif
+
+-- Test that retention policies that conflict with continuous aggs are disabled --
+CREATE TABLE conflict_test (time TIMESTAMPTZ, location TEXT, temperature DOUBLE PRECISION);
+SELECT create_hypertable('conflict_test', 'time', chunk_time_interval => INTERVAL '1 week');
+
+DO LANGUAGE PLPGSQL $$
+DECLARE
+  ts_version TEXT;
+  retention_jobid INTEGER;
+BEGIN
+  SELECT extversion INTO ts_version FROM pg_extension WHERE extname = 'timescaledb';
+
+  IF ts_version < '2.0.0' THEN
+    CREATE VIEW mat_conflict
+    WITH ( timescaledb.continuous, timescaledb.materialized_only=true,
+           timescaledb.refresh_lag='1 day',
+           timescaledb.ignore_invalidation_older_than = '28 days',
+           timescaledb.refresh_interval='12 hours' )
+    AS 
+      SELECT time_bucket('10 minute', time) as bucket, location, min(temperature) as min_temp,
+        max(temperature) as max_temp, avg(temperature) as avg_temp
+      FROM conflict_test
+      GROUP BY bucket, location;
+
+    PERFORM add_drop_chunks_policy('conflict_test', '14 days'::interval);
+  ELSE
+    CREATE MATERIALIZED VIEW mat_conflict
+    WITH ( timescaledb.continuous, timescaledb.materialized_only=true )
+    AS 
+      SELECT time_bucket('10 minute', time) as bucket, location, min(temperature) as min_temp, 
+        max(temperature) as max_temp, avg(temperature) as avg_temp
+      FROM conflict_test
+      GROUP BY bucket, location WITH NO DATA;
+
+    PERFORM add_continuous_aggregate_policy('mat_conflict', '28 days', '1 day', '12 hours');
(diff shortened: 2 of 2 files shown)


sql/updates/latest-dev.sql
```diff
@@ -63,3 +63,21 @@ CREATE INDEX continuous_aggs_materialization_invalidation_log_idx ON
...
+BEGIN
+    FOR jobid IN
+        SELECT c.id
+        FROM _timescaledb_config.bgw_job a
+        LEFT JOIN _timescaledb_catalog.continuous_agg b ON a.hypertable_id = b.mat_hypertable_id
+        INNER JOIN _timescaledb_config.bgw_job c ON c.hypertable_id = b.raw_hypertable_id
+        WHERE a.proc_name = 'policy_refresh_continuous_aggregate' AND c.proc_name = 'policy_retention' AND c.scheduled
+            AND (a.config->'start_offset')::text::interval > (c.config->'drop_after')::text::interval
+    LOOP
+        RAISE NOTICE 'Retention job with id % will drop chunks while a continuous aggregate is still running on them. Suspending this job for now. If this is the desired behavior, you can reenable the job with: SELECT alter_job(%, scheduled=>true);', jobid, jobid;
```
mfreed, line 80: I'm not sure this notice will suggest to users the potential risks from re-enabling --> they might just blindly do so.  Maybe something like the following, although it is long and may be shortened.

```suggestion
        RAISE NOTICE 'Data retention policy with job id % will drop chunks while a continuous aggregate may refresh the dropped region. Suspending this job for now, as otherwise data in the continuous aggregate view will be replaced with results from the now-deleted region.  To adjust the behavior, alter the 'drop_after' parameter of your data retention policy.  You can re-enable the data retention policy with: SELECT alter_job(%, scheduled=>true);', jobid, jobid;
```

sql/updates/latest-dev.sql
```diff
@@ -63,3 +63,21 @@ CREATE INDEX continuous_aggs_materialization_invalidation_log_idx ON
...
+DECLARE
+  jobid INTEGER;
+BEGIN
+    FOR jobid IN
+        SELECT c.id
+        FROM _timescaledb_config.bgw_job a
+        LEFT JOIN _timescaledb_catalog.continuous_agg b ON a.hypertable_id = b.mat_hypertable_id
+        INNER JOIN _timescaledb_config.bgw_job c ON c.hypertable_id = b.raw_hypertable_id
+        WHERE a.proc_name = 'policy_refresh_continuous_aggregate' AND c.proc_name = 'policy_retention' AND c.scheduled
+            AND (a.config->'start_offset')::text::interval > (c.config->'drop_after')::text::interval
```
gayyappan, line 78: we could have integer offsets as well for cont. aggregates that use integer time buckets.

svenklemm (approved):

sql/updates/latest-dev.sql
WireBaron: Actually it does work here (I tested this case).  Casting the int to text and then to interval converts the int to a seconds interval, which will do the right thing for determining if the cagg and retention policy overlap.

sql/updates/latest-dev.sql
erimatnor: I would prefer if we split this up in a simple main message and detail/hint as we do for normal errors/notices. Something like:

```
RAISE NOTICE 'suspending retention policy with ID %', jobid
      USING DETAIL = 'The retention policy was not running correctly so it was suspended. Enabling it again might remove data from the continuous aggregates.'
      USING HINT = 'Adjust the policy settings and enable the policy again with: SELECT alter_job(%, scheduled=>true).', jobid;
```

Please try to also follow the error/message style guide as any normal message.

sql/updates/latest-dev.sql
erimatnor: If this works as intended, I suggest a comment to make it clear that the behavior is intentional.

erimatnor: Do we need to check the `cascade_to_materializations` settings of the policy prior to being converted to the new format? I guess if that parameter was set to true, the policy was not failing and the intention was in fact to drop data from the continuous aggregates.

erimatnor:
@WireBaron I want to make sure that we are  on the same page w.r.t. the reason for suspending a policy.

The PR description seems to emphasize that there might be valid policies with `cascade_to_materializations=true`. But such policies shouldn't necessarily be suspended since the intention of those are to in fact drop data from the continuous aggregates.

The main case that we wanted to protect against in issue #2530 was one where you could have failing retention policies, i.e., those running with  `cascade_to_materializations=false` and at the same time conflicting with `ignore_invalidation_older_than`. 

For instance. This is one use case that we were thinking about:

1. User creates a hypertable "foo"
2. User creates a retention policy on  "foo" with `cascade_to_materialization=false`
3. At some point later, the user creates a cagg on "foo". The retention policy will now start to fail and the only way to know is that no data is dropped and a non-zero exit status reported in the postgres log (it is not marked as an error, so easy to miss). 
4. The user upgrades to 2.0.

sql/updates/latest-dev.sql
WireBaron: Hmm, I do like the idea of providing a more descriptive message here.  However we do encourage people to just just drop and recreate retention policies rather than try to update the jsonb config through alter_job.  Maybe this instead:
```        RAISE NOTICE 'Data retention policy (formerly drop_chunk policy) with job id % will drop chunks while a continuous aggregate is still running on them. Suspending this job as it will likely result in overwriting the aggregate with empty data. If this is actually the desired behavior, you can reenable the job with: SELECT alter_job(%, scheduled=>true);  Otherwise, please create a new rention_policy with a larger drop_after parameter and remove the old policy with: SELECT delete_job(%)', jobid, jobid, jobid;```

Comment (WireBaron) [reply sql/updates/latest-dev.sql:80]: Ahh, just saw Erik's comment, let me rework this a bit more.

File: sql/updates/latest-dev.sql (hunk 2 above)
Comment (WireBaron) [reply sql/updates/latest-dev.sql:78]: We actually can't check `cascade_to_materializations` at this point as we have to do this update after we complete the rc1 update.  So that parameter will no longer exist by the time we do this update.

File: sql/updates/latest-dev.sql (hunk 1 above)
Comment (WireBaron) [reply sql/updates/latest-dev.sql]:
```        
RAISE NOTICE 'Suspending data retention policy with job id %.', jobid
            USING DETAIL = 'The retention policy (formerly drop_chunks policy) will drop chunks while a continuous aggregate is still running on them. This will likely result in overwriting the aggregate with empty data.',
            HINT = ('To restore the retention policy, with the possibility of updating aggregates with dropped data, run: SELECT alter_job(%, scheduled=>true);  Otherwise, please create a new rention_policy with a larger drop_after parameter and remove the old policy with: SELECT delete_job(%);', jobid, jobid);
 ```

erimatnor: Note that the first letter of the main message should not be capitalized: https://www.postgresql.org/docs/current/error-style-guide.html

I don't think the hint should be in parentheses.

sql/updates/latest-dev.sql
erimatnor: But I guess it must be possible to check this during the update process at some point. I guess it is fine to not do this and just disable all policies, even those that weren't failing and "did the right thing". But we must decide that this is the semantics we'd like to have and then describe the behavior in the migration doc so that users can be advised to take appropriate action.

sql/updates/latest-dev.sql
```diff
@@ -63,3 +63,23 @@ CREATE INDEX continuous_aggs_materialization_invalidation_log_idx ON
...
+    LOOP
+        RAISE NOTICE 'Suspending data retention policy with job id %.', jobid
```
erimatnor, line 80:
```suggestion
        RAISE NOTICE 'suspending data retention policy with job id %.', jobid
```

WireBaron: Fixed

sql/updates/latest-dev.sql
WireBaron: I think the best approach at this point is to raise this notice and change all tables that have this conflict.  It will be annoying for any users who are intentionally doing this, but I can't actually think of a valid use case where a user should intentionally be doing this (it seems just dropping the aggregated chunk would be the solution they'd want to be using).

sql/updates/latest-dev.sql
WireBaron: Good catch, fixed

gayyappan (approved):

PR #2559, opened 2020-10-16
Move chunk size arguments to the end

k-rus:
Arguments chunk_target_size and chunk_sizing_func of create_hypertable
and create_distributed_hypertable are deprecated and undocumented.
Thus moving them to the end of the definitions to avoid confusions
with arguments in the documentation and in the definitions are not
following the same order.

### DIFF
--- a/sql/ddl_api.sql
+++ b/sql/ddl_api.sql
@@ -34,11 +34,11 @@ CREATE OR REPLACE FUNCTION  create_hypertable(
     if_not_exists           BOOLEAN = FALSE,
     partitioning_func       REGPROC = NULL,
     migrate_data            BOOLEAN = FALSE,
-    chunk_target_size       TEXT = NULL,
-    chunk_sizing_func       REGPROC = '_timescaledb_internal.calculate_chunk_interval'::regproc,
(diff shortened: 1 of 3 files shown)


erimatnor (commented):
I'd prefer that we fix this in documentation. We're just pushing the problem down the line as we will end up in the same situation when, and if, we're adding additional parameters in the future.

k-rus:
> I'd prefer that we fix this in documentation. We're just pushing the problem down the line as we will end up in the same situation when, and if, we're adding additional parameters in the future.

@erimatnor In any case I think it is good to have the deprecated arguments at the end of the list, so they are not pop up to the users, who unlikely to use them. Also New arguments can be added before the deprecated arguments, but this usually avoided due to need to change argument order numbers.

svenklemm (approved):

pmwkaa (approved):
Just to note: Highly unlikely but this might break API in case if the function called before without specifying argument names.

erimatnor:
We decided to remove adaptive chunking instead: https://github.com/timescale/timescaledb/issues/2574

Issue #2350, opened 2020-09-09
Can't drop schema containing continuous aggregates


treuherz:
**Relevant system information:**
 - OS: macOS 10.15.6
 - PostgreSQL version: 12.4
 - TimescaleDB version: 1.7.4
 - Installation method: Docker 19.03

**Describe the bug**
For a schema containing a continuous aggregate view, running `DROP SCHEMA my_cool_schema CASCADE` on the schema prints the list of things that the `DROP` would cascade to, but then fails, saying incorrectly that the schema does not exist. To delete the schema, you need to do `DROP VIEW cont_agg_view CASCADE` on all CA views and _then_ drop the schema.

**To Reproduce**
1. Create your schema, with a hypertable and a CA view
```sql
CREATE SCHEMA my_cool_schema;
SET search_path to my_cool_schema,public; -- timescaledb is created in the public schema

CREATE TABLE telemetry_raw (
  ts        TIMESTAMP WITH TIME ZONE,
  value     DOUBLE PRECISION
);

SELECT create_hypertable('telemetry_raw', 'ts');

CREATE VIEW telemetry_1s
  WITH (timescaledb.continuous)
    AS
SELECT time_bucket(INTERVAL '1 s', ts) AS ts_1s,
       avg(value)
  FROM telemetry_raw
 GROUP BY ts_1s;
```
1. Check schema exists:
```
postgres=# \dn+
                                  List of schemas
          Name           |  Owner   |  Access privileges   |      Description
-------------------------+----------+----------------------+------------------------
[...]
 my_cool_schema          | postgres |                      |
 public                  | postgres | postgres=UC/postgres+| standard public schema
                         |          | =UC/postgres         |
[...]
```
2. Try to delete your schema
```sql
DROP SCHEMA my_cool_schema CASCADE;
```

**Expected behavior**
Schema is dropped, e.g.
```
postgres=# drop schema my_cool_schema cascade;
NOTICE:  drop cascades to 4 other objects
DETAIL:  drop cascades to table my_cool_schema.telemetry_raw
drop cascades to view _timescaledb_internal._partial_view_47
drop cascades to view _timescaledb_internal._direct_view_47
drop cascades to view my_cool_schema.telemetry_1s
DROP SCHEMA
```
**Actual behavior**
Confusing error message
```
postgres=# drop schema my_cool_schema casdcade;
NOTICE:  drop cascades to 4 other objects
DETAIL:  drop cascades to table my_cool_schema.telemetry_raw
[...]

PR #2560, opened 2020-10-16
Fix SCHEMA DROP CASCADE with continuous aggregates

pmwkaa:
This change fixes the situation when schema object is dropped
before the cagg which leads to an error when it tries to
resolve it.

Fixes: #2350

### DIFF
--- a/src/continuous_agg.c
+++ b/src/continuous_agg.c
@@ -405,37 +405,71 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
 	return ts_continuous_agg_find_by_relid(relid);
 }
 
+static ObjectAddress
+get_and_lock_rel_by_name(const Name schema, const Name name, LOCKMODE mode)
+{
+	ObjectAddress addr;
+	Oid relid = InvalidOid;
+	Oid nspid = get_namespace_oid(NameStr(*schema), true);
+	if (OidIsValid(nspid))
+	{
+		relid = get_relname_relid(NameStr(*name), nspid);
+		if (OidIsValid(relid))
+			LockRelationOid(relid, mode);
+	}
+	ObjectAddressSet(addr, RelationRelationId, relid);
+	return addr;
+}
+
+static ObjectAddress
+get_and_lock_rel_by_hypertable_id(int32 hypertable_id, LOCKMODE mode)
+{
+	ObjectAddress addr;
+	Oid relid = ts_hypertable_id_to_relid(hypertable_id);
+	if (OidIsValid(relid))
+		LockRelationOid(relid, mode);
+	ObjectAddressSet(addr, RelationRelationId, relid);
+	return addr;
+}
+
 /*
  * Drops continuous aggs and all related objects.
  *
- * These objects are: the user view itself, the catalog entry in
- * continuous-agg , the partial view,
- * the materialization hypertable,
- * trigger on the raw hypertable (hypertable specified in the user view )
- * copy of the user view query (aka the direct view)
- * NOTE: The order in which the objects are dropped should be EXACTLY the same as in materialize.c"
+ * This function is intended to be run by event trigger during CASCADE,
+ * which implies that most of the dependent objects potentially could be
+ * dropped including associated schema.
+ *
+ * These objects are:
+ *
+ * - user view itself
+ * - continuous agg catalog entry
+ * - partial view
+ * - materialization hypertable
+ * - trigger on the raw hypertable (hypertable specified in the user view)
+ * - copy of the user view query (AKA the direct view)
+ *
+ * NOTE: The order in which the objects are dropped should be EXACTLY the
+ * same as in materialize.c
  *
  * drop_user_view indicates whether to drop the user view.
  *                (should be false if called as part of the drop-user-view callback)
  */
 static void
-drop_continuous_agg(ContinuousAgg *agg, bool drop_user_view)
+drop_continuous_agg(FormData_continuous_agg *cadata, bool drop_user_view)
 {
-	ScanIterator iterator =
-		ts_scan_iterator_create(CONTINUOUS_AGG, RowExclusiveLock, CurrentMemoryContext);
-	Catalog *catalog = ts_catalog_get();
-	ObjectAddress user_view = { .objectId = InvalidOid }, partial_view = { .objectId = InvalidOid },
-				  rawht_trig = { .objectId = InvalidOid }, direct_view = { .objectId = InvalidOid };
-	Hypertable *mat_hypertable, *raw_hypertable;
-	int32 count = 0;
-	bool raw_hypertable_has_other_caggs = true;
-	bool raw_hypertable_exists;
+	Catalog *catalog;
+	ScanIterator iterator;
+	ObjectAddress user_view = { 0 };
+	ObjectAddress partial_view = { 0 };
+	ObjectAddress direct_view = { 0 };
+	ObjectAddress raw_hypertable_trig = { 0 };
+	ObjectAddress raw_hypertable = { 0 };
+	ObjectAddress mat_hypertable = { 0 };
+	bool raw_hypertable_has_other_caggs;
 
-	/* NOTE: the lock order matters, see tsl/src/materialization.c. Perform all locking upfront */
-
-	/* delete the job before taking locks as it kills long-running jobs which we would otherwise
-	 * wait on */
-	List *jobs = ts_bgw_job_find_by_hypertable_id(agg->data.mat_hypertable_id);
+	/* Delete the job before taking locks as it kills long-running jobs
+	 * which we would otherwise wait on */
+	List *jobs = ts_bgw_job_find_by_hypertable_id(cadata->mat_hypertable_id);
 	ListCell *lc;
 
 	foreach (lc, jobs)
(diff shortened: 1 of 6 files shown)


tsl/test/expected/continuous_aggs_ddl.out
```diff
@@ -1110,5 +1110,36 @@ partial_view_owner | test_role_1
 tablespace         | 
 
 \x off
+--
+-- Test DROP SCHEMA CASCADE with continuous aggregates
+--
+-- Issue: #2350
+CREATE SCHEMA my_cool_schema;
+SET search_path to my_cool_schema, public; -- timescaledb is created in the public schema
```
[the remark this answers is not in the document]
svenklemm: or if you think search_path interaction is worth testing have extra tests that test it in addition to specifying schema directly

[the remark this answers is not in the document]
pmwkaa: Sure, just thought it makes sense to keep it since it is the same case as in the issue description

svenklemm (approved):

src/continuous_agg.c
```diff
@@ -577,7 +577,15 @@ ts_continuous_agg_drop_hypertable_callback(int32 hypertable_id)
...
 		{
-			continuous_agg_init(&ca, data);
+			/* schema might not exists at this point */
+			Oid nspid = get_namespace_oid(NameStr(data->user_view_schema), true);
+
+			if (OidIsValid(nspid))
+				ca.relid = get_relname_relid(NameStr(data->user_view_name), nspid);
+			else
+				ca.relid = InvalidOid;
+			ca.data = *data;
```
gayyappan, line 587: shouldn't these 2 lines be conditional?  i.e.
i.e. this is valid only if you cna find a view object.
```suggestion
if( OidIsValid(...) )
{
            ca.redlid = ...
			ca.data = *data;
			drop_continuous_agg(...)
}
```

erimatnor (changes requested):
I think we should try to hook into `DROP SCHEMA` before it is executed by PostgreSQL so that we can drop dependent objects while they are still valid. 

I think it is problematic to have functions that can create both fully valid objects (e.g., Hypertable) and "partial" objects (without all fields set) because it is hard to know which version of the object is produced for a particular case.

src/continuous_agg.c
```diff
@@ -577,7 +577,15 @@ ts_continuous_agg_drop_hypertable_callback(int32 hypertable_id)
 
 		if (data->raw_hypertable_id == hypertable_id)
 		{
-			continuous_agg_init(&ca, data);
+			/* schema might not exists at this point */
+			Oid nspid = get_namespace_oid(NameStr(data->user_view_schema), true);
+
+			if (OidIsValid(nspid))
```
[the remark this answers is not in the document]
pmwkaa: I understand your concern. I tried to make this fix as smaller as possible because we already have a lot of assumption that this functionality can be called during `CASCADE` and expects some of the objects to be deleted: https://github.com/timescale/timescaledb/blob/master/src/continuous_agg.c#L459

src/continuous_agg.c
pmwkaa: Yes, you right. Missed that. But it should not be used anyway I believe since the relation marked as not found.

src/continuous_agg.c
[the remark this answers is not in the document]
erimatnor: Well, those assumptions aren't great either, TBH.

I am actually most concerned about this PR introducing an ambiguous state for the `Hypertable` object, in a change below. We've had issues related to such things before where some fields could be set or not set, and we've fixed most of them to always be consistent. This change would move us back towards a situation where objects can again be in different "states".

pmwkaa:
`drop_continuous_agg()` refactored and now tries to work with catalog data only assuming that depending objects (including associated schema) might not exists anymore

k-rus:
@pmwkaa It seems that you have a linking issue on Windows. From Appveyor log:
```
  timescaledb-tests.vcxproj -> C:\projects\timescaledb\build\test\src\timescaledb-tests.dir\Debug\timescaledb-tests.lib
550     Creating library C:/projects/timescaledb/build/src/Debug/timescaledb-2.1.0-dev.lib and object C:/projects/timescaledb/build/src/Debug/timescaledb-2.1.0-dev.exp
551continuous_agg.obj : error LNK2001: unresolved external symbol InvalidObjectAddress [C:\projects\timescaledb\build\src\timescaledb.vcxproj]
552C:\projects\timescaledb\build\src\Debug\timescaledb-2.1.0-dev.dll : fatal error LNK1120: 1 unresolved externals [C:\projects\timescaledb\build\src\timescaledb.vcxproj]
```

src/continuous_agg.c
```diff
@@ -405,37 +405,71 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
...
-
-	/* delete the job before taking locks as it kills long-running jobs which we would otherwise
-	 * wait on */
-	List *jobs = ts_bgw_job_find_by_hypertable_id(agg->data.mat_hypertable_id);
+	ObjectAddress user_view = InvalidObjectAddress;
+	ObjectAddress partial_view = InvalidObjectAddress;
+	ObjectAddress direct_view = InvalidObjectAddress;
+	ObjectAddress rawht_trig = InvalidObjectAddress;
+	ObjectAddress raw_hypertable = InvalidObjectAddress;
+	ObjectAddress mat_hypertable = InvalidObjectAddress;
```
k-rus, line 460: I guess these lines introduce the linking error on Windows.

src/continuous_agg.c
```diff
@@ -405,37 +405,64 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
...
+ * which implies that most of the dependent objects potentially could be
+ * dropped including associated schema.
+ *
+ * These objects are: the user view itself, the catalog entry in continuous-agg,
+ * the partial view, the materialization hypertable, trigger on the raw hypertable
+ * (hypertable specified in the user view) copy of the user view query
+ * (aka the direct view).
+ *
+ * NOTE: The order in which the objects are dropped should be EXACTLY the
+ * same as in materialize.c
```
k-rus, line 444: Am I correct that the comment is just explicitly describes this assumption, which existed before this PR? Or is it brought by this PR?

k-rus (commented):
I have a comment about the signature helper functions. See the comments.

src/continuous_agg.c
pmwkaa: Yes, that was before. Just moved a couple of lines below close to the actual action

src/continuous_agg.c
```diff
@@ -405,37 +405,64 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
 	return ts_continuous_agg_find_by_relid(relid);
 }
 
+static void
+get_rel_by_name(ObjectAddress *addr, const char *schema, const char *name, LOCKMODE mode)
```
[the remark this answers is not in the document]
pmwkaa: Also updated the argument types to avoid `Name` casting

src/continuous_agg.c
```diff
@@ -405,37 +405,68 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
 	return ts_continuous_agg_find_by_relid(relid);
 }
 
+static ObjectAddress
+get_rel_by_name(Name schema, Name name, LOCKMODE mode)
```
gayyappan, line 409: Rename as lock_rel_by_name as  that is the primary function here.

src/continuous_agg.c
```diff
@@ -405,37 +405,68 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
...
+		relid = get_relname_relid(NameStr(*name), nspid);
+		if (OidIsValid(relid))
+			LockRelationOid(relid, mode);
+	}
+	ObjectAddressSet(addr, RelationRelationId, relid);
+	return addr;
+}
+
+static ObjectAddress
+get_rel_by_hypertable_id(int32 hypertable_id, LOCKMODE mode)
```
gayyappan, line 425:
```suggestion
lock_rel_by_hypertable_id(int32 hypertable_id, LOCKMODE mode)
```
could you rename this as well?

src/continuous_agg.c
```diff
@@ -444,80 +475,61 @@ drop_continuous_agg(ContinuousAgg *agg, bool drop_user_view)
...
-			get_relname_relid(NameStr(agg->data.user_view_name),
-							  get_namespace_oid(NameStr(agg->data.user_view_schema), false)),
-	};
-	/* The partial view may already be dropped by PG's dependency system (e.g. the raw table was
-	 * dropped) */
-	if (OidIsValid(user_view.objectId))
-		LockRelationOid(user_view.objectId, AccessExclusiveLock);
-
-	raw_hypertable = ts_hypertable_get_by_id(agg->data.raw_hypertable_id);
-	/* The raw hypertable might be already dropped if this is a cascade from that drop */
```
gayyappan, line 459: These comments  about locking and situtaions related to dependent objects being dropped, should be retained. Otherwise, it is not clear why it is ok to not throw an error if the object is not found.

src/continuous_agg.c
```diff
@@ -444,80 +475,61 @@ drop_continuous_agg(ContinuousAgg *agg, bool drop_user_view)
...
-		.classId = RelationRelationId,
-		.objectId =
-			get_relname_relid(NameStr(agg->data.direct_view_name),
-							  get_namespace_oid(NameStr(agg->data.direct_view_schema), false)),
-	};
-	if (OidIsValid(direct_view.objectId))
-		LockRelationOid(direct_view.objectId, AccessExclusiveLock);
-
-	/*  END OF LOCKING. Perform actual deletions now. */
+	partial_view = get_rel_by_name(&cadata->partial_view_schema,
```
gayyappan, line 523: Please retain the comment at line 500 (from the old code)

tsl/test/sql/continuous_aggs_ddl.sql
```diff
@@ -734,5 +734,31 @@ ALTER MATERIALIZED VIEW owner_check OWNER TO :ROLE_1;
...
+CREATE SCHEMA test_schema;
+
+CREATE TABLE test_schema.telemetry_raw (
+  ts        TIMESTAMP WITH TIME ZONE NOT NULL,
+  value     DOUBLE PRECISION
+);
+
+SELECT create_hypertable('test_schema.telemetry_raw', 'ts');
+
+CREATE MATERIALIZED VIEW test_schema.telemetry_1s
```
gayyappan, line 751: could you add 1 more mat view for the test case so that we have multiple caggs attached to the hypertable?

src/continuous_agg.c
pmwkaa: I don't think this is accurate, because the relation might not exists. In such case semantically it is more close to the get (or get and lock if exists) operation and it returns its address too. . Maybe `get_and_lock_rel_by_name`? not fan of long names

src/continuous_agg.c
pmwkaa: I've updated the function description in order to describe the situation. There were couple of duplicate messages about the locking, not sure how much value they have. I can put them back, if you would like

src/continuous_agg.c
gayyappan: yes, get_and_lock_rel_by_name sounds good.

src/continuous_agg.c
```diff
@@ -529,26 +541,33 @@ drop_continuous_agg(ContinuousAgg *agg, bool drop_user_view)
...
+		performDeletion(&user_view, DROP_RESTRICT, 0);
 
-	/* delete the materialization table */
-	ts_hypertable_drop(mat_hypertable, DROP_CASCADE);
+	if (OidIsValid(raw_hypertable_trig.objectId))
+		ts_hypertable_drop_trigger(raw_hypertable.objectId, CAGGINVAL_TRIGGER_NAME);
+
+	if (OidIsValid(mat_hypertable.objectId))
+	{
+		performDeletion(&mat_hypertable, DROP_CASCADE, 0);
```
erimatnor, line 564: Why does this need CASCADE, while others don't? It would be good to at least explain this.

src/continuous_agg.c
```diff
@@ -405,37 +405,68 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
...
- * continuous-agg , the partial view,
- * the materialization hypertable,
- * trigger on the raw hypertable (hypertable specified in the user view )
- * copy of the user view query (aka the direct view)
- * NOTE: The order in which the objects are dropped should be EXACTLY the same as in materialize.c"
+ * This function is intedended to be run by event trigger during CASCADE,
+ * which implies that most of the dependent objects potentially could be
+ * dropped including associated schema.
+ *
+ * These objects are: the user view itself, the catalog entry in continuous-agg,
```
erimatnor, line 442: I think this would be better described with a bullet-list of all objects.

src/continuous_agg.c
```diff
@@ -405,37 +405,68 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
...
+ * the partial view, the materialization hypertable, trigger on the raw hypertable
+ * (hypertable specified in the user view), copy of the user view query
+ * (aka the direct view).
```
erimatnor, line 445:
```suggestion
 * (AKA the direct view).
```

tsl/test/sql/continuous_aggs_ddl.sql
```diff
@@ -734,5 +734,31 @@ ALTER MATERIALIZED VIEW owner_check OWNER TO :ROLE_1;
...
+  WITH (timescaledb.continuous)
+    AS
+SELECT time_bucket(INTERVAL '1 s', ts) AS ts_1s,
+       avg(value)
+  FROM test_schema.telemetry_raw
+ GROUP BY ts_1s WITH NO DATA;
+
+DROP SCHEMA test_schema CASCADE;
+
+SELECT true FROM pg_catalog.pg_namespace WHERE nspname = 'test_schema';
```
erimatnor, line 761: Can you show that the internal objects (e.g., the materialized hypertable and views) are also removed from the internal namespaces (or whatever namespaces they existed in).

src/continuous_agg.c
```diff
@@ -405,37 +405,68 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
...
+ * same as in materialize.c
  *
  * drop_user_view indicates whether to drop the user view.
  *                (should be false if called as part of the drop-user-view callback)
  */
 static void
-drop_continuous_agg(ContinuousAgg *agg, bool drop_user_view)
+drop_continuous_agg(FormData_continuous_agg *cadata, bool drop_user_view)
 {
 	ScanIterator iterator =
```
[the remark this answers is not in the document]
pmwkaa: I can do it, but was worried if that has something to do with locking order

src/continuous_agg.c
pmwkaa: I don't have an explanation for that, rather then it was there before. I can only imaging some user-case depending on this object

pmwkaa: CASCADE is required there, it will not work with RESTRICT. Likely for the hypertable it is expected to have depenent objects? (chunks, etc)

src/continuous_agg.c
```diff
@@ -405,37 +405,71 @@ ts_continuous_agg_find_by_rv(const RangeVar *rv)
...
 /*
  * Drops continuous aggs and all related objects.
  *
- * These objects are: the user view itself, the catalog entry in
- * continuous-agg , the partial view,
- * the materialization hypertable,
- * trigger on the raw hypertable (hypertable specified in the user view )
- * copy of the user view query (aka the direct view)
- * NOTE: The order in which the objects are dropped should be EXACTLY the same as in materialize.c"
+ * This function is intedended to be run by event trigger during CASCADE,
```
erimatnor, line 438:
```suggestion
 * This function is intended to be run by event trigger during CASCADE,
```

src/continuous_agg.c
erimatnor: I don't think cascade is required normally when dropping a hypertable. This looks potentially   dangerous, but I guess it is mitigated by this being an internal table.

erimatnor (approved):
One comment nit.

src/continuous_agg.c
gayyappan: I think it is still good to keep these comments about locking and  objects that could have been potentially dropped (line 459 , 463 , 464 and 500)..

gayyappan (approved):
Changes look good. Request to retain additional comments.

PR #2577, opened 2020-10-20
Exclude compressed chunks from ANALYZE/VACUUM

### LINKED ISSUE #2576: ANALYZE and VACUUM clear pg_class stats for compressed chunks
When running a `ANALYZE` or `VACUUM` in postgresql without providing relations, the system automatically performs the operation on every table in the database.  For chunks which have been compressed this ends up erasing all the stat information we saved during compression.

WireBaron:
This change makes sure that ANALYZE and VACUUM commands run without
any relations will not clear the stats on compressed chunks that
were saved at compression time.  It also will skip any distributed
tables.

Fixes #2576

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -600,6 +600,81 @@ add_chunk_to_vacuum(Hypertable *ht, Oid chunk_relid, void *arg)
 	ctx->chunk_rels = lappend(ctx->chunk_rels, chunk_vacuum_rel);
 }
 
+/*
+ * Construct a list of VacuumRelations for all vacuumable rels in
+ * the current database.  This is similar to the PostgresQL get_all_vacuum_rels
+ * from vacuum.c, only it filters out distributed hypertables and chunks
+ * that have been compressed.
+ */
+static List *
+ts_get_all_vacuum_rels(bool is_vacuumcmd)
+{
+	List *vacrels = NIL;
+	Relation pgclass;
+	TableScanDesc scan;
+	HeapTuple tuple;
+	Cache *hcache = ts_hypertable_cache_pin();
+
+	pgclass = table_open(RelationRelationId, AccessShareLock);
+
+	scan = table_beginscan_catalog(pgclass, 0, NULL);
+
+	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	{
+		Form_pg_class classform = (Form_pg_class) GETSTRUCT(tuple);
+		Hypertable *ht;
+		Chunk *chunk;
+		Oid relid;
+
+#if PG12_GE
+		relid = classform->oid;
+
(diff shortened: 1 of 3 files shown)


src/process_utility.c
```diff
@@ -600,6 +600,70 @@ add_chunk_to_vacuum(Hypertable *ht, Oid chunk_relid, void *arg)
...
+	HeapTuple tuple;
+	Cache *hcache = ts_hypertable_cache_pin();
+
+	pgclass = table_open(RelationRelationId, AccessShareLock);
+
+	scan = table_beginscan_catalog(pgclass, 0, NULL);
+
+	while ((tuple = heap_getnext(scan, ForwardScanDirection)) != NULL)
+	{
+		Form_pg_class classForm = (Form_pg_class) GETSTRUCT(tuple);
```
erimatnor, line 624:
```suggestion
		Form_pg_class classform = (Form_pg_class) GETSTRUCT(tuple);
```

src/process_utility.c
```diff
@@ -600,6 +600,70 @@ add_chunk_to_vacuum(Hypertable *ht, Oid chunk_relid, void *arg)
...
+			classForm->relkind != RELKIND_PARTITIONED_TABLE)
+			continue;
+
+		ht = ts_hypertable_cache_get_entry(hcache, relid, CACHE_FLAG_MISSING_OK);
+		if (ht && hypertable_is_distributed(ht))
+			continue;
+
+		chunk = ts_chunk_get_by_relid(relid, false);
+		if (chunk && chunk->fd.compressed_chunk_id != INVALID_CHUNK_ID)
+			continue;
```
erimatnor, line 650: If a hypertable was found, no need to check for chunk:

```suggestion
  	ht = ts_hypertable_cache_get_entry(hcache, relid, CACHE_FLAG_MISSING_OK);
		
		if (ht)
		{
		    if (hypertable_is_distributed(ht))
			    continue;
	    }
	    else
	    {
           Chunk *chunk = ts_chunk_get_by_relid(relid, false);
             
		    if (chunk && chunk->fd.compressed_chunk_id != INVALID_CHUNK_ID)
			   continue;
	     }
```

src/process_utility.c
WireBaron: fixed

src/process_utility.c
WireBaron: fixed

erimatnor (approved):

erimatnor:
Approving, given that test failures are fixed.

svenklemm (changes requested):
I think this needs a discussion whether this is the best approach i think it would be better to make analyze work for compressed chunks because with this PR there is no way to fix weird stats. Stats are not preserved over dump/restore and a user would have no way to fix it.

PR #2589, opened 2020-10-21
Add test for continuous agg policies

gayyappan:
Add mock timestamp support for subtract_from_now.
Use mock timestamp to execute cont. agg policies for
date and timestamp based tables.

### DIFF
--- a/src/time_utils.c
+++ b/src/time_utils.c
@@ -12,8 +12,9 @@
 #include <parser/parse_coerce.h>
 #include <fmgr.h>
 
-#include "time_utils.h"
 #include "dimension.h"
+#include "guc.h"
+#include "time_utils.h"
 #include "utils.h"
 
 static Datum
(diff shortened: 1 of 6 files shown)


src/time_utils.c
```diff
@@ -535,3 +536,22 @@ ts_subtract_integer_from_now_saturating(Oid now_func, int64 interval, Oid timety
 		res = nowval - interval;
 	return res;
 }
+
+/* return mock time for testing */
+Datum
+ts_get_mock_time_or_current_time(void)
```
k-rus, line 542: Can the entire function be available in Debug, like the GUC used in it? I see that the function is only used in Debug.

k-rus (commented):
I suggest to have the mock function only in debug and not return current time.

src/time_utils.c
gayyappan: done.

WireBaron (approved):

k-rus (approved):

PR #2593, opened 2020-10-26
Set explicitly all lock parameters in alter_job

k-rus:
Sets the flag parameter to pass for tuple lock during alter_job
execution. Restores the lost comment explaining the value of the flag
parameter.

### DIFF
--- a/src/bgw/job.c
+++ b/src/bgw/job.c
@@ -990,6 +990,7 @@ ts_bgw_job_update_by_id(int32 job_id, BgwJob *job)
 	ScanTupLock scantuplock = {
 		.waitpolicy = LockWaitBlock,
 		.lockmode = LockTupleExclusive,
+		.lockflags = 0 /* don't follow updates, used in PG12 */
 	};
 	ScannerCtx scanctx = { .table = catalog_get_table_id(catalog, BGW_JOB),
 						   .index = catalog_get_index(catalog, BGW_JOB, BGW_JOB_PKEY_IDX),


src/bgw/job.c
```diff
@@ -990,6 +990,7 @@ ts_bgw_job_update_by_id(int32 job_id, BgwJob *job)
 	ScanTupLock scantuplock = {
 		.waitpolicy = LockWaitBlock,
 		.lockmode = LockTupleExclusive,
+		.lockflags = 0 /* don't follow updates, used in PG12 */
```
pmwkaa, line 993: I suggest to cover it up `#if PG12_GE ...`. See `ts_hypercube_from_constraints()` function

k-rus: @pmwkaa The flag is still present in the struct in all versions of PG. Thus I don't like the idea of not setting it explicitly, since it will be set in any case.

pmwkaa: Yes you right, it is better to initialize it in such case

pmwkaa (approved):

src/bgw/job.c
erimatnor: Not clear why this is important. The struct is zero initialized anyway and I don't see how the comments helps a lot. Still, this is a harmless change, so I am OK with it.

erimatnor (approved):

src/bgw/job.c
k-rus: @erimatnor I come up with this really minor update when I was looking to the code and trying to understand logic. And I've got impression that it was missed to be initialised until I found the comment in the old version of the code, thus I thought it will be helpful in future.

Thank you for approving.

PR #2600, opened 2020-10-28
Add user-defined actions to telemetry

### LINKED ISSUE #2596: Include user-defined actions into telemetry

svenklemm:
This patch adjusts the telemetry code to the job refactorings.
Additionally it adds telemetry entries for user-defined actions,
continuous aggregate policies, compression policies and renames
drop_chunks to retention to match the new policy names.

Fixes #2596

### DIFF
--- a/src/bgw_policy/policy.c
+++ b/src/bgw_policy/policy.c
@@ -61,8 +61,6 @@ ts_bgw_job_type_counts()
 				counts.policy_retention++;
 			else if (namestrcmp(&job->fd.proc_name, "policy_telemetry") == 0)
 				counts.policy_telemetry++;
-			else
-				Assert(false);
 		}
 		else
 		{
--- a/src/bgw_policy/policy.c
+++ b/src/bgw_policy/policy.c
@@ -5,6 +5,7 @@
  */
 
 #include <postgres.h>
+#include <utils/builtins.h>
 
 #include "bgw/job.h"
 #include "policy.h"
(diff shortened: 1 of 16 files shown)


pmwkaa (commented):
I think it requires testing, otherwise looks good

src/bgw_policy/policy.c
```diff
@@ -36,18 +37,38 @@ ts_bgw_policy_delete_row_only_tuple_found(TupleInfo *ti, void *const data)
...
+			else if (namestrcmp(&job->fd.proc_name, "policy_compression") == 0)
+				counts.policy_compression++;
+			else if (namestrcmp(&job->fd.proc_name, "policy_reorder") == 0)
+				counts.policy_reorder++;
+			else if (namestrcmp(&job->fd.proc_name, "policy_retention") == 0)
+				counts.policy_retention++;
+			else if (namestrcmp(&job->fd.proc_name, "policy_telemetry") == 0)
+				counts.policy_telemetry++;
+			else
+				Assert(false);
```
pmwkaa, line 65: Can a user create a job using internal schema?

pmwkaa (approved):

src/bgw_policy/policy.c
```diff
@@ -36,18 +37,38 @@ ts_bgw_policy_delete_row_only_tuple_found(TupleInfo *ti, void *const data)
...
+	{
+		BgwJob *job = lfirst(lc);
 
-int32
-ts_bgw_policy_retention_count()
-{
-	List *jobs = ts_bgw_job_find_by_proc("policy_retention", INTERNAL_SCHEMA_NAME);
+		if (namestrcmp(&job->fd.proc_schema, INTERNAL_SCHEMA_NAME) == 0)
+		{
+			if (namestrcmp(&job->fd.proc_name, "policy_refresh_continuous_aggregate") == 0)
```
erimatnor, line 54: I would prefer that this isn't hard coded with free-floating string  values. This will be cumbersome to maintain when we add new policies and it would be better to either use a hash table to collect policy counts, or use a static list of officially supported policies that we can reference throughout the code.

I guess this is a bigger refactoring, though, and not for this PR.

erimatnor (approved):
Approving with nits.

PR #2604, opened 2020-10-29
Fix chunk creation on hypertables with foreign key constraints

tylerfontaine:
This resolves an issue when using both constraints with index
tablespaces AND constraints that have no indexes on a single hypertable.

chunk_constraint_add_table_constraint() was attempting to add
`USING INDEX TABLESPACE` to all constraints when a given hypertable had
any constraint configured to use a tablespace, resulting in a SYNTAX
error and blocking the creation of new chunks.

The solution here is to limit index tablespace lookups to only the
constraint types which use indexes: primary key and unique so that only
those constraits will have `USING INDEX TABLESPACE` prepended when
necessary.

### DIFF
--- a/sql/chunk_constraint.sql
+++ b/sql/chunk_constraint.sql
@@ -35,7 +35,7 @@ BEGIN
 
         SELECT T.spcname INTO indx_tablespace 
         FROM pg_constraint C, pg_class I, pg_tablespace T
-        WHERE C.oid = constraint_oid AND I.oid = C.conindid AND I.reltablespace = T.oid;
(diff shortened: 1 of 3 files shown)


k-rus (commented):
LGTM!
Can you squash commits and create commit message according to the best practice rules? https://chris.beams.io/posts/git-commit/#seven-rules

tylerfontaine:
Commits squashed and message updated.

k-rus (approved):

erimatnor (approved):

Issue #2592, opened 2020-10-26
Refresh on drop can create outliers in continuous aggregates


k-rus:
The drop chunk, which can be called manually or by a retention policy on a hypertable, calls the refresh on all continuous aggregates of the hypertable for each chunk to drop. The refresh recomputes buckets, which fully covers a chunk to drop, if an invalidation is recorded for it. In the case when chunks and buckets are not align, a retention policy can result in animalities in aggregated values in continuous aggregates if the invalidations happen in the oldest bucket, which contains the start of the oldest dropped chunk in a retention policy call.

The following pictures illustrate the case when chunks and buckets are not aligned and a retention policy was called three times. For each call the recalculation buckets on dropped data is demonstrated.

![Drop on refresh with cagg](https://user-images.githubusercontent.com/1726044/97166276-48899f00-1785-11eb-9e7a-160c9273e37c.png)

The aggregates values in continuous aggregate might contain anomalies, e.g., if the aggregate operator was `sum` or `count`:

![Anomalities after drop on refresh with cagg](https://user-images.githubusercontent.com/1726044/97166376-6eaf3f00-1785-11eb-9529-910d30182849.jpg)

Invalidation in old data is not common. However, if it is enough common, so it is important to refresh on drop, then the described anomaly case will be also not so uncommon. So if the refresh on drop is kept, then this issue needs to be solved too and the solution is not just switching from expanding buckets window to shrinking buckets window, since refresh is done per chunk, while several chunks can be dropped together.

k-rus:
Since drop chunks will register invalidation, the above case will be common whenever buckets and chunks are not aligned.

k-rus:
Reproduced by https://github.com/timescale/timescaledb/pull/2606

k-rus:
The issue is also reproduced in 1.7, see https://github.com/timescale/timescaledb/pull/2618/files

PR #2606, opened 2020-10-30
Add test to verify refresh on drop in caggs

k-rus:
When drop_chunks is called on a hypertable, it will initiate refresh
in hypertable's continuous aggregates. The refresh will include all
buckets covering dropped chunks. If buckets are not aligned with
chunks, it results in outliers. This test demonstrates this. See #2592
for details.


k-rus:
Test can be found in https://github.com/timescale/timescaledb/pull/2613

PR #2610, opened 2020-10-30
Support analyze of internal compression table

gayyappan:
When an internal compression table is analyzed,
statistics from the compressed chunk (such as page count and
tuple count) is used to update the
statistics of the corresponding chunk parent.

### DIFF
--- a/src/compression_chunk_size.c
+++ b/src/compression_chunk_size.c
@@ -90,3 +90,35 @@ ts_compression_chunk_size_totals()
 
 	return sizes;
 }
+
+/* Return the pre-compression row count for the chunk */
+int64
+ts_compression_chunk_size_row_count(int32 uncompressed_chunk_id)
+{
+	int found_cnt = 0;
+	int64 rowcnt = 0;
+	ScanIterator iterator =
+		ts_scan_iterator_create(COMPRESSION_CHUNK_SIZE, AccessShareLock, CurrentMemoryContext);
+	init_scan_by_uncompressed_chunk_id(&iterator, uncompressed_chunk_id);
+	ts_scanner_foreach(&iterator)
+	{
+		bool nulls[Natts_compression_chunk_size];
+		Datum values[Natts_compression_chunk_size];
+		bool should_free;
+		HeapTuple tuple = ts_scan_iterator_fetch_heap_tuple(&iterator, false, &should_free);
+
+		heap_deform_tuple(tuple, ts_scan_iterator_tupledesc(&iterator), values, nulls);
+		if (!nulls[AttrNumberGetAttrOffset(Anum_compression_chunk_size_numrows_pre_compression)])
+			rowcnt = DatumGetInt64(values[AttrNumberGetAttrOffset(
+				Anum_compression_chunk_size_numrows_pre_compression)]);
+		if (should_free)
+			heap_freetuple(tuple);
+		found_cnt++;
+	}
+	if (found_cnt != 1)
+		elog(ERROR,
+			 "missing record for chunk with id %d in %s",
+			 uncompressed_chunk_id,
+			 COMPRESSION_CHUNK_SIZE_TABLE_NAME);
+	return rowcnt;
+}
--- a/src/compression_chunk_size.h
+++ b/src/compression_chunk_size.h
@@ -21,5 +21,6 @@ typedef struct TotalSizes
 } TotalSizes;
 
 extern TSDLLEXPORT TotalSizes ts_compression_chunk_size_totals(void);
+extern TSDLLEXPORT int64 ts_compression_chunk_size_row_count(int32 uncompressed_chunk_id);
 
 #endif
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -576,7 +576,38 @@ typedef struct VacuumCtx
 {
 	VacuumRelation *ht_vacuum_rel;
 	List *chunk_rels;
+	List *chunk_pairs;
 } VacuumCtx;
+
+typedef struct ChunkPair
+{
+	Oid uncompressed_relid;
+	Oid compressed_relid;
+} ChunkPair;
+
+static void
+add_compressed_chunk_to_vacuum(Hypertable *ht, Oid comp_chunk_relid, void *arg)
+{
+	VacuumCtx *ctx = (VacuumCtx *) arg;
+	Chunk *compressed_chunk = ts_chunk_get_by_relid(comp_chunk_relid, true);
+	VacuumRelation *chunk_vacuum_rel;
+
+	Chunk *chunk_parent;
+	/* chunk is from a compressed hypertable */
+	Assert(ht->fd.compressed);
+
+	/*chunks for internal compression table have a parent */
+	chunk_parent = ts_chunk_get_compressed_chunk_parent(compressed_chunk);
+	Assert(chunk_parent != NULL);
+
+	ChunkPair *cp = palloc(sizeof(ChunkPair));
+	cp->uncompressed_relid = chunk_parent->table_id;
+	cp->compressed_relid = comp_chunk_relid;
+	ctx->chunk_pairs = lappend(ctx->chunk_pairs, cp);
+	/* analyze/vacuum the compressed rel instead */
+	chunk_vacuum_rel = makeVacuumRelation(NULL, comp_chunk_relid, NIL);
+	ctx->chunk_rels = lappend(ctx->chunk_rels, chunk_vacuum_rel);
+}
 
 /* Adds a chunk to the list of tables to be vacuumed */
 static void
@@ -587,16 +618,30 @@ add_chunk_to_vacuum(Hypertable *ht, Oid chunk_relid, void *arg)
 	VacuumRelation *chunk_vacuum_rel;
 	RangeVar *chunk_range_var;
 
-	/* Skip vacuuming compressed chunks */
+	/* If the chunk has an associated compressed chunk, analyze that instead
+	 * When we compress a chunk, we save stats for the raw chunk, do
+	 * not modify that. Data now lives in the compressed chunk, so
+	 * analyze it.
+	 */
 	if (chunk->fd.compressed_chunk_id != INVALID_CHUNK_ID)
-		return;
-
-	chunk_range_var = copyObject(ctx->ht_vacuum_rel->relation);
-	chunk_range_var->relname = NameStr(chunk->fd.table_name);
-	chunk_range_var->schemaname = NameStr(chunk->fd.schema_name);
-	chunk_vacuum_rel =
-		makeVacuumRelation(chunk_range_var, chunk_relid, ctx->ht_vacuum_rel->va_cols);
-
+	{
+		Chunk *comp_chunk = ts_chunk_get_by_id(chunk->fd.compressed_chunk_id, true);
+		ChunkPair *cp = palloc(sizeof(ChunkPair));
+		cp->uncompressed_relid = chunk_relid;
+		cp->compressed_relid = comp_chunk->table_id;
+		ctx->chunk_pairs = lappend(ctx->chunk_pairs, cp);
+		/* analyze/vacuum the compressed rel instead */
+		chunk_vacuum_rel = makeVacuumRelation(NULL, comp_chunk->table_id, NIL);
+		ctx->chunk_rels = lappend(ctx->chunk_rels, chunk_vacuum_rel);
+	}
+	else
+	{
+		chunk_range_var = copyObject(ctx->ht_vacuum_rel->relation);
+		chunk_range_var->relname = NameStr(chunk->fd.table_name);
+		chunk_range_var->schemaname = NameStr(chunk->fd.schema_name);
+		chunk_vacuum_rel =
+			makeVacuumRelation(chunk_range_var, chunk_relid, ctx->ht_vacuum_rel->va_cols);
+	}
 	ctx->chunk_rels = lappend(ctx->chunk_rels, chunk_vacuum_rel);
 }
 
@@ -675,7 +720,7 @@ ts_get_all_vacuum_rels(bool is_vacuumcmd)
 	return vacrels;
 }
 
-/* Vacuums a hypertable and all of it's chunks */
+/* Vacuums/Analyzes a hypertable and all of it's chunks */
 static DDLResult
 process_vacuum(ProcessUtilityArgs *args)
 {
@@ -684,6 +729,7 @@ process_vacuum(ProcessUtilityArgs *args)
 	VacuumCtx ctx = {
 		.ht_vacuum_rel = NULL,
 		.chunk_rels = NIL,
+		.chunk_pairs = NIL,
 	};
 	ListCell *lc;
 	Hypertable *ht;
@@ -728,8 +774,16 @@ process_vacuum(ProcessUtilityArgs *args)
 					if (hypertable_is_distributed(ht))
 						continue;
 
-					ctx.ht_vacuum_rel = vacuum_rel;
-					foreach_chunk(ht, add_chunk_to_vacuum, &ctx);
+					if (ht->fd.compressed)
+					{
+						ctx.ht_vacuum_rel = vacuum_rel;
+						foreach_chunk(ht, add_compressed_chunk_to_vacuum, &ctx);
+					}
+					else
+					{
+						ctx.ht_vacuum_rel = vacuum_rel;
+						foreach_chunk(ht, add_chunk_to_vacuum, &ctx);
+					}
 				}
 			}
 			vacuum_rels = lappend(vacuum_rels, vacuum_rel);
@@ -752,8 +806,13 @@ process_vacuum(ProcessUtilityArgs *args)
 #endif
 			stmt,
 			is_toplevel);
+		foreach (lc, ctx.chunk_pairs)
+		{
+			ChunkPair *cp = (ChunkPair *) lfirst(lc);
+			ts_cm_functions->update_compressed_chunk_relstats(cp->uncompressed_relid,
+															  cp->compressed_relid);
+		}
 	}
-
 	return DDL_DONE;
 }
 
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -37,9 +37,11 @@
 #include "compat.h"
 
 #include "array.h"
+#include "chunk.h"
 #include "deltadelta.h"
 #include "dictionary.h"
 #include "gorilla.h"
+#include "compression_chunk_size.h"
 #include "create.h"
 #include "custom_type_cache.h"
 #include "segment_meta.h"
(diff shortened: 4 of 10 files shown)


src/process_utility.c
```diff
@@ -576,8 +576,38 @@ typedef struct VacuumCtx
 {
 	VacuumRelation *ht_vacuum_rel;
 	List *chunk_rels;
+	List *chunk_pairlst;
```
pmwkaa, line 579: `chunk_pairs` maybe?

src/process_utility.c
```diff
@@ -728,8 +757,16 @@ process_vacuum(ProcessUtilityArgs *args)
 					if (hypertable_is_distributed(ht))
 						continue;
 
-					ctx.ht_vacuum_rel = vacuum_rel;
-					foreach_chunk(ht, add_chunk_to_vacuum, &ctx);
+					if (ht->fd.compressed == true)
+					{
+						ctx.ht_vacuum_rel = vacuum_rel;
```
pmwkaa, line 762: Maybe move this statement up, since it is used in both if/else blocks?

tsl/src/compression/compression.c
```diff
@@ -1450,3 +1452,53 @@ compression_get_toast_storage(CompressionAlgorithms algorithm)
...
+update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
+{
+	bool update = true;
+	double rowcnt;
+	int comp_pages, uncomp_pages, comp_visible, uncomp_visible;
+	float comp_tuples, uncomp_tuples, out_tuples;
+	Chunk *uncompressed_chunk = ts_chunk_get_by_relid(uncompressed_relid, true);
+	Chunk *compressed_chunk = ts_chunk_get_by_relid(compressed_relid, true);
+
+	if (uncompressed_chunk->table_id != uncompressed_relid ||
```
pmwkaa, line 1469: Is this case possible, maybe `assert` instead?

src/process_utility.c
```diff
@@ -728,8 +757,16 @@ process_vacuum(ProcessUtilityArgs *args)
 					if (hypertable_is_distributed(ht))
 						continue;
 
-					ctx.ht_vacuum_rel = vacuum_rel;
-					foreach_chunk(ht, add_chunk_to_vacuum, &ctx);
+					if (ht->fd.compressed == true)
```
pmwkaa, line 760: nit: `if (ht->fd.compressed)`

tsl/src/compression/compression.c
```diff
@@ -1450,3 +1452,53 @@ compression_get_toast_storage(CompressionAlgorithms algorithm)
...
+						uncompressed_relid,
+						compressed_relid)));
+	}
+
+	/* we need page info from compressed relid */
+	capture_pgclass_stats(compressed_relid, &comp_pages, &comp_visible, &comp_tuples);
+	rowcnt = (double) get_before_compression_row_count(uncompressed_chunk->fd.id);
+	capture_pgclass_stats(uncompressed_relid, &uncomp_pages, &uncomp_visible, &uncomp_tuples);
+
+	if (uncomp_tuples == 0)
```
pmwkaa, line 1485: Maybe put a comment about the reasoning used here?

pmwkaa (commented):
looks good, just a couple of notes

tsl/src/compression/compression.c
gayyappan: want this to be generic safety check here. Should not happen with the  current code path unless there is some corruption.

src/compression_chunk_size.c
```diff
@@ -90,3 +90,34 @@ ts_compression_chunk_size_totals()
...
+		bool should_free;
+		HeapTuple tuple = ts_scan_iterator_fetch_heap_tuple(&iterator, false, &should_free);
+
+		heap_deform_tuple(tuple, ts_scan_iterator_tupledesc(&iterator), values, nulls);
+		if (!nulls[AttrNumberGetAttrOffset(Anum_compression_chunk_size_numrows_pre_compression)])
+			rowcnt = DatumGetInt64(values[AttrNumberGetAttrOffset(
+				Anum_compression_chunk_size_numrows_pre_compression)]);
+		if (should_free)
+			heap_freetuple(tuple);
+		found_cnt++;
```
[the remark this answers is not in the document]
gayyappan: Should not find > 1 as we are using the primary key. But we might not have a record, if the chunk is not compressed.

src/process_utility.c
gayyappan: modified.

tsl/src/compression/compression.c
gayyappan: added.

tsl/src/compression/compression.c
```diff
@@ -1450,3 +1452,58 @@ compression_get_toast_storage(CompressionAlgorithms algorithm)
...
+		if (rowcnt > 0)
+			out_tuples = (float4) rowcnt;
+		else
+			out_tuples = (float4) comp_tuples;
+	}
+	else
+	{
+		out_tuples = (float4) uncomp_tuples;
+		/* tuple count is unchanged. If num pages changed, we will update */
+		if (comp_pages == uncomp_pages)
```
WireBaron, line 1501: This doesn't seem right to me, as we'd expect comp_pages to be less than uncomp_pages here.  I'm not sure how this information is used in the planning, but given that it's passed up to the parent, I'd assume it's more correct to return the number of pages expected to be passed up by this child (so uncomp_pages).

Assuming I've convinced you that uncomp_pages is more correct here, I also think that if we're updating the number of rows (rowcnt != uncomp_tuples) then we should update the number of tuples to try to preserve the same tuple:page ratio.

src/process_utility.c
```diff
@@ -728,8 +757,16 @@ process_vacuum(ProcessUtilityArgs *args)
 					if (hypertable_is_distributed(ht))
 						continue;
 
-					ctx.ht_vacuum_rel = vacuum_rel;
-					foreach_chunk(ht, add_chunk_to_vacuum, &ctx);
+					if (ht->fd.compressed)
```
WireBaron, line 760: This code won't be run in the case of a database-wide analyze.  That case is handled by ts_get_all_vacuum_rels up above in the `stmt->rels == NIL` case.  You should modify that call to also add the the compressed chunk pairs (keep in mind that that call will iterate over both the compressed and uncompressed chunk tables).

WireBaron (changes requested):

tsl/src/compression/compression.c
```diff
@@ -1450,3 +1452,58 @@ compression_get_toast_storage(CompressionAlgorithms algorithm)
...
+	if (uncomp_tuples == 0)
+	{
+		if (rowcnt > 0)
+			out_tuples = (float4) rowcnt;
+		else
+			out_tuples = (float4) comp_tuples;
+	}
+	else
+	{
+		out_tuples = (float4) uncomp_tuples;
```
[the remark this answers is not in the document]
gayyappan: We would reach this path only in 2.0 as we set uncomp_tuples to a non zero value. If we delete rows from the compressed adta, it does not change rowcnt, but would modify comp_tuples. The comp_tuples count could be unexpectedly low (based on the compression ratio). I have chosen to over-estimate the output rows rather than under-estimate by retaining the old value for uncomp_tuples..

tsl/src/compression/compression.c
gayyappan: It would be equal on 2.0 due to the compress_chunk behavior that updates stats in 2.0? But we could vacuum the compressed hypertable and find that we now have fewer pages. That's the reason for the update. uncomp-apges might be 0 if it was vacuumed. Or it would be == comp_pages due to compress_chunk?

WireBaron (approved):

src/process_utility.c
```diff
@@ -752,8 +789,13 @@ process_vacuum(ProcessUtilityArgs *args)
 #endif
 			stmt,
 			is_toplevel);
+		foreach (lc, ctx.chunk_pairs)
+		{
+			chunk_pair *cp = (chunk_pair *) lfirst(lc);
+			ts_cm_functions->update_compressed_chunk_relstats(cp->uncompressed_relid,
+															  cp->compressed_relid);
```
k-rus, line 796: Since compression chunks cannot be updated, is it necessary to update compressed stat? Or is the name of the function misleading?

tsl/src/compression/compression.c
```diff
@@ -1450,3 +1452,47 @@ compression_get_toast_storage(CompressionAlgorithms algorithm)
 		elog(ERROR, "invalid compression algorithm %d", algorithm);
 	return definitions[algorithm].compressed_data_storage;
 }
+
+/* Get relstats from compressed chunk and insert into relstats for the
+ * corresponding uncompressed parent.
```
k-rus, line 1457: Does uncompressed parent mean the hypertable?

tsl/src/compression/compression.c
k-rus: Then why not assert?

tsl/src/compression/compression.c
```diff
@@ -1450,3 +1452,47 @@ compression_get_toast_storage(CompressionAlgorithms algorithm)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+				 errmsg("mismatched chunks for relstats update %d %d",
+						uncompressed_relid,
+						compressed_relid)));
+	}
+
+	/* we need page info from compressed relid */
+	capture_pgclass_stats(compressed_relid, &comp_pages, &comp_visible, &comp_tuples);
+	rowcnt = (double) ts_get_before_compression_row_count(uncompressed_chunk->fd.id);
```
k-rus, line 1481: I see that `rowcnt` is casted to `float4`. Can it be casted to `float4` already here?

tsl/src/compression/compression.c
```diff
@@ -1450,3 +1452,47 @@ compression_get_toast_storage(CompressionAlgorithms algorithm)
...
+	capture_pgclass_stats(uncompressed_relid, &uncomp_pages, &uncomp_visible, &uncomp_tuples);
+
+	/* Before compressing a chunk in 2.0, we save its stats. Prior
+	 * releases do not support this. So the stats on uncompressed relid
+	 * could be invalid. In this case, do the best that we can.
+	 * Alternately, the stats could match, don't update in that case.
+	 */
+	if (uncomp_tuples == 0)
```
k-rus, line 1489: The comment and the condition do not match. It doesn't update if the number of uncompressed tuples is set. However, it doesn't check if the number of uncompressed tuples is correct or not. What do I miss?

tsl/test/expected/compression.out
```diff
@@ -1363,13 +1363,116 @@ SELECT reloptions FROM pg_class WHERE relname = :statchunk;
...
+      relname       | reltuples | relpages | relallvisible 
+--------------------+-----------+----------+---------------
+ _hyper_29_58_chunk |       200 |        2 |             0
+ _hyper_29_59_chunk |         0 |        0 |             0
+(2 rows)
+
+\c :TEST_DBNAME :ROLE_SUPERUSER
+--overwrite pg_class stats for the compressed chunk.
+UPDATE pg_class
+SET reltuples = 0, relpages = 0
```
k-rus, line 1400: Is there a natural way that the value will be overwritten? Can it be overwritten to a number other than `0`?

k-rus (commented):
Few comments:
- It would good that the commit message explains what the commit does. This is unclear now.
- Is it possible that the number of uncompressed tuple will be incorrect after the chunk was compressed?

src/process_utility.c
k-rus: @pmwkaa This nit is a matter of taste. I can see that @erimatnor changes conditions of your suggestion into explicit conditions in https://github.com/timescale/timescaledb/pull/2625

src/compression_chunk_size.c
```diff
@@ -90,3 +90,34 @@ ts_compression_chunk_size_totals()
...
+	}
+	if (found_cnt != 1)
+		elog(ERROR,
+			 "missing record for chunk with id %d in %s",
```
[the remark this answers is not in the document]
gayyappan: This is a primary key.

src/process_utility.c
```diff
@@ -590,13 +620,11 @@ add_chunk_to_vacuum(Hypertable *ht, Oid chunk_relid, void *arg)
 	/* Skip vacuuming compressed chunks */
 	if (chunk->fd.compressed_chunk_id != INVALID_CHUNK_ID)
 		return;
```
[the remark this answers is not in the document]
gayyappan: No, this will be called when you pass in the hypertable. This skips the chunk if it has an associated compressed chunk.

src/process_utility.c
gayyappan: The purpose of calling analyze on compression hypertable is to update the relstats on that table. This would be called only if someone wants to explicitly do so (  e.g. upgrade to 2.0, we don't have stats on the compressed chunks. Now call analyze <_timescaledb_internal.compression_table> . This will update stats . This is the primary use case for this fix).

tsl/src/compression/compression.c
gayyappan: uncompressed parent = chunk for which this is the associated compressed chunk.

src/process_utility.c
gayyappan: more details about this in compression.c file changes.

tsl/src/compression/compression.c
gayyappan: We have no way of verifying the number of uncompressed tuples without actually decompressing the chunk.

src/compression_chunk_size.c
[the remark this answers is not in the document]
gayyappan: > Few comments:
> 
> * It would good that the commit message explains what the commit does. This is unclear now.
I am not sure what you mean. This commit supports ANALYZE of an compressed hypertable. The commit message says so.
> * Is it possible that the number of uncompressed tuple will be incorrect after the chunk was compressed?
Only in 2.0, we added support to save number of tuples before compression. So it would be an estimate and not exact in other cases and/or if rows were deleted from the chunk.

gayyappan:
> Few comments:
> 
> * It would good that the commit message explains what the commit does. This is unclear now.
I am not sure what you mean. This commit supports ANALYZE of an compressed hypertable. The commit message says so.
> * Is it possible that the number of uncompressed tuple will be incorrect after the chunk was compressed?
 we added support to save number of tuples before compression in 2.0.  The number would be an estimate and not exact in other cases and/or if rows were deleted from the chunk.

tsl/test/expected/compression.out
gayyappan: yes,  it would be 0 , if you are on a db from pre 2.0

src/compression_chunk_size.c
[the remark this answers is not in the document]
k-rus: Then I suggest to check that `found_cnt == 0` and assert for `found_cnt == 1` after the error. Or keep current condition and assert for `found_cnt == 0` before the printing the error.

I am surprise that you implemented the loop for primary key access, instead of just reading singe value, which will be matching by definition.

src/process_utility.c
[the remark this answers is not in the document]
k-rus: Is it possible to have compressed chunks in uncompressed hypertable? `add_chunk_to_vacuum` is only called when `ht->fd.compressed` is false.

src/process_utility.c
k-rus: Thank you for clarifying!

src/process_utility.c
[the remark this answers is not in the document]
k-rus: Now I realised that uncompressed hypertable will have chunks with compression pair, thus they are ignored. I forgot about it. Sorry, for confusion.

tsl/src/compression/compression.c
k-rus: This is confusing terminology. If I don't mistake compressed chunk inherits from compressed hypertable and the parent usually means the object to inherit from.

tsl/src/compression/compression.c
k-rus: Then I suggest to improve the comment. E.g.,
```suggestion
	 * Alternately, the stats could already exist, don't update in that case.
	 */
	if (uncomp_tuples == 0)
```

tsl/src/compression/compression.c
gayyappan: modified comment.

tsl/src/compression/compression.c
gayyappan: modified.

k-rus:
> * It would good that the commit message explains what the commit does. This is unclear now.
>   I am not sure what you mean. This commit supports ANALYZE of an compressed hypertable. The commit message says so.

Sorry, for not being clear. Now I see that the commit message describes how it works after the commit. It would be nice to make it explicit, e..g, by just saying, `after this fix/commit`, or using future tense.

tsl/test/sql/compression.sql
```diff
@@ -559,9 +570,80 @@ SELECT reloptions FROM pg_class WHERE relname = :statchunk;
...
+UPDATE pg_class
+SET reltuples = 0, relpages = 0
+ WHERE relname in ( SELECT ch.table_name FROM 
+    _timescaledb_catalog.chunk ch, 
+    _timescaledb_catalog.hypertable ht
+  WHERE ht.table_name = 'stattest2' AND ch.hypertable_id = ht.id 
+        AND ch.compressed_chunk_id > 0 );
+\c :TEST_DBNAME :ROLE_DEFAULT_PERM_USER
+
+SELECT relname, reltuples, relpages, relallvisible FROM pg_class
```
erimatnor, line 605: I suggest making this a view or similar for convenience since it is used repeatedly.

tsl/test/sql/compression.sql
```diff
@@ -559,9 +570,80 @@ SELECT reloptions FROM pg_class WHERE relname = :statchunk;
...
+  WHERE ht.table_name = 'stattest2' AND ch.hypertable_id = ht.id )
+order by relname;
+
+SELECT '_timescaledb_internal.' || compht.table_name as "STAT_COMP_TABLE",
+             compht.table_name  as "STAT_COMP_TABLE_NAME"
+FROM _timescaledb_catalog.hypertable ht, _timescaledb_catalog.hypertable compht
+WHERE ht.table_name = 'stattest2' AND ht.compressed_hypertable_id = compht.id \gset
+
+--analyze the compressed table, will update stats for the raw table.
+ANALYZE :STAT_COMP_TABLE;
```
erimatnor, line 617: The main feature/test case seems to be missing.

Normally, the user runs ANALYZE on the main hypertable, not the internal compressed table. This is the case we really care about. We want to make sure that ANALYZE on the main hypertable updates the row count for the parent chunks based on the internal compressed chunk.


While it is OK to test also ANALYZE on the internal compressed table, this is not the main case we want to work.

gayyappan: @erimatnor We already have a test case for ANALYZE <main table> just above this. Added by another PR that made changes to this area.

erimatnor:
Mostly nits on my end, but I am mostly concerned about having no test for the main case we want to support: namely running ANALYZE on the base hypertable (not the internal one) and see that the row count on the parent (empty) chunks are updated based on the information about the compressed chunks.

src/process_utility.c
```diff
@@ -576,8 +576,38 @@ typedef struct VacuumCtx
...
 	VacuumRelation *ht_vacuum_rel;
 	List *chunk_rels;
+	List *chunk_pairs;
 } VacuumCtx;
 
+typedef struct chunk_pair
+{
+	Oid uncompressed_relid;
+	Oid compressed_relid;
+} chunk_pair;
```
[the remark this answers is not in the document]
gayyappan: fixed.

src/compression_chunk_size.c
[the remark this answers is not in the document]
gayyappan: Maybe I am missing something here. But  how do you read without the iterator?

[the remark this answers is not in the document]
erimatnor: I think the main concern is the loop, not the iterator itself. Either use the PG scan API directly, our scan API. We actually already have a `ts_scanner_scan_one` function. With the iterator we can do something similar `ts_scan_iterator_get_one(&iterator)`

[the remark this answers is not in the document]
erimatnor: For instance, this could be added to `scan_iterator.h`:

```
static inline TupleInfo *
ts_scan_iterator_get_one(ScanIterator *iterator)
{
	TupleInfo *ti;
	
	ts_scanner_start_scan(&iterator->ctx, &iterator->ictx);
	ti = ts_scanner_next(&iterator->ctx, &iterator->ictx);
	
	if (ts_scanner_next(&iterator->ctx, &iterator->ictx) != NULL)
		elog(ERROR, "more than one tuple when one expected");
	
	return ti;
}
```

src/compression_chunk_size.h
```diff
@@ -21,5 +21,6 @@ typedef struct TotalSizes
 } TotalSizes;
 
 extern TSDLLEXPORT TotalSizes ts_compression_chunk_size_totals(void);
+extern TSDLLEXPORT int64 ts_get_before_compression_row_count(int32 uncompressed_chunk_id);
```
erimatnor, line 24: This function is still not consistent with the other two public functions in this module.

src/process_utility.c
erimatnor: If it matters, I have a habit of using explicit conditions for pointers, since that makes it clear that one is checking a pointer as opposed to a boolean. I have rarely, if ever, used explicit conditions for booleans.

erimatnor (approved):

src/compression_chunk_size.h
gayyappan: renamed.

PR #2612, opened 2020-11-02
Optimize internal cagg_watermark function

### LINKED ISSUE #2602: Continuous Aggregates watermark function can cause long scans on dates not yet materialized
Using official TimescaleDB Docker image for 2.0.0-rc2
 * Postgres 12.4
 * TimescaleDB 2.0.0-RC2
 * Docker

**Describe the bug**
Followed the tutorial for Continuous [Aggregates](https://docs.timescale.com/v2.0/tutorials/continuous-aggs-tutorial) but changing the SQL for the new 2.0 functions. The dataset for the tutorial is from Jan 2016 only (11 million rows). Upon creating the materialized view using `WITH NO DATA` as recommended with 2.0, attempting to query data from the view that is not yet materialized causes the query to scan in ways that result in a very long-running query.

If data is materialized manually for any period of Jan 2016, querying those date ranges specifically return quickly as expected. Querying any dates beyond that range (even including the materialized range) still results in long query scans.

The queries will run as expected if any of the following are done:
 * set to `materialized_only = true` on the view
 * create the view without specifying `WITH NO DATA` (data is materialized at runtime)
 * create a refresh policy with a `start_offset = NULL` (not a desirable long-term option) 

Upon further internal discussion, we believe this is tied to the `ts_continuous_agg_watermark` function, based on feedback from @gayyappan. 

**To Reproduce**
1. Follow the Continuous Aggregate tutorial linked above to create tables and import taxi data from Jan 2016.
2. For 2.0, the `CREATE MATERIALIZED VIEW` command I used was:

```sql
CREATE MATERIALIZED VIEW cagg_rides_view WITH
  (timescaledb.continuous)
AS
SELECT vendor_id, time_bucket('1h', pickup_datetime) as day,
  count(*) total_rides,
  avg(fare_amount) avg_fare,
  max(trip_distance) as max_trip_distance,
  min(trip_distance) as min_trip_distance
FROM rides
GROUP BY vendor_id, time_bucket('1h', pickup_datetime)
WITH NO DATA;
```
3. Run a simple `COUNT(*)` query on the data.

```sql
SELECT COUNT(*) FROM cagg_rides_view;
```

The query will run for a long time... maybe even not finish.

4. Now, manually refresh a portion of the Jan 2016 results, from Jan 1 to Jan 7

```sql
CALL refresh_continuous_aggregate('cagg_rides_view'::regclass,'2016-01-01','2016-01-07');
```

5. Run the same `COUNT(*)` query and it will still run for a long time:
```sql
SELECT COUNT(*) FROM cagg_rides_view;
```

6. Now add a `day` predicate that includes the materialized range and the query returns "instantly"

```sql
SELECT COUNT(*) FROM cagg_rides_view
WHERE day > '2016-01-01' AND day < '2016-01-08'
```

7. As you begin to expand that range of dates outside of the materialized area, the query time increases dramatically again:
```sql
SELECT COUNT(*) FROM cagg_rides_view
WHERE day > '2016-01-01' AND day < '2016-01-15'
```

**Expected behavior**
In TimescaleDB 1.7.4, creating the Continuous Aggregate with the `VIEW` function materialized the data at creation and set a refresh policy. All of the queries above work as expected even when the real-time aggregate data is added to the mix. The planner should be able to understand that there is only data for Jan 2016 in the hypertable and easily perform an unbounded query across that date ranges.

**Actual behavior**
If data is not properly materialized in 2.0 at creation time (or shortly thereafter), any queries to CAggs that involve real-time aggregates appear to mess with the planner in a way that causes scans. 

As long as data is materialized and up-to-date from the engine's perspective (it appears), the queries return as I would expect.

**Additional context**
Obviously this isn't the most real-world scenario because data is during a fixed time period. Two other methods for getting this to "work" as expected include:

1. Materialize the view `WITH NO DATA` and then add an  aggregation policy that has a `start_offset = NULL` for the CAgg policy. This does work, but obviously we don't want a `start_offset` set like that long-term. If you do this, when the policy runs, it will aggregate "all of time" and the planner seems to be happy then.

2. Materialize the view `WITH NO DATA` and set a reasonable aggregation policy, maybe `start_offset = 2h`. Then, manually refresh the old data up to the present. This also appears to set whatever metadata the watermark function is looking at and queries work as expected.

erimatnor:
The internal function `cagg_watermark` is often used as part of an
index scan condition and is therefore called on every row of such a
scan. Ideally, the planner should be able to reuse the result of
repeated calls of the same function since it is marked STABLE, but
this doesn't seem to be the case for C functions, making scans very
slow.

To deal with this, we cache the watermark value in the `fn_extra`
field of the function call info context. This avoids row-by-row max
value calculations, which otherwise makes things very slow.

The function could be optimized further by making the scan for the max
value faster (e.g., by doing an internal scan for max instead of
relying on SPI). However, by caching the max value, this is already
faster than the corresponding watermark function in 1.7.x.

Fixes #2602

### DIFF
--- a/src/continuous_agg.c
+++ b/src/continuous_agg.c
@@ -862,6 +862,12 @@ ts_continuous_agg_find_integer_now_func_by_materialization_id(int32 mat_htid)
 	return par_dim;
 }
 
+typedef struct Watermark
+{
+	int32 hyper_id;
+	int64 value;
+} Watermark;
+
 TS_FUNCTION_INFO_V1(ts_continuous_agg_watermark);
 
 /*
@@ -886,21 +892,36 @@ ts_continuous_agg_watermark(PG_FUNCTION_ARGS)
 	Dimension *dim;
 	Datum maxdat;
 	bool max_isnull;
-	int64 watermark;
+	Watermark *watermark = (Watermark *) fcinfo->flinfo->fn_extra;
 	Oid timetype;
 	AclResult aclresult;
 
 	if (PG_ARGISNULL(0))
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-				 errmsg("invalid continuous aggregate hypertable")));
+				 errmsg("materialized hypertable cannot be NULL")));
+
+	if (watermark != NULL)
+	{
+		/* The flinfo is probably reset when the input argument changes, but
+		 * the documentation isn't clear on this. Check that the hypertable ID
+		 * matches across repeated calls just to be safe. */
+		if (watermark->hyper_id == hyper_id)
+			PG_RETURN_INT64(watermark->value);
+
+		pfree(watermark);
+	}
 
 	cagg = ts_continuous_agg_find_by_mat_hypertable_id(hyper_id);
 
 	if (NULL == cagg)
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-				 errmsg("%d is not a materialized hypertable", hyper_id)));
+				 errmsg("invalid materialized hypertable ID: %d", hyper_id)));
+
+	watermark = MemoryContextAllocZero(fcinfo->flinfo->fn_mcxt, sizeof(Watermark));
+	watermark->hyper_id = hyper_id;
+	fcinfo->flinfo->fn_extra = watermark;
 
 	/* Preemptive permission check to ensure the function complains about lack
 	 * of permissions on the cagg rather than the materialized hypertable */
@@ -915,15 +936,17 @@ ts_continuous_agg_watermark(PG_FUNCTION_ARGS)
 
 	if (!max_isnull)
 	{
+		int64 value;
+
 		/* Add one bucket to get to the end of the last bucket */
-		watermark = ts_time_value_to_internal(maxdat, timetype);
-		watermark = ts_time_saturating_add(watermark, cagg->data.bucket_width, timetype);
+		value = ts_time_value_to_internal(maxdat, timetype);
+		watermark->value = ts_time_saturating_add(value, cagg->data.bucket_width, timetype);
 	}
 	else
 	{
 		/* Nothing materialized, so return min */
-		watermark = ts_time_get_min(timetype);
+		watermark->value = ts_time_get_min(timetype);
 	}
 
-	PG_RETURN_INT64(watermark);
+	PG_RETURN_INT64(watermark->value);
 }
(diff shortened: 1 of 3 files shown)


src/continuous_agg.c
```diff
@@ -886,21 +892,36 @@ ts_continuous_agg_watermark(PG_FUNCTION_ARGS)
...
-				 errmsg("invalid continuous aggregate hypertable")));
+				 errmsg("materialized hypertable cannot be NULL")));
+
+	if (watermark != NULL)
+	{
+		/* The flinfo is probably reset when the input argument changes, but
+		 * the documentation isn't clear on this. Check that the hypertable ID
+		 * matches across repeated calls just to be safe. */
+		if (watermark->hyper_id == hyper_id)
+			PG_RETURN_INT64(watermark->value);
```
pmwkaa, line 910: Last allocation will not be freed here I guess and live till fn_mcxt life time?

src/continuous_agg.c
```diff
@@ -886,21 +892,36 @@ ts_continuous_agg_watermark(PG_FUNCTION_ARGS)
 	Dimension *dim;
 	Datum maxdat;
 	bool max_isnull;
-	int64 watermark;
+	Watermark *watermark = (Watermark *) fcinfo->flinfo->fn_extra;
```
pmwkaa, line 895: Not really obvious the lifetime of this. For example, will this field always be NULL when this function is called directly? (in tests for example).

pmwkaa (approved):
Feels like this caching logic should happen outside this function by the function user, not really sure it is possible to achieve in a simple way. Maybe this function should be treated as an aggregate in such case? to properly handle context, if makes sense.

How it tested?

k-rus (commented):
LGTM
Can a test be added, which demonstrate correct reuse of watermark in multiple calls to this function in the same query? Or do you see other approach to prevent regression in future?

src/continuous_agg.c
erimatnor: `fn_extra` is always NULL on first call of the function. https://www.postgresql.org/docs/current/plhandler.html

src/continuous_agg.c
erimatnor: It only has to free if re-allocated. Last call will free automatically when the memory context is destroyed.

erimatnor (commented):
The function is already tested as part of existing test suites. It is called repeatedly in union view queries as part of index scans. If it did the wrong thing, the existing tests would fail. This is purely an optimization and I tested it manually on a large dataset (the cagg tutorial) and it works as well or better than 1.7.x.

k-rus:
@erimatnor By test I mean that the function is tested on reusing the watermark. I guess I can treat your reply as it is difficult to test this optimisation from regressing in the future.

k-rus (approved):

erimatnor:
> @erimatnor By test I mean that the function is tested on reusing the watermark. I guess I can treat your reply as it is difficult to test this optimisation from regressing in the future.

yeah, I don't know how to test this without somehow adding debug output to show that is using a cached value as opposed to calculating the max again. I did test the performance manually on a big dataset, as mentioned previously.

PR #2613, opened 2020-11-03
Refresh correct partial during refresh on drop

### LINKED ISSUE #2592: Refresh on drop can create outliers in continuous aggregates

k-rus:
When drop_chunks is called on a hypertable manually or through
a retention policy, it initiates refresh in hypertable's continuous
aggregates. The refresh includes all buckets of each dropped chunk.
Some buckets are inside a dropped chunk only partially. In such case,
this commit fixes to refresh only the correct partials of the buckets.

The fix passes chunk id to refresh on drop of a chunk. The chunk id is
used to identify and update correct partials. This is a workaround
that the invalidation outside of the refreshed chunk doesn't update
the other partial, which is outside the refreshed chunk.

The commit also adds the test demonstrating correct refresh.

Fixes #2592

### DIFF
--- a/tsl/src/continuous_aggs/materialize.c
+++ b/tsl/src/continuous_aggs/materialize.c
@@ -34,18 +34,20 @@ static Datum internal_to_time_value_or_infinite(int64 internal, Oid time_type,
 
 static void spi_update_materializations(SchemaAndName partial_view,
 										SchemaAndName materialization_table, Name time_column_name,
-										TimeRange invalidation_range);
+										TimeRange invalidation_range, int32 chunk_id);
 static void spi_delete_materializations(SchemaAndName materialization_table, Name time_column_name,
-										TimeRange invalidation_range);
+										TimeRange invalidation_range, const char *chunk_condition);
 static void spi_insert_materializations(SchemaAndName partial_view,
 										SchemaAndName materialization_table, Name time_column_name,
-										TimeRange materialization_range);
+										TimeRange materialization_range,
+										const char *chunk_condition);
 
 void
 continuous_agg_update_materialization(SchemaAndName partial_view,
 									  SchemaAndName materialization_table, Name time_column_name,
 									  InternalTimeRange new_materialization_range,
-									  InternalTimeRange invalidation_range, int64 bucket_width)
+									  InternalTimeRange invalidation_range, int64 bucket_width,
+									  const int32 chunk_id)
 {
 	InternalTimeRange combined_materialization_range = new_materialization_range;
 	bool materialize_invalidations_separately = range_length(invalidation_range) > 0;
@@ -90,19 +92,22 @@ continuous_agg_update_materialization(SchemaAndName partial_view,
 									materialization_table,
 									time_column_name,
 									internal_time_range_to_time_range(
-										combined_materialization_range));
+										combined_materialization_range),
+									chunk_id);
 	}
 	else
 	{
 		spi_update_materializations(partial_view,
 									materialization_table,
 									time_column_name,
-									internal_time_range_to_time_range(invalidation_range));
+									internal_time_range_to_time_range(invalidation_range),
+									chunk_id);
 
 		spi_update_materializations(partial_view,
 									materialization_table,
 									time_column_name,
-									internal_time_range_to_time_range(new_materialization_range));
+									internal_time_range_to_time_range(new_materialization_range),
+									chunk_id);
 	}
 
 	res = SPI_finish();
@@ -200,18 +205,33 @@ internal_time_range_to_time_range(InternalTimeRange internal)
 
 static void
 spi_update_materializations(SchemaAndName partial_view, SchemaAndName materialization_table,
-							Name time_column_name, TimeRange invalidation_range)
+							Name time_column_name, TimeRange invalidation_range,
+							const int32 chunk_id)
 {
-	spi_delete_materializations(materialization_table, time_column_name, invalidation_range);
+	StringInfo chunk_condition = makeStringInfo();
+
+	/* chunk_id is valid if the materializaion update should be done only the given chunk.
+	 * This is used currently for refresh on chunk drop only. In other cases, manual
+	 * call to refresh_continuous_aggregate or call from a refresh policy, chunk_id is
+	 * not provided, i.e., invlaid.
+	 */
+	if (chunk_id != INVALID_CHUNK_ID)
+		appendStringInfo(chunk_condition, "AND chunk_id = %d", chunk_id);
+
+	spi_delete_materializations(materialization_table,
+								time_column_name,
+								invalidation_range,
+								chunk_condition->data);
 	spi_insert_materializations(partial_view,
 								materialization_table,
 								time_column_name,
-								invalidation_range);
+								invalidation_range,
+								chunk_condition->data);
 }
 
 static void
 spi_delete_materializations(SchemaAndName materialization_table, Name time_column_name,
-							TimeRange invalidation_range)
+							TimeRange invalidation_range, const char *const chunk_condition)
 {
 	int res;
 	StringInfo command = makeStringInfo();
@@ -226,13 +246,14 @@ spi_delete_materializations(SchemaAndName materialization_table, Name time_colum
 
 	appendStringInfo(command,
 					 "DELETE FROM %s.%s AS D WHERE "
-					 "D.%s >= %s AND D.%s < %s;",
+					 "D.%s >= %s AND D.%s < %s %s;",
 					 quote_identifier(NameStr(*materialization_table.schema)),
 					 quote_identifier(NameStr(*materialization_table.name)),
 					 quote_identifier(NameStr(*time_column_name)),
 					 quote_literal_cstr(invalidation_start),
 					 quote_identifier(NameStr(*time_column_name)),
-					 quote_literal_cstr(invalidation_end));
+					 quote_literal_cstr(invalidation_end),
+					 chunk_condition);
 
 	res = SPI_execute_with_args(command->data,
 								0 /*=nargs*/,
@@ -247,7 +268,8 @@ spi_delete_materializations(SchemaAndName materialization_table, Name time_colum
 
 static void
 spi_insert_materializations(SchemaAndName partial_view, SchemaAndName materialization_table,
-							Name time_column_name, TimeRange materialization_range)
+							Name time_column_name, TimeRange materialization_range,
+							const char *const chunk_condition)
 {
 	int res;
 	StringInfo command = makeStringInfo();
@@ -262,7 +284,7 @@ spi_insert_materializations(SchemaAndName partial_view, SchemaAndName materializ
 
 	appendStringInfo(command,
 					 "INSERT INTO %s.%s SELECT * FROM %s.%s AS I "
-					 "WHERE I.%s >= %s AND I.%s < %s;",
+					 "WHERE I.%s >= %s AND I.%s < %s %s;",
 					 quote_identifier(NameStr(*materialization_table.schema)),
 					 quote_identifier(NameStr(*materialization_table.name)),
 					 quote_identifier(NameStr(*partial_view.schema)),
@@ -270,7 +292,8 @@ spi_insert_materializations(SchemaAndName partial_view, SchemaAndName materializ
 					 quote_identifier(NameStr(*time_column_name)),
 					 quote_literal_cstr(materialization_start),
 					 quote_identifier(NameStr(*time_column_name)),
-					 quote_literal_cstr(materialization_end));
+					 quote_literal_cstr(materialization_end),
+					 chunk_condition);
 
 	res = SPI_execute_with_args(command->data,
 								0 /*=nargs*/,
--- a/tsl/test/expected/continuous_aggs_drop_chunks.out
+++ b/tsl/test/expected/continuous_aggs_drop_chunks.out
@@ -139,3 +139,136 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
  _timescaledb_internal._hyper_1_15_chunk
 (15 rows)
 
+\set VERBOSITY terse
+DROP MATERIALIZED VIEW records_monthly;
+NOTICE:  drop cascades to 32 other objects
+DROP TABLE records;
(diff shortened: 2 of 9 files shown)


tsl/src/continuous_aggs/materialize.c
```diff
@@ -200,18 +205,27 @@ internal_time_range_to_time_range(InternalTimeRange internal)
...
-							Name time_column_name, TimeRange invalidation_range)
+							Name time_column_name, TimeRange invalidation_range,
+							const int32 chunk_id)
 {
-	spi_delete_materializations(materialization_table, time_column_name, invalidation_range);
+	StringInfo chunk_condition = makeStringInfo();
+
+	if (chunk_id != INVALID_CHUNK_ID)
+		appendStringInfo(chunk_condition, "AND chunk_id = %d", chunk_id);
+	spi_delete_materializations(materialization_table,
```
erimatnor, line 215: I recommend a newline between if clause and the following statements

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,136 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
  _timescaledb_internal._hyper_1_15_chunk
 (15 rows)
 
+\set VERBOSITY terse
+DROP MATERIALIZED VIEW records_monthly;
+NOTICE:  drop cascades to 32 other objects
+DROP TABLE records;
+DROP TABLE clients;
+\set VERBOSITY default
+-- Test how caggs are afected by drop_chunk due to refresh on drop
```
erimatnor, line 148:
```suggestion
-- Test how caggs are affected by drop_chunk due to refresh on drop
```

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,136 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+     28 | 21.98 |     7
+     35 | 21.98 |     7
+     42 | 21.98 |     7
+     49 | 21.98 |     7
+     56 | 21.98 |     7
+     63 | 21.98 |     7
+(10 rows)
+
+-- This is the simplest case, when the updated bucket is inside chunk, so it is expected
+-- always reflected.
```
erimatnor, line 197: I don't understand this comment. Specifically, "...so it is expected always reflected"

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,136 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+     28 | 21.98 |     7
+     35 | 21.98 |     7
+     42 | 21.98 |     7
+     49 | 21.98 |     7
+     56 | 21.98 |     7
+     63 | 21.98 |     7
+(10 rows)
+
+-- This case is update in a bucket, which crosses two chunks, which exist and both is going
+-- to be dropped.
```
erimatnor, line 222:
```suggestion
--  Test an UPDATE that affects a bucket that is split across two chunks. Both chunks are going to be dropped.
```

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,136 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+UPDATE conditions SET value = 4.00 WHERE time_int = 31;
+-- This update is in the bucket, which crosses two chunks. The first chunk is going to be 
+-- dropped now and the update is outside it, and thus should not be reflected. The second 
+-- chunk contains the update and will be dropped on the next call, thus it should be reflected 
+-- in the continuous aggregate only after the last drop.
+UPDATE conditions SET value = 4.00 WHERE time_int = 41;
+-- Now the update in 31 will be present, but not the update in 41.
```
erimatnor, line 229:
```suggestion
-- After drop_chunks, which refreshes the cagg, the update in 31 will be visible, but not the update in 41.
```

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,136 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+(10 rows)
+
+-- This case is update in a bucket, which crosses two chunks, which exist and both is going
+-- to be dropped.
+UPDATE conditions SET value = 4.00 WHERE time_int = 31;
+-- This update is in the bucket, which crosses two chunks. The first chunk is going to be 
+-- dropped now and the update is outside it, and thus should not be reflected. The second 
+-- chunk contains the update and will be dropped on the next call, thus it should be reflected 
+-- in the continuous aggregate only after the last drop.
+UPDATE conditions SET value = 4.00 WHERE time_int = 41;
```
erimatnor, line 228: Isn't a more interesting test case one where you update the _same_ bucket twice, but each update falls within a separate partial. Then when you drop the first chunk, the first partial of bucket X will be updated, but not the second partial of bucket X. For instance,

```
UPDATE conditions SET value = 4.00 WHERE time_int = 29;
UPDATE conditions SET value = 4.00 WHERE time_int = 32;
```

These two updates fall in the same bucket `[28,35)`, but are two different partials since the chunk border is at 30.

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,136 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+     21 | 21.98 |     7
+     28 | 22.84 |     7
+     35 | 21.98 |     7
+     42 | 21.98 |     7
+     49 | 21.98 |     7
+     56 | 21.98 |     7
+     63 | 21.98 |     7
+(10 rows)
+
+-- Now the update the update in 41 will be also present.
```
erimatnor, line 252:
```suggestion
-- Now the update in 41 will be also be visible.
```

tsl/test/expected/continuous_aggs_drop_chunks.out
k-rus: It is not more interested than the current tests. It doesn't allow to see if the fix recognises, which partials need to be updated, and this test was already working correctly before this fix.
I will add it.

erimatnor: Maybe I am missing something, but if a bucket can be split across two partials (one in each chunk), it seems reasonable to test updates in those two same-bucket partials to see that both cases are properly reflected in the bucket when dropping the corresponding chunks.

For instance, bucket 2 below is split across two chunks, thus consisting of two partials 2a and 2b. When dropping the first chunk we'd like to see that bucket 2 is updated to reflect changes in chunk 1 and when the second chunk is dropped we'd like to see that bucket 2 is updated to reflect changes in chunk 2.

```
  1    2     3
|----|-----|----| buckets
       a|b
|-------|-------| chunks
```

tsl/test/sql/continuous_aggs_drop_chunks.sql
```diff
@@ -51,3 +51,67 @@ CALL refresh_continuous_aggregate('records_monthly', NULL, NULL);
...
+SELECT time_val, time_val % 4, 3.14 FROM generate_series(0,100,1) AS time_val;
+
+SELECT set_integer_now_func('conditions', 'test_int_now');
+CREATE MATERIALIZED VIEW conditions_7
+    WITH (timescaledb.continuous, timescaledb.materialized_only = TRUE)
+    AS
+        SELECT time_bucket(7, time_int) as bucket,
+            SUM(value), COUNT(value)
+        FROM conditions GROUP BY bucket WITH DATA;
+SELECT materialization_hypertable_schema||'.'||materialization_hypertable_name AS mat_hyper
```
erimatnor, line 80: I don't think `mat_hyper` is ever used. Is this query used at all?

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,139 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+    $$ SELECT 125 $$;
+CREATE TABLE conditions(time_int INT NOT NULL, device INT, value FLOAT);
+SELECT create_hypertable('conditions', 'time_int', chunk_time_interval => 10);
+    create_hypertable    
+-------------------------
+ (3,public,conditions,t)
+(1 row)
+
+INSERT INTO conditions
+SELECT time_val, time_val % 4, 3.14 FROM generate_series(0,100,1) AS time_val;
```
erimatnor, line 160: Would be nice to use the new chunks view to show the ranges of the chunks before dropping them so that one can easily compare the ranges to the buckets in the cagg.

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,139 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+     42 | 21.98 |     7
+     49 | 21.98 |     7
+     56 | 21.98 |     7
+     63 | 21.98 |     7
+(10 rows)
+
+-- This is the simplest case, when the updated bucket is inside a chunk, so it is expected
+-- that the update is always reflected after the refresh on drop.
+UPDATE conditions SET value = 4.00 WHERE time_int = 2;
+-- This case updates values in the bucket, which affects two different partials in
```
erimatnor, line 197: The test case I was referring to previously was to drop these chunks separately, to see that the _same_ bucket is updated twice; first when dropping the first chunk, then updated again when dropping the second chunk.

tsl/test/expected/continuous_aggs_drop_chunks.out
k-rus: The described case is already producing correct result before this fix because it was refreshing bucket 2 two times, so the test doesn't demonstrated the corner case described by the issue.
This PR fixes that 2b is refreshed during drop of chunk b, and 2a is refreshed during drop of chunk a.  The regression test was already testing the most interesting cases. I added the test as you requested.

tsl/test/sql/continuous_aggs_drop_chunks.sql
k-rus: Thank you! It is left over after I queries the materialized table in the view, but it was redundant after I improved tests to cover corner cases better. I will remove this.

tsl/test/expected/continuous_aggs_drop_chunks.out
```diff
@@ -139,3 +139,139 @@ SELECT drop_chunks('records', '2000-03-16'::timestamptz);
...
+-----------------------------------------
+ _timescaledb_internal._hyper_3_65_chunk
+ _timescaledb_internal._hyper_3_66_chunk
+(2 rows)
+
+SELECT * FROM see_cagg;
+ bucket |  sum  | count 
+--------+-------+-------
+      0 | 22.84 |     7
+      7 |  23.7 |     7
```
erimatnor, line 212: If you drop the two first chunks separately, I would expect bucket 7 to first refresh to 22.84 and then when the second chunk is dropped it would be 23.7. Why not test that case as I suggested?

erimatnor: Ohh, I think this is tested below... sorry

tsl/test/expected/continuous_aggs_drop_chunks.out
k-rus: No, your suggestion was to drop from the chunks, which are dropped together. See your comment with your example: https://github.com/timescale/timescaledb/pull/2613#discussion_r516608352

tsl/test/expected/continuous_aggs_drop_chunks.out
k-rus: Your case is already tested and I commented two times that I am adding it. Please, check the code

erimatnor (approved):

tsl/test/expected/continuous_aggs_drop_chunks.out
erimatnor: My suggestion was to change the test  so that the chunks are not dropped together. But I see that the requested case is now covered in a different way

k-rus: I was mislead by your example, which was in the bucket of the chunks dropped together. Yes, the other test was also added.

gayyappan (approved):
Comments requested.

tsl/src/continuous_aggs/materialize.c
```diff
@@ -34,18 +34,20 @@ static Datum internal_to_time_value_or_infinite(int64 internal, Oid time_type,
...
-										TimeRange materialization_range);
+										TimeRange materialization_range,
+										const char *chunk_condition);
 
 void
 continuous_agg_update_materialization(SchemaAndName partial_view,
 									  SchemaAndName materialization_table, Name time_column_name,
 									  InternalTimeRange new_materialization_range,
-									  InternalTimeRange invalidation_range, int64 bucket_width)
+									  InternalTimeRange invalidation_range, int64 bucket_width,
```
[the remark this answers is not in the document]
k-rus: Added a comment

PR #2615, opened 2020-11-03
Ignore PostgreSQL's timetz test

erimatnor:
PostgreSQL's `timetz` test fails outside of daylight savings. The
test is ignored until fixed in PostgreSQL (likely to happen in
12.5).


pmwkaa (approved):

k-rus (changes requested):
Test `time` doesn't fail. Only `timetz` fails.

k-rus (approved):

PR #2617, opened 2020-11-03
Fix validation of available extensions on data node

niksajakovljevic:
We want to check all available extension versions
and not just the installed one. This is because we
might be setting up a cluster for a database that
has different extension version then the `postgres` or
`template1` database which  we actually use to perform
this validation.
So instead of using `pg_available_extensions` view we
use `pg_available_extension_versions` that should return
the same list of extension versions no matter which database
we connect to.

### DIFF
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -32,6 +32,7 @@
 #include <utils/inval.h>
 #include <utils/syscache.h>
 
+#include "config.h"
 #include "fdw/fdw.h"
 #include "remote/async.h"
 #include "remote/connection.h"
@@ -563,21 +564,26 @@ connect_for_bootstrapping(const char *node_name, const char *const host, int32 p
 }
 
 /**
- * Validate that extension is available and with the correct version.
+ * Validate that compatible extension is available on the data node.
  *
- * If the extension is not available on the data node, we will get strange
- * errors when we try to use functions, so we check that the extension is
- * available before attempting anything else.
+ * We check all available extension versions. Since we are connected to
+ * template DB when performing this check, it means we can't
+ * really tell if a compatible extension is installed in the database we
+ * are trying to add to the cluster. However we can make sure that a user
+ * will be able to manually upgrade the extension on the data node if needed.
  *
- * Will abort with error if there is an issue, otherwise do nothing.
+ * Will abort with error if there is no compatible version available, otherwise do nothing.
  */
 static void
 data_node_validate_extension_availability(TSConnection *conn)
 {
-	PGresult *res = remote_connection_execf(conn,
-											"SELECT default_version, installed_version FROM "
-											"pg_available_extensions WHERE name = %s",
-											quote_literal_cstr(EXTENSION_NAME));
+	bool compatible;
+
+	PGresult *res =
+		remote_connection_execf(conn,
+								"SELECT version FROM pg_available_extension_versions WHERE name = "
+								"%s AND version ~ '\\d+.\\d+.\\d+.*' ORDER BY version DESC",
+								quote_literal_cstr(EXTENSION_NAME));
 
 	if (PQresultStatus(res) != PGRES_TUPLES_OK)
 		ereport(ERROR,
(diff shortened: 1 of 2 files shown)


tsl/src/data_node.c
```diff
@@ -589,8 +589,18 @@ data_node_validate_extension_availability(TSConnection *conn)
...
+	/* If present we check the installed version first, otherwise check default version */
+	if (PQgetisnull(res, 0, 1) == 0)
+		remote_validate_extension_version(conn, PQgetvalue(res, 0, 1));
+	else if (PQgetisnull(res, 0, 0) == 0)
+		remote_validate_extension_version(conn, PQgetvalue(res, 0, 0));
+	else
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_DATA_NODE_INVALID_CONFIG),
+				 errmsg("No default version present for TimescaleDB extension on remote PostgreSQL "
+						"instance"),
```
erimatnor, line 602:
```suggestion
				 errmsg("no default version of the TimescaleDB extension present on remote PostgreSQL instance"),
```

erimatnor (approved):
Minor error formatting nit.

pmwkaa (approved):

niksajakovljevic:
@pmwkaa @erimatnor I've actually had to redo this PR. The reason is that while looking into Coverity test coverage I've found a strange thing: it marked a code line as not tested where I believed that our tests should cover it. Doing further investigation I've figured out that we use `postgres` or `template1` DB to perform extension availability checks. These DBs might have different versions then the database we are trying to add to the cluster ( especially here taking into account the ability to run ALTER EXTENSION UPDATE on data nodes), so might return ERROR even when the target database has the proper version present.

tsl/src/data_node.c
```diff
@@ -574,10 +575,13 @@ connect_for_bootstrapping(const char *node_name, const char *const host, int32 p
...
-	PGresult *res = remote_connection_execf(conn,
-											"SELECT default_version, installed_version FROM "
-											"pg_available_extensions WHERE name = %s",
-											quote_literal_cstr(EXTENSION_NAME));
+	bool compatible;
+
+	PGresult *res =
+		remote_connection_execf(conn,
+								"SELECT version FROM pg_available_extension_versions WHERE name = "
+								"%s AND version ~ '\\d+.\\d+.\\d+.*' ORDER BY version DESC",
```
erimatnor, line 583: Why is the version regexp needed here? Do we expect versions with a different format and why should those not be included in that case?

It is also unclear to me which use case this change addresses. I guess it is trying to check that there is at least one compatible version available for installation?

But there seems to be many different cases to handle here:

1. The database does not exist on the data node. Requirement: there must be a compatible extension version available
2. The database exists but does not have the extension installed. Same requirement as 1.
3. The database exists and extension is installed. Requirement: the installed extension must be a compatible version.


Which of these cases do we handle? It seems this change mostly targets case 1 and 2 above.

pmwkaa, line 583: Not sure how Postgres stores versions here, for example is `2.0` will be always `2.0.0`?

tsl/src/data_node.c
```diff
@@ -589,8 +593,29 @@ data_node_validate_extension_availability(TSConnection *conn)
...
+	int i;
+	StringInfo concat_versions = makeStringInfo();
+	bool old_version;
+	for (i = 0; i < PQntuples(res); i++)
+	{
+		appendStringInfo(concat_versions, "%s, ", PQgetvalue(res, i, 0));
+		compatible = dist_util_is_compatible_version(PQgetvalue(res, i, 0),
+													 TIMESCALEDB_VERSION,
+													 &old_version);
+		if (compatible)
```
pmwkaa, line 607: Isn't this version should be installed in such case during the `CREATE EXTENSION` operation on data node? I don't think we specify any version there

tsl/src/data_node.c
niksajakovljevic: Regexp is there to avoid tests breaking - since some of our so mocks don't follow our semantic versioning. If we would change the way we do versioning then that will require changes in few other places in our code (eg. logic that does version compatibility checking).

Right, your observations are pretty much correct. Let me explain a bit more. We have two types of checks: 1. check available versions 2. check installed version. Check 1 is what I was focused on here and that check if performed only
when trying to bootstrap DN. Check 2 is performed later on after we bootstrap DN and also each time we establish a connection.
This means that with this change in place the bootstrap process will not fail if it finds some available extension but can still fail later on if installed version is incompatible with version on AN. This would mean that user will need to do ALTER
EXTENSION UPDATE on problematic DN and then try adding the DN again. Hope this makes it more clear now.

niksajakovljevic: From what I can tell Postgres only stores string so we control the versioning. The order is not particularly important (code would work without ordering as well). I've only added it since most users would run the latest version so this would speed up the loop below.

tsl/src/data_node.c
niksajakovljevic: Right, we install the `default` version and that default might not always be compatible with the one on AN (eg. default is 1.7.4 and we have 2.0.0 as well). But in that case a user can ALTER EXTENSION UPDATE on DN and the problem will be solved.

tsl/src/data_node.c
erimatnor: Is this same function used for case 2? Do we support case 3 (and which function is used to validate the extension in that case)?

Would be good to make this clear in comments or the function description. For instance, it would be good to describe exactly how the function is used (i.e., which cases above).

niksajakovljevic: Well, this function is only used when bootstrapping parameter is true (that's by default). So it will run for all three case you've mentioned. No matter if DB/extension exists or not.

pmwkaa (approved):
I guess the main concern is that the validation function might not be in sync with `CREATE EXTENSION` executed after it.
For example, what if DN will have a version newer that AN in such case the latest version will be incompatible (likely), but `CREATE EXTENSION` will choose it anyway since it is the fresh one. We could pass `VERSION` parameter to the create statement in such case. I think we should be address this issues in separate PR with more testing, etc.

niksajakovljevic:
This function is responsible for checking if there are compatible versions available on the data node and does not provide any guarantees that compatible version will be installed (since that happens later on). So I guess we will deal with it separately. However it can only be addressed when installing new extension, the pre-installed extension still need to be manually updated.

PR #2619, opened 2020-11-04
Fix segfault in decompress_chunk for chunks with dropped columns

svenklemm:
This patch fixes a segfault in decompress_chunk for chunks with dropped
columns. Since dropped columns don't exists in the compressed chunk
the values for those columns were undefined in the decompressed tuple
leading to a segfault when trying to build the heap tuple.

### DIFF
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -1040,7 +1040,6 @@ decompress_chunk(Oid in_table, Oid out_table)
 
 	Oid compressed_data_type_oid = ts_custom_type_cache_get(CUSTOM_TYPE_COMPRESSED_DATA)->type_oid;
 
-	Assert(in_desc->natts >= out_desc->natts);
 	Assert(OidIsValid(compressed_data_type_oid));
 
 	{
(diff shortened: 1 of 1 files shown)


k-rus (commented):
Can a test be added for this case?

tsl/src/compression/compression.c
```diff
@@ -1061,6 +1060,14 @@ decompress_chunk(Oid in_table, Oid out_table)
...
+		/*
+		 * We need to make sure decompressed_is_nulls is in a defined state. While this
+		 * will get written for normal columns it will not get written for dropped columns
+		 * since dropped columns don't exist in the compressed chunk so we initiallize
+		 * with true here.
+		 */
+		memset(decompressor.decompressed_is_nulls, true, out_desc->natts);
+
 		Datum *compressed_datums = palloc(sizeof(*compressed_datums) * in_desc->natts);
 		bool *compressed_is_nulls = palloc(sizeof(*compressed_is_nulls) * in_desc->natts);
```
pmwkaa, line 1072: Do we need the same here?

tsl/src/compression/compression.c
```diff
@@ -1061,6 +1060,14 @@ decompress_chunk(Oid in_table, Oid out_table)
 			.decompressed_datums = palloc(sizeof(Datum) * out_desc->natts),
```
pmwkaa, line 1060: Don't we need to init this field as well in such case? Was thinking about something like `calloc` analogue for memory contexts, if such thing exists

pmwkaa (approved):
Is it possible to test it?

tsl/src/compression/compression.c
svenklemm: we could do palloc0 here but you are not supposed to look at datum when is_null is true

k-rus (approved):
It would be great to have a test.

tsl/src/compression/compression.c
svenklemm: no since those will be set by heap_deform_tuple

gayyappan (approved):

Issue #2608, opened 2020-10-30
Drop continuous aggregate with cascade fails.


gayyappan:
Applicable to 2.0 rc and master only
CREATE TABLE conditionsnm(time_int INT NOT NULL, device INT, value FLOAT);
SELECT create_hypertable('conditionsnm', 'time_int', chunk_time_interval => 10);

INSERT INTO conditionsnm
SELECT time_val, time_val % 4, 3.14 FROM generate_series(0,100,1) AS time_val;

SELECT set_integer_now_func('conditionsnm', 'test_int_now');
CREATE MATERIALIZED VIEW conditionsnm_4
    WITH (timescaledb.continuous, timescaledb.materialized_only = TRUE)
    AS
        SELECT time_bucket(7, time_int) as bucket,
            SUM(value), COUNT(value)
        FROM conditionsnm GROUP BY bucket WITH DATA;

Now try to drop the materialized view and view.

test=# drop materialized view conditionsnm_4 cascade;
ERROR:  "conditionsnm_4" is not a materialized view
HINT:  Use DROP VIEW to remove a view.

test=# drop materialized view conditionsnm_4 ;   <------ succeeds
NOTICE:  drop cascades to table _timescaledb_internal._hyper_12_137_chunk
DROP MATERIALIZED VIEW
er_11_135_chunk
drop cascades to table _timescaledb_internal._hyper_11_136_chunk
DROP MATERIALIZED VIEW

gayyappan:
Failure scenario 2:
I create view as
create view see_cagg as select * from conditions_nm4;
Now try to drop the materialized view with cascade. It fails. 

test=# DROP MATERIALIZED VIEW conditionsnm_4;
ERROR:  cannot drop view conditionsnm_4 because other objects depend on it
DETAIL:  view see_cagg depends on view conditionsnm_4

test=# DROP MATERIALIZED VIEW conditionsnm_4 cascade;
ERROR:  "conditionsnm_4" is not a materialized view
HINT:  Use DROP VIEW to remove a view.

PR #2620, opened 2020-11-04
Fix DROP CASCADE for continuous aggregate

pmwkaa:
Currently we treat cagg materialized view as a view internally, but that was missed in case of the drop cascade.

Issue: #2608

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -1278,9 +1278,6 @@ process_drop_continuous_aggregates(ProcessUtilityArgs *args, DropStmt *stmt)
 	ListCell *lc;
 	int caggs_count = 0;
 
-	if (stmt->behavior == DROP_CASCADE)
-		return;
-
 	foreach (lc, stmt->objects)
 	{
 		List *const object = lfirst(lc);
@@ -1289,16 +1286,6 @@ process_drop_continuous_aggregates(ProcessUtilityArgs *args, DropStmt *stmt)
 
 		if (cagg)
 		{
-			/* Add the materialization table to the arguments so that the
-			 * continuous aggregate and associated materialization table is
-			 * dropped together.
-			 *
-			 * If the table is missing, something is wrong, but we proceed
-			 * with dropping the view anyway since the user cannot get rid of
-			 * the broken view if we generate an error. */
-			Hypertable *ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
-			if (ht)
-				process_add_hypertable(args, ht);
 			/* If there is at least one cagg, the drop should be treated as a
 			 * DROP VIEW. */
 			stmt->removeType = OBJECT_VIEW;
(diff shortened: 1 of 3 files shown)


src/process_utility.c
```diff
@@ -1289,16 +1286,6 @@ process_drop_continuous_aggregates(ProcessUtilityArgs *args, DropStmt *stmt)
...
-			/* Add the materialization table to the arguments so that the
-			 * continuous aggregate and associated materialization table is
-			 * dropped together.
-			 *
-			 * If the table is missing, something is wrong, but we proceed
-			 * with dropping the view anyway since the user cannot get rid of
-			 * the broken view if we generate an error. */
-			Hypertable *ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
-			if (ht)
-				process_add_hypertable(args, ht);
```
pmwkaa, line 1301: I believe this part is not needed, because cagg is not supported by distributed hypertables

tsl/test/expected/continuous_aggs_ddl.out
```diff
@@ -1110,5 +1110,55 @@ partial_view_owner | test_role_1
...
+INSERT INTO conditionsnm
+SELECT time_val, time_val % 4, 3.14 FROM generate_series(0,100,1) AS time_val;
+CREATE MATERIALIZED VIEW conditionsnm_4
+WITH (timescaledb.continuous, timescaledb.materialized_only = TRUE)
+AS
+SELECT time_bucket(7, time_int) as bucket,
+SUM(value), COUNT(value)
+FROM conditionsnm GROUP BY bucket WITH DATA;
+NOTICE:  refreshing continuous aggregate "conditionsnm_4"
+DROP materialized view conditionsnm_4 CASCADE;
```
erimatnor, line 1145: When the dependent view does not exist, `CASCADE` should not be necessary. Therefore, I think this test line should not have `CASCADE` and we should still expect the drop to work.

pmwkaa: This is test specifically for the CASCADE, because it had the issue. It was taken from the reproduction case here: https://github.com/timescale/timescaledb/issues/2608. Usually you don't need to use cascade

erimatnor: Do we test DROP without CASCADE elsewhere?

I think it makes sense to have four test cases:

Dependent view does not exist:

- `DROP cagg;` -- success
- `DROP cagg CASCADE;`  --success

Dependent view exists:

- `DROP cagg;` -- failure
- `DROP cagg CASCADE;` -- success

pmwkaa: Yes, but it is not the matter of this test. I believe every other test which use caggs are doing just `DROP cagg`

erimatnor: I think it would be good to have these test cases in one place so that it is easy to see that all of them work as expected without having to sift through a lot of other tests.

pmwkaa: added more tests

erimatnor (approved):

k-rus (approved):

PR #2621, opened 2020-11-04
Rename process_add_hypertable()

pmwkaa:
This function is implemented in the process utility hook and used to
form a list of hypertables to pass to the tsl module for further
processing. Mostly used in the distributed DDL implementation.

This change renames the function to a more obvious name, regarding
what it does: `add_for_tsl_ddl_command_start()`.

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -267,7 +267,7 @@ relation_not_only(RangeVar *rv)
 }
 
 static void
-process_add_hypertable(ProcessUtilityArgs *args, Hypertable *ht)
+add_hypertable_to_process_args(ProcessUtilityArgs *args, Hypertable *ht)
 {
 	args->hypertable_list = lappend_oid(args->hypertable_list, ht->main_table_relid);
 }
@@ -407,7 +407,7 @@ process_altertableschema(ProcessUtilityArgs *args)
 	{
 		ts_hypertable_set_schema(ht, alterstmt->newschema);
 
-		process_add_hypertable(args, ht);
+		add_hypertable_to_process_args(args, ht);
 	}
 
 	ts_cache_release(hcache);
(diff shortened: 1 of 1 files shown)


erimatnor (approved):

k-rus (commented):
IMHO new name doesn't help much. It is complicated and can be violated by using it in non-tsl functionality. It also doesn't say what it does.
What's about:
```
add_hypertable_to_process_args
```

pmwkaa:
> IMHO new name doesn't help much. It is complicated and can be violated by using it in non-tsl functionality. It also doesn't say what it does.
> What's about:
> 
> ```
> add_hypertable_to_process_args
> ```

Process utility hook handling start from `timescaledb_ddl_command_start` function, we have a cross function named `ddl_command_start` as well. I believe this name should be represented here as well.

The purpose of this function is to create a list of involved hypertables in utility query and pass it further to the tsl module.

pmwkaa:
Any other thoughts?

k-rus:
IMHO the function name should talk about what it does. It doesn't need to mention where it can be used, especially if it is not required for the function to work correctly. 

>Process utility hook handling start from timescaledb_ddl_command_start function, we have a cross function named ddl_command_start as well. I believe this name should be represented here too.

I don't see anything in its implementation, which required to mention in the name that it is used in process start.

> The purpose of this function is to create a list of involved hypertables in the utility query and pass it further to the tsl module. 

The function only adds one hypertable to the arg list of hypertables.

svenklemm (approved):

pmwkaa:
@k-rus agree with you reasoning to use `add_hypertable_to_process_args` also make sense

k-rus (approved):

PR #2626, opened 2020-11-06
Fix incorrect total_table_pages setting for compressed scan

gayyappan:
The compressed scan path overwrites the total_table_pages
setting on PlannerInfo. This results in an assertion failure in
index_pages_fetched function at
Assert(T <= total_pages)
When a hypertable has a mix of compressed and uncompressed chunks,
the total_table_pages value is calculated in reenable_inheritance code
path and create_compressed_scan_paths. create_compressed_scan_path
should not overwrite the value previously computed by
reenable_inheritance.

### DIFF
--- a/tsl/src/nodes/decompress_chunk/decompress_chunk.c
+++ b/tsl/src/nodes/decompress_chunk/decompress_chunk.c
@@ -973,8 +973,10 @@ create_compressed_scan_paths(PlannerInfo *root, RelOptInfo *compressed_rel, int 
 	Path *compressed_path;
 
 	/* clamp total_table_pages to 10 pages since this is the
-	 * minimum estimate for number of pages */
-	root->total_table_pages = Max(compressed_rel->pages, 10);
+	 * minimum estimate for number of pages.
+	 * Add the value to any existing estimates
+	 */
+	root->total_table_pages += Max(compressed_rel->pages, 10);
 
 	/* create non parallel scan path */
 	compressed_path = create_seqscan_path(root, compressed_rel, NULL, 0);
--- a/tsl/test/expected/transparent_decompression-11.out
+++ b/tsl/test/expected/transparent_decompression-11.out
@@ -8695,3 +8695,67 @@ $sql$;
          ->  Seq Scan on compress_hyper_5_16_chunk (actual rows=5 loops=1)
 (7 rows)
 
+-- repro for core dump related to total_table_pages setting that get
+-- adjusted during decompress path.
+CREATE SEQUENCE vessel_id_seq 
+    INCREMENT 1
+    START 1    MINVALUE 1
+    MAXVALUE 9223372036854775807
+    CACHE 1;
+CREATE TABLE motion_table(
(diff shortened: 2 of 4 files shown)


tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -973,8 +973,10 @@ create_compressed_scan_paths(PlannerInfo *root, RelOptInfo *compressed_rel, int
 	Path *compressed_path;
 
 	/* clamp total_table_pages to 10 pages since this is the
-	 * minimum estimate for number of pages */
-	root->total_table_pages = Max(compressed_rel->pages, 10);
+	 * minimum estimate for number of pages.
+	 * Do not over-write any existing estimates
```
k-rus, line 977: This comment is misleading, since the actual code overwrites the existing value with added number of pages. `do not over-write` will do nothing if the value exists. Can you clarify the comment or fix the code?

k-rus (commented):
I have an issue to follow the commit message.

Can you explain what was incorrect and how it is fixed? Am I correct that the number of pages was set to the number of pages in a compressed chunk, which was last planned? Does the fix sums all total pages now?

You mention that the total number of pages is calculated by `reenable_inheritance`, however, `reenable_inheritance` doesn't exist in PG11 code. Does it mean that your fix might not estimate total number of pages in PG11 correctly.

erimatnor (approved):
LGTM, but suggest fixing the comment that @k-rus raised a concern about.

gayyappan:
> I have an issue to follow the commit message.
> 
> Can you explain what was incorrect and how it is fixed? Am I correct that the number of pages was set to the number of pages in a compressed chunk, which was last planned? Does the fix sums all total pages now?
> 
> You mention that the total number of pages is calculated by `reenable_inheritance`, however, `reenable_inheritance` doesn't exist in PG11 code. Does it mean that your fix might not estimate total number of pages in PG11 correctly.

This issue is specific to PG12. Modified the commit message to indicate that this occurs in the PG12 code path

tsl/src/nodes/decompress_chunk/decompress_chunk.c
gayyappan: modified comment.

k-rus:
Am I correct that now in both PG versions the total number of pages will add sum of total pages of each compressed chunk? I think it is worth to mention in the commit message, e.g., 
> This commit fixes that the total amount of pages is not overwritten, but instead total number of pages of compressed chunks are added.

k-rus (approved):
See suggestion in https://github.com/timescale/timescaledb/pull/2626#issuecomment-724282063

gayyappan:
> Am I correct that now in both PG versions the total number of pages will add sum of total pages of each compressed chunk? I think it is worth to mention in the commit message, e.g.,
> 
> > This commit fixes that the total amount of pages is not overwritten, but instead total number of pages of compressed chunks are added.

modified commit message.

PR #2639, opened 2020-11-11
Fix uninitialized variable in data node validation

erimatnor:
Initialize a boolean variable used to check for a compatible extension
on a data node. Leaving it uninitialized might lead to a potential
read of a garbage value and unpredictible behavior.


erimatnor:
This issue was found by a linter test.

gayyappan (approved):

WireBaron (approved):

PR #2643, opened 2020-11-13
Make compatible with old Windows compiler

k-rus:
Fixes usage of const to be identical in declarations and definitions,
since 15.9 VS compiler gets warnings otherwise.


pmwkaa (approved):

mkindahl (approved):
Looks good. Could you add the warnings you get as a comment to this PR?

`const` parameters does not make sense for function declarations (they are basically ignored), but they do make sense for function definitions so not sure if it is possible to keep them there. Seeing the warnings would make it clear if that is possible, and otherwise make it clear that it is not possible.

PR #2657, opened 2020-11-18
Add raw hypertable to continuous aggregate view

### LINKED ISSUE #2653: Include raw hypertable into continuous_aggregate information view
`timescaledb_information.continuous_aggregates` view doesn't include raw hypertable, while there might be cases when it is necessary to connect continuous aggregates with raw hypertables, e.g., to find all continuous aggregates belonging to a hypertable. An example of such need is refreshing a chunk before drop. The best is if the view includes `REGCLASS` of the raw hypertable, so a user policy to refresh and drop will look like:

```sql
CREATE OR REPLACE PROCEDURE refresh_and_drop_policy(job_id int, config jsonb) LANGUAGE PLPGSQL AS
$$
DECLARE
  drop_after INTERVAL;
  hypertable REGCLASS;
BEGIN
  SELECT jsonb_object_field_text(config, 'drop_after')::interval 
    INTO STRICT drop_after;
  SELECT jsonb_object_field_text(config, 'hypertable')::regclass
    INTO STRICT hypertable;
  BEGIN
    SELECT refresh_chunks(cagg, show_chunks(older_than => drop_after))
    FROM timescaledb_information.continuous_aggregates cagg
    WHERE cagg.raw_hypertable = hypertable;
    SELECT drop_chunks(hypertable, older_than => drop_after);
  END;
END
$$;
```

k-rus:
Including REGCLASS of raw hypertable in view
timescaledb_information.continuous_aggregates is important for some
use cases. This commit adds it to the view.

Fixes #2653

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -113,9 +113,11 @@ SELECT cagg.user_view_schema AS view_schema,
   cagg.materialized_only,
   ht.schema_name AS materialization_hypertable_schema,
   ht.table_name AS materialization_hypertable_name,
+  format('%I.%I', raw_ht.schema_name, raw_ht.table_name)::regclass AS raw_hypertable,
   directview.viewdefinition AS view_definition
 FROM _timescaledb_catalog.continuous_agg cagg,
   _timescaledb_catalog.hypertable ht,
+  _timescaledb_catalog.hypertable raw_ht,
   LATERAL (
     SELECT C.oid,
       pg_get_userbyid(C.relowner) AS viewowner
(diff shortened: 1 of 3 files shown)


sql/views.sql
```diff
@@ -113,9 +113,11 @@ SELECT cagg.user_view_schema AS view_schema,
   cagg.materialized_only,
   ht.schema_name AS materialization_hypertable_schema,
   ht.table_name AS materialization_hypertable_name,
+  format('%I.%I', raw_ht.schema_name, raw_ht.table_name)::regclass AS raw_hypertable,
```
erimatnor, line 116: We decided previously to not use regclass in views, as it had some undesirable properties when joining with other views and catalog info. This is also not consistent with other views that use `hypertable_schema` and `hypertable_name`.

IMO, we should have two field at the start of the view which is `hypertable_schema` and `hypertable_name` similar to what we have in, e.g., the `chunks`,`dimensions`, and `compression_settings` views.

k-rus: Thank you for reminding about this decision. I can clearly see that it is nightmare for usability to use the pair of strings.

k-rus:
I forgot about the decision to use separate names instead of regclass. Closing.

sql/views.sql
gayyappan: '<table_name>'::regclass returns the OID , if you need to pass the oid as arg to a function.

k-rus: @gayyappan It is necessary to concatinate the schema name to it to avoid unexpected ambiguity, right?

PR #2662, opened 2020-11-19
Save compression metadata settings on access node

### LINKED ISSUE #2660: compression_settings view does not report information for compressed hypertables
 - OS: macos docker `timescale/timescaledb:2.0.0-rc2-pg12`
 - PostgreSQL version: 12
 - TimescaleDB version: 2.0.0-rc2
 - Installation method: docker

**Describe the bug**
`timescaledb_information.compression_settings` does not display information for distributed hypertables.

**To Reproduce**
1. Create a distributed hypertable
2. Enable compression by running `ALTER TABLE ... SET (timescaledb.compress, ...)`
3. `SELECT * FROM timescaledb_information.compression_settings`

**Expected behavior**
The distributed hypertable should have an entry in the `compression_settings` view.

**Actual behavior**
The distributed hypertable does not have an entry in the `compression_settings` view.

**Additional context**
Example output
```SQL
> ALTER TABLE prom_data.test SET (
>    timescaledb.compress,
>    timescaledb.compress_segmentby = 'series_id',
>    timescaledb.compress_orderby = 'time, value'
> );
ALTER TABLE

> select * from timescaledb_information.compression_settings;
(0 rows)
```

gayyappan:
1. Add compression_state column for hypertable catalog
by renaming compressed column for the hypertable catalog
table. compression_state is a tri-state column.
This column indicates if the hypertable has
compression enabled (value = 2) or if it is an internal
compression table (value = 1).

2. Save compression settings on access node when compression
is turned on for a distributed hypertable
For a distributed hypertable, that has compresison enabled,
compression_state is set. We don't create any internal tables
on the access node.

Fixes #2660

### DIFF
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -37,8 +37,10 @@
 -- chunks.
 -- The unique constraint is table_name +schema_name. The ordering is
 -- important as we want index access when we filter by table_name
+CREATE SEQUENCE IF NOT EXISTS _timescaledb_catalog.hypertable_id_seq MINVALUE 1;
+
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.hypertable (
-  id serial PRIMARY KEY,
+  id INTEGER PRIMARY KEY DEFAULT nextval('_timescaledb_catalog.hypertable_id_seq'), 
   schema_name name NOT NULL CHECK (schema_name != '_timescaledb_catalog'),
   table_name name NOT NULL,
   associated_schema_name name NOT NULL,
@@ -47,18 +49,19 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.hypertable (
   chunk_sizing_func_schema name NOT NULL,
   chunk_sizing_func_name name NOT NULL,
   chunk_target_size bigint NOT NULL CHECK (chunk_target_size >= 0), -- size in bytes
-  compressed boolean NOT NULL DEFAULT FALSE,
+  compression_state smallint NOT NULL DEFAULT 0,
   compressed_hypertable_id integer REFERENCES _timescaledb_catalog.hypertable (id),
   replication_factor smallint NULL CHECK (replication_factor > 0),
   UNIQUE (associated_schema_name, associated_table_prefix),
   CONSTRAINT hypertable_table_name_schema_name_key UNIQUE (table_name, schema_name),
-  CONSTRAINT hypertable_dim_compress_check CHECK (num_dimensions > 0 OR compressed = TRUE),
-  CONSTRAINT hypertable_compress_check CHECK (compressed = FALSE OR (compressed = TRUE AND compressed_hypertable_id IS NULL))
+----internal compressed hypertables have compression state = 2
+  CONSTRAINT hypertable_dim_compress_check CHECK (num_dimensions > 0 OR compression_state = 2),
+  CONSTRAINT hypertable_compress_check CHECK ( (compression_state = 0 OR compression_state = 1 )  OR (compression_state = 2 AND compressed_hypertable_id IS NULL))
 );
+ALTER SEQUENCE _timescaledb_catalog.hypertable_id_seq OWNED BY _timescaledb_catalog.hypertable.id;
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.hypertable_id_seq', '');
 
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.hypertable', '');
-
-SELECT pg_catalog.pg_extension_config_dump(pg_get_serial_sequence('_timescaledb_catalog.hypertable', 'id'), '');
 
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.hypertable_data_node (
   hypertable_id integer NOT NULL REFERENCES _timescaledb_catalog.hypertable (id),
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -2,3 +2,125 @@ DROP VIEW IF EXISTS timescaledb_information.continuous_aggregates;
 DROP VIEW IF EXISTS timescaledb_information.job_stats;
 DROP FUNCTION IF EXISTS _timescaledb_internal.get_git_commit;
 
+-- Begin Modify hypertable table
+-- we make a copy of the data, remove dependencies, drop the table
+DROP VIEW IF EXISTS timescaledb_information.hypertables;
+DROP VIEW IF EXISTS timescaledb_information.chunks;
+DROP VIEW IF EXISTS timescaledb_information.dimensions;
+DROP VIEW IF EXISTS timescaledb_information.jobs;
+DROP VIEW IF EXISTS timescaledb_information.compression_settings;
+DROP VIEW IF EXISTS _timescaledb_internal.compressed_chunk_stats;
+DROP VIEW IF EXISTS _timescaledb_internal.hypertable_chunk_local_size;
+DROP FUNCTION IF EXISTS _timescaledb_internal.hypertable_from_main_table;
+
+CREATE TABLE _timescaledb_catalog.hypertable_tmp
+AS SELECT * from _timescaledb_catalog.hypertable;
+
+--drop foreign keys on hypertable 
+ ALTER TABLE _timescaledb_catalog.hypertable DROP CONSTRAINT hypertable_compressed_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.hypertable_data_node DROP CONSTRAINT hypertable_data_node_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.tablespace DROP CONSTRAINT tablespace_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.dimension DROP CONSTRAINT dimension_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.chunk DROP CONSTRAINT chunk_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.chunk_index DROP CONSTRAINT chunk_index_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_config.bgw_job DROP CONSTRAINT bgw_job_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.continuous_agg DROP CONSTRAINT continuous_agg_mat_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.continuous_agg DROP CONSTRAINT continuous_agg_raw_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.continuous_aggs_invalidation_threshold DROP CONSTRAINT continuous_aggs_invalidation_threshold_hypertable_id_fkey ; 
+ ALTER TABLE _timescaledb_catalog.hypertable_compression DROP CONSTRAINT hypertable_compression_hypertable_id_fkey ;
+
+ 
+CREATE TABLE tmp_hypertable_seq_value AS 
+SELECT last_value, is_called FROM _timescaledb_catalog.hypertable_id_seq;
+ALTER EXTENSION timescaledb DROP TABLE _timescaledb_catalog.hypertable;
+ALTER EXTENSION timescaledb DROP SEQUENCE _timescaledb_catalog.hypertable_id_seq;
+DROP TABLE _timescaledb_catalog.hypertable;
+
+CREATE SEQUENCE IF NOT EXISTS _timescaledb_catalog.hypertable_id_seq MINVALUE 1;
+
+-- now create table without self referential foreign key
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.hypertable(
+  id INTEGER PRIMARY KEY DEFAULT nextval('_timescaledb_catalog.hypertable_id_seq'),
+  schema_name name NOT NULL CHECK (schema_name != '_timescaledb_catalog'),
+  table_name name NOT NULL,
+  associated_schema_name name NOT NULL,
+  associated_table_prefix name NOT NULL,
+  num_dimensions smallint NOT NULL,
+  chunk_sizing_func_schema name NOT NULL,
+  chunk_sizing_func_name name NOT NULL,
+  chunk_target_size bigint NOT NULL CHECK (chunk_target_size >= 0), -- size in bytes
+  compression_state smallint NOT NULL DEFAULT 0,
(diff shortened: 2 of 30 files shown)


sql/updates/latest-dev.sql
```diff
@@ -1 +1,22 @@
 
+--add compression_state column and remove compressed column
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable ADD COLUMN compression_state integer NOT NULL DEFAULT 0;
```
erimatnor, line 3: To preserve correct column indexes, it is typically necessary to drop and recreate the whole table, including copying over data.

Otherwise there's a big risk the column indexes will be incorrect.

sql/pre_install/tables.sql
```diff
@@ -47,13 +47,13 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.hypertable (
   chunk_sizing_func_schema name NOT NULL,
   chunk_sizing_func_name name NOT NULL,
   chunk_target_size bigint NOT NULL CHECK (chunk_target_size >= 0), -- size in bytes
-  compressed boolean NOT NULL DEFAULT FALSE,
+  compression_state smallint NOT NULL DEFAULT 0,
```
erimatnor, line 50: Maybe add CHECK  (compression_state >= 0 and AND compresion_state < 3), or is this covered by CHECK constraint below? And, perhaps add explanation of states.

sql/updates/latest-dev.sql
gayyappan: @erimatnor This column (both new and old) don't have indexes. Do we still need to copy the data over?

sql/pre_install/tables.sql
gayyappan: Yes, it is covered by the check below.

sql/pre_install/tables.sql
```diff
@@ -47,13 +47,14 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.hypertable (
...
   compressed_hypertable_id integer REFERENCES _timescaledb_catalog.hypertable (id),
   replication_factor smallint NULL CHECK (replication_factor > 0),
   UNIQUE (associated_schema_name, associated_table_prefix),
   CONSTRAINT hypertable_table_name_schema_name_key UNIQUE (table_name, schema_name),
-  CONSTRAINT hypertable_dim_compress_check CHECK (num_dimensions > 0 OR compressed = TRUE),
-  CONSTRAINT hypertable_compress_check CHECK (compressed = FALSE OR (compressed = TRUE AND compressed_hypertable_id IS NULL))
+----internal compressed hypertables have compression state = 1
```
erimatnor, line 55: It would be more clear if this comment included all states:

```suggestion
-- Compression state is as follows:
-- compression_state = 0 --> hypertable or distributed hypertable with compression disabled, compressed_hypertable_id should be NULL.
-- compression_state = 1 --> internal compression hypertable that stores compressed data, compressed_hypertable_id should be NULL.
-- compression_state = 2 --> hypertable or distributed hypertable with compression enabled, compressed_hypertable_id is defined for local hypertables but not distributed hypertables
```

Ideally, I would change the order of state 1 and 2, although the exact ordinal number is not very important. I do think it is clearer that 0 and 1 is compression on/off on a non-internal hypertable. Otherwise, users will see this value go from 0 to 2 when enabling compression. Better to make the internal table the odd one.

sql/updates/latest-dev.sql
```diff
@@ -1,2 +1,162 @@
 DROP VIEW IF EXISTS timescaledb_information.continuous_aggregates;
 DROP VIEW IF EXISTS timescaledb_information.job_stats;
+
+--Modify hypertable table.
+DROP VIEW IF EXISTS timescaledb_information.hypertables;
+DROP VIEW IF EXISTS timescaledb_information.chunks;
+
+--alter compressed column type
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable
```
erimatnor, line 9: I would caution against modifying a table too much in place, as opposed to recreating it completely. We often see table indexes get messed up this way. Do we know for sure that, e.g., changing a columns type doesn't really do a drop/add column internally?

sql/updates/latest-dev.sql
```diff
@@ -1,2 +1,162 @@
...
+--add the constraints
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable
+ADD CONSTRAINT hypertable_dim_compress_check 
+CHECK (num_dimensions > 0 OR compression_state = 1);
+
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable
+ADD CONSTRAINT hypertable_compress_check 
+CHECK ( (compression_state = 0 OR compression_state =2 )  OR (compression_state = 1 AND compressed_hypertable_id IS NULL));
+--recreate timescaledb_information.hypertables and chunks view 
+CREATE OR REPLACE VIEW timescaledb_information.hypertables AS
```
erimatnor, line 40: It shouldn't be necessary to include the new views in the update script here. They will be recreated automatically.

src/hypertable.c
```diff
@@ -347,7 +349,7 @@ static bool
 hypertable_is_compressed_or_materialization(Hypertable *ht)
 {
 	ContinuousAggHypertableStatus status = ts_continuous_agg_hypertable_status(ht->fd.id);
-	return (ht->fd.compressed || status == HypertableIsMaterialization);
+	return (TS_HYPERTABLE_INTERNAL_COMPRESSION(ht) || status == HypertableIsMaterialization);
```
erimatnor, line 352: FWIW, this is also a readability improvement. Previously, it is too easy to believe that compressed=true means a regular hypertable with compression enabled.

src/hypertable.h
```diff
@@ -26,9 +26,20 @@ typedef struct SubspaceStore SubspaceStore;
...
+	HypertableInternalCompressionTable = 1,
+	HypertableCompressionEnabled = 2,
+};
+
+#define TS_HYPERTABLE_HAS_COMPRESSION(ht) ts_hypertable_has_compression(ht)
 
+#define TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht)                                                  \
+	((ht)->fd.compression_state == HypertableCompressionEnabled)
+
+#define TS_HYPERTABLE_INTERNAL_COMPRESSION(ht)                                                     \
```
erimatnor, line 41: Longer but less ambiguous:

```suggestion
#define TS_HYPERTABLE_IS_INTERNAL_COMPRESSION_TABLE(ht)                                                     \
```

src/hypertable.h
```diff
@@ -26,9 +26,20 @@ typedef struct SubspaceStore SubspaceStore;
 typedef struct Chunk Chunk;
 typedef struct Hypercube Hypercube;
 
-#define TS_HYPERTABLE_HAS_COMPRESSION(ht)                                                          \
-	((ht)->fd.compressed_hypertable_id != INVALID_HYPERTABLE_ID)
+enum
+{
+	HypertableCompressionOff = 0,
+	HypertableInternalCompressionTable = 1,
+	HypertableCompressionEnabled = 2,
```
erimatnor, line 33: Again, I think it makes more sense to make the enabled state `1` and the internal table `2`.

sql/updates/latest-dev.sql
```diff
@@ -1,2 +1,162 @@
...
+  (
+    SELECT count(1)
+    FROM _timescaledb_catalog.chunk ch
+    WHERE ch.hypertable_id = ht.id) AS num_chunks,
+  (
+    CASE WHEN ht.compressed_hypertable_id IS NULL THEN
+      FALSE
+    ELSE
+      TRUE
+    END) AS compression_enabled,
```
erimatnor, line 54: Is this correct now? Shouldn't  compression_enabled  check compression_state since a distributed hypertable can have compression enabled but compressed_hypertable_id is still NULL?

erimatnor (commented):
Some nits and suggestions. But also some open questions about update script and view definitions.

svenklemm (approved):

src/hypertable.h
svenklemm: Also couldnt hurt to have a text block describing what value this will be on different hypertables in singlenode/multinode scenario

svenklemm: Also agree with Erik on the ordering

sql/pre_install/tables.sql
gayyappan: I initially had the states listed as you describe. But I had a really hard time getting the update script working correctly. Keeping compression_state = 1, retains the original values for "compressed" column on update and only adds a new value for the new state.

sql/updates/latest-dev.sql
gayyappan: From PG docs : https://www.postgresql.org/docs/current/sql-altertable.html
``` Adding a column with a volatile DEFAULT or changing the type of an existing column will require the entire table and its indexes to be rewritten. As an exception, when changing the type of an existing column, if the USING clause does not change the column contents and the old type is either binary coercible to the new type or an unconstrained domain over the new type, a table rewrite is not needed; but any indexes on the affected columns must still be rebuilt. ```

File: sql/updates/latest-dev.sql (hunk 5 above)
Comment (gayyappan) [reply sql/updates/latest-dev.sql:40]: oh, ok. Will remove the view definitions.

File: src/hypertable.h (hunk 8 above)
Comment (gayyappan) [reply src/hypertable.h:33]: See my earlier comments on why I chose to retain the original mapped values for "compressed" in the "compression_state". Given that enum values should be used  by the C code, the internal values that it gets mapped to should not really matter.  
Will add comments.

File: sql/updates/latest-dev.sql (hunk 9 above)
Comment (gayyappan) [reply sql/updates/latest-dev.sql:54]: Good catch. Fixed.

File: src/hypertable.h (hunk 7 above)
Comment (gayyappan) [reply src/hypertable.h:41]: fixed.

File: sql/pre_install/tables.sql (hunk 3 above)
Comment (erimatnor) [reply sql/pre_install/tables.sql:55]: Probably another reason why in-table modifications in the update script isn't a good idea. It would be better, IMO, to recreate the table and insert data from the old table the way we want it in the new one.

File: sql/updates/latest-dev.sql (hunk 4 above)
Comment (erimatnor) [reply sql/updates/latest-dev.sql:9]: This sounds to me as if the modified column will actually not be strictly the same as a cleanly and newly created one. The column will stay with the old type but existing values converted to the new type. Is this really what we want? It seems unnecessarily risky to me.

How will this affect future update scripts, for instance, when the column definition can be different depending on whether the DB was previously upgraded or created fresh.

File: sql/updates/latest-dev.sql [hunk 10]
```diff
@@ -1,2 +1,39 @@
...
+--alter compressed column type
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable 
+ALTER COLUMN compressed DROP DEFAULT;
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable
+DROP CONSTRAINT hypertable_dim_compress_check;
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable
+DROP CONSTRAINT hypertable_compress_check;
+
+ALTER TABLE IF EXISTS _timescaledb_catalog.hypertable
+ALTER COLUMN compressed TYPE smallint USING compressed::int;
```
erimatnor, line 17: As stated above, this seems very risky to me.

If I understand this correctly, it means we keep the existing storage type of the column, which is `boolean`. Do we know that a `boolean` can hold a `smallint`?

Also, why do we cast the existing values to integer if the type is smallint?

This is also the reason why we cannot implement the more logical compression_state discussed above (0 meaning off, 1 on, and 2 internal table). So, it sounds like we are trading a potentially little bit simpler update script for worse semanctics going forward.

sql/pre_install/tables.sql
erimatnor: FWIW, we could get away with a char for compression_state. A char is often used by postgresql itself for similar things in its catalog tables.

src/hypertable.h
erimatnor: I agree that it doesn't technically matter for the C code. But the catalog will still show these values and even though the tables there are "internal" it is not uncommon to look into them or do views on them. If we have the opportunity to chose a more logical numbering here, I think it is worth it.

sql/updates/latest-dev.sql
svenklemm: might be a good idea to run cluster after the operation so it gets rewritten

src/hypertable.h
```diff
@@ -26,9 +26,24 @@ typedef struct SubspaceStore SubspaceStore;
...
+{
+	HypertableCompressionOff = 0,
+	HypertableCompressionEnabled = 1,
+	HypertableInternalCompressionTable = 2,
+};
 
+#define TS_HYPERTABLE_HAS_COMPRESSION(ht) ts_hypertable_has_compression(ht)
+
+#define TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht)                                                  \
+	((ht)->fd.compression_state == HypertableCompressionEnabled)
```
erimatnor, line 43: What is the difference between `TS_HYPERTABLE_HAS_COMPRESSION` and `TS_HYPERTABLE_HAS_COMPRESSION_ENABLED`? 

This is somewhat confusing and it is difficult to tell these two macros apart.

erimatnor (approved):
Approving, but there are some questions/nits I would like to see addressed before merging.

src/hypertable.h
gayyappan: TS_HYPERTABLE_HAS_COMPRESSION is used in places where we need to know that the internal table exists for the  user table. The action is done on the internal table.

erimatnor: Sorry, I should have been more clear/specific. The issue here is that the macros are hard to tell apart. I understand that they are used differently, although it was initially hard for me to realize the difference. This leads me to believe that this will be the case for others too, so just explaining in the PR is not enough.

If I look at the implementation of `ts_hypertable_has_compression` it is asserted that `ht->fd.compression_state == HypertableCompressionEnabled` is true when the function returns true. This is the same condition used in `TS_HYPERTABLE_HAS_COMPRESSION_ENABLED`. So it is really unclear what the difference is.

I think what we actually want to do here is distinguish between  distributed and non-distributed compressed hypertables:

```
#define TS_HYPERTABLE_IS_DISTRIBUTED_AND_COMPRESSED(ht) \
   (TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht) && \
     ht->fd.compressed_hypertable_id != INVALID_HYPERTABLE_ID)

```

This would make the naming more clear and also, looking at the definition, that this is a stricter condition than just having compression enabled.

gayyappan: The check is for the existence of a  "compressed" hypertable. I think a more appropriate rename is TS_HYPERTABLE_HAS_COMPRESSION_TABLE. If it is distributed, compressed_hypertable_id would be null.

PR #2674, opened 2020-11-23
Make job_stats view show jobs without hypertables

davidkohn88:
Using an inner join instead of a left join in the view definition
caused certain jobs to not be shown in the job_stats view because
they aren't associated with a hypertable (including any custom jobs).

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -78,7 +78,7 @@ SELECT ht.schema_name AS hypertable_schema,
   js.total_successes,
   js.total_failures
 FROM _timescaledb_config.bgw_job j
-  INNER JOIN _timescaledb_catalog.hypertable ht ON j.hypertable_id = ht.id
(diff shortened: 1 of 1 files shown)


davidkohn88:
I'm not sure if this is going to work, because I can't get tests to run because I can't get ssl compilation to work...so someone else may either have to take this or help me with my setup...

k-rus (commented):
Can you add a test, which shows that it works in the missed case and prevents future regression?

Now I noticed the comment:


> I'm not sure if this is going to work, because I can't get tests to run because I can't get ssl compilation to work...so someone else may either have to take this or help me with my setup...

Then may be create an issue, so @timescale/database-eng can fix the issue and implement the test?

k-rus:
Thank you for finding the bug and  the proposed fixed! It is merged with a test through https://github.com/timescale/timescaledb/pull/2678

PR #2682, opened 2020-11-25
Fix appveyor build

svenklemm:


erimatnor (approved):

mkindahl (approved):

k-rus (commented):
Will it work correctly when package for 13 will be pushed to apk repo?

svenklemm:
> Will it work correctly when package for 13 will be pushed to apk repo?

I changed it to bump the version it would probably fail when newer versions get added

PR #2683, opened 2020-11-25
Fix crash in add_job when given NULL interval

erimatnor:
This change fixes a crash in `add_job` when given an NULL
schedule interval as input. In addition, the following improvements
have been implemented for all job API functions:

* Validation of input parameters, specifically NULL checks
* Improved error messages for clarity and consistency
* Tests for bad input arguments

### DIFF
--- a/tsl/src/bgw_policy/job_api.c
+++ b/tsl/src/bgw_policy/job_api.c
@@ -55,11 +55,21 @@ job_add(PG_FUNCTION_ARGS)
 
 	TS_PREVENT_FUNC_IF_READ_ONLY();
 
+	if (PG_ARGISNULL(0))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("function or procedure cannot be NULL")));
+
+	if (NULL == schedule_interval)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("schedule interval cannot be NULL")));
+
 	func_name = get_func_name(proc);
 	if (func_name == NULL)
 		ereport(ERROR,
 				(errcode(ERRCODE_UNDEFINED_OBJECT),
-				 errmsg("function with OID %d does not exist", proc)));
+				 errmsg("function or procedure with OID %d does not exist", proc)));
 
 	if (pg_proc_aclcheck(proc, owner, ACL_EXECUTE) != ACLCHECK_OK)
 		ereport(ERROR,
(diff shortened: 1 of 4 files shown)


tsl/src/bgw_policy/job_api.c
```diff
@@ -55,6 +55,15 @@ job_add(PG_FUNCTION_ARGS)
 
 	TS_PREVENT_FUNC_IF_READ_ONLY();
 
+	if (!OidIsValid(proc))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("function cannot be NULL")));
```
k-rus, line 60: It is allowed to provide a function or a procedure, thus it will be good to be clear in the error message. Saying `function` can confuse a user that a procedure is not expected.

tsl/src/bgw_policy/job_api.c
```diff
@@ -106,11 +115,16 @@ Datum
...
+	Oid owner;
 
 	TS_PREVENT_FUNC_IF_READ_ONLY();
 
-	BgwJob *job = ts_bgw_job_find(job_id, CurrentMemoryContext, true);
-	Oid owner = get_role_oid(NameStr(job->fd.owner), false);
+	if (PG_ARGISNULL(0))
+		ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("job ID cannot be NULL")));
+
+	job = ts_bgw_job_find(job_id, CurrentMemoryContext, true);
```
k-rus, line 126: Checking job ID argument and retrieving the job is the same code in all job functions, except `add_job`. It might be beneficial to replace code copies with a call to a function or macro.

tsl/test/expected/bgw_custom.out
```diff
@@ -26,6 +26,19 @@ BEGIN
   COMMIT;
 END
 $$;
+\set ON_ERROR_STOP 0
+-- test bad input
+SELECT add_job(NULL, '1h');
+ERROR:  function cannot be NULL
+SELECT add_job('invalid_func', '1h');
+ERROR:  function "invalid_func" does not exist at character 16
```
k-rus, line 34: It can be a procedure too, so it is better to not focus just on function. For example, user call `add_job` for a procedure, but misspelled the name, then the error message can be interpreted as it is required to have a function, not a procedure.

k-rus (approved):
LGTM, but I highly recommend to fix error messages that they are not focused just on functions, since procedures are also allowed.
See also a comment about code copies.

mkindahl (approved):

tsl/src/bgw_policy/job_api.c
svenklemm: this code-path can be reached with non-NULL aswell so the error message is incorrect, probably should add a test for ```select add_job(0,'1day');```

erimatnor: @k-rus I was following the format of other errors in this function that refers to a "function". I would have to update also those messages for consistency. That's OK with me, but I wanted to minimize changes.

tsl/test/expected/bgw_custom.out
erimatnor: This is a PostgreSQL error so there is nothing we can do about it. I guess all errors are legacy (focusing on function) since procedures was a recent addition to PostgreSQL.

tsl/src/bgw_policy/job_api.c
k-rus: @erimatnor this PR contains changes to error message unrelated to this PR (removing hash in front of job number), so it is inline to improve other error messages as part of this PR.

PR #2688, opened 2020-11-26
Fix crash for concurrent drop and compress chunk

erimatnor:
This change fixes a segfault that occured when `drop_chunks` is
concurrently executed with `compress_chunk` and the same chunk that
gets dropped is also being compressed.

The crash happened because the tuple lock status function for a
dimension slice passed in a pointer to a dimension slice that was
always NULL.

An isolation test is also added to cover concurrent compression and
drop of the same chunk. To make the test pass with identical errors
for PG11 and PG12, additional changes are made to the scanner API to
pass on the lock failure data so that it is possible to distinguish
between an update and delete on PG11.

### DIFF
--- a/src/dimension_slice.c
+++ b/src/dimension_slice.c
@@ -99,8 +99,27 @@ ts_dimension_slice_cmp_coordinate(const DimensionSlice *slice, int64 coord)
 	return 0;
 }
 
+static bool
+tuple_is_deleted(TupleInfo *ti)
+{
+#if PG12_GE
+#ifdef USE_ASSERT_CHECKING
+	if (ti->lockresult == TM_Deleted)
+		Assert(ItemPointerEquals(ts_scanner_get_tuple_tid(ti), &ti->lockfd.ctid));
+#endif
+	return ti->lockresult == TM_Deleted;
+#else
+	/* If the tid and ctid in the lock failure data is the same, then this is
+	 * a delete. Otherwise it is an update and ctid is the new tuple ID. This
+	 * applies mostly to PG11, since PG12 has an explicit lockresult for
+	 * deleted tuples. */
+	return ti->lockresult == TM_Updated &&
+		   ItemPointerEquals(ts_scanner_get_tuple_tid(ti), &ti->lockfd.ctid);
+#endif
+}
+
 static void
-lock_result_ok_or_abort(TupleInfo *ti, DimensionSlice *slice)
+lock_result_ok_or_abort(TupleInfo *ti)
 {
 	switch (ti->lockresult)
 	{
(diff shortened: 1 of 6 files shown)


src/dimension_slice.c
```diff
@@ -114,23 +114,34 @@ lock_result_ok_or_abort(TupleInfo *ti, DimensionSlice *slice)
...
+			 * tuple ID. This applies mostly to PG11, since PG12 has an
+			 * explicit lockresult for delete tuples. */
+			if (memcmp(ts_scanner_get_tuple_tid(ti), &ti->lockfd.ctid, sizeof(ti->lockfd.ctid)) ==
+				0)
+				ereport(ERROR,
+						(errcode(ERRCODE_LOCK_NOT_AVAILABLE),
+						 errmsg("chunk deleted by other transaction"),
+						 errhint("Retry the operation again.")));
+			else
+				ereport(ERROR,
```
erimatnor, line 134: In practice, this case should never occur. Dimension slices are immutable so they should never be updated. But I am not sure we are confident enough to assert it here.

tsl/test/isolation/expected/deadlock_drop_chunks_compress.out
```diff
@@ -0,0 +1,237 @@
+Parsed test spec with 2 sessions
+
+starting permutation: s1_drop s1_commit s2_compress_chunk_1 s2_compress_chunk_2 s2_commit
```
pmwkaa, line 3: This iterates over all possible step order combinations, right?

pmwkaa (approved):
Looks good. I guess this could also be tested using wait points we used before to test concurrent behavior in a more deterministic way.

Maybe it makes sense to reference an issue related to this fix, if any?

tsl/test/isolation/expected/deadlock_drop_chunks_compress.out
erimatnor: Unless a test specifies explicit permutations, it will run all permutations.

src/scanner.h
```diff
@@ -42,6 +42,8 @@ typedef struct TupleInfo
 	 * in lockresult.
 	 */
 	TM_Result lockresult;
+	/* Failure data in case of failed tuple lock */
+	TM_FailureData lockfd;
```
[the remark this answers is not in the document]
erimatnor: It is usually embedded in a larger struct, which, AFAIK, are initialized properly in the places where used. Still, even if this was garbage, it is not something that one should look at unless lockresult tells you to, in which case lockfd should also be set accordingly.

erimatnor:
@pmwkaa I thought about using wait points, but couldn't see that it would bring additional benefits over the current test at this point. If you know how the test can be improved, I am happy to learn about that.

Regarding the deadlock issue, I looked at those, but I am not sure this is related. This fixes a crash, not a deadlock. Also, I am not sure what that deadlock issue is. I didn't encounter any deadlocks in the tests I created. But I guess in some cases deadlocks are expected if chunks are locked in different and incompatible orders. But that is to be expected in that case.

src/scanner.h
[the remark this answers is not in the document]
pmwkaa: Sure, I was mostly thinking about tools like valgrind which might bring additional warnings, etc. Anyway, I don't think that it is necessary to change anything

svenklemm (approved):

mkindahl (approved):
LGTM.

PR #2696, opened 2020-12-01
Support gapfill on distributed hypertable

k-rus:
This commits enables support for executing queries with
time_bucket_gapfill on distributed hypertables. Current implementation
always plans to execute time_bucket_gapfill on the access node. In
future, it can be optimized to be executed on data nodes in suitable
cases.

### DIFF
--- a/tsl/test/expected/dist_api_calls.out
+++ b/tsl/test/expected/dist_api_calls.out
@@ -216,27 +216,150 @@ _timescaledb_internal._dist_hyper_1_8_chunk
  
 (1 row)
 
--- Simple test of time_bucket_gapfill, which is disabled for now.
-\set ON_ERROR_STOP 0
+-- Simple tests of time_bucket_gapfill
 SELECT time_bucket_gapfill('3 hours', time, '2017-01-01 06:00', '2017-01-02 18:00'),
        first(value, time),
        avg(value)
 FROM disttable
 GROUP BY 1;
-ERROR:  time_bucket_gapfill not implemented for distributed hypertable
+     time_bucket_gapfill      | first |   avg   
+------------------------------+-------+---------
+ Sun Jan 01 04:00:00 2017 PST |   1.2 |     1.2
+ Sun Jan 01 07:00:00 2017 PST |   7.3 |     5.8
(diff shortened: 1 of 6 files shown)


tsl/test/expected/dist_api_calls.out
```diff
@@ -216,44 +216,278 @@ _timescaledb_internal._dist_hyper_1_8_chunk
...
--- Simple test of time_bucket_gapfill, which is disabled for now.
-\set ON_ERROR_STOP 0
+-- Simple tests of time_bucket_gapfill
 SELECT time_bucket_gapfill('3 hours', time, '2017-01-01 06:00', '2017-01-02 18:00'),
        first(value, time),
        avg(value)
 FROM disttable
 GROUP BY 1;
-ERROR:  time_bucket_gapfill not implemented for distributed hypertable
+     time_bucket_gapfill      | first |   avg
```
erimatnor, line 225: This test file (`dist_api_calls.c`) seems like the wrong place to test gapfill queries on distributed hypertables. I could maybe understand that the previous block functionality was here, but actual query output should probably be in `dist_query` or similar query-related test.

erimatnor (commented):
One question about where to put the tests. Otherwise, LGTM.

tsl/test/expected/dist_api_calls.out
k-rus: @erimatnor I agree I want to create separate test, which will also compare query results between single node, non-partitionwise distributed and partitionwise distributed. I feel it is good to make as a separate commit and may be a separate PR.

erimatnor: I guess that's fine, but is there a reason why it isn't possible to just move this to, e.g., `dist_query` right now and at least get some of those benefits?

erimatnor (approved):

tsl/test/expected/dist_api_calls.out
k-rus: Imho it is double work. I investigated previously and haven't found a natural place to put this test. So it will require some work.

mkindahl (approved):
LGTM

PR #2697, opened 2020-12-01
Fix postgres versions in build matrix

svenklemm:
Versions in the build matrix with significant leading/trailing zeroes
need the version string quoted otherwise it will get treated as number
and those zeroes will be removed. As currently those tests fixed by
this patch run PG 11.1 instead of 11.10.


k-rus (approved):
Thank you for fixing it!

gayyappan (approved):

PR #2703, opened 2020-12-02
Compare distributed gapfill test to single-node

k-rus:
The regression tests of gapfill on distributed hypertable compares
their results to executing the same gapfill queries against
non-distributed hypertable. The distributed gapfill test executes
queries with partitionwise_aggregate disabled and enabled and compares
both results.

The tests were moved from the previous location to new, more
appropriate files. The output was also moved with fixing the table
name.

This commit doesn't add new tests and doesn't include explain.


pmwkaa (approved):
Looks good overall

erimatnor (approved):
Approving, but I am wondering why this can't be added to an existing test (e.g., dist_query.sql)? Adding extra test files is always a trade-off since it also leads to longer test execution.

Issue #2528, opened 2020-10-12
tableoid not set correctly for queries on distributed hypertables


svenklemm:
Tableoid for queries on distributed hypertables is set to hypertable instead of chunk when timescaledb.enable_per_data_node_queries is true
```
dev=# select tableoid::regclass,* from metrics_dist where device_id = 1;
   tableoid   |          time          | device_id | value
--------------+------------------------+-----------+-------
 metrics_dist | 2000-01-01 00:00:00+01 |         1 |   0.5
(1 row)
dev=# set timescaledb.enable_per_data_node_queries to false;
SET
dev=# select tableoid::regclass,* from metrics_dist where device_id = 1;
                   tableoid                   |          time          | device_id | value
----------------------------------------------+------------------------+-----------+-------
 _timescaledb_internal._dist_hyper_9_14_chunk | 2000-01-01 00:00:00+01 |         1 |   0.5
(1 row)
```

erimatnor:
Could you provide some more information and a way to reproduce this issue?

svenklemm:
tableoid is a system column that identifies the relation a tuple is originating from. For a distributed hypertable this is set to the hypertable which is wrong since hypertable has no tuples.
To reproduce:
```
SELECT tableoid::regclass, * FROM dist_insert LIMIT 10;
  tableoid   |              inserted               | partkey |       value
-------------+-------------------------------------+---------+--------------------
 dist_insert | Wed Apr 17 16:03:48.482009 2019 PDT | 0       |  0.143692847820045
 dist_insert | Wed Apr 17 05:25:54.923609 2019 PDT | 0       |  0.919203089383576
 dist_insert | Tue Apr 16 16:51:13.461209 2019 PDT | 3       |  0.501142786979077
 dist_insert | Tue Apr 16 19:10:08.123609 2019 PDT | 3       |  0.977845790681446
 dist_insert | Sun Apr 14 14:47:26.776409 2019 PDT | 7       |  0.361208168566694
 dist_insert | Thu Feb 06 12:56:44.776409 2020 PST | 0       |  0.482509741016958
 dist_insert | Mon Feb 10 13:44:08.891609 2020 PST | 0       |  0.617650506760175
 dist_insert | Wed Feb 12 01:13:09.304409 2020 PST | 3       |  0.780974726880007
 dist_insert | Mon Feb 10 09:18:49.518809 2020 PST | 3       | 0.0343540287821931
[...]

PR #2705, opened 2020-12-02
Block tableoid access on distributed hypertable

pmwkaa:
This change adds a check which raises an error when system column (like
tableoid) being requested in case if data node scan is enabled.

Fixes: #2528

### DIFF
--- a/tsl/src/fdw/data_node_scan_exec.c
+++ b/tsl/src/fdw/data_node_scan_exec.c
@@ -71,22 +71,15 @@ data_node_scan_next(CustomScanState *node)
 	oldcontext = MemoryContextSwitchTo(econtext->ecxt_per_tuple_memory);
 	slot = fdw_scan_iterate(&node->ss, &sss->fsstate);
 	MemoryContextSwitchTo(oldcontext);
-	/*
-	 * If any system columns are requested, we have to force the tuple into
-	 * physical-tuple form to avoid "cannot extract system attribute from
-	 * virtual tuple" errors later.  We also insert a valid value for
-	 * tableoid, which is the only actually-useful system column.
-	 */
-	if (sss->systemcol && !TupIsNull(slot))
-	{
-#if PG12_LT
-		HeapTuple tup = ExecMaterializeSlot(slot);
 
-		tup->t_tableOid = RelationGetRelid(node->ss.ss_currentRelation);
-#else
-		slot->tts_tableOid = RelationGetRelid(node->ss.ss_currentRelation);
-#endif
-	}
+	/* Raise an error when system column is requsted, eg. tableoid */
+	if (sss->systemcol && !TupIsNull(slot))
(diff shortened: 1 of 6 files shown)


k-rus (dismissed):
LGTM
I don't understand why it is necessary to create new distributed hypertable instead of utilising the one existing in the test.

k-rus (commented):
I dismissed my review, since the PR breaks tests on PG11, where a system column OID is returned.

tsl/src/fdw/data_node_scan_exec.c
```diff
@@ -71,22 +71,13 @@ data_node_scan_next(CustomScanState *node)
...
-#else
-		slot->tts_tableOid = RelationGetRelid(node->ss.ss_currentRelation);
-#endif
-	}
+	/* Raise an error when system column is requsted, eg. tableoid */
+	if (sss->systemcol && !TupIsNull(slot))
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("system columns are not accessible on distributed hypertable "
+						"with enabled timescaledb.enable_per_data_node_queries")));
```
erimatnor, line 80: Main error should be short according to error style guide. I suggest splitting this up:

```suggestion
		ereport(ERROR,
				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
				 errmsg("system columns are not accessible on distributed hypertables"),
				 errhint("Set timescaledb.enable_per_data_node_queries=false to query system columns.")));
```

erimatnor (approved):
Approving, but I suggest fixing two things before merging:

* Split up error message so that main error is shorter
* Commit message should use "Fixes" or "Closes" to reference the issue as "Issue" is not a recognizable key word. https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

pmwkaa:
>     * Commit message should use "Fixes" or "Closes" to reference the issue as "Issue" is not a recognizable key word. https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

That's not entirely true, github recognizes issue references in format `#number`. I was aware about the tag, I just wanted to reference the issue without doing any action to it. Anyway, updated to the 'fixes' as suggested.

tsl/src/fdw/data_node_scan_exec.c
k-rus: > "system columns are not accessible on distributed hypertables"

This suggested text is not fully correct, since system columns are accessible on distributed hypetable if the optimization is disabled. So I don't think it should be stated in the suggested way.

pmwkaa: Updated the message to be more specific

k-rus (approved):

PR #2707, opened 2020-12-02
Introduce additional db for data node bootstrapping

niksajakovljevic:
We now try connecting to three databases before giving up:
postgres, template1 and defaultdb.

### DIFF
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -61,6 +61,9 @@ typedef struct DbInfo
 	NameData collation;
 } DbInfo;
 
+/* A list of databases we try to connect to when bootstrapping a data node */
+static const char *bootstrap_databases[] = { "postgres", "template1", "defaultdb" };
+
 static bool data_node_validate_database(TSConnection *conn, const DbInfo *database);
 
 /*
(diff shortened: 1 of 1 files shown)


tsl/src/data_node.c
```diff
@@ -541,22 +545,24 @@ add_distributed_id_to_data_node(TSConnection *conn)
...
-	TSConnection *conn = remote_connection_open_with_options_nothrow(node_name, node_options);
-
-	if (conn)
-		return conn;
-
-	node_options = create_data_node_options(host, port, "template1", username, password);
-	return remote_connection_open_with_options_nothrow(node_name, node_options);
+	TSConnection *conn = NULL;
+	int i;
+	for (i = 0; i < MAX_BOOTSTRAP_DATABASES; i++)
```
pmwkaa, line 557: Instead of introducing `MAX_BOOTSTRAP_DATABASES` macro, also could do something like this:

``` C
static const char *bootstrap_databases[] = { "postgres", "template1", "defaultdb", NULL };

for (i = 0; i < bootstrap_databases[i] != NULL; i++)
{
...
}
```

tsl/src/data_node.c
```diff
@@ -541,22 +545,24 @@ add_distributed_id_to_data_node(TSConnection *conn)
...
+	{
+		List *node_options =
+			create_data_node_options(host, port, bootstrap_databases[i], username, password);
+		conn = remote_connection_open_with_options_nothrow(node_name, node_options);
+		if (conn)
+			return conn;
+	}
+	return conn;
```
pmwkaa, line 565: nit: maybe just return NULL here and possible to move `TSConnection *conn` to the loop block

pmwkaa (approved):

tsl/src/data_node.c
erimatnor: You can also use `lengthof(bootstrap_databases)`

erimatnor (approved):
Fix commit title and use lengthof for array.

tsl/src/data_node.c
niksajakovljevic: Updated

erimatnor (approved):

pmwkaa (approved):

PR #2708, opened 2020-12-02
Rename macro TS_HYPERTABLE_HAS_COMPRESSION

gayyappan:
This PR cleans up some macro names.
Rename macro TS_HYPERTABLE_HAS_COMPRESSION
to TS_HYPERTABLE_HAS_COMPRESSION_TABLE.

### DIFF
--- a/src/hypertable.h
+++ b/src/hypertable.h
@@ -37,7 +37,7 @@ enum
 	HypertableInternalCompressionTable = 2,
 };
 
-#define TS_HYPERTABLE_HAS_COMPRESSION(ht) ts_hypertable_has_compression(ht)
+#define TS_HYPERTABLE_HAS_COMPRESSION_TABLE(ht) ts_hypertable_has_compression_table(ht)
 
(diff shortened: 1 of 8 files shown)


src/hypertable.h
```diff
@@ -37,7 +37,7 @@ enum
 	HypertableInternalCompressionTable = 2,
 };
 
-#define TS_HYPERTABLE_HAS_COMPRESSION(ht) ts_hypertable_has_compression(ht)
+#define TS_HYPERTABLE_HAS_COMPRESSION_TABLE(ht) ts_hypertable_has_compression_table(ht)
```
pmwkaa, line 40: why not to use function in such case without macro?

erimatnor (approved):
Wondering if we shouldn't just make all of these into functions, since at least the macro in this PR is just aliasing an C function.

src/hypertable.h
gayyappan: Using a macro keeps it consistent with the other compression related macros like TS_HYPERTABLE_HAS_COMPRESSION_ENABLED

pmwkaa (approved):

PR #2719, opened 2020-12-07
Support disabling compression on distributed hypertables

### LINKED ISSUE #2716: Turning off compression on distributed hypertable doesn't work.
 - OS: [e.g. Ubuntu 16.04, Windows 10 x64, etc] docker
 - PostgreSQL version (output of `postgres --version`): 12.x
 - TimescaleDB version (output of `\dx` in `psql`): 2.0.0-rc4
 - Installation method: docker

**Describe the bug**
Turning off compression on a distributed hypertable fails.

```
STDERR AN 2020-12-07 00:55:31.369 UTC [tmp_db_timescale_migrate_test]ERROR:  compressed hypertable not found
STDERR AN 2020-12-07 00:55:31.369 UTC [tmp_db_timescale_migrate_test]DETAIL:  compression was enabled on "test", but its internal compressed hypertable could not be found.
STDERR AN 2020-12-07 00:55:31.369 UTC [tmp_db_timescale_migrate_test]CONTEXT:  SQL statement "
STDERR              ALTER TABLE prom_data.test SET (
STDERR                  timescaledb.compress = false
STDERR              ); "
```

**Expected behavior**
No error

gayyappan:
Fixes #2716

### DIFF
--- a/tsl/src/compression/create.c
+++ b/tsl/src/compression/create.c
@@ -868,7 +868,7 @@ drop_existing_compression_table(Hypertable *ht)
 	 * thus the column types of compressed hypertable need to change) */
 	ts_hypertable_drop(compressed, DROP_RESTRICT);
 	ts_hypertable_compression_delete_by_hypertable_id(ht->fd.id);
-	ts_hypertable_unset_compressed_id(ht);
+	ts_hypertable_unset_compressed(ht);
 }
 
 static bool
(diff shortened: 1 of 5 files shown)


svenklemm (approved):

tsl/src/compression/create.c
```diff
@@ -888,8 +888,15 @@ disable_compression(Hypertable *ht, WithClauseResult *with_clause_options)
 		return false;
 	/* compression is enabled. can we turn it off? */
 	check_modify_compression_options(ht, with_clause_options);
-	drop_existing_compression_table(ht);
-	ts_hypertable_unset_compressed_id(ht);
+
+	/* distributed hypertables do not have compression table on the access node */
+	if (TS_HYPERTABLE_HAS_COMPRESSION_TABLE(ht))
+		drop_existing_compression_table(ht);
```
erimatnor, line 894: Given that `ts_hypertable_unset_compressed()` is called in the `else` case, I would have expected it is called also when you have a compression table. In fact, to unset the compression state seems to be independent of this if-clause. 

For clarity, I would move `ts_hypertable_unset_compressed()` out of this if clause and out of `drop_existing_compression_table` to the end of this function. Then it is set at the same level for both cases and it is clear it is always unset.

erimatnor (approved):
Approving with one suggestion for making code more clear.

tsl/src/compression/create.c
gayyappan: I did consider the option of making unset_compressed common to both paths. drop_existing_compression_table is also called from other places in the code.  So it makes sense to encapsulate all the things that need to happen during a drop in a single function.

PR #2723, opened 2020-12-08
Add support for PG13 List implementation

svenklemm:
PG13 changes the List implementation from a linked list to an array
while most of the API functions did not change a few them have slightly
different signature in PG13, additionally the list_make5 functions
got removed.

https://github.com/postgres/postgres/commit/1cff1b95ab

### DIFF
--- a/src/bgw/scheduler.c
+++ b/src/bgw/scheduler.c
@@ -456,7 +456,7 @@ ts_update_scheduled_jobs_list(List *cur_jobs_list, MemoryContext mctx)
 			 */
 			terminate_and_cleanup_job(cur_sjob);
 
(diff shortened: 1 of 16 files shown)


mkindahl (approved):

src/bgw/scheduler.c
```diff
@@ -488,7 +488,11 @@ ts_update_scheduled_jobs_list(List *cur_jobs_list, MemoryContext mctx)
 	{
 		ListCell *ptr;
 
+#if PG13_GE
```
erimatnor, line 491: I suggest we create a compat version of `for_each_cell` too in order to make this code look nicer.

erimatnor (approved):
One small nit.

PR #2729, opened 2020-12-09
Fix compressed status for distributed chunks

### LINKED ISSUE #2724: timescaledb_information.chunks is_compressed not reporting correctly on access node
 - OS: Debian 9
 - PostgreSQL version (output of `postgres --version`): postgres (PostgreSQL) 12.4 (Debian 12.4-1.pgdg90+1)
 - TimescaleDB version (output of `\dx` in `psql`): timescaledb | 2.0.0-rc4
 - Installation method: apt from packagecloud.io

**Describe the bug**
Compressed tables report as `is_compressed=f` on the access node.

**To Reproduce**
```
create table test (time timestamptz, i int);

select create_distributed_hypertable('test','time');

alter table test set (timescaledb.compress);

insert into test values (now(), 1);

select compress_chunk(chunk_schema || '.' || chunk_name) from timescaledb_information.chunks where hypertable_name='test';

select is_compressed from timescaledb_information.chunks where hypertable_name='test';

```

**Expected behavior**
```
 is_compressed 
---------------
 t
(1 row)
```

**Actual behavior**
```
 is_compressed 
---------------
 f
(1 row)
```

**Additional context**
The data node however properly reports the state:
```
# select is_compressed from timescaledb_information.chunks where hypertable_name='test';
 is_compressed 
---------------
 t
(1 row)
```

erimatnor:
The `chunks` view uses the presence of a local auxiliary compressed
chunk table to determine a chunk's compression status. This doesn't
work on distributed hypertables on the access node since no auxiliary
compressed chunk exists in that case.

This change fixes the issue by fetching a chunk's compression status
from one of its data nodes. In a future change, we might want to store
a chunk's compression status on the access node, e.g., as a column in
the chunk's metadata table or as a key-value pair in the chunk's
foreign table. However, that also risks having no source truth of
compression status in case it will still be possible to compress
locally on a data node.

Fixes #2724


erimatnor:
Closing since solution was rejected.

PR #2734, opened 2020-12-10
Adjust code to PG13 list sort changes

svenklemm:
PG13 changes the name of the list sorting function from list_qsort
to list_sort. Additionally PG13 does in-place sort.

https://github.com/postgres/postgres/commit/569ed7f483

### DIFF
--- a/src/bgw/scheduler.c
+++ b/src/bgw/scheduler.c
@@ -530,10 +530,15 @@ ts_populate_scheduled_job_tuple(ScheduledBgwJob *sjob, Datum *values)
 #endif
 
 static int
+#if PG13_LT
 cmp_next_start(const void *left, const void *right)
 {
 	const ListCell *left_cell = *((ListCell **) left);
 	const ListCell *right_cell = *((ListCell **) right);
+#else
+cmp_next_start(const ListCell *left_cell, const ListCell *right_cell)
+{
+#endif
 	ScheduledBgwJob *left_sjob = lfirst(left_cell);
 	ScheduledBgwJob *right_sjob = lfirst(right_cell);
 
@@ -549,10 +554,18 @@ cmp_next_start(const void *left, const void *right)
 static void
 start_scheduled_jobs(register_background_worker_callback_type bgw_register)
 {
+	List *ordered_scheduled_jobs;
 	ListCell *lc;
 	Assert(CurrentMemoryContext == scratch_mctx);
+
 	/* Order jobs by increasing next_start */
-	List *ordered_scheduled_jobs = list_qsort(scheduled_jobs, cmp_next_start);
+#if PG13_LT
+	ordered_scheduled_jobs = list_qsort(scheduled_jobs, cmp_next_start);
+#else
+	/* PG13 does in-place sort */
+	ordered_scheduled_jobs = scheduled_jobs;
+	list_sort(ordered_scheduled_jobs, cmp_next_start);
+#endif
 
 	foreach (lc, ordered_scheduled_jobs)
 	{


gayyappan (approved):

k-rus (commented):
Having PG directives directly in the code makes it more difficult to follow and also will bring issue of reusability.
To my understanding providing wrapper around the sort functions in `compat.h` should not be an issue. So it would be great to do it through compat module.

Can a function variable be re-casted? If so, then I believe it can be done inside the sot function wrapper.

Is there a good reason to not move PG difference into compat and keep here? I might miss something.

src/bgw/scheduler.c
```diff
@@ -548,10 +553,18 @@ cmp_next_start(const void *left, const void *right)
...
 	Assert(CurrentMemoryContext == scratch_mctx);
+
 	/* Order jobs by increasing next_start */
-	List *ordered_scheduled_jobs = list_qsort(scheduled_jobs, cmp_next_start);
+#if PG13_LT
+	ordered_scheduled_jobs = list_qsort(scheduled_jobs, cmp_next_start);
+#else
+	/* PG13 does in-place sort */
+	ordered_scheduled_jobs = scheduled_jobs;
+	list_sort(ordered_scheduled_jobs, cmp_next_start);
```
k-rus, line 566: This will modify `scheduled_jobs`, right? Should the PG13_LT version to replace `scheduled_jobs` with the result of sorting, so both code paths are semantically equivalent? (then it can be wrapped into compat function with the same semantics)

src/bgw/scheduler.c
```diff
@@ -529,10 +529,15 @@ ts_populate_scheduled_job_tuple(ScheduledBgwJob *sjob, Datum *values)
...
 static int
+#if PG13_LT
 cmp_next_start(const void *left, const void *right)
 {
 	const ListCell *left_cell = *((ListCell **) left);
 	const ListCell *right_cell = *((ListCell **) right);
+#else
+cmp_next_start(const ListCell *left_cell, const ListCell *right_cell)
+{
+#endif
```
k-rus, line 540: A suggestion to do it without directives:
```suggestion
static int
cmp_next_start_void(const void *left, const void *right)
{
	const ListCell *left_cell = *((ListCell **) left);
	const ListCell *right_cell = *((ListCell **) right);
	cmp_next_start(left_cell, right_cell);
}

static int
cmp_next_start(const ListCell *left_cell, const ListCell *right_cell)
{
```
`cmp_next_start` or its declaration should precede `cmp_next_start_void`. (the name is just suggestion)

src/bgw/scheduler.c
```diff
@@ -548,10 +553,18 @@ cmp_next_start(const void *left, const void *right)
...
+
 	/* Order jobs by increasing next_start */
-	List *ordered_scheduled_jobs = list_qsort(scheduled_jobs, cmp_next_start);
+#if PG13_LT
+	ordered_scheduled_jobs = list_qsort(scheduled_jobs, cmp_next_start);
+#else
+	/* PG13 does in-place sort */
+	ordered_scheduled_jobs = scheduled_jobs;
+	list_sort(ordered_scheduled_jobs, cmp_next_start);
+#endif
```
k-rus, line 567: This can be encapsulated into an ordering function, which will always produce a sorted copy. E.g.:
```c
static List *
list_sort_with_copy(List *origin) 
{
	List *new_list;

#if PG13_LT
	new_list = list_qsort(origin, cmp_next_start_void);
#else
	/* PG13 does in-place sort */
	new_list = list_copy(origin);
	list_sort(new_list, cmp_next_start);
#endif
}
```
Then:
```suggestion
	ordered_scheduled_jobs = order_list_with_copy(scheduled_jobs);
```

src/bgw/scheduler.c
k-rus: > Should the PG13_LT version to replace `scheduled_jobs` with the result of sorting

Dumb suggestion from me.

But copying the list before sorting by `list_sort` in PG13 should work.

k-rus (approved):
I really don't like the proposed implementation since:
1. PG13 and PG13_LT have different semantics: the first change the global list of scheduled jobs, while the second not.
2. Directives and PG version specific codes are not abstracted from the logic.

I added a proposal to address these in my comments, but the solution is not perfect.

PR #2736, opened 2020-12-10
Support ADD COLUMN for compressed hypertables

gayyappan:
Support ALTER TABLE .. ADD COLUMN <colname> <typname>
for hypertables with compressed chunks.


tsl/test/sql/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(TEST_FILES
   bgw_custom.sql
   bgw_policy.sql
   compression_bgw.sql
+  compression_alter.sql
```
erimatnor, line 5: Do we need a completely new test for this? We already have ~8 compression test files, including a `compression_ddl`. Every unique test file that we add will add to the test runtime. Maybe it can even be a shared test?

k-rus:
@gayyappan One test fails due to this PR.

tsl/test/sql/CMakeLists.txt
gayyappan: There will be more tests specific to alter table. I prefer to split the test files up. It is hard to figure out what is being tested with huge test files.
Moving it to a shared test might work.

k-rus (commented):
Can you add this PR to the change log? (after #2753 is merged)

tsl/src/compression/create.c
```diff
@@ -996,3 +1018,44 @@ tsl_process_compress_table(AlterTableCmd *cmd, Hypertable *ht,
 	/* do not release any locks, will get released by xact end */
 	return true;
 }
+
+/* Add a column to a compressed table */
```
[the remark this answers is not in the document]
gayyappan: Added a comment to clarify. I chose the name to be consistent with the other extern functions for this file.

src/cross_module_fn.c
```diff
@@ -160,6 +160,13 @@ process_compress_table_default(AlterTableCmd *cmd, Hypertable *ht,
 	pg_unreachable();
 }
 
+static void
+process_compress_table_add_column_default(int32 htid, int32 compress_htid, ColumnDef *orig_def)
```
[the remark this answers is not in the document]
gayyappan: added a new altercmd processing function to tsl/src/utility.c. This functionality has been moved there.

src/process_utility.c
```diff
@@ -238,6 +241,35 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
 	}
 }
 
+static void
+process_altertable_add_column_for_compressed(Hypertable *ht, AlterTableCmd *cmd)
```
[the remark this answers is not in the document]
gayyappan: done.

tsl/test/sql/CMakeLists.txt
gayyappan: This is now called from compression_ddl.sql so that we avoid the cost of a new db setup. This file moved to include directory.

src/process_utility.c
```diff
@@ -238,6 +241,35 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
...
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("cannot add column with constraints or defaults to a hypertable that has "
+						"compression enabled")));
+	}
+	/* not possible to get non-null value here this is set when
+	 * ALTER TABLE ALTER COLUMN ... SET TYPE < > USING ...
+	 * but check anyway.
+	 */
+	if (col->raw_default != NULL || col->cooked_default != NULL)
+		elog(ERROR, "unexpected default value found while adding column ");
```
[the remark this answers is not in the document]
gayyappan: modified.

tsl/src/compression/create.h
```diff
@@ -18,6 +18,8 @@
 
 bool tsl_process_compress_table(AlterTableCmd *cmd, Hypertable *ht,
 								WithClauseResult *with_clause_options);
+void tsl_process_compress_table_add_column(int32 orig_htid, int32 compress_htid,
+										   ColumnDef *orig_def);
```
k-rus, line 22: The name `orig_def` suggests that it will not be modified and it is not in the implementation, thus:
```suggestion
void tsl_process_compress_table_add_column(int32 orig_htid, int32 compress_htid,
										   const ColumnDef *orig_def);
```

tsl/src/process_utility.c
```diff
@@ -21,10 +22,24 @@ tsl_ddl_command_start(ProcessUtilityArgs *args)
 		DropdbStmt *stmt = castNode(DropdbStmt, args->parsetree);
 		remote_connection_cache_dropped_db_callback(stmt->dbname);
 	}
-
 	dist_ddl_start(args);
 }
 
+/* AlterTableCmds that need tsl side processing invoke this function
+ * we only process AddColumn command right now.
```
k-rus, line 29: It would be nice to improve the sentence grammatically to make it easier to read and have sentences syntactically correct, i.e., starting with a capital letter and a dot at the end. (I believe it is two sentences there)

tsl/src/process_utility.c
```diff
@@ -21,10 +22,24 @@ tsl_ddl_command_start(ProcessUtilityArgs *args)
 		DropdbStmt *stmt = castNode(DropdbStmt, args->parsetree);
 		remote_connection_cache_dropped_db_callback(stmt->dbname);
 	}
-
```
k-rus, line 24: Unrelated change, which is matter of taste.

tsl/src/process_utility.c
```diff
@@ -21,10 +22,24 @@ tsl_ddl_command_start(ProcessUtilityArgs *args)
...
 	dist_ddl_start(args);
 }
 
+/* AlterTableCmds that need tsl side processing invoke this function
+ * we only process AddColumn command right now.
+ */
+void
+tsl_process_altertable_cmd(Hypertable *ht, AlterTableCmd *cmd)
```
k-rus, line 32: Am I correct that both arguments are not changed in the function?
```suggestion
tsl_process_altertable_cmd(const Hypertable *ht, const AlterTableCmd *cmd)
```

tsl/test/expected/compression_ddl.out
```diff
@@ -685,3 +680,131 @@ WHERE hypertable.table_name like 'test1'  ORDER BY chunk.id ) as subq;
...
+CREATE TABLE test1 ("Time" timestamptz, intcol integer, bntcol bigint, txtcol text);
+SELECT table_name from create_hypertable('test1', 'Time', chunk_time_interval=> INTERVAL '1 day');
+psql:include/compression_alter.sql:6: NOTICE:  adding not-null constraint to column "Time"
+ table_name 
+------------
+ test1
+(1 row)
+
+INSERT INTO test1 SELECT t,  gen_rand_minstd(), gen_rand_minstd(), gen_rand_minstd()::text FROM generate_series('2018-03-02 1:00'::TIMESTAMPTZ, '2018-03-05 1:00', '1 hour') t;
+ALTER TABLE test1 set (timescaledb.compress, timescaledb.compress_segmentby = 'bntcol', timescaledb.compress_orderby = '"Time" DESC');
```
k-rus, line 698: It would be great if SQL statements fit the screen, e.g., 100 chars.

tsl/test/expected/compression_ddl.out
```diff
@@ -685,3 +680,131 @@ WHERE hypertable.table_name like 'test1'  ORDER BY chunk.id ) as subq;
...
+SELECT COUNT(*) AS count_compressed
+FROM
+(
+SELECT compress_chunk(chunk.schema_name|| '.' || chunk.table_name)
+FROM _timescaledb_catalog.chunk chunk
+INNER JOIN _timescaledb_catalog.hypertable hypertable ON (chunk.hypertable_id = hypertable.id)
+WHERE hypertable.table_name like 'test1' and chunk.compressed_chunk_id IS NULL ORDER BY chunk.id
```
k-rus, line 705: Can the subquery be indented?

tsl/test/expected/compression_ddl.out
```diff
@@ -685,3 +680,131 @@ WHERE hypertable.table_name like 'test1'  ORDER BY chunk.id ) as subq;
...
+
+--decompress 1 chunk and query again 
+SELECT COUNT(*) AS count_compressed
+FROM
+(
+SELECT decompress_chunk(chunk.schema_name|| '.' || chunk.table_name)
+FROM _timescaledb_catalog.chunk chunk
+INNER JOIN _timescaledb_catalog.hypertable hypertable ON (chunk.hypertable_id = hypertable.id)
+WHERE hypertable.table_name like 'test1' and chunk.compressed_chunk_id IS NOT NULL ORDER BY chunk.id
+LIMIT 1
```
k-rus, line 748: Can the subquery be indented?

tsl/test/expected/compression_ddl.out
```diff
@@ -685,3 +680,131 @@ WHERE hypertable.table_name like 'test1'  ORDER BY chunk.id ) as subq;
...
+    25
+(1 row)
+
+SELECT COUNT(*) AS count_compressed
+FROM
+(
+SELECT compress_chunk(chunk.schema_name|| '.' || chunk.table_name)
+FROM _timescaledb_catalog.chunk chunk
+INNER JOIN _timescaledb_catalog.hypertable hypertable ON (chunk.hypertable_id = hypertable.id)
+WHERE hypertable.table_name like 'test1' and chunk.compressed_chunk_id IS NULL ORDER BY chunk.id
```
k-rus, line 790: It will be great to indent

k-rus (commented):
Many nits, which need to be fixed.

I suggest to give comment what happens with the values of the added column, i.e., PG will set values to NULL per compressed row.

Can you add two tests?:
1. Select from the compressed hypertable to see some values including new column.
2. Insert data into new column in a not compressed chunk and do query, which involves both compressed and uncompressed chunks.
I expect that everything works, since it is nothing special with new column.
(I guess such tests already exist without adding column. One way is to run the same tests with no add column and with add column. However, such thing is better to do as separate PR, I believe)

gayyappan:
> Many nits, which need to be fixed.
> 
> I suggest to give comment what happens with the values of the added column, i.e., PG will set values to NULL per compressed row.
> 
> Can you add two tests?:
> 
> 1. Select from the compressed hypertable to see some values including new column.
> 2. Insert data into new column in a not compressed chunk and do query, which involves both compressed and uncompressed chunks.
>    I expect that everything works, since it is nothing special with new column.
>    (I guess such tests already exist without adding column. One way is to run the same tests with no add column and with add column. However, such thing is better to do as separate PR, I believe)

1. already have tests that selects from the table with filters on the new columns. 
2. Similarly have query which involves both compressed and uncompressed chunk after adding a column. Do you have a specific test in mind?

src/process_utility.c
```diff
@@ -239,6 +240,23 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
...
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("cannot add column with constraints or defaults to a hypertable that has "
+						"compression enabled")));
+	}
+	/* not possible to get non-null value here this is set when
+	 * ALTER TABLE ALTER COLUMN ... SET TYPE < > USING ...
+	 * but check anyway.
+	 */
+	Assert(col->raw_default == NULL && col->cooked_default == NULL);
```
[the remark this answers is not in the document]
gayyappan: I think the assert should be retained since this is a valid value in the struct. We might support DEFAULT values later.

tsl/src/compression/create.c
```diff
@@ -320,6 +322,26 @@ compresscolinfo_init(CompressColInfo *cc, Oid srctbl_relid, List *segmentby_cols
 	table_close(rel, AccessShareLock);
 }
 
+/* use this api for the case when you add a single column to a table that already has
+ * copression setup
```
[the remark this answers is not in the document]
gayyappan: fixed.

src/process_utility.c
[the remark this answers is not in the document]
k-rus: The values will not be not-null only for `ALTER COLUMN`, right? At least this is what I see in PG code for `raw_default`. Are you sure they can be not null in the case of `ADD COLUMN`?

tsl/src/process_utility.c
gayyappan: Right now, they do not. But this is supposed to be a more generic function i..e other AlterTablecmds that need tsl side processing would introduce additional if stmts here. So should not force it to be a const.

src/process_utility.c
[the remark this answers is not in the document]
gayyappan: The only path for this,  as far as I can tell, is via the stmt I mentioned in the comment. I prefer to be defensive with the code.

tsl/src/process_utility.c
```diff
@@ -21,10 +22,24 @@ tsl_ddl_command_start(ProcessUtilityArgs *args)
...
+{
+	if (cmd->subtype == AT_AddColumn || cmd->subtype == AT_AddColumnRecurse)
+	{
+		ColumnDef *orig_coldef = (ColumnDef *) cmd->def;
```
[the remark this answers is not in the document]
gayyappan: fixed.

tsl/src/process_utility.c
k-rus: `const` can be removed when the implementation changes. Currently it helps to understand the code.
btw, may be new lines between variable declarations, assertion and code can help for readability.

gayyappan: modified.

tsl/src/compression/create.c
```diff
@@ -320,6 +322,26 @@ compresscolinfo_init(CompressColInfo *cc, Oid srctbl_relid, List *segmentby_cols
...
+ * such as ALTER TABLE xyz ADD COLUMN .....
+ */
+static void
+compresscolinfo_init_singlecolumn(CompressColInfo *cc, char *colname, Oid typid)
```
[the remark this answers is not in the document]
gayyappan: modified.

src/process_utility.c
```diff
@@ -213,6 +213,7 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
 			case AT_SetTableSpace:
 				/* this is passed down in `process_altertable_set_tablespace_end` */
 			case AT_SetStatistics: /* should this be pushed down in some way? */
+			case AT_AddColumn:	 /* this is passed down */
```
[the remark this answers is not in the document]
gayyappan: being consistent with the comment above.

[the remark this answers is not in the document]
k-rus: However, the most of the rows above are without the comment. If the comment is important, it is good to have it more meaningful.

k-rus:
> 1. already have tests that selects from the table with filters on the new columns.

Sorry, I wasn't clear and missed to write that I thought about a test, which selects from compressed hypertable. However, I am not sure now if it is actually good idea.


> 2. Similarly have query which involves both compressed and uncompressed chunk after adding a column. Do you have a specific test in mind?

Nothing specific. Does this PR include test with mix of compressed and uncompressed chunks? I might not understand well enough the tests.

k-rus (commented):
I see that many my comments were not addressed, so I leave this comment to have an opportunity to re-request.

gayyappan:
> > 1. already have tests that selects from the table with filters on the new columns.
> 
> Sorry, I wasn't clear and missed to write that I thought about a test, which selects from compressed hypertable. However, I am not sure now if it is actually good idea.
> 
> > 1. Similarly have query which involves both compressed and uncompressed chunk after adding a column. Do you have a specific test in mind?
> 
> Nothing specific. Does this PR include test with mix of compressed and uncompressed chunks? I might not understand well enough the tests.

Yes, the tests have a mix of compressed and uncompressed chunks.

src/cross_module_fn.c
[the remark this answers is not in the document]
erimatnor: Sorry, I didn't mean to just move the function to `tsl/process_utility.c`; I meant that we do not need a cross-module function since we already forward ddl commands to TSL. If we add a cross-module function definition for every DDL command, including every alter-table variant, the cross-module interface is going to blow-up with a ton of functions.

Thus, unless I am missing something, you should be able to "capture" the alter table in TSL's process utility instead of capturing it outside TSL and then forwarding it to TSL via a cross-module function.

src/process_utility.c
```diff
@@ -3264,6 +3283,10 @@ process_altertable_end_subcmd(Hypertable *ht, Node *parsetree, ObjectAddress *ob
 			break;
 		case AT_AddColumn:
 		case AT_AddColumnRecurse:
+			Assert(IsA(cmd->def, ColumnDef));
+			if (TS_HYPERTABLE_HAS_COMPRESSION_TABLE(ht))
+				ts_cm_functions->process_altertable_cmd(ht, cmd);
```
erimatnor, line 3288: This cross-module call should not be necessary since we already forward DDL commands to TSL, which means you can capture it there instead of defining a new cross-module function. I don't believe it is a scalable approach to define new functions for every DDL and alter command.

tsl/src/compression/create.c
```diff
@@ -320,6 +322,26 @@ compresscolinfo_init(CompressColInfo *cc, Oid srctbl_relid, List *segmentby_cols
...
+{
+	int colno = 0;
+	Oid compresseddata_oid = ts_custom_type_cache_get(CUSTOM_TYPE_COMPRESSED_DATA)->type_oid;
+	ColumnDef *coldef;
+
+	cc->numcols = 1;
+	cc->col_meta = palloc0(sizeof(FormData_hypertable_compression) * 1);
```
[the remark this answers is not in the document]
svenklemm: agree with ruslan here this should use cc->numcols instead of constant

src/cross_module_fn.h
```diff
@@ -98,6 +98,7 @@ typedef struct CrossModuleFunctions
 	PGFunction compressed_data_out;
 	bool (*process_compress_table)(AlterTableCmd *cmd, Hypertable *ht,
 								   WithClauseResult *with_clause_options);
+	void (*process_altertable_cmd)(const Hypertable *ht, const AlterTableCmd *cmd);
```
svenklemm, line 101: agree with erik that ideally this should happen in tsl code to not further bloat the cross module stuff

gayyappan: The alter table cmd processing has been moved to tsl code. This is the function from cross_module that would reference tsl code.

src/process_utility.c
gayyappan: The earlier version of the function had subcmd specifc processing in src/process_utility.c. With this approach, I have  a more generic tsl side function for the ALTER TABLE ddl. This is invoked only when needed.

The downside of passing the entire processing over to ddl_command_start and ddl_command_end in the tsl code is: we would be replicating code for table lookup and subcmd processing on tsl side (see code path for process_altertable_end_table - has simple cmd and multiple cmds code path. If we were to do this in ddl_command_end, we  would replicate all this code,  but all this processing has already been done by the caller. I don't see this approach as simplifying the code) 

I can move the call at line 3288 to the end of the function process_altertable_end_subcmd to make it more generic. There is nothing that prevents the tsl  side function from being generic.

svenklemm (approved):

src/cross_module_fn.h
```diff
@@ -98,6 +98,7 @@ typedef struct CrossModuleFunctions
 	PGFunction compressed_data_out;
 	bool (*process_compress_table)(AlterTableCmd *cmd, Hypertable *ht,
 								   WithClauseResult *with_clause_options);
+	void (*process_altertable_cmd)(Hypertable *ht, const AlterTableCmd *cmd);
```
pmwkaa, line 101: I would try to avoid adding new functions here, for example we have `ddl_command_start` hook already which basically allows to handle process utility hook on the tsl side.

If you prefer this approach more, I suggest to make argument ordering consistent with the function below

tsl/src/compression/create.c
```diff
@@ -320,6 +322,26 @@ compresscolinfo_init(CompressColInfo *cc, Oid srctbl_relid, List *segmentby_cols
...
 }
 
+/* use this api for the case when you add a single column to a table that already has
+ * copression setup
+ * such as ALTER TABLE xyz ADD COLUMN .....
+ */
+static void
+compresscolinfo_init_singlecolumn(CompressColInfo *cc, char *colname, Oid typid)
+{
+	int colno = 0;
```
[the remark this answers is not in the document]
pmwkaa: Maybe just remove this variable at all, position logic can be identified by the function name

tsl/src/compression/create.c
```diff
@@ -900,6 +922,39 @@ disable_compression(Hypertable *ht, WithClauseResult *with_clause_options)
...
+	stmt->relkind = OBJECT_TABLE;
+	stmt->missing_ok = false;
+
+	/* alter the table and add column */
+	lockmode = AlterTableGetLockLevel(stmt->cmds);
+	compress_relid = AlterTableLookupRelation(stmt, lockmode);
+	Assert(compress_relid == compress_ht->main_table_relid);
+	AlterTable(compress_relid, AccessExclusiveLock, stmt);
+	modify_compressed_toast_table_storage(compress_cols, compress_relid);
+	return;
```
pmwkaa, line 955: Do we need that return?

tsl/src/compression/create.c
```diff
@@ -996,3 +1051,29 @@ tsl_process_compress_table(AlterTableCmd *cmd, Hypertable *ht,
...
+	coloid = LookupTypeNameOid(NULL, orig_typname, false);
+	compresscolinfo_init_singlecolumn(&compress_cols, colname, coloid);
+	if (TS_HYPERTABLE_HAS_COMPRESSION_TABLE(ht))
+	{
+		int32 compress_htid = ht->fd.compressed_hypertable_id;
+		Hypertable *compress_ht = ts_hypertable_get_by_id(compress_htid);
+		add_column_to_compression_table(compress_ht, &compress_cols);
+	}
+	else
+		Assert(TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht));
```
pmwkaa, line 1076: I believe it might fail if the extension will be build with `-DNDEBUG`  flag (or maybe not, if `;` will be kept), anyway I suggest to wrap this statement into the code block

pmwkaa (approved):
Couple of nits. I believe we had support for the DROP COLUMN before?

tsl/src/compression/create.c
gayyappan: removed.

tsl/src/compression/create.c
gayyappan: added

PR #2738, opened 2020-12-11
Adjust code to PG13 convert_tuples_by_name signature change

svenklemm:
PG13 removed the msg parameter from convert_tuples_by_name.

https://github.com/postgres/postgres/commit/fe66125974

### DIFF
--- a/src/chunk_insert_state.c
+++ b/src/chunk_insert_state.c
@@ -120,7 +120,7 @@ create_chunk_result_relation_info(ChunkDispatch *dispatch, Relation rel)
 
 static ProjectionInfo *
 get_adjusted_projection_info_returning(ProjectionInfo *orig, List *returning_clauses,
-									   AttrNumber *map, int map_size, Index varno, Oid rowtype,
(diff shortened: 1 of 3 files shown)


src/chunk_insert_state.c
```diff
@@ -379,8 +379,11 @@ setup_on_conflict_state(ChunkInsertState *state, ChunkDispatch *dispatch, AttrNu
 
 		if (NULL == chunk_attnos)
 			chunk_attnos = convert_tuples_by_name_map(RelationGetDescr(chunk_rel),
-													  RelationGetDescr(first_rel),
-													  gettext_noop("could not convert row type"));
+													  RelationGetDescr(first_rel)
+#if PG13_LT
+                            , gettext_noop("could not convert row type")
```
pmwkaa, line 384: Is there are a good reason why we use `gettext` here at all? maybe just pass this string as is

erimatnor: It's used to mark text for translation purposes.

erimatnor (approved):

gayyappan (approved):

PR #2742, opened 2020-12-11
Fix compression status in chunks view for distributed chunks

gayyappan:
The compression status of individual chunks for distributed tables is
unknown on the access node. Fix the view so that is_compressed is NULL
 instead of false for distributed chunks.

### DIFF
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -206,10 +206,12 @@ FROM (
     ELSE
       dimsl.range_end
     END AS integer_range_end,
-    CASE WHEN srcch.compressed_chunk_id IS NOT NULL THEN
-      TRUE
-    ELSE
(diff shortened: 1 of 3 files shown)


erimatnor (approved):
This also needs docs updates.

erimatnor (commented):
I suggest improving the title to be more descriptive of what is done. E.g., "Do not show compression status for distributed chunks"

pmwkaa (approved):

erimatnor (commented):
Does this need update script changes to remove/update the view? Seems like update tests passed, so maybe not...

mkindahl (approved):

gayyappan:
doc update PR 618

PR #2749, opened 2020-12-16
Run chunk_utils test in solo to reduce flakiness

k-rus:
chunk_utils test experiences some inconsistency in permissions
sometimes. To avoid this the test is run solo without other tests
interfering.

PR note: 
[link](https://github.com/timescale/timescaledb/runs/1563206122?check_suite_focus=true#step:15:25) to a failed test.

### DIFF
--- a/test/sql/CMakeLists.txt
+++ b/test/sql/CMakeLists.txt
@@ -59,6 +59,7 @@ set(SOLO_TESTS
   alter
   alternate_users
   bgw_launcher
+  chunk_utils.sql
   index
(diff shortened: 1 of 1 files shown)


pmwkaa (approved):

mkindahl (approved):
Since the table permissions for `drop_chunks_test1` does not seem to be modified in any other test (which would be running in parallel with this one, creating a race condition), I wonder if this is the issue, but approving since this does not create an issue and might solve one.

PR #2751, opened 2020-12-16
Fix crash and cancel when adding data node

erimatnor:
This change fixes two issues with `add_data_node`:

1. In one case, a check for a valid connection pointer was not done,
   causing a segmentation fault when connection attempts failed.

2. Connections were made with a blocking API that would hang
   indefinitely if the receiving end is not responding. The user
   couldn't cancel the connection attempt with CTRL-C, since no wait
   latch or interrupt checking was used. The code is now updated to
   use a non-blocking connection API, where it is possible to wait on
   the socket and latch, respecting interrupts.

This was tested locally by setting up a bogus listening socket with `netcat -l` that would not respond to connection attempts by `add_data_node`. Without the fix, the command could not be canceled with CTRL-C. With the update code, this works as expected. Naturally, this is a test case that is difficult to test for in the regression tests.

### DIFF
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -726,6 +726,12 @@ data_node_add_internal(PG_FUNCTION_ARGS, bool set_distid)
 		{
 			TSConnection *conn =
 				connect_for_bootstrapping(node_name, host, port, username, password);
+
+			if (NULL == conn)
+				ereport(ERROR,
+						(errcode(ERRCODE_SQLCLIENT_UNABLE_TO_ESTABLISH_SQLCONNECTION),
+						 errmsg("could not connect to \"%s\"", node_name)));
+
 			data_node_validate_extension_availability(conn);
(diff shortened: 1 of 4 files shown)


mkindahl (approved):

k-rus (approved):

pmwkaa (approved):
Looks good to me.

I think it might be possible to test it by setting statement timeout, it will likely go throw `CHECK_FOR_INTERRUPTS()` as well

erimatnor:
> statement

Good point. I'll try it.

erimatnor:
> > statement
> 
> Good point. I'll try it. Although I think the issue is having a remote endpoint that accepts a connection but is not responding.

Couldn't get this to work. I think one way to do it is to set up a listening socket in a separate thread (C code) and then connect to that. But that's a bit more work than I think we can afford right now, so I will look into that as a separate PR.

PR #2763, opened 2020-12-19
Fix check constraint on hypertable metadata table

erimatnor:
The `replication_factor` is set to `-1` on hypertables that are
created on data nodes as part of a larger distributed
hypertable. However, the check constraint on the hypertable metadata
table doesn't allow such values, causing update scripts to fail when
this check constraint is recreated as part of updating to version
`2.0.0-rc4`.

The reason it is possible to insert violating rows is because check
constraints aren't validated when inserting data using PostgreSQL's
internal catalog functions (in C). Therefore, the violating row can
exist until one tries to update a data node to `2.0.0-rc4`, at which
point the update script tries to recreate the `hypertable` metadata
table due to other changes that were made to the table.

This change fixes the check constraint to account for `-1` as a valid
value, and also changes the update scripts to account for the new
check constraint so that updates to the latest version will no longer
fail.

### DIFF
--- a/scripts/test_update_from_tag.sh
+++ b/scripts/test_update_from_tag.sh
@@ -80,8 +80,10 @@ docker_logs() {
 }
 
 docker_pgcmd() {
+    local database=${3:-single}
+    echo "executing pgcmd on database $database"
     set +e
-    docker_exec $1 "psql -h localhost -U postgres -d single -v VERBOSITY=verbose -c \"$2\""
+    docker_exec $1 "psql -h localhost -U postgres -d $database -v VERBOSITY=verbose -c \"$2\""
     if [ $? -ne 0 ]; then
       docker_logs $1
       exit 1
@@ -90,13 +92,15 @@ docker_pgcmd() {
 }
 
 docker_pgscript() {
-    docker_exec $1 "psql -h localhost -U postgres -v ON_ERROR_STOP=1 -f $2"
+    local database=${3:-postgres}
+    docker_exec $1 "psql -h localhost -U postgres -d $database -v ON_ERROR_STOP=1 -f $2"
 }
 
 docker_pgtest() {
+    local database=${3:-single}
     set +e
     >&2 echo -e "\033[1m$1\033[0m: $2"
-    docker exec $1 psql -X -v ECHO=ALL -v ON_ERROR_STOP=1 -h localhost -U postgres -d single -f $2 > ${TEST_TMPDIR}/$1.out
+    docker exec $1 psql -X -v ECHO=ALL -v ON_ERROR_STOP=1 -h localhost -U postgres -d $database -f $2 > ${TEST_TMPDIR}/$1.out
     if [ $? -ne 0 ]; then
       docker_logs $1
       exit 1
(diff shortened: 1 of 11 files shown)


k-rus:
@erimatnor Can you improve the commit message?
- It is difficult to understand what didn't work. Am I correct that the issue is that data nodes are not updated correctly when users update from RC1-3 to the release 2.0.0, and thus update TimescaleDB will bring multinode to non-working state?
- Can the message be neutral without negativity of `unfortunately` and code `wasn't updated`? IMHO both are pointing that we missed to do correct implementation, but this is internal for our improvements, and not for public message.

If I am correct about the first, then I believe a test is needed.

erimatnor:
> @erimatnor Can you improve the commit message?
> 
>     * It is difficult to understand what didn't work. Am I correct that the issue is that data nodes are not updated correctly when users update from RC1-3 to the release 2.0.0, and thus update TimescaleDB will bring multinode to non-working state?
> 

The update will fail, but the nodes will remain in working state on the old extension version.

>     * Can the message be neutral without negativity of `unfortunately` and code `wasn't updated`? IMHO both are pointing that we missed to do correct implementation, but this is internal for our improvements, and not for public message.
> 
We use the same PostgreSQL functions to update catalog tables as PostgreSQL uses itself, so this is not about us not doing correct implementation. This is PostgreSQL: it doesn't validate constraints unless you run a full query plan and executor. I will nevertheless try to change the commit message to your liking.

> 
> If I am correct about the first, then I believe a test is needed.

pmwkaa (approved):

mkindahl (approved):

erimatnor:
Working on a test for this updating the extension on data nodes, but I am not sure I will be done for this PR. I did manually test this fix, so can add the test case later.

sql/updates/latest-dev.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE _timescaledb_catalog.hypertable DROP CONSTRAINT hypertable_replication_factor_check, ADD CONSTRAINT hypertable_replication_factor_check CHECK (replication_factor > 0 OR replication_factor = -1);
```
k-rus, line 1: Why is this constraint needed? There is already unnamed constraint created.
And if it needed, can it be split to multiple lines for readability?

k-rus (commented):
See my comment.
Is it possible to test the update of multinode?

sql/updates/latest-dev.sql
erimatnor: This like alters the existing constraint, which is incorrect. The only way to "change" constraints is to drop and add it back again with the new definition.

k-rus: > This line alters the existing constraint, which is incorrect.

If this line is incorrect, why it is still kept. I guess you meant something else :)

The constraint is re-created during `_timescaledb_catalog.hypertable` re-creation after the drop. So I don't understand why constraint `hypertable_replication_factor_check` is needed. I guess I still miss something.

k-rus: It is still unclear why it is needed. See my comment above.

scripts/test_update_from_tag.sh
```diff
@@ -90,13 +92,15 @@ docker_pgcmd() {
 }
 
 docker_pgscript() {
-    docker_exec $1 "psql -h localhost -U postgres -v ON_ERROR_STOP=1 -f $2"
+    local database=${3:-postgres}
+    docker_exec $1 "psql -h localhost -U postgres -d $database -v ON_ERROR_STOP=1 -f $2"
 }
 
 docker_pgtest() {
+    local database=${3:-single}
```
k-rus, line 100: With current implementation the argument will be always set, so may be just use `$3` directly?

k-rus (commented):
I have only one question, which I still don't understand https://github.com/timescale/timescaledb/pull/2763#discussion_r546592996

Otherwise the update change seems good.

sql/updates/latest-dev.sql
erimatnor: The existing old constraint is incorrect. That's why it is recreated. I guess what you are saying is that this isn't needed because the update script for rc4 was already changed with this fix.

This change accounts for updates from RC4 installations to latest (2.0.0).

k-rus (approved):

sql/updates/2.0.0-rc3--2.0.0-rc4.sql
```diff
@@ -50,12 +50,16 @@ CREATE TABLE IF NOT EXISTS _timescaledb_catalog.hypertable(
...
   compression_state smallint NOT NULL DEFAULT 0,
   compressed_hypertable_id integer ,
-  replication_factor smallint NULL CHECK (replication_factor > 0),
+  replication_factor smallint NULL,
   UNIQUE (associated_schema_name, associated_table_prefix),
   CONSTRAINT hypertable_table_name_schema_name_key UNIQUE (table_name, schema_name),
-----internal compressed hypertables have compression state = 2
+  -- internal compressed hypertables have compression state = 2
   CONSTRAINT hypertable_dim_compress_check CHECK (num_dimensions > 0 OR compression_state = 2),
   CONSTRAINT hypertable_compress_check CHECK ( (compression_state = 0 OR compression_state = 1 )  OR (compression_state = 2 AND compressed_hypertable_id IS NULL))
```
k-rus, line 58:
```suggestion
  CONSTRAINT hypertable_compress_check CHECK ( (compression_state = 0 OR compression_state = 1 )  OR (compression_state = 2 AND compressed_hypertable_id IS NULL)),
```

PR #2772, opened 2020-12-22
Always validate existing database and extension

erimatnor:
This change ensures the database and extension is validated whenever
these objects aren't created, instead of only doing validation when
`bootstrap=>false` is passed when adding a data node.

This fixes a corner case where a data node could be added and removed
several times, even though the data node's database was already marked
as having been part of a multi-node setup.

### DIFF
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -454,7 +454,7 @@ data_node_validate_as_data_node(TSConnection *conn)
 	if (PQresultStatus(res) != PGRES_TUPLES_OK)
 		ereport(ERROR,
 				(errcode(ERRCODE_TS_DATA_NODE_INVALID_CONFIG),
-				 (errmsg("%s is not valid as data node", remote_connection_node_name(conn)),
+				 (errmsg("cannot add \"%s\" as a data node", remote_connection_node_name(conn)),
 				  errdetail("%s", PQresultErrorMessage(res)))));
 
 	remote_result_close(res);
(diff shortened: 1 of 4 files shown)


k-rus (approved):
I suggest to merge the commits into one, since the test commit checks that the non-bootrstrap option works and will fail without the first commit. I hope I haven't missed anything in the logic.

pmwkaa (approved):

erimatnor:
> I suggest to merge the commits into one, since the test commit checks that the non-bootrstrap option works and will fail without the first commit. I hope I haven't missed anything in the logic.

All tests pass in both commits, independently. The second commit adds a test case that covers the case that didn't work as expected. Although I can certainly squash this, I personally don't see any issues with separate change into logical units.

k-rus:
> > I suggest to merge the commits into one, since the test commit checks that the non-bootrstrap option works and will fail without the first commit. I hope I haven't missed anything in the logic.
> 
> All tests pass in both commits, independently. The second commit adds a test case that covers the case that didn't work as expected. Although I can certainly squash this, I personally don't see any issues with separating changes into logical units.

The first commit implements validation of database and extension if bootstrap is false. However, I see no tests in the this commit, which checks this. It is good practice to have test in the same commit as implementation for several reasons:
1. It helps to understand the commit better.
2. It allows to cherry pick or rollback the implementation with the regression test together.

If I understood the second commit correctly, it includes a test for the first commit. Correct me if I am wrong and test is missing.

Issue #2770, opened 2020-12-22
Unable to build from source


zdm:
**Relevant system information:**
 - OS: Fedora 33
 - PostgreSQL version 12.5
 - TimescaleDB version 2.0.0
 - Installation method: "source"

**Describe the bug**
I get the error below when trying to build it from source.

1. clone and checkout latest release;
2. run:
    ```
    ./bootstrap -DREGRESS_CHECKS=OFF
    cd build && gmake
    ```

```
[ 33%] Building C object src/CMakeFiles/timescaledb.dir/jsonb_utils.c.o
[ 34%] Building C object src/CMakeFiles/timescaledb.dir/license_guc.c.o
[ 34%] Building C object src/CMakeFiles/timescaledb.dir/partitioning.c.o
In file included from /usr/pgsql-12/include/server/postgres.h:46,
                 from /root/rpmbuild/BUILD/timescaledb/src/partitioning.c:6:
/root/rpmbuild/BUILD/timescaledb/src/partitioning.c: In function 'ts_partitioning_info_create':
/root/rpmbuild/BUILD/timescaledb/src/partitioning.c:187:2: error: 'strncpy' specified bound 64 equals destination size [-Werror=stringop-truncation]
  187 |  StrNCpy(pinfo->partfunc.name, partfunc, NAMEDATALEN);
      |  ^~~~~~~
/root/rpmbuild/BUILD/timescaledb/src/partitioning.c:188:2: error: 'strncpy' specified bound 64 equals destination size [-Werror=stringop-truncation]
  188 |  StrNCpy(pinfo->column, partcol, NAMEDATALEN);
      |  ^~~~~~~
/root/rpmbuild/BUILD/timescaledb/src/partitioning.c:196:2: error: 'strncpy' specified bound 64 equals destination size [-Werror=stringop-truncation]
  196 |  StrNCpy(pinfo->partfunc.schema, schema, NAMEDATALEN);
      |  ^~~~~~~
cc1: all warnings being treated as errors
gmake[2]: *** [src/CMakeFiles/timescaledb.dir/build.make:628: src/CMakeFiles/timescaledb.dir/partitioning.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:1011: src/CMakeFiles/timescaledb.dir/all] Error 2
gmake: *** [Makefile:149: all] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.u8LEud (%build)


RPM build errors:
[...]

PR #2776, opened 2020-12-28
Use cppflags and ldflags from pg_config.

alexeyklyukin:
Unlike projects built with PGXS infrastructure, Timescale provided
CMakeLists.txt doesn't include CPPFLAGS and LDFLAGS used when building
PostgreSQ. As a result, TimescaleDB compilation fails in the PostgreSQL
code, i.e. being unable to localte neccessary includes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,6 +253,11 @@ get_pg_config(PG_CPPFLAGS --cppflags)
 get_pg_config(PG_LDFLAGS --ldflags)
 get_pg_config(PG_LIBS --libs)
 
+string(REGEX REPLACE " -I" ";" EXTRA_INCLUDES   ${PG_CPPFLAGS})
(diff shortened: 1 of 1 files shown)


alexeyklyukin:
Note: this worked to resolve the issue for me, but I'm not a CMake expert; in particular, the setting of LDFLAGS is quite naive, as they don't take into account non -L options allowed there.

erimatnor:
@alexeyklyukin  Apologies for taking time to give you feedback on this. While I think the intuition here is correct (we do want to propagate flags), we are pondering other solutions. Since the PR currently isn't passing tests, I will close it and keep the issue open. I think there was also a suggestion to fix this as part of fixing a related issue: https://github.com/timescale/timescaledb/issues/2770

alexeyklyukin:
Yes @erimatnor , that was basically a demo, the tests indeed didn't pass.
Glad to know you are working on a better approach.

PR #2796, opened 2021-01-08
Handle AT_DropExpression in process_utility

svenklemm:
PG13 adds a new ALTER TABLE subcommand for dropping the generated
property from a column.

https://github.com/postgres/postgres/commit/f595117e24


gayyappan (approved):
Looks good. Would be good to add a test case for this.

pmwkaa (approved):
+1 for test case, if possible

PR #2822, opened 2021-01-14
Improve logic of placing dispatches into plan

k-rus:
TimescaleDB planner adds a chunk dispatch node to a query plan. If a
hypertable is distributed, a data node dispatch node is also added
into the query plan. However, it wasn't clear, since the logic flow
was either to add the chunk dispatch or the data node dispatch, which
encapsulated the chunk dispatch. This minor refactor makes it more
clear to help with code understanding.

### DIFF
--- a/src/hypertable_insert.c
+++ b/src/hypertable_insert.c
@@ -471,16 +471,16 @@ ts_hypertable_insert_path_create(PlannerInfo *root, ModifyTablePath *mtpath)
 								"constraints"),
 						 errhint("Use column names to infer indexes instead.")));
 
+			subpath = ts_chunk_dispatch_path_create(root, mtpath, rti, i);
 			if (hypertable_is_distributed(ht) && ts_guc_max_insert_batch_size > 0)
 			{
 				/* Remember that this will become a data node dispatch plan. We
 				 * need to know later whether or not to plan this using the
 				 * FDW API. */
 				data_node_dispatch_plans = bms_add_member(data_node_dispatch_plans, i);
-				subpath = ts_cm_functions->data_node_dispatch_path_create(root, mtpath, rti, i);
+				subpath =
+					ts_cm_functions->data_node_dispatch_path_create(root, mtpath, subpath, rti, i);
 			}
-			else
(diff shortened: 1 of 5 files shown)


gayyappan (commented):
FWIW I find the existing code with separate if-else for distributed versus non-distributed case easier to follow and maintain.  The subpath  created (by the if-else) is the one that gets added to ModifyTablePath's list of paths. 
DispatchPath decides how to handle paths below it. DispatchPlan and State then deal with them appropriately.

k-rus:
@gayyappan The original code:
```c
			if (hypertable_is_distributed(ht) && ts_guc_max_insert_batch_size > 0)
 			{
 				data_node_dispatch_plans = bms_add_member(data_node_dispatch_plans, i);
 				subpath = ts_cm_functions->data_node_dispatch_path_create(root, mtpath, rti, i);
 			}
 			else
 				subpath = ts_chunk_dispatch_path_create(root, mtpath, rti, i);
```
This clearly sounds that the data node dispatch path is created in the case of distributed hypertable, and the chunk dispatch path is created in the case of non-distributed hypertable. However, it is not correct. In the case of the distributed hypertable it creates first the chunk dispatch path and then the data node dispatch path. This needs to be clear in the code, since otherwise it is difficult to follow.
Can you propose a solution, which will improve the current code without drawback of my solution?

src/hypertable_insert.c
```diff
@@ -471,16 +471,16 @@ ts_hypertable_insert_path_create(PlannerInfo *root, ModifyTablePath *mtpath)
 								"constraints"),
 						 errhint("Use column names to infer indexes instead.")));
 
+			subpath = ts_chunk_dispatch_path_create(root, mtpath, rti, i);
```
erimatnor, line 474: I am OK with this change, but it is not an obvious improvement IMO. Usually, you build a tree of paths recursively, so it is logical that the parent creates the child, which in turn creates its own children and so on recursively. This change breaks that pattern where the hypertable_insert path creates the child of the child. This breaks the normal pattern.

erimatnor (approved):

PR #2832, opened 2021-01-15
Test building against PG13 in CI

svenklemm:


pmwkaa (approved):

mkindahl (approved):
Looks good. The failing PG11 and PG12 tests looks like flakes, but please make sure that there is nothing else.

PR #2862, opened 2021-01-25
Improve CI coredump detection

svenklemm:
Occasionally tests would segfault but the coredump would not be
uploaded as artefact because the coredump was still in progress.
This patch changes CI to wait for potential in progress coredumps
to upload coredumps more reliably.


k-rus (approved):

mkindahl (approved):
The best would be if there is any way to check that a coredump is expected (e.g., checking the signal of the crashed pid), but there seems to be no easy way to do this. My worry is that 10 seconds is not enough, but at least this will improve the situation, so approving.

PR #2865, opened 2021-01-26
Apply volatile function quals at decompresschunk

### LINKED ISSUE #2864: Filters with Function Calls not applied on compressed chunks
 - OS: [e.g. Ubuntu 16.04, Windows 10 x64, etc]: Docker / timescale/timescaledb:1.7.4-pg12

**Describe the bug**
When a function is called as part of a WHERE clause, it does not seem to evaulate that filter on compressed chunks. It's unclear if it's specific to this particular kind of function, or not, but it does reproduce reliably, and unexpectedly.

This function is being used as you [cannot use row level security with compressed hypertables](https://github.com/timescale/timescaledb/blob/3afdf7cc725d089d0e30b07459681f87c2068041/tsl/src/compression/create.c#L983-L987), and perhaps this is why?

**To Reproduce**
```sql
-- create tables and hypertables
create table user_acl (accessid int[], username text);

create table testfunchyper (time timestamp, accessid int, name text, val int);
select create_hypertable('testfunchyper', 'time');

create table testfunchypercompress (time timestamp, accessid int, name text, val int);
select create_hypertable('testfunchypercompress', 'time');

-- insert values into hypertabes. Note this pre-suposes the postgres user. If you're using another user, adjust as appropriate

insert into user_acl VALUES ('{1,2,3}', 'postgres');
insert into testfunchyper VALUES (now(), 1, 'a', 3), 
(now() - interval '1 month', 2, 'a', 3), 
(now() - interval '2 months', 4, 'a', 3);
insert into testfunchypercompress VALUES (now(), 1, 'a', 3), 
(now() - interval '1 month', 2, 'a', 3), 
(now() - interval '2 months', 4, 'a', 3);

--enable and then compress chunks
alter table testfunchypercompress set (timescaledb.compress);

SELECT compress_chunk(i) from show_chunks('testfunchypercompress', older_than => INTERVAL '1 week') i;

-- create the function that makes the check
CREATE OR REPLACE FUNCTION public.checkfunction(tid bigint)
RETURNS boolean
security invoker
LANGUAGE SQL
AS
$$
    select exists (select accessid from user_acl where username = current_user and tid = any(accessid));
$$ ;

-- this one works as expected, and its explain shows the filter being applied. The row with accessid = 4 is NOT shown.
select * from testfunchyper where checkfunction(accessid);
explain (analyze) select * from testfunchyper where checkfunction(accessid);


-- this one doesn't work as expected, all rows are shown. Its explain does NOT show the filter being applied.
select * from testfunchypercompress where checkfunction(accessid);
explain (analyze) select * from testfunchypercompress where checkfunction(accessid);
```
**Expected behavior**
I would expect this filter to be equally applied whether the chunk is compressed or not.

**Actual behavior**
The filter doesn't seem to be evaulated at all on compressed chunks, at least according to the explain plans:
Explain without compression (note the filter being applied):
```
Custom Scan (ChunkAppend) on testfunchyper  (cost=0.00..872.62 rows=1071 width=48) (actual time=0.204..0.462 rows=2 loops=1)
  ->  Seq Scan on _hyper_230_1435_chunk  (cost=0.00..290.88 rows=357 width=48) (actual time=0.203..0.204 rows=1 loops=1)
        Filter: checkfunction((accessid)::bigint)
  ->  Seq Scan on _hyper_230_1436_chunk  (cost=0.00..290.88 rows=357 width=48) (actual time=0.129..0.130 rows=1 loops=1)
        Filter: checkfunction((accessid)::bigint)
  ->  Seq Scan on _hyper_230_1437_chunk  (cost=0.00..290.88 rows=357 width=48) (actual time=0.125..0.125 rows=0 loops=1)
        Filter: checkfunction((accessid)::bigint)
        Rows Removed by Filter: 1
Planning Time: 0.347 ms
Execution Time: 0.562 ms
```

Explain with compression (note the filter is only applied on the uncompressed chunk):

```
Custom Scan (ChunkAppend) on testfunchypercompress  (cost=0.00..319.88 rows=900357 width=48) (actual time=0.175..0.200 rows=3 loops=1)
  ->  Seq Scan on _hyper_231_1438_chunk  (cost=0.00..290.88 rows=357 width=48) (actual time=0.174..0.175 rows=1 loops=1)
        Filter: checkfunction((accessid)::bigint)
  ->  Custom Scan (DecompressChunk) on _hyper_231_1439_chunk  (cost=0.03..14.50 rows=450000 width=48) (actual time=0.013..0.014 rows=1 loops=1)
        ->  Seq Scan on compress_hyper_232_1441_chunk  (cost=0.00..14.50 rows=450 width=148) (actual time=0.006..0.007 rows=1 loops=1)
  ->  Custom Scan (DecompressChunk) on _hyper_231_1440_chunk  (cost=0.03..14.50 rows=450000 width=48) (actual time=0.008..0.009 rows=1 loops=1)
        ->  Seq Scan on compress_hyper_232_1442_chunk  (cost=0.00..14.50 rows=450 width=148) (actual time=0.006..0.006 rows=1 loops=1)
Planning Time: 0.811 ms
Execution Time: 0.351 ms
```
**Additional context**
Simple filters get applied as expected, e.g. `WHERE NOT accessid = 4` (explain below). It seems like the part that fails is when it has to evaluate the function. 

```
Append  (cost=0.00..4557.70 rows=901065 width=48) (actual time=0.015..0.041 rows=2 loops=1)
  ->  Seq Scan on _hyper_231_1438_chunk  (cost=0.00..23.38 rows=1065 width=48) (actual time=0.014..0.015 rows=1 loops=1)
        Filter: (accessid <> 4)
  ->  Custom Scan (DecompressChunk) on _hyper_231_1439_chunk  (cost=0.03..14.50 rows=450000 width=48) (actual time=0.014..0.015 rows=1 loops=1)
        Filter: (accessid <> 4)
        ->  Seq Scan on compress_hyper_232_1441_chunk  (cost=0.00..14.50 rows=450 width=148) (actual time=0.006..0.006 rows=1 loops=1)
  ->  Custom Scan (DecompressChunk) on _hyper_231_1440_chunk  (cost=0.03..14.50 rows=450000 width=48) (actual time=0.009..0.009 rows=0 loops=1)
        Filter: (accessid <> 4)
        Rows Removed by Filter: 1
        ->  Seq Scan on compress_hyper_232_1442_chunk  (cost=0.00..14.50 rows=450 width=148) (actual time=0.005..0.006 rows=1 loops=1)
Planning Time: 0.729 ms
Execution Time: 0.091 ms
```

gayyappan:
Volatile functions that are in chunk's baserestrictinfo
list were not correctly handled when we inserted a
DecompressChunk node. This PR adds these quals as a filter
to the DecompressChunk node.

Fixes #2864

### DIFF
--- a/tsl/test/expected/transparent_decompression_queries.out
+++ b/tsl/test/expected/transparent_decompression_queries.out
@@ -163,3 +163,42 @@ GROUP BY 2, 3;
 (27 rows)
 
 RESET enable_hashagg;
+-- test if volatile function quals are applied to compressed chunks
+CREATE FUNCTION check_equal_228( intval INTEGER) RETURNS BOOL 
+LANGUAGE PLPGSQL AS
+$BODY$
+DECLARE
+    retval BOOL;
+BEGIN
+   IF intval = 228 THEN RETURN TRUE;
+   ELSE RETURN FALSE;
+   END IF;
+END;
+$BODY$;
+-- the function cannot be pushed down to compressed chunk
+-- but should be applied as a filter on decompresschunk
+SELECT * from test_chartab 
+WHERE check_equal_228(rtt) ORDER BY ts;
+ job_run_id |      mac_id      | rtt |              ts              
+------------+------------------+-----+------------------------------
+       8864 | 001407000001DD2E | 228 | Sat Dec 14 02:52:05.863 2019
+       8890 | 001407000001DD2E | 228 | Fri Dec 20 02:52:05.863 2019
+(2 rows)
+
+EXPLAIN (analyze,costs off,timing off,summary off) 
+SELECT * from test_chartab 
+WHERE check_equal_228(rtt) and ts < '2019-12-15 00:00:00' order by ts;
+                                                   QUERY PLAN                                                    
+-----------------------------------------------------------------------------------------------------------------
+ Sort (actual rows=1 loops=1)
+   Sort Key: test_chartab.ts
+   Sort Method: quicksort 
+   ->  Custom Scan (ChunkAppend) on test_chartab (actual rows=1 loops=1)
+         Chunks excluded during startup: 0
+         ->  Custom Scan (DecompressChunk) on _hyper_1_1_chunk (actual rows=1 loops=1)
+               Filter: ((ts < 'Sun Dec 15 00:00:00 2019'::timestamp without time zone) AND check_equal_228(rtt))
+               Rows Removed by Filter: 2
+               ->  Seq Scan on compress_hyper_2_3_chunk (actual rows=3 loops=1)
+                     Filter: (_ts_meta_min_1 < 'Sun Dec 15 00:00:00 2019'::timestamp without time zone)
+(10 rows)
+
--- a/tsl/test/sql/transparent_decompression_queries.sql
+++ b/tsl/test/sql/transparent_decompression_queries.sql
@@ -67,3 +67,24 @@ GROUP BY 2, 3;
 
 RESET enable_hashagg;
 
+-- test if volatile function quals are applied to compressed chunks
+CREATE FUNCTION check_equal_228( intval INTEGER) RETURNS BOOL 
+LANGUAGE PLPGSQL AS
+$BODY$
+DECLARE
+    retval BOOL;
+BEGIN
+   IF intval = 228 THEN RETURN TRUE;
+   ELSE RETURN FALSE;
+   END IF;
+END;
+$BODY$;
+
+-- the function cannot be pushed down to compressed chunk
+-- but should be applied as a filter on decompresschunk
+SELECT * from test_chartab 
+WHERE check_equal_228(rtt) ORDER BY ts;
+
+EXPLAIN (analyze,costs off,timing off,summary off) 
+SELECT * from test_chartab 
+WHERE check_equal_228(rtt) and ts < '2019-12-15 00:00:00' order by ts;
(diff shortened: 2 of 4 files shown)


svenklemm (approved):

tsl/test/sql/transparent_decompression_queries.sql
```diff
@@ -67,3 +67,22 @@ GROUP BY 2, 3;
 
 RESET enable_hashagg;
 
+-- test if volatile function quals are applied to compressed chunks
+CREATE FUNCTION check_equal_228( intval INTEGER) RETURNS BOOL 
+LANGUAGE PLPGSQL AS
```
erimatnor, line 72: I know that functions are volatile by default, but it might make sense to make it explicit given that this is what we test for.

```suggestion
LANGUAGE PLPGSQL VOLATILE AS
```

tsl/test/expected/transparent_decompression_queries.out
```diff
@@ -163,3 +163,40 @@ GROUP BY 2, 3;
...
+$BODY$
+DECLARE
+    retval BOOL;
+BEGIN
+   IF intval = 228 THEN RETURN TRUE;
+   ELSE RETURN FALSE;
+   END IF;
+END;
+$BODY$;
+SELECT * from test_chartab
```
erimatnor, line 178: What is the expected output? What is the output in the failure case prior to this fix? Would help to add a few notes about this in a comment so that one knows what to look for.

pmwkaa (approved):

erimatnor (approved):
Approving, but suggested some improvements for clarity.

tsl/test/expected/transparent_decompression_queries.out
gayyappan: added comments.

PR #2866, opened 2021-01-27
Avoid partitionwise planning of partialize_agg

### LINKED ISSUE #2849: Incorrect result of count with enabled partitionwise aggregate on multinode
Work around: set `enable_partitionwise_aggregate = off` on all data nodes.

 - OS: Mac
 - PostgreSQL version (output of `postgres --version`): 12.2
 - TimescaleDB version (output of `\dx` in `psql`): 2.0.0
 - Installation method: source

**Describe the bug**
A query with count over distributed hypertable produces incorrect result when `partitionwise_aggregate` is enabled.

**To Reproduce**
1. Enable partitionwise aggregation by `enable_partitionwise_aggregate = on` in `postgresql.conf`
2. Create a distributed hypertable with space dimension and no replication on 2 data nodes and insert data
3. Run `select count(*)` on the distributed hypertable, which will return incorrect result.

Setup
```sql
select add_data_node('dn1', host => 'localhost', database => 'dn1');
select add_data_node('dn2', host => 'localhost', database => 'dn2');
create table dist(time timestamptz not null, device int, value float);
select create_distributed_hypertable('dist','time','device');
insert into dist select t, (random()*20)::int, random()*100 - 50 from generate_series('2020-01-10'::timestamp, '2020-04-03'::timestamp, '1m') t;
INSERT 0 120961
``` 
Repro incorrect result
```sql
select count(*) from dist;
 count
-------
  8700
(1 row)

explain verbose select count(*) from dist;
                                                                                                                                                                   QUERY PLAN                                                                                                               
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=970931.95..970931.96 rows=1 width=8)
   Output: count(*)
   ->  Custom Scan (AsyncAppend)  (cost=78743.19..970931.95 rows=2 width=8)
         Output: (PARTIAL count(*))
         ->  Append  (cost=78743.19..970931.95 rows=2 width=8)
               ->  Custom Scan (DataNodeScan)  (cost=78743.19..485465.97 rows=1 width=8)
                     Output: (PARTIAL count(*))
                     Relations: Aggregate on (public.dist)
                     Data node: dn1
                     Chunks: _dist_hyper_1_2_chunk, _dist_hyper_1_3_chunk, _dist_hyper_1_6_chunk, _dist_hyper_1_7_chunk, _dist_hyper_1_10_chunk, _dist_hyper_1_12_chunk, _dist_hyper_1_14_chunk, _dist_hyper_1_16_chunk, _dist_hyper_1_18_chunk, _dist_hyper_1_20_chunk, _dist_hyper_1_21_chunk, _dist_hyper_1_23_chunk, _dist_hyper_1_25_chunk
                     Remote SQL: SELECT _timescaledb_internal.partialize_agg(count(*)) FROM public.dist WHERE _timescaledb_internal.chunks_in(public.dist.*, ARRAY[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13])
               ->  Custom Scan (DataNodeScan)  (cost=78743.19..485465.97 rows=1 width=8)
                     Output: (PARTIAL count(*))
                     Relations: Aggregate on (public.dist)
                     Data node: dn2
                     Chunks: _dist_hyper_1_1_chunk, _dist_hyper_1_4_chunk, _dist_hyper_1_5_chunk, _dist_hyper_1_8_chunk, _dist_hyper_1_9_chunk, _dist_hyper_1_11_chunk, _dist_hyper_1_13_chunk, _dist_hyper_1_15_chunk, _dist_hyper_1_17_chunk, _dist_hyper_1_19_chunk, _dist_hyper_1_22_chunk, _dist_hyper_1_24_chunk, _dist_hyper_1_26_chunk
                     Remote SQL: SELECT _timescaledb_internal.partialize_agg(count(*)) FROM public.dist WHERE _timescaledb_internal.chunks_in(public.dist.*, ARRAY[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13])
(17 rows)
```
**Expected behavior**
Count should return the amount of rows in hypertable, i.e., `120961` row.

**Actual behavior**
Count returns incorrect amount of rows `8700`.

**Additional context**
Disabling partitionwise aggregate allows to get correct result. After setting `enable_partitionwise_aggregate = off` in `postgresql.conf` and restarting the instance:
```sql
select count(*) from dist;
 count
--------
 120961
(1 row)

explain select count(*) from dist;
                                                  QUERY PLAN
--------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=2543795.63..2543795.63 rows=1 width=8)
   ->  Custom Scan (AsyncAppend)  (cost=100.00..2386509.25 rows=62914550 width=0)
         ->  Append  (cost=100.00..2386509.25 rows=62914550 width=0)
               ->  Custom Scan (DataNodeScan) on dist dist_1  (cost=100.00..1035968.25 rows=31457275 width=0)
               ->  Custom Scan (DataNodeScan) on dist dist_2  (cost=100.00..1035968.25 rows=31457275 width=0)
(5 rows)
```

### LINKED ISSUE #2858: Number of regression tests crash when partitionwise aggregate is enabled
 - OS: Mac
 - TimescaleDB version (output of `\dx` in `psql`): 2.1.0-dev
 - Installation method: source

**Describe the bug**
`continuous_aggs` test crashes if partitionwise aggregate is enabled in the postgres config file with core dump:
```
* thread #1, stop reason = signal SIGSTOP
  * frame #0: 0x00007fff20312462 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff20340610 libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007fff20293720 libsystem_c.dylib`abort + 120
    frame #3: 0x000000010a491690 postgres`ExceptionalCondition(conditionName="!(aggref->aggsplit == aggstate->aggsplit)", errorType="FailedAssertion", fileName="nodeAgg.c", lineNumber=2546) at assert.c:54:2
    frame #4: 0x000000010a0bae0a postgres`ExecInitAgg(node=0x00007fe58f80de58, estate=0x00007fe58f83d918, eflags=0) at nodeAgg.c:2546:3
    frame #5: 0x000000010a0ac7aa postgres`ExecInitNode(node=0x00007fe58f80de58, estate=0x00007fe58f83d918, eflags=0) at execProcnode.c:323:27
    frame #6: 0x000000010a0b8e48 postgres`ExecInitAppend(node=0x00007fe58f8bba40, estate=0x00007fe58f83d918, eflags=0) at nodeAppend.c:231:28
    frame #7: 0x000000010a0ac4b6 postgres`ExecInitNode(node=0x00007fe58f8bba40, estate=0x00007fe58f83d918, eflags=0) at execProcnode.c:179:27
    frame #8: 0x0000000113f6cd4d timescaledb-2.1.0-dev.so`chunk_dispatch_begin(node=0x00007fe58f83e508, estate=0x00007fe58f83d918, eflags=0) at chunk_dispatch_state.c:36:7
    frame #9: 0x000000010a0c34d8 postgres`ExecInitCustomScan(cscan=0x00007fe58f8bdd28, estate=0x00007fe58f83d918, eflags=0) at nodeCustom.c:101:2
    frame #10: 0x000000010a0ac6e6 postgres`ExecInitNode(node=0x00007fe58f8bdd28, estate=0x00007fe58f83d918, eflags=0) at execProcnode.c:282:27
    frame #11: 0x000000010a0d9581 postgres`ExecInitModifyTable(node=0x00007fe58cdf1748, estate=0x00007fe58f83d918, eflags=0) at nodeModifyTable.c:2304:26
    frame #12: 0x000000010a0ac49a postgres`ExecInitNode(node=0x00007fe58cdf1748, estate=0x00007fe58f83d918, eflags=0) at execProcnode.c:174:27
    frame #13: 0x0000000113f8c78a timescaledb-2.1.0-dev.so`hypertable_insert_begin(node=0x00007fe58f83de20, estate=0x00007fe58f83d918, eflags=0) at hypertable_insert.c:85:7
    frame #14: 0x000000010a0c34d8 postgres`ExecInitCustomScan(cscan=0x00007fe58cdf1920, estate=0x00007fe58f83d918, eflags=0) at nodeCustom.c:101:2
    frame #15: 0x000000010a0ac6e6 postgres`ExecInitNode(node=0x00007fe58cdf1920, estate=0x00007fe58f83d918, eflags=0) at execProcnode.c:282:27
    frame #16: 0x000000010a0a12c3 postgres`InitPlan(queryDesc=0x00007fe58f010f18, eflags=0) at execMain.c:1046:14
    frame #17: 0x000000010a0a09ad postgres`standard_ExecutorStart(queryDesc=0x00007fe58f010f18, eflags=0) at execMain.c:265:2
    frame #18: 0x000000010a0a06ee postgres`ExecutorStart(queryDesc=0x00007fe58f010f18, eflags=0) at execMain.c:147:3
    frame #19: 0x000000010a2de9e4 postgres`ProcessQuery(plan=0x00007fe58cdf3c48, sourceText="insert into \"_timescaledb_internal\".\"_materialized_hypertable_2\"\nselect a, _timescaledb_internal.partialize_agg(count(b)),\ntime_bucket(1, a)\n,1\nfrom foo\ngroup by time_bucket(1, a) , a ;", params=0x0000000000000000, queryEnv=0x0000000000000000, dest=0x00007fe58cdf3d60, completionTag="") at pquery.c:156:2
    frame #20: 0x000000010a2dda6e postgres`PortalRunMulti(portal=0x00007fe58e831318, isTopLevel=true, setHoldSnapshot=false, dest=0x00007fe58cdf3d60, altdest=0x00007fe58cdf3d60, completionTag="") at pquery.c:1286:5
    frame #21: 0x000000010a2dd13d postgres`PortalRun(portal=0x00007fe58e831318, count=9223372036854775807, isTopLevel=true, run_once=true, dest=0x00007fe58cdf3d60, altdest=0x00007fe58cdf3d60, completionTag="") at pquery.c:799:5
    frame #22: 0x000000010a2d863e postgres`exec_simple_query(query_string="insert into \"_timescaledb_internal\".\"_materialized_hypertable_2\"\nselect a, _timescaledb_internal.partialize_agg(count(b)),\ntime_bucket(1, a)\n,1\nfrom foo\ngroup by time_bucket(1, a) , a ;") at postgres.c:1145:10
    frame #23: 0x000000010a2d7802 postgres`PostgresMain(argc=1, argv=0x00007fe58d025940, dbname="db_continuous_aggs", username="super_user") at postgres.c:4193:7
    frame #24: 0x000000010a2155c6 postgres`BackendRun(port=0x00007fe58e005560) at postmaster.c:4364:2
    frame #25: 0x000000010a214965 postgres`BackendStartup(port=0x00007fe58e005560) at postmaster.c:4036:3
    frame #26: 0x000000010a2138b6 postgres`ServerLoop at postmaster.c:1707:7
    frame #27: 0x000000010a21124c postgres`PostmasterMain(argc=8, argv=0x00007fe58cc06be0) at postmaster.c:1380:11
    frame #28: 0x000000010a11c499 postgres`main(argc=8, argv=0x00007fe58cc06be0) at main.c:228:3
    frame #29: 0x00007fff2035b631 libdyld.dylib`start + 1
    frame #30: 0x00007fff2035b631 libdyld.dylib`start + 1
```

**To Reproduce**
1. Set `enable_partitionwise_aggregate = on` in `postgresql.conf`
2. Run `make installcheck TESTS='continuous_aggs'`
3. The test reports
```
+SSL SYSCALL error: EOF detected
+connection to server was lost
```
Or just connect to a database with TimescaleDB extension and run using psql:
```sql
create table foo (a integer, b integer, c integer);
select table_name from create_hypertable('foo', 'a', chunk_time_interval=> 10);
insert into foo values( 3 , 16 , 20);
insert into foo values( 1 , 10 , 20);
insert into foo values( 1 , 11 , 20);
insert into foo values( 1 , 12 , 20);
insert into foo values( 1 , 13 , 20);
insert into foo values( 1 , 14 , 20);
insert into foo values( 2 , 14 , 20);
insert into foo values( 2 , 15 , 20);
insert into foo values( 2 , 16 , 20);

select a, _timescaledb_internal.partialize_agg(count(b)),
time_bucket(1, a)
,1
from foo
group by time_bucket(1, a) , a ;
```

**Expected behavior**
No crash should happen and all query execution should return correct results (explain will differ).

**Actual behavior**
Crash happens

k-rus:
partialize_agg is an internal function, which serializes partial
aggregate results. It is used to prepare partials for materialization
in continuous aggregates and partial results on data nodes in
distributed query execution. paritalize_agg doesn't expect push down of
aggregates, which happens when partitionwise aggregate is enabled, and
produces a query plan, which either crashes on assert during execution
or produces incorrect result.

This fix avoids adding partition info if the function is present in the
query. This can be seen as a work around and it is good to fix planning
of partialize_agg in the case of pushed down aggregates.

This commit also contains few minor fixes of readability of comments
and code around the changes.

Fixes #2849 and fixes #2858

### DIFF
--- a/src/plan_partialize.c
+++ b/src/plan_partialize.c
@@ -35,6 +35,7 @@ typedef struct PartializeWalkerState
 	bool found_non_partial_agg;
 	bool looking_for_agg;
 	Oid fnoid;
+	PartializeAggFixAggref fix_aggref;
 } PartializeWalkerState;
 
 /*
@@ -70,13 +71,20 @@ check_for_partialize_function_call(Node *node, PartializeWalkerState *state)
 		if (state->looking_for_agg)
 		{
 			state->looking_for_agg = false;
-			aggref->aggsplit = AGGSPLIT_INITIAL_SERIAL;
+			if (state->fix_aggref == TS_FIX_AGGREF)
+			{
+				aggref->aggsplit = AGGSPLIT_INITIAL_SERIAL;
 
-			if (aggref->aggtranstype == INTERNALOID &&
-				DO_AGGSPLIT_SERIALIZE(AGGSPLIT_INITIAL_SERIAL))
-				aggref->aggtype = BYTEAOID;
-			else
-				aggref->aggtype = aggref->aggtranstype;
+				if (aggref->aggtranstype == INTERNALOID &&
+					DO_AGGSPLIT_SERIALIZE(AGGSPLIT_INITIAL_SERIAL))
+				{
+					aggref->aggtype = BYTEAOID;
+				}
+				else
+				{
+					aggref->aggtype = aggref->aggtranstype;
+				}
+			}
 		}
 
 		/* We currently cannot handle cases like
@@ -97,8 +105,8 @@ check_for_partialize_function_call(Node *node, PartializeWalkerState *state)
 	return expression_tree_walker(node, check_for_partialize_function_call, state);
 }
 
-static bool
-has_partialize_function(Query *parse)
+bool
+has_partialize_function(Query *parse, PartializeAggFixAggref fix_aggref)
 {
 	Oid partialfnoid = InvalidOid;
 	Oid argtyp[] = { ANYELEMENTOID };
@@ -106,6 +114,7 @@ has_partialize_function(Query *parse)
 	PartializeWalkerState state = { .found_partialize = false,
 									.found_non_partial_agg = false,
 									.looking_for_agg = false,
+									.fix_aggref = fix_aggref,
 									.fnoid = InvalidOid };
 	List *name = list_make2(makeString(INTERNAL_SCHEMA_NAME), makeString(TS_PARTIALFN));
 
@@ -145,7 +154,7 @@ partialize_agg_paths(RelOptInfo *rel)
  * instance:
  *
  *  SELECT time_bucket('1 day', time), device,
- *  __timescaledb_internal.partialize_agg(avg(temp))
+ *  _timescaledb_internal.partialize_agg(avg(temp))
  *  GROUP BY 1, 2;
  *
  * Would compute the partial aggregate of avg(temp).
@@ -174,9 +183,10 @@ partialize_agg_paths(RelOptInfo *rel)
  * simple filter (e.g., HAVING device > 3). In such cases, the HAVING clause is
  * removed and replaced by a filter on the input.
  * Returns : true if partial aggs were found, false otherwise.
+ * Modifies : output_rel if partials aggs were found.
  */
 bool
-ts_plan_process_partialize_agg(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo *output_rel)
+ts_plan_process_partialize_agg(PlannerInfo *root, RelOptInfo *output_rel)
 {
 	Query *parse = root->parse;
 	Assert(IS_UPPER_REL(output_rel));
@@ -184,7 +194,7 @@ ts_plan_process_partialize_agg(PlannerInfo *root, RelOptInfo *input_rel, RelOptI
 	if (CMD_SELECT != parse->commandType || !parse->hasAggs)
 		return false;
 
-	if (!has_partialize_function(parse))
+	if (!has_partialize_function(parse, TS_FIX_AGGREF))
 		return false;
 
 	/* We cannot check root->hasHavingqual here because sometimes the
--- a/tsl/test/expected/partialize_finalize.out
+++ b/tsl/test/expected/partialize_finalize.out
@@ -2,6 +2,13 @@
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-TIMESCALE for a copy of the license.
 -- TEST1 count with integers
+SHOW enable_partitionwise_aggregate;
+ enable_partitionwise_aggregate 
+--------------------------------
+ off
+(1 row)
+
+SET enable_partitionwise_aggregate = on;
 create table foo (a integer, b integer, c integer);
 insert into foo values( 1 , 10 , 20);
 insert into foo values( 1 , 11 , 20);
@@ -76,6 +83,7 @@ select a, _timescaledb_internal.finalize_agg( 'sum(numeric)', null, null, null, 
  5 |   80 |    0
 (4 rows)
 
+SET enable_partitionwise_aggregate = off;
 --TEST3 sum with expressions
 drop table t1;
 drop view v1;
--- a/tsl/test/sql/continuous_aggs.sql
+++ b/tsl/test/sql/continuous_aggs.sql
@@ -82,6 +82,9 @@ SET ROLE :ROLE_DEFAULT_PERM_USER;
 -- TEST2 ---
 DROP MATERIALIZED VIEW mat_m1;
 
+SHOW enable_partitionwise_aggregate;
+SET enable_partitionwise_aggregate = on;
+
 SELECT * FROM _timescaledb_config.bgw_job;
 
 CREATE TABLE conditions (
@@ -141,6 +144,8 @@ select time_bucket('1day', timec), min(location), sum(temperature), sum(humidity
 from conditions
 group by time_bucket('1day', timec)
 order by 1;
+
+SET enable_partitionwise_aggregate = off;
 
 -- TEST3 --
 -- drop on table conditions should cascade to materialized mat_v1
--- a/tsl/test/sql/partialize_finalize.sql
+++ b/tsl/test/sql/partialize_finalize.sql
@@ -3,6 +3,9 @@
 -- LICENSE-TIMESCALE for a copy of the license.
 
 -- TEST1 count with integers
+SHOW enable_partitionwise_aggregate;
+SET enable_partitionwise_aggregate = on;
+
 create table foo (a integer, b integer, c integer);
 insert into foo values( 1 , 10 , 20);
 insert into foo values( 1 , 11 , 20);
@@ -60,6 +63,8 @@ insert into foo values( 5, 40, 0);
 --sum aggfnoid 2114, min aggfnoid is 2136 oid  numeric is 1700
 insert into t1 select * from v1 where ( a = 3 ) or a = 5;
 select a, _timescaledb_internal.finalize_agg( 'sum(numeric)', null, null, null, partialb, cast('1' as numeric) ) sumb, _timescaledb_internal.finalize_agg( 'min(double precision)', null, null, null, partialminc, cast('1' as float8) ) minc from t1 group by a order by a ;
+
+SET enable_partitionwise_aggregate = off;
 
 --TEST3 sum with expressions
 drop table t1;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,11 @@ accidentally triggering the load of a previous DB version.**
 ## Latest
 
 **Bugfixes**
-* #2842 Do not mark job as started when seting next_start field
+* #2842 Do not mark job as started when setting next_start field
 * #2845 Fix continuous aggregate privileges during upgrade
 * #2851 Fix nested loop joins that involve compressed chunks
 * #2865 Apply volatile function quals at decompresschunk node
+* #2866 Avoid partitionwise planning of partialize_agg
 * #2868 Fix corruption in gapfill plan
 
 **Minor features**
--- a/src/plan_expand_hypertable.c
+++ b/src/plan_expand_hypertable.c
@@ -33,19 +33,20 @@
 #include <optimizer/optimizer.h>
 #endif
 
-#include "import/planner.h"
-#include "plan_expand_hypertable.h"
-#include "hypertable.h"
-#include "hypertable_restrict_info.h"
-#include "planner.h"
-#include "chunk_append/chunk_append.h"
+#include "chunk.h"
+#include "cross_module_fn.h"
 #include "guc.h"
 #include "extension.h"
-#include "chunk.h"
 #include "extension_constants.h"
+#include "hypertable.h"
+#include "hypertable_restrict_info.h"
 #include "partitioning.h"
-#include "cross_module_fn.h"
+#include "plan_expand_hypertable.h"
+#include "plan_partialize.h"
+#include "planner.h"
 #include "time_utils.h"
+#include "chunk_append/chunk_append.h"
+#include "import/planner.h"
 
 typedef struct CollectQualCtx
 {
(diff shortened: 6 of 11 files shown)


src/plan_partialize.c
```diff
@@ -174,17 +183,18 @@ partialize_agg_paths(RelOptInfo *rel)
...
+ts_plan_process_partialize_agg(PlannerInfo *root, RelOptInfo *output_rel)
 {
 	Query *parse = root->parse;
 	Assert(IS_UPPER_REL(output_rel));
 
 	if (CMD_SELECT != parse->commandType || !parse->hasAggs)
 		return false;
 
-	if (!has_partialize_function(parse))
+	if (!has_partialize_function(parse, TS_FIX_AGGREF))
```
erimatnor, line 197: Shouldn't we check `rel->relatarget->expr` here instead of parse tree? Otherwise it will partialize for any rel.

k-rus: It is expected to partialize for any rel. I believe it can be changes to do it on `reltarget`, but imho this refactoring is not necessary for this PR.

src/plan_partialize.c
```diff
@@ -70,13 +71,20 @@ check_for_partialize_function_call(Node *node, PartializeWalkerState *state)
...
 		{
 			state->looking_for_agg = false;
-			aggref->aggsplit = AGGSPLIT_INITIAL_SERIAL;
-
-			if (aggref->aggtranstype == INTERNALOID &&
-				DO_AGGSPLIT_SERIALIZE(AGGSPLIT_INITIAL_SERIAL))
-				aggref->aggtype = BYTEAOID;
-			else
-				aggref->aggtype = aggref->aggtranstype;
+			if (state->fix_aggref == TS_FIX_AGGREF)
```
pmwkaa, line 74: I suggest to write a comment here, why it is needed and what it actually do

k-rus: @pmwkaa It was originally there, I just re-use function without doing modification. I am guessing what it does, but I am not so fluent in the PG planning to be sure. So I am not sure I should comment here.

src/plan_partialize.c
erimatnor: Does this check avoid doing things for, e.g., declarative partitioning? It seems to me that we will partialize any rel, including child partitions during partitionwise aggregation for declarative partitions, so I am wondering if this bug will still be hit for other tables than hypertables?

erimatnor (approved):
Approving, since it was agreed to leave fixes for other table types to later.

src/plan_partialize.c
k-rus: I believe it will be hit on declarative partitioning. I created a separate internal issue to address it later.

erimatnor: FWIW, it might be easy to fix by checking the extras parameter in the upper_rel hook? It contains a `GroupPathExtraData` struct that contains info on whether this is called for a partitionwise aggregation. We can just exclude calling this function in that case.

k-rus: Removing function from the target list is not an easy fix if I remember correctly.

erimatnor: Is it necessary to remove it from the target list? Isn't it just a noop? Can we just clear the partitionwise paths so that we don't produce those plans in that case since they don't make sense anyway?

It is fine to leave for another PR I guess.

tsl/test/sql/continuous_aggs.sql
```diff
@@ -82,6 +82,9 @@ SET ROLE :ROLE_DEFAULT_PERM_USER;
 -- TEST2 ---
 DROP MATERIALIZED VIEW mat_m1;
 
+SHOW enable_partitionwise_aggregate;
```
gayyappan, line 85: This is not the best test case for this change as the mat view does not have an aggregate with partials in internal form.  avg with numeric or bigint type will have partials in internal form. So could you  add a test case specifically for this: ( I haven't been able to locate one yet)? It is similar to the test case at line 408 but the avg should be on a numeric type instead. (The INSERT stmt uses the partialize_agg stmt).

gayyappan (changes requested):

tsl/test/sql/continuous_aggs.sql
gayyappan: I see that sum(numeric) also has internal aggtranstype.

tsl/test/expected/partialize_finalize.out
```diff
@@ -2,6 +2,13 @@
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-TIMESCALE for a copy of the license.
 -- TEST1 count with integers
+SHOW enable_partitionwise_aggregate;
+ enable_partitionwise_aggregate 
+--------------------------------
+ off
+(1 row)
+
+SET enable_partitionwise_aggregate = on;
```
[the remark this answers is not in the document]
k-rus: @gayyappan The test in this file was crashing. It is repro of #2858. 

The fix doesn't affect aggregate functions, so it doesn't matter which to test. My suggestion is to update several test files to run with partitionwise enabled and disabled, however it is for different PR.

tsl/test/sql/continuous_aggs.sql
k-rus: The fix doesn't affect aggregate functions. Another approach, which I initial implemented, is to enable paritionwise aggregate in the PostgreSQL config. However, I don't think it will be appropriate. My recommendation is to do proper testing with comparison between disabled and enabled paritionwise in a separate PR.

tsl/test/expected/partialize_finalize.out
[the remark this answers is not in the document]
gayyappan: I am not that I follow  what you mean  : ```The fix doesn't affect aggregate functions ```.  This test perhaps caused the crash. But you still want to test that partialize_agg is doing the correct thing when enable_partitionwise_aggregate is  true.  A cont. agg view does all these steps : a) insert with partials and b) select using the finalize_agg. This way , you can verify that  partials is working correctly.

tsl/test/sql/partialize_finalize.sql
```diff
@@ -28,6 +31,8 @@ insert into t1 select * from t1;
 select a, _timescaledb_internal.finalize_agg( 'count("any")', null, null, null, partial, cast('1' as int8) ) from t1
 group by a order by a ;
 
+SET enable_partitionwise_aggregate = off;
```
gayyappan, line 34: Moving this to line 68 will cover the NUMERIC case with internal type as well.

tsl/test/sql/continuous_aggs.sql
gayyappan: Yes, this would be ideal.

tsl/test/sql/partialize_finalize.sql
k-rus: Moved.

gayyappan (approved):
Changes look good.  Missing partialize_finalize.out push after the set change.

PR #2868, opened 2021-01-27
Fix corruption in gapfill plan

### LINKED ISSUE #2232: Could not find pathkey item to sort
 - OS: Ubuntu 18.04 LTS
 - PostgreSQL 12.2
 - TimescaleDB 1.7.0

**Describe the bug**
When querying a hypertable and `GROUP BY` the hypertable's time column, the width of the time bucket determines whether the postgresql planner returns the error `Could not find pathkey item to sort`.

**To Reproduce**
It is not known if this is reproducible on other environments.

Example failure, note the time_bucket value:
```
postgres=# explain select
 time_bucket('83022ms',time),
 avg(seconds)
from metrics
where "time" > '2020-08-19T17:34:37.496Z' AND time < '2020-08-19T23:34:37.497Z'
  and cell_name = 'a particular cell'
group by 1;
ERROR:  could not find pathkey item to sort
```

Example success, note the time bucket value:
```
postgres=# explain select
 time_bucket('83021ms',time),
 avg(seconds)
from metrics
where "time" > '2020-08-19T17:34:37.496Z' AND time < '2020-08-19T23:34:37.497Z'
  and cell_name = 'a particular cell'
group by 1;
                                                                                               QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=127900.99..134242.67 rows=43909 width=16)
   Group Key: (time_bucket('00:01:23.021'::interval, _hyper_20_1182123_chunk."time"))
   ->  Gather Merge  (cost=127900.99..133364.48 rows=43910 width=40)
         Workers Planned: 5
         ->  Partial GroupAggregate  (cost=126900.91..127076.55 rows=8782 width=40)
               Group Key: (time_bucket('00:01:23.021'::interval, _hyper_20_1182123_chunk."time"))
               ->  Sort  (cost=126900.91..126922.86 rows=8782 width=16)
                     Sort Key: (time_bucket('00:01:23.021'::interval, _hyper_20_1182123_chunk."time"))
                     ->  Result  (cost=0.43..126325.67 rows=8782 width=16)
                           ->  Parallel Index Scan Backward using _hyper_20_1182123_chunk_this_index_name_is_long_enough_to_trunca on _hyper_20_1182123_chunk  (cost=0.43..126215.90 rows=8782 width=16)
                                 Index Cond: (("time" > '2020-08-19 10:34:37.496-07'::timestamp with time zone) AND ("time" < '2020-08-19 16:34:37.497-07'::timestamp with time zone))
                                 Filter: (cell_name = 'some particular cell'::text)
 JIT:
   Functions: 15
   Options: Inlining false, Optimization false, Expressions true, Deforming true
(15 rows)
```

**Expected behavior**
For any valid width of time_bucket, the planner can produce a query plan.  Alternatively, date_trunc can work with _both_ `minute` and `hour`.

**Actual behavior**
Any grouping of the hypertable's time column that is larger than 83021 milliseconds causes the query planner to error.

**Additional context**
Certain selections of the table for this time range do this while others do not.
* The hypertable is simple; there is no physical partitioning, continuous aggregate or compression.
* There are no additional indices beyond the one Timescaledb affixes to the time column when you create_hypertable.
* All columns are either `text` or `double precision`.
* There are 43 chunks, each 1 day wide with approximately 7 million rows each.
* There are no null values in the time column (per the hypertable constraint) and there are no null values in the aggregated numeric column.
* The query matches 46128 rows before grouping, no null values in either column.

erimatnor:
This change fixes a bug with gapfill that caused certain query plans
to fail with the error "could not find pathkey item to sort". This was
caused by a corruption in the query plan which happened as a result of
removing the last two arguments to the `time_bucket_gapfill` function
expression during query planning. Since the function expression was
modified via a reference to the original query plan tree, it affected
also the expression in the target list. When the planner couldn't
match the target list with the corresponding equivalence member (which
still included the two removed arguments), the error was generated.

The original reason for removing the two arguments was to avoid
passing them on to `time_bucket`, which is called internally by
`time_bucket_gapfill`. However, the last to arguments aren't passed on
anyway, so it isn't necessary to modify the original argument list.

Fixes #2232

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,15 @@ accidentally triggering the load of a previous DB version.**
 * #2845 Fix continuous aggregate privileges during upgrade
 * #2851 Fix nested loop joins that involve compressed chunks
 * #2865 Apply volatile function quals at decompresschunk node
+* #2868 Fix corruption in gapfill plan
 
 **Minor features**
 * #2736 Support adding columns to hypertables with compression enabled
+
+**Thanks**
+* @WarriorOfWire for reporting the bug with gapfill queries not being
+  able to find pathkey item to sort
+
 
 ## 2.0.0 (2020-12-18)
 
--- a/tsl/src/nodes/gapfill/planner.c
+++ b/tsl/src/nodes/gapfill/planner.c
@@ -196,9 +196,6 @@ gapfill_plan_create(PlannerInfo *root, RelOptInfo *rel, CustomPath *path, List *
 	cscan->custom_private =
 		list_make4(gfpath->func, root->parse->groupClause, root->parse->jointree, args);
 
-	/* remove start and end argument from time_bucket call */
-	gfpath->func->args = list_make2(linitial(gfpath->func->args), lsecond(gfpath->func->args));
-
 	return &cscan->scan.plan;
 }
 
--- a/tsl/test/shared/expected/gapfill.out
+++ b/tsl/test/shared/expected/gapfill.out
@@ -10,13 +10,13 @@ SELECT
 FROM (VALUES (now(),1),(now(),NULL),(now(),NULL)) as t(time,c2)
 GROUP BY 1
 ORDER BY 1;
-                                       QUERY PLAN                                        
------------------------------------------------------------------------------------------
+                                              QUERY PLAN                                               
+-------------------------------------------------------------------------------------------------------
  Custom Scan (GapFill)
    ->  GroupAggregate
-         Group Key: (time_bucket_gapfill('@ 5 mins'::interval, "*VALUES*".column1))
+         Group Key: (time_bucket_gapfill('@ 5 mins'::interval, "*VALUES*".column1, now(), now()))
          ->  Sort
-               Sort Key: (time_bucket_gapfill('@ 5 mins'::interval, "*VALUES*".column1))
+               Sort Key: (time_bucket_gapfill('@ 5 mins'::interval, "*VALUES*".column1, now(), now()))
                ->  Values Scan on "*VALUES*"
 (6 rows)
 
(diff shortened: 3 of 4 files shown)


k-rus (approved):
It looks good to me, but I cannot verify the example.
Can you run a query showing the actual data in the range (or absence of the data)?
For example:
```sql
SELECT time
FROM metrics_tstz
WHERE time >= '2017-01-01' AND time < '2018-01-01'
```

erimatnor:
> It looks good to me, but I cannot verify the example.
> Can you run a query showing the actual data in the range (or absence of the data)?
> For example:
> 
> ```sql
> SELECT time
> FROM metrics_tstz
> WHERE time >= '2017-01-01' AND time < '2018-01-01'
> ```

I'll add that, but the fix is for an error that occurred before, not correctness, which should also be covered by existing queries/tests in my understanding.

k-rus:
> Added improved queries, although the fix is for an error that occurred before, not correctness (which should also be covered by existing queries/tests in my understanding). Also, shouldn't the comparison query be time_bucket without gapfill? (Not sure what the value is of showing the raw, non-grouped data, for this bug fix)

@erimatnor I think you missed my point. If I see the data in table for the same range as the test query, I can judge if the test query result is reasonable or not. Thus I suggested to run the query:
```sql
SELECT time
FROM metrics_tstz
WHERE time >= '2017-01-01' AND time < '2018-01-01'
```

Since the query was failing before the fix, the code path wasn't tested and thus it is not strange to pay attention to the correctness of the test result.

svenklemm (approved):
Having those extra arguments in EXPLAIN looks a bit weird

erimatnor:
> Having those extra arguments in EXPLAIN looks a bit weird

IMO, it was more weird before when the arguments were given in the query but disappeared in the plan.

PR #2909, opened 2021-02-05
Support column renames on tables that have compression enabled.

gayyappan:
ALTER TABLE <hypertable> RENAME <column_name> TO <new_column_name>
is now supported for hypertables that have compression enabled.

To reviewers:
This PR has multiple commits to help with reviewing test changes. These will be merged.

### DIFF
--- a/src/cross_module_fn.h
+++ b/src/cross_module_fn.h
@@ -99,6 +99,7 @@ typedef struct CrossModuleFunctions
 	bool (*process_compress_table)(AlterTableCmd *cmd, Hypertable *ht,
 								   WithClauseResult *with_clause_options);
 	void (*process_altertable_cmd)(Hypertable *ht, const AlterTableCmd *cmd);
+	void (*process_rename_cmd)(Hypertable *ht, const RenameStmt *stmt);
 	PGFunction compress_chunk;
 	PGFunction decompress_chunk;
 	/* The compression functions below are not installed in SQL as part of create extension;
--- a/src/hypertable_compression.h
+++ b/src/hypertable_compression.h
@@ -17,5 +17,7 @@ ts_hypertable_compression_fill_tuple_values(FormData_hypertable_compression *fd,
 											bool *nulls);
 
 extern TSDLLEXPORT bool ts_hypertable_compression_delete_by_hypertable_id(int32 htid);
+extern TSDLLEXPORT void ts_hypertable_compression_rename_column(int32 htid, char *old_column_name,
+																char *new_column_name);
 
 #endif
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -1571,10 +1571,10 @@ process_rename_column(ProcessUtilityArgs *args, Cache *hcache, Oid relid, Rename
 
 	dim = ts_hyperspace_get_dimension_by_name(ht->space, DIMENSION_TYPE_ANY, stmt->subname);
 
-	if (NULL == dim)
-		return;
-
-	ts_dimension_set_name(dim, stmt->newname);
+	if (dim)
+		ts_dimension_set_name(dim, stmt->newname);
+	if (ts_cm_functions->process_rename_cmd)
+		ts_cm_functions->process_rename_cmd(ht, stmt);
 }
 
 static void
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -1134,6 +1134,8 @@ create_per_compressed_column(TupleDesc in_desc, TupleDesc out_desc, Oid out_reli
 		/* find the mapping from compressed column to uncompressed column, setting
 		 * the index of columns that don't have an uncompressed version
 		 * (such as metadata) to -1
+		 * Assumption: column names are the same on compressed and
+		 *       uncompressed chunk.
 		 */
 		AttrNumber decompressed_colnum = get_attnum(out_relid, col_name);
 		if (!AttributeNumberIsValid(decompressed_colnum))
--- a/tsl/src/compression/create.c
+++ b/tsl/src/compression/create.c
@@ -16,6 +16,7 @@
 #include <catalog/pg_constraint.h>
 #include <catalog/pg_type.h>
 #include <catalog/toasting.h>
+#include <commands/alter.h>
 #include <commands/defrem.h>
 #include <commands/tablecmds.h>
 #include <commands/tablespace.h>
@@ -950,6 +951,12 @@ add_column_to_compression_table(Hypertable *compress_ht, CompressColInfo *compre
  * Note: caller should check security permissions
  *
  * Return true if compression was enabled, false otherwise.
+ *
+ * Steps:
+ * 1. Check existing constraints on the table -> can we support them with compression?
+ * 2. Create internal compression table + mark hypertable as compression enabled
+ * 3. Add catalog entries to hypertable_compression to record compression settings.
+ * 4. Copy constraints to internal compression table
  */
 bool
 tsl_process_compress_table(AlterTableCmd *cmd, Hypertable *ht,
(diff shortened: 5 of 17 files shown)


tsl/test/sql/include/compression_alter.sql
```diff
@@ -67,21 +69,61 @@ SELECT count(*) from test1 where new_colv  = '101t';
...
+SELECT * from test1 WHERE bigintcol = 100;
+SELECT * from test1 WHERE bigintcol = 200;
 
+-- add a new chunk and compress
+INSERT INTO test1 SELECT '2019-03-04 2:00', 99, 800, 'newchunk' ; 
 
+SELECT COUNT(*) AS count_compressed
+FROM
+(
+SELECT compress_chunk(chunk.schema_name|| '.' || chunk.table_name)
```
mkindahl, line 100: Any reason to not use `format`? It is used elsewhere.

tsl/test/sql/include/compression_alter.sql
```diff
@@ -67,21 +69,61 @@ SELECT count(*) from test1 where new_colv  = '101t';
...
+FROM _timescaledb_catalog.chunk chunk
+INNER JOIN _timescaledb_catalog.hypertable hypertable ON (chunk.hypertable_id = hypertable.id)
+WHERE hypertable.table_name like 'test1' and chunk.compressed_chunk_id IS NULL ORDER BY chunk.id
```
mkindahl, line 103:
```suggestion
WHERE hypertable.table_name = 'test1' and chunk.compressed_chunk_id IS NULL ORDER BY chunk.id
```

tsl/test/sql/include/compression_alter.sql
```diff
@@ -67,21 +69,61 @@ SELECT count(*) from test1 where new_colv  = '101t';
...
+--the correct column name 
+SELECT format('%I.%I', cht.schema_name, cht.table_name) AS "COMPRESSION_TBLNM"
+FROM _timescaledb_catalog.hypertable ht, _timescaledb_catalog.hypertable cht
+WHERE ht.table_name = 'test1' and cht.id = ht.compressed_hypertable_id \gset
+
+SELECT count(*) 
+FROM ( SELECT attrelid::regclass, attname FROM pg_attribute 
+       WHERE attrelid in (SELECT inhrelid::regclass from pg_inherits 
+                          where inhparent = :'COMPRESSION_TBLNM'::regclass )
+       and attname = 'bigintcol' ) q;
```
mkindahl, line 128: Nit: you're only using the variable here so suggest to use a CTE: it's a lot easier to read.

mkindahl (commented):
Consider if you should move all the code to the `tsl/` directory.

svenklemm (commented):
Can you add some tests where identifiers are getting truncated.

ALTER TABLE t RENAME c TO cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc;

ALTER TABLE t RENAME cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
 TO accccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc;

tsl/src/compression/compression.c
```diff
@@ -1134,6 +1134,8 @@ create_per_compressed_column(TupleDesc in_desc, TupleDesc out_desc, Oid out_reli
 		/* find the mapping from compressed column to uncompressed column, setting
 		 * the index of columns that don't have an uncompressed version
 		 * (such as metadata) to -1
+		 * Assumption: column names are the same on compressed and
+		 *       uncompressed chunk.
```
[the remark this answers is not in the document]
gayyappan: This is a pre-existing assumption. This code does not change any pre-existing assumptions.

tsl/test/sql/compression_ddl.sql
```diff
@@ -446,5 +446,24 @@ WHERE hypertable.table_name like 'test1'  ORDER BY chunk.id ) as subq;
...
+
+ALTER TABLE i2844 SET (timescaledb.compress);
+
+SELECT compress_chunk(show_chunks) AS compressed_chunk FROM show_chunks('i2844');
+SELECT drop_chunks('i2844', older_than => '2000-01-01 18:00'::timestamptz);
+SELECT decompress_chunk(show_chunks, if_compressed => TRUE) AS decompressed_chunks FROM show_chunks('i2844');
+
+ALTER TABLE i2844 SET (timescaledb.compress = FALSE);
+
+-- compression alter schema tests
```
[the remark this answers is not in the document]
gayyappan: It isn't. The individual commits explain the changes to the test files.

tsl/src/compression/compression.c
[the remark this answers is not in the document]
mkindahl: What happens if this assumption is not correct, does the code crash or does it just give an error?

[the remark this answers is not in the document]
gayyappan: I added this comment to _clarify_ existing code and the assumptions made by it. This would crash if the rename were to go through without respecting the existing assumption.

src/hypertable_compression.h
```diff
@@ -17,5 +17,7 @@ ts_hypertable_compression_fill_tuple_values(FormData_hypertable_compression *fd,
 											bool *nulls);
 
 extern TSDLLEXPORT bool ts_hypertable_compression_delete_by_hypertable_id(int32 htid);
+extern TSDLLEXPORT void ts_hypertable_compression_rename_column(int32 htid, char *old_column_name,
+																char *new_column_name);
```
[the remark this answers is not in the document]
gayyappan: Catalog table related code is under src.

mkindahl (approved):

src/process_utility.c
```diff
@@ -1571,10 +1571,10 @@ process_rename_column(ProcessUtilityArgs *args, Cache *hcache, Oid relid, Rename
 
 	dim = ts_hyperspace_get_dimension_by_name(ht->space, DIMENSION_TYPE_ANY, stmt->subname);
 
-	if (NULL == dim)
-		return;
-
-	ts_dimension_set_name(dim, stmt->newname);
+	if (dim)
+		ts_dimension_set_name(dim, stmt->newname);
+	if (ts_cm_functions->process_rename_cmd)
```
pmwkaa, line 1576: Do we really this this new cross module function? We can use existing `tsl_ddl_command_start` hook instead

tsl/src/compression/create.c
```diff
@@ -1068,3 +1075,31 @@ tsl_process_compress_table_add_column(Hypertable *ht, ColumnDef *orig_def)
...
+	{
+		int32 compress_htid = ht->fd.compressed_hypertable_id;
+		Hypertable *compress_ht = ts_hypertable_get_by_id(compress_htid);
+		RenameStmt *compress_col_stmt = (RenameStmt *) copyObject(stmt);
+		compress_col_stmt->relation = makeRangeVar(NameStr(compress_ht->fd.schema_name),
+												   NameStr(compress_ht->fd.table_name),
+												   -1);
+		ExecRenameStmt(compress_col_stmt);
+	}
+	Assert(TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht));
```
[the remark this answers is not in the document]
gayyappan: done.

tsl/test/sql/dist_compression.sql
```diff
@@ -123,14 +123,14 @@ SELECT table_name, compression_state, compressed_hypertable_id
 FROM _timescaledb_catalog.hypertable
 ORDER BY 1;
 
-SELECT * FROM timescaledb_information.compression_settings;
+SELECT * FROM timescaledb_information.compression_settings order by attname;
```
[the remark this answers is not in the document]
gayyappan: If a segment by/order by  column is renamed, the new column name should be reflected in the view.

tsl/test/sql/include/compression_alter.sql
gayyappan: done.

src/process_utility.c
gayyappan: I tried to refactor the process_utility code to directly call tsl_ddl_command_start without any additional tsl support functions.  It doesn't work as expected as some ddl functions are executed in multi transaction contexts. For this particular ddl command, the original hypertable has already been modified and the hypertable cache cannot find it by the time we get here with the original name.

svenklemm (approved):

src/cross_module_fn.h
```diff
@@ -99,6 +99,7 @@ typedef struct CrossModuleFunctions
 	bool (*process_compress_table)(AlterTableCmd *cmd, Hypertable *ht,
 								   WithClauseResult *with_clause_options);
 	void (*process_altertable_cmd)(Hypertable *ht, const AlterTableCmd *cmd);
+	void (*process_rename_cmd)(Hypertable *ht, const RenameStmt *stmt);
```
erimatnor, line 102: I think this has come up before, and I don't know the conclusion so I'll reiterate.

I think creating a new cross-module function for every process utility (DDL) command is not a good idea and doesn't really scale. We already "forward" DDL commands to TSL and we should try to use that mechanism or improve it if it isn't sufficient in its current form.

The next best approach, IMO, is to have a "general" cross-module function that we can reuse for all DDL commands, e.g.:

```
void (*process_ddl_cmd)(Hypertable *ht, const Node *cmd);
```

then use a switch on it in the `process_ddl_cmd` function:

```

switch (nodeTag(cmd))
{
    case T_RenameStmt:
         ....
         break;
    case T_AlterTableCmd:
         ...
         break;
}
```

src/process_utility.c
erimatnor: I share @pmwkaa's concern (and I have given similar feedback in the past). If the existing DDL forwarding doesn't work for some reason (I am not sure I understand why not), then we should at least use just one cross-module function for all DDL statements instead of adding a new one for each DDL. The reason the the hypertable cannot be found in the cache is because `CommandCounterIncrement()`. We have catalog update functions to deal with that without doing command-counter increment.

gayyappan: @erimatnor This is the refactoring change we discussed. I spent a few days trying to refactor the code to get this to work: Setting the hypertable cache and passing it around didn't work as expected. I didn't want to spend more time fixing this at the cost of derailing this feature support. process_utility code needs a wider overhaul to work without adding new functions to the tsl code. We also have exactly the same issue calling out to sql functions that have implementations under tsl. This should also be addressed as part of the refactoring effort.

PR #2919, opened 2021-02-09
Add test to check PARALLEL VACUUM functionality with PG13

nikkhils:
PG13 has introduced the parallel vacuum functionality in which if the
table has two or more indexes (with sizes beyond 512KB and enough number
of parallel worker processes), then the vacuum on those indexes can be
done in parallel.


test/expected/vacuum_parallel-13.out
```diff
@@ -0,0 +1,52 @@
...
+CREATE INDEX vt_temp2 on vacuum_test (temp2);
+-- parallel vacuum needs the index size to be larger than 512KB to kick in
+INSERT INTO vacuum_test SELECT TIMESTAMP 'epoch' + (i * INTERVAL '100 second'),
+                i, i+1 FROM generate_series(1, 100000) as T(i);
+-- we should see two parallel workers for each chunk
+VACUUM (PARALLEL 3, VERBOSE) vacuum_test;
+INFO:  vacuuming "_timescaledb_internal._hyper_1_1_chunk"
+INFO:  launched 2 parallel vacuum workers for index cleanup (planned: 2)
+INFO:  index "_hyper_1_1_chunk_vacuum_test_time_idx" now contains 26279 row versions in 131 pages
+INFO:  index "_hyper_1_1_chunk_vt_temp1" now contains 26279 row versions in 74 pages
```
erimatnor, line 27: I am wondering if this output is fully predictable or if we need to mute it to ensure that the test won't fail under different circumstances (i.e., the  XX row versions in  YY pages would change depending on setting this is run under, leading to a test failure).

nikkhils: @erimatnor yeah, I thought of that. But I don't see a good way to mute the number of rows and pages info. Or is there something additional that I am not aware of that can help with this?

test/sql/.gitignore
```diff
@@ -28,3 +28,4 @@
 /sql_query_results_x_diff-*.sql
 /sql_query-*.sql
 /update-*.sql
+/vacuum_parallel-*.sql
```
svenklemm, line 31: this should not be part of this patch

svenklemm (commented):
I suggest naming the test vacuum_parallel without version number since it will be needed for all versions >= 13

test/sql/.gitignore
nikkhils: yes, removed

nikkhils:
> I suggest naming the test vacuum_parallel without version number since it will be needed for all versions >= 13

yeah, makes sense. Will rename.

pmwkaa (approved):
Looks good

test/expected/vacuum_parallel.out
```diff
@@ -0,0 +1,52 @@
...
+CREATE INDEX vt_temp1 on vacuum_test (temp1);
+CREATE INDEX vt_temp2 on vacuum_test (temp2);
+-- parallel vacuum needs the index size to be larger than 512KB to kick in
+INSERT INTO vacuum_test SELECT TIMESTAMP 'epoch' + (i * INTERVAL '100 second'),
+                i, i+1 FROM generate_series(1, 100000) as T(i);
+-- we should see two parallel workers for each chunk
+VACUUM (PARALLEL 3, VERBOSE) vacuum_test;
+INFO:  vacuuming "_timescaledb_internal._hyper_1_1_chunk"
+INFO:  launched 2 parallel vacuum workers for index cleanup (planned: 2)
+INFO:  index "_hyper_1_1_chunk_vacuum_test_time_idx" now contains 26279 row versions in 131 pages
```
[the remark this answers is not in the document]
nikkhils: yeah, there are other tests which have similar flaky output and it seems to be holding up.

test/sql/vacuum_parallel.sql
```diff
@@ -0,0 +1,26 @@
...
+-- create additional indexes on the temp columns
+CREATE INDEX vt_temp1 on vacuum_test (temp1);
+CREATE INDEX vt_temp2 on vacuum_test (temp2);
+
+-- parallel vacuum needs the index size to be larger than 512KB to kick in
+INSERT INTO vacuum_test SELECT TIMESTAMP 'epoch' + (i * INTERVAL '100 second'),
+                i, i+1 FROM generate_series(1, 100000) as T(i);
+
+-- we should see two parallel workers for each chunk
+VACUUM (PARALLEL 3, VERBOSE) vacuum_test;
```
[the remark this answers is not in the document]
nikkhils: nothing multinode specific AFAICS here.

erimatnor (approved):

PR #2922, opened 2021-02-10
Explicitly check for "-Wno-stringop-truncation" compiler flag

nikkhils:
Older versions of GCC do not have this flag, so check for it explicitly
before adding it.

Fix by Mats Kindahl


mkindahl (approved):
The common way to add co-authors are given in https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors so suggest that you use that.

nikkhils:
> The common way to add co-authors are given in https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors so suggest that you use that.

Ah, ok. Thanks will follow this next time around :-)

pmwkaa (approved):
Looks good

PR #2926, opened 2021-02-10
Optimize cagg refresh for small invalidations

### LINKED ISSUE #2867: TimescaleDB 2.x Continuous Aggrecation long recalculation
We have an issue with the Continuous Aggregation feature with TimescaleDB 2.x upgrade.
We storing data with 1 seconds granularity in hypertable and doing rollups to 10m and 1h intervals with continuous aggregation views.
Before 2.0 we have no issues with refreshes, but with 2.0 rollups get stuck after inserting a bunch of historical data.
We've looked through a code and was able to tell that the issue raises in invalidation logic. 

Intervals

Our system doing separate inserts of data points rounded to 1-second grid (shown in red on the diagram) and this action leads to creating invalidation records in the `_timescaledb_catalog.continuous_aggs_hypertable_invalidation_log` table. Because all of these records are 0 length intervals - the aggregation process cannot merge them into bigger intervals.
This leads to adding dozens of records into  `_timescaledb_catalog.continuous_aggs_materialization_invalidation_log` and the continuous aggregation job invalidates the same 10 minutes interval again and again for each changed record.
Maybe the invalidation process can be more clever and join intervals that fall into the same bucket to avoid recalculating the same intervals over and over again? Another thought is to recalculate the bucket once taking into account all intervals currently invalidating it and not once per invalidation log record.

I hope this problem is not ours only. Please help us to understand and investigate the problem properly.

```
postgres=# SELECT version();
                                                 version
---------------------------------------------------------------------------------------------------------
 PostgreSQL 12.2 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39), 64-bit
(1 row)
postgres=# \dx
                                      List of installed extensions
 timescaledb | 2.0.0   | public     | Enables scalable inserts and complex queries for time-series data
```

erimatnor:
This PR includes two changes:

**Change 1:**
 
The refreshing of a continuous aggregate is slow when many small
invalidations are generated by frequent single row insert
backfills. This change adds an optimization that merges small
invalidations by first expanding invalidations to full bucket
boundaries. There is really no reason to maintain invalidations that
aren't covering full buckets since refresh windows are already aligned
to buckets anyway.

**Change 2:**

When there are many small (e.g., single timestamp) invalidations that
cannot be merged despite expanding invalidations to full buckets
(e.g., invalidations are spread across every second bucket in the
worst case), it might no longer be beneficial to materialize every
invalidation separately.

Instead, this change adds a threshold for the number of invalidations
used by the refresh (currently 10 by default) above which
invalidations are merged into one range based on the lowest and
greatest invalidated time value.

The limit can be controlled by an anonymous session variable for
debugging and tweaking purposes. It might be considered for promotion
to an official GUC in the future.

Fixes #2867

### DIFF
--- a/tsl/src/continuous_aggs/invalidation.c
+++ b/tsl/src/continuous_aggs/invalidation.c
@@ -19,6 +19,8 @@
 #include <scanner.h>
 #include <scan_iterator.h>
 #include <utils.h>
+#include <time_utils.h>
+#include <time_bucket.h>
 
 #include "continuous_agg.h"
 #include "continuous_aggs/materialize.h"
@@ -55,8 +57,8 @@
  * merged during processing to reduce the number of entries in the logs. This
  * typically happens during a refresh of a continuous aggregate, which also
  * cuts invalidations along the refresh window. The cutting will leave some
- * parts of entries in the ivnalidation log while the entries that fall within
- * the refresh window are stored in an invalidation store and used for
+ * parts of entries in the invalidation log while the entries that fall
+ * within the refresh window are stored in an invalidation store and used for
  * refreshing:
  *
  *       |-------------|    refresh window
@@ -78,6 +80,8 @@
 typedef struct CaggInvalidationState
 {
 	ContinuousAgg cagg;
+	Oid dimtype; /* The type of the underlying hypertable's time dimension
+				  * that is bucketed */
 	MemoryContext per_tuple_mctx;
 	Relation cagg_log_rel;
 	Snapshot snapshot;
@@ -459,6 +463,70 @@ invalidation_entry_reset(Invalidation *entry)
 }
 
 /*
+ * Expand an invalidation to bucket boundaries.
+ *
+ * Since a refresh always materializes full buckets, we can safely expand an
+ * invalidation to bucket boundaries and in the process merge a lot more
+ * invalidations.
+ */
+static void
+invalidation_expand_to_bucket_boundaries(Invalidation *inv, Oid timetype, int64 bucket_width)
+{
+	const int64 min_for_type = ts_time_get_min(timetype);
+	const int64 max_for_type = ts_time_get_max(timetype);
+	int64 min_bucket_start;
+	int64 max_bucket_end;
+
+	Assert(bucket_width > 0);
+
+	/* Compute the start of the "first" bucket for the type. The min value
+	 * must be at the start of the "first" bucket or somewhere in the
+	 * bucket. If the min value falls on the exact start of the bucket we are
+	 * good. Otherwise, we need to move to the next full bucket. */
+	min_bucket_start = ts_time_saturating_add(min_for_type, bucket_width - 1, timetype);
+	min_bucket_start = ts_time_bucket_by_type(bucket_width, min_bucket_start, timetype);
+
+	/* Compute the end of the "last" bucket for the time type. Remember that
+	 * invalidations are inclusive, so the "greatest" value should be the last
+	 * value of the last full bucket. Either the max value is already the last
+	 * value of the last bucket, or we need to return the last value of the
+	 * previous full bucket.  */
+	max_bucket_end = ts_time_bucket_by_type(bucket_width, max_for_type, timetype);
+
+	/* Check if the max value was already the last value of the last bucket */
+	if (ts_time_saturating_add(max_bucket_end, bucket_width - 1, timetype) == max_for_type)
+		max_bucket_end = max_for_type;
+	else
+		/* The last bucket was partial. To get the end of previous bucket, we
+		 * need to move one step down from the partial last bucket. */
+		max_bucket_end = ts_time_saturating_sub(max_bucket_end, 1, timetype);
+
+	if (inv->lowest_modified_value < min_bucket_start)
+		/* Below the min bucket, so treat as invalid to -infinity. */
+		inv->lowest_modified_value = INVAL_NEG_INFINITY;
+	else if (inv->lowest_modified_value > max_bucket_end)
+		/* Above the max bucket, so treat as invalid to +infinity. */
+		inv->lowest_modified_value = INVAL_POS_INFINITY;
+	else
+		inv->lowest_modified_value =
+			ts_time_bucket_by_type(bucket_width, inv->lowest_modified_value, timetype);
+
+	if (inv->greatest_modified_value < min_bucket_start)
+		/* Below the min bucket, so treat as invalid to -infinity. */
+		inv->greatest_modified_value = INVAL_NEG_INFINITY;
+	else if (inv->greatest_modified_value > max_bucket_end)
+		/* Above the max bucket, so treat as invalid to +infinity. */
+		inv->greatest_modified_value = INVAL_POS_INFINITY;
+	else
+	{
+		inv->greatest_modified_value =
+			ts_time_bucket_by_type(bucket_width, inv->greatest_modified_value, timetype);
+		inv->greatest_modified_value =
+			ts_time_saturating_add(inv->greatest_modified_value, bucket_width - 1, timetype);
+	}
+}
+
+/*
  * Macro to set an Invalidation from a tuple. The tuple can either have the
  * format of the hypertable invalidation log or the continuous aggregate
  * invalidation log (as determined by the type parameter).
@@ -480,9 +548,9 @@ invalidation_entry_reset(Invalidation *entry)
 			heap_freetuple(tuple);                                                                 \
 	} while (0);
 
-void
+static void
 invalidation_entry_set_from_hyper_invalidation(Invalidation *entry, const TupleInfo *ti,
-											   int32 hyper_id)
+											   int32 hyper_id, Oid dimtype, int64 bucket_width)
 {
 	INVALIDATION_ENTRY_SET(entry,
 						   ti,
@@ -492,15 +560,27 @@ invalidation_entry_set_from_hyper_invalidation(Invalidation *entry, const TupleI
 	 * invalidation log, a different hypertable ID must be set (the ID of the
 	 * materialized hypertable). */
 	entry->hyper_id = hyper_id;
+	invalidation_expand_to_bucket_boundaries(entry, dimtype, bucket_width);
 }
 
 static void
-invalidation_entry_set_from_cagg_invalidation(Invalidation *entry, const TupleInfo *ti)
+invalidation_entry_set_from_cagg_invalidation(Invalidation *entry, const TupleInfo *ti, Oid dimtype,
+											  int64 bucket_width)
 {
 	INVALIDATION_ENTRY_SET(entry,
 						   ti,
 						   materialization_id,
 						   Form_continuous_aggs_materialization_invalidation_log);
+
+	/* It isn't strictly necessary to expand the invalidation to bucket
+	 * boundaries here since all invalidations were already expanded when
+	 * copied from the hypertable invalidation log. However, since
+	 * invalidation expansion wasn't implemented in early 2.0.x versions of
+	 * the extension, there might be unexpanded entries in the cagg
+	 * invalidation log for some users. Therefore we try to expand
+	 * invalidation entries also here, although in most cases it would do
+	 * nothing. */
+	invalidation_expand_to_bucket_boundaries(entry, dimtype, bucket_width);
 }
 
 /*
@@ -617,6 +697,7 @@ move_invalidations_from_hyper_to_cagg_log(const CaggInvalidationState *state)
 	foreach (lc, cagg_ids)
 	{
 		int32 cagg_hyper_id = lfirst_int(lc);
+		ContinuousAgg *cagg = ts_continuous_agg_find_by_mat_hypertable_id(cagg_hyper_id);
 		Invalidation mergedentry;
 		ScanIterator iterator;
 
@@ -634,7 +715,11 @@ move_invalidations_from_hyper_to_cagg_log(const CaggInvalidationState *state)
 			oldmctx = MemoryContextSwitchTo(state->per_tuple_mctx);
 			ti = ts_scan_iterator_tuple_info(&iterator);
 
-			invalidation_entry_set_from_hyper_invalidation(&logentry, ti, cagg_hyper_id);
+			invalidation_entry_set_from_hyper_invalidation(&logentry,
+														   ti,
+														   cagg_hyper_id,
+														   state->dimtype,
+														   cagg->data.bucket_width);
 
 			if (!IS_VALID_INVALIDATION(&mergedentry))
 			{
@@ -799,7 +884,10 @@ clear_cagg_invalidations_for_refresh(const CaggInvalidationState *state,
 		Invalidation logentry;
 
 		oldmctx = MemoryContextSwitchTo(state->per_tuple_mctx);
-		invalidation_entry_set_from_cagg_invalidation(&logentry, ti);
+		invalidation_entry_set_from_cagg_invalidation(&logentry,
+													  ti,
+													  state->dimtype,
+													  state->cagg.data.bucket_width);
 
 		if (!IS_VALID_INVALIDATION(&mergedentry))
 			mergedentry = logentry;
@@ -838,9 +926,10 @@ clear_cagg_invalidations_for_refresh(const CaggInvalidationState *state,
 }
 
 static void
-invalidation_state_init(CaggInvalidationState *state, const ContinuousAgg *cagg)
+invalidation_state_init(CaggInvalidationState *state, const ContinuousAgg *cagg, Oid dimtype)
 {
 	state->cagg = *cagg;
+	state->dimtype = dimtype;
 	state->cagg_log_rel = open_invalidation_log(LOG_CAGG, RowExclusiveLock);
 	state->per_tuple_mctx = AllocSetContextCreate(CurrentMemoryContext,
 												  "Continuous aggregate invalidations",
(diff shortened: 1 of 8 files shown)


tsl/src/continuous_aggs/refresh.c
```diff
@@ -290,15 +291,71 @@ log_refresh_window(int elevel, const ContinuousAgg *cagg, const InternalTimeRang
 		 DatumGetCString(OidFunctionCall1(outfuncid, end_ts)));
 }
 
+#define MAX_INVALIDATIONS_PER_REFRESH 10
```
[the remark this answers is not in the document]
erimatnor: The semantics here are different from `max_interval_per_job`. The old `max_interval_per_job` setting is actually already similar to the user- or policy-specified refresh window; i.e., user controls the max amount materialized by specifying a specific refresh window size.

This max value is about how we optimize the materialization of the specified refresh window. Pre-2.0, everything from the earliest invalidation and forward was re-materialized no matter whether the entire range needed re-materialization or not. Post-2.0 we break that down so that we only materialize the regions/buckets that are actually invalid. However, if there are a lot of small regions that are invalidated (e.g., every other bucket in the worst case), it is cheaper to merge that into a large range. The session variable below controls the threshold.

tsl/src/continuous_aggs/invalidation.c
```diff
@@ -458,6 +462,53 @@ invalidation_entry_reset(Invalidation *entry)
...
+	 * and move one step down to the end value of the previous
+	 * bucket. Remember that invalidations are inclusive, so the "greatest"
+	 * value should be the last value of the last full bucket. */
+	max_bucket_end = ts_time_bucket_by_type(bucket_width, ts_time_get_max(timetype), timetype);
+	max_bucket_end = ts_time_saturating_sub(max_bucket_end, 1, timetype);
+
+	if (inv->lowest_modified_value < min_bucket_start)
+		inv->lowest_modified_value = PG_INT64_MIN;
+	else if (inv->lowest_modified_value > max_bucket_end)
+		inv->lowest_modified_value = PG_INT64_MAX;
```
mkindahl, line 494: Isn't this a problem if `PG_INT16_MIN` or `PG_INT64_MAX` happen to be in the middle of a bucket?

tsl/src/continuous_aggs/refresh.c
```diff
@@ -290,15 +291,71 @@ log_refresh_window(int elevel, const ContinuousAgg *cagg, const InternalTimeRang
...
+		GetConfigOption("timescaledb.max_invalidations_per_refresh", true, false);
+	long max_invalidations = MAX_INVALIDATIONS_PER_REFRESH;
+
+	if (max_invalidations_setting)
+	{
+		char *endptr = NULL;
+
+		/* Not using pg_strtol here since we don't want to throw error in case
+		 * of parsing issue */
+		max_invalidations = strtol(max_invalidations_setting, &endptr, 10);
```
mkindahl, line 315: Why not? If the variable is set to a strange string, isn't it better to raise an error early rather than running with a default value? A user making a subtle mistake might be befuddled about why things are behaving strange.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -290,15 +291,71 @@ log_refresh_window(int elevel, const ContinuousAgg *cagg, const InternalTimeRang
...
+
+		if (*endptr != '\0')
+			max_invalidations = MAX_INVALIDATIONS_PER_REFRESH;
```
mkindahl, line 318: What about trailing spaces? Shouldn't that be handled? Also, AFAICT, the empty string will result in 0 max invalidation, which might not want be what you want.

tsl/src/continuous_aggs/refresh.c
```diff
@@ -326,8 +383,39 @@ continuous_agg_refresh_with_window(const ContinuousAgg *cagg,
...
+	if (do_merged_refresh && count > 0)
+	{
+		Assert(merged_refresh_window.type == refresh_window->type);
+		Assert(merged_refresh_window.start >= refresh_window->start);
+		Assert(merged_refresh_window.end <= refresh_window->end);
+
+		log_refresh_window(DEBUG1,
+						   cagg,
+						   &merged_refresh_window,
+						   psprintf("merged %ld invalidations for refresh on", count));
```
mkindahl, line 417: I think this should be a notice (https://www.postgresql.org/docs/12/runtime-config-logging.html#RUNTIME-CONFIG-SEVERITY-LEVELS):

> NOTICE | Provides information that might be helpful to users, e.g., notice of truncation of long identifiers.

mkindahl (commented):
Found a few things that are not clear, could be, or is a problem. If you do not think they are, please respond to them.

I also think that is not a good idea to have the `max_invalidations_per_refresh` for the user function and especially having it to default to 10 (which it does, from my understanding). For the policy runs, this is just a way to throttle the background workers, which is fine, but when calling this from the command line the result will be that the entire range is not refreshed and the user will have to type the command several times, which will be annoying. I think it might be better to add a parameter to the function which defaults to do a full refresh and ensure that the policy set a value for this parameter instead. This will make sure that policies can be throttled and have a good default value, while command-line calls will refresh the entire range.

tsl/src/continuous_aggs/refresh.c
pmwkaa: Can this be checked by the GUC callback or something like that? If there is none, we should validate it

tsl/test/sql/continuous_aggs_invalidation.sql
```diff
@@ -619,3 +619,49 @@ INSERT INTO conditions VALUES(6, 1, 1.0);
...
+SELECT * FROM cagg_invals
+WHERE cagg_id = :cond_10_id;
+
+-- should trigger two individual refreshes
+SET client_min_messages TO DEBUG1;
+CALL refresh_continuous_aggregate('cond_10', 0, 200);
+RESET client_min_messages;
+
+-- Allow at most 5 individual invalidations per refreshe
+SET timescaledb.max_invalidations_per_refresh=5;
```
pmwkaa, line 654: Suggest to add test, that checks this value is correct and produce an error if not

tsl/src/continuous_aggs/invalidation.c
```diff
@@ -458,6 +462,53 @@ invalidation_entry_reset(Invalidation *entry)
...
+
+	/* To get the end of the last full bucket, we need to bucket the max value
+	 * and move one step down to the end value of the previous
+	 * bucket. Remember that invalidations are inclusive, so the "greatest"
+	 * value should be the last value of the last full bucket. */
+	max_bucket_end = ts_time_bucket_by_type(bucket_width, ts_time_get_max(timetype), timetype);
+	max_bucket_end = ts_time_saturating_sub(max_bucket_end, 1, timetype);
+
+	if (inv->lowest_modified_value < min_bucket_start)
+		inv->lowest_modified_value = PG_INT64_MIN;
```
pmwkaa, line 492: Do we have tests somewhere which use those MIN/MAX as a values?

pmwkaa (approved):
Looks good overall, don't have much to add, just a couple of comments

erimatnor:
> I also think that is not a good idea to have the `max_invalidations_per_refresh` for the user function and especially having it to default to 10 (which it does, from my understanding). For the policy runs, this is just a way to throttle the background workers, which is fine, but when calling this from the command line the result will be that the entire range is not refreshed and the user will have to type the command several times, which will be annoying. 

So, this seems to be a misunderstanding of what the changes do, and that's on me for failing to make clear and perhaps a badly named variable.

Two things that I'd like to clarify:

* The changes in this PR in no way changes the semantics of `refresh_continuous_aggregate`. After a call to this function, everything in the given window will be up-to-date. This hasn't changed. What happens internally, though, is that we break a refresh down into a set of multiple sub-refreshes on the ranges (buckets) that are actually out-of-date. But if that turns out to be 100 separate refreshes, this is likely to be slower than one single refresh across the range determined by the min invalidated time and max invalidated time. The `max_invalidations_per_refresh` just sets the threshold of where we merge those separate ranges into one big range. It does _not_ mean that we leave some ranges out-of-date.
* None of the changes in this PR has anything to do with throttling policies. Not sure where that impression came from.

tsl/src/continuous_aggs/refresh.c
erimatnor: I can add handling for trailing spaces, but I don't think this is necessary for something that is not an official GUC. Falling back to the default value or even 0 is OK. All of these will work.

tsl/src/continuous_aggs/refresh.c
erimatnor: I disagree. A notice is sent to the client and this is not normally valuable information to the user since it is implementation detail (debug information). Users care that the result of the refresh is what they expect, not _how_ it was executed. If they want to see this info, they can always change the debug level.

tsl/test/sql/continuous_aggs_invalidation.sql
erimatnor: I can do that if deemed important. But I don't think we should waste time on that for something that is just a session variable for debugging and not an real GUC. Also, it is not possible to parse session variables when set, so the failure would appear when running refresh and persist until the variable is changed.

tsl/src/continuous_aggs/invalidation.c
erimatnor: I am actually unsure what is asked. MIN/MAX is used to mean invalid to -infinity and +infinity. There are several tests (e.g., `continuous_aggs_invalidation`) which show such invalidations in the invalidation log.

pmwkaa (approved):

tsl/src/continuous_aggs/refresh.c
pmwkaa: Sorry, this was not clear, that we are not going to document it

tsl/src/continuous_aggs/refresh.c
pmwkaa: Do we need to check `strtol` for overflows as well? errno `ERANGE` or endptr is enought?

erimatnor: In my understanding, checking endptr is enough unless we want to understand more about what failed.

tsl/src/continuous_aggs/refresh.c
erimatnor: Sorry it wasn't clear. Tried to make clear in the function: "* Use a session variable for now, for debugging and tweaking. Might promote to full GUC if useful."

tsl/src/continuous_aggs/invalidation.c
```diff
@@ -458,6 +462,53 @@ invalidation_entry_reset(Invalidation *entry)
...
+	 * good. Otherwise, we need to move to the next full bucket. */
+	min_bucket_start =
+		ts_time_saturating_add(ts_time_get_min(timetype), bucket_width - 1, timetype);
+	min_bucket_start = ts_time_bucket_by_type(bucket_width, min_bucket_start, timetype);
+
+	/* To get the end of the last full bucket, we need to bucket the max value
+	 * and move one step down to the end value of the previous
+	 * bucket. Remember that invalidations are inclusive, so the "greatest"
+	 * value should be the last value of the last full bucket. */
+	max_bucket_end = ts_time_bucket_by_type(bucket_width, ts_time_get_max(timetype), timetype);
```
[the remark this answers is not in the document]
erimatnor: I thought about it, but figured it wasn't worth it. Perhaps you have a suggestion of how to easily cache it without making this unnecessarily complex?

tsl/src/continuous_aggs/refresh.c
mkindahl: The example given in the manual is where something happened that might change the behavior but not likely affect the result, e.g., truncation of identifier. The PostgreSQL code uses notices internally for these things (and others, but these are examples):

- Operations are skipped for statements, e.g., table is not dropped in a drop statement
- Constraints are merged with inherited definition
- Merging multiple inherited definitions

I think this message belongs in the same category.

tsl/src/continuous_aggs/invalidation.c
```diff
@@ -458,6 +462,53 @@ invalidation_entry_reset(Invalidation *entry)
...
+invalidation_expand_to_bucket_boundaries(Invalidation *inv, Oid timetype, int64 bucket_width)
+{
+	int64 min_bucket_start;
+	int64 max_bucket_end;
+
+	/* If the min value falls on the exact start of a full bucket we are
+	 * good. Otherwise, we need to move to the next full bucket. */
+	min_bucket_start =
+		ts_time_saturating_add(ts_time_get_min(timetype), bucket_width - 1, timetype);
+	min_bucket_start = ts_time_bucket_by_type(bucket_width, min_bucket_start, timetype);
```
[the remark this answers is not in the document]
mkindahl: The comment was hard to read, you might want to clarify it.

tsl/src/continuous_aggs/refresh.c
mkindahl: The point of raising an error here is to ensure that the value is corrected rather than using a default value, so yes, it should throw an error until the value is changed. Validating it when setting it is (of course) better.

mkindahl:
> > I also think that is not a good idea to have the `max_invalidations_per_refresh` for the user function and especially having it to default to 10 (which it does, from my understanding). For the policy runs, this is just a way to throttle the background workers, which is fine, but when calling this from the command line the result will be that the entire range is not refreshed and the user will have to type the command several times, which will be annoying.
> 
> So, this seems to be a misunderstanding of what the changes do, and that's on me for failing to make clear and perhaps a badly named variable.
> 
> Two things that I'd like to clarify:
> 
>     * The changes in this PR in no way changes the semantics of `refresh_continuous_aggregate`. After a call to this function, everything in the given window will be up-to-date. This hasn't changed. What happens internally, though, is that we break a refresh down into a set of multiple sub-refreshes (materialization) on the ranges (buckets) that are actually out-of-date. But if that turns out to be 100 separate refreshes, this is likely to be slower than one single refresh across the range determined by the min invalidated time and max invalidated time. The `max_invalidations_per_refresh` just sets the threshold of where we merge those separate ranges into one big range. It does _not_ mean that we leave some ranges out-of-date.
> 
>     * None of the changes in this PR has anything to do with throttling policies. Not sure where that impression came from.
> 
> 
> I've tried to change the comments and variable names to make this more clear.

I understand what the variable do, but IMHO the problem is that the default behavior is to not process the entire range when it is "too complicated" (when we have a lot of buckets that are not connected). We had several comments before when `REFRESH MATERIALIZED VIEW` didn't refresh everything, so I think we should make sure that we, by default, refresh the entire range even if that happens to take a long time.

erimatnor:
> I understand what the variable do, but IMHO the problem is that the default behavior is to not process the entire range when it is "too complicated" (when we have a lot of buckets that are not connected). We had several comments before when `REFRESH MATERIALIZED VIEW` didn't refresh everything, so I think we should make sure that we, by default, refresh the entire range even if that happens to take a long time.

Above the threshold we actually process more than necessary, so it is actually the opposite of your understanding. When you say we had comments about `REFRESH MATERIALIZED VIEW`, I assume you are referring to pre-2.0 behavior. That functionality was entirely different from what we do here, so there is no clear comparison. The semantics of `refresh_continuous_aggregate` here is to leave everything in the window/range up-to-date after the call completes. So, your statement about not processing the entire range is incorrect.

mkindahl (approved):
Thanks for the clarification that the changes does not affect the range that the refresh function process. LGTM.

tsl/src/continuous_aggs/refresh.c
erimatnor: Added a warning as agreed.

tsl/src/continuous_aggs/refresh.c
erimatnor: FWIW, if the integer value is too big it will just return LONG_MAX. Added test.

tsl/src/continuous_aggs/refresh.c
erimatnor: This is about how we internally process invalidations. Users might not even know what an invalidation is, and most won't care.

tsl/src/continuous_aggs/invalidation.c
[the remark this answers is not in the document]
erimatnor: which comment are you referring to? Is it the one at line 478 or the comment on the function? How can it be clarified?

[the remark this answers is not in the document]
mkindahl: It is the comment on line 478. AFAICT, you're computing the beginning of the first bucket completely inside the range of the type, but this is not clear from the comment.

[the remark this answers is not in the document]
erimatnor: Made an attempt to clarify.

erimatnor:
@mkindahl @pmwkaa Changed the following based on feedback.

* Updated comments for min/max bucket calculation (also made a small tweak to calculating the end of the last bucket)
* Added warning for the session variable parsing, and also added support for trailing whitespaces
* Added tests for non-parsable input of session variable

tsl/src/continuous_aggs/invalidation.c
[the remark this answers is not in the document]
mkindahl: LGTM. Good that you did the same change for the max variable.

mkindahl:
> > I understand what the variable do, but IMHO the problem is that the default behavior is to not process the entire range when it is "too complicated" (when we have a lot of buckets that are not connected). We had several comments before when `REFRESH MATERIALIZED VIEW` didn't refresh everything, so I think we should make sure that we, by default, refresh the entire range even if that happens to take a long time.
> 
> Above the threshold we actually process more than necessary, so it is actually the opposite of your understanding. When you say we had comments about `REFRESH MATERIALIZED VIEW`, I assume you are referring to pre-2.0 behavior. That functionality was entirely different from what we do here, so there is no clear comparison. The semantics of `refresh_continuous_aggregate` here is to leave everything in the user-specified window/range up-to-date after the call completes. So, your statement about not processing the entire range is not the way it works.

Good to hear. Then I misunderstood how it worked.

PR #2927, opened 2021-02-10
Add tests for ALTER COLUMN .. DROP EXPRESSION

gayyappan:
This PR adds tests for dropping defaults for generated
 columns. This is a feature introduced in PG13. This is done by executing
 ALTER TABLE .. ALTER COLUMN .. DROP EXPRESSION


svenklemm (approved):
maybe change tablenames to not be foo/bar

mkindahl (approved):
We generate test for PG12 even though generated columns do not exist in PG12, which causes the error. Better would be to not have the test generated at all and just have one for PG13. 

Approving, but ideally we should not run more tests than necessary since they add up to the test time.

PR #2928, opened 2021-02-11
Add test to check incremental sort functionality with PG13+

nikkhils:
Starting with PG13, when a relation is already sorted by
(key1, key2) and we need to sort it by (key1, key2, key3) we can simply
split the input rows into groups having equal values in (key1, key2),
and only sort/compare the remaining column key3.

This commit adds a test for hypertables.

### DIFF
--- a/tsl/test/shared/expected/incremental_sort.out
+++ b/tsl/test/shared/expected/incremental_sort.out
@@ -1,0 +1,98 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+SELECT
+       format('include/%s.sql', :'TEST_BASE_NAME') as "TEST_QUERY_NAME",
+       format('%s/shared/results/%s_results_unoptimized.out', :'TEST_OUTPUT_DIR', :'TEST_BASE_NAME') as "TEST_RESULTS_UNOPTIMIZED",
+       format('%s/shared/results/%s_results_incremental_sort.out', :'TEST_OUTPUT_DIR', :'TEST_BASE_NAME') as "TEST_RESULTS_INCREMENTAL_SORT",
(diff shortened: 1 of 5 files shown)


k-rus (commented):
LGTM
I suggest to test that the query results are equivalent between enabled and disabled incremental sort to prevent future regressions if new optimisations will be implemented in TimescaleDB.

nikkhils:
> LGTM
> I suggest to test that the query results are equivalent between enabled and disabled incremental sort to prevent future regressions if new optimisations will be implemented in TimescaleDB.

Sure. Added actual SELECT queries to run and get the output in both the case.

svenklemm (changes requested):
This needs more tests, especially interaction with ordered append should be tested

nikkhils:
> This needs more tests, especially interaction with ordered append should be tested

Added a query which uses incremental sort with `ChunkAppend`

gayyappan (approved):

svenklemm (commented):
I think you should move this test to regresscheck-shared and try to use the tables already existing in that environment. When testing query output we usually diff the output of optimized/unoptimized output to make the test output itself less noisy.

You can have a look at https://github.com/timescale/timescaledb/blob/master/tsl/test/shared/sql/ordered_append.sql.in for a test doing something like that

nikkhils:
> I think you should move this test to regresscheck-shared and try to use the tables already existing in that environment. When testing query output we usually diff the output of optimized/unoptimized output to make the test output itself less noisy.
> 
> You can have a look at https://github.com/timescale/timescaledb/blob/master/tsl/test/shared/sql/ordered_append.sql.in for a test doing something like that

@svenklemm I have moved the test to shared directory and used the same format as others. Please take a look.

nikkhils:
@svenklemm pointed out that "Incremental Sort" plans are already present in other test files. So, adding a small subset here in a new test file does not achieve much in terms of test coverage. So based on that, we are abandoning this PR.

PR #2931, opened 2021-02-11

svenklemm:
Run tests on PG 11.11, 12.6 and 13.2


mkindahl (approved):
Changes look fine, but is it possible to have a tag denoting "latest PG11" (for example) instead of having to update this every time?

gayyappan (approved):

PR #2933, opened 2021-02-11
Add Docker authentication to Travis builds

erimatnor:
Docker has added rate limits for image pulls, which breaks some Travis
builds when the rate limit is reached. By authenticating with
DockerHub we get a higher rate limit that we don't share with other
anonymous users.

Note that PRs have to be created from a branch in the main repo (not from a fork) after this PR is merged. This is because PRs from forks cannot access encrypted environment variables needed for Docker credentials.


.travis.yml
```diff
@@ -59,11 +60,12 @@ jobs:
   include:
 
     # This tests the formatting of a PR.
-    - if: (type = pull_request) OR (type = cron) OR NOT (branch = master)
+    - if: (type = pull_request) OR (type = cron) OR (branch = prerelease_test)
```
erimatnor, line 63: This seems to have had the wrong logic since it triggers duplicate builds when one pushes a branch to the main repo and makes a pull request from there. (I.e., this used to build on both PR and updates to any branch except master).

svenklemm (approved):

NunoFilipeSantos (approved):

pmwkaa (approved):

PR #2940, opened 2021-02-12
Add PG13 to 32bit tests, coverity and cron tests

svenklemm:


.github/workflows/coverity.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pg: ["11.11", "12.6"]
+        pg: ["12.6", "13.2"]
```
nikkhils, line 16: We don't care about coverity on PG11 anymore?

nikkhils (approved):
LGTM other than my query about coverity on PG11.

.github/workflows/coverity.yaml
svenklemm: coverity has a rate limit so i wanted to limit the number of runs and PG11 support will be removed eventually

mkindahl (approved):

PR #2941, opened 2021-02-14
Set minimum required PG13 version to 13.2

svenklemm:
The minimum required PG13 version is 13.2 because we require a
bugfix only present in that version otherwise decompression might
not work properly for certain queries.


nikkhils:
LGTM.

nikkhils (approved):

mkindahl (approved):

PR #2943, opened 2021-02-15
Fix detection of stringop truncation compiler support

svenklemm:
When checking for -Wno-stringop-truncation gcc reports success even
though it doesn't support the flag. This patch changes the check
to check for actual flag instead of the flag that turns it off
which produces the correct result.


mkindahl (approved):

erimatnor (approved):

nikkhils (approved):
LGTM

PR #2948, opened 2021-02-16
Fix off by 4 error in histogram deserialize

### LINKED ISSUE #2684: Off by 4 error in histogram deserialize
In [the histogram deserialization function](https://github.com/timescale/timescaledb/blob/825645b2f5c3685ce50d643f657e48d35677d015/src/histogram.c#L194-L196l) the `StringInfoData`'s pointer is set to `VARDATA(...)` which _does not_ include the `varlen` header, while the it's length is set to `VARSIZE(...)` which _does_ include the `varlen` header's length. This probably does not result in issues in practice because the actual deserialization is driven off the stored `nbuckets` field.

svenklemm:
In the histogram deserialization function the StringInfoData's pointer
is set to VARDATA(...) which does not include the varlen header, while
the length is set to VARSIZE(...) which does include the header's
length.

Fixes #2684

### DIFF
--- a/src/histogram.c
+++ b/src/histogram.c
@@ -192,8 +192,8 @@ ts_hist_deserializefunc(PG_FUNCTION_ARGS)
 	serialized = PG_GETARG_BYTEA_P(0);
 
 	buf.data = VARDATA(serialized);
-	buf.len = VARSIZE(serialized);
-	buf.maxlen = VARSIZE(serialized);
+	buf.len = VARSIZE(serialized) - VARHDRSZ;
+	buf.maxlen = VARSIZE(serialized) - VARHDRSZ;
 	buf.cursor = 0; /* used by pq_getmsgint*/
 
 	nbuckets = pq_getmsgint(&buf, 4);


pmwkaa (approved):

erimatnor (approved):

mkindahl (approved):
LGTM

src/histogram.c
```diff
@@ -192,8 +192,8 @@ ts_hist_deserializefunc(PG_FUNCTION_ARGS)
 	serialized = PG_GETARG_BYTEA_P(0);
 
 	buf.data = VARDATA(serialized);
-	buf.len = VARSIZE(serialized);
-	buf.maxlen = VARSIZE(serialized);
+	buf.len = VARSIZE(serialized) - VARHDRSZ;
```
erimatnor, line 195: I guess it would be good to have a test that could trigger the bad behavior without this fix.

erimatnor: Since this is a stringinfo data type, we should use the API `appendBinaryStringInfo` to set these fields instead of directly modifying the struct.

erimatnor: I missed the maxlen, but still think it is good to use stringinfo functions. I don't think the extra copy is a problem.

PR #2956, opened 2021-02-17
Disable strict-overflow for gcc < 8

svenklemm:
The strict-overflow check on gcc < 8 produces false positives
leading to build failuren when compiling with -Werror


mkindahl (approved):

svenklemm:
Example for failure: https://github.com/timescale/timescaledb/runs/1918721832?check_suite_focus=true

gayyappan (approved):

PR #2986, opened 2021-02-26
Fix mismatched dsm_attach/dsm_detach calls

nikkhils:
A backend is allowed to only do one dsm_attach call to attach to
any given dynamic shared memory segment (it needs to detach before
calling attach again). The code in params_open_wrapper took care of
calling it once. However the code in params_close_wrapper was doing an
unconditional dsm_detach everytime which can cause the dsm refcnt to go
down unnecessarily and was causing weird crashes

The fix is to track when we should be calling params_close_wrapper
instead of doing an unconditional detach always.

### DIFF
--- a/test/src/bgw/params.c
+++ b/test/src/bgw/params.c
@@ -99,19 +99,26 @@ params_get_dsm_handle()
 }
 
 static TestParamsWrapper *
-params_open_wrapper()
+params_open_wrapper(bool *do_close)
 {
 	dsm_segment *seg;
 	dsm_handle handle = params_get_dsm_handle();
 	TestParamsWrapper *wrapper;
 
+	/*
+	 * If segment is returned via the mapping then there's no need to call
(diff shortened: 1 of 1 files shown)


test/src/bgw/params.c
```diff
@@ -99,19 +99,26 @@ params_get_dsm_handle()
 }
 
 static TestParamsWrapper *
-params_open_wrapper()
+params_open_wrapper(bool *do_close)
```
pmwkaa, line 102: nit: maybe `openned`?

pmwkaa (approved):
Looks good

svenklemm (approved):

test/src/bgw/params.c
nikkhils: just trying to be consistent with the corresponding `params_close_wrapper` call ;-)

PR #2987, opened 2021-02-26
Fix crash due to txns in emit_log_hook_callback

nikkhils:
A weird crash was coming up while running the bgw_db_scheduler test
while using log_min_messages with DEBUG3 setting. So what's happening
here is that we register a RegisterXactCallback callback which gets
called at post-commit time. This remote_connection_xact_end callback
was issuing a debug3 level message. As part of its processing, the
emit_log_hook_callback function was trying to write this message
to a catalog table via a transaction. Obviously trying to start a
transaction at this post-commit stage will error out and leads to
weird behavior and crashes. In general it's not a good idea to have
emit_log_hook_callback functions start transactions because messages
can be emitted at any point of a transaction cycle, but we have to live
with the current code for now.
    
The fix is to reset the emit_log_hook in these xact_end callback
functions. Additionally the emit_log_hook_callback function is only
interested in elog/ereport messages with elevel LOG or above. So add
that check in the emit_log_hook_callback as well.
    
Fixes https://github.com/timescale/timescaledb-private/issues/899

### DIFF
--- a/test/src/bgw/log.c
+++ b/test/src/bgw/log.c
@@ -61,6 +61,11 @@ bgw_log_insert(char *msg)
 
 static emit_log_hook_type prev_emit_log_hook = NULL;
 
+/*
+ * NOTE: using transactions in emit_log_hook functions is not recommended.
+ * However we rely on this current functionality for our test verifications,
+ * so have to live with it for now.
+ */
 static void
 emit_log_hook_callback(ErrorData *edata)
 {
@@ -70,6 +75,10 @@ emit_log_hook_callback(ErrorData *edata)
 	 * once proc_exit has started we may no longer be able to start transactions
 	 */
 	if (MyProc == NULL)
+		return;
+
+	/* We are only interested in elevel LOG and above. */
+	if (edata->elevel < LOG)
 		return;
 
 	/*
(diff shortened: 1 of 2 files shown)


pmwkaa:
Is it possible to add a test case for it?

nikkhils:
> Is it possible to add a test case for it?

This came up when I was running `bgw_db_scheduler` test with debug3 `log_min_messages`. So the test case is already there

mkindahl (approved):
LGTM

test/src/bgw/log.c
```diff
@@ -107,6 +116,10 @@ emit_log_hook_callback(ErrorData *edata)
 	}
 	PG_CATCH();
 	{
+		/* If there was an error, rollback what was done before the error */
+		if (IsTransactionState())
+			AbortCurrentTransaction();
```
pmwkaa, line 121: Is that OK do it in the error handler here? Is it possible that `AbortCurrentTransaction()` can also throw an error?

pmwkaa (approved):

svenklemm:
please squash commits before merging

test/src/bgw/log.c
nikkhils: It's possible in which case a `RETHROW` might be required to handle it. But that's not needed here.

PR #2998, opened 2021-03-02
Mark timescaledb as trusted extension

svenklemm:
PG13 adds the trusted attribute to extensions. Extensions marked
as trusted do not require superuser privileges, but can be installed
by a non-superuser with CREATE privilege in the database.
The objects within the extension will be owned by the bootstrap
superuser, but the extension itself will be owned by the calling user.

https://github.com/postgres/postgres/commit/50fc694e

### DIFF
--- a/test/sql/CMakeLists.txt
+++ b/test/sql/CMakeLists.txt
@@ -98,6 +98,9 @@ if (CMAKE_BUILD_TYPE MATCHES Debug)
     timestamp_limits.sql
     with_clause_parser.sql
   )
+  list(APPEND TEST_TEMPLATES
+    multi_transaction_index.sql.in
+  )
 endif (CMAKE_BUILD_TYPE MATCHES Debug)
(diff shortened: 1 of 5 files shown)


mkindahl (approved):

test/sql/CMakeLists.txt
```diff
@@ -98,6 +98,9 @@ if (CMAKE_BUILD_TYPE MATCHES Debug)
     timestamp_limits.sql
     with_clause_parser.sql
   )
+  list(APPEND TEST_TEMPLATES
```
pmwkaa, line 101: Is that change of this PR?

pmwkaa (approved):

test/sql/CMakeLists.txt
svenklemm: just  combines the two if (debug) into 1

PR #3020, opened 2021-03-11
Fix flaky remote_txn.sql test

nikkhils:
The remote_txn.sql test used to be consistently flaky in CI runs.

The RegisterXactCallback function in the test does the following on the
access node:

1) Send an async command to the remote node
2) Invoke the eventcallback, for example:
eventcallback(DTXN_EVENT_WAIT_COMMIT) which kills the remote node
backend
3) Wait for async responses for (1) from remote node and expect it to
fail due to the killing of the remote backend in (2) above.

The above works most of the times but it's possible that (1) completes
successfully and returns successful async response via (3) before the
remote backend gets killed via (2).

To make things more predictable, we now use the DEBUG_WAITPOINT
infrastructure.

To ensure predictability, we want to kill the remote backend when it's
in the midst of processing the transaction. To ensure that the access
node sets the event handler and then takes an exclusive lock on the
"remote_conn_xact_end" advisory lock via "debug_waitpoint_enable"
function

The "RegisterXactCallback" callback on the remote backend tries to take
this same advisory lock in shared mode and waits. This allows the event
handler enough time to kill this remote backend at the right time.

The DEBUG_WAITPOINT infrastructure was modified to provide a "blocking"
and "retry loop" based implementation to take the advisory lock. This
is needed in cases where this function gets called from deep down
inside a transaction where interrupts are not being served currently.
    
The "retry loop" part also returns with a WARNING if the "retry_count"
is exhausted. We could provide an implementation later which decides to
retry indefinitely but the present implementation is good enough for
the current requirements.

### DIFF
--- a/src/debug_wait.c
+++ b/src/debug_wait.c
@@ -11,6 +11,7 @@
 #include <fmgr.h>
 
 #include <access/hash.h>
+#include <storage/ipc.h>
 #include <storage/lock.h>
 #include <miscadmin.h>
 #include <utils/builtins.h>
@@ -31,7 +32,7 @@ ts_debug_waitpoint_init(DebugWait *waitpoint, const char *tagname)
 
 	SET_LOCKTAG_ADVISORY(waitpoint->tag, MyDatabaseId, (uint32)(hash >> 32), (uint32) hash, 1);
 	waitpoint->tagname = pstrdup(tagname);
-	ereport(DEBUG1,
+	ereport(DEBUG3,
 			(errmsg("initializing waitpoint '%s' to use " UINT64_FORMAT,
 					waitpoint->tagname,
 					hash)));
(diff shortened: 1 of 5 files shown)


src/debug_wait.c
```diff
@@ -44,23 +45,64 @@ ts_debug_waitpoint_init(DebugWait *waitpoint, const char *tagname)
...
+		lock_acquire_result = LockAcquire(&waitpoint->tag, ShareLock, true, true);
+		while (lock_acquire_result == LOCKACQUIRE_NOT_AVAIL)
+		{
+			/* don't dare to take a lock when the proc is exiting! */
+			if (proc_exit_inprogress || ProcDiePending)
+				return;
+
+			if (retry_count == 0)
+			{
+				elog(WARNING, "timeout while acquiring shared lock");
```
erimatnor, line 88: Would be good to give some context that this is a waitpoint lock:

```suggestion
				elog(WARNING, "timeout while acquiring waitpoint lock");
```

src/debug_wait.c
```diff
@@ -44,23 +45,64 @@ ts_debug_waitpoint_init(DebugWait *waitpoint, const char *tagname)
...
+		 *
+		 * If required, in future, we could take an additional option to decide
+		 * if the caller wants to retry indefinitely or return with a warning.
+		 * But the current behavior based on the "blocking" argument is ok for
+		 * now.
+		 */
+		unsigned int retry_count = 1000;
+
+		/* try to acquire the lock without waiting. */
+		lock_acquire_result = LockAcquire(&waitpoint->tag, ShareLock, true, true);
```
erimatnor, line 79: since this statement is repeated at the end of the loop, maybe we can just have one statement within the loop instead, and then break out or if it on the right conditions?

src/debug_wait.c
```diff
@@ -44,23 +45,64 @@ ts_debug_waitpoint_init(DebugWait *waitpoint, const char *tagname)
...
+		 * Trying to wait indefinitely here could lead to hangs. The current
+		 * behavior is to retry for retry_count and return with a warning
+		 * if that's crossed.
+		 *
+		 * If required, in future, we could take an additional option to decide
+		 * if the caller wants to retry indefinitely or return with a warning.
+		 * But the current behavior based on the "blocking" argument is ok for
+		 * now.
+		 */
+		unsigned int retry_count = 1000;
```
erimatnor, line 76: Perhaps it makes sense make the max retry count a function parameter?

src/debug_wait.c
```diff
@@ -44,23 +45,64 @@ ts_debug_waitpoint_init(DebugWait *waitpoint, const char *tagname)
...
+
+			if (retry_count == 0)
+			{
+				elog(WARNING, "timeout while acquiring shared lock");
+				return;
+			}
+			retry_count--;
+
+			/* retry after some time */
+			pg_usleep(100L);
```
erimatnor, line 94: Maybe this should also be a function parameter like the retry count. (I guess we could still use a macro that uses default constants.

src/debug_wait.c
```diff
@@ -44,23 +45,64 @@ ts_debug_waitpoint_init(DebugWait *waitpoint, const char *tagname)
...
+		 * If required, in future, we could take an additional option to decide
+		 * if the caller wants to retry indefinitely or return with a warning.
+		 * But the current behavior based on the "blocking" argument is ok for
+		 * now.
+		 */
+		unsigned int retry_count = 1000;
+
+		/* try to acquire the lock without waiting. */
+		lock_acquire_result = LockAcquire(&waitpoint->tag, ShareLock, true, true);
+		while (lock_acquire_result == LOCKACQUIRE_NOT_AVAIL)
```
erimatnor, line 80: Do we need a `CHECK_FOR_INTERRUPTS()` in this loop to ensure we break out if canceled?

tsl/test/sql/remote_txn.sql
```diff
@@ -116,8 +138,10 @@ FROM _timescaledb_internal.show_connection_cache() ORDER BY 1,4;
 --and be rolled back with no unresolved state
 BEGIN;
     SELECT remote_node_killer_set_event('pre-commit', 'loopback');
+    SELECT debug_waitpoint_enable('remote_conn_xact_end');
```
erimatnor, line 141: Can we make enabling the waitpoint part of `remote_node_killer_set_event`? These two functions are always paired together.

Similarly, can the node killer release the waitpoint automatically after it completes? Then we wouldn't have to add these extra calls.

gayyappan (approved):

src/debug_wait.c
nikkhils: Ok, using a `do..while` loop now.

src/debug_wait.c
nikkhils: We could, but then all the callers of `DEBUG_WAITPOINT` would need to provide that and it will be code churn. Given that all of this is just for testing purposes, I am inclined to keep it as simple as possible.

src/debug_wait.c
nikkhils: same comment as above. As you said we can pass defaults anyways via the macros. So unless the actual calling sites need to specify something that makes sense, I don't see the value add of churning the function signature.

src/debug_wait.c
nikkhils: `CHECK_FOR_INTERRUPTS()` is not useful if you are deep down inside CommitTransaction code path. That's the reason I added a comment on top of this function. And that is also the reason why I check `proc_exit_inprogress` explicitly in this loop.

tsl/test/sql/remote_txn.sql
nikkhils: The waitpoint has to be released only after transaction completion. We cannot release it earlier because then we won't be able to make the remote backend wait inside the committransaction code paths as desired.

erimatnor: Isn't `XACT_EVENT_COMMIT` "after" (or during) commit? So, it would be a good place to also release the waitpoint?

erimatnor (approved):

tsl/test/sql/remote_txn.sql
nikkhils: The idea is for the remote backend to wait deep down in the committransaction.

PR #3028, opened 2021-03-15
Add update test for PG13

svenklemm:


mkindahl (approved):

gayyappan (approved):

scripts/test_updates_pg13.sh
```diff
@@ -0,0 +1,16 @@
...
+
+SCRIPT_DIR=$(dirname $0)
+echo $SCRIPT_DIR
+
+TAGS="2.1.0-pg13"
+TEST_VERSION="v7"
+
+TAGS=$TAGS TEST_VERSION=$TEST_VERSION bash ${SCRIPT_DIR}/test_updates.sh "$@"
+EXIT_CODE=$?
+if [ $EXIT_CODE -ne 0 ]; then
```
pmwkaa, line 14: Do we still need this check? since we use `set -e` above which suppose to stop script execution in case if command return non zero status

PR #3029, opened 2021-03-15
Switch CI to Ubuntu 20.04

svenklemm:


pmwkaa (approved):

k-rus (approved):

.github/workflows/code_style.yaml
```diff
@@ -12,7 +12,7 @@ on:
 jobs:
   misc_checks:
     name: Check formatting, license and git hooks
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
pmwkaa, line 15: We can't set `ubuntu-latest` here as well?

scripts/gh_matrix_builder.py
```diff
@@ -51,7 +51,7 @@ def build_debug_config(overrides):
     "coverage": True,
     "llvm_config": "llvm-config-9",
     "clang": "clang-9",
-    "os": "ubuntu-18.04",
+    "os": "ubuntu-20.04",
```
pmwkaa, line 54: Is it possible that `ubuntu-20.04` will not match `ubuntu-latest` in such case?

k-rus (commented):
Can you add commit message and explain the fix, i.e., why it is appropriate to use latest in some cases and use the specific version in other cases? In particular, I am wonder if new version of ubuntu is expected to work in the tests, which use latest.

scripts/gh_matrix_builder.py
svenklemm: atm it matches ubuntu-latest but once it switches the llvm and clang versions will probably need to be adjusted thats why i pinned all jobs that directly build stuff in ubuntu to a specific version but changed the ones that run the build processes in docker to ubuntu-latest cause they dont rely on any packages from the environment they run in

svenklemm:
> Can you add commit message and explain the fix, i.e., why it is appropriate to use latest in some cases and use the specific version in other cases? In particular, I am wonder if new version of ubuntu is expected to work in the tests, which use latest.

Yes those I changed to ubuntu-latest are expected to work when latest changes cause they dont have dependencies on the environment they run in cause they run actions in docker containers

PR #3045, opened 2021-03-22
Make data_fetcher solo test to fix failing tests

erimatnor:
The `data_fetcher` test currently causes tests to fail regularly when
run in parallel. Therefore, this change moves `data_fetcher` to solo
tests.


svenklemm (approved):

mkindahl (approved):
LGTM and all tests are passing.

PR #3053, opened 2021-03-24
Fix gapfill/hashagg planner interaction

### LINKED ISSUE #3048: time_bucket_gapfill is skipping some buckets
 - OS: TimescaleDB Docker image
 - Installation method: Docker

**Describe the bug**
`time_bucket_gapfill` is not filling gaps properly anymore.

Screenshot 2021-03-22 at 22 47 49

**Additional information**
We upgraded from 1.7.1 to 2.1.0 recently on this node. It was working before the upgrade.

When not using any aggregate functions in the `SELECT`, it appears to work fine.

svenklemm:
The hashagg optimization adds a hashagg plan with modified costs
to the list of paths. This happens after gapfill pathes have been
created so those newly created pathes would miss the GapFill node.
If those created pathes would turn out to be cheaper than other
pathes GapFill would fail to work as no GapFill node would be
executed.
This patch changes hashagg path creation to skip adding pathes
when there is a GapFill path.

Fixes #3048

### DIFF
--- a/src/plan_add_hashagg.c
+++ b/src/plan_add_hashagg.c
@@ -4,14 +4,15 @@
  * LICENSE-APACHE for a copy of the license.
  */
 #include <postgres.h>
+#include <catalog/namespace.h>
+#include <miscadmin.h>
+#include <nodes/extensible.h>
 #include <nodes/plannodes.h>
-#include <parser/parsetree.h>
 #include <optimizer/clauses.h>
 #include <optimizer/pathnode.h>
 #include <optimizer/paths.h>
 #include <optimizer/tlist.h>
-#include <catalog/namespace.h>
-#include <miscadmin.h>
+#include <parser/parsetree.h>
 
 #include "compat-msvc-enter.h"
 #include <optimizer/cost.h>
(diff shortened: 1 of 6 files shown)


k-rus (dismissed):
LGTM

Can you use the same constant during the creation of the custom path too? It will be in nodes/gapfill/planner.c.

src/plan_add_hashagg.c
```diff
@@ -24,6 +25,8 @@
 #include "guc.h"
 #include "estimate.h"
 
+#define GAPFILL_PATH_NAME "GapFill"
```
k-rus, line 28: Can this constant be shared with Custom Node creation and used there too?

src/plan_add_hashagg.c
```diff
@@ -130,6 +133,14 @@ ts_plan_add_hashagg(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo *output
 	if (parse->groupingSets || !parse->hasAggs || parse->groupClause == NIL)
 		return;
 
+	/* Don't add HashAgg path if this is a gapfill query */
+	if (IsA(linitial(output_rel->pathlist), CustomPath))
```
erimatnor, line 137: It would be nice to put this whole clause in a function or macro `is_gapfill_path()`

erimatnor (approved):

src/plan_add_hashagg.c
svenklemm: ideally yes but that requires some more refactoring because gapfill is tsl and we need this check to be in apache code

k-rus (approved):

mkindahl (approved):

PR #3058, opened 2021-03-25
Reduce memory usage for distributed inserts

### LINKED ISSUE #3049: OOM killer triggered when creatng a distributed hyper table of 90M rows
 - OS: Azure Kubernetes - running docker images
 - PostgreSQL 12.6
 - TimescaleDB 2.1
 - Installation method: using timescaledev multinode

**Describe the bug**
Importing the Nyc Taxi data for 2019 and 2020 lead to OOM killer being triggered

**To Reproduce**
TimescaleDb is installed in a multi node env 3 AKS Nodes, running 1 access node and 3 data nodes
I tried to import the nyc data limiting to 2020 and 2019 with yellow and green cab only comming from 
https://github.com/toddwschneider/nyc-taxi-data

Every thing works fine when data is imported to a regular postgres table

Then I create an HyperTable based on pickup datetime, copying data from the regular trips data to the hypertable works fine too

If I create a distributed hypertable using pickup datetime for tmie serie and id as a distribution key, the copy of data (using insert into select) consumes the whole node memory and lead to a 137 error code in psql and a restart of the access node

Trips table contains 90M rows
If I copy a first 10M rows, the postgres process server side consumes up to 4Gb of RAM but does not give it back once insert is processed and as such a second insert statement lead to OOM killer freeing memory

**Expected behavior**
I would expect the distributed table to works as a regular hypertable allowing to spread the load of reading and writing accross multiple nodes

**Actual behavior**
OOM Killler retarting acces node and preventing data from being inserted

erimatnor:
The data node dispatcher used a per-tuple memory context that wasn't
automatically reset as part of each iteration of the execution
loop. This lead to linearly growing memory usage over a transaction
that inserted a lot of data.

This change switches to using the right per-tuple memory context and
also introduces a new "batch" memory context that maintains the batch
data until it is fully processed and the next batch is read. These
changes mean that memory usage never grows higher than what is needed
for one batch, which will avoid out-of-memory situations.

Fixes #3049

### DIFF
--- a/tsl/src/data_node_dispatch.c
+++ b/tsl/src/data_node_dispatch.c
@@ -156,16 +156,17 @@ typedef struct DataNodeDispatchState
 	DeparsedInsertStmt stmt; /* Partially deparsed insert statement */
 	const char *sql_stmt;	/* Fully deparsed insert statement */
 	TupleFactory *tupfactory;
-	List *target_attrs;		 /* The attributes to send to remote data nodes */
-	List *responses;		 /* List of responses to process in RETURNING state */
-	HTAB *nodestates;		 /* Hashtable of per-nodestate (tuple stores) */
-	MemoryContext mcxt;		 /* Memory context for per-node state */
-	int64 num_tuples;		 /* Total number of tuples flushed each round */
-	int64 next_tuple;		 /* Next tuple to return to the parent node when in
-							  * RETURNING state */
-	int replication_factor;  /* > 1 if we replicate tuples across data nodes */
-	StmtParams *stmt_params; /* Parameters to send with statement. Format can be binary or text */
-	int flush_threshold;	 /* Batch size used for this dispatch state */
+	List *target_attrs;		  /* The attributes to send to remote data nodes */
+	List *responses;		  /* List of responses to process in RETURNING state */
+	HTAB *nodestates;		  /* Hashtable of per-nodestate (tuple stores) */
+	MemoryContext mcxt;		  /* Memory context for per-node state */
+	MemoryContext batch_mcxt; /* Memory context for batches of data */
+	int64 num_tuples;		  /* Total number of tuples flushed each round */
+	int64 next_tuple;		  /* Next tuple to return to the parent node when in
+							   * RETURNING state */
+	int replication_factor;   /* > 1 if we replicate tuples across data nodes */
+	StmtParams *stmt_params;  /* Parameters to send with statement. Format can be binary or text */
+	int flush_threshold;	  /* Batch size used for this dispatch state */
 	TupleTableSlot *batch_slot; /* Slot used for sending tuples to data
 								 * nodes. Note that this needs to be a
 								 * MinimalTuple slot, so we cannot use the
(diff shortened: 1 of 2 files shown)


erimatnor:
**Memory usage without the change:**

dist-insert-memory-usage-bad

**Memory usage with the change:**
dist-insert-memory-usage-good2

k-rus:
@erimatnor Does it fix OOM, so it cannot happened? Or does it reduce chance of getting it, since it is less likely to run out of memory after the fix? It is not clear from the commit message.

erimatnor:
> @erimatnor Does it fix OOM, so it cannot happened? Or does it reduce chance of getting it, since it is less likely to run out of memory after the fix? It is not clear from the commit message.

It fixes the referenced issue. Memory will no longer grow indefinitely in a single transaction. I posted screenshot to show the behavior.

Whether you OOM or not can also depend on your workload, OS, and PostgreSQL settings.

I nevertheless tried to clarify in the description.

pmwkaa (approved):
Looks good. I am curious if we can use any kind of PG internal introspection for memory pools for testing

tsl/src/data_node_dispatch.c
```diff
@@ -156,16 +156,17 @@ typedef struct DataNodeDispatchState
...
+	List *responses;		  /* List of responses to process in RETURNING state */
+	HTAB *nodestates;		  /* Hashtable of per-nodestate (tuple stores) */
+	MemoryContext mcxt;		  /* Memory context for per-node state */
+	MemoryContext batch_mcxt; /* Memory context for batches of data */
+	int64 num_tuples;		  /* Total number of tuples flushed each round */
+	int64 next_tuple;		  /* Next tuple to return to the parent node when in
+							   * RETURNING state */
+	int replication_factor;   /* > 1 if we replicate tuples across data nodes */
+	StmtParams *stmt_params;  /* Parameters to send with statement. Format can be binary or text */
+	int flush_threshold;	  /* Batch size used for this dispatch state */
```
mkindahl, line 169: Nit: Doesn't seem necessary and also create a lot of different indents for the comments (one more than the one before and one less than the one after).

tsl/src/data_node_dispatch.c
```diff
@@ -331,6 +332,7 @@ data_node_dispatch_begin(CustomScanState *node, EState *estate, int eflags)
 	sds->flush_threshold = intVal(list_nth(cscan->custom_private, CustomScanPrivateFlushThreshold));
 
 	sds->mcxt = mcxt;
```
mkindahl, line 334: Maybe you want to call this `stmt_mcxt` to make the usage clear and distinguish it from the batch context.

mkindahl (approved):
LGTM

tsl/src/data_node_dispatch.c
erimatnor: This is `make format`'s doing, so is needed to pass formatting checks.

tsl/src/data_node_dispatch.c
erimatnor: Why do you suggest `stmt_mcxt` specifically. This is the memory context used for executor state, not only the statement.

PR #3076, opened 2021-03-31
Refactor chunk creation for future extension

k-rus:
Separates chunk preparation and metadata update. Separates preparation
of constraints names, since there is no overlap between preparing
names for dimension constraints and other constraints. Factors out 
creation of json string describing dimension slices of a chunk.

This refactoring is preparation for implementing new functionalities.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -209,7 +209,7 @@ ts_chunk_primary_dimension_end(const Chunk *chunk)
 }
 
 static void
-chunk_insert_relation(Relation rel, Chunk *chunk)
+chunk_insert_relation(Relation rel, const Chunk *chunk)
 {
 	HeapTuple new_tuple;
 	CatalogSecurityContext sec_ctx;
@@ -224,7 +224,7 @@ chunk_insert_relation(Relation rel, Chunk *chunk)
 }
 
 void
-ts_chunk_insert_lock(Chunk *chunk, LOCKMODE lock)
+ts_chunk_insert_lock(const Chunk *chunk, LOCKMODE lock)
 {
 	Catalog *catalog = ts_catalog_get();
 	Relation rel;
@@ -862,7 +862,7 @@ ts_chunk_create_table(Chunk *chunk, Hypertable *ht, const char *tablespacename)
 }
 
 static List *
-chunk_assign_data_nodes(Chunk *chunk, Hypertable *ht)
+chunk_assign_data_nodes(Chunk *chunk, const Hypertable *ht)
 {
 	List *htnodes;
 	List *chunk_data_nodes = NIL;
@@ -921,7 +921,10 @@ ts_chunk_get_data_node_name_list(const Chunk *chunk)
 }
 
 /*
- * Create a chunk from the dimensional constraints in the given hypercube.
+ * Create a chunk object from the dimensional constraints in the given hypercube.
+ *
+ * The chunk object is then used to create the actual chunk table and update the
+ * metadata separately.
  *
  * The table name for the chunk can be given explicitly, or generated if
  * table_name is NULL. If the table name is generated, it will use the given
@@ -930,11 +933,11 @@ ts_chunk_get_data_node_name_list(const Chunk *chunk)
  * the chunk.
  */
 static Chunk *
-chunk_create_metadata_after_lock(Hypertable *ht, Hypercube *cube, const char *schema_name,
-								 const char *table_name, const char *prefix)
+chunk_create_object(const Hypertable *ht, Hypercube *cube, const char *schema_name,
+					const char *table_name, const char *prefix)
 {
-	Hyperspace *hs = ht->space;
-	Catalog *catalog = ts_catalog_get();
+	const Hyperspace *hs = ht->space;
+	const Catalog *catalog = ts_catalog_get();
 	CatalogSecurityContext sec_ctx;
 	Chunk *chunk;
 	const char relkind = hypertable_chunk_relkind(ht);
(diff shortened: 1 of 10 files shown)


src/chunk.c
```diff
@@ -930,8 +932,8 @@ ts_chunk_get_data_node_name_list(const Chunk *chunk)
  * the chunk.
  */
 static Chunk *
-chunk_create_metadata_after_lock(Hypertable *ht, Hypercube *cube, const char *schema_name,
-								 const char *table_name, const char *prefix)
+chunk_create_object(const Hypertable *const ht, Hypercube *const cube, const char *schema_name,
```
erimatnor, line 935: Out of curiosity, why do you bother with `const` pointers?

I do get why it makes sense to have `const` objects in the parameter list since it signals no side-effects on these objects. But `const` pointers seem almost more trouble than it is worth, especially the way it is applied here.

A couple of observations and thoughts on this:

* Since the pointer is passed by value it has no side effects to set it in the function. 
* Often the use of `const` pointers is inconsistent. For instance, why isn't at least `table_name` also a `const` pointer when you applied it to most of the other parameters (it doesn't seem to be set in the function). Another example: In the parameter list of `chunk_insert_into_metadata_after_lock` you made the object `const` but not the pointer.
* Why does this pattern seem to apply to only function parameters but not, e.g., function variable pointers? If it is really important to not change pointers, it should apply also to pointer variables and pointers to non-const objects. Am I missing something?
* Personally, I don't find a whole lot of value in making pointers `const`; in fact, it mostly makes the code much more verbose. I do think we should be better at making objects `const` though.
* I can't remember having seen any C projects/code that use `const` pointers this way, probably for the reasons I listed.

Interested to hear your thoughts on this. We can discuss separately, but just wanted to leave my thoughts on this while in context.

src/chunk.c
```diff
@@ -921,7 +921,9 @@ ts_chunk_get_data_node_name_list(const Chunk *chunk)
 }
 
 /*
- * Create a chunk from the dimensional constraints in the given hypercube.
+ * Create a chunk object from the dimensional constraints in the given hypercube.
+ *
+ * The chunk object is then used to create actual chunk table.
```
erimatnor, line 926:
```suggestion
 * The chunk object is then used to create the actual chunk table. Note that this function
 * does not persist the chunk object. The metadata for the chunk object is created
 * in a separate step.
```

src/chunk.c
```diff
@@ -970,34 +972,25 @@ chunk_create_metadata_after_lock(Hypertable *ht, Hypercube *cube, const char *sc
...
 	if (chunk->relkind == RELKIND_FOREIGN_TABLE)
 		chunk->data_nodes = chunk_assign_data_nodes(chunk, ht);
 
 	return chunk;
 }
 
-static Oid
-chunk_create_table_after_lock(Chunk *chunk, Hypertable *ht)
+static void
+chunk_insert_into_metadata_after_lock(const Chunk *chunk)
```
erimatnor, line 982: const object, but not const pointer, which seems inconsistent (see separate comment on this topic).

Also, maybe this name:

```suggestion
/*
* Persist a chunk object by writing it to metadata tables.
*/
static void
chunk_persist_after_lock(const Chunk *chunk)
```

src/chunk.c
```diff
@@ -970,34 +972,25 @@ chunk_create_metadata_after_lock(Hypertable *ht, Hypercube *cube, const char *sc
...
+	ts_chunk_insert_lock(chunk, RowExclusiveLock);
 
-	if (!OidIsValid(chunk->table_id))
-		elog(ERROR, "could not create chunk table");
+	/* Add metadata for dimensional and inheritable constraints */
+	ts_chunk_constraints_insert_metadata(chunk->constraints);
+}
 
+static void
+chunk_create_table_constraints(Chunk *chunk)
```
erimatnor, line 992: Can the object be `const` here (or won't `ts_trigger_create_all_on_chunk` allow it)? I guess at least the pointer could be `const`, if that's important.

```suggestion
chunk_create_table_constraints(const Chunk *chunk)
```

src/chunk.c
```diff
@@ -1013,6 +1006,18 @@ chunk_create_table_after_lock(Chunk *chunk, Hypertable *ht)
...
+
+static Oid
+chunk_create_table(Chunk *chunk, Hypertable *ht)
+{
+	/* Create the actual table relation for the chunk */
+	const char *tablespace = ts_hypertable_select_tablespace_name(ht, chunk);
+
+	chunk->table_id = ts_chunk_create_table(chunk, ht, tablespace);
+
+	if (!OidIsValid(chunk->table_id))
```
erimatnor, line 1019: If this failure case is even possible, it might make sense to move this into `ts_chunk_create_table` immediately after`DefineRelation`. It isn't clear that that function can even return `InvalidOid`, so we might be able to get rid of this check altogether.

erimatnor (approved):
LGTM. Just a bunch of nits and suggestions.

src/chunk.c
k-rus: The second const means that the variable cannot be reused for another pointer of the same type.

src/chunk_constraint.c
```diff
@@ -80,31 +80,29 @@ chunk_constraints_expand(ChunkConstraints *ccs, int16 new_capacity)
...
-	{
-		char constrname[100];
-		CatalogSecurityContext sec_ctx;
+	snprintf(NameStr(*dst), NAMEDATALEN, "constraint_%d", dimension_slice_id);
+}
+
+static void
+chunk_constraint_choose_name(Name dst, const char *hypertable_constraint_name, int32 chunk_id)
+{
+	char constrname[100];
```
svenklemm, line 91: why is this hardcoded 100, shouldnt this be NAMEDATALEN anything past that would get truncated anyway and might even create unrecoverable errors if we start creating names that get truncated because identifier names are no longer unique and chunk creation would fail as a result

k-rus: It is a move of the original code, which contained this magic number. I can look into fixing it, but it is not something I introduced.

src/chunk.c
svenklemm: I agree with erik here this just adds line noise and additional development overhead with little practical benefit

k-rus: It is not a noise. In general, it provides semantics how the variable will be used in the code. I.e., that it cannot be reused to store different value. Thus I personally prefer to see `const` more often, so it is clear when to expect variable reuse.
There is a drawback that it doesn't play well with old Microsoft compilers.

src/chunk.c
```diff
@@ -970,34 +972,25 @@ chunk_create_metadata_after_lock(Hypertable *ht, Hypercube *cube, const char *sc
...
-	/* Insert any new dimension slices */
-	ts_dimension_slice_insert_multi(cube->slices, cube->num_slices);
-
-	/* Add metadata for dimensional and inheritable constraints */
-	chunk_add_constraints(chunk);
-	ts_chunk_constraints_insert_metadata(chunk->constraints);
-
-	/* If this is a remote chunk we assign data nodes */
 	if (chunk->relkind == RELKIND_FOREIGN_TABLE)
 		chunk->data_nodes = chunk_assign_data_nodes(chunk, ht);
```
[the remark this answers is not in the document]
k-rus: No, this will only assign data nodes into the `chunk` object. The metadata is not be modified at this phase.

src/chunk.c
k-rus: > * why isn't at least `table_name` also a `const` pointer when you applied it to most of the other parameters (it doesn't seem to be set in the function).

@erimatnor `table_name` is set in the function when it is not provided. So this actually demonstrates that it is good to be explicit about using `const` as you missed to find the usage in the code.

src/chunk.c
k-rus: It doesn't allow to use `const` here due to string conversion by a PG function deeply insde..

src/chunk.c
k-rus: I prefer the original name as it catches me better than `persist` and thus doesn't require a comment.

src/chunk.c
k-rus: Since this function is long, I find it useful to know about variables reuse. I haven't changed other functions as they are simple enough.

src/chunk.c
k-rus: Improved the comment.

src/chunk.c
k-rus: Correct, it should not be possible. Changed to assertion for now.

src/chunk_constraint.c
k-rus: `NAMEDATALEN` should be correct here. So changed to it.

pmwkaa (approved):

svenklemm (approved):

PR #3083, opened 2021-04-03
Keep Append pathkeys in ChunkAppend

### LINKED ISSUE #3030: ChunkAppend does not propagate pathkeys when subpath is Append node
 - OS: CentOS7.6
 - PostgreSQL version (output of `postgres --version`): 12 & 13
 - Installation method: source

**Describe the bug**
When there's a hypertable that has just one child and a query on it is done that uses an ORDER BY with LIMIT 1 that matches some index, let's say:

```
set random_page_cost=1;
create table t1 (a timestamptz, b int, c int);
select create_hypertable('t1', 'a', chunk_time_interval=>interval '1d', create_default_indexes=>false);
create index on t1 (b,a);
insert into t1 (a,b) select (current_date+make_interval(mins=>b))::timestamptz, a from generate_series (1,30) a, generate_series(60*12,60*15) b;
analyze t1;
```

```
explain select * from t1 where a between now()-interval '5d' and now() + interval '2d' order by b,a limit 1;
```

We should end up with a plan like:

```
Limit
    Index Scan on chunk
```

Internal, during planning, Postgres still has an Append or MergeAppend node. This is only removed as one of the last steps during planning. Thus, when it arrives at `apply_optimizations` function in Timescale hook, it looks more something like this:

```
Limit
    Append
        Index Scan on chunk
```

Note the Append node here. Postgres does detect that there is only one child of the hypertable, so it chooses an Append instead of a MergeAppend (there's only one child, so there's nothing to merge). If we had two chunks, the plan would look more like this instead:

```
Limit
    MergeAppend
        Index Scan on chunk1
        Index Scan on chunk2
```

TimescaleDb in `apply_optimizations` has code which calls `should_chunk_append`. This detects that there's an Append subpath, that there's some mutable functions due to the usage of `now()`, so it decides to pick the ChunkAppend.
The problem here is that ChunkAppend does not maintain the ordering for such a node. Therefore, a ChunkAppend node is inserted, but Postgres then after that needs to put a Sort node. This is hugely inefficient.

```
Limit  (cost=189.03..189.03 rows=1 width=16)
  ->  Sort  (cost=189.03..202.60 rows=5429 width=16)
        Sort Key: t1.b, t1.a
        ->  Custom Scan (ChunkAppend) on t1  (cost=0.29..161.88 rows=5429 width=16)
              ->  Index Scan using _hyper_558_55769_chunk_t1_b_a_idx on _hyper_558_55769_chunk  (cost=0.29..161.88 rows=5429 width=16)
                    Index Cond: ((a >= (now() - '5 days'::interval)) AND (a <= (now() + '2 days'::interval)))
```

Turning off `enable_chunk_append` leads to the much faster plan:

```
Limit  (cost=0.29..0.32 rows=1 width=16)
  ->  Index Scan using _hyper_558_55769_chunk_t1_b_a_idx on _hyper_558_55769_chunk  (cost=0.29..161.88 rows=5429 width=16)
        Index Cond: ((a >= (now() - '5 days'::interval)) AND (a <= (now() + '2 days'::interval)))
```

There are two ways in which this can be fixed:
- Disable ChunkAppend when you know there is only 1 child. ConstraintAwareAppend actually has such logic in `ts_constraint_aware_append_possible`. If this logic is put in `should_chunk_append` it fixes my issue. The ChunkAppend node is not chosen at all now.
- ChunkAppend could also be made aware of the pathkeys of the Append node, therefore telling Postgres that the order is maintained. In `ts_chunk_append_path_create()` the pathkeys of the Append subpath can be copied to the ChunkAppend path. This works for my particular query and is probably a nicer solution, but there may be side-effects here I'm unaware of. I haven't looked deep into this option. With this fix, the planner will choose the ChunkAppend node, however it will not put the Sort node on top of it.

I realize this is a bit of an edge case in planning that is probably quite rare to hit in real life. However, it's leading to some very unpredictable plans for us as the planner behaves so different for tables with just one child compared to tables with more than one child. The plan that is chosen with the Sort on top is so inefficient for large tables that it can be really annoying when the planner chooses this (sorting the whole table rather than just picking the first row from the index).

svenklemm:
This changes ChunkAppend to always keep the pathkeys from the original
path because the original path was either a MergeAppendPath and this
will become an ordered append or the original path is an AppendPath
and since we do not reorder children the order will be kept intact.
For the AppendPath case with pathkeys it was most likely an Append
with only a single child. We could skip the ChunkAppend path creation
if there is only a single child but we decided earlier that ChunkAppend
would be beneficial for this query so we treat it the same as if it
had multiple children.

Fixes #3030

### DIFF
--- a/test/sql/include/append_query.sql
+++ b/test/sql/include/append_query.sql
@@ -346,3 +346,16 @@ LEFT JOIN i2661 ht ON
   (FLOOR(EXTRACT (EPOCH FROM ht."timestamp") / 300) * 300 = EXTRACT (EPOCH FROM ts.timestamp))
   AND ht.timestamp > '2019-12-30T00:00:00Z'::timestamp;
 
+-- #3030 test chunkappend keeps pathkeys when subpath is append
+-- on PG11 this will not use ChunkAppend but MergeAppend
+SET enable_seqscan TO FALSE;
+CREATE TABLE i3030(time timestamptz NOT NULL, a int, b int);
+SELECT table_name FROM create_hypertable('i3030', 'time', create_default_indexes=>false);
+CREATE INDEX ON i3030(a,time);
+INSERT INTO i3030 (time,a) SELECT time, a FROM generate_series('2000-01-01'::timestamptz,'2000-01-01 3:00:00'::timestamptz,'1min'::interval) time, generate_series(1,30) a;
+ANALYZE i3030;
+
+:PREFIX SELECT * FROM i3030 where time BETWEEN '2000-01-01'::text::timestamptz AND '2000-01-03'::text::timestamptz ORDER BY a,time LIMIT 1;
+DROP TABLE i3030;
+RESET enable_seqscan;
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,10 @@ accidentally triggering the load of a previous DB version.**
 * #2989 Refactor and harden size and stats functions
 * #3058 Reduce memory usage for distributed inserts
 * #3067 Fix extremely slow multi-node order by queries
+* #3083 Keep Append pathkeys in ChunkAppend
 
 **Thanks**
+* @fvannee for reporting an issue with ChunkAppend pathkeys
 * @pedrokost and @RobAtticus for reporting an issue with size
   functions on empty hypertables
 * @stephane-moreau for reporting an issue with high memory usage during
--- a/src/chunk_append/chunk_append.c
+++ b/src/chunk_append/chunk_append.c
@@ -57,6 +57,21 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
 	path->cpath.path.pathtarget = rel->reltarget;
 	path->cpath.path.param_info = subpath->param_info;
 
+	/*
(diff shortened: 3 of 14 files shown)


test/sql/include/append_query.sql
```diff
@@ -346,3 +346,15 @@ LEFT JOIN i2661 ht ON
   (FLOOR(EXTRACT (EPOCH FROM ht."timestamp") / 300) * 300 = EXTRACT (EPOCH FROM ts.timestamp))
   AND ht.timestamp > '2019-12-30T00:00:00Z'::timestamp;
 
+-- #3030 test chunkappend keeps pathkeys when subpath is append
+-- on PG11 this will not use ChunkAppend but MergeAppend
+SET random_page_cost=1;
```
gayyappan, line 351: I think it it better to just enable_indexscan and disable seqscan. The intent is clearer.
Need to reset at the end of this test case.

gayyappan (approved):
Comment about parameter in test case. Prefer use of enable_indexscan/seqscan versus setting page costs.

test/sql/include/append_query.sql
svenklemm: changed

mkindahl (approved):

Issue #3093, opened 2021-04-07
dist_restore_point test is flaky


k-rus:
`dist_restore_point` test fails rather often in cron jobs. The diff from [the last failure[(https://github.com/timescale/timescaledb/runs/2281680997?check_suite_focus=true#step:16:103):
```diff
--- /home/runner/work/timescaledb/timescaledb/tsl/test/isolation/expected/dist_restore_point.out	2021-04-06 20:09:19.682309040 +0000
+++ /home/runner/work/timescaledb/timescaledb/build/tsl/test/isolation/results/dist_restore_point.out	2021-04-06 20:13:12.372284002 +0000
@@ -323,8 +323,6 @@
 
                
 step s1_create_dist_rp: SELECT restore_point > pg_lsn('0/0') as valid_lsn FROM create_distributed_restore_point('s1_test'); <waiting ...>
-s2: NOTICE:  the number of partitions in dimension "device" was decreased to 3
-DETAIL:  To make efficient use of all attached data nodes, the number of space partitions was set to match the number of data nodes.
 step s2_del_dn: SELECT * FROM delete_data_node('data_node_4'); <waiting ...>
 step s3_lock_count: 
 	SELECT waitpoint_locks('create_distributed_restore_point_lock') as cdrp_locks,
```

PR #3096, opened 2021-04-08
Fix dist_restore_point test on PG11.0

pmwkaa:
The test fails on PG11.0 since there is no notice during data node deletion. Reduce client notice level to warning, to make it compatible with other versions.

Fixes https://github.com/timescale/timescaledb/issues/3093


k-rus (approved):
Can you post the PR message into the commit message too, so it explains the commit?
Otherwise looks good 👍

svenklemm (approved):

PR #3107, opened 2021-04-12
Fix Skip Scan cost calculation for queries with excluded chunks

svenklemm:
If the row estimate for the scan below Skip Scan is 1 we assume that the
estimate got clamped to 1 and no rows would be returned by this scan and
this chunk will most likely be excluded by runtime exclusion.
Otherwise the cost for this path would be highly inflated due to
(ndistinct / rows) * total leading to SkipScan not being chosen for
queries on hypertables with a lot of excluded chunks.


nikkhils (approved):
LGTM

pmwkaa (approved):

PR #3112, opened 2021-04-13
Use %u to format Oid instead of %d

svenklemm:
Since Oid is unsigned int we have to use %u to print it otherwise
oids >= 2^31 will not work correctly.

### DIFF
--- a/src/hypertable_restrict_info.c
+++ b/src/hypertable_restrict_info.c
@@ -8,6 +8,7 @@
 #include <utils/lsyscache.h>
 #include <parser/parsetree.h>
 #include <utils/array.h>
+#include <utils/builtins.h>
 
 #include "compat.h"
 #if PG12_LT
@@ -442,7 +443,9 @@ dimension_values_create_from_array(Const *c, bool user_or)
 	/* it's an array type, lets get the base element type */
 	base_el_type = get_element_type(c->consttype);
 	if (base_el_type == InvalidOid)
-		elog(ERROR, "Couldn't get base element type from array type: %d", c->consttype);
+		elog(ERROR,
+			 "invalid base element type for array type: \"%s\"",
+			 format_type_be(c->consttype));
 
 	return dimension_values_create(values, base_el_type, user_or);
 }
--- a/src/plan_agg_bookend.c
+++ b/src/plan_agg_bookend.c
@@ -52,7 +52,9 @@
 #include <parser/parse_clause.h>
 #include <parser/parse_func.h>
 #include <rewrite/rewriteManip.h>
+#include <utils/builtins.h>
 #include <utils/lsyscache.h>
+#include <utils/regproc.h>
 #include <utils/syscache.h>
 #include <utils/typcache.h>
 
@@ -484,9 +486,9 @@ find_first_last_aggs_walker(Node *node, List **context)
 			get_opfamily_member(sort_tce->btree_opf, sort_oid, sort_oid, func_strategy->strategy);
 		if (aggsortop == InvalidOid)
 			elog(ERROR,
-				 "Can't resolve sort operator oid for function oid: %d and type: %d",
-				 aggref->aggfnoid,
-				 sort_oid);
+				 "Cannot resolve sort operator for function \"%s\" and type \"%s\"",
+				 format_procedure(aggref->aggfnoid),
+				 format_type_be(sort_oid));
 
 		/* Used in projection */
 		value = (TargetEntry *) linitial(aggref->args);
--- a/src/time_bucket.c
+++ b/src/time_bucket.c
@@ -5,11 +5,12 @@
  */
 #include <postgres.h>
 #include <catalog/pg_type.h>
-#include <utils/timestamp.h>
-#include <utils/datetime.h>
-#include <utils/date.h>
-#include <utils/fmgrprotos.h>
 #include <fmgr.h>
+#include <utils/builtins.h>
+#include <utils/date.h>
+#include <utils/datetime.h>
+#include <utils/fmgrprotos.h>
+#include <utils/timestamp.h>
 
 #include "utils.h"
 #include "time_bucket.h"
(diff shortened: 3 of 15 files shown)


mkindahl (approved):

src/plan_agg_bookend.c
```diff
@@ -484,9 +486,9 @@ find_first_last_aggs_walker(Node *node, List **context)
 			get_opfamily_member(sort_tce->btree_opf, sort_oid, sort_oid, func_strategy->strategy);
 		if (aggsortop == InvalidOid)
 			elog(ERROR,
-				 "Can't resolve sort operator oid for function oid: %d and type: %d",
-				 aggref->aggfnoid,
-				 sort_oid);
+				 "Can't resolve sort operator oid for function \"%s\" and type \"%s\"",
+				 format_procedure(aggref->aggfnoid),
```
mfundul, line 490: The result is supposed to be a `palloc`ed string. Should we maybe explicitly deallocate this or is this a non issue with the current memory context?

svenklemm: It is a non-issue. All those allocations are on error pathes meaning the memory context will be reset anyway.

mfundul (approved):

pmwkaa (approved):

src/plan_agg_bookend.c
erimatnor: I guess in many cases, errors return numerical OIDs because the reason for the error was that some of those OIDs didn't correspond to actual types or objects. So, I am wondering if it is a good idea in general to turn lots of error calls into `format_...` instead of printing numerical values. I think in case of `format_procedure` it is handled by that function, but is this always the case?

This is the reason why cache lookup failures print something like:

```
cache lookup failed for attribute %d of relation %u
```
since the whole reason for the cache lookup failure likely was that the relation or attribute didn't exist.

src/time_bucket.c
```diff
@@ -283,7 +284,7 @@ ts_time_bucket_by_type(int64 interval, int64 timestamp, Oid timestamp_type)
 			bucket_function = ts_date_bucket;
 			break;
 		default:
-			elog(ERROR, "invalid time_bucket Oid %d", timestamp_type);
+			elog(ERROR, "invalid time_bucket type \"%s\"", format_type_be(timestamp_type));
```
erimatnor, line 287: This might not make sense because if the type doesn't exist, `format_type_be` will generate a generic cache lookup failure/error on its own. Is this what we want?

src/utils.c
```diff
@@ -126,7 +127,7 @@ ts_time_value_to_internal(Datum time_val, Oid type_oid)
 		if (ts_type_is_int8_binary_compatible(type_oid))
 			return DatumGetInt64(time_val);
 
-		elog(ERROR, "unknown time type OID %d", type_oid);
+		elog(ERROR, "unknown time type \"%s\"", format_type_be(type_oid));
```
erimatnor, line 130: I think there are two cases here:

1. The type exists but is not a valid time type --- all is good and this error makes sense
2. The type does not exist, and then `format_type_be` will throw another error.

In particular for the second case, is this what we want here?

I think this concern applies to a lot of other suggested changes in this PR.

svenklemm: hmm i could have sworn format_type_be returns OID when it cannot resolve but it does not and while there is a flag to make it not fail when it cannot resolve it returns '???' instead of the OID in that case. I'll revert the format_type_be changes.

src/plan_agg_bookend.c
```diff
@@ -484,9 +486,9 @@ find_first_last_aggs_walker(Node *node, List **context)
...
+				 format_type_be(sort_oid));
```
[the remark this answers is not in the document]
svenklemm: no the failure is when there is no btree opfamily member for this datatype, both datatype and function have to be valid at this point

src/time_bucket.c
svenklemm: i think the intent here is for it to error when being called with types that we don't support in time_bucket like float or decimal not that the type in general is invalid, but i dont think you can reach this part without a bug somewhere in our code

src/utils.c
svenklemm: Actually regarding your 2nd point i dont think it applies this is all internal code and not being called with random type ids but with actual oids handed to us from postgres so the types are validated but its just that its a type we do not support.

src/hypertable_restrict_info.c
```diff
@@ -442,7 +443,9 @@ dimension_values_create_from_array(Const *c, bool user_or)
 	/* it's an array type, lets get the base element type */
 	base_el_type = get_element_type(c->consttype);
 	if (base_el_type == InvalidOid)
-		elog(ERROR, "Couldn't get base element type from array type: %d", c->consttype);
+		elog(ERROR,
+			 "Couldn't get base element type from array type: \"%s\"",
```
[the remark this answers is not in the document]
svenklemm: in theory yes, array types have their own type id and they could have the type of the array elements be something invalid

src/plan_agg_bookend.c
svenklemm: actually most of these errors are around datatypes not supported with our functions not about general bogus type ids. e.g. time_bucket with decimal or with domains/custom types

tsl/src/compression/compression.c
```diff
@@ -453,7 +453,7 @@ compress_chunk_populate_sort_info_for_column(Oid table, const ColumnCompressionI
 
 	tp = SearchSysCacheAttName(table, NameStr(column->attname));
 	if (!HeapTupleIsValid(tp))
-		elog(ERROR, "table %d does not have column \"%s\"", table, NameStr(column->attname));
+		elog(ERROR, "table %u does not have column \"%s\"", table, NameStr(column->attname));
```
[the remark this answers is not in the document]
svenklemm: the table is valid at this point

PR #3118, opened 2021-04-14
Fix use after free in cache

pmwkaa:
When calling hash_search with HASH_REMOVE the returned pointer should not
be dereferenced because it returns a dangling pointer

Issue: https://github.com/timescale/timescaledb-private/issues/914

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -209,16 +209,19 @@ bool
 ts_cache_remove(Cache *cache, void *key)
 {
 	bool found;
-	void *entry;
 
-	entry = hash_search(cache->htab, key, HASH_REMOVE, &found);
-
-	if (found)
+	if (cache->remove_entry != NULL)
 	{
-		if (cache->remove_entry != NULL)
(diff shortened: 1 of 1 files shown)


src/cache.c
```diff
@@ -209,17 +209,18 @@ bool
...
 	bool found;
-	void *entry;
 
-	entry = hash_search(cache->htab, key, HASH_REMOVE, &found);
-
-	if (found)
+	if (cache->remove_entry != NULL)
 	{
-		if (cache->remove_entry != NULL)
+		void *entry = hash_search(cache->htab, key, HASH_FIND, &found);
```
svenklemm, line 215: maybe add a command why we do this here

svenklemm (approved):

erimatnor (approved):
The commit log references the wrong issue since the issue referenced is in a private repo. I suggest not referencing the private issue or move it to the public repo.

Otherwise, LGTM.

PR #3123, opened 2021-04-15
Fix crash while using REINDEX TABLE CONCURRENTLY

### LINKED ISSUE #3122: REINDEX CONCURRENTLY leads to crash
Seems to be index concurrently implementation supposed to be run without active transaction, which is being violated while iterating chunks during REINDEX.

A way to reproduce:

```
CREATE TABLE reindex_test(time timestamp, temp float, PRIMARY KEY(time, temp));
CREATE UNIQUE INDEX reindex_test_time_unique_idx ON reindex_test(time);

SELECT create_hypertable('reindex_test', 'time', chunk_time_interval => 2628000000000);

INSERT INTO reindex_test VALUES ('2017-01-20T09:00:01', 17.5),
                                ('2017-01-21T09:00:01', 19.1),
                                ('2017-04-20T09:00:01', 89.5),
                                ('2017-04-21T09:00:01', 17.1),
                                ('2017-06-20T09:00:01', 18.5),
                                ('2017-06-21T09:00:01', 11.0);

REINDEX TABLE CONCURRENTLY reindex_test;
```

PostgreSQL logs:

Single node

```
TRAP: FailedAssertion("!cp->cache->release_on_commit", File: "/timescaledb/src/cache.c", Line: 313)

309│                                 /*
310│                                  * This assert makes sure that that we don't have a cache
311│                                  * leak when running with debugging
312│                                  */
313├───────────────────────────────> Assert(!cp->cache->release_on_commit);
```

Multi node (executed on a data node)
```
TRAP: FailedAssertion("GetTopTransactionIdIfAny() == InvalidTransactionId", File: "index.c", Line: 3343)
```

pmwkaa:
Fix crash while using REINDEX TABLE CONCURRENTLY

CONCURRENTLY option of the REINDEX TABLE is not supported and
leads to a crash when using with hypertable. Block it and print
the error message.

Fix #3122

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -1478,13 +1478,18 @@ process_reindex(ProcessUtilityArgs *args)
 			{
 				PreventCommandDuringRecovery("REINDEX");
 				ts_hypertable_permissions_check_by_id(ht->fd.id);
-
+#if PG12_GE
+				if (stmt->concurrent)
+					ereport(ERROR,
+							(errmsg("concurrent index creation on hypertables is not supported")));
+#endif
 				if (foreach_chunk(ht, reindex_chunk, args) >= 0)
 					result = DDL_DONE;
 
 				add_hypertable_to_process_args(args, ht);
 			}
 			break;
+
 		case REINDEX_OBJECT_INDEX:
 			ht = ts_hypertable_cache_get_entry(hcache,
 											   IndexGetRelation(relid, true),
--- a/test/sql/CMakeLists.txt
+++ b/test/sql/CMakeLists.txt
@@ -20,7 +20,6 @@ set(TEST_FILES
   grant_hypertable.sql
   hash.sql
   histogram_test.sql
-  index.sql
   insert_many.sql
   insert_single.sql
   join.sql
@@ -66,6 +65,7 @@ set(TEST_TEMPLATES
   sort_optimization.sql.in
   sql_query.sql.in
   update.sql.in
+  index.sql.in
 )
 
 # tests that fail or are unreliable when run in parallel
@@ -74,8 +74,11 @@ set(SOLO_TESTS
   alter
   alternate_users
   bgw_launcher
-  chunk_utils.sql
-  index
+  chunk_utils
+  index-11
+  index-12
+  index-13
+  loader
   loader
   net
   pg_dump_unprivileged
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ accidentally triggering the load of a previous DB version.**
 * #3101 Use commit date in get_git_commit()
 * #3104 Fix use after free in add_reorder_policy
 * #3106 Fix use after free in chunk_api_get_chunk_stats
+* #3123 Fix crash while using REINDEX TABLE CONCURRENTLY
 
 ## 2.2.0 (2021-04-13)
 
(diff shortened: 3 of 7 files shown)


pmwkaa:
Index test on PG11 is failing, working on a template test file

svenklemm (approved):

src/process_utility.c
```diff
@@ -1478,13 +1478,22 @@ process_reindex(ProcessUtilityArgs *args)
 			{
 				PreventCommandDuringRecovery("REINDEX");
 				ts_hypertable_permissions_check_by_id(ht->fd.id);
-
+#if PG12_GE
+				if (stmt->concurrent)
+				{
+					stmt->concurrent = false;
+					ereport(NOTICE,
+							(errmsg("REINDEX TABLE CONCURRENTLY is unsupported on a hypertable"),
```
erimatnor, line 1486: It is unclear why this isn't an error. People can just drop CONCURRENTLY if they want to reindex. 

I also suggest formulating similar to this PostgreSQL error: ""concurrent index creation on system catalog tables is not supported":

```suggestion
							(errmsg("concurrent index creation on hypertables is not supported"),
```

src/process_utility.c
```diff
@@ -1478,13 +1478,22 @@ process_reindex(ProcessUtilityArgs *args)
...
+							 errdetail("Using non-concurrent implementation.")));
```
erimatnor, line 1487: Again, not sure we need this since we should probably just generate an error above.

```suggestion
							 errdetail("Performing non-concurrent reindexing instead.")));
```

src/process_utility.c
pmwkaa: I am fine to throw an error here, I've seen that PostgreSQL actually also uses regular implementation in some cases and shows a NOTICE in such cases. I guess we should decide for yourselves

erimatnor: Can you point to the place in the PostgreSQL code where they do this? 

It seems odd to me, because if PostgreSQL has implemented  `CONCURRENTLY` I don't understand why they wouldn't use it, unless they do to this bait-and-switch depending on object type.

pmwkaa: Sorry, I misunderstood when checked the code for first time, I had a different impression. Switched to produce the error as suggest. Please take a look at the updated PR info above

test/sql/CMakeLists.txt
```diff
@@ -74,8 +74,11 @@ set(SOLO_TESTS
   alter
   alternate_users
   bgw_launcher
-  chunk_utils.sql
-  index
+  chunk_utils
+  index-11
```
gayyappan, line 78: Could you clarify why we need to add this to SOLO_TESTS as well? We already have this in the templated list (line 68).

gayyappan (approved):
Question about test MakeList.

test/sql/CMakeLists.txt
svenklemm: SOLO_TESTS doesnt actually run any tests it's just a marker that those tests will not run in parallel with other tests but it still needs to be in one of the other lists to actually run

svenklemm:
Needs changelog entry.

test/sql/CMakeLists.txt
pmwkaa: Well we had this before for `index` test. Since index test made as a template, I thought it is necessary to update it here as well. I am not aware why index test was put in the SOLO tests list in the first place, likely it has something to do with tablespace

PR #3126, opened 2021-04-16
Add GitHub action for posting daily CI summary

svenklemm:
This patch adds a GitHub action that posts all failed CI jobs in
the last 24 hours with link to the failed job.

Disable-check: commit-count

### DIFF
--- a/.github/workflows/ci_summary.yaml
+++ b/.github/workflows/ci_summary.yaml
@@ -1,0 +1,23 @@
+name: CI Summary
+on:
+  schedule:
+    # run daily 8:00 CET on master branch
+    - cron: '0 7 * * *'
+jobs:
+  ci_summary:
+    if: github.repository == 'timescale/timescaledb'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: "Build message"
+      env:
+        SLACK_CHANNEL: "CEKV5LMK3"
+      run: python scripts/gh_ci_summary.py > message.json
+
+    - name: "Notify Slack"
+      uses: krider2010/slack-bot-action@1.0.1
+      env:
+        SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+        MESSAGE_FILE: message.json
+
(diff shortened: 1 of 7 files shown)


.github/workflows/ci_summary.yaml
```diff
@@ -0,0 +1,22 @@
+name: CI Summary
+on:
+  schedule:
+    # run daily 8:00 CET on master branch
+    - cron: '0 7 * * *'
+jobs:
+  build_summary:
```
erimatnor, line 7: If I am not mistaken, this will run on every fork of this repo, including our private one.

We've seen issues with this previously, where we have a lot of runs on the private repo that eats into our credits. We also don't want other forks to run this so should add some type of conditional:

https://github.community/t/do-not-run-cron-workflows-in-forks/17636

scripts/gh_ci_summary.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/python3
+
+from datetime import datetime
```
erimatnor, line 3: Can we add basic unit tests for this code similar to what we have for the Git hooks: https://github.com/timescale/timescaledb/blob/master/scripts/githooks/commit_msg_tests.py

Otherwise, there's a lot of Python code to "just trust" here.

mkindahl (commented):
LGTM, but you need to make sure that it only runs on the master branch.

.github/workflows/ci_summary.yaml
k-rus: We have cron jobs with the same setup, which runs every night and it doesn't run on forks. According to the [documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows):
>Scheduled workflows run on the latest commit on the default or base branch.

So I don't see an issue.

k-rus:
> LGTM, but you need to make sure that it only runs on the master branch.

@mkindahl It runs only on the master/default branch. From [docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows):
>Scheduled workflows run on the latest commit on the default or base branch.

mkindahl (approved):

gayyappan (approved):

PR #3135, opened 2021-04-20
Fix SkipScan path generation with expressions

### LINKED ISSUE #3134: SkipScan plan being chosen even though DISTINCT is on a function expression.
Consider a regular hypertable:

```
single=# \d metrics
                        Table "public.metrics"
  Column   |           Type           | Collation | Nullable | Default 
-----------+--------------------------+-----------+----------+---------
 time      | timestamp with time zone |           | not null | 
 device_id | integer                  |           |          | 
 v0        | integer                  |           |          | 
 v1        | integer                  |           |          | 
 v2        | double precision         |           |          | 
 v3        | double precision         |           |          | 
Indexes:
    "metrics_device_id_time_idx" btree (device_id, "time" DESC)
    "metrics_time_idx" btree ("time" DESC)
Triggers:
    ts_insert_blocker BEFORE INSERT ON metrics FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker()
Number of child tables: 3 (Use \d+ to list them.)
```

The following query errors out on this above hypertable:

```
single=# SELECT DISTINCT time_bucket('1h',time) col1                                                                                     
FROM metrics;                                                                                                                         
ERROR:  skip column not found in targetlist
```

SkipScan plan should not be chosen because the DISTINCT is on a function expression.

svenklemm:
In DISTINCT when deciding about adding a SkipScan path we try
to match the PathKey to the targetlist which would fail if
PathKey was an expression but the index not an expression
index. The check for the DISTINCT being a Var would only happen
after this check leading to a "skip column not found in targetlist"
error under these circumstances.

Fixes #3134

Disable-check: commit-count


erimatnor (commented):
Can you provide a PR description explaining the problem. Perhaps an example query statement?

tsl/test/expected/skip_scan.out
```diff
@@ -117,6 +117,11 @@ CREATE INDEX ON :TABLE(time,dev,val);
 :PREFIX SELECT DISTINCT *, 'q1_9' FROM :TABLE ORDER BY dev;
 :PREFIX SELECT DISTINCT dev, time, 'q1_10' FROM :TABLE ORDER BY dev;
 :PREFIX SELECT DISTINCT dev, NULL, 'q1_11' FROM :TABLE ORDER BY dev;
+-- distinct on expressions not supported
+:PREFIX SELECT DISTINCT time_bucket(10,time), 'q1_12' FROM :TABLE;
```
nikkhils, line 121: Could add a test case where the table has an INDEX on the expression `time_bucket('10,time)` where we can see that SkipScan doesn't get used.

svenklemm: there is a test for expression index couple lines above

nikkhils: Ah, ok. LGTM

nikkhils (approved):

mkindahl (approved):

mfundul:
What about `max_prepared_transactions=100 #set same as max_connections`, is this a problem?

svenklemm:
> What about `max_prepared_transactions=100 #set same as max_connections`, is this a problem?

In my testing max_connections was the limiting factor.

PR #3137, opened 2021-04-20
Create an empty chunk table

k-rus:
Adds an internal API function to create an empty chunk table according 
the given hypertable for the given chunk table name and dimension 
slices. This functions creates a chunk table inheriting from the 
hypertable, so it guarantees the same schema. No TimescaleDB's 
metadata is updated.

To be able to create the chunk table in a tablespace attached to the 
hyeprtable, this commit allows calculating the tablespace id without 
the dimension slice to exist in the catalog.

If there is already a chunk, which collides on dimension slices, the 
function fails to create the chunk table.

The function will be used internally in multi-node to be able to 
replicate a chunk from one data node to another.

### DIFF
--- a/sql/chunk.sql
+++ b/sql/chunk.sql
@@ -67,3 +67,10 @@ RETURNS TABLE(chunk_id INTEGER, hypertable_id INTEGER, att_num INTEGER, nullfrac
 slot1numbers FLOAT4[], slot2numbers FLOAT4[], slot3numbers FLOAT4[], slot4numbers FLOAT4[], slot5numbers FLOAT4[],
 slotvaluetypetrings CSTRING[], slot1values CSTRING[], slot2values CSTRING[], slot3values CSTRING[], slot4values CSTRING[], slot5values CSTRING[])
 AS '@MODULE_PATHNAME@', 'ts_chunk_get_colstats' LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.create_chunk_table(
+       hypertable REGCLASS,
+       slices JSONB,
+       schema_name NAME,
+       table_name NAME)
+RETURNS BOOL AS '@MODULE_PATHNAME@', 'ts_chunk_create_empty_table' LANGUAGE C VOLATILE STRICT;
--- a/src/dimension.c
+++ b/src/dimension.c
@@ -381,9 +381,19 @@ ts_dimension_get_open_slice_ordinal(Dimension *dim, DimensionSlice *slice)
 	/* Find the index (ordinal) of the chunk's slice in the open dimension */
 	i = ts_dimension_vec_find_slice_index(vec, slice->fd.id);
 
-	Assert(i >= 0);
-
-	return i;
+	if (i >= 0)
+		return i;
+	else
+	{
+		/*
+		 * Returns the number of slices if the slice not found, i.e., i = -1.
+		 * Dimension slice might not exist if a chunk table is created without
+		 * modifying metadata. It happens only during copy/move chunk for distributed
+		 * hypertable, thus this code, which is used when no space dimension exists,
+		 * is unlikely to be used.
+		 */
+		return vec->num_slices;
+	}
 }
 
 /*
--- a/tsl/src/chunk_api.c
+++ b/tsl/src/chunk_api.c
@@ -1627,3 +1627,45 @@ chunk_api_get_chunk_relstats(PG_FUNCTION_ARGS)
 {
 	return chunk_api_get_chunk_stats(fcinfo, false);
 }
+
+Datum
+chunk_create_empty_table(PG_FUNCTION_ARGS)
+{
+	const Oid hypertable_relid = PG_GETARG_OID(0);
+	Jsonb *const slices = PG_GETARG_JSONB_P(1);
+	const char *const schema_name = PG_GETARG_CSTRING(2);
+	const char *const table_name = PG_GETARG_CSTRING(3);
+	Cache *const hcache = ts_hypertable_cache_pin();
+	Hypertable *const ht = ts_hypertable_cache_get_entry(hcache, hypertable_relid, CACHE_FLAG_NONE);
+	Hypercube *hc;
+	const char *parse_err;
+	AclResult acl_result;
+
+	Assert(!PG_ARGISNULL(0));
+	Assert(!PG_ARGISNULL(1));
+	Assert(!PG_ARGISNULL(2));
+	Assert(!PG_ARGISNULL(3));
+	Assert(ht != NULL);
+
+	acl_result = pg_class_aclcheck(hypertable_relid, GetUserId(), ACL_INSERT);
+	if (acl_result != ACLCHECK_OK)
+		ereport(ERROR,
+				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+				 errmsg("permission denied for table \"%s\"", get_rel_name(hypertable_relid)),
+				 errdetail("Insert privileges required on \"%s\" to create chunks.",
+						   get_rel_name(hypertable_relid))));
+
+	hc = hypercube_from_jsonb(slices, ht->space, &parse_err);
+
+	if (hc == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid hypercube for hypertable \"%s\"", get_rel_name(hypertable_relid)),
+				 errdetail("%s", parse_err)));
+
+	ts_chunk_create_only_table(ht, hc, schema_name, table_name);
+
+	ts_cache_release(hcache);
+
+	PG_RETURN_BOOL(true);
+}
--- a/tsl/test/expected/chunk_api-12.out
+++ b/tsl/test/expected/chunk_api-12.out
@@ -85,6 +85,56 @@ SELECT * FROM _timescaledb_internal.create_chunk('chunkapi',' {"time": [15150240
 ERROR:  permission denied for table "chunkapi"
 DETAIL:  Insert privileges required on "chunkapi" to create chunks.
 \set ON_ERROR_STOP 1
+SET ROLE :ROLE_DEFAULT_PERM_USER;
+-- Test create_chunk_table is STRICT
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi', NULL, '_timescaledb_internal','_hyper_1_1_chunk');
+ create_chunk_table 
+--------------------
+ 
+(1 row)
+
+-- Test create_chunk_table for errors
+\set ON_ERROR_STOP 0
+-- Modified time constraint should fail with collision
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": [1514419600000000, 1515024000000000], "device": [-9223372036854775808, 1073741823]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  chunk table creation failed due to dimension slice collision
+-- Missing dimension
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": [1514419600000000, 1515024000000000]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  invalid hypercube for hypertable "chunkapi"
+DETAIL:  invalid number of hypercube dimensions
+-- Extra dimension
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": [1514419600000000, 1515024000000000],  "device": [-9223372036854775808, 1073741823], "time2": [1514419600000000, 1515024000000000]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  invalid hypercube for hypertable "chunkapi"
+DETAIL:  invalid number of hypercube dimensions
+-- Bad dimension name
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": [1514419600000000, 1515024000000000],  "dev": [-9223372036854775808, 1073741823]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  invalid hypercube for hypertable "chunkapi"
+DETAIL:  dimension "dev" does not exist in hypertable
+-- Same dimension twice
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": [1514419600000000, 1515024000000000], "time": [1514419600000000, 1515024000000000]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  invalid hypercube for hypertable "chunkapi"
+DETAIL:  invalid number of hypercube dimensions
+-- Bad bounds format
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": ["1514419200000000", 1515024000000000], "device": [-9223372036854775808, 1073741823]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  invalid hypercube for hypertable "chunkapi"
+DETAIL:  constraint for dimension "time" is not numeric
+-- Bad slices format
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": [1515024000000000], "device": [-9223372036854775808, 1073741823]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  invalid hypercube for hypertable "chunkapi"
+DETAIL:  unexpected number of dimensional bounds for dimension "time"
+-- Bad slices json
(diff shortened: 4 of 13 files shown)


tsl/src/chunk_api.c
```diff
@@ -1627,3 +1627,63 @@ chunk_api_get_chunk_relstats(PG_FUNCTION_ARGS)
...
+	const char *const schema_name = PG_GETARG_CSTRING(2);
+	const char *const table_name = PG_GETARG_CSTRING(3);
+	Cache *const hcache = ts_hypertable_cache_pin();
+	Hypertable *const ht = ts_hypertable_cache_get_entry(hcache, hypertable_relid, CACHE_FLAG_NONE);
+	Hypercube *hc;
+	const char *parse_err;
+	Chunk *chunk;
+	AclResult acl_result;
+
+	Assert(!PG_ARGISNULL(0));
```
erimatnor, line 1662: I suggest adding not that we can assert non-nullness because the function is `STRICT`. Although I really wonder if we want `STRICT` in this case as opposed to generating a useful error? With `STRICT` this function will succeed on NULL input and return a NULL result, which then might need handling by the caller as opposed to failing with a descriptive error.

tsl/src/chunk_api.c
```diff
@@ -1627,3 +1627,63 @@ chunk_api_get_chunk_relstats(PG_FUNCTION_ARGS)
...
+	const Chunk *chunk = ts_chunk_get_by_relid(relid, false);
+	Hypertable *ht;
+
+	if (NULL == chunk)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("must be a hypertable or chunk")));
+
+	ht = ts_hypertable_get_by_id(chunk->fd.hypertable_id);
+	PG_RETURN_TEXT_P(cstring_to_text(chunk_api_dimension_slices_json(chunk, ht)));
```
erimatnor, line 1645: Why not return JSONB here as opposed to text?

tsl/src/chunk_api.c
```diff
@@ -1627,3 +1627,63 @@ chunk_api_get_chunk_relstats(PG_FUNCTION_ARGS)
 {
 	return chunk_api_get_chunk_stats(fcinfo, false);
 }
+
+TS_FUNCTION_INFO_V1(ts_chunk_get_dimension_slices);
+Datum
+ts_chunk_get_dimension_slices(PG_FUNCTION_ARGS)
```
erimatnor, line 1633: Is this function needed?

There's already an internal `show_chunk` function that shows the slices:

```
nyc_data=# select slices from _timescaledb_internal.show_chunk('_timescaledb_internal._dist_hyper_1_150_chunk');
                                                    slices                                                    
--------------------------------------------------------------------------------------------------------------
 {"payment_type": [1073741823, 9223372036854775807], "pickup_datetime": [1452124800000000, 1452729600000000]}
(1 row)

```

tsl/test/expected/chunk_api-12.out
```diff
@@ -85,6 +94,58 @@ SELECT * FROM _timescaledb_internal.create_chunk('chunkapi',' {"time": [15150240
...
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time: [1515024000000000] "device": [-9223372036854775808, 1073741823]}', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  invalid input syntax for type json
+LINE 1: ...imescaledb_internal.create_chunk_table('chunkapi',' {"time: ...
+                                                             ^
+DETAIL:  Token "device" is invalid.
+CONTEXT:  JSON data, line 1:  {"time: [1515024000000000] "device...
+-- Valid chunk, but no permissions
+SET ROLE :ROLE_DEFAULT_PERM_USER_2;
+SELECT * FROM _timescaledb_internal.create_chunk_table('chunkapi',' {"time": [1515024000000000, 1519024000000000], "device": [-9223372036854775808, 1073741823]}', 'ChunkSchema', 'My_chunk_Table_name', '_timescaledb_internal','_hyper_1_1_chunk');
+ERROR:  function _timescaledb_internal.create_chunk_table(unknown, unknown, unknown, unknown, unknown, unknown) does not exist
```
erimatnor, line 144: Is this expected output? Doesn't seem to match comment above.

erimatnor (commented):
I realize this is in draft, but figured I'd give early feedback.

I am not sure I fully understand what this new function is doing and how it is different from the previous create_chunk function.

It seems to create metadata (dimension slices), but it is not clear why or what other metadata is _not_ created.

It seems to me that, when we create the chunk table, we can validate that it matches the hypertable schema and other things (collision), but ultimately the only "real" check can be done when the table is  "attached" to become a full member chunk.

Other comments inline.

k-rus:
> I am not sure I fully understand what this new function is doing and how it is different from the previous create_chunk function.
@erimatnor I don't understand what you try to say by this comment. Do you mean that we don't need a function to create a chunk table and should use the function, which creates the chunk? We decided that it is important to just create a table for the chunk and don't make it part of the hypertable. What do you miss in the PR to make it clear?

> 
> It seems to create metadata (dimension slices), but it is not clear why or what other metadata is _not_ created.
Do you miss a better commit message? Or what is the point of your statement?

> 
> It seems to me that, when we create the chunk table, we can validate that it matches the hypertable schema and other things (collision), but ultimately the only "real" check can be done when the table is "attached" to become a full member chunk.

The chunk is created according the hypertable schema. No chunk schema is provided during the call to function. So I don't follow this statment.

tsl/src/chunk_api.c
k-rus: I don't really follow why it is important to invest in descriptive error messages for NULL values in this internal function. Can you elaborate why it is important in the context of implementing copy/move functionality and the function being internal?

tsl/src/chunk_api.c
k-rus: Good point! I will look into it.

tsl/test/expected/chunk_api-12.out
k-rus: Thank you for noticing this! Fixed.

tsl/src/chunk_api.c
k-rus: Replaced with the suggestion.

tsl/src/chunk_api.c
k-rus: The function is removed as not needed.

k-rus:
@erimatnor I improved the commit message. I don't know if it helps, since I don't fully understand what your points are.

k-rus:
It was decided to use the default tablespace and do not add dimension slices to the catalog meta-table.

tsl/src/chunk_api.c
k-rus: See separate PR https://github.com/timescale/timescaledb/pull/3149

sql/chunk.sql
```diff
@@ -67,3 +67,10 @@ RETURNS TABLE(chunk_id INTEGER, hypertable_id INTEGER, att_num INTEGER, nullfrac
 slot1numbers FLOAT4[], slot2numbers FLOAT4[], slot3numbers FLOAT4[], slot4numbers FLOAT4[], slot5numbers FLOAT4[],
 slotvaluetypetrings CSTRING[], slot1values CSTRING[], slot2values CSTRING[], slot3values CSTRING[], slot4values CSTRING[], slot5values CSTRING[])
 AS '@MODULE_PATHNAME@', 'ts_chunk_get_colstats' LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.create_chunk_table(
+       hypertable REGCLASS,
+       slices JSONB,
+       schema_name NAME,
+       table_name NAME)
+RETURNS OID AS '@MODULE_PATHNAME@', 'ts_chunk_create_empty_table' LANGUAGE C VOLATILE STRICT;
```
erimatnor, line 76: When this function is called remotely from the AN, then the OID can't be used. Just pointing this out since it is easy to make the mistake if this is the signature. Other functions, like `create_chunk` elided the return of an OID in favor of schema+name for this reason.

tsl/test/expected/chunk_api-12.out
```diff
@@ -517,3 +567,211 @@ WARNING:  insufficient number of data nodes for distributed hypertable "disttabl
...
+(1 row)
+
+SELECT count(*) FROM 
+   _timescaledb_internal.create_chunk_table('chunkapi', :'SLICES', :'CHUNK_SCHEMA', :'CHUNK_NAME');
+ count 
+-------
+     1
+(1 row)
+
+-- Demonstrate that current settings for dimensions doesn't affect create_chunk_table
```
erimatnor, line 636:
```suggestion
-- Demonstrate that current settings for dimensions don't affect create_chunk_table
```

tsl/test/expected/chunk_api-12.out
```diff
@@ -517,3 +567,211 @@ WARNING:  insufficient number of data nodes for distributed hypertable "disttabl
...
+(1 row)
+
+SELECT * FROM _timescaledb_catalog.dimension_slice ORDER BY id;
+ id | dimension_id |     range_start      |    range_end     
+----+--------------+----------------------+------------------
+  2 |            2 | -9223372036854775808 |       1073741823
+  3 |            1 |     1515024000000000 | 1519024000000000
+(2 rows)
+
+-- Test create_chunk_table to a chunk table in a hypertable where chunk didn't exist before
```
erimatnor, line 619: Not sure I understand this sentence. Should it be:

```suggestion
-- Test create_chunk_table on a hypertable where the chunk didn't exist before
```

tsl/test/expected/chunk_api-12.out
```diff
@@ -517,3 +567,211 @@ WARNING:  insufficient number of data nodes for distributed hypertable "disttabl
...
+DROP TABLE chunkapi;
+-- Use the time partition to calculate the tablespace id to use
+CREATE TABLE chunkapi (time timestamptz not null, device int, temp float);
+SELECT * FROM create_hypertable('chunkapi', 'time');
+ hypertable_id | schema_name | table_name | created 
+---------------+-------------+------------+---------
+             8 | public      | chunkapi   | t
+(1 row)
+
+INSERT INTO chunkapi VALUES ('2018-01-01 05:00:00-8', 1, 23.4);
```
erimatnor, line 733: I would also like to see a test where the chunk does not collide due to being a different space partition (i.e., it still overlaps in the time dimension). For instance, if you instead insert this row:
```
INSERT INTO chunkapi VALUES ('2018-01-01 05:00:00-8', 2, 23.4);
```

Then, assuming "device 2" is another space partition than "device 1", the created chunk table should not collide.

erimatnor (commented):
Overall looks good. I have some inline suggestions for additional tests. I would also like to see how other functions are affected by this change. For instance, if you successfully create a chunk table with the new function, what happens when `create_chunk` tries to create the same chunk?

sql/chunk.sql
erimatnor: Also wondering if the return type should actually be `REGCLASS` (if we stick with OIDs)?

k-rus: Good point about not returning `OID` or `REGCLASS` for remotely called function.
Will it be suitable to return `BOOL`?

k-rus:
> For instance, if you successfully create a chunk table with the new function, what happens when `create_chunk` tries to create the same chunk?

@erimatnor I added a test to create chunk with the same name and it fails. If the name will be different or not provided, `create_chunk` will obviously succeed.

erimatnor (approved):

sql/chunk.sql
```diff
@@ -67,3 +67,10 @@ RETURNS TABLE(chunk_id INTEGER, hypertable_id INTEGER, att_num INTEGER, nullfrac
 slot1numbers FLOAT4[], slot2numbers FLOAT4[], slot3numbers FLOAT4[], slot4numbers FLOAT4[], slot5numbers FLOAT4[],
 slotvaluetypetrings CSTRING[], slot1values CSTRING[], slot2values CSTRING[], slot3values CSTRING[], slot4values CSTRING[], slot5values CSTRING[])
 AS '@MODULE_PATHNAME@', 'ts_chunk_get_colstats' LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.create_chunk_table(
```
nikkhils, line 71: We are doing all of this in the context of `chunk copy/move`. So all these new APIs should reflect that. IMO, the name of this function should be `create_chunk_replica` and it needs to take the destination data node name as an input argument. The other argument that's needed is the `chunk_id`.  We should get all the dimension slices info for this chunk.

Then we should call something similar to `chunk_api_create_on_data_nodes` passing this target data node as input. I think we should modify CHUNK_CREATE_STMT to take in an additional argument to not create hypertable/metadata linkages on the DN.

IIUC, isn't the above enough to do what we want here?

src/dimension.c
```diff
@@ -381,9 +381,19 @@ ts_dimension_get_open_slice_ordinal(Dimension *dim, DimensionSlice *slice)
...
 
-	Assert(i >= 0);
-
-	return i;
+	if (i >= 0)
+		return i;
+	else
+	{
+		/*
+		 * Returns the number of slices if the slice not found, i.e., i = -1.
```
nikkhils, line 389: Better to error out since we don't expect this code path to be reached for regular tables?

sql/chunk.sql
k-rus: > We are doing all of this in the context of `chunk copy/move`. So all these new APIs should reflect that. IMO, the name of this function should be `create_chunk_replica` and it needs to take the destination data node name as an input argument. The other argument that's needed is the `chunk_id`. We should get all the dimension slices info for this chunk.

This is a local function, which needs to exist to be able to implement the function on the access node. It is separate task to create the function on the access node, which will take distributed chunk as input. So it is outside the scope of this PR.



> I think we should modify CHUNK_CREATE_STMT/chunk_create to take in an additional argument to not create hypertable/metadata linkages on the DN. You have already made good changes in `chunk_create_from_hypercube_after_lock` and other functions to support this.

This requires to change the signature of `create_chunk` and it brings different semantics to it. Such change requires a team discussion. Also the refactoring of underlying C function will make it more complicated. So imho I am not in favor of this idea.



> Here the assumption is that the hypertable exists on this DN. If it does not then we would need to create that using `attach_data_node` as part of copy/move chunk.

It is outside the scope of this PR.

src/dimension.c
k-rus: I don't follow. Returning error means that creating a chunk table will always fail if no space dimension exists.

sql/chunk.sql
nikkhils: > This is a local function, which needs to exist to be able to implement the function on the access node. It is separate task to create the function on the access node, which will take distributed chunk as input. So it is outside the scope of this PR.
> 

Ok, I see you have a separate work item for the distributed function call.

> > I think we should modify CHUNK_CREATE_STMT/chunk_create to take in an additional argument to not create hypertable/metadata linkages on the DN. You have already made good changes in `chunk_create_from_hypercube_after_lock` and other functions to support this.
> 
> This requires to change the signature of `create_chunk` and it brings different semantics to it. Such change requires a team discussion. Also the refactoring of underlying C function will make it more complicated. So imho I am not in favor of this idea.
> 

Ok. Looks like we should discuss this because your function is basically a cut down version of `create_chunk` AIUI. We could always default this additional argument (if we choose to modify create_chunk) to true. And I don't see this`create_chunk` function in the documentation, so, I don't think it's supposed to be used by users directly anyways.

> > Here the assumption is that the hypertable exists on this DN. If it does not then we would need to create that using `attach_data_node` as part of copy/move chunk.
> 
> It is outside the scope of this PR.

Sure, if not already added, we should add a work item for this in the main copy/move chunk issue.

k-rus: > Looks like we should discuss this because your function is basically a cut down version of `create_chunk` AIUI. We could always default this additional argument (if we choose to modify create_chunk) to true. And I don't see this`create_chunk` function in the documentation, so, I don't think it's supposed to be used by users directly anyways.

It is not just cut down. It also behaves differently if the chunk already exists. So few considerations:
1. The existing `chunk_table` either creates a chunk or returns the existing chunk. `create_chunk_table` either creates the table or fails.
2. The existing `chunk_table` returns `chunk_id`, while `create_chunk_table` will never return `chunk_id`, since it is not allocated at this step and will be allocated during attaching the chunk table.
3. Attaching chunk table will also use a part of `chunk_table` implementation. So if `create_chunk_table` is extended with creating an empty chunk table, then it is logical to include attaching chunk into the same API. IMHO it will make implementation of `chunk_table` more difficult to follow.

The primary usage of `create_chunk_table` is to insert data into a distributed hypertable.

I suggest to implement separate functions first and have a separate discussion about the API, since the functions needed for copy/move chunk might be exposed  as public API in future.

nikkhils (approved):

Issue #3025, opened 2021-03-12
INSERT on a distributed hypertable is significantly slower than COPY


ryanbooz:
**Relevant system information:**
 - OS: Timescale Forge (although non-Forge experiences similar performance)
 - PostgreSQL version: 12.5
 - TimescaleDB version: 2.0.1
 - Installation method: Timescale Forge

**Describe the bug**
Using INSERT on a distributed hypertable is significantly slower than inserting into a regular hypertable with INSERT. Based on feedback in the linked Slack conversation, I tested INSERT/COPY performance using TSBS on a regular and distributed hypertable. Across 8 different tests (results below), it is evident that INSERT actually performs worse on a distributed hypertable performs worse than a single-node. Even though we expect a batched INSERT statement to be slower than COPY, the expectation is that more nodes would still, in some way, improve ingest performance with INSERT in a similar fashion as COPY. That does not appear to be the case.

All tests inserted 36,000,000 rows, in batches of 4,000 rows, with TSBS from a pre-generated file (using `FILE`mode)

### Single-node, regular hypertable (8cpu/32gb)
| Number of Workers | Method | Performance |
|-----|----|----|
| 1 | COPY | loaded 36000000 rows in 470.899sec with 1 workers (mean rate 76449.48 rows/sec) |
| 1 | INSERT | loaded 36000000 rows in 1144.999sec with 1 workers (mean rate 31441.09 rows/sec) |
| 10 | COPY | loaded 36000000 rows in 119.607sec with 10 workers (mean rate 300985.30 rows/sec) |
| 10 | INSERT | loaded 36000000 rows in 195.681sec with 10 workers (mean rate 183972.47 rows/sec) |

### Multi-node (1AN/4DN), distributed hypertable (4cpu/16gb) 
| Number of Workers | Method | Performance |
|-----|----|----|
| 1 | COPY | loaded 36000000 rows in 403.772sec with 1 workers (mean rate 89159.29 rows/sec) |
| 1 | INSERT | loaded 36000000 rows in 2043.873sec with 1 workers (mean rate 17613.62 rows/sec) |
| 10 | COPY | loaded 36000000 rows in 95.850sec with 10 workers (mean rate 375585.24 rows/sec) |
| 10 | INSERT | loaded 36000000 rows in 272.150sec with 10 workers (mean rate 132280.02 rows/sec) |

[...]

PR #3138, opened 2021-04-20
Improve INSERT performance on distributed hypertables

erimatnor:
Refactor the data node dispatch node to send batches of tuples to data
nodes in columnar format. The columnar format is implemented by
rolling up multiple rows of values into one row of arrays of
values. So, an insert like

```
INSERT INTO hypertable (time, value)
VALUES ('2021-04-19', 1), (2021-04-20, 2);
```
is deparsed to

```
INSERT INTO hypertable (time, value)
SELECT * FROM unnest(ARRAY['2021-04-19', '2021-04-20'], ARRAY[1, 2]) a(t,v);
```

when sent to data nodes. This approach allows sending an arbitrary
number of rows using the same prepared statement:

```
INSERT INTO hypertable (time, value)
SELECT * FROM unnest($1::timestamptz[], $2::integer[]) a(t,v)
```

Previously, the number of rows that could be sent in a batch was
limited by the number of parameters allowed in a prepared statement.

Special care has been taken to handle custom types (e.g., composite)
as the internal fields of such types are unnested along with the
array. For such cases, the deparsed insert statement must look like:

```
INSERT INTO hypertable (time, value)
SELECT time, (high, low)::customtype
FROM unnest($1::timestamptz[], $2::custom_type[]) a(t,high,low);
```

Array columns present another challange. In such cases we must deparse
into an array of arrays. This puts a limit on the number of dimensions
we can support, as rolling up an array adds an extra dimension.

### DIFF
--- a/tsl/src/fdw/deparse.c
+++ b/tsl/src/fdw/deparse.c
@@ -77,6 +77,7 @@
 
 #include <func_cache.h>
 #include <remote/utils.h>
+#include <guc.h>
 
 #include "relinfo.h"
 #include "deparse.h"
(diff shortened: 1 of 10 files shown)


erimatnor:
Still need to implement tests for custom/composite types and add support for tables with array columns.

erimatnor:
Support for array columns have proven problematic. Details in the issue https://github.com/timescale/timescaledb/issues/3025.

Closing until a solution can be implemented.

PR #3148, opened 2021-04-22
Make SELECT DISTINCT handle non-var targetlists

nikkhils:
The current SELECT DISTINCT pushdown code assumed that the targetlist
will always contain references to column attributes of the target
table.

So, even a simple "SELECT DISTINCT 1 from table;" causes a segmentation
fault because the "varno" field is not assigned. Fix this oversight.

Issue reported by @svenklemm

Fixes timescale/timescaledb-private#920

### DIFF
--- a/tsl/test/shared/expected/dist_distinct.out
+++ b/tsl/test/shared/expected/dist_distinct.out
@@ -888,6 +888,34 @@ LIMIT 10;
                            Remote SQL: SELECT "time" FROM public.metrics_dist WHERE _timescaledb_internal.chunks_in(public.metrics_dist.*, ARRAY[1, 2, 3]) ORDER BY public.time_bucket('01:00:00'::interval, "time") ASC NULLS LAST
 (23 rows)
 
+SELECT DISTINCT without any var references is handled correctly
+EXPLAIN (verbose, costs off)
+SELECT DISTINCT 1, 'constx'
+FROM metrics_dist;
+                                                                    QUERY PLAN                                                                     
+---------------------------------------------------------------------------------------------------------------------------------------------------
+ Unique
+   Output: 1, 'constx'::text
+   ->  Custom Scan (AsyncAppend)
+         Output: 1, 'constx'::text
+         ->  Append
+               ->  Custom Scan (DataNodeScan) on public.metrics_dist metrics_dist_1
+                     Output: 1, 'constx'::text
+                     Data node: data_node_1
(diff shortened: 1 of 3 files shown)


svenklemm:
Hmm SELECT DISTINCT Const returns 0 or 1 rows, but SELECT Const returns Const as many times as there are rows so i think it would be beneficial to push down the distinct here

tsl/test/shared/expected/dist_distinct.out
```diff
@@ -888,6 +888,34 @@ LIMIT 10;
...
+ Unique
+   Output: 1, 'constx'::text
+   ->  Custom Scan (AsyncAppend)
+         Output: 1, 'constx'::text
+         ->  Append
+               ->  Custom Scan (DataNodeScan) on public.metrics_dist metrics_dist_1
+                     Output: 1, 'constx'::text
+                     Data node: data_node_1
+                     Chunks: _dist_hyper_7_37_chunk, _dist_hyper_7_40_chunk, _dist_hyper_7_43_chunk
+                     Remote SQL: SELECT NULL FROM public.metrics_dist WHERE _timescaledb_internal.chunks_in(public.metrics_dist.*, ARRAY[1, 2, 3])
```
erimatnor, line 906: I guess it is weird that we even do a query against data nodes in this case. But perhaps that is another fix or just a weirdness with the PostgreSQL query plan?

erimatnor (approved):

nikkhils:
> Hmm SELECT DISTINCT Const returns 0 or 1 rows, but SELECT Const returns Const as many times as there are rows so i think it would be beneficial to push down the distinct here

The underlying remote query is:

```
SELECT NULL FROM public.hyper WHERE _timescaledb_internal.chunks_in(public.hyper.*, ARRAY[1, 2]);
```

We could check if `varno_assigned` is false and blindly append the `DISTINCT` if so. But don't think we need to optimize for this use case.

tsl/test/shared/expected/dist_distinct.out
nikkhils: Another fix in the fdw layer. PostgreSQL is flexible in terms of the myriad queries it allows. 

`SELECT DISTINCT const1 from table` 

is allowed and that's what we are trying to not crash for here. IMO, nothing else needed to optimize for such queries.

svenklemm (approved):

tsl/test/shared/expected/dist_distinct.out
nikkhils: Added the DISTINCT pushdown here now.

Issue #3139, opened 2021-04-20
Improve error reporting of connection failures in multinode tests


svenklemm:
When running regresscheck-t and checking postgres log there are multiple instances of
```
cluster_super_user [12405] db_dist_commands_1 FATAL:  connection requires a valid client certificate
```
This errors are never reported to the accessnode and get swallowed somewhere and are probably the cause
of the flaky multinode tests.

svenklemm:
To reproduce: lower max_connections in postgres.conf and run regresscheck-t

svenklemm:
multinode has been removed

PR #3152, opened 2021-04-23
Fix flaky dist_hypertable test

nikkhils:
The chunks of the datanodes seem to contain very few rows. That can
cause PostgreSQL to choose a SeqScan plan in some cases. Add a few
more rows to make it choose IndexScan all of the time.


tsl/test/expected/dist_hypertable-11.out
```diff
@@ -192,12 +192,22 @@ INSERT INTO disttable VALUES
 INSERT INTO disttable VALUES
        ('2017-01-01 06:01', 1, 1.1),
        ('2017-01-01 09:11', 3, 2.1),
+       ('2017-01-01 09:21', 3, 2.2),
```
erimatnor, line 195: How do we know this is enough? Is it purely incidental?

nikkhils: That's why this is a draft. Will try with a few variations if this doesn't work.

nikkhils: @erimatnor I re-ran the PG 11/12/13 regression tests here multiple times and `dist_hypertable` test is holding up. Looks like this is a good start.

mkindahl (approved):

tsl/test/expected/dist_hypertable-11.out
svenklemm: i'm having doubts this will fix it, have a look at https://github.com/timescale/timescaledb/issues/3139
you can reproduce the error locally by reverting PARALLEL_GROUP_SIZE and max_connection change in https://github.com/timescale/timescaledb/commit/202cceee56d1f8fb5844b0d7be6da2f9d65120f8 
and lowering max_connections to 50

nikkhils: @svenklemm as you can see, this specific failure is a PLAN failure and doesn't seem to have anything to do with max connections.

svenklemm: i could reproduce it doing the wrong plan with those changed settings

svenklemm (approved):

tsl/test/expected/dist_hypertable-11.out
nikkhils: Thanks for confirming. As I said, we can always iterate this further if issues persist on CI.

PR #3153, opened 2021-04-23
First steps toward PostgreSQL 14

afiskon:
`StrNCpy` and `namecpy` are gone in PostgreSQL 14. Apparently, the community uses `strlcpy` instead. Plus minor changes to get rid of warnings about redefining macros when using MacOS and Clang.


afiskon:
As it turned out @svenklemm is already working on porting Timescale to PostgreSQL 14. We agreed to make a separate feature branch.

PR #3159, opened 2021-04-25
Introduce function to copy chunk data between data nodes

pmwkaa:
Add internal copy_chunk_data() function which implements a way
to copy chunk data between data nodes using logical
replication.

Some moments to consider

* this functionality needs to be tested using separate instances,  `test_copy_chunk_data.sh` been introduced for that purpose
* no chunk locking been made during the copy operation
* no tests for corner cases such as connection failure during the copy or operation retry after failure
* there is no undo/clean procedure in case if operation fails during its operation

### DIFF
--- a/sql/ddl_internal.sql
+++ b/sql/ddl_internal.sql
@@ -37,3 +37,33 @@ CREATE OR REPLACE FUNCTION  _timescaledb_internal.chunk_drop_replica(
     node_name               NAME
 ) RETURNS VOID
 AS '@MODULE_PATHNAME@', 'ts_chunk_drop_replica' LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE PROCEDURE _timescaledb_internal.wait_subscription_sync(
+    schema_name    NAME,
+    table_name     NAME,
+    retry_count    INT DEFAULT 18000,
+    retry_delay_ms NUMERIC DEFAULT 0.200
+)
+LANGUAGE PLPGSQL AS
+$BODY$
+DECLARE
+    in_sync BOOLEAN;
+BEGIN
+    FOR i in 1 .. retry_count
+    LOOP
+        SELECT pgs.srsubstate = 'r'
+        INTO in_sync
+        FROM pg_subscription_rel pgs
+        JOIN pg_class pgc ON relname = table_name
+        JOIN pg_namespace n ON (n.OID = pgc.relnamespace)
+        WHERE pgs.srrelid = pgc.oid AND schema_name = n.nspname;
+
+        if (in_sync IS NULL OR NOT in_sync) THEN
+          PERFORM pg_sleep(retry_delay_ms);
+        ELSE
+          RETURN;
+        END IF;
+    END LOOP;
+    RAISE 'subscription sync wait timedout';
+END
+$BODY$;
--- a/tsl/src/chunk.c
+++ b/tsl/src/chunk.c
@@ -41,6 +41,7 @@
 #include "data_node.h"
 #include "deparse.h"
 #include "remote/dist_commands.h"
+#include "dist_util.h"
 
 static bool
 chunk_match_data_node_by_server(const Chunk *chunk, const ForeignServer *server)
@@ -410,7 +411,7 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
 						quote_identifier(chunk->fd.schema_name.data),
 						quote_identifier(chunk->fd.table_name.data));
 	data_nodes = list_make1((char *) node_name);
-	ts_dist_cmd_run_on_data_nodes(drop_cmd, data_nodes);
+	ts_dist_cmd_run_on_data_nodes(drop_cmd, data_nodes, true);
 
 	/*
 	 * This chunk might have this data node as primary, change that association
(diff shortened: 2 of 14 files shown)


tsl/src/chunk_api.c
```diff
@@ -1627,3 +1629,81 @@ chunk_api_get_chunk_relstats(PG_FUNCTION_ARGS)
 {
 	return chunk_api_get_chunk_stats(fcinfo, false);
 }
+
+Datum
+chunk_api_copy_data(PG_FUNCTION_ARGS)
```
erimatnor, line 1634: Nit: Usually, `chunk_api` contains functions that are executed on the DNs, not the functions called from the AN.

tsl/src/chunk_api.c
```diff
@@ -1627,3 +1629,81 @@ chunk_api_get_chunk_relstats(PG_FUNCTION_ARGS)
...
+	src_node = lappend(NIL, pstrdup(src_node_name));
+	dst_node = lappend(NIL, pstrdup(dst_node_name));
+
+	/* prepare shared operation id name for publication and subscription */
+	operation_id = psprintf("ts_copy_chunk_%s",
+							quote_qualified_identifier(NameStr(chunk->fd.schema_name),
+													   NameStr(chunk->fd.table_name))),
+
+	/* create publication on the source data node */
+		ts_dist_cmd_run_on_data_nodes(psprintf("CREATE PUBLICATION %s FOR TABLE %s",
```
erimatnor, line 1681: I am wondering if we should wrap these remote SQL commands in four API functions to be executed on the DNs:

```
chunk_create_subscription()
chunk_drop_subscription()
chunk_create_publication()
chunk_drop_publication()
```

Then a this  `copy_chunk` function is executed on the AN  and will call the above functions.

Why wrap in functions? I imagine that it makes testing easier, we can to parameter validation, and raise more appropriate errors for our use case.

Or is this over-doing it?

pmwkaa:
> Might make sense to provide sensible defaults for `retry_count` and `retry_delay_ms`

It is set to 1 hour by default. Do we want to make a macro from it?

sql/ddl_internal.sql
```diff
@@ -37,3 +37,31 @@ CREATE OR REPLACE FUNCTION  _timescaledb_internal.chunk_drop_replica(
...
+$BODY$
+DECLARE
+    in_sync BOOLEAN;
+BEGIN
+    FOR i in 1 .. retry_count
+    LOOP
+        SELECT pgs.srsubstate = 'r'
+        INTO in_sync
+        FROM pg_subscription_rel pgs
+        JOIN pg_class pgc ON relname = table_name
```
nikkhils, line 57: We'd also need to use the `schema_name`.

tsl/src/chunk.c
```diff
@@ -421,3 +422,134 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
 	PG_RETURN_VOID();
 }
+
+typedef struct
+{
+	Chunk *chunk;
+	NameData operation_id;
+	List *src_node;
+	List *dst_node;
+} ChunkCopyData;
```
afiskon, line 432: I believe several comments to the structure and its fields would be appropriate.

scripts/test_copy_chunk_data.sh
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env bash
+
+#
+# Test copy_chunk_data(): using logical replication to copy chunk data
+# between data nodes
```
afiskon, line 5: Any chance we could utilize TAP tests in this one? I realize this is additional and probably unnecessary work and don't insist on it. On the other hand, Perl code is much easier to maintain than Bash and would be nice to be consistent with future tests.

tsl/src/chunk.c
```diff
@@ -421,3 +422,134 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+
+	/* create subscription on the destination data node */
+
+	/*
+	 * CREATE SUBSCRIPTION from a database within the same database cluster will hang,
+	 * create the replication slot separately before creating the subscription
+	 */
+	cmd = psprintf("SELECT pg_create_logical_replication_slot('%s', 'pgoutput')",
+				   NameStr(ccd->operation_id));
+	ts_dist_cmd_run_on_data_nodes(cmd, ccd->src_node, false);
```
afiskon, line 455: Just curious, what will happen on data nodes when the access node fails after executing this code?

tsl/src/remote/connection.c
```diff
@@ -1390,6 +1408,60 @@ add_userinfo_to_server_options(ForeignServer *server, Oid user_id)
...
+	server = data_node_get_foreign_server(node_name, ACL_NO_CHECK, false, false);
+	connection_options = add_userinfo_to_server_options(server, GetUserId());
+	setup_full_connection_options(connection_options, &keywords, &values);
+
+	/* Cycle through the options and create the connection string */
+	initStringInfo(&buf);
+	i = 0;
+	while (keywords[i] != NULL)
+	{
+		appendStringInfo(&buf, "%s=%s ", keywords[i], escape_param_str(values[i]));
```
afiskon, line 1453: Might be a good idea to call pfree() here on the string returned by escape_param_str()

svenklemm (changes requested):
The current implementation of the connection string handling allows for sql injection in connection string parameters. You shouldnt roll your own escape function. The correct escape for `'` is `''` unless you mark the string as having escape sequences.

tsl/src/chunk.c
nikkhils: The replication slot needs to be created on the destination node and not on the source node. I wonder how your test script is succeeding with this..

.github/workflows/cron-tests.yaml
```diff
@@ -81,3 +82,22 @@ jobs:
     - name: Test recovery using restore points
       run: |
         ./scripts/docker-run-restore-points-test.sh
+
+  copy_chunk_data:
```
nikkhils, line 86: This needs to be removed now that tap test infra is in place

tsl/src/chunk.c
```diff
@@ -421,3 +422,134 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+chunk_copy_data_perform(ChunkCopyData *ccd)
+{
+	const char *cmd;
+
+	/* start data transfer on the destination node */
+	cmd = psprintf("ALTER SUBSCRIPTION %s ENABLE", NameStr(ccd->operation_id));
+	ts_dist_cmd_run_on_data_nodes(cmd, ccd->dst_node, false);
+
+	/* wait until data transfer finishes */
+	cmd = psprintf("CALL _timescaledb_internal.wait_subscription_sync(%s, %s)",
```
nikkhils, line 478: This will create an infinite loop wait because you are creating SUBSCRIPTION in the same transaction, so there won't be an entry that is visible in the `pg_subscription_rel` catalog yet. The wait has to happen after the CREATE SUBS/ALTER ENABLE has committed..

tsl/src/chunk.c
pmwkaa: Previous effect of this command will still be remained and has to be manually cleaned up. We decided to address this issue separately

pmwkaa:
Latest updates:

* Removed escape function and used `PQescapeString` analogue instead @svenklemm
* Removed bash/docker scripts to test copy chunk data and added similar TAP test instead @nikkhils, @afiskon

tsl/src/chunk.c
pmwkaa: I guess we figured it out during discussions :)

sql/ddl_internal.sql
```diff
@@ -37,3 +37,33 @@ CREATE OR REPLACE FUNCTION  _timescaledb_internal.chunk_drop_replica(
     node_name               NAME
 ) RETURNS VOID
 AS '@MODULE_PATHNAME@', 'ts_chunk_drop_replica' LANGUAGE C VOLATILE;
+
+CREATE OR REPLACE PROCEDURE _timescaledb_internal.wait_subscription_sync(
+    schema_name    NAME,
+    table_name     NAME,
+    retry_count    INT DEFAULT 18000,
+    retry_delay_ms NUMERIC DEFAULT 0.200
```
nikkhils, line 45: `pg_sleep` takes seconds as input. So you are basically passing 0.2 seconds as default. Given than retry count is 10000, this will wait for a very long time.

pmwkaa: This supposed to be an hour in total

tsl/src/remote/connection.c
```diff
@@ -1468,6 +1486,45 @@ add_userinfo_to_server_options(ForeignServer *server, Oid user_id)
...
+	{
+		int value_len = strlen(values[i]);
+
+		appendStringInfo(&buf, " %s=", keywords[i]);
+
+		/* ensure there is enough space reserved for escaped value */
+		enlargeStringInfo(&buf, value_len * 2 + 1);
+
+		/* escape value */
+		buf.len += PQescapeString(buf.data + buf.len, values[i], value_len);
```
svenklemm, line 1516: i think the correct function to use here is appendConnStrVal and then call PQescapeString on the whole string you pass as conninfo into CREATE SUBSCRIPTION, you are not treating the individual key-value pairs of the connection string correctly here, values have to be wrapped in quotes if they contain spaces.

pmwkaa: It seems to be `appendConnStrVal` is part of the libpq source which is not part of imported PostgreSQL include directory we are using and sits in the `interfaces/libpq` directory. It is not clear how we should add this include path unless we always provide `PG_SOURCE_DIR` which we don't. Looks like it is easier to use `PQescapeString()` in both cases here

afiskon (approved):

tsl/src/remote/connection.c
```diff
@@ -1468,6 +1486,90 @@ add_userinfo_to_server_options(ForeignServer *server, Oid user_id)
...
+	{
+		appendStringInfo(&connstr, " %s=", keywords[i]);
+		remote_connection_append_connstr_value(&connstr, values[i]);
+		i++;
+	}
+	Assert(keywords[i] == NULL && values[i] == NULL);
+
+	initStringInfo(&connstr_escape);
+	enlargeStringInfo(&connstr_escape, connstr.len * 2 + 1);
+	connstr_escape.len += PQescapeString(connstr_escape.data, connstr.data, connstr.len);
```
svenklemm, line 1563: i would prefer to see the sql escaping in the caller where you build the sql string, but other than that looks good

PR #3160, opened 2021-04-26
Implement drop_chunk_replica API

nikkhils:
This function drops a chunk on a specified data node. It then removes
the metadata about the datanode, chunk association on the access node.

This function is meant for internal use as part of the "move chunk"
functionality.

If only one chunk replica remains then this function refuses to drop it
to avoid data loss.

### DIFF
--- a/sql/ddl_internal.sql
+++ b/sql/ddl_internal.sql
@@ -30,3 +30,10 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.create_chunk_replica_table(
     chunk REGCLASS,
     data_node_name NAME
 ) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_chunk_create_replica_table' LANGUAGE C VOLATILE;
+
+-- Drop the specified chunk replica on the specified data node
+CREATE OR REPLACE FUNCTION  _timescaledb_internal.chunk_drop_replica(
+    chunk                   REGCLASS,
+    node_name               NAME
+) RETURNS VOID
+AS '@MODULE_PATHNAME@', 'ts_chunk_drop_replica' LANGUAGE C VOLATILE;
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -925,6 +925,30 @@ ts_chunk_get_data_node_name_list(const Chunk *chunk)
 	return datanodes;
 }
 
+bool
+ts_chunk_has_data_node(const Chunk *chunk, const char *node_name)
+{
+	ListCell *lc;
+	ChunkDataNode *cdn;
+	bool found = false;
+
+	if (chunk == NULL || node_name == NULL)
+		return false;
+
+	/* check that the chunk is indeed present on the specified data node */
+	foreach (lc, chunk->data_nodes)
+	{
+		cdn = lfirst(lc);
+		if (namestrcmp(&cdn->fd.node_name, node_name) == 0)
+		{
+			found = true;
+			break;
+		}
+	}
+
+	return found;
+}
+
 static int32
 get_next_chunk_id()
 {
--- a/src/cross_module_fn.h
+++ b/src/cross_module_fn.h
@@ -129,6 +129,7 @@ typedef struct CrossModuleFunctions
 	Datum (*chunk_set_default_data_node)(PG_FUNCTION_ARGS);
 	Datum (*create_chunk)(PG_FUNCTION_ARGS);
 	Datum (*show_chunk)(PG_FUNCTION_ARGS);
+
 	List *(*get_and_validate_data_node_list)(ArrayType *nodearr);
 	void (*hypertable_make_distributed)(Hypertable *ht, List *data_node_names);
 	Datum (*timescaledb_fdw_handler)(PG_FUNCTION_ARGS);
@@ -159,6 +160,7 @@ typedef struct CrossModuleFunctions
 	PGFunction hypertable_distributed_set_replication_factor;
 	PGFunction chunk_create_empty_table;
 	PGFunction chunk_create_replica_table;
+	PGFunction chunk_drop_replica;
 	void (*update_compressed_chunk_relstats)(Oid uncompressed_relid, Oid compressed_relid);
 } CrossModuleFunctions;
 
--- a/tsl/src/chunk.c
+++ b/tsl/src/chunk.c
@@ -13,6 +13,7 @@
 #include <access/htup_details.h>
 #include <access/xact.h>
 #include <nodes/makefuncs.h>
+#include <utils/builtins.h>
 #include <utils/syscache.h>
 #include <utils/inval.h>
 #include <utils/tuplestore.h>
@@ -328,6 +329,95 @@ chunk_create_replica_table(PG_FUNCTION_ARGS)
 	chunk_api_call_create_empty_chunk_table(ht, chunk, data_node_name);
 
 	ts_cache_release(hcache);
+
+	PG_RETURN_VOID();
+}
+
+/*
+ * chunk_drop_replica:
+ *
+ * This function drops a chunk on a specified data node. It then
+ * removes the metadata about the association of the chunk to this
+ * data node on the access node.
+ */
+Datum
+chunk_drop_replica(PG_FUNCTION_ARGS)
+{
+	Oid chunk_relid = PG_ARGISNULL(0) ? InvalidOid : PG_GETARG_OID(0);
+	const char *node_name = PG_ARGISNULL(1) ? NULL : NameStr(*PG_GETARG_NAME(1));
+	List *data_nodes;
+	ForeignServer *server;
+	Chunk *chunk;
+	const char *drop_cmd;
+
+	TS_PREVENT_FUNC_IF_READ_ONLY();
+
+	if (!OidIsValid(chunk_relid))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("invalid chunk relation")));
+
+	chunk = ts_chunk_get_by_relid(chunk_relid, false);
+
+	if (NULL == chunk)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid chunk relation"),
+				 errdetail("Object with OID %u is not a chunk relation", chunk_relid)));
+
+	/* It has to be a foreign table chunk */
+	if (chunk->relkind != RELKIND_FOREIGN_TABLE)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("\"%s\" is not a valid remote chunk", get_rel_name(chunk_relid))));
+
+	server = data_node_get_foreign_server(node_name, ACL_USAGE, true, false);
+	Assert(NULL != server);
+
+	/* Early abort on missing permissions */
+	ts_hypertable_permissions_check(chunk_relid, GetUserId());
+
+	if (!ts_chunk_has_data_node(chunk, node_name))
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("chunk \"%s\" does not exist on data node \"%s\"",
+						get_rel_name(chunk_relid),
+						node_name)));
+
+	/*
+	 * There should be at least one surviving replica after the deletion here.
+	 *
+	 * We could fetch the corresponding hypertable and check its
+	 * replication_factor. But the user of this function is using it
+	 * to move chunk from one data node to another and is well aware of
+	 * the replication_factor requirements
+	 */
+	if (list_length(chunk->data_nodes) <= 1)
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_INSUFFICIENT_NUM_DATA_NODES),
+				 errmsg("cannot drop the last chunk replica"),
+				 errdetail("Dropping the last chunk replica could lead to data loss.")));
+
+	/*
+	 * Drop chunk on the data node using a regular "DROP TABLE".
+	 * Note that CASCADE is not required as it takes care of dropping compressed
+	 * chunk (if any).
+	 *
+	 * If there are any other non-TimescaleDB objects attached to this table due
+	 * to some manual user activity then they should be dropped by the user
+	 * before invoking this function.
+	 */
+	drop_cmd = psprintf("DROP TABLE %s.%s",
+						quote_identifier(chunk->fd.schema_name.data),
+						quote_identifier(chunk->fd.table_name.data));
+	data_nodes = list_make1((char *) node_name);
+	ts_dist_cmd_run_on_data_nodes(drop_cmd, data_nodes);
+
+	/*
+	 * This chunk might have this data node as primary, change that association
+	 * if so. Then delete the chunk_id and node_name association.
+	 */
+	chunk_update_foreign_server_if_needed(chunk->fd.id, server->serverid);
+	ts_chunk_data_node_delete_by_chunk_id_and_node_name(chunk->fd.id, node_name);
 
 	PG_RETURN_VOID();
 }
--- a/tsl/test/shared/expected/chunk_move_copy.out
+++ b/tsl/test/shared/expected/chunk_move_copy.out
@@ -1,0 +1,134 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+CREATE TABLE mvcp_hyper (time bigint NOT NULL, value integer);
+SELECT table_name FROM create_distributed_hypertable('mvcp_hyper', 'time',
+        chunk_time_interval => 200, replication_factor => 3);
+ table_name 
+------------
+ mvcp_hyper
+(1 row)
+
+-- Enable compression so that we can test dropping of compressed chunks
+ALTER TABLE mvcp_hyper  SET (timescaledb.compress, timescaledb.compress_orderby='time DESC');
+INSERT INTO mvcp_hyper SELECT g, g FROM generate_series(0,1000) g;
+-- Sanity checking of the chunk_drop_replica API
+\set ON_ERROR_STOP 0
+-- Check that it doesn't work in a read only transaction
+SET default_transaction_read_only TO on;
+SELECT _timescaledb_internal.chunk_drop_replica(NULL, NULL);
+ERROR:  cannot execute chunk_drop_replica() in a read-only transaction
+RESET default_transaction_read_only;
+-- NULL input for chunk id errors out
+SELECT _timescaledb_internal.chunk_drop_replica(NULL, NULL);
+ERROR:  invalid chunk relation
+-- Specifying any regular hypertable instead of chunk errors out
+SELECT _timescaledb_internal.chunk_drop_replica('public.metrics', NULL);
+ERROR:  invalid chunk relation
+-- Specifying regular hypertable chunk on a proper data node errors out
+SELECT _timescaledb_internal.chunk_drop_replica('_timescaledb_internal._hyper_1_1_chunk', 'data_node_1');
+ERROR:  "_hyper_1_1_chunk" is not a valid remote chunk
+-- Specifying non-existent chunk on a proper data node errors out
+SELECT _timescaledb_internal.chunk_drop_replica('_timescaledb_internal._dist_hyper_700_38_chunk', 'data_node_1');
+ERROR:  relation "_timescaledb_internal._dist_hyper_700_38_chunk" does not exist at character 49
+-- Get the last chunk for this hypertable
+SELECT ch1.schema_name|| '.' || ch1.table_name as "CHUNK_NAME", ch1.id "CHUNK_ID" FROM
+_timescaledb_catalog.chunk ch1, _timescaledb_catalog.hypertable ht WHERE ch1.hypertable_id = ht.id
+AND ht.table_name = 'mvcp_hyper' ORDER BY ch1.id desc LIMIT 1 \gset
+-- Specifying wrong node name errors out
+SELECT _timescaledb_internal.chunk_drop_replica(:'CHUNK_NAME', 'bad_node');
+ERROR:  server "bad_node" does not exist
+-- This chunk contains only one entry as of now
+SELECT * FROM :CHUNK_NAME;
+ time | value 
+------+-------
+ 1000 |  1000
+(1 row)
+
+-- Specifying NULL node name along with proper chunk errors out
+SELECT _timescaledb_internal.chunk_drop_replica(:'CHUNK_NAME', NULL);
+ERROR:  data node name cannot be NULL
+\set ON_ERROR_STOP 1
+-- Check the current primary foreign server for this chunk, that will change
+-- post the chunk_drop_replica call
+SELECT foreign_server_name FROM information_schema.foreign_tables WHERE
+    foreign_table_name = split_part(:'CHUNK_NAME', '.', 2);
+ foreign_server_name 
+---------------------
+ data_node_3
+(1 row)
+
+-- Drop one replica of a valid chunk. Should succeed
+SELECT _timescaledb_internal.chunk_drop_replica(:'CHUNK_NAME', 'data_node_3');
+ chunk_drop_replica 
+--------------------
+ 
+(1 row)
+
+-- The primary foreign server should be updated now
+SELECT foreign_server_name FROM information_schema.foreign_tables WHERE
+    foreign_table_name = split_part(:'CHUNK_NAME', '.', 2);
+ foreign_server_name 
+---------------------
+ data_node_1
+(1 row)
+
+-- Number of replicas should have been reduced by 1
+SELECT count(*) FROM _timescaledb_catalog.chunk_data_node WHERE chunk_id = :'CHUNK_ID';
+ count 
+-------
+     2
+(1 row)
+
+-- Ensure that INSERTs still work on this mvcp_hyper table into this chunk
+INSERT INTO mvcp_hyper VALUES (1001, 1001);
+-- Ensure that SELECTs are able to query data from the above chunk
+SELECT count(*) FROM mvcp_hyper WHERE time >= 1000;
+ count 
+-------
+     2
+(1 row)
+
+-- Check that chunk_drop_replica works with compressed chunk
+SELECT substr(compress_chunk(:'CHUNK_NAME')::TEXT, 1, 29);
+            substr             
+-------------------------------
+ _timescaledb_internal._dist_h
+(1 row)
+
+-- Drop one replica of a valid chunk. Should succeed on another datanode
+SELECT _timescaledb_internal.chunk_drop_replica(:'CHUNK_NAME', 'data_node_2');
+ chunk_drop_replica 
+--------------------
+ 
+(1 row)
+
+-- Number of replicas should have been reduced by 1
+SELECT count(*) FROM _timescaledb_catalog.chunk_data_node WHERE chunk_id = :'CHUNK_ID';
+ count 
+-------
+     1
+(1 row)
+
+-- Decompress before checking INSERTs
+SELECT substr(decompress_chunk(:'CHUNK_NAME')::TEXT, 1, 29);
+            substr             
+-------------------------------
+ _timescaledb_internal._dist_h
+(1 row)
+
+-- Ensure that INSERTs still work on this mvcp_hyper table into this chunk
+INSERT INTO mvcp_hyper VALUES (1002, 1002);
+-- Ensure that SELECTs are able to query data from the above chunk
+SELECT count(*) FROM mvcp_hyper WHERE time >= 1000;
+ count 
+-------
+     3
+(1 row)
+
+\set ON_ERROR_STOP 0
+-- Drop one replica of a valid chunk. Should not succeed on last datanode
+SELECT _timescaledb_internal.chunk_drop_replica(:'CHUNK_NAME', 'data_node_1');
+ERROR:  cannot drop the last chunk replica
+\set ON_ERROR_STOP 1
+DROP table mvcp_hyper;
--- a/tsl/test/shared/sql/chunk_move_copy.sql
+++ b/tsl/test/shared/sql/chunk_move_copy.sql
@@ -1,0 +1,91 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+
+CREATE TABLE mvcp_hyper (time bigint NOT NULL, value integer);
+SELECT table_name FROM create_distributed_hypertable('mvcp_hyper', 'time',
+        chunk_time_interval => 200, replication_factor => 3);
+
+-- Enable compression so that we can test dropping of compressed chunks
+ALTER TABLE mvcp_hyper  SET (timescaledb.compress, timescaledb.compress_orderby='time DESC');
+
+INSERT INTO mvcp_hyper SELECT g, g FROM generate_series(0,1000) g;
+
+-- Sanity checking of the chunk_drop_replica API
+
+\set ON_ERROR_STOP 0
+-- Check that it doesn't work in a read only transaction
+SET default_transaction_read_only TO on;
+SELECT _timescaledb_internal.chunk_drop_replica(NULL, NULL);
+RESET default_transaction_read_only;
+
+-- NULL input for chunk id errors out
+SELECT _timescaledb_internal.chunk_drop_replica(NULL, NULL);
+
+-- Specifying any regular hypertable instead of chunk errors out
+SELECT _timescaledb_internal.chunk_drop_replica('public.metrics', NULL);
+
+-- Specifying regular hypertable chunk on a proper data node errors out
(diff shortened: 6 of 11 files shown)


sql/ddl_api.sql
```diff
@@ -213,3 +213,10 @@ CREATE OR REPLACE PROCEDURE refresh_continuous_aggregate(
     window_start             "any",
     window_end               "any"
 ) LANGUAGE C AS '@MODULE_PATHNAME@', 'ts_continuous_agg_refresh';
+
+-- Drop the specified chunk replica on the specified data node
+CREATE OR REPLACE FUNCTION  drop_chunk_replica(
```
k-rus, line 218: This file contains only public API. I suggest to move this function to `ddl_internal.sql`

k-rus (commented):
Can you test when a chunk is dropped from the data node, which is used as the primate foreign server, and demonstrate that the chunk's foreign table is updated?

Can you test that the distributed hypertable and the chunk are still operational after the removal, i.e., it is possible to insert into the chunk and select?

nikkhils:
> Can you test when a chunk is dropped from the data node, which is used as the primate foreign server, and demonstrate that the chunk's foreign table is updated?
> 

I didn't see a way to check which foreign server is primary for a given chunk. Is there a function to check that? Otherwise this will be difficult to implement at the test case level.

> Can you test that the distributed hypertable and the chunk are still operational after the removal, i.e., it is possible to insert into the chunk and select?

Ok, good idea. Will add this.

sql/ddl_api.sql
nikkhils: Ok, done

k-rus:
> > Can you test when a chunk is dropped from the data node, which is used as the primate foreign server, and demonstrate that the chunk's foreign table is updated?
> 
> I didn't see a way to check which foreign server is primary for a given chunk. Is there a function to check that? Otherwise this will be difficult to implement at the test case level.

@nikkhils You can print this information by selecting  from `information_schema.foreign_tables`. For example:
```sql
select * from information_schema.foreign_tables;
 foreign_table_catalog | foreign_table_schema  |  foreign_table_name   | foreign_server_catalog | foreign_server_name
-----------------------+-----------------------+-----------------------+------------------------+---------------------
 access_node           | _timescaledb_internal | _dist_hyper_1_2_chunk | access_node            | data_node_2
 access_node           | _timescaledb_internal | _dist_hyper_1_6_chunk | access_node            | data_node_2
 access_node           | _timescaledb_internal | _dist_hyper_1_1_chunk | access_node            | data_node_1
 access_node           | _timescaledb_internal | _dist_hyper_1_5_chunk | access_node            | data_node_1
(4 rows)
```

tsl/src/chunk.c
```diff
@@ -259,3 +260,94 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+						get_rel_name(chunk_id))));
+
+	/*
+	 * Drop chunk on the data node using a regular "DROP TABLE". CASCADE is
+	 * not required and compressed chunk (if any) should also get dropped.
+	 */
+	drop_cmd = psprintf("DROP TABLE %s.%s",
+						quote_identifier(chunk->fd.schema_name.data),
+						quote_identifier(chunk->fd.table_name.data));
+	data_nodes = lappend(data_nodes, (char *) node_name);
```
k-rus, line 342: `lappend` is confusing here, since the list is supposed to contain only one node, right? I suggest to create the list with one element instead. E.g., with `make_list1`.

tsl/src/chunk.c
```diff
@@ -259,3 +260,94 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
 
 	return num_results;
 }
+
+/*
+ * drop_chunk_replica:
+ *
+ * This function drops a chunk on a specified data node. It then
+ * removes the metadata about the datanode, chunk association on the
```
k-rus, line 268:
```suggestion
 * removes the metadata about the data node, chunk association on the
```

I am confused by this comment. Will it be correct to say?:
```suggestion
 * removes the metadata about the association of the chunk to the data node on the
```

k-rus (commented):
LGTM. Some nit comments.

The regression tests are failing on CI, thus I haven't reviewed the tests in details. It would be great to fix the regression tests before next review round.

nikkhils:
> > > Can you test when a chunk is dropped from the data node, which is used as the primate foreign server, and demonstrate that the chunk's foreign table is updated?
> > 
> > 
> > I didn't see a way to check which foreign server is primary for a given chunk. Is there a function to check that? Otherwise this will be difficult to implement at the test case level.
> 
> @nikkhils You can print this information by selecting from `information_schema.foreign_tables`. For example:
> 
> ```sql
> select * from information_schema.foreign_tables;
>  foreign_table_catalog | foreign_table_schema  |  foreign_table_name   | foreign_server_catalog | foreign_server_name
> -----------------------+-----------------------+-----------------------+------------------------+---------------------
>  access_node           | _timescaledb_internal | _dist_hyper_1_2_chunk | access_node            | data_node_2
>  access_node           | _timescaledb_internal | _dist_hyper_1_6_chunk | access_node            | data_node_2
>  access_node           | _timescaledb_internal | _dist_hyper_1_1_chunk | access_node            | data_node_1
>  access_node           | _timescaledb_internal | _dist_hyper_1_5_chunk | access_node            | data_node_1
> (4 rows)
> ```

Ok, thanks. I have used it in the test now.

tsl/src/chunk.c
nikkhils: Ok, will use `list_make1`. Thanks

tsl/src/chunk.c
nikkhils: Done

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+ *
+ * This function drops a chunk on a specified data node. It then
+ * removes the metadata about the association of the chunk to this
+ * datanode on the access node.
+ */
+Datum
+drop_chunk_replica(PG_FUNCTION_ARGS)
+{
+	const char *node_name = PG_ARGISNULL(0) ? NULL : NameStr(*PG_GETARG_NAME(0));
+	Oid chunk_id = PG_ARGISNULL(1) ? InvalidOid : PG_GETARG_OID(1);
```
k-rus, line 275: I just noticed this line. 
`chunk_id` usually means the chunk id allocated by TimescaleDB. So I suggest to rename it, since it is relational id, e.g.:
```suggestion
	Oid chunk_relid = PG_ARGISNULL(1) ? InvalidOid : PG_GETARG_OID(1);
```

nikkhils: Ok, done.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+
+	/*
+	 * There should be atleast one surviving replica after the deletion here.
+	 *
+	 * We could fetch the corresponding hypertable and check its
+	 * replication_factor. But the user of this function is using it
+	 * to move chunk from one datanode to another and is well aware of
+	 * the replication_factor requirements
+	 */
+	if (list_length(replicas) <= 1)
```
k-rus, line 328: Can you also add a warning if the number of replicas will be less than the replication factor? E.g.
```c
if (list_length(replicas) < ht->fd.replication_factor)
		ereport(WARNING,
				(errcode(ERRCODE_TS_INSUFFICIENT_NUM_DATA_NODES),
				 errmsg("insufficient number of data nodes"),
				 errdetail("There are not enough data nodes to replicate chunks according to the"
						   " configured replication factor."),
```

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_INSUFFICIENT_NUM_DATA_NODES),
+				 errmsg("dropping last chunk could lead to data loss")));
```
k-rus, line 331: I suggest to give an error in the error message and the current text can go into details instead. For example:
```suggestion
				 errmsg("cannot drop the last chunk replica"),
				 errdetail("Dropping the last chunk  replica could lead to data loss.")));
```

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+
+	/*
+	 * Drop chunk on the data node using a regular "DROP TABLE". CASCADE is
```
k-rus, line 334: I am not sure if the comment about `CASCADE` is fully correct. If a user created any database objects on the chunk directly, then `CASCADE` is necessary to be able to successfully execute the drop table. The drop chunk, which is called by `DROP TABLE`, only removes objects created by TimescaleDB.

The current approach is that the user should separately drop user-defined objects on the chunk before, e.g., calling this API function.

I suggest to improve the comment that it is not needed for removing database objects such as compressed chunk. Otherwise it might mislead the code reader.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+	 * not required and compressed chunk (if any) should also get dropped.
+	 */
+	drop_cmd = psprintf("DROP TABLE %s.%s",
+						quote_identifier(chunk->fd.schema_name.data),
+						quote_identifier(chunk->fd.table_name.data));
+	data_nodes = list_make1((char *) node_name);
+	ts_dist_cmd_invoke_on_data_nodes(drop_cmd, data_nodes, true);
+
+	/*
+	 * This chunk might have this datanode as primary, remove that association
```
k-rus, line 344:
```suggestion
	 * This chunk might have this data node as primary, change that association
```
It will be consistent with other comments to use `data node` instead of `datanode`.
The association is not just removed by changed to another data node.

sql/ddl_internal.sql
```diff
@@ -25,3 +25,10 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.refresh_continuous_aggregate(
     continuous_aggregate     REGCLASS,
     hypertable_chunk         REGCLASS
 ) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_continuous_agg_refresh_chunk' LANGUAGE C VOLATILE;
+
+-- Drop the specified chunk replica on the specified data node
+CREATE OR REPLACE FUNCTION  _timescaledb_internal.drop_chunk_replica(
+    node_name               NAME,
+    chunk                   REGCLASS
+) RETURNS VOID
```
k-rus, line 33: Since the function name is about chunk, then I suggest to put the chunk as the first argument and then the data node name. Other API functions in this file are about data nodes and those the data node is the first argument.

tsl/test/shared/expected/chunk_move_copy.out
```diff
@@ -0,0 +1,112 @@
...
+ count 
+-------
+     1
+(1 row)
+
+\set ON_ERROR_STOP 0
+-- Drop one replica of a valid chunk. Should not succeed on last datanode
+SELECT _timescaledb_internal.drop_chunk_replica('data_node_1', :'CHUNK_NAME');
+ERROR:  dropping last chunk could lead to data loss
+DROP table hyper;
```
k-rus, line 112: Can you set back to stop on error?

tsl/test/shared/sql/chunk_move_copy.sql
```diff
@@ -0,0 +1,78 @@
...
+
+-- Drop one replica of a valid chunk. Should succeed
+SELECT _timescaledb_internal.drop_chunk_replica('data_node_3', :'CHUNK_NAME');
+
+-- The primary foreign server should be updated now
+SELECT foreign_server_name FROM information_schema.foreign_tables WHERE
+    foreign_table_name = split_part(:'CHUNK_NAME', '.', 2);
+
+-- Number of replicas should have been reduced by 1
+SELECT count(*) from _timescaledb_catalog.chunk_data_node where chunk_id = :'CHUNK_ID';
```
k-rus, line 56:
```suggestion
SELECT count(*) FROM _timescaledb_catalog.chunk_data_node where chunk_id = :'CHUNK_ID';
```

k-rus (commented):
Thank you for fixing the regression tests. Finally this PR is ready for final review :)

I have more comments in the code.
Also I noticed inconsistent formatting of SQL code in the test. It is good to be consistent within the file and actually with other files. There is no standard decided yet, but many uses `pg_format`.

Can you add two more tests:
- Test when the chunk replica to remove is compressed
- Try to call this function from a read only transaction.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+Datum
+drop_chunk_replica(PG_FUNCTION_ARGS)
+{
+	const char *node_name = PG_ARGISNULL(0) ? NULL : NameStr(*PG_GETARG_NAME(0));
+	Oid chunk_relid = PG_ARGISNULL(1) ? InvalidOid : PG_GETARG_OID(1);
+	List *data_nodes, *replicas;
+	ForeignServer *server;
+	Chunk *chunk;
+	ChunkDataNode *cdn;
+	char *drop_cmd;
```
k-rus, line 280: Will it be correct?
```suggestion
	const char *drop_cmd;
```

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+	if (!OidIsValid(chunk_relid))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("Invalid chunk specification %u", chunk_relid)));
+
+	chunk = ts_chunk_get_by_relid(chunk_relid, false);
+
+	if (NULL == chunk)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("Invalid chunk specified")));
```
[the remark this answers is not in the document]
nikkhils: The chunk_relid could point to anything, not just a relation. Also the above will crash if `get_rel_name` returns NULL, which is a possibility.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+
+	/* It has to be a foreign table chunk */
+	if (chunk->relkind != RELKIND_FOREIGN_TABLE)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("\"%s\" is not a valid remote chunk", get_rel_name(chunk_relid))));
+
+	server = data_node_get_foreign_server(node_name, ACL_USAGE, true, false);
```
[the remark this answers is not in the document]
nikkhils: It's good to validate the server via ACL and all.

tsl/src/chunk.c
nikkhils: Please read the comment above this. We are not fetching the HT at all here. Also, this API will be used as part of `chunk move` where the user knows that temporarily replication factor might not be honored.

tsl/src/chunk.c
nikkhils: Fine.. changed

tsl/src/chunk.c
nikkhils: As per the FORGE environment, DNs are not supposed to be directly usable by users. If users want to still go ahead and do things on the DN then it's on them to also clean up appropriately.

tsl/src/chunk.c
nikkhils: There are other locations in other files where `datanode` is used. But changed to `data node` now..

sql/ddl_internal.sql
nikkhils: Hmm, will have to change lots of files for this. Will do..

tsl/test/shared/expected/chunk_move_copy.out
nikkhils: Done

tsl/test/shared/sql/chunk_move_copy.sql
nikkhils: why not the "where"? :-)

sql/ddl_internal.sql
erimatnor: I agree with Ruslan. We often try to make the main object the first parameter in our API functions. I think it is good with consistence, otherwise it is a slippery slope IMO.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+	List *data_nodes, *replicas;
+	ForeignServer *server;
+	Chunk *chunk;
+	ChunkDataNode *cdn;
+	char *drop_cmd;
+
+	if (!OidIsValid(chunk_relid))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("Invalid chunk specification %u", chunk_relid)));
```
erimatnor, line 285: We follow the PostgreSQL error style guide. Specifically in this case: "Use lower case for message wording, including the first letter of a primary error message" https://www.postgresql.org/docs/current/error-style-guide.html

Not sure what is meant by "specification" as this concerns a chunk relid. Also, there was already a check that this relid was `InvalidOid` (i.e, `0`), so no need to print it:

```suggestion
				 errmsg("invalid chunk relation")));
```

tsl/src/chunk.c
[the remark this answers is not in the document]
erimatnor: Same as above regarding error style. Also, FWIW, PostgreSQL errors often print the numerical Oid in cases such as this. For instance, for cache lookup failures:

```
elog(ERROR, "cache lookup failed for relation %u", relationId);
```

I often try to take inspiration from these PostgreSQL errors.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+				 errmsg("\"%s\" is not a valid remote chunk", get_rel_name(chunk_relid))));
+
+	server = data_node_get_foreign_server(node_name, ACL_USAGE, true, false);
+	Assert(NULL != server);
+
+	/* Early abort on missing permissions */
+	ts_hypertable_permissions_check(chunk_relid, GetUserId());
+
+	/* check that the chunk is indeed present on the specified data node */
+	cdn = ts_chunk_data_node_scan_by_chunk_id_and_node_name(chunk->fd.id,
```
erimatnor, line 306: I think we already have this information in `chunk->data_nodes`, so could just check `node_name` against that list.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("Chunk \"%s\" does not exist on data node \"%s\"",
+						get_rel_name(chunk_relid),
+						node_name)));
+
+	/* Ensure that this chunk remains replicated so as to avoid data loss */
+	replicas = ts_chunk_data_node_scan_by_chunk_id(chunk->fd.id, CurrentMemoryContext);
+
+	/*
+	 * There should be atleast one surviving replica after the deletion here.
```
erimatnor, line 321: Nit:

```suggestion
	 * There should be at least one surviving replica after the deletion here.
```

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+
+	if (!cdn)
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("Chunk \"%s\" does not exist on data node \"%s\"",
+						get_rel_name(chunk_relid),
+						node_name)));
+
+	/* Ensure that this chunk remains replicated so as to avoid data loss */
+	replicas = ts_chunk_data_node_scan_by_chunk_id(chunk->fd.id, CurrentMemoryContext);
```
erimatnor, line 318: Again, check if you cannot already rely in `chunk->data_nodes` list for this.

tsl/src/chunk.c
erimatnor: I agree with @nikkhils that we probably should elide these  types of warnings. Especially considering the fact that this might be invoked on many chunks as part of a single "move" operation. We don't want to print one warning per chunk in that case.

tsl/src/chunk.c
erimatnor: I think it is fine to elide CASCADE, especially since it is clear from the comment that we do this consciously.  We can always revisit if it turns out to be a problem.

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+
+	/* check that the chunk is indeed present on the specified data node */
+	cdn = ts_chunk_data_node_scan_by_chunk_id_and_node_name(chunk->fd.id,
+															server->servername,
+															CurrentMemoryContext);
+
+	if (!cdn)
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("Chunk \"%s\" does not exist on data node \"%s\"",
```
erimatnor, line 313:
```suggestion
				 errmsg("chunk \"%s\" does not exist on data node \"%s\"",
```

tsl/src/chunk.c
```diff
@@ -259,3 +260,92 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+
+	/*
+	 * Drop chunk on the data node using a regular "DROP TABLE". CASCADE is
+	 * not required and compressed chunk (if any) should also get dropped.
+	 */
+	drop_cmd = psprintf("DROP TABLE %s.%s",
+						quote_identifier(chunk->fd.schema_name.data),
+						quote_identifier(chunk->fd.table_name.data));
+	data_nodes = list_make1((char *) node_name);
+	ts_dist_cmd_invoke_on_data_nodes(drop_cmd, data_nodes, true);
```
erimatnor, line 341: I think you might need (should) handle the result, check for errors, and clean it up (`ts_dist_cmd_close_response`).

erimatnor (commented):
Reviewed the code part, will look into tests later. LGTM, although a few nits and error message tweaks. The one thing to look into is whether you need to handle the request results when invoking the remote command.

tsl/test/shared/sql/chunk_move_copy.sql
k-rus: I haven't noticed that it is there 😃 Yes, definitely

tsl/src/chunk.c
k-rus: My note is about the code comment. Now it sounds that `CASCADE` will take care of non-TimescaleDB objects too. The comment is important for code maintainability and should help a developer to think in the correct direction and not mislead.

tsl/src/chunk.c
k-rus: > this API will be used as part of `chunk move` where the user knows that temporarily replication factor might not be honored.

In the case of the move the replica will be already added to this moment. It is similar to dropping the last chunk replica. Using your reasoning no error should be produced on dropping the last replica, however as I noted the chunk already should be copied, so it is correct to produce the error.

tsl/src/chunk.c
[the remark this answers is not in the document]
k-rus: Can you add a comment explaining the need for this call?

sql/ddl_internal.sql
erimatnor: Actually, thinking about this some more, maybe the parameter order makes sense if we'd like to expand this to drop multiple chunk replicas in one go:

```
drop_chunk_replicas('dn1', 'chunk1');
drop_chunk_replicas('dn1', 'chunk1', 'chunk2');
```

For the latter case we could make it a vararg function. Or, we could have one `drop_chunk_replica` and one `drop_chunk_replicas` function (obviuosly, the former would just call the latter with one chunk).

tsl/src/chunk.c
k-rus: @erimatnor In the case of move the warning will be fired only if the chunk was under-replicated before calling the move.

tsl/src/chunk.c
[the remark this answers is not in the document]
nikkhils: An indirect challenge due to printing the OID in the error message is that test regression output file cannot handle variation in the OID values. We cannot expect the OID to remain the same, especially in shared test cases.

tsl/src/chunk.c
nikkhils: `CASCADE` will indeed drop other objects associated with this chunk. That's what `CASCADE` is supposed to do.

sql/ddl_internal.sql
nikkhils: Have moved `chunk` as the first argument now..

tsl/src/chunk.c
nikkhils: Ok

tsl/src/chunk.c
nikkhils: Changed

tsl/src/chunk.c
k-rus: My suggestion is to improve the comment, so it will not mislead.

tsl/src/chunk.c
nikkhils: Fixed.

tsl/src/chunk.c
nikkhils: I think we can. Fixed.

tsl/src/chunk.c
nikkhils: Ok, that's better.

tsl/src/chunk.c
nikkhils: done

tsl/src/chunk.c
nikkhils: Handled by using `ts_dist_cmd_run_on_data_nodes`

tsl/test/shared/sql/chunk_move_copy.sql
nikkhils: There are many other test files with such inconsistencies. Anyways, fixed...

tsl/test/shared/sql/chunk_move_copy.sql
```diff
@@ -0,0 +1,78 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+
+CREATE TABLE hyper (time bigint NOT NULL, value integer);
```
k-rus, line 5: My apology for bringing a new comment. I just realised that this test will share the same database and run in parallel with other tests, thus it is good create a table with more unique name, e.g., the name of the test file, which is what you see in other test files.

k-rus: Also note that `include/shared_setup.sql` is executed before calling this file.

k-rus: > Also note that `include/shared_setup.sql` is executed before calling this file.

I believe you are already utilising the fact this shared script creates data nodes.

tsl/src/chunk.c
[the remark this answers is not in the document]
nikkhils: Solved this by printing the OID in the `errdetail` part.

tsl/src/chunk.c
nikkhils: I don't think it's our job to add a comment about users adding additional objects.. however, since you are insisting still, I am adding it...

k-rus: @nikkhils My comment is that current formulation is misleading. You might remove completely this sentence. You can just say that `DROP TABLE` will drop the chunk. What the drop chunk does can be checked in the corresponding implementation and is not necessary to describe here, imho.

sql/ddl_internal.sql
erimatnor: That works for now @nikkhils. We fan always change later.

tsl/src/chunk.c
erimatnor: @k-rus but if we assume the chunk replica is moved (and not just deleted), then we aren't changing the replication state (i.e., not making it worse) compared to before the operation. So, it doesn't seem worthwhile to warn about a situation that already existed prior to the operation. Or am I missing something?

tsl/test/shared/expected/chunk_move_copy.out
```diff
@@ -0,0 +1,112 @@
+-- This file and its contents are licensed under the Timescale License.
```
erimatnor, line 1: Mental note; I think Ruslan added a test file `dist_chunk.sql` so we might want to merge this test into that one later.

tsl/test/shared/sql/chunk_move_copy.sql
nikkhils: done...

src/cross_module_fn.h
```diff
@@ -129,6 +129,8 @@ typedef struct CrossModuleFunctions
 	Datum (*chunk_set_default_data_node)(PG_FUNCTION_ARGS);
 	Datum (*create_chunk)(PG_FUNCTION_ARGS);
 	Datum (*show_chunk)(PG_FUNCTION_ARGS);
+	Datum (*drop_chunk_replica)(PG_FUNCTION_ARGS);
+
```
erimatnor, line 133: Optional: would recommend using `PGFunction `, but we aren't consistent about this in existing code, so we could just clean up later.

Note the extra line, however.

```suggestion
	Datum (*drop_chunk_replica)(PG_FUNCTION_ARGS);
```

tsl/src/chunk.c
```diff
@@ -259,3 +260,102 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+				 errmsg("\"%s\" is not a valid remote chunk", get_rel_name(chunk_relid))));
+
+	server = data_node_get_foreign_server(node_name, ACL_USAGE, true, false);
+	Assert(NULL != server);
+
+	/* Early abort on missing permissions */
+	ts_hypertable_permissions_check(chunk_relid, GetUserId());
+
+	/* check that the chunk is indeed present on the specified data node */
+	foreach (lc, chunk->data_nodes)
```
erimatnor, line 308: Optional: Could expose this is `ts_chunk_has_data_node` and put in `chunk.c`

erimatnor (approved):
Did second pass to include tests. LGTM, so approving although I left a couple of nits and comments.

tsl/src/chunk.c
k-rus: It is worth to warn if the function is used without move. I mean it is the same as returning the error of deleting the last chunk replica. It sounds strange to me to do one but not the other, while the other places do both.

The only case when it might be confusing to give the warning is when the chunk was under-replicated before the move. May be it is good to pass a parameter to this SQL function to avoid warning?

tsl/src/chunk.c
nikkhils: Done

src/cross_module_fn.h
nikkhils: Done. Using `PGFunction` now.

tsl/test/shared/expected/chunk_move_copy.out
nikkhils: I see that he has added his tests in `chunk_api.sql.in`

nikkhils:
Finally, renamed the API to `chunk_drop_replica` from the earlier `drop_chunk_replica` for uniformity with other APIs

tsl/src/chunk.c
erimatnor: I believe this function will be use in two cases, either you:

1. Move the chunk (number of chunk replicas doesn't change)
2. Copy the chunk (number of chunk replicas increases).

In none of the cases will we delete the last chunk replica.

I think, in practice, the error raised when trying to delete the last chunk replica will never be raised if this function is used properly; the chunk would have been copied or moved prior to calling this function, so there should be at least one an extra replica in addition to this one. So, the previous check for last chunk replica is a pure safety check in case this is called outside the scope of the two cases above.

tsl/test/shared/expected/chunk_move_copy.out
erimatnor: Sorry, I meant in his yet-to-be-merged PR.

src/chunk.c
```diff
@@ -925,6 +925,30 @@ ts_chunk_get_data_node_name_list(const Chunk *chunk)
 	return datanodes;
 }
 
+bool
+ts_chunk_has_data_node(Chunk *chunk, const char *node_name)
```
erimatnor, line 929: Nit: I guess chunk isn't modified so can be const too.

nikkhils: Done

k-rus (approved):

pmwkaa (approved):

PR #3165, opened 2021-04-28
Use no-unused-command-line-argument flag

nikkhils:
We started inheriting CFLAGS from postgresql via ba51adc3. Although we
remove "-W" flags, we still end up accepting other options like "-L"
from postgresql. This can cause compilation to fail with compilers
like Clang which enable reporting of errors on unused command line
arguments.

This patch conditionally adds -Wno-unused-command-line-argument in
our cmake files to avoid unhelpful unused-command-line-argument
warnings/errors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,11 @@ if(CMAKE_C_COMPILER_ID MATCHES "GNU|AppleClang|Clang")
 
   # These flags are just supported on some of the compilers, so we
   # check them before adding them.
+  check_c_compiler_flag(-Wno-unused-command-line-argument CC_SUPPORTS_NO_UNUSED_CLI_ARG)
+  if(CC_SUPPORTS_NO_UNUSED_CLI_ARG)
+    add_compile_options(-Wno-unused-command-line-argument)
+  endif()
+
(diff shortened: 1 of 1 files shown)


mkindahl (approved):

CMakeLists.txt
```diff
@@ -133,6 +133,13 @@ if(CMAKE_C_COMPILER_ID MATCHES "GNU|AppleClang|Clang")
 
   # These flags are just supported on some of the compilers, so we
   # check them before adding them.
+  check_c_compiler_flag(-Wno-unused-command-line-argument CC_SUPPORTS_NO_UNUSED_CLI_ARG)
+  if(CC_SUPPORTS_NO_UNUSED_CLI_ARG)
+    add_compile_options(-Wno-unused-command-line-argument)
+  else()
+    message(STATUS "Compiler does not support -Wunused-command-line-argument")
```
svenklemm, line 140:
```suggestion
```
I think we can skip the status message

svenklemm (approved):

CMakeLists.txt
nikkhils: Ok, done.

PR #3171, opened 2021-04-28
Create chunk table from access node

k-rus:
Creates a table for chunk replica on the given data node. The table 
gets the same schema and name as the chunk. The created chunk replica 
table is not added into metadata on the access node or data node.

The primary goal is to use it during copy/move chunk.

### DIFF
--- a/tsl/src/chunk.c
+++ b/tsl/src/chunk.c
@@ -32,11 +32,34 @@
 #include <chunk_data_node.h>
 #include <extension.h>
 #include <errors.h>
+#include <error_utils.h>
+#include <hypertable_cache.h>
 
 #include "chunk.h"
+#include "chunk_api.h"
 #include "data_node.h"
 #include "deparse.h"
 #include "remote/dist_commands.h"
+
+static bool
+chunk_match_data_node_by_server(const Chunk *chunk, const ForeignServer *server)
+{
+	bool server_found = false;
+	ListCell *lc;
+
+	foreach (lc, chunk->data_nodes)
+	{
+		ChunkDataNode *cdn = lfirst(lc);
+
+		if (cdn->foreign_server_oid == server->serverid)
+		{
+			server_found = true;
+			break;
+		}
+	}
+
+	return server_found;
+}
 
 static bool
 chunk_set_foreign_server(Chunk *chunk, ForeignServer *new_server)
@@ -49,21 +72,8 @@ chunk_set_foreign_server(Chunk *chunk, ForeignServer *new_server)
 	CatalogSecurityContext sec_ctx;
 	Oid old_server_id;
 	long updated;
-	ListCell *lc;
-	bool new_server_found = false;
 
-	foreach (lc, chunk->data_nodes)
-	{
-		ChunkDataNode *cdn = lfirst(lc);
-
-		if (cdn->foreign_server_oid == new_server->serverid)
-		{
-			new_server_found = true;
-			break;
-		}
-	}
-
-	if (!new_server_found)
+	if (!chunk_match_data_node_by_server(chunk, new_server))
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 				 errmsg("chunk \"%s\" does not exist on data node \"%s\"",
@@ -258,4 +268,66 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
 	FreeExecutorState(estate);
 
 	return num_results;
+}
+
+static bool
+chunk_is_distributed(const Chunk *chunk)
+{
+	return chunk->relkind == RELKIND_FOREIGN_TABLE;
+}
+
+Datum
+chunk_create_replica_table(PG_FUNCTION_ARGS)
+{
+	Oid chunk_relid;
+	const char *data_node_name;
+	const Chunk *chunk;
+	const Hypertable *ht;
+	const ForeignServer *server;
+	Cache *hcache = ts_hypertable_cache_pin();
+
+	TS_PREVENT_FUNC_IF_READ_ONLY();
+
+	GETARG_NOTNULL_OID(chunk_relid, 0, "chunk");
+	GETARG_NOTNULL_NULLABLE(data_node_name, 1, "data node name", CSTRING);
+
+	chunk = ts_chunk_get_by_relid(chunk_relid, false);
+	if (chunk == NULL)
+	{
+		const char *rel_name = get_rel_name(chunk_relid);
+		if (rel_name == NULL)
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("oid \"%u\" is not a chunk", chunk_relid)));
+		else
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("relation \"%s\" is not a chunk", rel_name)));
+	}
+	if (!chunk_is_distributed(chunk))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("chunk \"%s\" doesn't belong to a distributed hypertable",
+						get_rel_name(chunk_relid))));
+
+	ht = ts_hypertable_cache_get_entry(hcache, chunk->hypertable_relid, CACHE_FLAG_NONE);
+	ts_hypertable_permissions_check(ht->main_table_relid, GetUserId());
+
+	/* Check the given data node exists */
+	server = data_node_get_foreign_server(data_node_name, ACL_USAGE, true, false);
+	/* Find if hypertable is attached to the data node and return an error otherwise */
+	data_node_hypertable_get_by_node_name(ht, data_node_name, true);
+
+	if (chunk_match_data_node_by_server(chunk, server))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("chunk \"%s\" already exists on data node \"%s\"",
+						get_rel_name(chunk_relid),
+						data_node_name)));
+
+	chunk_api_call_create_empty_chunk_table(ht, chunk, data_node_name);
+
+	ts_cache_release(hcache);
+
+	PG_RETURN_VOID();
 }
--- a/tsl/src/chunk_api.c
+++ b/tsl/src/chunk_api.c
@@ -27,12 +27,10 @@
 
 #include <catalog.h>
 #include <compat.h>
-#include <chunk.h>
 #include <chunk_data_node.h>
 #include <errors.h>
 #include <error_utils.h>
 #include <hypercube.h>
-#include <hypertable.h>
 #include <hypertable_cache.h>
 
 #include "remote/async.h"
@@ -1671,4 +1669,25 @@ chunk_create_empty_table(PG_FUNCTION_ARGS)
 	ts_cache_release(hcache);
 
 	PG_RETURN_BOOL(true);
+}
+
+#define CREATE_CHUNK_TABLE_NAME "create_chunk_table"
+
+void
+chunk_api_call_create_empty_chunk_table(const Hypertable *ht, const Chunk *chunk,
+										const char *node_name)
+{
+	const char *create_cmd =
+		psprintf("SELECT %s.%s($1, $2, $3, $4)", INTERNAL_SCHEMA_NAME, CREATE_CHUNK_TABLE_NAME);
+	const char *params[4] = { quote_qualified_identifier(NameStr(ht->fd.schema_name),
+														 NameStr(ht->fd.table_name)),
+							  chunk_api_dimension_slices_json(chunk, ht),
+							  NameStr(chunk->fd.schema_name),
+							  NameStr(chunk->fd.table_name) };
+
+	ts_dist_cmd_close_response(
+		ts_dist_cmd_params_invoke_on_data_nodes(create_cmd,
+												stmt_params_create_from_values(params, 4),
+												list_make1((void *) node_name),
+												true));
 }
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -40,7 +40,6 @@
 #include "remote/connection_cache.h"
 #include "data_node.h"
 #include "remote/utils.h"
-#include "hypertable.h"
 #include "hypertable_cache.h"
 #include "errors.h"
 #include "dist_util.h"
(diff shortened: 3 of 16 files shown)


tsl/src/chunk.c
```diff
@@ -259,3 +262,60 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("chunk \"%s\" doesn't belong to a distributed hypertable",
+						get_rel_name(chunk_relid))));
+
+	ht = ts_hypertable_cache_get_entry(hcache, chunk->hypertable_relid, CACHE_FLAG_NONE);
+	ts_hypertable_permissions_check(ht->main_table_relid, GetUserId());
+
+	/* Check the given data node exists */
+	data_node_get_foreign_server(data_node_name, ACL_USAGE, true, false);
```
erimatnor, line 302: I don't recall what the `true` and `false` arguments mean (which is the problem of such boolean parameters), but I assume one of them means "fail if not found", correct?

tsl/src/chunk.c
```diff
@@ -259,3 +262,60 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+	/* Find if hypertable is attached to the data node and return an error otherwise */
+	data_node_hypertable_get(ht, data_node_name, true);
+
+	cdn = ts_chunk_data_node_scan_by_chunk_id_and_node_name(chunk->fd.id,
```
erimatnor, line 306: The `chunk` object has a list of data nodes `chunk->data_nodes` already. I suggest checking against that list of nodes since that would avoid the extra scan.

tsl/src/chunk_api.c
```diff
@@ -1672,3 +1670,24 @@ chunk_create_empty_table(PG_FUNCTION_ARGS)
 
 	PG_RETURN_BOOL(true);
 }
+
+#define CREATE_CHUNK_TABLE_NAME "create_chunk_table"
+
+void
+chunk_api_call_create_empty_chunk_table(const Hypertable *ht, const Chunk *chunk,
```
erimatnor, line 1677: I am wondering if this code doesn't belong in the other function or `tsl/src/chunk.c`?

In most cases, `chunk_api.c` implements the SQL function API (PG_FUNCTIONS) that is called on the data node side, not the call-out code on the access node.

tsl/src/data_node.c
```diff
@@ -1185,16 +1184,31 @@ get_hypertable_data_node(Oid table_id, const char *node_name, bool owner_check,
...
 							node_name,
-							get_rel_name(table_id))));
+							get_rel_name(ht->main_table_relid))));
 	}
 
+	return hdn;
+}
+
+static HypertableDataNode *
+get_hypertable_data_node(Oid table_id, const char *node_name, bool owner_check, bool attach_check)
```
erimatnor, line 1201: I would prefer that this function share name with the other function but with, e.g., a suffix `_from_relid`. It is a bit confusing to just change the order of the words. So, the suggested names would be something like:

```
hypertable_data_node_get
hypertable_data_node_get_from_relid
```

tsl/test/shared/expected/dist_chunk.out
```diff
@@ -0,0 +1,55 @@
...
+SELECT _timescaledb_internal.create_chunk_replica_table('_timescaledb_internal._dist_hyper_15_27_chunk', 'data_node_1');
+ERROR:  relation "_timescaledb_internal._dist_hyper_15_27_chunk" does not exist at character 57
+SELECT _timescaledb_internal.create_chunk_replica_table('_timescaledb_internal._dist_hyper_15_67_chunk', 'data_node_4');
+ERROR:  server "data_node_4" does not exist
+BEGIN READ ONLY;
+SELECT _timescaledb_internal.create_chunk_replica_table('_timescaledb_internal._dist_hyper_15_67_chunk', 'data_node_3');
+ERROR:  cannot execute create_chunk_replica_table() in a read-only transaction
+COMMIT;
+\set ON_ERROR_STOP 1
+SELECT _timescaledb_internal.create_chunk_replica_table('_timescaledb_internal._dist_hyper_15_67_chunk', 'data_node_3');
```
erimatnor, line 45: It would be good with e.g., a `remote_exec` to see that this chunk table was actually created on the data node.

erimatnor:
LGTM. Just have a few suggestions for nits and tweaks. Functionality-wise, looks good.

tsl/src/chunk.c
k-rus: `true` is for `fail_on_aclcheck` and `false` is for `missing_ok`, i.e., opposite to fail if not found.

tsl/src/chunk.c
k-rus: Do you know if there is already a helper function for this check?

tsl/src/chunk_api.c
k-rus: Remotely calling `create_table` is also done here, so it was the main reason to place it here. Also I use few static functions from this file.

tsl/test/shared/expected/dist_chunk.out
k-rus: The second call gives an error on the corresponding data node that the relation already exists. Isn't it enough?

tsl/src/chunk_api.c
k-rus: And, yes, I first thought to have the complete implementation in `chunk.c`.

tsl/src/chunk.c
k-rus: Does it resolve your question?

tsl/src/chunk.c
k-rus: I created one

tsl/src/data_node.c
k-rus: This is the original name before this PR. It is an issue with GH that it shows it as a new name.

k-rus: I updated names of newly added functions in this PR to specify by what they retrieve or match. I didn't update this static function as it was named such originally.

tsl/src/chunk_api.c
```diff
@@ -1672,3 +1670,24 @@ chunk_create_empty_table(PG_FUNCTION_ARGS)
...
+{
+	const char *create_cmd =
+		psprintf("SELECT %s.%s($1, $2, $3, $4)", INTERNAL_SCHEMA_NAME, CREATE_CHUNK_TABLE_NAME);
+	const char *params[4] = { quote_qualified_identifier(NameStr(ht->fd.schema_name),
+														 NameStr(ht->fd.table_name)),
+							  chunk_api_dimension_slices_json(chunk, ht),
+							  NameStr(chunk->fd.schema_name),
+							  NameStr(chunk->fd.table_name) };
+
+	ts_dist_cmd_close_response(
```
erimatnor, line 1688: I think there's a macro for this `ts_dist_cmd_run_on_data_node` or similar.

tsl/test/shared/expected/dist_chunk.out
erimatnor: I guess, although as a reviewer it is non-obvious that the failure is generated by a table conflict on the DN as opposed to some kind of early check on the AN. For instance, the AN could have created some local metadata for the table so it fails on checking for the metadata. But it could still have failed to create the table object on the DN. You'd have to know the implementation details to understand whether this is the case or not. 

IMO, it is mentally easier to follow when there is a follow-up check to show that the object in fact was created.

erimatnor (approved):

tsl/src/chunk_api.c
k-rus: This macro doesn't with parameters.

tsl/src/chunk.c
erimatnor: Suggestion: add a comment to explain that this fails, similar to the comment for the lookup below.

tsl/test/shared/expected/dist_chunk.out
k-rus: @erimatnor Have you looked to the error?
```
ERROR:  [data_node_3]: relation "_dist_hyper_15_67_chunk" already exists
```
It clearly says that the failure is on the data node.

erimatnor: I am giving the feedback from the point of view of a reviewer. And, Yes, reviewers might not always notice or capture every detail, so just letting you know what I "expected to see" and things that are easy to miss otherwise.

Like with code, even if something is technically correct, the behavior isn't always obvious even though all the information is technically there (however subtly).

I approved, so this is certainly not a show-stopper so if you believe it is more effort than it is worth, then I'll let you be the judge of that. Just giving feedback.

tsl/src/chunk.c
k-rus: I haven't seen that we add comments for internal API function calls to explain its arguments. I checked that none of the calls to this function has such comment. So the suggestion is not consistent with current practice and might be good to discuss on the team level.

tsl/test/shared/expected/dist_chunk.out
k-rus: I added a comment to clarify this in the test.

tsl/src/chunk.c
erimatnor: Just a suggestion for clarity since I didn't understand myself. Never intended for this to be controversial or requiring team-level discussions.

tsl/src/data_node.c
erimatnor: Yes, the there was a function there previously that had this name, but you also introduced a _new_ public function with a different name but sharing the same basic functionality, So, _previously_ there was no inconsistency (since there was only on internal function) but _after_ this change there is an inconsistency. 

Just noting, you can keep as is.

tsl/src/chunk.c
```diff
@@ -259,3 +269,57 @@ chunk_invoke_drop_chunks(Oid relid, Datum older_than, Datum older_than_type)
...
+	TS_PREVENT_FUNC_IF_READ_ONLY();
+
+	GETARG_NOTNULL_OID(chunk_relid, 0, "chunk");
+	GETARG_NOTNULL_NULLABLE(data_node_name, 1, "data node name", CSTRING);
+
+	chunk = ts_chunk_get_by_relid(chunk_relid, false);
+	if (chunk == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("relation \"%s\" is not a chunk", get_rel_name(chunk_relid))));
```
nikkhils, line 298: get_rel_name returns NULL if it's not a relid. So the above ereport will crash.

tsl/test/shared/expected/dist_chunk.out
```diff
@@ -0,0 +1,59 @@
...
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+-- Test function _timescaledb_internal.create_chunk_replica_table
+-- A table for the first chunk will be created on the data node, where it is not present.
+SELECT chunk_name, data_nodes 
+FROM timescaledb_information.chunks 
+WHERE hypertable_name = 'dist_chunk_copy';
+       chunk_name        |        data_nodes         
+-------------------------+---------------------------
+ _dist_hyper_15_67_chunk | {data_node_1,data_node_2}
```
nikkhils, line 11: These chunk names are not guaranteed to be the same in `shared` test cases.

tsl/src/chunk.c
k-rus: @nikkhils It does not crash. It returns NULL. It is common way in our code now.

tsl/test/shared/expected/dist_chunk.out
k-rus: @nikkhils Since the data are inserted in the shared setup, the chunk name is guaranteed.

tsl/src/chunk.c
nikkhils: errmsg("relation NULL is not a chunk") <- this will crash

k-rus: It does not crash. It returns an error with not very nice message. It is good to fix on a general level, but I find it helpful to get name of the relation if the OID is valid.

nikkhils: @k-rus  I don't know why you don't want to fix this..

```sql
single=# select oid from pg_database where datname = 'single';
  oid  
-------
 16384
(1 row)
single=# SELECT _timescaledb_internal.create_chunk_replica_table(16384, 'data_node_1');
SSL SYSCALL error: EOF detected
```

here's the backtrace:

```
(lldb) bt
* thread #1, stop reason = signal SIGSTOP
  * frame #0: 0x00007fff6b9f333a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff6baafe60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff6b97a808 libsystem_c.dylib`abort + 120
    frame #3: 0x0000000100ae29bd postgres`ExceptionalCondition(conditionName="strvalue != NULL", errorType="FailedAssertion", fileName="snprintf.c", lineNumber=442) at assert.c:67:2
    frame #4: 0x0000000100b6a950 postgres`dopr(target=0x00007ffeef88b338, format="\" is not a chunk", args=0x00007ffeef88b5d0) at snprintf.c:442:4
    frame #5: 0x0000000100b6a715 postgres`pg_vsnprintf(str="relation \"\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7
```

k-rus: Thank you for demonstrating the crash. I didn't get it when I tested myself.
I will look into a general fix.

nikkhils: Please add a test case for this in your patch. Thanks

tsl/test/shared/sql/dist_chunk.sql
```diff
@@ -0,0 +1,38 @@
...
+-- A table for the first chunk will be created on the data node, where it is not present.
+SELECT chunk_name, data_nodes 
+FROM timescaledb_information.chunks 
+WHERE hypertable_name = 'dist_chunk_copy';
+
+-- Non-distributed chunk will be used to test an error
+SELECT chunk_name 
+FROM timescaledb_information.chunks 
+WHERE hypertable_name = 'conditions';
+
```
nikkhils, line 16: What is the expected behavior for compressed chunks? Please add a test for compressed chunk.

k-rus: Compressed chunks are ignored. They are not even available on the access node.

nikkhils: We should be able to copy/move compressed chunks. So, creation of a compressed chunk replica table on the DN needs to work, right?

k-rus: @nikkhils According to the decided plan copying compressed data will be next step after completing initial implementation of copy and move non-compressed chunk. It is more work to support compressed chunk, thus the work is done incrementally.

nikkhils (approved):
LGTM after fixing a few nits I pointed out.

tsl/test/shared/sql/dist_chunk.sql
nikkhils: Ok. Thanks for the explanation.

tsl/src/chunk.c
k-rus: Just for the record, when I do above steps I get:
```
ERROR:  relation "(null)" is not a chunk
```

tsl/test/shared/sql/dist_chunk.sql
k-rus: I added a test with a compressed chunk, which demonstrates that it doesn't affect the result.

tsl/src/chunk.c
k-rus: Fixed and tested

tsl/test/shared/expected/dist_chunk.out
k-rus: Using `remote_exec` is nontrivial in shared tests. After spending considerable time I fall back to connecting and running the query on the data node.

PR #3174, opened 2021-04-29
Drop inherit from chunk table to hypertable

k-rus:
When an empty chunk table is created, it is not associated with its
hypertable in metadata, however it was inheriting from the hypertable.
This commit removes inheritance, so the chunk table is completely
standalone and cannot affect queries on the hypertable.

In future, this fix of dropping inherit can be replaced with the chunk
table creation, which does not create inheritance in the first place.
Since it is larger amount of work, it was not considered now.


erimatnor (approved):

nikkhils (approved):
LGTM

PR #3183, opened 2021-05-01
Fix segfault in calculate_chunk_interval

svenklemm:
Block calling calculate_chunk_interval on distributed hypertables
to prevent segfault when finding min and max as distributed chunks
have no tableam since they are foreign tables.


pmwkaa (approved):
I guess this one requires a test case

mkindahl (approved):
LGTM

PR #3193, opened 2021-05-05
Fix broken links in README.md

### LINKED ISSUE #3192: broken links in readme.md
links to installation on os's other than windows are broken as the docs page has moved.

Blobcode:
fixes #3192, where links for installation for RHEL, Ubuntu and Debian as well as MacOS and Docker were broken due to the pages being moved.


NunoFilipeSantos (approved):
All the links are correct. Approved.

erimatnor:
@Blobcode Thanks for this contribution. However, the existing links work now. Maybe it was a temporary failure or it got fixed with the recent revamp of the docs pages?

In any case, I think we should keep the shortened links as it allows future redirection without updating the README.

mkindahl (approved):
Approving, but we are not going to be able to keep links consistent everywhere and this will change in the future, so question is if we should just not have links to documentation in the timescaledb repository, or have a more general link that is unlikely to change.

Blobcode:
Okay. The links seem to work now, so this is unnecessary.

PR #3199, opened 2021-05-06
Add GENERATED column support on distributed hypertables

erimatnor:
Add support for deparsing `GENERATED` columns when creating
distributed hypertables, including support for generating values
"on-the-fly" when a `RETURNING` clause is used with an `INSERT`
statement.

Previously, using generated columns on a hypertable led to errors
or had unpredictible results.


tsl/src/nodes/data_node_dispatch.c
```diff
@@ -1063,8 +1077,14 @@ get_insert_attrs(Relation rel)
 	{
 		Form_pg_attribute attr = TupleDescAttr(tupdesc, i);
 
-		if (!attr->attisdropped)
-			attrs = lappend_int(attrs, AttrOffsetGetAttrNumber(i));
+		if (attr->attisdropped
+#if PG12_GE
+			|| attr->attgenerated != '\0'
```
nikkhils, line 1082: Nit: Even  a check for `attr->attgenerated` should work instead of comparing with `'\0'`

tsl/test/sql/dist_hypertable_pg12.sql
```diff
@@ -29,7 +30,7 @@ CREATE ACCESS METHOD test_am TYPE TABLE HANDLER heap_tableam_handler;
 $$);
 
 -- Create distributed hypertable using non-default access method
-CREATE TABLE disttable(time timestamptz NOT NULL, device int) USING test_am;
+CREATE TABLE disttable(time timestamptz NOT NULL, device int, temp_c float, temp_f float GENERATED ALWAYS AS (temp_c * 9 / 5 + 32) STORED) USING test_am;
```
nikkhils, line 33: I don't see the connection between `test_am` and testing `GENERATED` columns support. Is there?

tsl/src/nodes/data_node_dispatch.c
erimatnor: I am not sure I follow what you are asking/suggesting.

`attgenerated` is a char that map to different types of generated columns. Currently, only `ATTRIBUTE_GENERATED_STORED` is supported/exists. I assume that there will be more types of generated columns in the future, so I figured it is better to check for the '\0' case in order to be forward compatible with new `ATTRIBUTE_GENERATED_` in the future.

tsl/test/sql/dist_hypertable_pg12.sql
erimatnor: No connection. Just reusing existing table instead of adding a new one.

This test file previously contained only tests for AM, as this was a new feature in PG12, thus can't run on PG11. I figured it is better to put all tests that require PG12 or higher in the same test file as the test cannot go into `dist_hypertable`.

nikkhils (approved):
LGTM

pmwkaa (approved):

PR #3207, opened 2021-05-07
Support COPY into compressed chunk

svenklemm:


src/copy.c
```diff
@@ -396,14 +396,41 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, void (*call
...
-			/* OK, store the tuple and create index entries for it */
-			table_tuple_insert(resultRelInfo->ri_RelationDesc, myslot, mycid, ti_options, bistate);
-
-			if (resultRelInfo->ri_NumIndices > 0)
-				recheckIndexes = ExecInsertIndexTuplesCompat(myslot, estate, false, NULL, NIL);
+			if (cis->compress_state)
+			{
+				TupleTableSlot *compress_slot =
+					ts_cm_functions->compress_row_exec(cis->compress_state, myslot);
+				table_tuple_insert(resultRelInfo->ri_RelationDesc,
```
gayyappan, line 413: I think we need some additional tests. 1)  a test or copy into compressed chunk +  a) before insert row triggers, b) after insert row triggers and c) stmt triggers (see line 363 above, I think we need a specific check for compressed chunk and do the trigger processing there).
2) A test where copy inserts rows into 2 or more chunks  where one of them is compressed.

gayyappan (approved):

PR #3210, opened 2021-05-10
Add trigger support on distributed hypertables

### LINKED ISSUE #3201: Enable triggers on distributed hypertables
Triggers don't work as expected on distributed hypertables because `CREATE TRIGGER` and `DROP TRIGGER` are not white-listed DDL commands to be forwarded to data nodes. The `distributed_exec` can be used as a workaround however.

To fix this issue, we should whitelist the trigger-related DDL and add tests to ensure that triggers work as expected.

erimatnor:
This change makes it possible to create and use triggers on
distributed hypertables. The DDL for `CREATE TRIGGER` and `DROP
TRIGGER` are now forwarded to data nodes. Previously, `CREATE TRIGGER`
worked, but didn't forward the DDL command.

Note that triggers exist only on the data nodes and on the data node's
root hypertable. Triggers need to execute on the data nodes where data
is stored; in particular, row-based "before" triggers might modify the
tuple. On the access node, chunks are foreign tables and such tables
do not support triggers.

Also note that statement-level triggers execute once on each node
involved, including the access node. So, if, e.g., two rows are
inserted into two different data nodes in the same statement, a
statement-level insert trigger will fire once on the access node and
once on each data node.

An issue with `RETURNING` handling for distributed insert is also
fixed (in the `DataNodeDispatch` executor node). The fix ensures that
the modified tuple is returned when a trigger modifies the tuple on a
data node.

Fixes #3201

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -268,7 +268,7 @@ relation_not_only(RangeVar *rv)
 }
 
 static void
-add_hypertable_to_process_args(ProcessUtilityArgs *args, Hypertable *ht)
+add_hypertable_to_process_args(ProcessUtilityArgs *args, const Hypertable *ht)
 {
 	args->hypertable_list = lappend_oid(args->hypertable_list, ht->main_table_relid);
 }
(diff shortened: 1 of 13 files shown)


erimatnor:
Currently looking into adding more tests for other types of triggers. But this might be improved in a separate PR.

src/process_utility.c
```diff
@@ -333,6 +333,42 @@ process_drop_foreign_server_start(DropStmt *stmt)
 	}
 }
 
+static void
+process_drop_trigger_start(ProcessUtilityArgs *args, DropStmt *stmt)
```
nikkhils, line 337: So, creation happens manually or via distributed exec. But drop is handled internally?

erimatnor:
Moved most tests to `dist_triggers` and added more tests to show trigger behavior on distributed hypertables; in particular statement-level triggers.

tsl/test/expected/dist_triggers.out
```diff
@@ -0,0 +1,511 @@
...
+AFTER  |ROW      |INSERT|_0_test_trigger_insert_after_when_dev1|    1
+AFTER  |ROW      |INSERT|z_test_trigger_all_after              |    1
+AFTER  |STATEMENT|INSERT|_0_test_trigger_insert_s_after        |    1
+BEFORE |ROW      |INSERT|_0_test_trigger_insert                |    1
+BEFORE |ROW      |INSERT|z_test_trigger_all                    |    1
+BEFORE |STATEMENT|INSERT|_0_test_trigger_insert_s_before       |    1
+(7 rows)
+
+
+NOTICE:  [db_dist_triggers_2]:
```
nikkhils, line 167: Why don't we see the firing of the triggers on DN2 and DN3? Because rows didn't get inserted in them? Might be a good idea to add inserts which insert into these datanodes as well.

nikkhils (approved):
A few comments, but LGTM overall.

PR #3217, opened 2021-05-12
Support for compressing chunks that are in unordered state

gayyappan:
After inserts go into a compressed chunk, the chunk is marked as unordered.
This PR adds a new function recompress_chunk that compresses the data and sets the status back to
compressed. Further optimizations for this function are planned but not part of this PR.
This  function can be invoked by calling `SELECT recompress_chunk(<chunk_name>)`.

recompress_chunk function is automatically invoked by the compression policy job, when it sees that a chunk is in unordered state.

### DIFF
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -125,5 +125,5 @@ UPDATE pg_class
   FROM chunks
  WHERE oid = chunk_oid;
 
-DROP FUNCTION _timescaledb_internal.chunk_dml_blocker CASCADE;
+DROP FUNCTION IF EXISTS _timescaledb_internal.chunk_dml_blocker CASCADE;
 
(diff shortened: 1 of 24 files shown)


sql/updates/latest-dev.sql
```diff
@@ -90,6 +90,13 @@ REFERENCES _timescaledb_catalog.chunk(id) ON DELETE CASCADE;
 DROP TABLE _timescaledb_catalog.chunk_tmp;
 DROP TABLE tmp_chunk_seq_value;
 
+--recompress_chunk function
+DROP FUNCTION IF EXISTS recompress_chunk( regclass, boolean);
+CREATE OR REPLACE FUNCTION recompress_chunk(
+    uncompressed_chunk REGCLASS,
+    if_not_compressed BOOLEAN = false
+) RETURNS REGCLASS AS '@MODULE_PATHNAME@', 'ts_recompress_chunk' LANGUAGE C STRICT VOLATILE;
+
```
svenklemm, line 99: this should not be needed

svenklemm (approved):

PR #3222, opened 2021-05-13
Copy rebase

svenklemm:


pmwkaa (approved):

src/copy.c
```diff
@@ -396,14 +396,41 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, void (*call
 				resultRelInfo->ri_RelationDesc->rd_att->constr)
 			{
 				Assert(resultRelInfo->ri_RangeTableIndex > 0 && estate->es_range_table);
-				ExecConstraints(resultRelInfo, myslot, estate);
+				ExecConstraints(saved_resultRelInfo, myslot, estate);
```
gayyappan, line 399: why is this on the saved_resultRelInfo (this would be the hypertable, isn't it?). Shouldn't this be on the chunk itself? We need to special case only for the compressed chunk case.

src/copy.c
```diff
@@ -396,14 +396,41 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, void (*call
...
+			}
+			else if (cis->compress_state &&
```
gayyappan, line 401: Also would be good to add a comment on why this is necessary (as part of this PR or the follow-up).

gayyappan (approved):
Question about use of saved_resultRelInfo for the regular chunk case.
Otherwise, looks good.

src/copy.c
svenklemm: reverted not sure why it was there might have been testing something

PR #3227, opened 2021-05-16
Test support for copy on distributed hypertables with compressed chunks

gayyappan:
This PR adds a test case for this case, verifies that recompress_chunk
and compression policy work as expected.

### DIFF
--- a/tsl/src/compression/compress_utils.c
+++ b/tsl/src/compression/compress_utils.c
@@ -541,7 +541,7 @@ get_compression_fcinfo(char *fname, FmgrInfo *decompfn, FunctionCallInfo *decomp
 }
 
 static Datum
-tsl_recompress_remote_chunk(Chunk *uncompressed_chunk, FunctionCallInfo fcinfo)
+tsl_recompress_remote_chunk(Chunk *uncompressed_chunk, FunctionCallInfo fcinfo, bool if_compressed)
 {
 	FmgrInfo decompfn;
 	FmgrInfo compfn;
@@ -552,20 +552,25 @@ tsl_recompress_remote_chunk(Chunk *uncompressed_chunk, FunctionCallInfo fcinfo)
 	FunctionCallInvoke(decompfn_fcinfo);
 	if (decompfn_fcinfo->isnull)
 	{
-		ereport(WARNING,
+		ereport((if_compressed ? NOTICE : ERROR),
 				(errcode(ERRCODE_INTERNAL_ERROR),
-				 errmsg("decompression failed for recompress_chunk for %u",
-						uncompressed_chunk->fd.id)));
+				 errmsg("decompression failed for chunk \"%s\"",
+						get_rel_name(uncompressed_chunk->table_id)),
+				 errdetail("The compression status for the chunk is %d",
+						   uncompressed_chunk->fd.status)));
+
 		PG_RETURN_NULL();
 	}
 	get_compression_fcinfo(COMPRESS_CHUNK_FUNCNAME, &compfn, &compfn_fcinfo, fcinfo);
 	Datum compoid = FunctionCallInvoke(compfn_fcinfo);
 	if (compfn_fcinfo->isnull)
 	{
-		ereport(WARNING,
+		ereport((if_compressed ? NOTICE : ERROR),
 				(errcode(ERRCODE_INTERNAL_ERROR),
-				 errmsg("compression failed for recompress_chunk for %u",
-						uncompressed_chunk->fd.id)));
+				 errmsg("compression failed for chunk \"%s\"",
+						get_rel_name(uncompressed_chunk->table_id)),
+				 errdetail("The compression status for the chunk is %d",
+						   uncompressed_chunk->fd.status)));
 		PG_RETURN_NULL();
 	}
 	return compoid;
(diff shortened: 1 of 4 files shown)


tsl/src/compression/compress_utils.c
```diff
@@ -552,20 +552,22 @@ tsl_recompress_remote_chunk(Chunk *uncompressed_chunk, FunctionCallInfo fcinfo)
...
 	if (decompfn_fcinfo->isnull)
 	{
-		ereport(WARNING,
+		ereport((if_compressed ? NOTICE : ERROR),
 				(errcode(ERRCODE_INTERNAL_ERROR),
-				 errmsg("decompression failed for recompress_chunk for %u",
-						uncompressed_chunk->fd.id)));
+				 errmsg("decompression failed for recompress_chunk for %u status=%d",
+						uncompressed_chunk->fd.id,
+						uncompressed_chunk->fd.status)));
```
erimatnor, line 559: I would suggest putting the status in a detailed message. I don't think the user will understand what this integer means.

Other than that, I am wondering if we shouldn't also reword this error. The current formulation sounds as if there was a failure in decompressing the function "recompress_chunk", although a chunk ID is given. I would expect the error to be about the chunk that failed, not the function. It is already obvious what function is being called. Ideally, we would also print the chunk name and not the internal ID.

Suggestion:
```suggestion
				 errmsg("decompression failed for chunk \"%s\"", get_rel_name(uncompressed_chunk->table_id),
				 errdetail("The compression status for the chunk is %d", uncompressed_chunk->fd.status)));
				 
```

tsl/src/compression/compress_utils.c
```diff
@@ -552,20 +552,22 @@ tsl_recompress_remote_chunk(Chunk *uncompressed_chunk, FunctionCallInfo fcinfo)
...
 	get_compression_fcinfo(COMPRESS_CHUNK_FUNCNAME, &compfn, &compfn_fcinfo, fcinfo);
 	Datum compoid = FunctionCallInvoke(compfn_fcinfo);
 	if (compfn_fcinfo->isnull)
 	{
-		ereport(WARNING,
+		ereport((if_compressed ? NOTICE : ERROR),
 				(errcode(ERRCODE_INTERNAL_ERROR),
-				 errmsg("compression failed for recompress_chunk for %u",
-						uncompressed_chunk->fd.id)));
+				 errmsg("compression failed for recompress_chunk for %u with status=%d",
```
erimatnor, line 568: Same here as above (for decompress case)

tsl/src/remote/dist_copy.c
```diff
@@ -813,9 +813,16 @@ remote_copy_process_and_send_data(RemoteCopyContext *context)
 	else
 		point = get_current_point_for_text_copy(context->data_context);
 
-	chunk = get_target_chunk(ht, point, &context->connection_state);
+	chunk = (Chunk *) get_target_chunk(ht, point, &context->connection_state);
```
erimatnor, line 816: If we really need a mutable chunk here I think we should change `get_target_chunk` to return a mutable chunk  or move the mutating part into `get_target_chunk`.

tsl/src/remote/dist_copy.c
```diff
@@ -813,9 +813,16 @@ remote_copy_process_and_send_data(RemoteCopyContext *context)
...
 	connections = get_connections_for_chunk(context, chunk->fd.id, chunk->data_nodes, GetUserId());
 
+	/* for remote copy, we don't use chunk insert states on the AN.
+	 * so we need to explicitly set the chunk as unordered when copies
+	 * are directed to previously compressed chunks
+	 */
+	if (ts_chunk_is_compressed(chunk) && (!ts_chunk_is_unordered(chunk)))
+		ts_chunk_set_unordered(chunk);
```
erimatnor, line 824: I think this should probably be part of `get_target_chunk`, which also fixes the const issue above.

tsl/test/expected/dist_compression.out
```diff
@@ -880,3 +880,98 @@ SELECT * from test_recomp_int_chunk_status ORDER BY 1;
...
+FROM show_chunks('test_recomp_int') AS chunk
+ORDER BY chunk;
+                compress_chunk                
+----------------------------------------------
+ _timescaledb_internal._dist_hyper_4_14_chunk
+ _timescaledb_internal._dist_hyper_4_15_chunk
+ _timescaledb_internal._dist_hyper_4_16_chunk
+(3 rows)
+
+--interleave copy into 3 different chunks and check status--
```
erimatnor, line 916: Shouldn't status be checked immediately after compression above rather than after COPY?

tsl/src/remote/dist_copy.c
gayyappan: Won't that be setting it prematurely before the connections are created and available? I changed the status o ts_get_chunk to avoid the casting away of const.

tsl/test/expected/dist_compression.out
gayyappan: added a check after call to compress_chunk as well

tsl/src/compression/compress_utils.c
gayyappan: fixed

tsl/src/compression/compress_utils.c
gayyappan: fixed.

svenklemm (approved):

PR #3250, opened 2021-05-20
Fix constraint trigger on hypertables

### LINKED ISSUE #3235: Chunk creation fails after creating a CONSTRAINT TRIGGER
After creating a constraint trigger chunk creation fails for the hypertable.

```
CREATE CONSTRAINT TRIGGER t4_constraint AFTER INSERT ON trigger_test FOR EACH ROW EXECUTE FUNCTION stmt_trig_info();

INSERT INTO trigger_test SELECT '2001-01-01',1,0;
ERROR:  syntax error at or near "TRIGGER" at character 91
```

svenklemm:
When creating a constraint trigger on a hypertable the command
succeeds and the constraint trigger works correctly for existing
chunks but any chunk creation after that would fail with an
error while copying chunk constraints.

Fixes #3235


mkindahl (approved):
LGTM but I would appreciate  if you could add two things:

1. A description about what the function does, e.g., "copies constraints from the hypertable to the chunk. Unique constraints and primary keys need to be..." (These are handled specially here since they can be in a tablespace. Why not other constraints?)
2. A paragraph that briefly describe what didn't work before (why was the error generated, not only that there were an error) and the change you make at a high level (e.g., "for primary keys and unique constraints, the index tablespace is missing, causing and error and this commit fix this by ....".

gayyappan (approved):

gayyappan (approved):

PR #3252, opened 2021-05-20
Fix blocking triggers with transition tables

### LINKED ISSUE #3234: Block triggers with transition tables during trigger creation
Currently it is possible to create a trigger with transition tables, which will fire for already existing chunks. But whenever
new chunks are created chunk creation will fail with `ERROR:  hypertables do not support transition tables in triggers`.
We should change this so the error happens when trying to create the trigger and not when the next chunk is created.

```
CREATE TRIGGER t4_transition AFTER INSERT ON trigger_test REFERENCING NEW TABLE AS tg_new FOR EACH STATEMENT EXECUTE FUNCTION stmt_trig_info();
INSERT INTO trigger_test SELECT '2001-01-01',1,0;
ERROR:  hypertables do not support transition tables in triggers
```

svenklemm:
Block creation of triggers with transition tables on trigger creation
instead of erroring out when a chunk is created.

Fixes #3234


test/sql/triggers.sql
```diff
@@ -293,3 +293,26 @@ DROP TRIGGER create_vehicle_trigger ON location;
 SELECT count(1) FROM pg_depend d WHERE d.classid = 'pg_trigger'::regclass AND NOT EXISTS (SELECT 1 FROM pg_trigger WHERE oid = d.objid);
 DROP TABLE location;
 
+-- test triggers with transition tables
+-- test creaing hypertable from table with triggers with transition tables
```
mfundul, line 297: `creaing` typo

mfundul (approved):

test/sql/triggers.sql
svenklemm: fixed

PR #3264, opened 2021-05-22
Fix broken links in notice

JuanitoFatas:
Hi 👋🏼 🐯

I was trying Timescale and found these links were moved to somewhere else after installed the extension:

CleanShot 2021-05-22 at 17 01 06@2x

We could either:

1) Add page redirects to existing links and keep these links unchanged, then close this PR
2) Change these links to something that should never change. Add redirects of these links so the `notice.sql` won’t need to be changed in the future:

- https://docs.timescale.com/telemetry
- https://docs.timescale.com/getting-started
- https://docs.timescale.com/api
- https://docs.timescale.com/core-concepts

3) Merge this PR and call it a day 😅

I am happy with any options.

Thanks!

### DIFF
--- a/sql/notice.sql
+++ b/sql/notice.sql
@@ -10,7 +10,7 @@ BEGIN
   THEN
     telemetry_string = E'Note: Please enable telemetry to help us improve our product by running: ALTER DATABASE "' || current_database() || E'" SET timescaledb.telemetry_level = ''basic'';';
   ELSE
-    telemetry_string = E'Note: TimescaleDB collects anonymous reports to better understand and assist our users.\nFor more information and how to disable, please see our docs https://docs.timescaledb.com/using-timescaledb/telemetry.';
+    telemetry_string = E'Note: TimescaleDB collects anonymous reports to better understand and assist our users.\nFor more information and how to disable, please see our docs https://docs.timescale.com/timescaledb/latest/how-to-guides/configuration/telemetry.';
   END IF;
 
   RAISE WARNING E'%\n%\n',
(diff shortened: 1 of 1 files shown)


sql/notice.sql
```diff
@@ -10,7 +10,7 @@ BEGIN
   THEN
     telemetry_string = E'Note: Please enable telemetry to help us improve our product by running: ALTER DATABASE "' || current_database() || E'" SET timescaledb.telemetry_level = ''basic'';';
   ELSE
-    telemetry_string = E'Note: TimescaleDB collects anonymous reports to better understand and assist our users.\nFor more information and how to disable, please see our docs https://docs.timescaledb.com/using-timescaledb/telemetry.';
+    telemetry_string = E'Note: TimescaleDB collects anonymous reports to better understand and assist our users.\nFor more information and how to disable, please see our docs https://docs.timescale.com/timescaledb/latest/how-to-guides/configuration/telemetry/#disabling-telemetry.';
```
mfreed, line 13: The link includes "more information", so should probably still point to the same top-level section, and just be updated for the new link.  Thanks!

```suggestion
    telemetry_string = E'Note: TimescaleDB collects anonymous reports to better understand and assist our users.\nFor more information and how to disable, please see our docs https://docs.timescale.com/timescaledb/latest/how-to-guides/configuration/telemetry/.';
```

mfreed:
@Loquacity Given we are updating the links in TimescaleDB NOTICE, I wanted to confirm that the proposed links are likely the correct place to direct users to?

In particular, e.g., should we point to https://docs.timescale.com/api/latest/ vs. https://docs.timescale.com/api/?

(Aside, the redirect from https://docs.timescale.com/api/ appears to be broken.  Will create another issue in docs for that.)

Loquacity:
> @Loquacity Given we are updating the links in TimescaleDB NOTICE, I wanted to confirm that the proposed links are likely the correct place to direct users to?

Yes, these look right to me.

> 
> In particular, e.g., should we point to https://docs.timescale.com/api/latest/ vs. https://docs.timescale.com/api/?

Yes, always use the `/latest/` notation. When we get versions working properly, it will become important. 

> 
> (Aside, the redirect from https://docs.timescale.com/api/ appears to be broken. Will create another issue in docs for that.)

👍

erimatnor:
@Loquacity @mfreed Can you approve if this looks correct to you?

ryanbooz (approved):
Just FYI, these redirects were also fixed yesterday in a separate update to the web code. Any older links that do not include "latest" (but were otherwise valid as of the April 30 cut-over) should redirect to the appropriate place now.

Loquacity (approved):

svenklemm:
Thank you for your contribution.

PR #3269, opened 2021-05-24
Enable 1PC and DN reads for read only AN txns

nikkhils:
In high availability setups, a streaming replica of the access node can 
be configured. Since it's a hot standby users should be able to run 
read-only queries from this replica and should be able to query the 
distributed hypertables from it. 

To allow this, we need to disable the "timescaledb.enable_2pc" GUC on
the standby.

Additionally, the internal code which sets up a connection from the 
standby access node to the data nodes now marks the ongoing transaction
explicitly as READ ONLY ones. This ensures that any functions or any 
activity which tries to make modifications on the data nodes errors
out.

In normal Postgres, the ongoing transaction becomes READ WRITE
immediately on promotion whereas in our case the ongoing transaction
will remain READ ONLY till completion. New transactions on the same
session will be READ WRITE as expected.

### DIFF
--- a/tsl/src/dist_backup.c
+++ b/tsl/src/dist_backup.c
@@ -125,11 +125,11 @@ create_distributed_restore_point(PG_FUNCTION_ARGS)
 		 * inconsistent state when the distributed database is restored from a backup
 		 * using the restore point.
 		 *
-		 * To do that we take an exclusive lock on the remote transaction
+		 * To do that we take an access exclusive lock on the remote transaction
 		 * table, which will force any concurrent transaction
 		 * wait during their PREPARE phase.
 		 */
-		LockRelationOid(ts_catalog_get()->tables[REMOTE_TXN].id, ExclusiveLock);
+		LockRelationOid(ts_catalog_get()->tables[REMOTE_TXN].id, AccessExclusiveLock);
 
 		/* Prevent situation when new data node added during the execution */
 		LockRelationOid(ForeignServerRelationId, ExclusiveLock);
--- a/tsl/src/remote/dist_txn.c
+++ b/tsl/src/remote/dist_txn.c
@@ -5,15 +5,17 @@
  */
 #include <postgres.h>
 #include <access/htup_details.h>
-#include <access/xact.h>
+#include <access/xlog.h>
+#include <storage/lmgr.h>
 #include <utils/hsearch.h>
 #include <utils/builtins.h>
 #include <utils/memutils.h>
 #include <utils/syscache.h>
-
 #include "dist_txn.h"
+#include "catalog.h"
 #include "connection.h"
 #include "async.h"
+#include "errors.h"
 #include "txn.h"
 #include "txn_store.h"
 #include "guc.h"
@@ -126,9 +128,20 @@ static void
 dist_txn_xact_callback_1pc_pre_commit()
 {
 	RemoteTxn *remote_txn;
+	Catalog *catalog = ts_catalog_get();
 	AsyncRequestSet *ars = async_request_set_create();
 
 	eventcallback(DTXN_EVENT_PRE_COMMIT);
+
+	/*
+	 * In 1PC, we don't need to add entries to the remote_txn table. However
+	 * we do need to take a SHARE lock on it to interlock with any distributed
+	 * restore point activity that might be happening in parallel.
+	 *
+	 * The catalog table lock is kept until the transaction completes in order to
+	 * synchronize with distributed restore point creation
+	 */
+	LockRelationOid(catalog->tables[REMOTE_TXN].id, AccessShareLock);
 
 	/* send a commit to all connections */
 	remote_txn_store_foreach(store, remote_txn)
@@ -474,11 +487,39 @@ dist_txn_xact_callback_2pc(XactEvent event, void *arg)
 static void
 dist_txn_xact_callback(XactEvent event, void *arg)
 {
+	bool use_2pc;
+	char *xactReadOnly;
+
 	/* Quick exit if no connections were touched in this transaction. */
 	if (store == NULL)
 		return;
 
-	if (ts_guc_enable_2pc)
+	/*
+	 * Windows MSVC builds have linking issues for GUC variables from postgres for
+	 * use inside this extension. So we use GetConfigOptionByName
+	 */
+	xactReadOnly = GetConfigOptionByName("transaction_read_only", NULL, false);
+
+	/*
+	 * The decision to use 2PC rests on multiple factors:
+	 *
+	 * 1) if ts_guc_enable_2pc is enabled and it's a regular backend use it
+	 *
+	 * 2) if ts_guc_enable_2pc is enabled but we are running a read only txn, don't use it
+	 *
+	 * We might be tempted to use 1PC if just one DN is involved in the transaction.
+	 * However, it's possible that a transaction which involves data on AN and the one DN could get
+	 * a failure at the end of the COMMIT processing on the AN due to issues in local AN data. In
+	 * such a case since we send a COMMIT at "XACT_EVENT_PRE_COMMIT" event time to the DN, we might
+	 * end up with a COMMITTED DN but an aborted AN! Hence this optimization is not possible to
+	 * guarantee transactional semantics.
+	 */
+	use_2pc = (ts_guc_enable_2pc && strncmp(xactReadOnly, "on", sizeof("on")) != 0);
+#ifdef TS_DEBUG
+	ereport(DEBUG3, (errmsg("use 2PC: %s", use_2pc ? "true" : "false")));
+#endif
+
+	if (use_2pc)
 		dist_txn_xact_callback_2pc(event, arg);
 	else
 		dist_txn_xact_callback_1pc(event, arg);
--- a/tsl/src/remote/txn.c
+++ b/tsl/src/remote/txn.c
@@ -83,20 +83,51 @@ remote_txn_begin(RemoteTxn *entry, int curlevel)
 	/* Start main transaction if we haven't yet */
 	if (xact_depth == 0)
 	{
-		const char *sql;
+		StringInfoData sql;
+		char *xactReadOnly;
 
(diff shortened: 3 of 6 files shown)


tsl/src/remote/connection.c
```diff
@@ -697,6 +697,16 @@ remote_connection_configure(TSConnection *conn)
...
 	}
 
+	/*
+	 * If we are initiating connection from a standby (of an AN for example),
+	 * then the remote connection session needs to be also set up as a
+	 * READ ONLY one. This will catch any commands that are sent from the
+	 * read only AN to datanodes but which could have potential read-write
+	 * side effects on data nodes.
+	 */
+	if (RecoveryInProgress())
```
erimatnor, line 707: Shouldn't we also check for other read-only situations, not only recovery-in-progress. For instance, the session on the AN could also have done `SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY`, or  `default_transaction_read_only=true` is set.

erimatnor: Is it enough to check `XactReadOnly`?

nikkhils: > Is it enough to check `XactReadOnly`?

On AN, `XactReadOnly` can be set for a specific transaction whereas we want to set the entire remote SESSION as read only for the rest of its life. 

Also, we could intercept future `SET SESSION` or `set transaction` kinda calls on the AN backend and do the same on the remote sessions, but then we will also need to trap RESET commands. There is a `remote_connection_configure_if_changed` function which could be modified to handle this.

akuzm (approved):
Makes sense.

This looks somewhat similar to the "don't use 2pc for read-only transactions" idea we discussed earlier (https://github.com/timescale/timescaledb-private/issues/482#issuecomment-949743700)

nikkhils:
> Makes sense.
> 
> This looks somewhat similar to the "don't use 2pc for read-only transactions" idea we discussed earlier ([timescale/timescaledb-private#482 (comment)](https://github.com/timescale/timescaledb-private/issues/482#issuecomment-949743700))

Yeah, that basic use case when the user specifies read only will be similarly optimized.

tsl/src/remote/dist_txn.c
```diff
@@ -543,6 +545,19 @@ dist_txn_subxact_callback(SubXactEvent event, SubTransactionId mySubid,
...
+	 * Disable 2PC if you are running as a standby
+	 *
+	 * It's good for performance since all queries will be read only on the standby
+	 * Additionally we try to assign xids when 2PC is enabled which doesn't work
+	 * on a standby
+	 */
+	if (RecoveryInProgress() && ts_guc_enable_2pc)
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_OPERATION_NOT_SUPPORTED),
+				 errmsg("disable two-phase commit transactions"),
```
erimatnor, line 558: I am not sure I think it is great to issue this error only when the first distributed transaction is run. Is there a way we can do it much earlier?

Otherwise I think there's a risk that someone will believe they have successfully set up a standby only to find that it later fails.

Another, probably better, option is to just simply not do 2pc when we know it is not needed, even if `ts_guc_enabled_2pc` is TRUE. I guess the GUC doesn't imply that one has to _always_ do 2pc just because it is enabled.

tsl/src/remote/txn.c
```diff
@@ -83,20 +83,40 @@ remote_txn_begin(RemoteTxn *entry, int curlevel)
...
+		StringInfoData sql;
 
 		Assert(remote_connection_get_status(entry->conn) == CONN_IDLE);
 		elog(DEBUG3, "starting remote transaction on connection %p", entry->conn);
 
+		initStringInfo(&sql);
+		appendStringInfo(&sql, "%s", "START TRANSACTION ISOLATION LEVEL");
 		if (IsolationIsSerializable())
-			sql = "START TRANSACTION ISOLATION LEVEL SERIALIZABLE";
+			appendStringInfo(&sql, "%s", " SERIALIZABLE");
```
erimatnor, line 94: We should add test for serializable.

tsl/src/remote/dist_txn.c
```diff
@@ -543,6 +545,19 @@ dist_txn_subxact_callback(SubXactEvent event, SubTransactionId mySubid,
...
 {
+	/*
+	 * Disable 2PC if you are running as a standby
+	 *
+	 * It's good for performance since all queries will be read only on the standby
+	 * Additionally we try to assign xids when 2PC is enabled which doesn't work
+	 * on a standby
+	 */
+	if (RecoveryInProgress() && ts_guc_enable_2pc)
+		ereport(ERROR,
```
erimatnor, line 556: Should test that we hit this error.

tsl/src/remote/dist_txn.c
nikkhils: @erimatnor I have refactored this now. Please re-check the new code.

tsl/src/remote/dist_txn.c
```diff
@@ -474,11 +476,27 @@ dist_txn_xact_callback_2pc(XactEvent event, void *arg)
...
+	/*
+	 * The decision to use 2PC rests on multiple factors:
+	 *
+	 * 1) if ts_guc_enable_2pc is enabled and it's a regular backend use it
+	 *
+	 * 2) if (1) is true but only ONE DN is involved in the remote transaction, don't use it
+	 *
+	 * 3) if ts_guc_enable_2pc is enabled but we are running a read only txn, don't use it
+	 */
+	use_2pc = (ts_guc_enable_2pc && !XactReadOnly && (hash_get_num_entries(store->hashtable) != 1));
```
mfundul, line 494: I don't think 1 DN doing 1PC is safe. For example:

- AN pre-commit: remotely commits DN
- AN rolls-back transaction due to conflicts
- No way for DN to rollback because no 2PC

tsl/src/remote/dist_txn.c
```diff
@@ -474,11 +490,40 @@ dist_txn_xact_callback_2pc(XactEvent event, void *arg)
...
+	 *
+	 * 1) if ts_guc_enable_2pc is enabled and it's a regular backend use it
+	 *
+	 * 2) if ts_guc_enable_2pc is enabled but we are running a read only txn, don't use it
+	 *
+	 * 3) if (1) is true but only ONE DN is involved in the remote transaction, don't use it
+	 * For now, we hold off on (3) because it's possible that a transaction which involves
+	 * data on AN and DNs could get a failure at the fag end of the COMMIT processing on the
+	 * AN due to issues in local AN data. In such a case since we send a COMMIT at
+	 * "XACT_EVENT_PRE_COMMIT" event time to the DN, we might end up with a COMMITTED DN but
```
erimatnor, line 514: Do you know why the commit is sent in the pre-commit hook as opposed to the commit hook? I guess it is because we need to be able to roll back the AN transaction if the commit on the DN fails, right? So, is (3) even a realistic option? Maybe we should just skip all mention of it here or, even better, clearly state that it is not a possibility if we want  to guarantee transactional semantics.

erimatnor (approved):
Approving, but with nits.

It would also be good to add a test for executing under SERIALIZABLE mode, since there is a code path for it that is not tested.

tsl/src/dist_backup.c
```diff
@@ -129,7 +129,7 @@ create_distributed_restore_point(PG_FUNCTION_ARGS)
 		 * table, which will force any concurrent transaction
 		 * wait during their PREPARE phase.
 		 */
-		LockRelationOid(ts_catalog_get()->tables[REMOTE_TXN].id, ExclusiveLock);
+		LockRelationOid(ts_catalog_get()->tables[REMOTE_TXN].id, AccessExclusiveLock);
```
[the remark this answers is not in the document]
nikkhils: Done.

tsl/src/remote/dist_txn.c
```diff
@@ -126,10 +127,19 @@ static void
...
 	AsyncRequestSet *ars = async_request_set_create();
 
 	eventcallback(DTXN_EVENT_PRE_COMMIT);
 
+	/*
+	 * In 1PC, we don't need to add entries to the remote_txn table. However
+	 * we do need to take a SHARE lock on it to interlock with any distributed
+	 * restore point activity that might be happening in parallel.
+	 */
+	rel = table_open(catalog->tables[REMOTE_TXN].id, AccessShareLock);
```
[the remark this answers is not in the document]
nikkhils: Good idea. DOne.

tsl/src/remote/dist_txn.c
```diff
@@ -474,11 +490,40 @@ dist_txn_xact_callback_2pc(XactEvent event, void *arg)
...
 {
+	bool use_2pc;
+	char *xactReadOnly;
+
 	/* Quick exit if no connections were touched in this transaction. */
 	if (store == NULL)
 		return;
 
-	if (ts_guc_enable_2pc)
+	/* Windows related shenanigans.. */
```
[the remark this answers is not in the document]
nikkhils: Done

tsl/src/remote/dist_txn.c
nikkhils: Ok, removed this as a possibility.

tsl/src/remote/txn.c
```diff
@@ -83,20 +83,48 @@ remote_txn_begin(RemoteTxn *entry, int curlevel)
...
+		 * READ ONLY one. This will catch any commands that are sent from the
+		 * read only AN to datanodes but which could have potential read-write
+		 * side effects on data nodes.
+		 *
+		 * Note that when the STANDBY gets promoted then the ongoing transaction
+		 * will remain READ ONLY till its completion. New transactions will be
+		 * suitably READ WRITE. This is a slight change in behavior as compared to
+		 * regular Postgres, but promotion is not a routine activity, so it should
+		 * be acceptable and typically users would be reconnecting to the new
+		 * promoted AN anyways
```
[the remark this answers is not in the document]
nikkhils: DOne

PR #3279, opened 2021-05-25
Add some more randomness to chunk assignment

cevian:
Previously the assignment of data nodes to chunks had a bit
of a thundering-herd problem for multiple hypertables
without space partions: the data node assigned for the
first chunk was always the same across hypertables.
We fix this by adding the hypertable_id to the
index into the datanode array. This de-synchronizes
across hypertables but maintains consistency for any
given hypertable.

We could make this consistent for space partitioned tables
as well but avoid doing so now to prevent partitions
jumping nodes due to this change.

This also effects tablespace selection in the same way.

### DIFF
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -1191,10 +1191,11 @@ ts_hypertable_has_tablespace(Hypertable *ht, Oid tspc_oid)
 }
 
 static int
-hypertable_get_chunk_slice_ordinal(const Hypertable *ht, const Hypercube *hc)
+hypertable_get_chunk_round_robin_index(const Hypertable *ht, const Hypercube *hc)
 {
 	Dimension *dim;
 	DimensionSlice *slice;
+	int offset = 0;
 
 	Assert(NULL != ht);
 	Assert(NULL != hc);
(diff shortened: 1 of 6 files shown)


src/hypertable.c
```diff
@@ -1193,26 +1193,33 @@ ts_hypertable_has_tablespace(Hypertable *ht, Oid tspc_oid)
...
 	Assert(NULL != hc);
 
 	dim = hyperspace_get_closed_dimension(ht->space, 0);
 
 	if (NULL == dim)
+	{
 		dim = hyperspace_get_open_dimension(ht->space, 0);
+		/* add some randomness between hypertables so that
+		 * if there is no space partitions, but multiple hypertables
+		 * the initial index is different for different hypertables. */
```
erimatnor, line 1212: I suggest expanding:

```suggestion
		/* Add some randomness between hypertables so that
		 * if there is no space partitions, but multiple hypertables
		 * the initial index is different for different hypertables. 
		 * This protects against creating a lot of chunks on the same
		 * data node when many hypertables are created at roughly
		 * the same time, e.g., from a bootstrap script.
		 */
```

erimatnor (approved):
Just one suggestion for clarifying comment.

pmwkaa (approved):

mkindahl (approved):
LGTM

PR #3301, opened 2021-06-04
Use ExecComputeStoredGeneratedCompat macro

svenklemm:
When adding support for PG13 we introduced a macro to  hide the
function signature differences of ExecComputeStoredGenerated
but the callers of this function never got adjusted to use the
macro.


gayyappan (approved):

mkindahl (approved):
LGTM. Checked if there were other missing locations, but `chunk_dispatch_state.c` was already correct.

PR #3302, opened 2021-06-04
Move internal API functions to experimental schema

erimatnor:
Move the "block new chunks" functions and the chunk-based continuous
aggregate refresh function to the new experimental schema.


mkindahl (approved):
LGTM

mfundul (approved):

PR #3305, opened 2021-06-07
Fix pull_varnos miscomputation of relids set

svenklemm:
Upstream fixed a bug with miscomputation of relids for
PlaceHolderVar. Those fixes changed the signature of pull_varnos,
make_simple_restrictinfo and make_restrictinfo.
The fixes got backported to the PG12 and PG13 branch but to not
break compatibility with extensions the old functions were left in.
This patch makes our code use the new functions when compiling
against a postgres version that has them.

https://github.com/postgres/postgres/commit/1cce024fd2
https://github.com/postgres/postgres/commit/73fc2e5bab

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ accidentally triggering the load of a previous DB version.**
 
 ## Unreleased
 
+**Bugfixes**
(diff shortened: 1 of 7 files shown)


gayyappan (approved):

mkindahl (approved):
LGTM

PR #3306, opened 2021-06-07
Run ABI tests with TAP support

erimatnor:
Install the Perl prerequisites when building the image for ABI tests,
including the `prove` binary.

Although the ABI tests currently don't run TAP tests, CMake still
failed the configuration since it expects the prerequisites to be
installed unless it is run with `-DTAP_CHECKS=off`.


mkindahl (approved):
LGTM

pmwkaa (approved):

nikkhils (approved):
LGTM

svenklemm (approved):

PR #3311, opened 2021-06-08
Make provecheck run specific tap tests

nikkhils:
Introduce a shell wrapper around perl prove utility to control running
of TAP tests. The following control variable is supported:

PROVE_TESTS: only run TAP tests from this list

e.g make provecheck PROVE_TESTS="t/foo.pl t/bar.pl"

Note that you can also use regular expressions to run multiple taps
tests matching the pattern:

e.g make provecheck PROVE_TESTS="t/\*chunk\*"

If the existing "TESTS=" option is used along with PROVE_TESTS then
the subset represented by PROVE_TESTS will also get run. Otherwise tap
tests will be skipped if "TESTS=" is specified.

e.g make installcheck TESTS=dist_hyper* PROVE_TESTS="t/001_*"

### DIFF
--- a/test/pg_prove.sh
+++ b/test/pg_prove.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Wrapper around perl prove utility to control running of TAP tests
+#
+# The following control variable is supported:
+#
+# PROVE_TESTS  only run TAP tests from this list
+# e.g make provecheck PROVE_TESTS="t/foo.pl t/bar.pl"
+#
+# Note that you can also use regular expressions to run multiple
+# taps tests matching the pattern:
+#
+# e.g make provecheck PROVE_TESTS="t/*chunk*"
+#
+
+PROVE_TESTS=${PROVE_TESTS:-}
+PROVE=${PROVE:-prove}
+
+# If PROVE_TESTS is specified then run those subset of TAP tests even if
+# TESTS is also specified
+if [ -z "$PROVE_TESTS" ]
+then
+    # Exit early if we are running with TESTS=expr
+    if [ ! -z "$TESTS" ]
+    then
+        exit 0
+    fi
+    FINAL_TESTS="t/*.pl"
+else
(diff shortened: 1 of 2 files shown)


test/pg_prove.sh
```diff
@@ -0,0 +1,35 @@
...
+# Note that you can also use regular expressions to run multiple
+# taps tests matching the pattern:
+#
+# e.g make provecheck PROVE_TESTS="t/*chunk*"
+#
+
+PROVE_TESTS=${PROVE_TESTS:-}
+PROVE=${PROVE:-prove}
+
+echo "PROVE_TESTS ${PROVE_TESTS}"
```
erimatnor, line 19: Looks like debugging. Should we always show this? Most of the times it is empty I guess?

erimatnor (approved):
LGTM, just one question/nit.

test/pg_prove.sh
```diff
@@ -0,0 +1,35 @@
...
+    if [ ! -z "$TESTS" ]
+    then
+        exit 0
+    fi
+    FINAL_TESTS="t/*.pl"
+else
+    FINAL_TESTS=$PROVE_TESTS
+fi
+
+${PROVE} -I "${SRC_DIR}/src/test/perl" -I "${CM_SRC_DIR}/test/perl" $FINAL_TESTS
```
pmwkaa, line 35: Do we need put `"$FINAL_TESTS"` in quotes here or is it intentional?

pmwkaa (approved):

mkindahl (approved):
LGTM

test/pg_prove.sh
nikkhils: Yeah, more like debugging. Can remove it.

test/pg_prove.sh
nikkhils: it's intentional. The `prove` binary doesn't need the input in quotes.

PR #3322, opened 2021-06-10
Remove unused code

afiskon:
```
Remove unused code

While looking for code that relies on fixed bucket_width I discovered
several files that don't seem to be used anymore. This patch removes
these files. Even if there is a value in them we can't rely on fixed
bucket_width in the future anyway.
```


gayyappan (approved):

erimatnor (commented):
LGTM. Just a couple of reflections.

* I would avoid personal reflections/stories in the commit message, e.g., "I found...", "When doing this I thought about...". Instead stick to describing changes and why they're made.
* I am curious as to why this code is not used and whether it was ever used. What changed, if anything? How is the job managed now without this code?

afiskon:
> LGTM. Just a couple of reflections.
> 
>     * I would avoid personal reflections/stories in the commit message, e.g., "I found...", "When doing this I thought about...". Instead stick to describing changes and why they're made.
> 
>     * I am curious as to why this code is not used and whether it was ever used. What changed, if anything? How is the job managed now without this code?

@gayyappan told me that it was used in the past, then the functionality moved elsewhere (didn't remember all the details, sorry) but some dead code has left 🤷‍♂️ 

I'm fine with avoiding the "I found..." style of commit messages. If you don't mind, I will not edit it this time to not trigger all the checks once again.

svenklemm (approved):

PR #3327, opened 2021-06-12
Make aggregate in caggs fully qualified

### LINKED ISSUE #1919: function "last(anyelement,"any")" does not exist when accessing continuous aggregate via fdw
 - OS: ubuntu:18.04 - docker
 - PostgreSQL version (output of `postgres --version`): 10.12
 - Installation method: using docker from source
**Describe the bug**
When I want to query a continuous aggregate on a different database via the postgres_fdw I get the error `ERROR: function "last(anyelement,"any")" does not exist`.
I have created a foreign table that points to the continuous aggregate in the remote database.

**To Reproduce**
1. Create a continuous aggregate with a column that uses any timescaledb specific aggregate like last() or first().
2. Create connect from a different database via postgres_fdw to the database which has the continuous aggregate.
3. Create a foreign table that points to the continuous aggregate.
4. Do a query that includes a collum which has a timescaledb specific aggregate.
5. The query fails with `ERROR: function "last(anyelement,"any")" does not exist`

**Expected behavior**
The query should return a result as if it is queried on the database where the continuous aggregate is defined.

**Actual behavior**
A clear and concise description of what _actually_ happened.

**Additional context**

### LINKED ISSUE #3326: Make all references in cagg view definitions fully qualified
When using timescaledb aggregate functions like first()/last() querying the cagg view through postgres_fdw will not work because postgres_fdw resets the search_path to pg_catalog. We should make all object references in the cagg view fully qualified so querying through postgres_fdw works without modifications.

svenklemm:
When querying continuous aggregate views with a search_path not
including public the query will fail cause the function reference
in the finalize call is not fully qualified.

This can surface when querying caggs through postgres_fdw which
resets search_path to contain only pg_catalog.

Fixes #1919
Fixes #3326


sql/updates/post-update.sql
```diff
@@ -2,12 +2,13 @@
...
 DECLARE
  vname regclass;
+ materialized_only bool;
  altercmd text;
  ts_version TEXT;
 BEGIN
     SELECT extversion INTO ts_version FROM pg_extension WHERE extname = 'timescaledb';
 
-    FOR vname IN select format('%I.%I', user_view_schema, user_view_name)::regclass from _timescaledb_catalog.continuous_agg where materialized_only = false
+    FOR vname, materialized_only IN select format('%I.%I', cagg.user_view_schema, cagg.user_view_name)::regclass, cagg.materialized_only from _timescaledb_catalog.continuous_agg cagg
```
erimatnor, line 11: Not sure I follow what this change has to do with making aggregates fully qualified. Seem unrelated.

svenklemm: this rebuilds the view definition so caggs created in earlier versions have fully qualified aggs in view too, we used to only rebuild caggs with realtime aggregation but this changes the code to rebuild any cagg view definitions

mkindahl (approved):
This qualifies all symbols, even those that are in `pg_catalog`.

You might want to avoid unnecessary qualifications of functions, as this includes even built-in functions like `min` and `max`, you could use `format_procedure` and set the `search_path` to `pg_catalog` prior to calling this function. I see that as an improvement though, so approving anyway.

sql/updates/post-update.sql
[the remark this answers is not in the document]
svenklemm: this runs at the end of the update script so by this time everybody has materialized_only

svenklemm:
> This qualifies all symbols, even those that are in `pg_catalog`.
> 
> You might want to avoid unnecessary qualifications of functions, as this includes even built-in functions like `min` and `max`, and could in that case use `format_procedure` and set the `search_path` to `pg_catalog` prior to calling this function. I see that as an improvement though, so approving anyway.

I think this is actually a plus and this way we are completely independent of search_path

gayyappan (approved):

PR #3329, opened 2021-06-14
Fix flaky pg_dump test

svenklemm:
Call pg_terminate_backend on all active sessions on the database to
be used as template database to prevent on error on CREATE DATABASE.

### DIFF
--- a/test/sql/pg_dump.sql
+++ b/test/sql/pg_dump.sql
@@ -183,6 +183,13 @@ drop function get_sqlstate(TEXT);
 
 \c postgres :ROLE_SUPERUSER
 --need to shutdown workers to use db as template
+--<exclude_from_test>
+SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = :'TEST_DBNAME';
+--</exclude_from_test>
+-- there should be no sessions for that database
+-- columns are explicit so the output is the same on PG12 and PG13 since PG13 has extra column leader_pid
+SELECT datid,datname,pid,usesysid,usename,application_name,client_addr,client_hostname,client_port,backend_start,xact_start,query_start,state_change,wait_event_type,wait_event,state,backend_xid,backend_xmin,query,backend_type FROM pg_stat_activity WHERE datname = :'TEST_DBNAME';
+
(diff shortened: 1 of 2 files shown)


test/sql/pg_dump.sql
```diff
@@ -183,6 +183,10 @@ drop function get_sqlstate(TEXT);
 
 \c postgres :ROLE_SUPERUSER
 --need to shutdown workers to use db as template
+--<exclude_from_test>
+SELECT pg_cancel_backend(pid) FROM pg_stat_activity WHERE datname = :'TEST_DBNAME';
+--</exclude_from_test>
```
nikkhils, line 188: Hmm, this still leaves a mystery about what these `backends` are doing in the first place..

svenklemm: i could put a select after so a failed test would have the listing of the sessions to that database which would help us diagnosing this further
even with pg_terminate_backend there is still potential for race condition since pg_cancel_backend doesnt wait for the process to quit

afiskon (approved):

test/sql/pg_dump.sql
mfundul: Can we do something like this?
https://github.com/postgres/postgres/blob/817bb0a7d1e02df4643d37304aed8574cf43f629/contrib/dblink/sql/dblink.sql#L125

svenklemm: wait_pid is not a function that is installed by default so i'll punt on that. The additional query should help us pinpoint what is actually blocking the create should it still fail with this patch.
```
crash=# \df+ wait_pid
                                                                          List of functions
 Schema | Name | Result data type | Argument data types | Type | Volatility | Parallel | Owner | Security | Access privileges | Language | Source code | Description
--------+------+------------------+---------------------+------+------------+----------+-------+----------+-------------------+----------+-------------+-------------
(0 rows)
```

mfundul: Would it make sense to copy the `wait_pid()` code for usage in our own testing frameworks? I do not know if this is possible though.

mfundul (approved):

PR #3332, opened 2021-06-14
Rename dist_hypertable_pg12 test

svenklemm:
Combine dist_hypertable_pg12 test since all the tests in that file
can run on all supported PG versions now.
We also rename the views test to information_views to make it clearer
what the test is about and rename the join test to pg_join since
this is the postgres join test ported to hypertables.

Disable-check: commit-count


erimatnor (commented):
This test is not only testing table access methods. It is also testing other features introduced with PG12, like GENERATED columns. Renaming to `tableam` is thus not accurate.

In fact, I believe this was once named tableam, but was changed once other tests were added.

erimatnor:
If tableam and generated columns are now supported on all our supported PG versions, I think it is better to integrate this test with the regular dist_hypertable test instead of renaming to tableam.

nikkhils (approved):
LGTM

mfundul (approved):

PR #3334, opened 2021-06-14
Remove two_buckets_to_str() procedure

afiskon:
```
Remove two_buckets_to_str() procedure

The two_buckets_to_str() procedure relies on fixed bucket_width
which is not going to be fixed in the future. Since the procedure
is used only to generate a hint that accompanies an error message,
the simplest solution is to remove this procedure. We can improve
error messages later if that would be necessary.
```

### DIFF
--- a/tsl/src/bgw_policy/continuous_aggregate_api.c
+++ b/tsl/src/bgw_policy/continuous_aggregate_api.c
@@ -342,32 +342,6 @@ interval_to_int64(Datum interval, Oid type)
 	return 0;
 }
 
-static const char *
-two_buckets_to_str(const ContinuousAgg *cagg)
-{
-	Oid bucket_type;
-	Oid outfuncid;
-	int64 two_buckets;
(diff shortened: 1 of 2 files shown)


mfundul (approved):

erimatnor (commented):
I guess it is fine, but it is a little bit unclear why this error and function cannot be replaced with the variable size whenever that's available. Whey does it need to be removed now?

afiskon:
> I guess it is fine, but it is a little bit unclear why this error and function cannot be replaced with the variable size whenever that's available. Whey does it need to be removed now?

I generally try to keep the code / pull requests / feature implementation as small as possible (== MVP). If we keep this code it means that in further pull requests I will have to modify it as well which seems like an unnecessary time investment for me and reviewers, and next to zero gain for the users. This being said I'm fine with keeping the code if the team believes it's worth it.

svenklemm (approved):

PR #3317, opened 2021-06-09
Fix memory overflow

mfundul:
Fix heap buffer overflow in remote connection cache.

Fixes #3315

PR #3335, opened 2021-06-14
Copy/move chunk refactoring

pmwkaa:
Remove `copy_chunk_data()` function and code needed to support it,
such as the 'transactional' argument.

Introduce `copy_chunk()` API function as an internal wrapper for
the `move_chunk()`.


sql/maintenance_utils.sql
```diff
@@ -42,3 +34,17 @@ CREATE OR REPLACE FUNCTION recompress_chunk(
     if_not_compressed BOOLEAN = false
 ) RETURNS REGCLASS AS '@MODULE_PATHNAME@', 'ts_recompress_chunk' LANGUAGE C STRICT VOLATILE;
 
+-- Experimental features
+CREATE OR REPLACE PROCEDURE timescaledb_experimental.move_chunk(
```
erimatnor, line 38: I suggest putting these functions in `ddl_experimental.sql` so that we collect all experimental DDL commands in one file.

src/chunk.c
```diff
@@ -2305,7 +2304,7 @@ ts_chunk_data_nodes_copy(const Chunk *chunk)
 }
 
 Chunk *
-ts_chunk_copy(const Chunk *chunk)
+ts_chunk_duplicate(const Chunk *chunk)
```
erimatnor, line 2307: Why this change? Is it to avoid confusion with the DDL function with similar name? 

I worry that "duplicate" is somewhat confusing. Think we also use "copy" semantics for other internal objects.

tsl/src/chunk.c
```diff
@@ -336,6 +336,24 @@ chunk_create_replica_table(PG_FUNCTION_ARGS)
...
+	snprintf(application_name.data,
+			 sizeof(application_name.data),
+			 "%s:%s",
+			 ccd->operation_id.data,
+			 ccd->completed_stage);
+
+	pgstat_report_appname(application_name.data);
+
+	chunk_copy_activity_update(ccd);
+}
```
erimatnor, line 355: Is there any case where we shouldn't commit and start a new transaction when we transition to another stage? Wondering if it makes sense to put that at the end of this function instead of repeating every time we call the function.

tsl/src/chunk.h
```diff
@@ -8,17 +8,36 @@
 
 #include <postgres.h>
 #include <fmgr.h>
-
 #include <chunk.h>
 
+/* To track a chunk move or copy activity */
+typedef struct ChunkCopyData
```
erimatnor, line 14: So, I think we want to embed the FormData in this struct since they overlapped a lot in contents. This is also the typical pattern we have for data structures we persist to metadata tables, like `Hypertable` or `Dimension`.

As pointed out in the previous PR, the FormData is defined in catalog.h but is currently not used for anything.

erimatnor (commented):
Overall looks good. I have some comments/suggestions inline, though.

src/chunk.c
pmwkaa: I wanted to reuse chunk_copy name and found this collision which I found a bit confusing regarding its name and what it does. Thought that naming it as a duplicate is more representative actually

tsl/src/chunk.h
pmwkaa: So the idea to remove it completely?

tsl/src/chunk_copy.c
```diff
@@ -0,0 +1,581 @@
...
+
+struct ChunkCopyStage
+{
+	const char *name;
+	chunk_copy_stage_func function;
+	/* todo: abort function */
+};
+
+/* To track a chunk move or copy activity */
+struct ChunkCopy
```
erimatnor, line 72: I think we should keep the FormData struct in catalog.h and instead follow the established pattern of emdedding the FormData in a larger struct  like follows:


```
typedef struct FormData_chunk_copy_activity
    {
      
    	int32 id;
    	NameData operation_id;
    	int32 backend_pid;
    	text *completed_stage;
    	TimestampTz time_start;
    	int32 chunk_id;
    	NameData source_node_name;
    	NameData dest_node_name;
    	bool delete_on_src_node;
};

struct ChunkCopy
{
     struct FormData_chunk_copy_activity fd;
     ChunkCopyStage *stage;
     Chunk *chunk;
     ForeignServer *src_server;
     ForeignServer *dst_server;
     MemoryContext mctx;
}
```

I would also consider making the `completed_stage` a NameData type instead of text (I don't think we need more than 63  chars for the stage names), since that would make the FormData struct fixed size. That in turn allows using GETSTRUCT on the tuple instead of `heap_deform`.

tsl/src/chunk_copy.c
```diff
@@ -0,0 +1,581 @@
...
+static void
+chunk_copy_stage_delete_chunk(ChunkCopy *cc)
+{
+	if (!cc->delete_on_src_node)
+		return;
+
+	chunk_api_call_chunk_drop_replica(cc->chunk, cc->src_node, cc->src_server->serverid);
+}
+
+static ChunkCopyStage copy_stages[] = {
```
erimatnor, line 527: `👍 for this.

But I think this should be const, right? Or is it not possible for some reason?

```suggestion
static const ChunkCopyStage copy_stages[] = {
```

tsl/src/chunk_copy.c
```diff
@@ -0,0 +1,581 @@
...
+	/* Commit to get out of starting transaction */
+	PopActiveSnapshot();
+	CommitTransactionCommand();
+}
+
+static void
+chunk_copy_cleanup(ChunkCopy *cc)
+{
+	/* Done using this long lived memory context */
+	MemoryContextReset(cc->mcxt);
```
erimatnor, line 367: Since you allocated a new context, should this be:

```suggestion
	MemoryContextDelete(cc->mcxt);
```

erimatnor (commented):
Generally, this looks very nice and a lot cleaner than previously. 👍 

A few nits, and potential fixes needed, however.

src/catalog.h
```diff
@@ -1201,21 +1201,6 @@ enum Anum_chunk_copy_activity
 
 #define Natts_chunk_copy_activity (_Anum_chunk_copy_activity_max - 1)
 
-typedef struct FormData_chunk_copy_activity
```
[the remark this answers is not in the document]
pmwkaa: I guess I misunderstood the idea of embedding then

src/chunk.c
nikkhils: +1 for keeping the earlier name. `duplicate` is confusing.

src/catalog.h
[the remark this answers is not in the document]
erimatnor: > I guess I misunderstood the idea of embedding then, sorry

I probably wasn't clear :) I put an example further down.

tsl/src/chunk_copy.c
```diff
@@ -0,0 +1,581 @@
...
+{
+	ChunkCopy *cc = data;
+	Datum values[Natts_chunk_copy_activity];
+	bool nulls[Natts_chunk_copy_activity];
+	CatalogSecurityContext sec_ctx;
+	bool should_free;
+	HeapTuple tuple = ts_scanner_fetch_heap_tuple(ti, false, &should_free);
+	HeapTuple new_tuple;
+
+	heap_deform_tuple(tuple, ts_scanner_get_tupledesc(ti), values, nulls);
```
erimatnor, line 151: This is completely optional, but now it is possible to copy the tuple and update it directly with GETSTRUCT to get a pointer to the FormData. This might simplify the code a bit, but this deform and form works too.

tsl/src/chunk_copy.c
```diff
@@ -0,0 +1,581 @@
...
+
+	/* We only update the "completed_stage" field */
+	Assert(NULL != cc->stage);
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_completed_stage)] =
+		CStringGetDatum(cc->stage->name);
```
erimatnor, line 156: Note that this is now Name type. In other places we `DirectFunctionaCall1(namein, name)` to construct a name type. Alternatively, create a `NameData` on the stack and `namestrcpy()` the name into it followed by `NameGetDatum`

erimatnor (approved):
Approving, but two small things to look at/fix.

tsl/src/chunk_copy.c
```diff
@@ -0,0 +1,581 @@
...
+	/* Create the publication in autocommit mode */
+	ts_dist_cmd_run_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), false);
+}
+
+static void
+chunk_copy_stage_create_replication_slot(ChunkCopy *cc)
+{
+	const char *cmd;
+
+	/* Create replication slot on the destination data node in a new transaction block */
```
nikkhils, line 411: This comment should be removed now.

tsl/src/chunk_copy.c
```diff
@@ -0,0 +1,581 @@
...
+	for (stage = &chunk_copy_stages[0]; stage->name != NULL; stage++)
+	{
+		StartTransactionCommand();
+
+		cc.stage = stage;
+		cc.stage->function(&cc);
+
+		/* Mark current stage as completed and update the catalog */
+		chunk_copy_activity_update(&cc);
+
```
nikkhils, line 575: The `PG_TRY/PG_CATCH` block changes are pending still.

tsl/src/chunk_copy.c
pmwkaa: But it still requires to fill values in in order to update tuple, not sure that it is more convenient here to use

nikkhils (approved):
A couple of additional comments from me. Looks pretty good overall

tsl/src/chunk_copy.c
pmwkaa: `NameGetDatum` translates to the `CStringGetDatum` in the macro if I am not mistaken

mfundul: It's `#define NameGetDatum(X) CStringGetDatum(NameStr(*(X)))` from:
```
/*
 * Representation of a Name: effectively just a C string, but null-padded to
 * exactly NAMEDATALEN bytes.  The use of a struct is historical.
 */
typedef struct nameData
{
	char		data[NAMEDATALEN];
} NameData;
typedef NameData *Name;

#define NameStr(name)	((name).data)
```
A recent bugfix in master was so that we don't get memory overflows by not allocating enough `char` entries. See [here](https://github.com/timescale/timescaledb/pull/3317#issuecomment-858459915)

pmwkaa: I guess the problem was that the length of the string pointer for name type was larger then `NAMEDATALEN`?

mfundul: The opposite actually, it would `memcpy()` from the source which didn't have `NAMEDATALEN` characters allocated because it was a C string, and it would do the memory overflow.

pmwkaa: Ok cool, using `namein` function as suggested

pmwkaa: > The opposite actually, it would `memcpy()` from the source which didn't have `NAMEDATALEN` characters allocated because it was a C string, and it would do the memory overflow.

This is exactly what I meant to say :) good to know

PR #3336, opened 2021-06-15
Fix pg_init_privs objsubid handling

### LINKED ISSUE #3331: error during upgrade to 2.2.1
 - OS: CentOS 
 - TimescaleDB version (output of `\dx` in `psql`): 1.7.3 -> 2.2.1 upgrade
 - Installation method: source

**Describe the bug**
When upgrading timescaledb, I ran into the following error:

```
db=# alter extension timescaledb update;
ERROR:  duplicate key value violates unique constraint "saved_privs_tmpnsp_tmpname_key"
DETAIL:  Key (tmpnsp, tmpname)=(_timescaledb_catalog, chunk_constraint_name) already exists.
```

During the pre-upgrade phase, TSDB saves the pg_init_privs temporarily into a table, with the following SQL.

```
-- Table for ACL and initprivs of tables. 
CREATE TABLE saved_privs(
       tmpnsp name,
       tmpname name,
       tmpacl aclitem[],
       tmpini aclitem[],
       UNIQUE (tmpnsp, tmpname));

-- We save away both the ACL and the initprivs for all tables and
-- views in the extension (but not for chunks and internal objects) so
-- that we can restore them to the proper state after the update.
INSERT INTO saved_privs
SELECT nspname, relname, relacl, initprivs
  FROM pg_class cl JOIN pg_namespace ns ON ns.oid = relnamespace
                   JOIN pg_init_privs ip ON ip.objoid = cl.oid
WHERE nspname IN ('_timescaledb_catalog', '_timescaledb_config')
   OR nspname = '_timescaledb_internal'
  AND relname IN ('hypertable_chunk_local_size', 'compressed_chunk_stats',
                  'bgw_job_stat', 'bgw_policy_chunk_stats');
```

The UNIQUE constraint of this table was validated. Indeed, when looking into the catalog, I found multiple rows in pg_init_privs for the same `(tmpnsp, tmpname)`.
I "fixed" it manually by removing one row from the pg_init_privs table so that it was unique again, and then I ran the upgrade. But I wonder why the upgrade script does not support multiple rows here. At first, I was going to do dismiss it as some strange one-off, but I've encountered this twice now during an upgrade. I'm not sure under what circumstances this occurs exactly though.

svenklemm:
pg_init_privs can have multiple entries per relation if the relation
has per column privileges. An objsubid different from 0 means that
the entry is for a column privilege. Since we do not currently
restore column privileges we have to ignore those rows otherwise
the update script will fail when tables with column privileges are
present.

Fixes #3331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,12 @@ accidentally triggering the load of a previous DB version.**
 **Bugfixes**
 * #3305 Fix pull_varnos miscomputation of relids set
 * #3327 Make aggregates in caggs fully qualified
+* #3336 Fix pg_init_privs objsubid handling
 
 **Thanks**
 * @db-adrian for reporting an issue when accessing cagg view through postgres_fdw
 * @fncaldas and @pgwhalen for reporting an issue accessing caggs when public is not in search_path
+* @fvannee, @mglonnro and @ebreijo for reporting an issue with the upgrade script
 
 ## 2.3.0 (2021-05-25)
 
--- a/sql/updates/post-update.sql
+++ b/sql/updates/post-update.sql
@@ -42,12 +42,12 @@ DROP FUNCTION IF EXISTS _timescaledb_internal.cagg_watermark(oid);
 -- first installation.
 INSERT INTO saved_privs
      SELECT nspname, relname, relacl,
- 	    (SELECT tmpini FROM saved_privs
-	      WHERE tmpnsp = '_timescaledb_catalog' AND tmpname = 'chunk')
+       (SELECT tmpini FROM saved_privs
+        WHERE tmpnsp = '_timescaledb_catalog' AND tmpname = 'chunk')
        FROM pg_class JOIN pg_namespace ns ON ns.oid = relnamespace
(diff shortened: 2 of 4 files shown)


mkindahl (approved):
LGTM

nikkhils (approved):
LGTM

fvannee:
To me this looks like it fixes the case where `objsubid != 0`, however I don't think that was the actual issue in our case.
I don't have the full details anymore I'm afraid.. But what I know was that, in both cases that I ran into the error, I had two rows in `pg_init_privs` with the same `objoid` and a different `classoid`. In none of the cases I had set any kind of per-column privileges.

In order to fix the catalog, I ran this:

```
delete from pg_init_privs where objoid=16405 and classoid=1259;
```

This deleted 1 row from the catalog table and the other row with `objoid=16405` remained.

svenklemm:
@fvannee  Thanks for the update i've created a PR to limit the query to only pg_class classoids to prevent this from happening.

https://github.com/timescale/timescaledb/pull/3338

PR #3338, opened 2021-06-15
Harden pg_init_privs query

svenklemm:
Since we are only interested in entries with classoid pg_class
our queries should reflect that. Without these restrictions
objects that have entries for multiple classoids can cause the
extension update to fail.


mkindahl (approved):

nikkhils (approved):
LGTM

PR #3339, opened 2021-06-16
Fix updates to handle sequences

nikkhils:
The post-update script was handling preserving initprivs for newly
added catalog tables and views. However, newly added catalog sequences
need separate handling otherwise update tests start failing. We also
now grant privileges for all future sequences in the update tests.

In passing, default the PG_VERSION in the update tests to 12 since we
don't work with PG11 anymore.


mkindahl (approved):
LGTM. You need to fix the commit message though so that it doesn't look like the extension sets default privileges for all installations. The change you have is only for the update test.

nikkhils:
> LGTM. You need to fix the commit message though so that it doesn't look like the extension sets default privileges for all installations. The change you have is only for the update test.

Ah yes, the commit message suggests that whole sale. Will change

svenklemm (approved):

PR #3342, opened 2021-06-16
Filter out chunk and hypertable ids from test output

svenklemm:
Currently any tests in regresscheck-shared can only include
EXPLAIN output if they only access the precreated hypertables
as hypertables and chunks created in the test itself will
have ids depending on execution order of the shared tests.
This patch filters out those ids so shared tests can include
EXPLAIN output for local hypertables.


gayyappan (approved):

nikkhils (approved):
Nice trick!

PR #3346, opened 2021-06-20
Move timestamp_limits and with_clause_parser test

svenklemm:
Move the timestamp_limits and with_clause_parser test to
regresscheck-shared since those tests don't need a private
database incurring less overhead to run those tests.

### DIFF
--- a/tsl/test/shared/sql/include/shared_setup.sql
+++ b/tsl/test/shared/sql/include/shared_setup.sql
@@ -6,6 +6,8 @@ SET client_min_messages TO ERROR;
 
 SET ROLE :ROLE_DEFAULT_PERM_USER;
 
+CREATE SCHEMA test;
+
 -- create normal hypertable with dropped columns, each chunk will have different attribute numbers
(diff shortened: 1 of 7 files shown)


tsl/test/shared/sql/include/shared_setup.sql
```diff
@@ -6,6 +6,8 @@ SET client_min_messages TO ERROR;
 
 SET ROLE :ROLE_DEFAULT_PERM_USER;
 
+CREATE SCHEMA test;
```
gayyappan, line 9: why are we adding this?

gayyappan (approved):
changes look good. Question about a test change.

tsl/test/shared/sql/include/shared_setup.sql
svenklemm: timestamp_limits test uses this and on normal regresscheck we precreate that schema too

mfundul (approved):

PR #3350, opened 2021-06-21
time_bucket_ng() may be IMMUTABLE depending on arguments

afiskon:
```
time_bucket_ng() may be IMMUTABLE depending on arguments

This function is IMMUTABLE when it doesn't accept timestamptz arguments,
and STABLE otherwise. See the comments in sql/time_bucket_ng.sql for
more details.
```

### DIFF
--- a/sql/time_bucket_ng.sql
+++ b/sql/time_bucket_ng.sql
@@ -9,18 +9,32 @@
 -- and the interface of this function are subjects to change. There could
 -- be bugs, and the implementation doesn't claim to be complete. Use at
 -- your own risk.
+--
+-- This function is IMMUTABLE when it doesn't accept timestamptz arguments,
+-- and STABLE otherwise. The reason is that occasionally timezones change.
+-- When dealing with timestamptz's from the far future, it's possible that
+-- between now and then the rules for given TZ will change by local laws. Which
+-- makes the function STABLE by the definition [1].
+--
+-- We don't forbid users to work with timestamptz's from the future, nor warn
+-- about this corner case. This behavior is consistent with PostgreSQL
+-- behavior [2].
+--
+-- [1]: https://www.postgresql.org/docs/current/xfunc-volatility.html
+-- [2]: https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-TIMEZONES
+--
 CREATE OR REPLACE FUNCTION timescaledb_experimental.time_bucket_ng(bucket_width INTERVAL, ts DATE) RETURNS DATE
-	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng' LANGUAGE C STABLE PARALLEL SAFE STRICT;
+	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng' LANGUAGE C IMMUTABLE PARALLEL SAFE STRICT;
 
 CREATE OR REPLACE FUNCTION timescaledb_experimental.time_bucket_ng(bucket_width INTERVAL, ts DATE, origin DATE) RETURNS DATE
-	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng' LANGUAGE C STABLE PARALLEL SAFE STRICT;
+	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng' LANGUAGE C IMMUTABLE PARALLEL SAFE STRICT;
 
 -- utility functions
 CREATE OR REPLACE FUNCTION timescaledb_experimental.time_bucket_ng(bucket_width INTERVAL, ts TIMESTAMP) RETURNS TIMESTAMP
-	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng_timestamp' LANGUAGE C STABLE PARALLEL SAFE STRICT;
+	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng_timestamp' LANGUAGE C IMMUTABLE PARALLEL SAFE STRICT;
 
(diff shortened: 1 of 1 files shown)


sql/time_bucket_ng.sql
```diff
@@ -9,21 +9,32 @@
...
+-- because it's going to get timezone explicitly in text format, e.g.:
+--
+--   SELECT time_bucket_ng('3 months', now() :: timestamptz, 'Europe/Berlin');
+--
+-- In other words, there will be separate continuous aggregates for each
+-- timezone. The timestamptz argument is not a problem, since internally it's
+-- represented as a simple UTC timestamp.
 CREATE OR REPLACE FUNCTION timescaledb_experimental.time_bucket_ng(bucket_width INTERVAL, ts DATE) RETURNS DATE
-	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng' LANGUAGE C STABLE PARALLEL SAFE STRICT;
+	AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng' LANGUAGE C IMMUTABLE PARALLEL SAFE STRICT;
```
erimatnor, line 24: The function isn't taking an explicit time zone as argument right now, so why make it IMMUTABLE before that functionality is available. 

What is the behavior in the meantime? Do we hard-code a time zone as if the function was always called with that explicit time zone? Or, will the behavior not reflect the volatility class?

FWIW, it makes more sense to me to first add time zone argument support, then change the volatility.

afiskon: @erimatnor Thanks for the good question!

The reason is that I'm already facing problems because of the fact that the procedure is STABLE. If you try to use it in continuous aggregates you get "only immutable functions supported in continuous aggregate view" error and this is not easy to bypass because functions you could use to do it are declared as static in PostgreSQL code.

svenklemm: From https://www.postgresql.org/docs/current/xfunc-volatility.html
```
An IMMUTABLE function cannot modify the database and is guaranteed to return the same results given the same arguments forever.
```
Any function dealing with timestamptz and timezones cannot be IMMUTABLE since rules about timezones change all the time.

xzilla:
Be aware that since this change is explicitly lying to Postgres, it will be possible to introduce subtle bugs into user space code when the function returns the wrong result, but that probably went happen very often, so maybe that's not a concern for you ¯\_(ツ)_/¯

afiskon:
> Be aware that since this change is explicitly lying to Postgres
There is no lie. This function is and should be IMMUTABLE, otherwise, we have a bug in our code

sql/time_bucket_ng.sql
afiskon: Thanks, this is actually not a bad point. But the rules change only in the future. If now the time is 2021-06-22 11:39 MSK it will never change 🤷‍♂️ I don't think that any of our users going to work with timestamptz's from the future, but we can add a corresponding check and throw an error if they do. Then we are fine to use IMMUTABLE, right?

afiskon: I realized that we are not, by the definition. Have to think a little about it.

erimatnor:
@afiskon do you know why we don't support stable functions in continuous aggregates? Regular materialized views certainly works with stable functions, e.g., date_trunc. Perhaps it is because we continuously update? Still, I am not sure why we care about that because it seems odd for someone to change time zone on a running system and if they do I think it is on them to make sure their continuous aggregates are refreshed to reflect that.

afiskon:
> @afiskon do you know why we don't support stable functions in continuous aggregates? Regular materialized views certainly works with stable functions, e.g., date_trunc. Perhaps it is because we continuously update?

That's because of continuous update, yes. With regular views or materialized views STABLE functions are not a problem. STABLE functions guarantee to return the same results for the same rows in a given statement. When you refresh the view, that's another statement, and it's OK for a STABLE function to return other values. We don't refresh the entire aggregate thus it's a problem for us and that's why we need this function to be IMMUTABLE.

> Still, I am not sure why we care about that because it seems odd for someone to change time zone on a running system and if they do I think it is on them to make sure their continuous aggregates are refreshed to reflect that.

Good point. Personally, I'm fine with this approach too.

afiskon:
Or we can do both. Throw an error by default and explain the reason, but if the user wants to override it there is an optional `allow_timestamptz_from_the_future => true` flag 🤷‍♂️

afiskon:
@erimatnor @svenklemm It occurred to me that we don't have to choose the exact solution straight away since the procedure doesn't support timezones yet. I documented the discussion in the comments for time_bucket_ng() procedure and possible approaches to the timestamptz-from-the-future problem. Please let me know what do you think.

erimatnor:
> > Still, I am not sure why we care about that because it seems odd for someone to change time zone on a running system and if they do I think it is on them to make sure their continuous aggregates are refreshed to reflect that.
> 
> Good point. As I understand @svenklemm 's concern is that it will violate the definition of IMMUTABLE. But personally, I'm fine with either approach.

I am not suggesting we violate the definition of IMMUTABLE. I am suggesting we use the "right" volatility class for bucketing functions, including STABLE, but then leave it up to users to deal with any inconsistencies related to refreshing across time zone changes in the database. If we still don't want to support STABLE functions in caggs, we shouldn't make things IMMUTABLE when they shouldn't be just to make it work. In that case I think it is OK to only allow the immutable versions of the bucketing functions (i.e., those with an explicit time zone argument).

afiskon:
For the record, we had a Zoom call to discuss this issue. We agreed that the function should be STABLE when accepting timestamptz arguments, and IMMUTABLE otherwise. We also agreed that Caggs should work with IMMUTABLE and STABLE functions.

mfundul (approved):

svenklemm (approved):

PR #3366, opened 2021-06-25
Allow to use experimental functions in caggs

afiskon:
```
Allow to use experimental functions in caggs

Current implementation of caggs can't find a bucketing function
if it's declared in the experimental schema. This patch fixes it.

Also the patch adds `debug_notice` test to IGNORE list on AppVeyor.
The corresponding test generates an extra "DEBUG: rehashing catalog
cache" message which is not critical. It seems to be stable on Linux.
````

### CODE REVIEW

File: src/continuous_agg.c [hunk 1]
```diff
@@ -1093,4 +1093,4 @@ int64
 ts_continuous_agg_max_bucket_width(const ContinuousAgg *agg)
 {
 	return agg->data.bucket_width;
-}
\ No newline at end of file
+}
```
afiskon, line 1096: These changes are made by `make format` 🤷‍♂️

mfundul (approved):

mfundul:
Maybe figure out a way to fix those "no new line at the end of file".

tsl/src/continuous_aggs/create.c
```diff
@@ -1071,7 +1071,7 @@ is_valid_bucketing_function(Oid funcid)
 {
 	FuncInfo *finfo = ts_func_cache_get_bucketing_func(funcid);
 
-	return finfo != NULL && finfo->is_timescaledb_func && finfo->nargs == 2;
+	return finfo != NULL && (finfo->origin != ORIGIN_POSTGRES) && finfo->nargs == 2;
```
gayyappan, line 1074: This affects the public cagg api.

afiskon: I don't see how exactly. Maybe you could give an example?

gayyappan (commented):
My concern about this PR is that it introduces changes to the public cagg implementation. This exposes the experimental time bucket api via caggs. We need to discuss if this is the correct approach before committing into master.

tsl/src/continuous_aggs/create.c
afiskon: I mean: it changes the implementation for sure. I don't see how the API is affected.

src/extension.c
```diff
@@ -202,6 +202,12 @@ ts_extension_schema_name(void)
 	return get_namespace_name(ts_extension_schema_oid());
 }
 
+const char *
+ts_experimental_schema_name(void)
+{
+	return "timescaledb_experimental";
```
pmwkaa, line 208: Feels like this name should be put into macro instead of hardcoding it here

tsl/src/continuous_aggs/create.c
pmwkaa: Maybe check `ORIGIN_TIMESCALE` || `ORIGIN_TIMESCALE_EXPERIMENTAL` in such case? Feel like this change allows any function, not only provided by the extension, if I understood correctly :)

afiskon: @pmwkaa Thanks for the feedback. Fixed.

tsl/src/continuous_aggs/create.c
```diff
@@ -1069,9 +1069,18 @@ get_partialize_funcexpr(Aggref *agg)
 static bool
 is_valid_bucketing_function(Oid funcid)
 {
+	bool is_timescale;
 	FuncInfo *finfo = ts_func_cache_get_bucketing_func(funcid);
 
-	return finfo != NULL && finfo->is_timescaledb_func && finfo->nargs == 2;
+	if (finfo != NULL)
```
pmwkaa, line 1075: I think this should be `finfo == NULL`

tsl/test/sql/continuous_aggs_experimental.sql
```diff
@@ -0,0 +1,47 @@
...
+       MIN(temperature),
+       MAX(temperature)
+FROM conditions
+GROUP BY city, bucket;
+
+SELECT to_char(bucket, 'YYYY-MM-DD'), city, min, max
+FROM conditions_summary_weekly
+ORDER BY bucket;
+
+DROP TABLE conditions CASCADE;
```
pmwkaa, line 47: nit: And here

tsl/test/sql/continuous_aggs.sql
```diff
@@ -1026,4 +1026,4 @@ FROM water_consumption
 GROUP BY sensor_id, time_bucket(INTERVAL '1 minute', timestamp)
 ORDER BY water_consumption;
 
-DROP TABLE water_consumption CASCADE;
+DROP TABLE water_consumption CASCADE;
```
pmwkaa, line 1029: nit: maybe add a new line here, so the github would not complain about it :)

pmwkaa (approved):
Need to fix the finfo comparison, otherwise looks good

tsl/src/continuous_aggs/create.c
afiskon: Oops my bad :)

PR #3372, opened 2021-06-29
Make move_chunk use AN txns on DN

nikkhils:
We used to run transactions in autocommit mode on DN while running the
chunk copy/move activity. This meant that any failures on the access
node were de-coupled from the activity on the DN. This can make future
cleanup messy since we wouldn't know what's failed/succeeded on the
data nodes.

We now drive the entire activity via transactions started on the access
node.

### DIFF
--- a/tsl/src/chunk_copy.c
+++ b/tsl/src/chunk_copy.c
@@ -400,7 +400,7 @@ chunk_copy_stage_create_publication(ChunkCopy *cc)
 											  NameStr(cc->chunk->fd.table_name)));
 
 	/* Create the publication in autocommit mode */
-	ts_dist_cmd_run_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), false);
+	ts_dist_cmd_run_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), true);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


tsl/src/chunk_copy.c
```diff
@@ -442,29 +442,52 @@ chunk_copy_stage_sync_start(ChunkCopy *cc)
...
 }
 
 static void
 chunk_copy_stage_drop_subscription(ChunkCopy *cc)
 {
-	const char *cmd;
+	char *cmd;
 
+	/* Stop data transfer on the destination node */
+	cmd = psprintf("ALTER SUBSCRIPTION %s DISABLE", NameStr(cc->fd.operation_id));
```
pmwkaa, line 478: It might makes sense to separate all of those steps in three separate stages, since we have a separate stage to create the replication slot

pmwkaa (approved):

tsl/src/chunk_copy.c
nikkhils: We created these additional substeps just to allow things to work in a transaction block. Semantically it's all about dropping the subscription.

pmwkaa: Yes, but from point of view of the healing function which way is more convenient to operate?

nikkhils: it's not an issue. The cleanup function will be exactly the same as this one with these 3 steps part of a transaction block

erimatnor (approved):

tsl/src/chunk_copy.c
```diff
@@ -400,7 +400,7 @@ chunk_copy_stage_create_publication(ChunkCopy *cc)
 											  NameStr(cc->chunk->fd.table_name)));
 
 	/* Create the publication in autocommit mode */
-	ts_dist_cmd_run_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), false);
+	ts_dist_cmd_run_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), true);
```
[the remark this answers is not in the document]
nikkhils: Agreed.

PR #3394, opened 2021-07-06
Remove copy_chunk_data from downgrade script

erimatnor:
The internal function `copy_chunk_data` was removed as part of
refactoring and is no longer necessary to remove in the downgrade
script since the function was never part of a release.


nikkhils (approved):
LGTM

pmwkaa (approved):

PR #3395, opened 2021-07-06
Fix issues in the dist_chunk test

erimatnor:
A test that drops the primary chunk replica was hard coded to drop the
chunk on a specific data node, which wasn't the primary. Thus the
chunk's primary node wasn't changed as expected after the 
chunk replica was dropped. This also caused flakiness because the
ordering of foreign servers for the chunk could vary. These issues
have been fixed.

### DIFF
--- a/tsl/test/shared/expected/dist_chunk.out
+++ b/tsl/test/shared/expected/dist_chunk.out
@@ -3,30 +3,38 @@
 -- LICENSE-TIMESCALE for a copy of the license.
 -- This file contains tests for all features that will be used as part
 -- of the chunk move/copy multi-node functionality
--- Test function _timescaledb_internal.create_chunk_replica_table
 -- A table for the first chunk will be created on the data node, where it is not present.
 SELECT chunk_name, data_nodes 
 FROM timescaledb_information.chunks 
-WHERE hypertable_name = 'dist_chunk_copy';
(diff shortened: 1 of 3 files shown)


nikkhils (approved):
LGTM

pmwkaa (approved):

tsl/test/shared/expected/dist_chunk.out
```diff
@@ -3,30 +3,38 @@
...
  _dist_hyper_X_X_chunk | {data_node_2,data_node_3}
-(5 rows)
+ _dist_hyper_X_X_chunk | {data_node_1,data_node_3}
+(11 rows)
 
 -- Non-distributed chunk will be used to test an error
 SELECT chunk_name 
 FROM timescaledb_information.chunks 
-WHERE hypertable_name = 'conditions';
+WHERE hypertable_name = 'conditions'
```
nikkhils, line 29: I guess we don't need an order by with that output ;-) but good for consistency I guess

nikkhils (approved):
LGTM!

pmwkaa (approved):

tsl/test/shared/expected/dist_chunk.out
erimatnor: I'll add just to be sure

PR #3396, opened 2021-07-06
Implement cleanup for chunk copy/move

nikkhils:
A chunk copy/move activity is carried out in stages and it can
fail in any of them. We track the last completed stage in the
"chunk_copy_activity" catalog table. In case of failure, a
"chunk_copy_cleanup" function can be invoked to bring the chunk back
to its original state on the source datanode and all transient objects
like replication slot, publication, subscription, empty chunk, metadata
updates, etc are cleaned up.

Includes test case changes for each and every stage induced failure.

### DIFF
--- a/sql/ddl_experimental.sql
+++ b/sql/ddl_experimental.sql
@@ -37,3 +37,12 @@ CREATE OR REPLACE PROCEDURE timescaledb_experimental.copy_chunk(
     source_node NAME = NULL,
     destination_node NAME = NULL)
 AS '@MODULE_PATHNAME@', 'ts_copy_chunk_proc' LANGUAGE C;
+
+-- A copy_chunk or move_chunk procedure call involves multiple nodes and
+-- depending on the data size can take a long time. Failures are possible
+-- when this long running activity is ongoing. We need to be able to recover
+-- and cleanup such failed chunk copy/move activities and it's done via this
+-- procedure
+CREATE OR REPLACE PROCEDURE timescaledb_experimental.cleanup_copy_chunk_operation(
+    operation_id NAME)
+AS '@MODULE_PATHNAME@', 'ts_copy_chunk_cleanup_proc' LANGUAGE C;
--- a/tsl/src/chunk_copy.c
+++ b/tsl/src/chunk_copy.c
@@ -41,6 +41,7 @@
 #include "chunk_api.h"
 #include "chunk_copy.h"
 #include "data_node.h"
+#include "debug_point.h"
 #include "remote/dist_commands.h"
 #include "dist_util.h"
 
@@ -65,14 +66,14 @@ struct ChunkCopyStage
 {
 	const char *name;
 	chunk_copy_stage_func function;
-	/* todo: abort function */
+	chunk_copy_stage_func function_cleanup;
 };
 
 /* To track a chunk move or copy activity */
 struct ChunkCopy
 {
 	/* catalog data */
-	FormData_chunk_copy_activity fd;
+	FormData_chunk_copy_operation fd;
 	/* current stage being executed */
 	const ChunkCopyStage *stage;
 	/* chunk to copy */
@@ -85,38 +86,37 @@ struct ChunkCopy
 };
 
 static HeapTuple
-chunk_copy_activity_make_tuple(const FormData_chunk_copy_activity *fd, TupleDesc desc)
+chunk_copy_operation_make_tuple(const FormData_chunk_copy_operation *fd, TupleDesc desc)
 {
-	Datum values[Natts_chunk_copy_activity];
-	bool nulls[Natts_chunk_copy_activity] = { false };
+	Datum values[Natts_chunk_copy_operation];
+	bool nulls[Natts_chunk_copy_operation] = { false };
 	memset(values, 0, sizeof(values));
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_id)] = Int32GetDatum(fd->id);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_operation_id)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_operation_id)] =
 		NameGetDatum(&fd->operation_id);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_backend_pid)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_backend_pid)] =
 		Int32GetDatum(fd->backend_pid);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_completed_stage)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_completed_stage)] =
 		NameGetDatum(&fd->completed_stage);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_time_start)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_time_start)] =
 		TimestampTzGetDatum(fd->time_start);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_chunk_id)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_chunk_id)] =
 		Int32GetDatum(fd->chunk_id);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_source_node_name)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_source_node_name)] =
 		NameGetDatum(&fd->source_node_name);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_dest_node_name)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_dest_node_name)] =
 		NameGetDatum(&fd->dest_node_name);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_delete_on_src_node)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_delete_on_src_node)] =
 		BoolGetDatum(fd->delete_on_src_node);
 	return heap_form_tuple(desc, values, nulls);
 }
 
 static void
-chunk_copy_activity_insert_rel(Relation rel, const FormData_chunk_copy_activity *fd)
+chunk_copy_operation_insert_rel(Relation rel, const FormData_chunk_copy_operation *fd)
 {
 	CatalogSecurityContext sec_ctx;
 	HeapTuple new_tuple;
 
-	new_tuple = chunk_copy_activity_make_tuple(fd, RelationGetDescr(rel));
+	new_tuple = chunk_copy_operation_make_tuple(fd, RelationGetDescr(rel));
 
 	ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
 	ts_catalog_insert(rel, new_tuple);
@@ -125,24 +125,24 @@ chunk_copy_activity_insert_rel(Relation rel, const FormData_chunk_copy_activity 
 }
 
 static void
-chunk_copy_activity_insert(const FormData_chunk_copy_activity *fd)
+chunk_copy_operation_insert(const FormData_chunk_copy_operation *fd)
 {
 	Catalog *catalog;
 	Relation rel;
 
 	catalog = ts_catalog_get();
-	rel = table_open(catalog_get_table_id(catalog, CHUNK_COPY_ACTIVITY), RowExclusiveLock);
+	rel = table_open(catalog_get_table_id(catalog, CHUNK_COPY_OPERATION), RowExclusiveLock);
 
-	chunk_copy_activity_insert_rel(rel, fd);
+	chunk_copy_operation_insert_rel(rel, fd);
 	table_close(rel, RowExclusiveLock);
 }
 
 static ScanTupleResult
-chunk_copy_activity_tuple_update(TupleInfo *ti, void *data)
+chunk_copy_operation_tuple_update(TupleInfo *ti, void *data)
 {
 	ChunkCopy *cc = data;
-	Datum values[Natts_chunk_copy_activity];
-	bool nulls[Natts_chunk_copy_activity];
+	Datum values[Natts_chunk_copy_operation];
+	bool nulls[Natts_chunk_copy_operation];
 	CatalogSecurityContext sec_ctx;
 	bool should_free;
 	HeapTuple tuple = ts_scanner_fetch_heap_tuple(ti, false, &should_free);
@@ -152,7 +152,7 @@ chunk_copy_activity_tuple_update(TupleInfo *ti, void *data)
 
 	/* We only update the "completed_stage" field */
 	Assert(NULL != cc->stage);
-	values[AttrNumberGetAttrOffset(Anum_chunk_copy_activity_completed_stage)] =
+	values[AttrNumberGetAttrOffset(Anum_chunk_copy_operation_completed_stage)] =
 		DirectFunctionCall1(namein, CStringGetDatum((cc->stage->name)));
 
 	new_tuple = heap_form_tuple(ts_scanner_get_tupledesc(ti), values, nulls);
@@ -168,14 +168,14 @@ chunk_copy_activity_tuple_update(TupleInfo *ti, void *data)
 }
 
 static int
-chunk_copy_activity_scan_update_by_id(int32 id, tuple_found_func tuple_found, void *data,
-									  LOCKMODE lockmode)
+chunk_copy_operation_scan_update_by_id(const char *operation_id, tuple_found_func tuple_found,
+									   void *data, LOCKMODE lockmode)
 {
 	Catalog *catalog = ts_catalog_get();
 	ScanKeyData scankey[1];
 	ScannerCtx scanctx = {
-		.table = catalog_get_table_id(catalog, CHUNK_COPY_ACTIVITY),
-		.index = catalog_get_index(catalog, CHUNK_COPY_ACTIVITY, CHUNK_COPY_ACTIVITY_PKEY_IDX),
+		.table = catalog_get_table_id(catalog, CHUNK_COPY_OPERATION),
+		.index = catalog_get_index(catalog, CHUNK_COPY_OPERATION, CHUNK_COPY_OPERATION_PKEY_IDX),
 		.nkeys = 1,
 		.limit = 1,
 		.scankey = scankey,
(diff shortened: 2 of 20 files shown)


tsl/src/chunk_copy.c
```diff
@@ -399,7 +416,7 @@ chunk_copy_stage_create_publication(ChunkCopy *cc)
 				   quote_qualified_identifier(NameStr(cc->chunk->fd.schema_name),
 											  NameStr(cc->chunk->fd.table_name)));
 
-	/* Create the publication in autocommit mode */
+	/* Create the publication */
 	ts_dist_cmd_run_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), true);
```
pmwkaa, line 420: I am wondering if we should execute those commands in the auto-commit mode since we check if the object exist anyway during the cleanup

tsl/src/chunk_copy.c
```diff
@@ -418,6 +435,71 @@ chunk_copy_stage_create_replication_slot(ChunkCopy *cc)
...
+	dist_res =
+		ts_dist_cmd_invoke_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), true);
+	res = ts_dist_cmd_get_result_by_node_name(dist_res, NameStr(cc->fd.source_node_name));
+
+	if (PQresultStatus(res) != PGRES_TUPLES_OK)
+		ereport(ERROR,
+				(errcode(ERRCODE_CONNECTION_EXCEPTION), errmsg("%s", PQresultErrorMessage(res))));
+
+	/* Drop replication slot on the source data node only if it exists */
+	if (PQntuples(res) != 0)
```
pmwkaa, line 457: I believe it should be impossible to have slots with the same name. I am thinking about the situation, when drop slot executed correctly inside `BEGIN/COMMIT` block but it still shows in that view, because the slot drop procedure is async, if that makes sense

tsl/src/chunk_copy.c
```diff
@@ -547,33 +698,66 @@ chunk_copy_stage_delete_chunk(ChunkCopy *cc)
...
+	{ CCS_DROP_PUBLICATION, chunk_copy_stage_drop_publication, NULL },
 
-	/* Attach chunk to the hypertable on the dst_node */
-	{ CCS_ATTACH_CHUNK, chunk_copy_stage_attach_chunk },
+	/*
+	 * Attach chunk to the hypertable on the dst_node.
+	 * The operation has succeeded from the destination data node perspective.
+	 * No cleanup required here.
+	 */
+	{ CCS_ATTACH_CHUNK, chunk_copy_stage_attach_chunk, NULL },
```
pmwkaa, line 747: We don't need clean for it, because it is being run in the transaction?

tsl/src/chunk_copy.c
```diff
@@ -547,33 +698,66 @@ chunk_copy_stage_delete_chunk(ChunkCopy *cc)
 }
 
 static const ChunkCopyStage chunk_copy_stages[] = {
+	/* Initial Marker */
+	{ CCS_INIT, NULL, chunk_copy_stage_init_cleanup },
```
pmwkaa, line 702: Curious if we want to have different stage here instead of init, basically one that inserts first activity record: `chunk_copy_activity_insert` to make it orthogonal with the cleanup function.

I think this would require to run this in separate transaction as well and complete current transaction in the `chunk_copy_init()` etc. This will allow us to remove the START also.

Another approach is to treat this as a special case and not a stage, since we update the activity table in init function, which is not a stage. This would simplify stuff a bit and the cleanup function logic as well

tsl/src/chunk_copy.c
```diff
@@ -625,6 +811,197 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+chunk_copy_activity_tuple_found(TupleInfo *ti, void *const data)
+{
+	ChunkCopy **cc = data;
+
+	*cc = STRUCT_FROM_SLOT(ti->slot, ti->mctx, ChunkCopy, FormData_chunk_copy_activity);
+	return SCAN_CONTINUE;
+}
+
+static ChunkCopy *
+chunk_copy_activity_get(int32 activity_id, const char *operation_id)
```
pmwkaa, line 828: I think we should simplify it a bit and just use the `activity_id`

tsl/src/chunk_copy.c
```diff
@@ -625,6 +811,197 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+	 * ChunkCopy structure
+	 */
+	if (cc)
+	{
+		cc->mcxt = mcxt;
+		cc->chunk = ts_chunk_get_by_id(cc->fd.chunk_id, true);
+		cc->stage = NULL;
+
+		/* Was the chunk compressed in the meanwhile? */
+		if (ts_chunk_is_compressed(cc->chunk))
```
pmwkaa, line 884: Interesting point, I think this one could require testing. Not sure if this should be done here. Also we might want to hold some kind of lock/status on the chunk to avoid it maybe

tsl/src/chunk_copy.c
```diff
@@ -625,6 +811,197 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+		stage_idx++;
+		if (namestrcmp(&cc->fd.completed_stage, stage->name) == 0)
+		{
+			found = true;
+			break;
+		}
+	}
+
+	/* should always find an entry, add ereport to quell compiler warning */
+	Assert(found == true);
```
pmwkaa, line 976: This can be triggered manually by updating the complete_stage in the table, right?

tsl/test/t/002_chunk_copy_move.pl
```diff
@@ -28,33 +28,59 @@
...
+
+	#The earlier debug error point gets released automatically since it's a session lock
+	#Call the cleanup procedure to make things right
+	$operation_id = "ts_copy_" . $curr_index . "_1";
+	$an->safe_psql('postgres',
+		"CALL timescaledb_experimental.copy_chunk_cleanup(operation_id=>'$operation_id');"
+	);
+
+	#Check chunk state is as before the move
+	check_pre_move_chunk_states();
```
pmwkaa, line 80: nit: just a thought that this could be tested in the same way as we run stages. Basically add a separate function which check the state after cleanup and iterate over them instead

pmwkaa (commented):
Wondering if we should do the cleanup automatically in case of error. Is there are serious drawbacks for doing that?

nikkhils:
> Wondering if we should do the cleanup automatically in case of error. Is there are serious drawbacks for doing that?

We could do that when we implement failure tolerant cleanup. Maybe the error occurred in the first place because one of the involved nodes is down right now and is experiencing network related issues and so running cleanup immediately might fail.

sql/ddl_experimental.sql
```diff
@@ -37,3 +37,8 @@ CREATE OR REPLACE PROCEDURE timescaledb_experimental.copy_chunk(
     source_node NAME = NULL,
     destination_node NAME = NULL)
 AS '@MODULE_PATHNAME@', 'ts_copy_chunk_proc' LANGUAGE C;
+
+CREATE OR REPLACE PROCEDURE timescaledb_experimental.copy_chunk_cleanup(
+    activity_id INT = NULL,
+    operation_id Name = NULL)
```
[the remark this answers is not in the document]
nikkhils: Yeah, the default for the other is NULL, so yeah, they can use either. We can discuss the API signature in a standup.

tsl/src/chunk_copy.c
nikkhils: We are not trying to speed things up here. We might add a AN controlled 1PC transaction if only one datanode is involved in the future. So, I would like to keep things controlled via AN as much as possible.

tsl/src/chunk_copy.c
nikkhils: The cleanup typically as of now is done much later. So if the replication slot is around, it's around and not due to async cleanup.

tsl/src/chunk_copy.c
nikkhils: The cleanup will actually happen when the "DROP TABLE" and metadata cleanup happens in other cleanup function.

tsl/src/chunk_copy.c
nikkhils: Goes back to my API signature comment. If we decide on one then we can simplify this.

tsl/src/chunk_copy.c
nikkhils: Locking will be looked at separately. But the thing is, we started a copy/move which failed. Then compression happened and then much later cleanup was called.

tsl/src/chunk_copy.c
nikkhils: Yeah, adding a bogus stage will trigger this indeed.

tsl/test/t/002_chunk_copy_move.pl
nikkhils: Didn't get this.

tsl/src/chunk_copy.c
```diff
@@ -625,6 +811,197 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+		cc->stage = NULL;
+
+		/* Was the chunk compressed in the meanwhile? */
+		if (ts_chunk_is_compressed(cc->chunk))
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("\"%s\" is a compressed remote chunk. Chunk copy/move cleanup not "
+							"supported"
+							" currently on compressed chunks",
+							get_rel_name(cc->chunk->table_id))));
```
erimatnor, line 890: Should not have several sentences in main error message according to the error style guide. I suggest moving additional information to an `errdetail` or `errhint`. For example:

```suggestion
			ereport(ERROR,
					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
					 errmsg("copy chunk cleanup not supported on compressed chunks"),
					 errhint("Decompress chunk \"%s\" before cleaning up the copy chunk operation.",
							get_rel_name(cc->chunk->table_id))));
```

tsl/src/chunk_copy.c
erimatnor: I am also concerned about this case. What if someone compresses a chunk after a copy operation has failed and we still have a registered (unfinished, but not running) activity for it? Will we end up with this error then and how will someone be able to cleanup?

Also, is there any technical reason we cannot cleanup the copy activity if the chunk is compressed post failure of the activity? I assume we should still be able to cleanup the destination node since the metadata there was created prior to compression of the chunk. So, do we even need a check and error if the chunk is compressed?

Should we add a check in `compress_chunk` that there is no ongoing (and potentially failed) copy activity for the chunk before compressing?

I can imagine adding an error to `compress_chunk` that says: "cannot compress chunk due to simultaneous copy activity" or similar.

tsl/src/chunk_copy.c
nikkhils: Ok, we are in agreement to only use `activity_id`, removed the other argument.

sql/ddl_experimental.sql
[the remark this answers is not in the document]
nikkhils: Ok, name the API `cleanup_copy_chunk_operation` and remove `operation_id` argument.

tsl/src/chunk_copy.c
nikkhils: Ok, yeah. This makes sense. On further reflection, even if the source chunk gets compressed, it should be ok to cleanup the chunk on the destination node. Removing this check.

sql/ddl_experimental.sql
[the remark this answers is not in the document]
nikkhils: Done.

tsl/src/reorder.c
```diff
@@ -244,6 +244,28 @@ tsl_copy_chunk_proc(PG_FUNCTION_ARGS)
 	PG_RETURN_VOID();
 }
 
+Datum
+tsl_copy_chunk_cleanup_proc(PG_FUNCTION_ARGS)
+{
+	Oid activity_id = PG_ARGISNULL(0) ? -1 : PG_GETARG_INT32(0);
```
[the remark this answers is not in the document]
nikkhils: Not needed anymore

tsl/src/chunk_copy.c
```diff
@@ -625,6 +811,197 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+			break;
+		}
+	}
+
+	/* should always find an entry, add ereport to quell compiler warning */
+	Assert(found == true);
+	if (!found)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("Entry for stage '%s' not found", NameStr(cc->fd.completed_stage))));
```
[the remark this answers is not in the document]
nikkhils: Done.

tsl/src/chunk_copy.c
nikkhils: Removed this error message now as per earlier discussions.

tsl/test/t/002_chunk_copy_move.pl
pmwkaa: I thought it makes sense to test each stage separately, but seems to be it will be complicated. I guess we need only to test that it properly transition to the correct stage after cleanup, which is there.

tsl/src/chunk_copy.c
```diff
@@ -624,13 +817,185 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
 		ErrorData *edata;
 		MemoryContextSwitchTo(oldcontext);
 		edata = CopyErrorData();
-		edata->hint =
-			psprintf("chunk copy operation id: %d (%s).", cc.fd.id, NameStr(cc.fd.operation_id));
+		edata->hint = psprintf("chunk copy operation id: (%s).", NameStr(cc.fd.operation_id));
```
pmwkaa, line 820: nit: I guess we can remove `()` as well or use quotes

pmwkaa (approved):
Looks good

sql/ddl_experimental.sql
```diff
@@ -37,3 +37,12 @@ CREATE OR REPLACE PROCEDURE timescaledb_experimental.copy_chunk(
     source_node NAME = NULL,
     destination_node NAME = NULL)
 AS '@MODULE_PATHNAME@', 'ts_copy_chunk_proc' LANGUAGE C;
+
+-- A copy_chunk or move_chunk procedure call involves multiple nodes and
+-- depending on the data size can take a long time. Failures are possible
+-- when this long running activity is ongoing. We need to be able to recover
+-- and cleanup such failed chunk copy/move activities and it's done via this
+-- procedure
+CREATE OR REPLACE PROCEDURE timescaledb_experimental.cleanup_copy_chunk_activity(
```
erimatnor, line 46: I think we should be consistent about terminology, e.g., "activity" vs "operation" in this case. It makes more sense, I think, to name the function `cleanup_copy_chunk_operation` when the parameter name is `operation_id`. Otherwise it is more difficult to make the mental connection.

```suggestion
CREATE OR REPLACE PROCEDURE timescaledb_experimental.cleanup_copy_chunk_operation(
```

tsl/src/chunk_copy.c
erimatnor, line 820: Nit: Hints start with capitol letter. Also, should this perhaps be a detailed message rather than a hint? According to error style guide: "Use a hint message for suggestions about what to do to fix the problem, ..." and I don't think this message qualifies.

```suggestion
		edata->hint = psprintf("Chunk copy operation id: %s.", NameStr(cc.fd.operation_id));
```

tsl/src/chunk_copy.c
```diff
@@ -624,13 +817,185 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+					Anum_chunk_copy_activity_idx_operation_id,
+					BTEqualStrategyNumber,
+					F_NAMEEQ,
+					DirectFunctionCall1(namein, CStringGetDatum(operation_id)));
+		indexid = CHUNK_COPY_ACTIVITY_PKEY_IDX;
+	}
+	else
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid chunk copy activity identifier")));
```
erimatnor, line 864:
```suggestion
				 errmsg("invalid chunk copy operation identifier")));
```

tsl/src/chunk_copy.c
```diff
@@ -625,6 +811,197 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("function must be run on the access node only")));
+
+	cc = chunk_copy_activity_get(activity_id, operation_id);
+
+	if (cc == NULL)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid chunk copy activity identifier. Entry not found")));
```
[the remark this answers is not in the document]
erimatnor: Still think this needs fixing, especially since we now only have the operation_id

tsl/src/chunk_copy.c
```diff
@@ -624,13 +817,185 @@ chunk_copy(Oid chunk_relid, const char *src_node, const char *dst_node, bool del
...
+	{
+		chunk_copy_cleanup_internal(cc, stage_idx);
+	}
+	PG_CATCH();
+	{
+		/* Include chunk copy id to the error message */
+		ErrorData *edata;
+		MemoryContextSwitchTo(oldcontext);
+		edata = CopyErrorData();
+		edata->hint = psprintf("While cleaning up chunk copy operation id: (%s).",
```
erimatnor, line 992: This should probably also be an errdetail, rather than hint, since it is not suggesting how to fix a problem.

erimatnor (approved):
Approving, but have a number of nits and suggestions for improvement that I think should be fixed before merging.

In particular, I think we need to be consistent about use of "operation" vs "activity". The code, APIs and error message are a bit confused about this and use mixed terminology (sometimes referring to operation, and other times to activity). We should stick to only operation if we give an `operation_id`.

Also, I think we should use errdetails  in some places where the PR now uses hints. Hints should be a suggestion how to fix the problem/error. 

Otherwise, LGTM.

sql/ddl_experimental.sql
nikkhils: Ok, done.

tsl/src/chunk_copy.c
nikkhils: Done.

tsl/src/chunk_copy.c
nikkhils: Done.

tsl/src/chunk_copy.c
nikkhils: done

PR #3407, opened 2021-07-12
Ignore debug_notice test during sanitizer runs

afiskon:
This test sometimes fails because it generates extra "rehashing
catalog cache" debug messages, which is not very helpful.


mfundul:
Is this only a problem for the sanitizer tests? I would expect this to be flaky in general.

mfundul (approved):

afiskon:
@mfundul I disabled in in AppVeyor not a long time ago for the same reason.

mkindahl (approved):
LGTM

PR #3412, opened 2021-07-15
Add perl module files to PG install

nikkhils:
We have added additional functionality in timescaledb extension to
use in tap tests. Install these perl files in the PG installation
directory so that external modules (the current "forge_ext" extension
as an example) can use this functionality without having to have an
additional dependency on the timescaledb extension source code. Note
that these perl files should be installed only if PG installation has
the relevant "${PG_PKGLIBDIR}/pgxs/src/test/perl" subdirectory.

Also rejig the configuration directory variable that's used for the
tap tests so that external modules can specify their own locations by
using their own values for it (the current variable was tightly tied to
timescaledb source code location).

### DIFF
--- a/test/perl/CMakeLists.txt
+++ b/test/perl/CMakeLists.txt
@@ -1,0 +1,6 @@
+set(PERL_FILES AccessNode.pm DataNode.pm TimescaleNode.pm)
+
+# Check if PostgreSQL was compiled with --enable-tap-tests
+if(EXISTS "${PG_PKGLIBDIR}/pgxs/src/test/perl")
+  install(FILES ${PERL_FILES} DESTINATION "${PG_PKGLIBDIR}/pgxs/src/test/perl")
+endif()
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -125,3 +125,5 @@ endif(PG_SOURCE_DIR)
 if(CMAKE_BUILD_TYPE MATCHES Debug)
   add_subdirectory(src)
 endif(CMAKE_BUILD_TYPE MATCHES Debug)
+
+add_subdirectory(perl)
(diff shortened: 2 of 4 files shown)


pmwkaa:
I think its a bit odd to install PostgreSQL files from one extension and assume they exists in another. What if PostgreSQL version changes? For example, if we have files installed for one version, but current PG version differs and some functionality changed.

Maybe it makes sense to properly integrate $PG_SOURCE_DIR into the `forge_ext` instead? Basically do proper support to find PostgreSQL sources in forge extension as we do for timescaledb.

nikkhils:
> I think its a bit odd to install PostgreSQL files from one extension and assume they exists in another. What if PostgreSQL version changes? For example, if we have files installed for one version, but current PG version differs and some functionality changed.
> 

TS sources were compiled with this same PG version, right? So the install will go into the corresponding PG version only. 

> Maybe it makes sense to properly integrate $PG_SOURCE_DIR into the `forge_ext` instead? Basically do proper support to find PostgreSQL sources in the forge extension as we do for timescaledb, if makes sense

We only need the `pm` files from the timescale extension. One option is to add an environment variable pointing to the timescaledb extension sources in `forget_ext` and then pass it as an additional include location while invoking `prove` from the forge extension.

test/perl/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(PERL_FILES AccessNode.pm DataNode.pm TimescaleNode.pm)
+
+foreach(P_FILE ${PERL_FILES})
+  configure_file(${P_FILE} ${CMAKE_CURRENT_BINARY_DIR}/${P_FILE} COPYONLY)
```
erimatnor, line 4: Why do we need to copy the files to the binary (build) dir?

nikkhils: Removed. This is indeed unwanted.

erimatnor (approved):

pmwkaa (approved):

PR #3489, opened 2021-08-16
Fix continuous agg refresh procedure failure with 12.8

gayyappan:


mkindahl (approved):

afiskon:
Looks OK but don't we have to add a test on this?

afiskon (approved):

PR #3491, opened 2021-08-16
Bump postgres versions used in CI to latest

svenklemm:


mkindahl (commented):
Wouldn't it make more sense to bump this in the PR where we actually fix the bug that currently causes 12.8 failures? Ideally, we would like to have a controlled update so that we do not bump the version until we are ready to support and can get the CI to pass: otherwise we will be blocked like this every time we have a minor version update.

mkindahl (approved):

gayyappan (approved):

PR #3497, opened 2021-08-18
Test Apache-only packages in CI

svenklemm:
Disable-check: commit-count


nikkhils (approved):
LGTM!

erimatnor (approved):
LGTM, but it would be nice with some commit log messages and then I guess you need to think about how to handle the multiple commits.

PR #3509, opened 2021-08-23
Push cagg invalidation trigger to datanodes

mfundul:


mfundul:
After inserting it crashes with:
image

tsl/src/continuous_aggs/create.c
```diff
@@ -368,6 +370,39 @@ cagg_add_trigger_hypertable(Oid relid, char *trigarg)
...
+
+		int i = 0;
+		cmds = palloc(list_length(data_node_list) * sizeof(*cmds));
+		foreach (cell, ht->data_nodes)
+		{
+			HypertableDataNode *node = lfirst(cell);
+			StringInfo command = makeStringInfo();
+			appendStringInfo(command,
+							"CREATE TRIGGER %s AFTER INSERT OR UPDATE OR DELETE ON %s.%s FOR EACH ROW EXECUTE FUNCTION %s.%s(%d)",
+							CAGGINVAL_TRIGGER_NAME,
```
mkindahl, line 388: Probably good to quote this as well so that it work even with strange trigger names (not likely to happen, just a precaution).
```suggestion
							quote_identifier(CAGGINVAL_TRIGGER_NAME),
```

tsl/src/continuous_aggs/create.c
```diff
@@ -368,6 +370,39 @@ cagg_add_trigger_hypertable(Oid relid, char *trigarg)
...
+							quote_identifier(NameStr(ht->fd.schema_name)),
+							quote_identifier(NameStr(ht->fd.table_name)),
+							quote_identifier(INTERNAL_SCHEMA_NAME),
+							quote_identifier(CAGG_INVALIDATION_TRIGGER),
+							node->fd.node_hypertable_id);
+			cmds[i++] = command->data;
+			elog(INFO, "[%s] hypertable_id=%d node_hypertable_id=%d", node->fd.node_name.data, node->fd.hypertable_id, node->fd.node_hypertable_id);
```
mkindahl, line 395: Maybe use `DEBUG2` instead? Will both allow us to control if it should be printed and also allow us to find and remove debug messages going forward.
```suggestion
			elog(DEBUG2, "[%s] hypertable_id=%d node_hypertable_id=%d", node->fd.node_name.data, node->fd.hypertable_id, node->fd.node_hypertable_id);
```

tsl/src/continuous_aggs/create.c
```diff
@@ -368,6 +370,39 @@ cagg_add_trigger_hypertable(Oid relid, char *trigarg)
...
+		}
+
+		result = ts_dist_multi_cmds_params_invoke_on_data_nodes(cmds, NULL, data_node_list, true,
+																true);
+		if (result)
+			ts_dist_cmd_close_response(result);
+
+		ts_cache_release(hcache);
+		return;
+	}
```
mkindahl, line 405: Nit: would suggest to use an else branch here instead and use the `ts_cache_release(hcache)` that is already there.

tsl/src/remote/dist_commands.c
```diff
@@ -74,18 +74,21 @@ ts_dist_cmd_collect_responses(List *requests)
...
+ * an explicit 2PC from the access node.
+ * 
+ * If "multiple_cmds" is false then "sql" and "params" are not iterated over.
  */
 DistCmdResult *
-ts_dist_cmd_params_invoke_on_data_nodes(const char *sql, StmtParams *params, List *data_nodes,
-										bool transactional)
+ts_dist_multi_cmds_params_invoke_on_data_nodes(const char **sql, StmtParams **params,
+											   List *data_nodes, bool multiple_cmds,
+											   bool transactional)
```
mkindahl, line 91: I think it might be better to make this a little more generic, but we can take that in a separate commit.

mkindahl (commented):
Some minor things, but otherwise looks good.

PR #3519, opened 2021-08-25
Run CI against PG snapshots

svenklemm:
Run CI against PG snapshots to discover potential issues with
upcoming PG releases earlier.

### DIFF
--- a/.github/workflows/linux-build-and-test.yaml
+++ b/.github/workflows/linux-build-and-test.yaml
@@ -19,7 +19,7 @@ jobs:
       run: python scripts/gh_matrix_builder.py ${{ github.event_name }}
 
   regress:
-    name: PG${{ matrix.pg }} ${{ matrix.name }} ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


mkindahl (approved):

gayyappan (approved):

.github/workflows/linux-build-and-test.yaml
```diff
@@ -66,7 +66,11 @@ jobs:
     - name: Build PostgreSQL ${{ matrix.pg }} ${{ matrix.build_type }}
```
mfundul, line 66: Shouldn't this be `Build PostgreSQL ${{ matrix.pg }} ${{matrix.snapshot}} ${{ matrix.build_type }}`?

And more importantly, shouldn't the cache above it include the `snapshot` otherwise it will conflict on the same key?

svenklemm: The key won't conflict since matrix.pg is either 12 or 13 for snapshots while it's exact version for non-snapshots e.g. 13.4
But i should probably disable cache entirely for snapshots.

PR #3542, opened 2021-09-01
time_bucket_ng() version with `origin` argument

afiskon:
This patch adds a version of time_bucket_ng() with 'origin' argument.

It doesn't address any other known issues. E.g. volatility of the function will be changed
in another patch. The error messages are going to be improved when the feature gets
 a little more stable.

### DIFF
--- a/src/time_bucket.c
+++ b/src/time_bucket.c
@@ -415,7 +415,9 @@ ts_time_bucket_ng_date(PG_FUNCTION_ARGS)
 	{
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-				 errmsg("origin must be the first day of the month")));
+				 errmsg("origin must be the first day of the month"),
+				 errhint("When using timestamptz-version of the function, 'origin' is "
+						 "converted to provided 'timezone'.")));
 	}
 
 	if (DATE_NOT_FINITE(date))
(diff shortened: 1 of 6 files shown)


test/expected/timestamp.out
```diff
@@ -1243,6 +1243,9 @@ ERROR:  interval can't combine months with minutes or hours
 -- timestamp is less than the default 'origin' value
 SELECT timescaledb_experimental.time_bucket_ng('1 day', '1999-01-01 12:34:56 MSK' :: timestamptz, timezone => 'MSK');
 ERROR:  origin must be before the given date
+-- 'origin' is not a first day of the month after converting from provided time zone to UTC
+select timescaledb_experimental.time_bucket_ng('1 month', '2021-07-12 12:34:56 Europe/Moscow' :: timestamptz, origin => '2021-06-01 00:00:00 Europe/Moscow' :: timestamptz, timezone => 'UTC');
+ERROR:  origin must be the first day of the month
```
gayyappan, line 1248: It would be helpful to add a hint explaining why it fails.
Something along the lines of:
errhint('origin must be first day of the month after applying timezone conversions specified by `timezone` argument).

The error messages usually specify the argument value that causes the failure.
e.g. :  elog(ERROR, "invalid time_bucket type \"%s\"", format_type_be(timestamp_type));

So for this error, I would expect something like:
`origin '2021-06-01 00:00:00 Europe/Moscow' must be the first day of the month.`
along with a hint why it is not so.

We have a similar issue with the error message: `origin must be before given date`. It doesn't give additional info on what origin or date is being referred to.

afiskon: Thanks for the feedback, Gayathri! I did what I could without refactoring the rest of the code (and the tests) too much. I suggest later we invest some time into reviewing and improving all the error messages of time_bucket_ng(). There was not too much effort put into them at this point. I agree that this case was especially confusing.

src/time_bucket.c
```diff
@@ -415,7 +415,12 @@ ts_time_bucket_ng_date(PG_FUNCTION_ARGS)
 	{
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
-				 errmsg("origin must be the first day of the month")));
+				 errmsg("origin \"%04d-%02d-%02d\" must be the first day of the month",
```
erimatnor, line 418: I would recommend against displaying a date in this hard-coded way. Different regions of the world use different format to display year, month, and day, so this can be confusing. US, for instance, is mm-dd-yyyy.

src/time_bucket.c
```diff
@@ -415,7 +415,12 @@ ts_time_bucket_ng_date(PG_FUNCTION_ARGS)
...
+						origin_year,
+						origin_month,
+						origin_day),
+				 errhint("when using timestamptz-version of time_bucket_ng(), 'origin' is "
+						 "converted to provided 'timezone'")));
```
erimatnor, line 423: A hint should start capitalized and end with a period. See PostgreSQL error style guide. There it might be worth also looking at how and when to use quotes. I would also avoid using the function name in the message as it is self-evident which function this concerns.

```suggestion
				 errhint("When using timestamptz-version of time_bucket_ng(), 'origin' is "
						 "converted to provided 'timezone'".)));
```

src/time_bucket.c
```diff
@@ -487,3 +492,32 @@ ts_time_bucket_ng_timezone(PG_FUNCTION_ARGS)
...
+TS_FUNCTION_INFO_V1(ts_time_bucket_ng_timezone_origin);
+TSDLLEXPORT Datum
+ts_time_bucket_ng_timezone_origin(PG_FUNCTION_ARGS)
+{
+	Timestamp result;
+	Datum timestamp, origin;
+	Datum interval = PG_GETARG_DATUM(0);
+	Datum timestamptz = PG_GETARG_DATUM(1);
+	Datum origintz = PG_GETARG_DATUM(2);
+	Datum tzname = PG_GETARG_DATUM(3);
```
mfundul, line 505: Shouldn't these be of the corresponding datatype instead of `Datum`?

mfundul: e.g. `PG_GETARG_TIMESTAMPTZ`

afiskon: Here there is no point converting them to TImestampTz and back to Datum. We never access these arguments as TimestampTz. We only pass them to other functions, which accept Datum when using C API.

src/time_bucket.c
afiskon: OK, then I remove the date from the message for now. @gayyappan @erimatnor  If you don't mind, I would prefer not to dive into localizing error messages at this stage of the game. I'm going to refactor the error messages when the feature will be more stable.

src/time_bucket.c
afiskon: Good point, thanks!

src/time_bucket.c
afiskon: I added a corresponding note to the commit message.

mfundul (approved):

gayyappan (approved):

PR #3574, opened 2021-09-10
Fix failure on job execution by background worker

fabriziomello:
If the registered procedure has COMMIT inside the code, the execution
of the job by the background worker was failing.

Fixed it checking if there are a PortalContext defined, if no then
create a Portal from scratch and execute the code using PortalRun, and
if there are a PortalContext then use the existing job execution code.

### DIFF
--- a/tsl/src/bgw_policy/job.c
+++ b/tsl/src/bgw_policy/job.c
@@ -16,8 +16,10 @@
 #include <nodes/primnodes.h>
 #include <parser/parse_func.h>
 #include <parser/parser.h>
+#include <tcop/pquery.h>
 #include <utils/builtins.h>
 #include <utils/lsyscache.h>
+#include <utils/portal.h>
 #include <utils/syscache.h>
 #include <utils/snapmgr.h>
 #include <utils/timestamp.h>
@@ -32,6 +34,7 @@
 #include "bgw_policy/policy_utils.h"
 #include "bgw_policy/reorder_api.h"
 #include "bgw_policy/retention_api.h"
+#include "compat/compat.h"
 #include "compression/compress_utils.h"
 #include "continuous_aggs/materialize.h"
 #include "continuous_aggs/refresh.h"
@@ -752,25 +755,30 @@ bool
 job_execute(BgwJob *job)
 {
 	Const *arg1, *arg2;
-	bool transaction_started = false;
-	bool pushed_snapshot = false;
+	bool portal_created = false;
 	char prokind;
 	Oid proc;
 	ObjectWithArgs *object;
 	FuncExpr *funcexpr;
 	MemoryContext parent_ctx = CurrentMemoryContext;
 	StringInfo query;
-	if (!IsTransactionOrTransactionBlock())
-	{
-		transaction_started = true;
-		StartTransactionCommand();
-	}
+	Portal portal = ActivePortal;
 
-	/* executing sql functions requires snapshot. */
-	if (!ActiveSnapshotSet())
+	/* Create a portal if there's no active */
+	if (!PortalIsValid(portal))
 	{
-		pushed_snapshot = true;
+		portal_created = true;
+		portal = CreatePortal("", true, true);
+		portal->visible = false;
+		portal->resowner = CurrentResourceOwner;
+		ActivePortal = portal;
+
+		StartTransactionCommand();
+#if (PG12 && PG_VERSION_NUM >= 120008) || (PG13 && PG_VERSION_NUM >= 130004) || PG14_GE
+		EnsurePortalSnapshotExists();
+#else
 		PushActiveSnapshot(GetTransactionSnapshot());
+#endif
 	}
 
 	object = makeNode(ObjectWithArgs);
@@ -829,16 +837,15 @@ job_execute(BgwJob *job)
 			break;
 	}
 
-	/* Both checks are needed: if the executed procedure commit the
-	 * transaction---which `continuous_agg_refresh_internal` does, for
-	 * example---it will remove the active snapshot and start a new
-	 * transaction with no active snapshots. In that case, we should not pop a
-	 * snapshot. */
-	if (pushed_snapshot && ActiveSnapshotSet())
-		PopActiveSnapshot();
-
-	if (transaction_started)
+	/* Drop portal if it was created */
+	if (portal_created)
+	{
+		if (ActiveSnapshotSet())
+			PopActiveSnapshot();
 		CommitTransactionCommand();
+		PortalDrop(portal, false);
+		ActivePortal = NULL;
+	}
 
 	return true;
 }
--- a/tsl/test/expected/bgw_custom.out
+++ b/tsl/test/expected/bgw_custom.out
@@ -245,7 +245,7 @@ SELECT add_job( proc=>'custom_func',
     1005
 (1 row)
 
-SELECT job_id, next_start, scheduled, schedule_interval 
+SELECT job_id, next_start, scheduled, schedule_interval
 FROM timescaledb_information.jobs WHERE job_id > 1000;
  job_id |          next_start          | scheduled | schedule_interval 
 --------+------------------------------+-----------+-------------------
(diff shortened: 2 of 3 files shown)


tsl/src/bgw_policy/job.c
```diff
@@ -839,6 +815,233 @@ job_execute(BgwJob *job)
 
 	if (transaction_started)
 		CommitTransactionCommand();
+}
+
+/*
+ * This code is based on:
```
gayyappan, line 821: Would be good to outline the steps involved here and why it is needed.

tsl/src/bgw_policy/job.c
```diff
@@ -839,6 +815,233 @@ job_execute(BgwJob *job)
...
+										 ALLOCSET_DEFAULT_MAXSIZE);
+	oldcontext = MemoryContextSwitchTo(parsecontext);
+	raw_parsetree_list = pg_parse_query(query);
+	commands_remaining = list_length(raw_parsetree_list);
+	isTopLevel = commands_remaining == 1;
+	MemoryContextSwitchTo(oldcontext);
+
+	/*
+	 * Do parse analysis, rule rewrite, planning, and execution for each raw
+	 * parsetree.  We must fully execute each query before beginning parse
```
gayyappan, line 872: This is a CALL or SELECT fncall(), we should have exactly 1 parsetree?  (We do not support generic sql strings) . Would be better to remove the loop and add an assert instead?

tsl/src/bgw_policy/job.c
```diff
@@ -839,6 +815,233 @@ job_execute(BgwJob *job)
...
+		Portal portal;
+		DestReceiver *receiver;
+		int16 format = 1;
+
+		/*
+		 * We don't allow transaction-control commands like COMMIT and ABORT
+		 * here.  The entire SQL statement is executed as a single transaction
+		 * which commits if no errors are encountered.
+		 */
+		if (IsA(parsetree, TransactionStmt))
```
gayyappan, line 899: The comment is a bit misleading since this can only be a Call or Select  Stmt?

tsl/src/bgw_policy/job.c
```diff
@@ -839,6 +815,233 @@ job_execute(BgwJob *job)
...
+		 * destination.
+		 */
+		commandTag = CreateCommandTag(parsetree->stmt);
+
+		BeginCommand(commandTag, DestNone);
+
+		/* Set up a snapshot if parse analysis/planning will need one. */
+		if (analyze_requires_snapshot(parsetree))
+		{
+			PushActiveSnapshot(GetTransactionSnapshot());
```
gayyappan, line 917: need a test case for this.

svenklemm (commented):
Hmm going through sql generation -> parsing allows for the introduction of subtle security bugs. Can we avoid this and call the respective job function directly?

tsl/src/bgw_policy/job.c
fabriziomello: Sure

tsl/src/bgw_policy/job.c
fabriziomello: Fixed

tsl/src/bgw_policy/job.c
fabriziomello: Removed.

tsl/src/bgw_policy/job.c
fabriziomello: You're correct... but first I'm trying a way to avoid the parser ...

fabriziomello:
> Hmm going through sql generation -> parsing allows for the introduction of subtle security bugs. Can we avoid this and call the respective job function directly?

You're correct... But I'm not sure (yet) if we can do it by creating a Portal from scratch... still trying to figure out if there are a way to avoid the parser... Anyway according Tom Lane [1] we should create Portal to execute SQL properly from background workers.

[1] https://www.postgresql.org/message-id/477877.1631561419%40sss.pgh.pa.us

tsl/src/bgw_policy/job.c
```diff
@@ -770,7 +785,11 @@ job_execute(BgwJob *job)
 	if (!ActiveSnapshotSet())
 	{
 		pushed_snapshot = true;
+#if (PG_VERSION_NUM >= 120800) || (PG_VERSION_NUM >= 130400)
```
svenklemm, line 788:
```suggestion
#if (PG12 && PG_VERSION_NUM >= 120800) || (PG13 && PG_VERSION_NUM >= 130400) || PG14_GE
```
Have to include compat/compath.h for PGXX macros to work

fabriziomello: Fixed, thanks!

tsl/src/bgw_policy/job.c
```diff
@@ -770,7 +786,11 @@ job_execute(BgwJob *job)
 	if (!ActiveSnapshotSet())
 	{
 		pushed_snapshot = true;
+#if (PG12 && PG_VERSION_NUM >= 120800) || (PG13 && PG_VERSION_NUM >= 130400) || PG14_GE
```
svenklemm, line 789:
```suggestion
#if (PG12 && PG_VERSION_NUM >= 120008) || (PG13 && PG_VERSION_NUM >= 130004) || PG14_GE
```
Middle bits of numeric version are not used past 9.6
```
dev=# select current_setting('server_version_num'),version();
 current_setting |                                   version
-----------------+------------------------------------------------------------------------------
 130003          | PostgreSQL 13.3 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 11.1.0, 64-bit
(1 row)
```

tsl/src/bgw_policy/job.c
```diff
@@ -840,5 +860,9 @@ job_execute(BgwJob *job)
 	if (transaction_started)
 		CommitTransactionCommand();
 
+	/* Drop portal if it was created */
+	if (portal_created)
+		PortalDrop(portal, false);
```
afiskon, line 865: Leaving ActivePortal pointing to pfree'd memory doesn't strike me as a great idea.

afiskon (changes requested):

tsl/src/bgw_policy/job.c
afiskon: Another question is what will happen if an exception will be thrown between creating a portal and this line.

tsl/src/bgw_policy/job.c
fabriziomello: Ooops ... my bad ... fixed!

tsl/src/bgw_policy/job.c
fabriziomello: > Leaving ActivePortal pointing to pfree'd memory doesn't strike me as a great idea.

You're correct... fixed!

fabriziomello: > Another question is what will happen if an exception will be thrown between creating a portal and this line.

This is responsibility of `PostgresMain` by calling `PortalErrorCleanup` no? https://github.com/postgres/postgres/blob/master/src/backend/tcop/postgres.c#L4245

svenklemm (approved):

tsl/test/expected/bgw_custom.out
```diff
@@ -269,3 +269,121 @@ total_successes        | 0
...
+-- Check results
+SELECT * FROM custom_log ORDER BY job_id, extra;
+ job_id |         args          |            extra            |      runner       
+--------+-----------------------+-----------------------------+-------------------
+   1000 | {"type": "procedure"} | custom_proc 1 COMMIT        | default_perm_user
+   1000 | {"type": "procedure"} | custom_proc 3 COMMIT        | default_perm_user
+   1001 | {"type": "procedure"} | custom_proc_nested 1 COMMIT | default_perm_user
+   1001 | {"type": "procedure"} | custom_proc_nested 3 COMMIT | default_perm_user
+(4 rows)
+
```
gayyappan, line 363: If we delete jobs 1 and 2 and then truncate custom_log, the output will be cleaner for the next job run.
select delete_job(:job_id_1);

fabriziomello: Fixed!

mkindahl (approved):
LGTM.

afiskon (approved):

tsl/src/bgw_policy/job.c
afiskon: If you are sure that job_execute() is never called from any PG_TRY's then probably that's the case.

tsl/src/bgw_policy/job.c
```diff
@@ -770,7 +786,11 @@ job_execute(BgwJob *job)
 	if (!ActiveSnapshotSet())
 	{
 		pushed_snapshot = true;
+#if (PG12 && PG_VERSION_NUM >= 120008) || (PG13 && PG_VERSION_NUM >= 130004) || PG14_GE
```
gayyappan, line 789: This is possible only if we have to setup a Portal, right? I think we can combine if conditions at 771, 779 and 786. If not, could you explain where else we might have to start a TransactionBlock ?

fabriziomello: Definitely we can do it. Fixed!

gayyappan (approved):

PR #3575, opened 2021-09-11
Optimize approximate_row_count

svenklemm:
Rewrite approximate_row_count to SQL instead of PLpgSQL and remove
superfluous JOINs against pg_namespace. Adjust tuple calculation
for PG14 since in PG14 reltuples for partitioned tables is the sum
of it's children so we need to exclude those from calculation to
not doublecount.

### DIFF
--- a/sql/size_utils.sql
+++ b/sql/size_utils.sql
@@ -368,43 +368,30 @@ $BODY$;
 -- Estimated number of rows according to catalog tables
 CREATE OR REPLACE FUNCTION approximate_row_count(relation REGCLASS)
 RETURNS BIGINT
-LANGUAGE PLPGSQL VOLATILE STRICT AS
+LANGUAGE SQL VOLATILE STRICT AS
 $BODY$
-DECLARE
-	table_name       NAME;
-	schema_name      NAME;
-	row_count_parent BIGINT;
-	row_count        BIGINT;
-BEGIN
-	SELECT relname, nspname, CASE WHEN c.reltuples > 0 THEN c.reltuples::bigint ELSE 0 END
-	INTO table_name, schema_name, row_count_parent
-	FROM pg_class c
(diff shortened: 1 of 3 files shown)


afiskon (approved):

sql/size_utils.sql
```diff
@@ -368,43 +368,30 @@ $BODY$;
...
+    WHERE i.inhparent = relation
+    UNION
+    SELECT i.inhrelid AS oid
+    FROM pg_inherits i
+    JOIN inherited_id b ON i.inhparent = b.oid
+  ),
+  parent AS
+  (
+    -- reltuples for partitioned tables is the sum of it's children in pg14 so we need to filter those out
+    SELECT CASE WHEN c.reltuples > 0 AND c.relkind <> 'p' THEN c.reltuples::bigint ELSE 0 END AS reltuples
```
gayyappan, line 386: changes look fine. But won't this give incorrect counts for PG < PG14?

svenklemm: No, before PG14 reltuples would just be 0 for partitioned table parent.

mkindahl (approved):
LGTM. Not sure if we need the recursive CTE (see the comment), but we can deal with this later.

sql/size_utils.sql
```diff
@@ -368,43 +368,30 @@ $BODY$;
...
+  WITH RECURSIVE inherited_id AS
+  (
+    SELECT i.inhrelid AS oid
+    FROM pg_inherits i
+    WHERE i.inhparent = relation
+    UNION
+    SELECT i.inhrelid AS oid
+    FROM pg_inherits i
+    JOIN inherited_id b ON i.inhparent = b.oid
+  ),
```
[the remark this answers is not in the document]
svenklemm: This is a generic convenience functions that supports hypertables, normal tables and partitioned tables. So the recursion is needed. I even added an example with nested partitioned table as part of this patch.

PR #3587, opened 2021-09-16
Fix find_ec_member_matching_expr backport for 13.2

svenklemm:
find_ec_member_matching_expr is not available in PG13.2 so this patch
adjusts the macro that enables the function for our code.

https://github.com/postgres/postgres/commit/7bfba4f1

### DIFF
--- a/src/import/planner.c
+++ b/src/import/planner.c
@@ -38,7 +38,7 @@
 #include "compat/compat.h"
 #include "planner.h"
 
-#if PG12
+#if PG_VERSION_NUM < 130003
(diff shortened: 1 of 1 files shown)


afiskon (approved):

src/import/planner.c
```diff
@@ -38,7 +38,7 @@
 #include "compat/compat.h"
 #include "planner.h"
 
-#if PG12
+#if PG12 || (PG13 && PG_VERSION_NUM < 130003)
```
erimatnor, line 41: Nit: PG13 check seems superfluous:

```suggestion
#if PG12 || (PG_VERSION_NUM < 130003)
```

erimatnor (approved):

PR #3588, opened 2021-09-16
Backport ec member function for PG v13.2 or less

erimatnor:
A couple of planner functions related to equivalance members were
introduced in PG 13.3, but was only backported to PG 12 or less. This
meant that builds for PG 13.2 broke.

Fix this by backporting to all versions < PG 13.3.


erimatnor:
Closing in favor of https://github.com/timescale/timescaledb/pull/3587

PR #3592, opened 2021-09-17
Allow alter column type on distributed hypertable

### LINKED ISSUE #3582: Alter table alter type on distributed hypertable
 - OS: Debian 10/Buster
 - PostgreSQL version (output of `postgres --version`): postgres (PostgreSQL) 13.4 (Debian 13.4-1.pgdg100+1)
 - TimescaleDB version (output of `\dx` in `psql`): 2.4.1
 - Installation method: apt

**Describe the bug**
Attempting to change the data type of a distributed hypertable results in `ERROR:  operation not supported on distributed hypertable`

**To Reproduce**
1. `create table test (time timestamp, v int);`
2. `select create_distributed_hypertable('test','time');`
3. `alter table test alter v type bigint;`

**Expected behavior**
Success

**Actual behavior**

**Additional context**
[The documentation](https://docs.timescale.com/timescaledb/latest/how-to-guides/distributed-hypertables/alter/) indicates that ALTER is supported, and gives no indication of any limitations.

I also [raised the issue in the slack channel](https://timescaledb.slack.com/archives/CPSD82EMU/p1631621121081600), and was directed to report it here.

pmwkaa:
Unlock and support ALTER TABLE ALTER TYPE on distributed
hypertable.

Fix #3582

### DIFF
--- a/tsl/test/expected/dist_ddl.out
+++ b/tsl/test/expected/dist_ddl.out
@@ -39,9 +39,9 @@ CREATE SCHEMA disttable_schema AUTHORIZATION :ROLE_1;
 CREATE SCHEMA some_schema AUTHORIZATION :ROLE_1;
 SET ROLE :ROLE_1;
 CREATE TABLE disttable(time timestamptz, device int, color int CONSTRAINT color_check CHECK (color > 0), temp float);
-CREATE UNIQUE INDEX disttable_pk ON disttable(time);
+CREATE UNIQUE INDEX disttable_pk ON disttable(time, temp);
 -- CREATE TABLE
-SELECT * FROM create_distributed_hypertable('disttable', 'time', replication_factor => 3);
+SELECT * FROM create_distributed_hypertable('disttable', 'time', 'temp', replication_factor => 3);
 NOTICE:  adding not-null constraint to column "time"
  hypertable_id | schema_name | table_name | created 
 ---------------+-------------+------------+---------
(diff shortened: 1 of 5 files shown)


svenklemm (commented):
I think this needs a couple more tests. Any ALTER TYPE on partitioning columns should be blocked. Also we should have an example with a conversion function with USING, e.g. `alter table x alter COLUMN float_col type int USING float_col::int;`

gayyappan (commented):
We need a test with distributed hypertables that have compressed chunks: ALTER type after chunks are compressed

tsl/test/expected/dist_ddl.out
```diff
@@ -475,9 +487,19 @@ ERROR:  cannot drop a hypertable along with other objects
...
+ALTER TABLE disttable ALTER COLUMN description TYPE VARCHAR(10);
+ALTER TABLE disttable ADD COLUMN float_col float;
+ALTER TABLE disttable ALTER COLUMN float_col TYPE INT USING float_col::int;
+\set ON_ERROR_STOP 0
+-- Changing the type of a hash-partitioned column should not be supported
+ALTER TABLE disttable ALTER COLUMN temp TYPE numeric;
+ERROR:  cannot change the type of a hash-partitioned column
+\set ON_ERROR_STOP 1
+-- Should be able to change if not hash partitioned though
+ALTER TABLE disttable ALTER COLUMN time TYPE timestamp;
```
erimatnor, line 502: Are we sure it is safe to do this? For instance, changing the time type from bigint to int might change the min/max allowed value and expand or shrink the check constraints on chunks.

I think it might be safest to block changes in types of all partitioning columns. Perhaps going from int to bigint is safe, but we'd have to think through the cases.

I am actually surprised this isn't already blocked in TimescaleDB. Doesn't seem to be a multi-node specific issue.

pmwkaa: I was actually following the same test used in `test/partition.sql` for this

gayyappan (approved):
Changes look good to me.

fabriziomello (approved):
LGTM

tsl/test/expected/dist_ddl.out
svenklemm: It is not safe to do in general, it will only work for a limited subset of alter table alter type

PR #3596, opened 2021-09-17
Refactor SQL function approximate_row_count

fabriziomello:
Simplify the CTE to recursively inspect all partitions of a relation
and calculate the sum of `pg_class.reltuples` taking in account the
differences introduced by PG14.


afiskon (approved):

mkindahl (approved):
LGTM

PR #3599, opened 2021-09-20
Update connection cache on role changes

erimatnor:
The access node maintains a cache of connections to its data
nodes. Each entry in the cache is a connection for a user and remote
data node pair. Currently, a cache entry is invalidated when a foreign
server object representing a data node is changed (e.g., the port
could have been updated). The connection will remain in the cache for
the duration of the current command, but will be remade with the
updated parameters the next time it is fetched from the cache.

This change invalidates a connection cache entry if the connection's
role/user changes and drops an entry if the role is dropped. One
reason for invalidating a connection is that a role rename invalidates
the certificate the connection is using in case client certificate
authentication is used. Thus, connections that have been
authenticated with a certificate that is no longer valid will be
remade. In some cases, this extra invalidation leads to purging
connections when not strictly necessary. However, this is not a big
problem in practice since role objects don't change often.

### DIFF
--- a/tsl/test/t/003_connections.pl
+++ b/tsl/test/t/003_connections.pl
@@ -1,0 +1,68 @@
+# This file and its contents are licensed under the Timescale License.
+# Please see the included NOTICE for copyright information and
+# LICENSE-TIMESCALE for a copy of the license.
+
+# test a simple multi node cluster creation and basic operations
+use strict;
+use warnings;
+use AccessNode;
+use DataNode;
+use TestLib;
+use Test::More tests => 1;
+
+#Initialize all the multi-node instances
+my $an  = AccessNode->create('an');
+my $dn1 = DataNode->create('dn1');
+my $dn2 = DataNode->create('dn2');
+
+$an->add_data_node($dn1);
+$an->add_data_node($dn2);
+
+$dn1->append_conf('postgresql.conf', 'log_connections=true');
+$dn2->append_conf('postgresql.conf', 'log_connections=true');
+
+my $output = $an->safe_psql(
+	'postgres',
+	qq[
+	CREATE ROLE alice;
+	CALL distributed_exec('CREATE ROLE alice LOGIN');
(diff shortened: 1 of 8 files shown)


nikkhils (approved):
LGTM. For completeness, we should check for `DROP ROLE` invalidating the cache entry as well.

pmwkaa (approved):
Looks good. Agree with Nikhil that we might want to test DROP ROLE here as well, if possible

erimatnor:
Testing DROP ROLE is a good idea. I added a TAP test for it. 

It required actually adding code functionality to purge a connection cache entry entirely when the role is dropped since otherwise it will leave stale entries in the cache.

I think we might want to do the same when a data node (foreign server) is dropped. I'll leave that for a separate PR though.

tsl/test/t/003_connections.pl
```diff
@@ -0,0 +1,66 @@
...
+# * for alice.
+#
+# * Second row is the output from the connection cache after SELECT
+#   and prior to dropping alice.
+#
+# * There is no row for the third connection cache query since the
+# * connections for alice have been purged.
+is($stdout,
+q[2021-09-20 00:00:00+02|1|23.4
+dn1|alice|f]);
```
nikkhils, line 66: Shouldn't we see an additional `dn1|alice|t` entry in the middle before the final false entry?

nikkhils: Additionally TAP tests are failing. Might make sense to remove the timestamp to keep things static.

erimatnor: It is explained in the comment. 

There is one entry for the initial SELECT query, then one entry for showing the connection cache. Once alice is dropped, the entry is removed, so the last show_connection_cache shows nothing. I am not 100% satisfied with this test, but couldn't figure out a better way to do it without significantly complicating the TAP test.

erimatnor: Fixed TAP tests by setting explicit time zone.

PR #3602, opened 2021-09-20
Post 2.4.2 release

svenklemm:
Update version numbers and add 2.4.2 to update tests.


gayyappan (approved):

sql/updates/latest-dev.sql
```diff
@@ -0,0 +1,2 @@
+DROP FUNCTION IF EXISTS _timescaledb_internal.time_col_name_for_chunk(name,name);
```
erimatnor, line 1: Can you explain why these two functions are dropped in the update script when adding 2.4.2 to update test? It probably warrants an explanation in the commit message as well.

mfundul (approved):

PR #3615, opened 2021-09-23
Make ssl_dir/passfile gucs accessible for read

### LINKED ISSUE #3608: Make `timescaledb.ssl_dir` GUC visible to all users
`timescaledb.ssl_dir` GUC is currently only visible for superusers. This makes it impossible for non superusers to read GUC value which might be useful in some cases (eg. troubleshooting or building multinode automation)
Since this GUC does not contain any sensitive information it makes sense to make it public.

pmwkaa:
`timescaledb.ssl_dir/passfile` GUC's is currently only visible for superusers. This makes it impossible for non superusers to read GUC value which might be useful in some cases (eg. troubleshooting or building multinode automation).

Since this GUC's does not contain any sensitive information it makes sense to make it public.

Fixes #3608

### DIFF
--- a/tsl/test/sql/dist_util.sql
+++ b/tsl/test/sql/dist_util.sql
@@ -278,6 +278,19 @@ SELECT * FROM chunk_compression_stats('nondisttable') ORDER BY chunk_schema, chu
 SELECT * FROM hypertable_index_size('disttable_pkey');
 SELECT * FROM hypertable_index_size('nondisttable_pkey');
 
+-- Make sure timescaledb.ssl_dir and passfile gucs can be read by a non-superuser
+\c :TEST_DBNAME :ROLE_1
+\unset ECHO
+\o /dev/null
+SHOW timescaledb.ssl_dir;
+SHOW timescaledb.passfile;
+\o
+\set ECHO all
+\set ON_ERROR_STOP 0
+SET timescaledb.ssl_dir TO 'ssldir';
+SET timescaledb.passfile TO 'passfile';
+\set ON_ERROR_STOP 1
+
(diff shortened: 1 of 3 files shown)


tsl/test/sql/dist_util.sql
```diff
@@ -278,6 +278,19 @@ SELECT * FROM chunk_compression_stats('nondisttable') ORDER BY chunk_schema, chu
 SELECT * FROM hypertable_index_size('disttable_pkey');
 SELECT * FROM hypertable_index_size('nondisttable_pkey');
 
+-- Make sure timescaledb.ssl_dir and passfile gucs can be read by a non-superuser
+\c :TEST_DBNAME :ROLE_1
+\unset ECHO
+\o /dev/null
+SHOW timescaledb.ssl_dir;
```
erimatnor, line 285: The output isn't shown in the test. Is this intentional. Now it is hard to see that this works.

pmwkaa: This is intentional. Otherwise test would be non-deterministic, because it shows full path which depends on an environment. In this test case we are interested only that the guc is accessible, without this fix it would generate and error and stop execution.

nikkhils: I think the default output for `timescaledb.ssl_dir` is ""? Which should be ok and will be deterministic. But I do get your point that if permissions are not there then it would error out.

nikkhils (approved):
If default output is shown for `timescaledb.ssl_dir` then we can use that in the output file. Otherwise LGTM.

erimatnor (approved):

PR #3635, opened 2021-10-01
Disable per data node queries in distributed CAGG

mfundul:
Disable per-data-node optimization so that 'tableoid' system column is evaluated in the Access Node to generate Access Node chunk-IDs for the materialization table.

### DIFF
--- a/tsl/src/continuous_aggs/refresh.c
+++ b/tsl/src/continuous_aggs/refresh.c
@@ -28,6 +28,7 @@
 #include "materialize.h"
 #include "invalidation.h"
 #include "invalidation_threshold.h"
+#include "guc.h"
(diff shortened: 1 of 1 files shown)


erimatnor (commented):
FWIW: disabling this optimization will be really bad for performance.

mfundul:
> 
> 
> FWIW: disabling this optimization will be really bad for performance.

That is an obvious optimization to do next for sure. I am not sure if it's going to be "really bad" in this scenario, I was hoping it would be "simply worse". I guess we need some benchmarking to measure the impact on the materialization query.

mkindahl (approved):
LGTM for merging with feature branch.

PR #3638, opened 2021-10-03
Allow owner change of a data node

### LINKED ISSUE #3617: Don’t block operations on local objects
In a multinode setup each data node is represented as a foreign server on the access node. 
Currently it is not possible to change the owner of the foreign server:
`ERROR:  alter owner not supported on a TimescaleDB data node`

Being able to manage ownership of foreign servers is useful for various things. For example by changing a
foreign server owner one can block/unblock data node removal.

This is maybe applicable to other objects as well.

pmwkaa:
Allow previously blocked `ALTER SERVER OWNER TO` operation
on a data node foreign server.

Fix #3617


erimatnor (commented):
I am wondering if we shouldn't support this via an `alter_data_node` function instead of opening up the foreign server API?

I guess it doesn't hurt to enable this, but it is a bit awkward to mix UIs (create via `add_data_node` but alter via `ALTER SERVER`).

pmwkaa:
> I am wondering if we shouldn't support this via an `alter_data_node` function instead of opening up the foreign server API?
> 
> I guess it doesn't hurt to enable this, but it is a bit awkward to mix UIs (create via `add_data_node` but alter via `ALTER SERVER`).

Agree, that's why we blocked it before I believe. I guess currently there are not much other things to support, beside maybe the `rename` support. also altering server options directly, might be unsafe

nikkhils:
@pmwkaa are you going to update this PR based on Erik's comments above?

PR #3648, opened 2021-10-05
Add PG14 to our CI

svenklemm:
This only adds the infrastructe for running make installcheck on PG14 we still need the INSERT and UPSERT fixes for PG14 merged to master for these tests to pass.

Disable-check: commit-count


gayyappan (approved):
Did not review the .out files as they don't have the correct expected output yet (since inserts don't work on PG14 yet).

fabriziomello (approved):
LGTM

PR #3657, opened 2021-10-06
Clarify the comment for invalidation threshold

afiskon:
Explicitly state that the invalidation threshold is ignored by the transactions
that use an isolation level above READ COMMITTED.


tsl/src/continuous_aggs/invalidation_threshold.c
```diff
@@ -57,6 +57,9 @@
  *                                        |
  *                               invalidation threshold
  *
+ * Transactions that use an isolation level above READ COMMITED don't have
+ * the access to the recent invalidation threshold. Thus such transactions
+ * always create records in the hypertable invalidation log.
  */
```
mfundul, line 63: You should point to the original comment which is in `cache_inval_entry_write()` in `tsl/src/continuous_aggs/insert.c`. Also, `COMMITTED` needs 2 `T`.

mfundul (changes requested):

mfundul (approved):

afiskon:
@mkindahl could you please elaborate on the complications in respect of maintenance? I don't see why we should gather and track completely unrelated and independent changed in feature branches instead of merging them right to the master branch. This will complicate the maintenance for sure.

PR #3661, opened 2021-10-06
Fix SkipScan path generation with constant DISTINCT column

### LINKED ISSUE #3629: DISTINCT ON causes error when using with array of single value and order by second parameter
Currently I'm using PostgreSQL 12.8 with TimescaleDB 2.2.0. 

Yesterday I test the migration to the latest TimescaleDB version and got the error with few queries.

I create the repo to reproduce the problem: https://github.com/binakot/timescaledb-error-distinct-on-single-object-array-and-second-order-by.

---

## Summary

```sql
SELECT DISTINCT ON (id) * FROM test
WHERE id IN (2)
ORDER BY id ASC, time DESC;
```

When we use `DISTINCT ON` with `IN array` of single object and second condition in `ORDER BY` then query doesn't work:

```sql
ERROR:  column not present in index: 1
```

The most interesting that the next queries work:

```sql
SELECT DISTINCT ON (id) * FROM test
WHERE id IN (2,3) -- few values in array
ORDER BY id ASC, time DESC;
```

```sql
SELECT DISTINCT ON (id) * FROM test
WHERE id IN (2)
ORDER BY id ASC; -- no second condition
```

It happens in hypertables only. For default tables of PostgreSQL problem doesn't happen.

svenklemm:
When a DISTINCT query has a WHERE clause that constifies the
DISTINCT column the query might use an index that does not have
include the DISTINCT column even though it is referenced in the
ORDER BY clause. The skipscan path generation would error on any
path with such a configuration. This patch changes the path
generation code to skip generating SkipScan path under these
circumstances.

Fixes #3629

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ accidentally triggering the load of a previous DB version.**
 **Bugfixes**
 * #3580 Fix memory context bug executing TRUNCATE
 * #3654 Fix index attnum mapping in reorder_chunk
+* #3661 Fix SkipScan path generation with constant DISTINCT column
+
+**Thanks**
+* @binakot and @sebvett for reporting an issue with DISTINCT queries
 
 **Thanks**
 * @hardikm10, @DavidPavlicek and @pafiti for reporting bugs on TRUNCATE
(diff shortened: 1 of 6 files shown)


fabriziomello:
Just need to add `tsl/test/expected/plan_skip_scan-12.out`, other than that LGTM

fabriziomello (approved):

PR #3678, opened 2021-10-11
Add handling of new ALTER TABLE options in PG14

fabriziomello:
PG14 introduced new `ALTER TABLE` sub-commands:

* `.. ALTER COLUMN .. SET COMPRESSION`: handled it properly on
  `process_utility` hook code and added related regression tests

* `.. DETACH PARTITION .. {CONCURRENTLY | FINALIZE}`: handled it
  properly on `process_utility` hook code but there's no need to add
  regression tests because we don't rely to native partitioning in
  hypertables.


fabriziomello:
> Could we also add a test to show that you cannot ALTER compression on a non-toasted column?

Added!

PR #3717, opened 2021-10-18
Support transparent decompression on individual chunks

### LINKED ISSUE #3714: cannot read compressed chunks from distributed hypertables
 - OS: Ubuntu 20.04
 - TimescaleDB version (output of `\dx` in `psql`): 2.4.2
 - Installation method: source

**Describe the bug**
I cannot read any data from compressed chunks if I disable per data node queries. See the steps below.

**To Reproduce**
1. Create a compressed distributed hypertable:
```
CREATE TABLE test_table_int(time BIGINT, val INT);
SELECT create_distributed_hypertable('test_table_int', 'time', chunk_time_interval => 1, replication_factor => 1);

CREATE OR REPLACE FUNCTION dummy_now() RETURNS BIGINT LANGUAGE SQL IMMUTABLE as 'SELECT 5::BIGINT';
CALL distributed_exec($$
CREATE OR REPLACE FUNCTION dummy_now() RETURNS BIGINT LANGUAGE SQL IMMUTABLE as 'SELECT 5::BIGINT'
$$);

SELECT set_integer_now_func('test_table_int', 'dummy_now');
INSERT INTO test_table_int SELECT generate_series(1,5), 10;
ALTER TABLE test_table_int SET (timescaledb.compress);
```
2. Compress the first chunk:
```
SELECT compress_chunk(chunk, true)
FROM show_chunks('test_table_int') AS chunk
ORDER BY chunk LIMIT 1;
```
3. See the data:
```
postgres=# SELECT * FROM test_table_int;
 time | val 
------+-----
    1 |  10
    4 |  10
    2 |  10
    5 |  10
    3 |  10
(5 rows)
```
4. Disable per data node queries:
```
SET timescaledb.enable_per_data_node_queries=FALSE;
```
5. Observe that the data of the first chunk are missing:
```
postgres=# SELECT * FROM test_table_int;
 time | val 
------+-----
    2 |  10
    3 |  10
    4 |  10
    5 |  10
(4 rows)
```

**Expected behavior**
No data should be missing.

**Actual behavior**
Compressed chunks cannot be read.

**Additional context**
This is needed to implement continuous aggregates over distributed hypertables that can be compressed because we need to disable per data node queries to access system columns.

svenklemm:
In queries on distributed hypertables with per datanode queries
disabled transparent decompression would not work because the
query sent to the data node was for individual chunks. To support
queries on distributed hypertables with per datanode queries
disabled this patch adds support for transparent decompression
in queries on individual chunks.

Fixes #3714

### DIFF
--- a/.github/workflows/linux-32bit-build-and-test.yaml
+++ b/.github/workflows/linux-32bit-build-and-test.yaml
@@ -23,13 +23,13 @@ jobs:
         build_type: [ Debug ]
         include:
           - pg: "12.8"
-            ignores: append-12 debug_notice transparent_decompression-12 plan_skip_scan-12 pg_dump
+            ignores: append-12 debug_notice transparent_decompression-12 transparent_decompress_chunk-12 plan_skip_scan-12 pg_dump
             pg_major: 12
           - pg: "13.4"
-            ignores: append-13 debug_notice transparent_decompression-13 pg_dump
+            ignores: append-13 debug_notice transparent_decompression-13 transparent_decompress_chunk-13 pg_dump
             pg_major: 13
           - pg: "14.0"
-            ignores: append-14 debug_notice transparent_decompression-14 pg_dump
+            ignores: append-14 debug_notice transparent_decompression-14 transparent_decompress_chunk-14 pg_dump
             pg_major: 14
 
     steps:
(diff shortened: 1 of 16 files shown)


svenklemm:
> Wasn't this broken also on single node if you queried a chunk directly? So, it doesn't seem specific to per-datanode queries being disabled, or am I missing something?
> 
> I would also expect new tests to cover transparent decompressions on individual chunks, but don't see any.

There wasn't really a usecase for this on singlenode but it is a required feature for querying distributed hypertable without per datanode queries. But you are correct that it didn't work at all previously independent of singlenode/multinode.

PR #3726, opened 2021-10-21
Ship any stable functions to data nodes, not only the whitelisted timestamp-related ones.

akuzm:
Let's see what breaks.

Fixes https://github.com/timescale/timescaledb-private/issues/509

TODO

- [x] add setting to disable shipping stable functions and switch back to a whitelist. This keeps the old code path, but it's simple, and integration is literally one `if`. Probably makes sense to have an enum settings like `push_down_quals = 'whitelist'/'stable'/'const'/whatever we might come up with`.
- [x] if we start doing this by default, this is backward-incompatible change in the sense that it can break some queries and require tweaking the settings on the data nodes. Mention this in changelog.
- [x] investigate what happens with expressions like `to_tsvector(column)`. Are they calculated on access node or on data nodes? <details>Seems that we only fetch bare Vars from the foreign rel, no matter what the foreign rel reltarget is on the access node (`build_tlist_to_deparse`). So `to_tsvector(column)` is calculated on the access node. If we start shipping quals with stable functions to data nodes, they start to be evaluated at data nodes. Some examples with to_tsvector below. It looks really weird when the stable functions are evaluated differently, probably have to start shipping the changed settings first. </details>
- [x] think about settings influencing evaluation of stable functions, create an issue for that https://github.com/timescale/timescaledb/issues/3771
- [ ] look at the existing stable functions to see which of them might be generally useful. `select pronamespace, prorettype, proargtypes, proname, (select array_agg(oprname) from pg_operator where oprcode=pg_proc.oid) t from pg_proc where provolatile = 's' order by pronamespace, proname;`. Some obvious stuff is text search functions (not widely useful), date/time functions (mostly whitelisted), concat.

<details><summary>some examples with to_tsvector</summary>

```
select to_tsvector('testing');
 to_tsvector 
─────────────
 'test':1

set default_text_search_config = 'simple';
select to_tsvector('testing');
 to_tsvector 
─────────────
 'testing':1


create table test_ts (ts timestamp, text text);
select create_distributed_hypertable('test_ts', 'ts');
insert into test_ts values ('2021-01-01 00:00:00', 'testing');
select * from test_ts;
         ts          │  text   
─────────────────────┼─────────
 2021-01-01 00:00:00 │ testing


-- 0 rows with different text search config
explain (analyze, verbose) select * from test_ts where to_tsvector(text) = to_tsvector('testing');
                                                                                 QUERY PLAN                                                                                 
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Custom Scan (DataNodeScan) on public.test_ts  (cost=100.00..22768203.16 rows=218604 width=40) (actual time=2.596..2.597 rows=0 loops=1)
   Output: test_ts.ts, test_ts.text
   Data node: data1
   Chunks: _dist_hyper_8_123_chunk
   Remote SQL: SELECT ts, text FROM public.test_ts WHERE _timescaledb_internal.chunks_in(public.test_ts.*, ARRAY[54]) AND ((to_tsvector(text) = '''testing'':1'::tsvector))
 Planning Time: 1.761 ms
 Execution Time: 4.349 ms
(7 rows)

Time: 24,341 ms


-- 1 row with same text search config
set default_text_search_config = 'english';
explain (analyze, verbose) select * from test_ts where to_tsvector(text) = to_tsvector('testing');
                                                                               QUERY PLAN                                                                                
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Custom Scan (DataNodeScan) on public.test_ts  (cost=100.00..22768203.16 rows=218604 width=40) (actual time=3.145..3.148 rows=1 loops=1)
   Output: test_ts.ts, test_ts.text
   Data node: data1
   Chunks: _dist_hyper_8_123_chunk
   Remote SQL: SELECT ts, text FROM public.test_ts WHERE _timescaledb_internal.chunks_in(public.test_ts.*, ARRAY[54]) AND ((to_tsvector(text) = '''test'':1'::tsvector))
 Planning Time: 1.875 ms
 Execution Time: 4.905 ms

select * from test_ts where to_tsvector(text) = to_tsvector('testing');
         ts          │  text   
─────────────────────┼─────────
 2021-01-01 00:00:00 │ testing
```

</details>


tsl/test/expected/dist_function_pushdown.out
```diff
@@ -0,0 +1,147 @@
...
+from words where to_tsvector(text) != to_tsvector('testing');
+         to_tsvector          | to_tsvector 
+------------------------------+-------------
+ 'testing':1                  | 'testing':1
+ 'other':2 'some':1 'words':3 | 'testing':1
+(2 rows)
+
+explain (analyze, verbose, costs off, timing off, summary off)
+select to_tsvector(text), to_tsvector('testing')
+from words where to_tsvector(text) != to_tsvector('testing');
```
fabriziomello, line 126: What about also add tests mixing stable and non-stable functions to see the differences between pushed and not pushed down functions?

akuzm: Technically it can already be seen comparing the 'stable' vs the 'whitelisted' variants, but I agree that having them in the same query might be more illustrative. Added a query with `random()`.

I also added the most important case I forgot, where we push down the to_tsvector and it works correctly because the settings are the same.

tsl/test/sql/dist_function_pushdown.sql
```diff
@@ -0,0 +1,124 @@
...
+
+-- Time for the weirdly broken stuff. This configuration will evaluate the
+-- to_tsvector('testing') on the access node, giving 'testing':1,
+-- and to_tsvector(text) on the data nodes, giving 'test':1. So this row will
+-- be returned from the data node. The access node won't check the filter again.
+-- To fix, we should start shipping the changed settings to the data nodes.
+-- For the time being, either configure the data nodes manually or limit the
+-- pushdown of functions to the whitelist (timezone stuff).
+set default_text_search_config = 'simple';
+set timescaledb.pushdown_functions = 'stable';
```
pmwkaa, line 111: I would suggest simple check to set this guc to something arbitrary, to make sure the value is properly validated by postgres

pmwkaa (approved):
looks good

tsl/test/sql/dist_function_pushdown.sql
akuzm: Not sure it would add anything -- it's a enum GUC, int-valued, there's no way around validation.

erimatnor (commented):
Overall this looks good. But I think there are two open questions to resolve:

1. Should we use a GUC and what should it look like? For example, with the `whitelist` option, is it possible to modify the whitelist?
2. Given that we don't forward SET statements right now, it seems too easy to break push down of stable functions in subtle ways. I am therefore wondering if this should really be the default right now.
3. It is unclear what happens with certain stable functions that users might intend to evaluate on the access node. For example
     ```
         SELECT *, current_setting('timezone') FROM distributed_hypertable;
     ```
Since `current_setting` is marked `STABLE`, does it return the setting on the AN or the DNs? I am wondering if there are other stable functions that users might actually intend to evaluate on the access node. 

I ran this test on your code:

```
select time_bucket('1 hour', time) as hour, device_id, current_setting('timezone') from devices_data group by 1,2;
        hour         | device_id | current_setting  
---------------------+-----------+------------------
 2021-02-18 13:00:00 |         1 | Europe/Stockholm
....

explain verbose select time_bucket('1 hour', time) as hour, device_id, current_setting('timezone') from devices_data group by 1,2;
                                                                                                                  QUERY PLAN                
                                                                                                   
--------------------------------------------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------------------
 Custom Scan (DataNodeScan)  (cost=2605.61..17517.42 rows=40000 width=44)
   Output: (time_bucket('01:00:00'::interval, devices_data."time")), devices_data.device_id, (current_setting('timezone'::text))
   Relations: Aggregate on (public.devices_data)
   Data node: dn1
   Chunks: _dist_hyper_1_1_chunk, _dist_hyper_1_2_chunk, _dist_hyper_1_3_chunk, _dist_hyper_1_4_chunk, _dist_hyper_1_5_chunk
   Remote SQL: SELECT public.time_bucket('01:00:00'::interval, "time"), device_id, current_setting('timezone'::text) FROM public.devices_dat
a WHERE _timescaledb_internal.chunks_in(public.devices_data.*, ARRAY[1, 2, 3, 4, 5]) GROUP BY 1, 2
(6 rows)

postgres=# set timezone=utc;
SET
postgres=# select time_bucket('1 hour', time) as hour, device_id, current_setting('timezone') from devices_data group by 1,2;
        hour         | device_id | current_setting 
---------------------+-----------+-----------------
 2021-02-18 13:00:00 |         1 | UTC
```

The behavior is quite strange. First off, should we really push down `current_setting` in this case? It is not clear. I guess we can argue that timezones should be the same on all nodes, so it shouldn't matter. But there might be other stable functions that return, e.g., the result of a query on another table that might have different state on the AN vs DNs (or the table doesn't exist at all).

Also, in the above test, when I changed the timezone on the AN so that the AN and DN time zone were different, the query returned the timezone of the AN despite the explain showing the expression being part of the remote SQL. How does that work?

akuzm:
@phemmer Do you have some stable functions in mind that could benefit from this pushdown? We were discussing this PR with @erimatnor and thought that maybe if we don't have any real-world examples that improve, we shouldn't implement this at all. Or just extend the whitelist. I posted an example with `to_tsvector` earlier, but it is somewhat synthetic.

phemmer:
I use `concat()` and `format()` a fair amount. Sometimes `array_to_string()`. I'm not sure how much it would benefit to push these down, since I imagine a similar amount of bytes comes back, just structured/formatted differently. It probably just comes down to processing power.

akuzm:
> I use `concat()` and `format()` a fair amount. Sometimes `array_to_string()`. I'm not sure how much it would benefit to push these down, since I imagine a similar amount of bytes comes back, just structured/formatted differently. It probably just comes down to processing power.
> 
> Edit: And thinking about it, I'm typically using `array_to_string()` on joins, so that's not a pushdown candidate anyway. So scratch that one.

Note that this optimization applies only to WHERE. The expressions from SELECT are still going to be evaluated on the access node. The benefit comes from evaluating the WHERE clauses with stable functions on data nodes, instead of sending all these rows to the access node like we do now. In the synthetic example with `to_tsvector` I posted above the speedup is huge because this condition filters out half of the rows, but not sure how well this translates to the real-life queries.

phemmer:
Ah, in that case, no, I can't think of many cases where I'm using stable functions inside where clauses (other than the time/date functions which are already addressed). The json/jsonb functions might be useful for others, but I don't use those data types.

akuzm:
I'll close this one and extend the whitelist here: https://github.com/timescale/timescaledb/pull/3815

PR #3104, opened 2021-04-12
Fix use after free in add_reorder_policy

svenklemm:
The hypertable object in policy_reorder_add was still used after
releasing the cache leading to occasional segfaults in CI.

https://github.com/timescale/timescaledb/pull/3100/checks?check_run_id=2321566721
Fixes https://github.com/timescale/timescaledb-private/issues/906

PR #3743, opened 2021-10-26
Add coccinelle workflow to CI

svenklemm:
This patch adds a workflow to run coccinelle scripts against our
codebase. This can be used to find errors in the code base.
The initial patch adds a check to find unbalanced heap_form_tuple
heap_freetuple occurances.

https://coccinelle.gitlabpages.inria.fr/website/

### DIFF
--- a/.github/workflows/coccinelle.yaml
+++ b/.github/workflows/coccinelle.yaml
@@ -1,0 +1,32 @@
+name: Coccinelle
+on:
+  pull_request:
+  push:
+jobs:
+  coccinelle:
+    name: Coccinelle
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Install Dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install software-properties-common
+        sudo add-apt-repository ppa:npalix/coccinelle
+        sudo apt-get update
+        sudo apt-get install coccinelle
+
+    - name: Checkout TimescaleDB
+      uses: actions/checkout@v2
+
+    - name: Run coccinelle
+      run: |
+        ./scripts/coccinelle.sh
+
+    - name: Save coccinelle.diff
+      if: always()
+      uses: actions/upload-artifact@v2
+      with:
+        name: coccinelle.diff
+        path: coccinelle.diff
+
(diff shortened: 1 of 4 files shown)


coccinelle/heap_freetuple.cocci
```diff
@@ -0,0 +1,29 @@
...
+(
+return tuple;
+|
+heap_freetuple(tuple)
+|
+HeapTupleGetDatum(tuple)
+)
+@depends on !safelist@
+expression tuple;
+position heap_form_tuple.p;
```
akuzm, line 25: OK, I think I kinda understand what it does after looking at it for a couple of minutes. Maybe some high-level comment about this language would help, since it's not the main language we use.

By the way, it looks like we first match `@heap_form_tuple@`, then `@safelist@`, then `@depends on !safelist@`? Does the first match add anything, e.g. can we instead match `@safelist@` and then `@depends on !safelist@`?

Also, position in `@depends on !safelist@` is `heap_form_tuple.p`, but then we use just `p` in the diff, is it intentional?

akuzm (approved):
Not sure how useful it is, but having some stuff from Inria in our CI is good blessing :smile:

.github/workflows/coccinelle.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Checkout TimescaleDB
+      uses: actions/checkout@v2
+
+    - name: Run coccinelle
+      run: |
+        for f in coccinelle/*.cocci; do
+          find . -name '*.c' | xargs spatch --very-quiet -sp_file $f | tee -a coccinelle.diff
+        done
+
+        if [ -s coccinelle.diff ] ; then exit 1; fi
```
akuzm, line 28: Is this diff uploaded somewhere?

akuzm: I see that it goes to stderr, OK https://github.com/timescale/timescaledb/runs/4024942886?check_suite_focus=true

svenklemm:
> Not sure how useful it is, but having some stuff from Inria in our CI is good blessing 😄

It found a couple bugs already e.g. #3104, #3106, #3732

coccinelle/heap_freetuple.cocci
svenklemm: I'll add a readme with some short intro. The first block is the anchor which gets referenced in the safelist and the output block.  inside the @@ is the block header which has variable declarations which can reference other blocks. Inside the diff part of the block you can't namespace variables as that is set up in the block header.

PR #3753, opened 2021-10-28
Monthly buckets support in CAGGs

afiskon:
```
Monthly buckets support in CAGGs

This patch allows using time_bucket_ng("N month", ...) in CAGGs. Users can also
specify years, or months AND years. CAGGs on top of distributed hypertables
are supported as well.
```

The RFC section that clarifies the changes in the catalog schema:

```
While reviewing Monthly buckets support in CAGGs pull request the team agreed
on the following iterative changes in the catalog schema.

When user creates a CAGG with a variable bucket, the corresponding `bucket_width`
value in the `_timescaledb_catalog.continuous_agg` table is -1 (declared as
BUCKET_WIDTH_VARIABLE in continuous_agg.h). This indicates that the bucket width
is variable and the information about the bucketing function is stored in the
`_timescaledb_catalog.continuous_aggs_bucket_function` table:

CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
  mat_hypertable_id integer PRIMARY KEY
    REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
  -- The schema of the function.
  -- Equals TRUE for "timescaledb_experimental", FALSE otherwise.
  experimental bool NOT NULL,
  -- Name of the bucketing function
  name text NOT NULL,
  -- `bucket_width` argument of the function, e.g. "1 month"
  bucket_width text NOT NULL,
  -- `origin` argument of the function provided by the user
  origin text NOT NULL,
  -- `timezone` argument of the function provided by the user
  timezone text NOT NULL);

This is a TEMPORARY SCHEMA that will be used while the feature evolves. It is
flexible because of using TEXT fields, and also because basically, it stores all
the data provided by the user when creating the CAGG. On the flip side, there
are no constraints and/or type checks.

Sooner or later we are going to graduate time_bucket_ng() from the experimental
schema. We discovered at least three different ways to do it, all with their
pros and cons. The least risky one is to let time_bucket() and time_bucket_ng()
co-exist for several releases and let the users do the migration manually. This
is why there are `experimental` and `name` columns in the 
`continuous_aggs_bucket_function` table:

- experimental = true, name = “time_bucket_ng” says that
  timescaledb_experimental.time_bucket_ng() is used as a bucketing function for
  the given CAGG;
- experimental = false, name = “time_bucket_ng” indicates public.time_bucket_ng();
- experimental = false, name = “time_bucket” indicates public.time_bucket();

These columns can be deleted after the graduation of time_bucket_ng(). There
will be only one possibility - public.time_bucket().

`bucket_with` will be converted to INTERVAL type, `origin` --- to TIMESTAMP and 
`timezone` --- to TEXT with CHECK. We almost certainly will be able to use the 
same algorithms for variable-sized and fixed-sized buckets, thus `bucket_width`
in the `_timescaledb_catalog.continuous_agg` will be deleted. The table
`continuous_aggs_bucket_function` will be deleted and its `bucket_width`,
`origin` and `timezone` columns will be transferred to `continuous_agg`.

At least, that’s the plan. Since there are way too many unknowns at this point,
we prefer to use a more flexible and less constraining schema until the feature
stabilizes.
```

### DIFF
--- a/sql/cagg_utils.sql
+++ b/sql/cagg_utils.sql
@@ -53,6 +53,7 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.drop_dist_ht_invalidation_trigg
 --                 'raw_hypertable_id'
 -- max_bucket_widths - The array of the maximum time bucket widths for all the CAGGs that belong
 --                     to 'raw_hypertable_id'
+-- bucket_functions - (Optional) The array of serialized information about bucket functions
 CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_hypertable_log(
     mat_hypertable_id INTEGER,
     raw_hypertable_id INTEGER,
@@ -60,6 +61,16 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_hypertable
     mat_hypertable_ids INTEGER[],
     bucket_widths BIGINT[],
     max_bucket_widths BIGINT[]
+) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_invalidation_process_hypertable_log' LANGUAGE C STRICT VOLATILE;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_hypertable_log(
+    mat_hypertable_id INTEGER,
+    raw_hypertable_id INTEGER,
+    dimtype REGTYPE,
+    mat_hypertable_ids INTEGER[],
+    bucket_widths BIGINT[],
+    max_bucket_widths BIGINT[],
+    bucket_functions TEXT[]
 ) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_invalidation_process_hypertable_log' LANGUAGE C STRICT VOLATILE;
 
 -- Processes the materialization invalidation log in a data node for the CAGG being refreshed that
@@ -78,6 +89,7 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_hypertable
 --                 'raw_hypertable_id'
 -- max_bucket_widths - The array of the maximum time bucket widths for all the CAGGs that belong
 --                     to 'raw_hypertable_id'
+-- bucket_functions - (Optional) The array of serialized information about bucket functions
 --
 -- Returns a tuple of:
 -- ret_window_start - The merged refresh window starting time
@@ -91,6 +103,20 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
     mat_hypertable_ids INTEGER[],
     bucket_widths BIGINT[],
     max_bucket_widths BIGINT[],
+    OUT ret_window_start BIGINT,
+    OUT ret_window_end BIGINT
+) RETURNS RECORD AS '@MODULE_PATHNAME@', 'ts_invalidation_process_cagg_log' LANGUAGE C STRICT VOLATILE;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
+    mat_hypertable_id INTEGER,
+    raw_hypertable_id INTEGER,
+    dimtype REGTYPE,
+    window_start BIGINT,
+    window_end BIGINT,
+    mat_hypertable_ids INTEGER[],
+    bucket_widths BIGINT[],
+    max_bucket_widths BIGINT[],
+    bucket_functions TEXT[],
     OUT ret_window_start BIGINT,
     OUT ret_window_end BIGINT
 ) RETURNS RECORD AS '@MODULE_PATHNAME@', 'ts_invalidation_process_cagg_log' LANGUAGE C STRICT VOLATILE;
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -280,6 +280,23 @@ CREATE INDEX IF NOT EXISTS continuous_agg_raw_hypertable_id_idx ON _timescaledb_
 
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_agg', '');
 
+-- See the comments for ContinuousAggsBucketFunction structure.
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
+  mat_hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
+  -- The schema of the function. Equals TRUE for "timescaledb_experimental", FALSE otherwise.
+  experimental bool NOT NULL,
+  -- Name of the bucketing function, e.g. "time_bucket" or "time_bucket_ng"
+  name text NOT NULL,
+  -- `bucket_width` argument of the function, e.g. "1 month"
+  bucket_width text NOT NULL,
+  -- `origin` argument of the function provided by the user
+  origin text NOT NULL,
+  -- `timezone` argument of the function provided by the user
+  timezone text NOT NULL
+);
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_bucket_function', '');
+
 CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_invalidation_threshold (
   hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
   watermark bigint NOT NULL
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -1,2 +1,26 @@
 DROP FUNCTION IF EXISTS recompress_chunk;
 
+
+-- Also see the comments for ContinuousAggsBucketFunction structure.
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
+  mat_hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
+  -- The schema of the function. Equals TRUE for "timescaledb_experimental", FALSE otherwise.
+  experimental bool NOT NULL,
+  -- Name of the bucketing function, e.g. "time_bucket" or "time_bucket_ng"
+  name text NOT NULL,
+  -- `bucket_width` argument of the function, e.g. "1 month"
+  bucket_width text NOT NULL,
+  -- `origin` argument of the function provided by the user
+  origin text NOT NULL,
+  -- `timezone` argument of the function provided by the user
+  timezone text NOT NULL
+);
+
+-- in tables.sql the same is done with GRANT SELECT ON ALL TABLES IN SCHEMA
+GRANT SELECT ON _timescaledb_catalog.continuous_aggs_bucket_function TO PUBLIC;
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_bucket_function', '');
+
+-- Adding overloaded versions of invalidation_process_hypertable_log() and invalidation_process_cagg_log()
+-- with bucket_functions argument is done in cagg_utils.sql. Note that this file is included when building
+-- the update scripts, so we don't have to do it here.
--- a/sql/updates/reverse-dev.sql
+++ b/sql/updates/reverse-dev.sql
@@ -1,2 +1,77 @@
+
 DROP PROCEDURE IF EXISTS recompress_chunk;
 DROP FUNCTION IF EXISTS _timescaledb_internal.chunk_status;
+
+DO $$
+DECLARE
+    caggs text[];
+    caggs_nr int;
+BEGIN
+    SELECT array_agg(format('%I.%I', user_view_schema, user_view_name)) FROM _timescaledb_catalog.continuous_agg WHERE bucket_width < 0 INTO caggs;
+    SELECT array_length(caggs, 1) INTO caggs_nr;
+    IF caggs_nr > 0 THEN
+        RAISE EXCEPTION 'Downgrade is impossible since % continuous aggregates exist which use variable buckets: %', caggs_nr, caggs
+              USING HINT = 'Remove the corresponding continuous aggregates manually before downgrading';
+    END IF;
+
+    -- It's safe to drop the table.
+    -- ALTER EXTENSION is required to revert the effect of pg_extension_config_dump()
+    -- See "The list of tables configured to be dumped" test in test/sql/updates/post.catalog.sql
+    ALTER EXTENSION timescaledb DROP TABLE _timescaledb_catalog.continuous_aggs_bucket_function;
+
+    -- Actually drop the table.
+    -- ALTER EXTENSION only removes the table from the extension but doesn't drop it.
+    DROP TABLE IF EXISTS _timescaledb_catalog.continuous_aggs_bucket_function;
+
+    -- Drop overloaded versions of invalidation_process_hypertable_log() and invalidation_process_cagg_log()
+    -- with bucket_functions argument.
+
+    ALTER EXTENSION timescaledb DROP FUNCTION _timescaledb_internal.invalidation_process_hypertable_log(
+        mat_hypertable_id INTEGER,
+        raw_hypertable_id INTEGER,
+        dimtype REGTYPE,
+        mat_hypertable_ids INTEGER[],
+        bucket_widths BIGINT[],
+        max_bucket_widths BIGINT[],
+        bucket_functions TEXT[]
+    );
+
+    DROP FUNCTION IF EXISTS _timescaledb_internal.invalidation_process_hypertable_log(
+        mat_hypertable_id INTEGER,
+        raw_hypertable_id INTEGER,
+        dimtype REGTYPE,
+        mat_hypertable_ids INTEGER[],
+        bucket_widths BIGINT[],
+        max_bucket_widths BIGINT[],
+        bucket_functions TEXT[]
+    );
+
+    ALTER EXTENSION timescaledb DROP FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
+        mat_hypertable_id INTEGER,
+        raw_hypertable_id INTEGER,
+        dimtype REGTYPE,
+        window_start BIGINT,
+        window_end BIGINT,
+        mat_hypertable_ids INTEGER[],
+        bucket_widths BIGINT[],
+        max_bucket_widths BIGINT[],
+        bucket_functions TEXT[],
+        OUT ret_window_start BIGINT,
+        OUT ret_window_end BIGINT
+    );
+
+    DROP FUNCTION IF EXISTS _timescaledb_internal.invalidation_process_cagg_log(
+        mat_hypertable_id INTEGER,
+        raw_hypertable_id INTEGER,
+        dimtype REGTYPE,
+        window_start BIGINT,
+        window_end BIGINT,
+        mat_hypertable_ids INTEGER[],
+        bucket_widths BIGINT[],
+        max_bucket_widths BIGINT[],
+        bucket_functions TEXT[],
+        OUT ret_window_start BIGINT,
+        OUT ret_window_end BIGINT
+    );
+END
+$$ LANGUAGE 'plpgsql';
--- a/src/catalog.h
+++ b/src/catalog.h
@@ -54,6 +54,8 @@ typedef enum CatalogTable
 	COMPRESSION_CHUNK_SIZE,
 	REMOTE_TXN,
 	CHUNK_COPY_OPERATION,
+	CONTINUOUS_AGGS_BUCKET_FUNCTION,
+	/* Don't forget updating catalog.c when adding new tables! */
 	_MAX_CATALOG_TABLES,
 } CatalogTable;
 
@@ -873,10 +875,17 @@ typedef struct FormData_continuous_agg
 	NameData partial_view_schema;
 	NameData partial_view_name;
 	/*
-	 * Don't access bucket_width directly to determine the width of the bucket.
-	 * Use corresponding procedures instead:
+	 * bucket_width is BUCKET_WIDTH_VARIABLE (see continuous_agg.h) for buckets
+	 * with variable size - monthly buckets, buckets with timezone, etc. For such
+	 * buckets the information about the bucketing function is stored in
+	 * _timescaledb_catalog.continuous_aggs_bucket_function.
+	 *
+	 * When possible, don't access bucket_width directly. Use corresponding
+	 * procedures instead, such as:
+	 * - ts_continuous_agg_bucket_width_variable
 	 * - ts_continuous_agg_bucket_width
 	 * - ts_continuous_agg_max_bucket_width
+	 * - ts_bucket_function_to_bucket_width_in_months
 	 */
 	int64 bucket_width;
 	NameData direct_view_schema;
@@ -931,6 +940,37 @@ typedef enum Anum_continuous_agg_raw_hypertable_id_idx
 
 #define Natts_continuous_agg_raw_hypertable_id_idx                                                 \
 	(_Anum_continuous_agg_raw_hypertable_id_idx_max - 1)
+
+/*** continuous_aggs_bucket_function table definitions ***/
+
+#define CONTINUOUS_AGGS_BUCKET_FUNCTION_TABLE_NAME "continuous_aggs_bucket_function"
+typedef enum Anum_continuous_aggs_bucket_function
+{
+	Anum_continuous_aggs_bucket_function_mat_hypertable_id = 1,
+	Anum_continuous_aggs_bucket_function_experimental,
+	Anum_continuous_aggs_bucket_function_name,
+	Anum_continuous_aggs_bucket_function_bucket_width,
+	Anum_continuous_aggs_bucket_function_origin,
+	Anum_continuous_aggs_bucket_function_timezone,
+	_Anum_continuous_aggs_bucket_function_max,
+} Anum_continuous_aggs_bucket_function;
+
+#define Natts_continuous_aggs_bucket_function (_Anum_continuous_aggs_bucket_function_max - 1)
+
+enum
+{
+	CONTINUOUS_AGGS_BUCKET_FUNCTION_PKEY_IDX = 0,
+	_MAX_CONTINUOUS_AGGS_BUCKET_FUNCTION_INDEX,
+};
+
+typedef enum Anum_continuous_aggs_bucket_function_pkey
+{
+	Anum_continuous_aggs_bucket_function_pkey_mat_hypertable_id = 1,
+	_Anum_continuous_aggs_bucket_function_pkey_max,
+} Anum_continuous_aggs_bucket_function_pkey;
+
+#define Natts_continuous_aggs_bucket_function_pkey                                                 \
+	(_Anum_continuous_aggs_bucket_function_pkey_max - 1)
 
 /****** CONTINUOUS_AGGS_HYPERTABLE_INVALIDATION_LOG_TABLE definitions*/
 #define CONTINUOUS_AGGS_HYPERTABLE_INVALIDATION_LOG_TABLE_NAME                                     \
--- a/src/continuous_agg.c
+++ b/src/continuous_agg.c
@@ -38,6 +38,7 @@
 #include "time_utils.h"
 #include "catalog.h"
 
+#define BUCKET_FUNCTION_SERIALIZE_VERSION 1
 #define CHECK_NAME_MATCH(name1, name2) (namestrcmp(name1, name2) == 0)
 
 static const WithClauseDefinition continuous_aggregate_with_clause_def[] = {
@@ -72,6 +73,21 @@ init_scan_by_mat_hypertable_id(ScanIterator *iterator, const int32 mat_hypertabl
 
 	ts_scan_iterator_scan_key_init(iterator,
 								   Anum_continuous_agg_pkey_mat_hypertable_id,
+								   BTEqualStrategyNumber,
+								   F_INT4EQ,
+								   Int32GetDatum(mat_hypertable_id));
+}
+
+static void
+init_scan_cagg_bucket_function_by_mat_hypertable_id(ScanIterator *iterator,
+													const int32 mat_hypertable_id)
+{
+	iterator->ctx.index = catalog_get_index(ts_catalog_get(),
+											CONTINUOUS_AGGS_BUCKET_FUNCTION,
+											CONTINUOUS_AGGS_BUCKET_FUNCTION_PKEY_IDX);
+
+	ts_scan_iterator_scan_key_init(iterator,
+								   Anum_continuous_aggs_bucket_function_pkey_mat_hypertable_id,
 								   BTEqualStrategyNumber,
 								   F_INT4EQ,
 								   Int32GetDatum(mat_hypertable_id));
@@ -166,6 +182,22 @@ invalidation_threshold_delete(int32 raw_hypertable_id)
 }
 
 static void
+cagg_bucket_function_delete(int32 mat_hypertable_id)
+{
+	ScanIterator iterator = ts_scan_iterator_create(CONTINUOUS_AGGS_BUCKET_FUNCTION,
+													RowExclusiveLock,
+													CurrentMemoryContext);
+
+	init_scan_cagg_bucket_function_by_mat_hypertable_id(&iterator, mat_hypertable_id);
+
+	ts_scanner_foreach(&iterator)
+	{
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		ts_catalog_delete_tid(ti->scanrel, ts_scanner_get_tuple_tid(ti));
+	}
+}
+
+static void
 hypertable_invalidation_log_delete(int32 raw_hypertable_id)
 {
 	ScanIterator iterator = ts_scan_iterator_create(CONTINUOUS_AGGS_HYPERTABLE_INVALIDATION_LOG,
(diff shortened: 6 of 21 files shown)


mfundul:
My first impression is that this code wouldn't work if the Access-Node runs the older version of the code and the data nodes run the new version of the code.

afiskon:
@mfundul Do you have any particular scenario in mind? Did you have a chance to take a look at `bucket_functions_default_argument()` and the comments there?

sql/updates/reverse-dev.sql
```diff
@@ -1,2 +1,49 @@
...
+        mat_hypertable_id INTEGER,
+        raw_hypertable_id INTEGER,
+        dimtype REGTYPE,
+        mat_hypertable_ids INTEGER[],
+        bucket_widths BIGINT[],
+        max_bucket_widths BIGINT[],
+        bucket_functions TEXT[]
+    );
+
+    DROP FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
```
mfundul, line 35: I think those 2 should be `DROP FUNCTION IF EXISTS` for downgrades. At least that's what we have been doing.

src/continuous_agg.c
```diff
@@ -259,59 +358,165 @@ ts_continuous_agg_get_all_caggs_info(int32 raw_hypertable_id)
...
+ * necessary.
+ */
+static const char *
+bucket_function_serialize(const ContinuousAggsBucketFunction *bf)
+{
+	const int ver = 1;
+	StringInfo str;
+
+	if (bf == NULL)
+		return pstrdup("");
```
mfundul, line 418: Is there a reason we have to use `pstrdup()` here? Why is there a requirement for the return value to only be `palloc()`ed?

src/continuous_agg.c
```diff
@@ -259,59 +358,165 @@ ts_continuous_agg_get_all_caggs_info(int32 raw_hypertable_id)
...
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("failed to deserialize \"%s\" into a bucketing function", str),
+				 errdetail("unsupported format version")));
+	}
+
+	bf = palloc(sizeof(ContinuousAggsBucketFunction));
+	bf->experimental = true;
+	bf->name = pstrdup("time_bucket_ng");
```
mfundul, line 479: Does it have to be `pstrdup()` here?

src/continuous_agg.c
```diff
@@ -259,59 +358,165 @@ ts_continuous_agg_get_all_caggs_info(int32 raw_hypertable_id)
...
+ */
+static ContinuousAggsBucketFunction *
+bucket_function_deserialize(const char *str)
+{
+	int i;
+	char *begin, *end, *strings[4];
+	ContinuousAggsBucketFunction *bf;
+
+	/* empty string stands for serialized NULL */
+	if (*str == '\0')
```
mfundul, line 446: We could allow for `const char *str` to be `NULL` since we return `NULL`. This way we won't need all those arrays and structures be populated.

sql/updates/reverse-dev.sql
afiskon: 👌

src/continuous_agg.c
afiskon: Not sure if I understand you. Could you please clarify which arrays and structures you propose not to populate and how allowing str == NULL will help with this?

src/continuous_agg.c
afiskon: Since the result is `const` anyway I have to agree that pstrdup() is redundant here.

src/continuous_agg.c
afiskon: I have mixed feelings about storing static value in a `char *` field. But I will check `const ContinuousAggsBucketFunction` can be returned here.

mfundul: The CPU cycle difference is significant here. It's either a pointer assignment or a full allocator call plus memory copy. Also, if this is called 100 times then it has 100x the memory footprint, whereas the static string is only allocated once. Plus memory allocator internal and external fragmentation overheads.

I can think of 2 reasons it would need to be in heap memory. 1 is if it needs to be modified and 2 is if it is explicitly `pfree()`ed somewhere.

afiskon: So I think it would be better to keep pstrdup() here for now. I added the comment that clarifies why it's used here.

src/continuous_agg.c
afiskon: Fixed 👍

tsl/src/continuous_aggs/invalidation.c
```diff
@@ -1060,6 +1111,30 @@ invalidation_process_hypertable_log(int32 mat_hypertable_id, int32 raw_hypertabl
...
+ * generate the default argument, and execute the rest of the logic as usual.
+ */
+static ArrayType *
+bucket_functions_default_argument(int ndim)
+{
+	int i;
+	Datum *bucketfunctions = palloc(sizeof(Datum) * ndim);
+
+	for (i = 0; i < ndim; i++)
+		bucketfunctions[i] = CStringGetTextDatum(pstrdup(""));
```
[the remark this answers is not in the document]
afiskon: Fixed.

src/continuous_agg.c
afiskon: @mfundul Personally, I don't think we should invest too much time into optimizing variable-sized buckets now especially since no one reported this particular bottleneck yet. There could be hundreds of places that affect the performance even worse (network, disks, ... to name a few). Our current goal is to receive feedback from the early adopters. Also, I suspect that when the feature stabilizes we may end up without this field at all.

But if the team feels otherwise, I'm OK with it. Just let me know if you have strong objections here and I will replace it with a pointer to a static char[].

sql/updates/latest-dev.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
+  mat_hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
```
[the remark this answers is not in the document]
afiskon: Just wanted to add that I disagree on this one:

> A constant origin does not make sense for caggs

Firstly, `origin` will be useful when people will migrate from time_bucket to time_bucket_ng() since these functions have different values of default `origin`.

Monday, Jan 3 was chosen for time_bucket() because it makes more sense for weekly buckets, and Jan 1 for time_bucket_ng() because Jan 3 doesn't make sense for monthly buckets. We didn't choose a particular strategy of promoting time_bukcet_ng() from the experiment schema yet, but one of the strategies is to keep both functions for several releases and let users do the migration manually. They will be unable to do it without specifying `origin`.

Secondly, in some countries Monday is considered to be the beginning of the week, while in others it's Sunday. Specifying `origin` will allow users to choose what is more convenient for them.

Finally, there is little reason NOT to support this feature. It's easy to implement and it costs nothing to us.

src/continuous_agg.c
mfundul: I have no strong objections. But it's not really investing too much time in optimizing anything. It would be a trivial code simplification:
```
bf->name = pstrdup("time_bucket_ng");
```
vs
```
bf->name = "time_bucket_ng";
```

afiskon: OK, you convinced me. I also changed the return type of the function to `const ContinuousAggsBucketFunction` just to be on the safe side. 👍

src/continuous_agg.c
```diff
@@ -233,6 +264,68 @@ ts_materialization_invalidation_log_delete(PG_FUNCTION_ARGS)
...
+		bool should_free;
+		HeapTuple tuple = ts_scan_iterator_fetch_heap_tuple(&iterator, false, &should_free);
+
+		/*
+		 * Our usual GETSTRUCT() approach doesn't work when TEXT fields are involved,
+		 * thus a more robust approach with heap_deform_tuple() is used here.
+		 */
+		heap_deform_tuple(tuple, ts_scan_iterator_tupledesc(&iterator), values, isnull);
+
+		Assert(!isnull[Anum_continuous_aggs_bucket_function_experimental - 1]);
```
fabriziomello, line 291:
```suggestion
		Assert(!isnull[AttrNumberGetAttrOffset(Anum_continuous_aggs_bucket_function_experimental)]);
```

afiskon: @fabriziomello Thanks for the suggestion, but I'm afraid I can't accept it. Firstly, this would be inconsistent with the rest of the code (see the lines above). Secondly, this would be inconsistent with the PostgreSQL code style which we try to follow.

sql/updates/latest-dev.sql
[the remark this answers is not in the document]
mfundul: Maybe instead of having an `experimental` field the relation itself could be named experimental-something? This would show that this metadata table is not part of the stable schema, instead of having per record state.

sql/updates/reverse-dev.sql
```diff
@@ -1,2 +1,49 @@
 -- Remove old compression policy procedures
 DROP PROCEDURE IF EXISTS _timescaledb_internal.policy_compression_execute(INTEGER, INTEGER, ANYELEMENT, INTEGER, BOOLEAN, BOOLEAN);
+
+DO $$
+DECLARE
+    caggs_nr int;
+BEGIN
+    SELECT count(*) FROM _timescaledb_catalog.continuous_agg WHERE bucket_width < 0 INTO caggs_nr;
+    IF caggs_nr > 0 THEN
+        RAISE EXCEPTION 'Downgrade is impossible since % continuous aggregates exist which use variable buckets', caggs_nr;
```
mfundul, line 10: Since this is an experimental feature you should be able to downgrade.

afiskon: @mfundul Sure. You are able to downgrade, but first, you have to _manually_ remove the CAGGs that were created using the feature you are about to downgrade. Don't think that implicitly removing user's data is a good idea.

mfundul: You should add this as a hint.

sql/updates/latest-dev.sql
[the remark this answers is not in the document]
afiskon: @mfundul Experimental column here is a compact way to store the schema of the bucketing function. This information is not used at the moment for anything, but will be needed when we will be switching from time_bucket() to time_bucket_ng().

In other words:

time_bukcet() = [experimental = false, name = "time_bucket"]
time_bucket_ng() = [experimental = true, name = "time_bucket_ng"]

During the promotion of time_bucket_ng() we may also have a situation when:

time_bucket_ng() = [experimental = false, name = "time_bucket_ng"]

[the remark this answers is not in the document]
afiskon: This is not quite obvious from the code though. I will add extra comments to the ContinuousAggsBucketFunction structure 👍

sql/updates/reverse-dev.sql
afiskon: Done! 👍

src/continuous_agg.c
```diff
@@ -233,6 +264,68 @@ ts_materialization_invalidation_log_delete(PG_FUNCTION_ARGS)
...
+		bf->experimental =
+			DatumGetBool(values[Anum_continuous_aggs_bucket_function_experimental - 1]);
+
+		Assert(!isnull[Anum_continuous_aggs_bucket_function_name - 1]);
```
mfundul, line 295: I don't particularly like this because one silent disk corruption will cause the extension to crash. I would prefer an exception to be thrown.

afiskon: I don't think I can agree. In case of disc corruption (and assuming heap checksums are turned off) the crash of TimescaleDB will be the least of the user's problems. Not mentioning that TimescaleDB is not the right place to check for disc corruption. If we start doing this there will be so many if's in the code that we can forget about the performance.

mfundul: Even after the disk has been repaired the TimescaleDB extension will still crash so it doesn't really matter what the right place for corruption handling is. This is not a transient bit-flip that will correct itself after the process restarts. This will keep crashing every time. Obviously, I don't feel strongly about this, but I am interested in how this is handled in our extension in general.

src/continuous_agg.c
```diff
@@ -1242,28 +1489,78 @@ ts_continuous_agg_watermark(PG_FUNCTION_ARGS)
...
-/*
- * Determines maximum possible bucket width for given continuous aggregate.
- *
- * E.g. for monthly continuous aggreagtes this procedure will return 31 days.
- * This is not true for ts_continuous_agg_bucket_width which may use additional
- * arguments to determine the width of a concrete bucket.
- */
+/* Determines maximum possible bucket width for given continuous aggregate. */
 int64
 ts_continuous_agg_max_bucket_width(const ContinuousAgg *agg)
```
mfundul, line 1516: The maximum widths were introduced by https://github.com/timescale/timescaledb/commit/3f9adffaf0a74d8d33a3de9f2994f86ee1c9bdf5

If I understand correctly we are abandoning them. In that case I think you should get rid of them completely.

src/continuous_agg.c
afiskon: > Obviously I don't feel strongly about this, but I am interested in how this is handled in our extension in general.

I have some experience with PostgreSQL extensions and also a very unpleasant experience of restoring databases from corrupted disks... To my knowledge, no one handles this in the extensions. I don't think this is even handled in PostgreSQL core, except for the heap checksums. (The core is huge thus I can't be 100% sure.)

Typically when the disk corruption is detected what you do is remove the DBMS instance from the cluster, repair the hardware, restore the instance from the latest backup and fill it using synchronous replication. If you don't have a backup (sadly this happens too) either your data was not that valuable and you create a new database, or you restore what you can manually, using tools like pg_filedump.

src/continuous_agg.c
afiskon: max_bucket_width existed long before 3f9adff, see CaggsInfo and CaggInvalidationState structures. In 3f9adff the access to max_bucket_with was only wrapped in a procedure, which was very helpful during developing this feature.

You are right that we don't need max_bucket_width because it always equals bucket_with. The first thing I'm going to do after merging this PR is to submit a corresponding refactoring. I have a corresponding deliverable in the epic task https://github.com/timescale/timescaledb-private/issues/926

mfundul: `CaggsInfo` was introduced quite recently and `CaggInvalidationState` does not have something either, look here:
https://github.com/timescale/timescaledb/blob/3f9adffaf0a74d8d33a3de9f2994f86ee1c9bdf5/tsl/src/continuous_aggs/invalidation.c#L80

The above is a snapshot of https://github.com/timescale/timescaledb/commit/3f9adffaf0a74d8d33a3de9f2994f86ee1c9bdf5
and there is nothing in `CaggInvalidationState`

afiskon: Hmmm... you are right, it looks like we overlooked adding max_bucket_width to CaggInvalidationState during the code review of 221437e8. There was no need for it. Anyway, I will refactor the code and get rid entirely of max_bucket_width after merging this PR.

mfundul (approved):

src/continuous_agg.c
fabriziomello: @afiskon I agree It's an old style (dated from 1996) Postgres stuff used in few places:

```sh
main on  main [$?] 
➜ git grep AttrNumberGetAttrOffset
src/backend/catalog/index.c:                                 AttrNumberGetAttrOffset(atnum));
src/backend/replication/logical/relation.c:         attnum = AttrNumberGetAttrOffset(attnum);
src/include/access/attnum.h: * AttrNumberGetAttrOffset
src/include/access/attnum.h:#define AttrNumberGetAttrOffset(attNum) \
```

But I just suggested because it is used in a lot of places in our codebase:

```sh
timescaledb on  master [$!] 
➜ git grep AttrNumberGetAttrOffset | wc -l
414
```

mkindahl (commented):
This is a partial review with a few comments that might be good to discuss/address.

sql/updates/latest-dev.sql
```diff
@@ -1,3 +1,45 @@
...
+
+-- in tables.sql the same is done with GRANT SELECT ON ALL TABLES IN SCHEMA
+GRANT SELECT ON _timescaledb_catalog.continuous_aggs_bucket_function TO PUBLIC;
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_bucket_function', '');
+
+-- Add overloaded versions of invalidation_process_hypertable_log() and invalidation_process_cagg_log()
+-- with bucket_functions argument.
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_hypertable_log(
```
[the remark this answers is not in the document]
afiskon: Yes, this is done for backward compatibility between data nodes and the access node.

sql/pre_install/tables.sql
```diff
@@ -280,6 +280,17 @@ CREATE INDEX IF NOT EXISTS continuous_agg_raw_hypertable_id_idx ON _timescaledb_
 
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_agg', '');
 
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
+  mat_hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
+  experimental bool NOT NULL,
```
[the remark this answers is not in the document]
afiskon: Ok, there was a comment in latest-dev.sql on this, but I forgot to add it here. I will fix it. The flag doesn't indicate that the feature is experimental, it indicates the schema of the function used for bucketing.

sql/updates/latest-dev.sql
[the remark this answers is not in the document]
afiskon: There is a good chance that when the feature stabilizes (after the time_bucket_ng() promotion and whatnot) we will end up without `experimental` and `name` columns. So I wouldn't invest too much time into discussing the exact schema until then.

sql/updates/reverse-dev.sql
```diff
@@ -1,2 +1,50 @@
...
+    ALTER EXTENSION timescaledb DROP TABLE _timescaledb_catalog.continuous_aggs_bucket_function;
+
+    -- Actually drop the table.
+    -- ALTER EXTENSION only removes the table from the extension but doesn't drop it.
+    DROP TABLE IF EXISTS _timescaledb_catalog.continuous_aggs_bucket_function;
+
+    -- Drop overloaded versions of invalidation_process_hypertable_log() and invalidation_process_cagg_log()
+    -- with bucket_functions argument.
+
+    DROP FUNCTION IF EXISTS _timescaledb_internal.invalidation_process_hypertable_log(
```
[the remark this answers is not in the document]
afiskon: I did some checking and it seems like I do need to do this. Here comes the corresponding commit.

sql/pre_install/tables.sql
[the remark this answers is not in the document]
afiskon: It's fixed now. The comment here says:
```
-- See the comments for ContinuousAggsBucketFunction structure.
```
Please let me know if you think it should be more detailed here.

sql/updates/latest-dev.sql
```diff
@@ -1,3 +1,45 @@
...
+    bucket_functions TEXT[]
+) RETURNS VOID AS '@MODULE_PATHNAME@', 'ts_invalidation_process_hypertable_log' LANGUAGE C STRICT VOLATILE;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
+    mat_hypertable_id INTEGER,
+    raw_hypertable_id INTEGER,
+    dimtype REGTYPE,
+    window_start BIGINT,
+    window_end BIGINT,
+    mat_hypertable_ids INTEGER[],
```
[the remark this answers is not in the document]
afiskon: Oh, I didn't notice that. Let's give it a try. Here comes the corresponding commit.

sql/pre_install/tables.sql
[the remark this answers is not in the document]
mkindahl: If the only reason for the flag is the schema, then it is probably better to follow the general scheme of having the schema name and function name in the table, similar to how it is done in other tables.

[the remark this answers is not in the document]
afiskon: @mkindahl OK, personally I don't have a strong opinion on this 🤷‍♂️ If you insist, I'll change this. Just to clarify, do you suggest changing the ContinuousAggsBucketFunction structure as well or keeping it as is for now?

src/catalog.h
```diff
@@ -873,10 +875,17 @@ typedef struct FormData_continuous_agg
...
 	NameData partial_view_name;
 	/*
-	 * Don't access bucket_width directly to determine the width of the bucket.
-	 * Use corresponding procedures instead:
+	 * bucket_width is BUCKET_WIDTH_VARIABLE (see continuous_agg.h) for buckets
+	 * with variable size - monthly buckets, buckets with timezone, etc. For such
+	 * buckets the information about the bucketing function is stored in
+	 * _timescaledb_catalog.continuous_aggs_bucket_function.
+	 *
+	 * When possible, don't access bucket_width directly. Use corresponding
```
davidkohn88, line 883: I'm wondering if we can just replace all of this with some sort of function call, sort of like we do with partitioning time function. So that we aren't calculating any of this ourselves and instead just calling a function that is a function of a certain type?

mkindahl (commented):
Hi Aleksander,

I have a few minor things concerning typos and test coverage, but my main concerns regarding this patch are these:

- There is some usage of double-indirection pointer that does not seem warranted and potentially also has a bug. Please store the data as a list of pointers rather than a list of pointers to pointers.
- Using text for all the schema fields can potentially result in difficult to find bugs and/or security problems and is also harder to fix going forward. It would be good to use proper types for the columns in the new table. If necessary, you can convert this internally, but I do not think that is necessary.
- You should follow the conventions elsewhere in the code, in particular:
  - Schema name + object name for storing function and table references (personally, I think we should store the OID of all such objects instead, but that is a different discussion). 
  - Use PostgreSQL naming conventions, i.e., `Form_...` and `FormData_...` for the data coming from tables.
- You have re-implemented arithmetic functions and serialization functions for handling months and timestamps. It is easier to maintain and less risk of introducing bugs to use existing types and functions to represent the new data, e.g., `Interval` and `Timestamp`.

sql/updates/latest-dev.sql
```diff
@@ -1,2 +1,21 @@
 DROP FUNCTION IF EXISTS recompress_chunk;
 
+
+-- See the comments for ContinuousAggsBucketFunction structure.
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
+  mat_hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
+  experimental bool NOT NULL,
+  name text NOT NULL,
```
mkindahl, line 8:
```suggestion
  name name NOT NULL,
```

tsl/src/continuous_aggs/invalidation.c
```diff
@@ -772,10 +805,18 @@ move_invalidations_from_hyper_to_cagg_log(const CaggInvalidationState *state)
...
+			  all_caggs->mat_hypertable_ids,
+			  lc2,
+			  all_caggs->bucket_widths,
+			  lc3,
+			  all_caggs->bucket_functions)
 	{
 		int32 cagg_hyper_id = lfirst_int(lc1);
 		int64 bucket_width = DatumGetInt64(PointerGetDatum(lfirst(lc2)));
+		const ContinuousAggsBucketFunction *bucket_function =
+			*(const ContinuousAggsBucketFunction **) lfirst(lc3);
```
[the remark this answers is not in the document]
afiskon: Fixed 👍

src/continuous_agg.h
```diff
@@ -62,20 +106,27 @@ typedef struct ContinuousAggMatOptions
...
-	List *bucket_widths;	  /* (int64) Datum */
-	List *max_bucket_widths;  /* (int64) Datum */
+	/* (int32) elements */
+	List *mat_hypertable_ids;
+	/* (int64) Datum elements; stores BUCKET_WIDTH_VARIABLE for variable buckets */
+	List *bucket_widths;
+	/* (int64) Datum elements; stores BUCKET_WIDTH_VARIABLE for variable buckets */
+	List *max_bucket_widths;
+	/* (const ContinuousAggsBucketFunction *) elements; stores NULL for fixed buckets */
+	List *bucket_functions;
```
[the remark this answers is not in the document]
afiskon: Fixed 👍

afiskon:
For the record, @mkindahl and I just had a call and agreed on the following:
* I will address the issues Mats mentioned except the change of the schema
* The plan of changing the schema and merging `continuous_aggs_bucket_function` and `continuous_aggs` tables will be documented in the RFC
* The corresponding part of the RFC will be copied to this PR discussion (e.g. in the description) in order to be transparent before the community
* Since Mats is going on vacation his approval is not needed to merge the PR, any other +1 will suffice

sql/updates/latest-dev.sql
afiskon: I'm marking this as resolved, see https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

tsl/src/continuous_aggs/create.c
```diff
@@ -180,7 +180,8 @@ typedef struct CAggTimebucketInfo
 							/* This should also be the column used by time_bucket */
 	Oid htpartcoltype;
 	int64 htpartcol_interval_len; /* interval length setting for primary partitioning column */
-	int64 bucket_width;			  /*bucket_width of time_bucket */
+	int64 bucket_width;			  /* bucket_width of time_bucket */
+	bool bucket_width_months;	 /* if true, bucket_width stores the amount of months */
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/pre_install/tables.sql
```diff
@@ -280,6 +280,18 @@ CREATE INDEX IF NOT EXISTS continuous_agg_raw_hypertable_id_idx ON _timescaledb_
 
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_agg', '');
 
+-- See the comments for ContinuousAggsBucketFunction structure.
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
+  mat_hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
+  experimental bool NOT NULL,
+  name text NOT NULL,
+  bucket_width text NOT NULL,
+  origin text NOT NULL,
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

src/continuous_agg.h
```diff
@@ -34,9 +34,53 @@ typedef enum ContinuousAggViewType
...
+	 * Schema of the bucketing function.
+	 * Equals TRUE for "timescaledb_experimental", FALSE otherwise.
+	 */
+	bool experimental;
+	/* Name of the bucketing function, e.g. "time_bucket" or "time_bucket_ng" */
+	char *name;
+	/* `bucket_width` argument of the function, e.g. "1 month" */
+	char *bucket_width;
+	/* `origin` argument of the function provided by the user */
+	char *origin;
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

src/continuous_agg.h
```diff
@@ -34,9 +34,53 @@ typedef enum ContinuousAggViewType
...
+ */
+typedef struct ContinuousAggsBucketFunction
+{
+	/*
+	 * Schema of the bucketing function.
+	 * Equals TRUE for "timescaledb_experimental", FALSE otherwise.
+	 */
+	bool experimental;
+	/* Name of the bucketing function, e.g. "time_bucket" or "time_bucket_ng" */
+	char *name;
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/pre_install/tables.sql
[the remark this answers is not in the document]
afiskon: I'm marking this long discussion as resolved, see https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/updates/latest-dev.sql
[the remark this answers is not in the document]
afiskon: I'm marking this long discussion as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

src/continuous_agg.h
```diff
@@ -34,9 +34,53 @@ typedef enum ContinuousAggViewType
...
 typedef struct ContinuousAgg
 {
 	FormData_continuous_agg data;
+
+	/*
+	 * bucket_function is NULL unless the bucket is variable in size,
+	 * e.g. monthly bucket or a bucket with a timezone.
+	 * In this case data.bucket_with stores BUCKET_WIDTH_VARIABLE.
+	 */
+	ContinuousAggsBucketFunction *bucket_function;
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/cagg_utils.sql
```diff
@@ -94,3 +106,17 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
...
 ) RETURNS RECORD AS '@MODULE_PATHNAME@', 'ts_invalidation_process_cagg_log' LANGUAGE C STRICT VOLATILE;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
+    mat_hypertable_id INTEGER,
+    raw_hypertable_id INTEGER,
+    dimtype REGTYPE,
+    window_start BIGINT,
+    window_end BIGINT,
+    mat_hypertable_ids INTEGER[],
+    bucket_widths BIGINT[],
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/cagg_utils.sql
```diff
@@ -94,3 +106,17 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
...
+    max_bucket_widths BIGINT[],
+    bucket_functions TEXT[],
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/pre_install/tables.sql
```diff
@@ -280,6 +280,18 @@ CREATE INDEX IF NOT EXISTS continuous_agg_raw_hypertable_id_idx ON _timescaledb_
 
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_agg', '');
 
+-- See the comments for ContinuousAggsBucketFunction structure.
+CREATE TABLE IF NOT EXISTS _timescaledb_catalog.continuous_aggs_bucket_function(
+  mat_hypertable_id integer PRIMARY KEY REFERENCES _timescaledb_catalog.hypertable (id) ON DELETE CASCADE,
+  experimental bool NOT NULL,
+  name text NOT NULL,
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/pre_install/tables.sql
```diff
@@ -280,6 +280,18 @@ CREATE INDEX IF NOT EXISTS continuous_agg_raw_hypertable_id_idx ON _timescaledb_
...
+  bucket_width text NOT NULL,
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/pre_install/tables.sql
```diff
@@ -280,6 +280,18 @@ CREATE INDEX IF NOT EXISTS continuous_agg_raw_hypertable_id_idx ON _timescaledb_
...
+  origin text NOT NULL,
+  timezone text NOT NULL
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

src/continuous_agg.c
afiskon: I'm marking this as resolved 🤷‍♂️

src/continuous_agg.h
```diff
@@ -34,9 +34,53 @@ typedef enum ContinuousAggViewType
...
+ * - bucket_function_deserialize()
+ *
+ * Serialized data is used as an input of the following procedures:
+ *
+ * - _timescaledb_internal.invalidation_process_hypertable_log()
+ * - _timescaledb_internal.invalidation_process_cagg_log()
+ *
+ * See bucket_functions[] argument.
+ */
+typedef struct ContinuousAggsBucketFunction
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

src/continuous_agg.h
```diff
@@ -34,9 +34,53 @@ typedef enum ContinuousAggViewType
...
+{
+	/*
+	 * Schema of the bucketing function.
+	 * Equals TRUE for "timescaledb_experimental", FALSE otherwise.
+	 */
+	bool experimental;
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

src/continuous_agg.h
```diff
@@ -34,9 +34,53 @@ typedef enum ContinuousAggViewType
...
+	/* Name of the bucketing function, e.g. "time_bucket" or "time_bucket_ng" */
+	char *name;
+	/* `bucket_width` argument of the function, e.g. "1 month" */
+	char *bucket_width;
```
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved, see the summary https://github.com/timescale/timescaledb/pull/3753#issuecomment-991073467

sql/updates/reverse-dev.sql
```diff
@@ -0,0 +1,71 @@
+DO $$
+DECLARE
+    caggs_nr int;
+BEGIN
+    SELECT count(*) FROM _timescaledb_catalog.continuous_agg WHERE bucket_width < 0 INTO caggs_nr;
+    IF caggs_nr > 0 THEN
+        RAISE EXCEPTION 'Downgrade is impossible since % continuous aggregates exist which use variable buckets', caggs_nr
+              USING HINT = 'Remove the corresponding continuous aggregates manually before downgrading';
+    END IF;
```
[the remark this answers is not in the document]
afiskon: 👍 Fixed.

src/continuous_agg.c
```diff
@@ -233,6 +264,68 @@ ts_materialization_invalidation_log_delete(PG_FUNCTION_ARGS)
...
+	}
+
+	/*
+	 * This function should never be called unless we know that the corresponding
+	 * cagg exists and uses a variable-sized bucket. There should be exactly one
+	 * entry in .continuous_aggs_bucket_function catalog table for such a cagg.
+	 */
+	if (count != 1)
+	{
+		ereport(ERROR,
```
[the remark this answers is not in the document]
afiskon: 👍 Done.

sql/updates/reverse-dev.sql
```diff
@@ -1,2 +1,77 @@
+
 DROP PROCEDURE IF EXISTS recompress_chunk;
 DROP FUNCTION IF EXISTS _timescaledb_internal.chunk_status;
+
+DO $$
+DECLARE
+    caggs text[];
+    caggs_nr int;
+BEGIN
+    SELECT array_agg(user_view_schema || '.' || user_view_name) FROM _timescaledb_catalog.continuous_agg WHERE bucket_width < 0 INTO caggs;
```
fabriziomello, line 10:
```suggestion
    SELECT array_agg(format('%I.%I', user_view_schema, user_view_name)) FROM _timescaledb_catalog.continuous_agg WHERE bucket_width < 0 INTO caggs;
```

src/continuous_agg.c
```diff
@@ -263,55 +362,159 @@ ts_continuous_agg_get_all_caggs_info(int32 raw_hypertable_id)
...
+ *
+ * Note that the schema and the name of the function are not serialized. This
+ * is intentional since this information is currently not used for anything.
+ * We can serialize the name of the function as well when and if this would be
+ * necessary.
+ */
+static const char *
+bucket_function_serialize(const ContinuousAggsBucketFunction *bf)
+{
+	const int ver = 1;
```
fabriziomello, line 412: What about have a macro for this?
```c
#define CAGG_BUCKET_FUNCTION_VERSION 1
```

src/continuous_agg.c
```diff
@@ -263,55 +362,159 @@ ts_continuous_agg_get_all_caggs_info(int32 raw_hypertable_id)
...
+	StringInfo str;
+
+	if (bf == NULL)
```
fabriziomello, line 415: Didn't we agreed is better to use [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions)?

mfundul: It's not obligatory in [timescaledb](https://github.com/timescale/timescaledb/blob/master/docs/StyleGuide.md). I use them sometimes but most developers don't.

src/continuous_agg.c
afiskon: 👍

src/continuous_agg.c
afiskon: Yep, that would be more consistent with the rest of the code. Fixed.

PR #3766, opened 2021-10-31
Fix segfault in ts_hist_sfunc

svenklemm:
Since the number of buckets of the histogram function is an argument
to the function call it is possible to initialize the histogram
state with a lower number than actually needed in further calls
leading to a segfault. This patch changes the memory access to use
the number the state was initialized with instead of the number
passed to the call. This also changes the function to error when
the passed number differs from the initialized state.

Fixes: https://github.com/timescale/timescaledb-private/issues/935

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ accidentally triggering the load of a previous DB version.**
 
 ## Unreleased
 
+**Bugfixes**
+* #3766 Fix segfault in ts_hist_sfunc
+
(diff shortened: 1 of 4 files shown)


mfundul (approved):

mkindahl (approved):
LGTM.

Strange that you cannot pass parameters that are only used to initialize the state when doing aggregation. Something like the direct arguments to ordered-set aggregation but without the requirement on ordering.

PR #3767, opened 2021-10-31
Remove namein calls from scankey initialization

svenklemm:
A lot of our scankey initialization when scanning indexes for
a name had a superfluous namein call.  This patch remove those
unnecessary calls.

Disable-check: commit-count

### DIFF
--- a/src/metadata.c
+++ b/src/metadata.c
@@ -24,22 +24,18 @@
 	elog(ERROR, "ts_metadata: no %s function for type %u", inout, typeid);
 
 static Datum
-convert_type(PGFunction func, Datum value, Oid from_type)
+convert_type_to_text(Datum value, Oid from_type)
 {
-	bool value_is_varlena;
-	Oid value_out;
+	bool is_varlena;
+	Oid outfunc;
 
-	getTypeOutputInfo(from_type, &value_out, &value_is_varlena);
+	getTypeOutputInfo(from_type, &outfunc, &is_varlena);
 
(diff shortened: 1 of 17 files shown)


src/metadata.c
```diff
@@ -161,8 +153,7 @@ ts_metadata_insert(Datum metadata_key, Oid key_type, Datum metadata_value, Oid v
 	}
 
 	/* Insert into the catalog table for persistence */
-	values[AttrNumberGetAttrOffset(Anum_metadata_key)] =
-		convert_type_to_name(metadata_key, key_type);
+	values[AttrNumberGetAttrOffset(Anum_metadata_key)] = CStringGetDatum(metadata_key);
```
mfundul, line 156: This is going to crash with ASAN, since it's a `NameData` and `heap_form_tuple()` is going to `memcpy()` it without checking the length. You should have a `palloc0`ed name here.

mfundul (changes requested):

src/metadata.c
svenklemm: changed

mfundul: `  strlcpy(metadata_key, key_data, NAMEDATALEN);` are destination and source in the wrong order?

mfundul: You could use the structure:
```
typedef struct nameData
{
	char		data[NAMEDATALEN];
} NameData;
```

mfundul (approved):

fabriziomello (approved):

PR #3778, opened 2021-11-04
Alter constraints on compressed hypertables

erthalion:
Hi,

From what I see currently it's not possible e.g. to drop a foreign key on a
compressed hypertable (among other DDL limitations). I haven't found any
detailed description on why is it like that in the original commit a399a57, and
few simple experiments show that at least dropping or adding a foreign key
actually works without any extra changes (except allowing those). Are there any
particular reasons to forbid such constraint related AlterTableCmd?

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -255,20 +255,33 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt 
 
 				if (constraint->contype == CONSTR_UNIQUE)
 				{
-					List *segmentby = NIL, *info = NIL;
+					List *allowed = NIL, *info = NIL;
 					ListCell *lc;
+					int i;
 
 					info = ts_hypertable_compression_get(ht->fd.id);
+
+					/* Unique constraints on segmentby columns are supported */
 					foreach (lc, info)
 					{
 						FormData_hypertable_compression *fd = lfirst(lc);
-						if (fd->segmentby_column_index > 0 ||
-							fd->orderby_column_index > 0)
-							segmentby = lappend(segmentby,
-									makeString(fd->attname.data));
+						if (fd->segmentby_column_index > 0)
+							allowed = lappend(allowed,
+											  makeString(NameStr(fd->attname)));
 					}
 
-					if(list_difference(constraint->keys, segmentby) != NIL)
+					/*
+					 * A unique constraint has to include dimention columns,
+					 * otherwise index creation is not possible.
+					 */
+					for (i = 0; i < ht->space->num_dimensions; i++)
+					{
+						Dimension *dim = &ht->space->dimensions[i];
+						allowed = lappend(allowed,
+										  makeString(NameStr(dim->fd.column_name)));
+					}
+
+					if(list_difference(constraint->keys, allowed) != NIL)
 						ereport(ERROR,
 								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 								 errmsg("operation not supported on hypertables "
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -235,6 +235,10 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt 
 			case AT_AddColumn:	 /* this is passed down */
 			case AT_ColumnDefault: /* this is passed down */
 			case AT_DropColumn:	/* this is passed down */
+
+			case AT_DropConstraint:
+			case AT_AddConstraint:
+			case AT_AlterConstraint:
 #if PG14_GE
 			case AT_ReAddStatistics:
 			case AT_SetCompression:
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -55,6 +55,7 @@
 #include "hypercube.h"
 #include "hypertable.h"
 #include "hypertable_cache.h"
+#include "hypertable_compression.h"
 #include "hypertable_data_node.h"
 #include "dimension_vector.h"
 #include "indexing.h"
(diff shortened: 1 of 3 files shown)


svenklemm:
I think allowing DROP CONSTRAINT and ALTER CONSTRAINT is fine but ADD CONSTRAINT is not simple yes/no. But all 3 of those commands are probably blocked to be symmetrical.
Since we don't currently support UNIQUE constraints that include non-segmentby columns the range of allowed ADD CONSTRAINT commands need to reflect that.

erthalion:
> On Thu, Nov 04, 2021 at 02:04:37PM -0700, Sven Klemm wrote:
> I think allowing DROP CONSTRAINT and ALTER CONSTRAINT is fine but ADD CONSTRAINT is not simple yes/no. But all 3 of those commands are probably blocked to be symmetrical.
> Since we don't currently support UNIQUE constraints that include non-segmentby columns the range of allowed ADD CONSTRAINT commands need to reflect that.

Yes, makes sense. I did a quick experiment (without writing any tests
yet), but not sure if the result breaks any modularity concerns. Is
there a better way to extract information about "segment by" columns?

src/process_utility.c
```diff
@@ -233,11 +234,51 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
...
+				{
+					List *segmentby = NIL, *info = NIL;
+					ListCell *lc;
+
+					info = ts_hypertable_compression_get(ht->fd.id);
+					foreach (lc, info)
+					{
+						FormData_hypertable_compression *fd = lfirst(lc);
+						if (fd->segmentby_column_index > 0 ||
+							fd->orderby_column_index > 0)
```
svenklemm, line 264:
```suggestion
						if (fd->segmentby_column_index > 0)
```
Unique constraints including orderby columns are not supported

erthalion: Right, I've confused it with the dimension columns, which have to be included here, otherwise creating a unique constraint still fails later on in indexing.

src/process_utility.c
```diff
@@ -233,11 +234,64 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
...
+
+					info = ts_hypertable_compression_get(ht->fd.id);
+
+					/* Unique constraints on segmentby columns are supported */
+					foreach (lc, info)
+					{
+						FormData_hypertable_compression *fd = lfirst(lc);
+						if (fd->segmentby_column_index > 0)
+							allowed = lappend(allowed,
+											  makeString(NameStr(fd->attname)));
```
akuzm, line 268: Codecov complains about this line not being covered (you should see the annotation below). Could you add a test for this?

src/process_utility.c
```diff
@@ -233,11 +234,64 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
...
+					{
+						Dimension *dim = &ht->space->dimensions[i];
+						allowed = lappend(allowed,
+										  makeString(NameStr(dim->fd.column_name)));
+					}
+
+					if(list_difference(constraint->keys, allowed) != NIL)
+						ereport(ERROR,
+								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+								 errmsg("operation not supported on hypertables "
```
akuzm, line 285: The error message is incorrect now (we do support it under some conditions). Probably the error message should say this and have a hint that lists the disallowed columns or someting.

src/process_utility.c
```diff
@@ -233,11 +234,64 @@ check_alter_table_allowed_on_ht_with_compression(Hypertable *ht, AlterTableStmt
...
+										"that have compression enabled")));
+
+				}
+				else
+					continue;
+
+				break;
```
akuzm, line 292: Again, a codecov warning about this line. This is not important but having less of these warnings makes it easier to notice the important ones. Looks like the 'else/continue' can be removed.

erthalion:
> On Mon, Nov 22, 2021 at 05:21:51AM -0800, Alexander Kuzmenkov wrote:
> @akuzm commented on this pull request.
>
>
>
> > +
> +				if (constraint->contype == CONSTR_UNIQUE)
> +				{
> +					List *allowed = NIL, *info = NIL;
> +					ListCell *lc;
> +					int i;
> +
> +					info = ts_hypertable_compression_get(ht->fd.id);
> +
> +					/* Unique constraints on segmentby columns are supported */
> +					foreach (lc, info)
> +					{
> +						FormData_hypertable_compression *fd = lfirst(lc);
> +						if (fd->segmentby_column_index > 0)
> +							allowed = lappend(allowed,
> +											  makeString(NameStr(fd->attname)));
>
> Codecov complains about this line not being covered (you should see the annotation below). Could you add a test for this?

Yes, makes sense. I've added few more tests around it, which
unfortunately turned out to be more verbose in the diff due to increment
in the chunk number, I hope it's fine.

erthalion:
> On Mon, Nov 22, 2021 at 05:22:42AM -0800, Alexander Kuzmenkov wrote:
> @akuzm commented on this pull request.
>
>
>
> > +
> +					/*
> +					 * A unique constraint has to include dimention columns,
> +					 * otherwise index creation is not possible.
> +					 */
> +					for (i = 0; i < ht->space->num_dimensions; i++)
> +					{
> +						Dimension *dim = &ht->space->dimensions[i];
> +						allowed = lappend(allowed,
> +										  makeString(NameStr(dim->fd.column_name)));
> +					}
> +
> +					if(list_difference(constraint->keys, allowed) != NIL)
> +						ereport(ERROR,
> +								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
> +								 errmsg("operation not supported on hypertables "
>
> The error message is incorrect now (we do support it under some conditions). Probably the error message should say this and have a hint that lists the disallowed columns or something.

I've changed it to "UNIQUE constraints on columns not included in
compress_segmentby are not supported" with hinting on not allowed
columns, which should probably be better. Do you have better suggestion?

akuzm (approved):

svenklemm (changes requested):
I think this needs some further changes, while you added a check for unique constraint, this PR also allows other constraints but those are not checked in the same way e.g. CHECK, FOREIGN KEY which do have the same constraints.

erthalion:
> On Thu, Jan 27, 2022 at 10:14:30AM -0800, Sven Klemm wrote:
> @svenklemm requested changes on this pull request.
>
> I think this needs some further changes, while you added a check for unique constraint, this PR also allows other constraints but those are not checked in the same way e.g. CHECK, FOREIGN KEY which do have the same constraints.

Do you mean TS doesn't support a CHECK/FOREIGN KEY constraints on
non-segmentby columns for compressed hypertables? To make it a bit
easier, could you point me the documentation where I can find a
comprehensive list of what type of constraints are supported in this
case?

Issue #3784, opened 2021-11-05
DISTINCT ON queries return incorrect results on Distributed Hypertables


tylerfontaine:
**Relevant system information:**
 - OS: [e.g. Ubuntu 16.04, Windows 10 x64, etc]
 - PostgreSQL version (output of `postgres --version`): `PostgreSQL 13.4 (Ubuntu 13.4-4.pgdg21.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 10.3.0-1ubuntu1) 10.3.0, 64-bit` 
 - TimescaleDB version (output of `\dx` in `psql`): 2.5.0
 - Installation method: Timescale Cloud
 - 
**Describe the bug**
When issuing DISTINCT ON queries with an ORDER BY against a distributed hypertable, it produces an incorrect result.

When reformatting the same query to do a LIMIT 1, it returns the correct query.

**To Reproduce**
On a multi-node system (Mine has 1 AN and 3 DNs): 
I was not able to reproduce this with a small amount of data, so here is a CSV of 100,000 rows which reliably reproduces the incorrect behavior.
[distinct_on_repro.csv](https://github.com/timescale/timescaledb/files/7488570/distinct_on_repro.csv)

Create two identical tables. Make one distributed, one a regular hypertable:

```sql
CREATE TABLE test_distributed (
    time timestamp without time zone NOT NULL,
    id text,
    val integer,
    label text
);
select create_distributed_hypertable('test_distributed', 'time');

CREATE TABLE test_hypertable (
    time timestamp without time zone NOT NULL,
    id text,
    val integer,
    label text
);
select create_hypertable('test_hypertable', 'time', 'label', 3);
```
Import the CSV above into both tables.

You can observe the different results with these queries:

```sql
select distinct on (id) id, val, time from test_distributed
where id in ('a')
order by id asc, time desc;

select distinct on (id) id, val, time from test_hypertable
where id in ('a')
order by id asc, time desc;

select id, val, time from test_distributed
where id in ('a')
order by id asc, time desc limit 1;

select id, val, time from test_hypertable
where id in ('a')
order by id asc, time desc limit 1;
```

```
tsdb=> select distinct on (id) id, val, time from test_distributed
where id in ('a')
order by id asc, time desc;
 id | val  |        time
----+------+---------------------
 a  | 7524 | 2021-10-13 03:10:24
(1 row)


tsdb=> select distinct on (id) id, val, time from test_hypertable
where id in ('a')
[...]

PR #3797, opened 2021-11-10
Fix DISTINCT ON pushdown

akuzm:
Fixes https://github.com/timescale/timescaledb/issues/3784

### DIFF
--- a/tsl/src/fdw/deparse.c
+++ b/tsl/src/fdw/deparse.c
@@ -179,7 +179,7 @@ static void printRemoteParam(int paramindex, Oid paramtype, int32 paramtypmod,
 							 deparse_expr_cxt *context);
 static void printRemotePlaceholder(Oid paramtype, int32 paramtypmod, deparse_expr_cxt *context);
 static void deparseSelectSql(List *tlist, bool is_subquery, List **retrieved_attrs,
-							 deparse_expr_cxt *context);
+							 deparse_expr_cxt *context, List *pathkeys);
 static void deparseLockingClause(deparse_expr_cxt *context);
 static void appendOrderByClause(List *pathkeys, deparse_expr_cxt *context);
 static void appendLimit(deparse_expr_cxt *context, List *pathkeys);
@@ -771,7 +771,7 @@ deparseSelectStmtForRel(StringInfo buf, PlannerInfo *root, RelOptInfo *rel, List
 	context.sca = sca;
 
 	/* Construct SELECT clause */
-	deparseSelectSql(tlist, is_subquery, retrieved_attrs, &context);
+	deparseSelectSql(tlist, is_subquery, retrieved_attrs, &context, pathkeys);
 
 	/* Construct FROM and WHERE clauses */
 	deparseFromExpr(remote_where, &context);
@@ -822,7 +822,7 @@ deparseSelectStmtForRel(StringInfo buf, PlannerInfo *root, RelOptInfo *rel, List
  * deparsing which happens later is good enough
  */
 static void
-deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
+deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context, List *pathkeys)
 {
 	PlannerInfo *root = context->root;
 	Query *query = root->parse;
@@ -865,6 +865,60 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
 		/* We only allow constants apart from vars, but we ignore them */
 		else if (!IsA(tle->expr, Const))
 			return;
+	}
+
+	if (query->hasDistinctOn)
+	{
+		/*
+		 * Pushing down DISTINCT ON is more complex than plain DISTINCT.
+		 * The DISTINCT ON columns must be a prefix of the ORDER BY columns.
+		 * Without this, the DISTINCT ON would return an unpredictable row
+		 * each time. There is a diagnostic for the case where the ORDER BY
+		 * clause doesn't match the DISTINCT ON clause, so in this case we
+		 * would get an error on the data node. There is no diagnostic for
+		 * the case where the ORDER BY is absent, so in this case we would
+		 * get a wrong result.
+		 * The remote ORDER BY clause is created from the pathkeys of the
+		 * corresponding relation. If the DISTINCT ON columns are not a prefix
+		 * of these pathkeys, we cannot push it down.
+		 */
+		ListCell *distinct_cell, *pathkey_cell;
+		forboth (distinct_cell, query->distinctClause, pathkey_cell, pathkeys)
+		{
+			SortGroupClause *sgc = lfirst_node(SortGroupClause, distinct_cell);
+			TargetEntry *tle = get_sortgroupclause_tle(sgc, query->targetList);
+
+			PathKey *pk = lfirst_node(PathKey, pathkey_cell);
+			EquivalenceClass *ec = pk->pk_eclass;
+
+			/*
+			 * The find_ec_member_matching_expr() has many checks that don't seem
+			 * to be relevant here. Enumerate the pathkey EquivalenceMembers by
+			 * hand and find the one that matches the DISTINCT ON expression.
+			 */
+			ListCell *ec_member_cell;
+			foreach (ec_member_cell, ec->ec_members)
+			{
+				EquivalenceMember *ec_member = lfirst_node(EquivalenceMember, ec_member_cell);
+				if (equal(ec_member->em_expr, tle->expr))
+					break;
+			}
+
+			if (ec_member_cell == NULL)
+			{
+				/*
+				 * Went through all the equivalence class members and didn't
+				 * find a match.
+				 */
+				return;
+			}
+		}
+
+		if (pathkey_cell == NULL && distinct_cell != NULL)
+		{
+			/* Ran out of pathkeys before we matched all the DISTINCT ON columns. */
+			return;
+		}
 	}
 
 	/* If there are no varno entries in the distinctClause, we are done */
@@ -937,7 +991,8 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
  * Read prologue of deparseSelectStmtForRel() for details.
  */
 static void
-deparseSelectSql(List *tlist, bool is_subquery, List **retrieved_attrs, deparse_expr_cxt *context)
+deparseSelectSql(List *tlist, bool is_subquery, List **retrieved_attrs, deparse_expr_cxt *context,
+				 List *pathkeys)
 {
 	StringInfo buf = context->buf;
 	RelOptInfo *foreignrel = context->foreignrel;
@@ -981,7 +1036,7 @@ deparseSelectSql(List *tlist, bool is_subquery, List **retrieved_attrs, deparse_
 		Relation rel = table_open(rte->relid, NoLock);
 
 		if (root->parse->distinctClause != NIL)
-			deparseDistinctClause(buf, context);
+			deparseDistinctClause(buf, context, pathkeys);
 
 		deparseTargetList(buf,
 						  rte,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ accidentally triggering the load of a previous DB version.**
 * #3739 Fix compression policy on tables using INTEGER
 * #3766 Fix segfault in ts_hist_sfunc
 * #3789 Fix time_bucket comparison transformation
+* #3784 Fix DISTINCT ON queries returning incorrect results on distributed hypertables
 * #3801 Fail size utility functions when data nodes do not respond
 
 **Thanks**
--- a/tsl/test/shared/expected/dist_distinct_pushdown.out
+++ b/tsl/test/shared/expected/dist_distinct_pushdown.out
@@ -1,0 +1,178 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+-- Test DISTINCT ON pushdown.
+-- The case with LIMIT serves as a reference.
+select ts, id from distinct_on_hypertable order by id, ts desc limit 1;
+            ts            | id 
+--------------------------+----
+ Fri Jan 01 03:47:41 2021 |  0
+(1 row)
+
+select ts, id from distinct_on_distributed order by id, ts desc limit 1;
+            ts            | id 
+--------------------------+----
+ Fri Jan 01 03:47:41 2021 |  0
+(1 row)
+
+-- DISTINCT ON should match the above LIMIT for the first id.
+select distinct on (id) ts, id from distinct_on_hypertable order by id, ts desc;
+            ts            | id 
+--------------------------+----
+ Fri Jan 01 03:47:41 2021 |  0
+ Fri Jan 01 03:47:38 2021 |  1
+ Fri Jan 01 03:47:39 2021 |  2
+ Fri Jan 01 03:47:40 2021 |  3
+(4 rows)
+
+select distinct on (id) ts, id from distinct_on_distributed order by id, ts desc;
+            ts            | id 
+--------------------------+----
+ Fri Jan 01 03:47:41 2021 |  0
+ Fri Jan 01 03:47:38 2021 |  1
+ Fri Jan 01 03:47:39 2021 |  2
+ Fri Jan 01 03:47:40 2021 |  3
+(4 rows)
+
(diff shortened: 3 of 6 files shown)


tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,59 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
...
+	if (query->hasDistinctOn)
+	{
+        /*
+         * Pushing down DISTINCT ON is more complex than plain DISTINCT.
+         * The former requires a matching ORDER BY clause, or else the result is
+         * undefined. There is a diagnostic for the case where the ORDER BY
+         * clause doesn't match the DISTINCT ON clause, so in this case we would
+         * get an error on the data node. There is no diagnostic for the case
+         * where the ORDER BY is absent, so in this case we would get a wrong
+         * result.
```
erimatnor, line 879: If the ORDER BY is absent, can't we just add one? The ORDER BY should not matter for anything else, unless I am missing something.

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,59 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
...
+         * The remote ORDER BY clause is created from the pathkeys of the
+         * corresponding relation. If these pathkeys do not match the DISTINCT ON
+         * columns, we cannot push it down.
```
erimatnor, line 882: Note that we typically add both ordered and unordered paths, so there should be a case where we can push down the DISTINCT ON and then do s Sort above it, either on the DNs or the AN.

I think this is about costing, and I think there is currently no costing implemented for pushing down DISTINCT and DISTINCT ON. This seems to be missing.

Pusing down DISTINCT means we retrieve less tuples over the network, which should always be a huge cost saving (unless all tuples are already distinct). I think we need to account for this in cost estimation when we create paths, and then let the planner decide when it makes sense to use the push-down paths or the non-push down paths.

erimatnor (commented):
I think there's something fundamentally missing from how DISTINCT push-down is implemented. Currently, there's only the deparsing implemented, but there is no costing or paths created. I believe this was missing also from the previous PR that implemented push down.

So, it looks like there is actually no planner decision involved where we weigh costing of different paths---some that push-down distinct and some that don't.

The way this _should_ be implemented, AFAIK, is to handle the UPPERREL_DISTINCT event in `GetForeignUpperPaths` (FDW hook) or `create_upper_paths_hook`  (standard extension hook). Then we should add DISTINCT paths for the foreign/data node scan rels, including costing.

A DISTINCT push down should be beneficial in terms of costing since it reduces the amount of tuples sent over the network (unless all tuples are already distinct).

tsl/src/fdw/deparse.c
akuzm: If it's _always_ beneficial (which I agree on), we don't need the costing and should just do it unconditionally.

tsl/src/fdw/deparse.c
akuzm: I think so, if we have a DISTINCT ON, we'll have to sort accordingly anyway, and better do it on the data nodes. This is a further optimization that should be done somewhere at the planning stage, not at deparsing. This check I added just restores correctness, preventing us from generating an obviously wrong query. We might decide to switch it to an assertion later.

akuzm:
>I think there's something fundamentally missing from how DISTINCT push-down is implemented.

I agree in general. It needs a proper planner support to ensure that the DISTINCT ON is always pushed down, and the underlying paths on data nodes is sorted appropriately. Probably this should be unconditional because it's always better and we don't need to compare costs each time to know that.

However, this is a high-priority user-facing bug, so I suggest first fixing the obviously broken part and only then thinking how we could optimize/implement it properly, in another PR.

erimatnor:
> > I think there's something fundamentally missing from how DISTINCT push-down is implemented.
> 
> I agree in general. It needs a proper planner support to ensure that the DISTINCT ON is always pushed down, and the underlying paths on data nodes is sorted appropriately. Probably this should be unconditional because it's always better and we don't need to compare costs each time to know that.
> 
> However, this is a high-priority user-facing bug, so I suggest first fixing the obviously broken part and only then thinking how we could optimize/implement it properly, in another PR.

We could go with a quick fix, let's discuss separately.

But I suspect that the planning is a bit wonky if we don't do this via the proper planner Paths. Unless I am mistaken, the planner will run another DISTINCT (Unique) node on the access node since there is no understanding at the planner level that a DISTINCT has already been executed at the remote side. While that will still work, it is suboptimal and somewhat weird from a planning perspective.

akuzm:
> We could go with a quick fix, let's discuss separately.

Sure. I intended this PR to be this quick fix. Note that it's not a hack, it's actually improving correctness, even if the overall approach we have now is somewhat weird.

> But I suspect that the planning is a bit wonky if we don't do this via the proper planner Paths. Unless I am mistaken, the planner will run another DISTINCT (Unique) node on the access node since there is no understanding at the planner level that a DISTINCT has already been executed at the remote side. While that will still work, it is suboptimal and somewhat weird from a planning perspective.
> 
> I guess in some cases, the AN Unique node has to be there in any case when uniqueness cannot be guaranteed across nodes (e.g., when the DISTINCT is not on a space partition key).

This is the generally correct thing to do, unless the DISTINCT ON key is proven not to repeat across shards (e.g. it's the sharding key). Proving this is even a further level of optimization.

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,59 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
 			return;
 	}
 
+	if (query->hasDistinctOn)
+	{
+		/*
+		 * Pushing down DISTINCT ON is more complex than plain DISTINCT.
+		 * The former requires a matching ORDER BY clause, or else the result is
+		 * undefined. There is a diagnostic for the case where the ORDER BY
```
svenklemm, line 875: Hmm DISTINCT ON without ORDER BY has the ordering undefined like any other query without ORDER BY

akuzm: I reworded the comment.

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,60 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
 			return;
 	}
 
+	if (query->hasDistinctOn)
+	{
+		/*
+		 * Pushing down DISTINCT ON is more complex than plain DISTINCT.
+		 * The DISTINCT ON columns must be a prefix of the ORDER BY columns.
```
erimatnor, line 874:
```suggestion
		 * The DISTINCT ON columns must be the prefix of the ORDER BY columns.
```

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,60 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
...
+		int next_pathkey_index = 0;
+		foreach (l, query->distinctClause)
+		{
+			SortGroupClause *sgc = lfirst_node(SortGroupClause, l);
+			TargetEntry *tle = get_sortgroupclause_tle(sgc, query->targetList);
+
+			if (next_pathkey_index >= list_length(pathkeys))
+			{
+				return;
+			}
```
erimatnor, line 894:
```suggestion
			if (next_pathkey_index >= list_length(pathkeys))
				return;
```

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,60 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
...
+				{
+					matching_em = ec_member;
+					break;
+				}
+			}
+
+			if (matching_em == NULL)
+			{
+				return;
+			}
```
erimatnor, line 920:
```suggestion
			if (matching_em == NULL)
				return;
```

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,60 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
...
+
+			PathKey *pk = list_nth(pathkeys, next_pathkey_index);
+			EquivalenceClass *ec = pk->pk_eclass;
+			++next_pathkey_index;
+
+			/*
+			 * The find_ec_member_matching_expr() has many checks that don't seem
+			 * to be relevant here. Enumarete the pathkey EquivalenceMembers by
```
erimatnor, line 902:
```suggestion
			 * to be relevant here. Enumerate the pathkey EquivalenceMembers by
```

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,60 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
...
+		{
+			SortGroupClause *sgc = lfirst_node(SortGroupClause, l);
+			TargetEntry *tle = get_sortgroupclause_tle(sgc, query->targetList);
+
+			if (next_pathkey_index >= list_length(pathkeys))
+			{
+				return;
+			}
+
+			PathKey *pk = list_nth(pathkeys, next_pathkey_index);
```
erimatnor, line 896: instead of keeping an extra index that you increment, can you loop both lists at the same time with `forboth` instead of `foreach`? It seems that would simplify the code and make it more clear that the two lists are iterated in lockstep.

tsl/src/fdw/deparse.c
```diff
@@ -867,6 +867,60 @@ deparseDistinctClause(StringInfo buf, deparse_expr_cxt *context)
...
+			ListCell *ec_member_cell;
+			foreach (ec_member_cell, ec->ec_members)
+			{
+				EquivalenceMember *ec_member = lfirst_node(EquivalenceMember, ec_member_cell);
+				if (equal(ec_member->em_expr, tle->expr))
+				{
+					matching_em = ec_member;
+					break;
+				}
+			}
```
erimatnor, line 915: I suggest moving this to an appropriately named function, similar to `find_ec_member_matching_expr()` but without the extra functionality that you refer to above.

erimatnor (approved):
Approving, but with nits and suggestions. This is also approved with the understanding that this is a quickfix, but we need to refactor DISTINCT handling to use proper upperrel planning via `UPPERREL_DISTINCT`, including cost estimation, etc.

Also remember to add an appropriate commit message before merging. Usually, it is good to add that as part of the original PR as that helps a reviewer understand the context of the PR and what to expect and look for when reviewing the code.

tsl/test/expected/dist_distinct_pushdown.out
```diff
@@ -0,0 +1,132 @@
...
+select create_distributed_hypertable('test_distributed', 'ts');
+NOTICE:  adding not-null constraint to column "ts"
+ create_distributed_hypertable 
+-------------------------------
+ (2,public,test_distributed,t)
+(1 row)
+
+insert into test_distributed select * from test_hypertable;
+-- The case with LIMIT serves as a reference.
+select ts, id from test_hypertable order by id, ts desc limit 1;
```
erimatnor, line 51: It would be nice to see some explains here as well, especially to validate the remote SQL.

tsl/src/fdw/deparse.c
akuzm: ORDER BY has many prefixes of different length, and DISTINCT ON has to be one of these prefixes. That's an "a". We know that it is "the" particular prefix that has matching length, but that's not specified anywhere in the context or the sentence itself. In the next sentence, we could write "the prefix", because now we would be talking about the particular one that matches DISTINCT ON.

tsl/src/fdw/deparse.c
akuzm: I'm going to fix this and I know that this is postgres style we have to follow, but: this is widely considered a bad practice (maybe you remember the Apple's `goto fail;`). I personally tend to have a lot of problems with this while editing the postgres code, because I forget to add the braces, in the same `goto fail` style. Unfortunately, clang-format can't fix this, and clang-tidy can only add braces, not remove.

tsl/src/fdw/deparse.c
akuzm: It's just one loop with one if. Even if I needed it somewhere else, I'd probably just write it anew. Too small/meaningless for a function.

tsl/src/fdw/deparse.c
akuzm: Good idea, it's more concise this way.

tsl/test/expected/dist_distinct_pushdown.out
akuzm: Added.

tsl/src/fdw/deparse.c
erimatnor: A function name also conveys what this code block does, so there's a readability aspect of trying to understand the code by "reading" it using aptly named functions as opposed to just doing it for code reuse purposes.

For example, that function you referred to called `find_ec_member_matching_expr()` made it immediately clear what happened here just by reading the name, while it is non-obvious from the code block what it does.

svenklemm (commented):
Why are the tests not in dist_hypertable or dist_query instead of adding new one?

akuzm:
> Why are the tests not in dist_hypertable or dist_query instead of adding new one?

I would prefer to increase the granularity of our tests. This is easier for: merging (less conflicts), running in parallel, disabling a particular test, reading the test output diff. For example, IIRC we now have several very big tests disabled for PG12, and who knows what actually fails there and what new failures we are introducing.

tsl/src/fdw/deparse.c
akuzm: There's a comment about this. It's more readable than a function name, and also I don't have to jump between screens if I want to see what is actually done. And I will want to see it, because the function names lie even more than the comments do. So extracting small functions is a net readability decrease, at least for my code reading habits.

> that function you referred to called find_ec_member_matching_expr() made it immediately clear what happened here just by reading the name

Yeah, and it looks like it's suitable for our use case, but in fact it's not, because it also does three unrelated checks useful for some other case. That's exactly what I'm talking about, there's no way to encode stuff like this into the function name, you have to read the comments/code anyway.

svenklemm:
> > Why are the tests not in dist_hypertable or dist_query instead of adding new one?
> 
> I would prefer to increase the granularity of our tests. This is easier for: merging (less conflicts), running in parallel, disabling a particular test, reading the test output diff. For example, IIRC we now have several very big tests disabled for PG12, and who knows what actually fails there and what new failures we are introducing.

Not sure where that claim that we don't run certain tests on PG12 is coming from but it is not true. All tests are run on 12.9, most of the isolation tests are disabled for 12.0 because there was an upstream output format change but we run them on latest released PG12. If you wanna to make more granular tests then you should use the shared tests instead of regresscheck-t. regresscheck-t has considerable overhead in the form of database creation which is not true for regresscheck-shared.

erimatnor:
@akuzm Please also add CHANGELOG entries and thanks.

akuzm:
> Not sure where that claim that we don't run certain tests on PG12 is coming from but it is not true. All tests are run on 12.9, most of the isolation tests are disabled for 12.0 because there was an upstream output format change but we run them on latest released PG12.

Probably I was thinking about the ignored tests, 6 of them for PG12. There are also 2.5k lines of regression diff, so not sure if making them more granular would help. (looking here https://github.com/timescale/timescaledb/runs/4189824270?check_suite_focus=true)

> If you wanna to make more granular tests then you should use the shared tests instead of regresscheck-t. regresscheck-t has considerable overhead in the form of database creation which is not true for regresscheck-shared.

Thanks for the pointer. Looks like the table creation must happen in the setup script, or else the chunk ids in e.g. remote SQL will be different each time. For many tests, the creation of the table is the heaviest parts, so this limits our ability to parallelize. For this one it's 0.4 s out of the total 0.6 ms the test runs. Still, running it as a separate test is much slower:
```
start  duration name
53.307 0.4      initdb & postmaster
55.668 1.2      roles & dbs before test script
56.708 1.3      test script, the setup part
57.990 0.6      main test (table creation is 0.4 s)
58.594
```
Of course, running as a part of a giant test may be as fast as shared, but looks like it's not sustainable in the long run.

akuzm:
Shared also has bad usability for running a single test manually, because you have to create all these tables for all the tests you don't need.

akuzm:
> @akuzm Please also add CHANGELOG entries and thanks.

Added the changelog entry. The issue was reported by our tech support, we don't mention this in Thanks section as far as I can see.

nikkhils (commented):
LGTM

akuzm:
> LGTM

Looks like you chose "Comment" instead of "Approve" so it's not counted as approval :)

pmwkaa (approved):
Looks good :)

PR #3819, opened 2021-11-17
Fix reading garbage value from TSConnectionError

afiskon:
clang-tidy reports that after eed2810d under certain conditions
remote_connection_error_elog() macro can read from the memory that was not
initialized. Make sure this can't happen.

### DIFF
--- a/tsl/src/remote/connection.c
+++ b/tsl/src/remote/connection.c
@@ -1846,6 +1846,8 @@ remote_connection_cancel_query(TSConnection *conn)
 	if (!conn)
 		return true;
 
+	memset(&err, 0, sizeof(TSConnectionError));
+
 	/*
 	 * Catch exceptions so that we can ensure the status is IDLE after the
 	 * cancel operation even in case of errors being thrown. Note that we


afiskon:
clang-tidy report https://pastebin.com/3pShir2T

akuzm (approved):
Oh, so this is in the very fresh code from @pmwkaa  . Maybe zero-initializing the error should be the responsibility of fill_result_error? Not sure.

We should enable clang-tidy in CI, maybe with a limited set of warnings and then gradually add them.

afiskon:
@akuzm 👍 Currently we have only this single warning.

tsl/src/remote/connection.c
```diff
@@ -289,7 +289,15 @@ fill_result_error(TSConnectionError *err, int errcode, const char *errmsg, const
 	const char *sqlstate;
 
 	if (NULL == err || NULL == res || NULL == entry)
+	{
+		if (NULL != err)
+		{
+			/* err should be filled unless we return true */
```
mfundul, line 295: Does this function ever return `true`?

afiskon: Good point, thanks. I changed the comment.

mfundul: On that point, if it always returns false, maybe it should become `void`? Maybe it needs some refactoring.

tsl/src/remote/connection.c
```diff
@@ -289,7 +289,15 @@ fill_result_error(TSConnectionError *err, int errcode, const char *errmsg, const
 	const char *sqlstate;
 
 	if (NULL == err || NULL == res || NULL == entry)
+	{
+		if (NULL != err)
```
fabriziomello, line 293: I always have a mind break when read this type of comparison between value and variable, why not use just:

```C
if (!err || !res || !entry)
{
    if (err)
...
```

I think we don't have a clear definition for this but to me is weird and @svenklemm also complained about it in other PR that I worked.

erimatnor: I guess this is a preference of style. 

While I agree that the suggested change is somewhat easier on the eye, it is also somewhat ambiguous. 

By explicitly comparing with `NULL` one knows that the variables are pointers without having to look at the variable definition (which might be outside this context, especially true in code-review snippets).

Let's take the variable `err` for example. From just reading the code, it could easily be an `int` , a `boolean`, a pointer, or even an `enum` type. So, by writing:

```
if (NULL == err)
```
I immediately know that `err` is a pointer and this is a `NULL` check. Similarly, if I see this:

```
if (0 == err)
```
I know err is an integer and the non-error condition is 0. Or I see this to identify enum types"

```
if (NO_ERROR == err)
```

However, if I see this:

```
if (!err)
```

then I don't know the type of `err` (is it int, pointer, bool, or even an enum type?) and I don't know if, e.g., 0 is the non-error condition. 

Just explaining some reasoning here. Don't feel super-strongly about it though, and I am sure we have a mix of styles in the code.

afiskon: Personally, I'm fine with any code style. In this case, I just followed the style that was already in use.

fabriziomello: Understood your point. Can we at least invert the comparison??? 

```C
if (err == NULL)
```

IMHO is more clear compare a variable with a value.

erimatnor: There's actually a point to this too :)

writing 

```
if (NULL == err)
```

protects against accidental assignment, e.g.:

```
if (err = NULL)
```

Most compilers today detect that case, however.

afiskon: @fabriziomello this is very a common approach in C world 🤷‍♂️ Again, I simply followed the code style that was chosen in the project (or this particular part of it) a long time before I joined the project. If you believe the code style can be improved, I'm sure we all are open to suggestions. However, this should be discussed not in a separate pull request.

pmwkaa:
@afiskon could you please share the link to the clang error about this? also I think it would make more sense to fix this at the moment of variable definition instead of actual usage

afiskon:
@pmwkaa the link is attached above, in the second message from the top ;) To clarify, it may be a false-positive, I didn't invest too much time into researching this. But considering this is the only warning from clang-tidy we currently have I suggest silencing it anyway (other warnings can be useful, and this one can hide them).

I'm fine with calling memset() in the place where the variable is defined. However it will cause some performance degradation, thus I choose to do it in the place that introduced the problem.

pmwkaa:
> @pmwkaa the link is attached above, in the second message from the top ;) To clarify, it may be a false-positive, I didn't invest too much time into researching this. But considering this is the only warning from clang-tidy we currently have I suggest silencing it anyway (other warnings can be useful, and this one can hide them).
> 
> I'm fine with calling memset() in the place where the variable is declared. However it will cause some performance degradation, thus I choose to do it in the place that introduced the problem.

Sorry, missed that :) The problem with fixing in place, is that it actually might not solve the problem if this struct being used somewhere else on the path, etc

afiskon:
@pmwkaa OK here is the updated version of the patch. BTW do you think we should initialize TSConnectionError in the rest of the code, just to be on the safe side?

tsl/src/remote/connection.c
afiskon: Aaaand it's gone http://www.failure.at/wp-content/uploads/2013/07/4387467_460s.jpg

mfundul: @fabriziomello source: https://en.wikipedia.org/wiki/Yoda_conditions

mfundul:
https://github.com/timescale/timescaledb/pull/3819#discussion_r751376143

any comments on the above?

afiskon:
@mfundul You are probably right, but I don't feel like making refactorings in this PR, sorry :)

mfundul:
> 
> 
> @mfundul You are probably right, but I don't feel like making refactorings in this PR, sorry :)

Just making a comment in case there was an error somewhere. Maybe people will notice something.

pmwkaa (approved):

PR #3820, opened 2021-11-17
Treat clang-tidy warnings as errors

afiskon:
This change makes the build fail on macos-* if clang-tidy detects any bugs.
This improves our process in two ways.

Firstly, currently we execute clang-tidy on macos-* but ignore the warnings.
As a result we waste CPU resources. Also, somebody who uses MacOS has to fix
the warnings manually while this could be done by the person who introduced
corresponding changes.

Secondly, running clang-tidy on a laptop is relatively expensive. This change
allows to disable clang-tidy locally and speed up the development process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -507,6 +507,9 @@ endif()
 option(LINTER "Enable linter support using clang-tidy (ON when using clang)"
        ${LINTER_DEFAULT})
 
+set(LINTER_STRICT_DEFAULT OFF)
+option(LINTER_STRICT "Treat linter warnings as errors" ${LINTER_STRICT_DEFAULT})
+
 if(LINTER)
   find_program(
     CLANG_TIDY clang-tidy
@@ -516,7 +519,13 @@ if(LINTER)
 
   if(CLANG_TIDY)
     message(STATUS "Linter support (clang-tidy) enabled")
-    set(CMAKE_C_CLANG_TIDY "${CLANG_TIDY};--quiet")
+    if(LINTER_STRICT)
+      set(CMAKE_C_CLANG_TIDY
+          "${CLANG_TIDY};--warnings-as-errors=clang-diagnostic-*,clang-analyzer-*,-*,clang-analyzer-core.*,clang-diagnostic-*"
+      )
+    else()
+      set(CMAKE_C_CLANG_TIDY "${CLANG_TIDY};--quiet")
+    endif(LINTER_STRICT)
   else()
     message(STATUS "Install clang-tidy to enable code linting")
   endif(CLANG_TIDY)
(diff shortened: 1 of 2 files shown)


afiskon:
The fail https://github.com/timescale/timescaledb/runs/4240260614?check_suite_focus=true is expected because https://github.com/timescale/timescaledb/pull/3819 should be merged first.

svenklemm (changes requested):
This should be handled differently and optional but CI should probably run with this enabled if we care about this

afiskon:
@svenklemm this can be always disabled locally by setting LINTER=OFF.  I didn't change anything compared to the way it is now, except the fact that the generated errors are not ignored anymore.

afiskon:
To clarify, clang-tydy doesn't have -Werror equivalent, thus I used available flags according to the documentation.

CMakeLists.txt
```diff
@@ -516,7 +516,9 @@ if(LINTER)
 
   if(CLANG_TIDY)
     message(STATUS "Linter support (clang-tidy) enabled")
-    set(CMAKE_C_CLANG_TIDY "${CLANG_TIDY};--quiet")
+    set(CMAKE_C_CLANG_TIDY
+        "${CLANG_TIDY};--warnings-as-errors=clang-diagnostic-*,clang-analyzer-*,-*,clang-analyzer-core.*,clang-diagnostic-*"
```
akuzm, line 520: Probably we can write `-warnings-as-errors=*` here. It doesn't enable the checks, and we won't have to maintain the list of checks in two different places.

Reading the docs here: https://clang.llvm.org/extra/clang-tidy/

akuzm: And according to Sven's comment, it might make sense to put this not here, but in the CI build settings, so that the warnings are treated as errors only the in CI. Not sure if people need clang-tidy at all in their default MacOS build settings. I don't, but I use Linux.

akuzm (approved):
I like the idea but we should figure out the details.

CMakeLists.txt
afiskon: OK, I will see if we can add it only on CI.

Regarding `-warnings-as-errors=*` I tried it locally and discovered that we have too many warnings in this case. Don't think we have the resources to fix them now, nor am I sure that this is necessary.

akuzm: Oh, I thought we have `-checks=<some closed list>`, no warnings with this list, and then `-warning-as-errors=*` could be enabled. I'd treat the clang-tidy warnings as normal warnings and make sure that there are none, so that the new ones are  easy to notice. A configuration that emits a lot of warnings feels inconvenient to use. But never mind, do as you see fit.

afiskon: The list I used here is a default list of checks (at least as I understood it from the docs). Ideally, I would like to have something like -Werror, but unfortunately clang-tidy doesn't seem to have such a flag. Thus I had to specify the list of checks that should be treated as errors manually 🤷‍♂️

afiskon:
@svenklemm I updated the PR according to your and Alexander's feedback. Please LMK what do you think?

svenklemm (approved):

PR #3827, opened 2021-11-18
Add an issue template for reporting flaky tests

erimatnor:
Tests are sometimes flaky, so add an issue template to report such
tests. The template ensures the issue is labeled and has the required
information we need to fix the flake.


mkindahl (approved):

akuzm (approved):
For all the issues intended to be created by the developers, I'd avoid forms or even templates at all, because we know better what we need to reproduce/fix the issue. But let's try this way too.

mkindahl (approved):

svenklemm (approved):

Issue #2807, opened 2021-01-12
Improve query performance by enabling parallel plans on data nodes


erimatnor:
The standard way we transfer data between the access node and data nodes is using a CURSOR, which supports large data sets, rewind, and interleaving multiple queries on the same connection (e.g., when doing subqueries and/or joins, multiple cursors can be open over the same connection). However, when executing a query via a cursor, PostgreSQL cannot do parallel execution plans, which slows down query performance compared to a single node instance.

The data fetcher implementation for multi-node supports two modes to fetch data: rowbyrow and cursor. Row-by-row is faster and supports parallel queries on data nodes, but doesn't work for some queries that include sub-queries.

It would be good to have an "auto" mode that picks the best data fetcher depending on the query.

akuzm:
Why the cursors do not support parallel operation: https://www.postgresql.org/message-id/flat/98760523-a0d2-8705-38e3-c4602ecf2448%402ndquadrant.com

In short, execution of cursor can be suspended and a user then can run arbitrary sql, and then resume execution. This kind  concurrency is not safe, postgres doesn't account for it.

akuzm:
And the reason TS needs cursors is to multiplex several subqueries to the same data node on the same connection. This is needed e.g. if we later join these subqueries w/ a nested loop.

PR #3852, opened 2021-11-24
use row-by-row fetcher for some queries to enable parallel plans on data nodes

akuzm:
Fixes https://github.com/timescale/timescaledb/issues/2807

The row-by-row fetcher is more efficient, so we want to use it when we can -- that is, when the have to read only one table from the data node, without interleaving it with anything else. This patch adds an option of choosing the fetcher type automatically. It detects the simplest case of only one distributed table in the entire query, and enables row-by-row fetcher. For other cases, the cursor fetcher is used.

### DIFF
--- a/src/planner.c
+++ b/src/planner.c
@@ -85,6 +85,18 @@ static void cagg_reorder_groupby_clause(RangeTblEntry *subq_rte, int rtno, List 
  */
 static const char *TS_CTE_EXPAND = "ts_expand";
 
+/*
+ * Controls which type of fetcher to use to fetch data from the data nodes.
+ * There is no place to store planner-global custom information (such as in
+ * PlannerInfo). Because of this, we have to use the global variable that is
+ * valid inside the scope of timescaledb_planner().
+ * Note that that function can be called recursively, e.g. when evaluating a
+ * SQL function at the planning time. We only have to determine the fetcher type
+ * in the outermost scope, so we distinguish it by that the fetcher type is set
+ * to the invalid value of 'auto'.
+ */
+DataFetcherType ts_data_node_fetcher_scan_type = AutoFetcherType;
+
 static void
 rte_mark_for_expansion(RangeTblEntry *rte)
 {
@@ -192,6 +204,20 @@ ts_rte_is_hypertable(const RangeTblEntry *rte, bool *isdistributed)
 #define IS_UPDL_CMD(parse)                                                                         \
 	((parse)->commandType == CMD_UPDATE || (parse)->commandType == CMD_DELETE)
 
+typedef struct
+{
+	Query *rootquery;
+	/*
+	 * The number of distributed hypertables in the query and its subqueries.
+	 * Specifically, we count range table entries here, so using the same
+	 * distributed table twice counts as two tables. No matter whether it's the
+	 * same physical table or not, the range table entries can be scanned
+	 * concurrently, and more than one of them being distributed means we have
+	 * to use the cursor fetcher so that these scans can be interleaved.
+	 */
+	int num_distributed_tables;
+} PreprocessQueryContext;
+
 /*
  * Preprocess the query tree, including, e.g., subqueries.
  *
@@ -207,7 +233,7 @@ ts_rte_is_hypertable(const RangeTblEntry *rte, bool *isdistributed)
  * 3. Reordering of GROUP BY clauses for continuous aggregates.
  */
 static bool
-preprocess_query(Node *node, Query *rootquery)
+preprocess_query(Node *node, PreprocessQueryContext *context)
 {
 	if (node == NULL)
 		return false;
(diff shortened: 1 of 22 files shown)


src/planner.c
```diff
@@ -312,8 +331,45 @@ timescaledb_planner(Query *parse, int cursor_opts, ParamListInfo bound_params)
...
+			 * a connection per table, we could avoid this requirement.
+			 */
+			if (ts_guc_remote_data_fetcher == AutoFetcherType)
+			{
+				if (context.num_distributed_tables >= 2)
+				{
+					data_node_scan_fetcher_type = CursorFetcherType;
+				}
+				else
+				{
```
nikkhils, line 364: We can have a case when there are Zero distributed hypertables, no?

tsl/test/shared/sql/dist_fetcher_type.sql
```diff
@@ -0,0 +1,29 @@
...
+-- Test that we use the correct type of remote data fetcher.
+set timescaledb.remote_data_fetcher = 'auto';
+select 1 x from distinct_on_distributed t1, distinct_on_distributed t2
+where t1.id = t2.id + 1
+limit 1;
+
+set timescaledb.remote_data_fetcher = 'cursor';
+select 1 x from distinct_on_distributed t1, distinct_on_distributed t2
+where t1.id = t2.id + 1
+limit 1;
```
nikkhils, line 16: Can we have EXPLAIN output show the cursor being used?

src/planner.c
akuzm: True, I fixed the comment.

tsl/test/shared/sql/dist_fetcher_type.sql
akuzm: Currently there is no way to tell from EXPLAIN. You mean, add it to the EXPLAIN output in this PR? I'll try that, I guess it might be useful for debugging performance. Maybe I'll only add the output for verbose mode, and also only for 'cursor' mode. 'rowbyrow' should be somewhat like the default now, and 'cursor' is the slower mode for complex queries. So we can only show this slower mode in EXPLAIN to avoid too much verbosity and test reference changes.

tsl/src/fdw/scan_exec.c
```diff
@@ -414,6 +429,11 @@ fdw_scan_explain(ScanState *ss, List *fdw_private, ExplainState *es, TsFdwScanSt
 
 		ExplainPropertyText("Data node", server->servername, es);
 
+		if (fsstate->fetcher_type == CursorFetcherType)
+		{
+			ExplainPropertyText("Fetcher Type", "Cursor", es);
```
erimatnor, line 434: If we output this info, we should also do it when row-by-row is used.

src/planner.c
```diff
@@ -312,8 +336,55 @@ timescaledb_planner(Query *parse, int cursor_opts, ParamListInfo bound_params)
...
+			 * only set/reset the fetcher type at the topmost level, that's why
+			 * we check it's not already set.
+			 */
+			if (ts_data_node_fetcher_scan_type == AutoFetcherType)
+			{
+				reset_fetcher_type = true;
+
+				if (ts_guc_remote_data_fetcher == AutoFetcherType)
+				{
+					if (context.num_distributed_tables >= 2)
```
erimatnor, line 373: Does this cover the case when you have a subquery from the same table? E.g., something like:

```sql
SELECT * FROM myhypertable WHERE  device=(SELECT device FROM myhypertable LIMIT 1);
```

In other words, does this count as two distributed hypertables or just one?

tsl/test/shared/sql/dist_fetcher_type.sql
```diff
@@ -0,0 +1,39 @@
...
+
+-- Check once again that 'auto' is used after 'rowbyrow'.
+set timescaledb.remote_data_fetcher = 'auto';
+select 1 x from distinct_on_distributed t1, distinct_on_distributed t2
+where t1.id = t2.id + 1
+limit 1;
+
+reset timescaledb.remote_data_fetcher;
```
erimatnor, line 39: missing new line

tsl/src/fdw/scan_exec.c
```diff
@@ -126,11 +128,24 @@ create_data_fetcher(ScanState *ss, TsFdwScanState *fsstate)
...
+	{
+		fetcher = cursor_fetcher_create_for_scan(fsstate->conn,
+												 ss,
+												 fsstate->retrieved_attrs,
+												 fsstate->query,
+												 params);
+	}
+	else
+	{
+		Assert(fsstate->fetcher_type == RowByRowFetcherType);
```
erimatnor, line 141: Suggest adding a comment on why "auto" is not possible at this point.

erimatnor (commented):
Looks OK, but have a couple of suggestions and questions.

Also, please provide a PR description and commit log entry. It's good to do this prior to requesting reviews given that this helps reviewers understand the context of the PR. There are some good comments in the code, so suggest using that as basis for the PR description.

Try to follow the seven rules of a great commit message: https://chris.beams.io/posts/git-commit/

src/planner.c
akuzm: Two, because they might be scanned concurrently, no matter whether it's the same physical table or not. Added a comment for num_distributed_tables.

erimatnor (approved):

nikkhils (approved):
LGTM!

PR #3853, opened 2021-11-24
Handle default privileges on distributed databases

### LINKED ISSUE #3851: [Enhancement]: Handle default privileges on distributed databases

User experience

Command processing, Multi-node

When altering the default privileges for, e.g., schemas or tables, those default privileges should also be set on data nodes.

### Implementation challenges

A role or a schema might not exist on a data node, and in that case an error will be raised. Normally, it is up to the DB admin to ensure the roles and schemas required exist in a consistent manner across nodes in the multi-node cluster.

However, there might be legitimate cases when a role or a schema does not exist on a data node while it exists on the access node. In that case, there's a risk the command for granting default privileges for such a schema or role might fail when the expectation is that it should succeed.

So, either we have to accept that users have to handle default privileges manually using `distributed_exec` or we mandate that all schemas and roles should exist on all nodes.

erimatnor:
Statements that alter default privileges are now by default forwarded
to all data nodes. For instance:

```sql
ALTER DEFAULT PRIVILEGES IN SCHEMA myschema GRANT INSERT ON TABLES TO myrole;
```

is applied across the entire cluster of data nodes.

Fixes #3851

### DIFF
--- a/tsl/src/remote/dist_ddl.c
+++ b/tsl/src/remote/dist_ddl.c
@@ -152,7 +152,7 @@ dist_ddl_state_set_exec_type(DistDDLExecType type)
  * execution on data nodes.
  */
 static inline void
-dist_ddl_state_schedule(DistDDLExecType type, ProcessUtilityArgs *args)
+dist_ddl_state_schedule(DistDDLExecType type, const ProcessUtilityArgs *args)
 {
 	Assert(dist_ddl_state.exec_type == DIST_DDL_EXEC_NONE);
 	Assert(type != DIST_DDL_EXEC_NONE);
@@ -232,7 +232,7 @@ dist_ddl_state_add_current_data_node_list(void)
 }
 
 static void
-dist_ddl_inspect_hypertable_list(ProcessUtilityArgs *args, Cache *hcache,
+dist_ddl_inspect_hypertable_list(const ProcessUtilityArgs *args, Cache *hcache,
 								 unsigned int *num_hypertables, unsigned int *num_dist_hypertables,
 								 unsigned int *num_dist_hypertable_members)
 {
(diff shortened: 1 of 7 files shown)


tsl/src/remote/dist_ddl.c
```diff
@@ -603,21 +601,71 @@ dist_ddl_process_grant_on_table(ProcessUtilityArgs *args)
...
+
+static void
+dist_ddl_process_grant_all_in_schema(const ProcessUtilityArgs *args)
+{
+	const GrantStmt *stmt = castNode(GrantStmt, args->parsetree);
+
+	switch (stmt->objtype)
+	{
+		case OBJECT_TABLE:
+			dist_ddl_process_grant_on_schema(args);
```
pmwkaa, line 629: Just to note that the naming of this function is actually not correct, I've fixed it in my PR. This is still grant on table command but with all in schema clause. `GRANT ON SCHEMA` is a different command

tsl/src/remote/dist_ddl.c
```diff
@@ -603,21 +601,71 @@ dist_ddl_process_grant_on_table(ProcessUtilityArgs *args)
...
+			break;
+	}
+}
+
+static void
+dist_ddl_process_grant(const ProcessUtilityArgs *args)
+{
+	const GrantStmt *stmt = castNode(GrantStmt, args->parsetree);
+
+	switch (stmt->targtype)
```
pmwkaa, line 641: To be honest, I found older way to be less confusing because the semantics on the grant is to be executed on the specific object in the first place. Now you have a reverse mapping, which requires to check each function separately

pmwkaa (approved):

tsl/src/remote/dist_ddl.c
erimatnor: Ok, I'll let you fix it then...

nikkhils (approved):
LGTM

Issue #3690, opened 2021-10-13
Multinode: CREATE SCHEMA on datanodes


nikkhils:
The following approaches can be used for schema creation on datanodes:

A) Ship the `CREATE SCHEMA` command the same time as AN to the DNs
If `AUTHORIZATION role_name` is specified then it's possible that that role does not exist on the DNs. The command will then fail on DN and the user can decide about creating the role on the DN as well.

B) Create the schema (if it does not exist on DN) the first time the distributed hypertable gets created on the DN. But in that case all the GRANTs on the schema on the AN need to be forwarded to the DN as well. Again the issue arises here if the roles which are part of the GRANTs are not present on the DN.

Note that if `cloudutils` extension is INSTALLED then the roles will always exist on the DNs as well

PR #3857, opened 2021-11-25
Support CREATE SCHEMA on distributed database

pmwkaa:
This PR makes CREATE SCHEMA command shippable to all data nodes
in the distributed database. Additional support for `GRANT ON SCHEMA` command added.

Issue: https://github.com/timescale/timescaledb/issues/3690


tsl/src/remote/dist_ddl.c
```diff
@@ -343,6 +343,19 @@ dist_ddl_state_get_hypertable_type(void)
...
 }
 
+static void
+dist_ddl_process_create_schema(ProcessUtilityArgs *args)
+{
+	CreateSchemaStmt *stmt = castNode(CreateSchemaStmt, args->parsetree);
+
+	if (dist_util_membership() != DIST_MEMBER_ACCESS_NODE)
+		return;
+	(void) stmt;
```
erimatnor, line 353:
```suggestion

```

tsl/src/remote/dist_ddl.c
```diff
@@ -343,6 +343,19 @@ dist_ddl_state_get_hypertable_type(void)
 	return type;
 }
 
+static void
+dist_ddl_process_create_schema(ProcessUtilityArgs *args)
+{
+	CreateSchemaStmt *stmt = castNode(CreateSchemaStmt, args->parsetree);
```
erimatnor, line 349:
```suggestion
	Assert(IsA(CreateSchemaStmt, args->parsetree));
```

tsl/src/remote/dist_ddl.c
```diff
@@ -343,6 +343,19 @@ dist_ddl_state_get_hypertable_type(void)
 	return type;
 }
 
+static void
+dist_ddl_process_create_schema(ProcessUtilityArgs *args)
```
erimatnor, line 347:
```suggestion
dist_ddl_process_create_schema(const ProcessUtilityArgs *args)
```

tsl/src/remote/dist_ddl.c
```diff
@@ -343,6 +343,19 @@ dist_ddl_state_get_hypertable_type(void)
...
+{
+	CreateSchemaStmt *stmt = castNode(CreateSchemaStmt, args->parsetree);
+
+	if (dist_util_membership() != DIST_MEMBER_ACCESS_NODE)
```
erimatnor, line 351: Why do we need this check here? 

I don't think we should even get to this point if we are not an access node since we shouldn't process dist ddl on data nodes. This should have been handled much higher up in the call-stack IMO.

Also, I don't see this check in other dist_ddl_process_ functions, so it seems inconsistent.

I think maybe Asserting is OK.

tsl/src/remote/dist_ddl.c
```diff
@@ -602,19 +614,37 @@ dist_ddl_process_grant_on_table(ProcessUtilityArgs *args)
 	dist_ddl_state_schedule(DIST_DDL_EXEC_ON_START, args);
 }
 
+static void
+dist_ddl_process_grant_on_schema(ProcessUtilityArgs *args)
+{
+	if (dist_util_membership() != DIST_MEMBER_ACCESS_NODE)
```
erimatnor, line 620: Same question here as above.

tsl/test/expected/dist_ddl.out
```diff
@@ -34,9 +35,6 @@ CALL distributed_exec('CREATE TABLE local(time timestamptz, measure int)', '{ "d
 CALL distributed_exec($$ SELECT create_hypertable('local', 'time') $$, '{ "data_node_1", "data_node_3" }');
 -- Import testsupport.sql file to data nodes
 \unset ECHO
--- This SCHEMA will not be created on data nodes
-CREATE SCHEMA disttable_schema AUTHORIZATION :ROLE_1;
```
erimatnor, line 38: Why was this removed? Is `disttable_schema` not used?

tsl/test/expected/dist_ddl.out
```diff
@@ -34,9 +35,6 @@ CALL distributed_exec('CREATE TABLE local(time timestamptz, measure int)', '{ "d
...
-CREATE SCHEMA some_schema AUTHORIZATION :ROLE_1;
```
erimatnor, line 39: `some_schema` is  now created above instead, but without AUTHORIZATION. Is this intentional?

tsl/test/expected/dist_ddl.out
```diff
@@ -479,7 +477,7 @@ ERROR:  operation not supported on distributed hypertable
 ALTER TABLE disttable SET SCHEMA some_unexist_schema;
 ERROR:  schema "some_unexist_schema" does not exist
 ALTER TABLE disttable SET SCHEMA some_schema;
-ERROR:  operation not supported on distributed hypertable
+ERROR:  permission denied for schema some_schema
```
erimatnor, line 480: I guess this error happens now because the CREATE SCHEMA above lacks AUTHORIZATION for ROLE_1? Is this intentional, e.g., to test lack of authorization? Would be good to add some comments to explain the intention of the test in that case.

tsl/src/remote/dist_ddl.c
pmwkaa: This can be triggered for a regular database, not only a data node

erimatnor: Yeah, we shouldn't process dist_util stuff at all for anything but an access node. So, having the check at this low level is confusing.

tsl/test/expected/dist_ddl.out
pmwkaa: I've updated the test and moved this clause back

tsl/test/expected/dist_ddl.out
pmwkaa: Because it would not work now, the tests bellow assumed that this schema was not created on the data nodes. I've moved schema creation down in the code, where the CREATE SCHEMA is actually tested

tsl/src/remote/dist_ddl.c
```diff
@@ -978,6 +1000,10 @@ dist_ddl_start(ProcessUtilityArgs *args)
 	if (dist_ddl_scheduled_for_execution())
 		return;
 
+	/* Process remote DDL only on the distributed database */
+	if (!IsTransactionState() || dist_util_membership() == DIST_MEMBER_NONE)
```
akuzm, line 1004: Two questions:
- why is this called for a non-distributed database at all?
- what does the `!IsTransactionState()` check mean?

akuzm (approved):

tsl/src/remote/dist_ddl.c
pmwkaa: 1) We have checks which block DDL operations executed directly on data nodes from a client. Basically we have a guc which allows or blocks those operations if they are not executed from the access node connection. This cases are handled there. So currently globally we prevent only non-distributed databases.

2) This is something I stood today. Since the dist ddl code executed after being already executed from main process utility hook in `src/process_utility.c` we could end in state, when transaction is already completed by `ROLLBACK` statement. Since `dist_util_membership()` function reads meta data catalog, it is required to have an active transaction there.

akuzm: Thanks. Maybe it makes sense to put this entire comment into code, I'll forget it again :)

erimatnor (approved):

PR #3863, opened 2021-11-29
Fix remote txn heal logic

### LINKED ISSUE #3219: Remote transactions cleanup/healing job fails on every run
 - OS: Centos 7.5
 - PostgreSQL version: `12.6`
 - TimescaleDB version (output of `\dx` in `psql`): `2.1.0`
 - Installation method: `yum`

**Describe the bug**
Cleanup of unfinished transactions always fails/stops at the very first transaction that causes error, which in turn causes it to not cleanup anything and the number of unfinished transactions keeps building up.

**To Reproduce**
1. Have a distributed hypertable with a constant non-stop flow of data (in our case about 40MB/s [bytes, not bits] of inbound data for writing in one table)
2. Have a maintenance job configured as per https://docs.timescale.com/timescaledb/latest/how-to-guides/multinode-timescaledb/multinode-maintenance/#multi-node-maintenance-tasks
3. Check logs
4. Check `select count(*) from _timescaledb_catalog.remote_txn;`

**Expected behavior**
The number of remote transactions over time should be a saw pattern (such as build up until the cleanup job runs which finishes/heals all/most of the unfinished transactions). The logs should not have any errors unless something critical happens.

**Actual behavior**
The number of remote transaction over time keeps building up and never being flushed. The configured maintenance task fails with following entries in logs intermixed (note that tx ids are changing, but in some cases are repeating non-stop i.e. stuck on one bad transaction) (note timestamps are removed and hostnames are edited):

```
[22822] LOG:  job 1000 threw an error
[22822] ERROR:  [datanode01.domain.name]: prepared transaction with identifier "ts-1-540621498-20818-20824" is busy
[22822] CONTEXT:  SQL function "data_node_maintenance" statement 1
[37139] LOG:  background worker "User-Defined Action [1000]" (PID 22822) exited with exit code 1

[22833] LOG:  job 1000 threw an error
[22833] ERROR:  [datanode03.domain.name]: prepared transaction with identifier "ts-1-540625107-20762-20824" does not exist
[22833] CONTEXT:  SQL function "data_node_maintenance" statement 1
[37139] LOG:  background worker "User-Defined Action [1000]" (PID 22833) exited with exit code 1
```

**Additional context**
This is causing growing disk space usage on access node (some of databases on access node grew to 300GB+ in size just because of the size of forever-growing `_timescaledb_catalog.remote_txn`

I do think that the job should skip TX if it is busy/does not exist/some other issue. And should continue to the cleanup of transactions that it can process still. Otherwise (as noticed by performance metrics) the inserts become even slower for distributed setup (from ~100ms per 1000 rows to ~150ms with remote_txn size being over 200GB).

nikkhils:
* A few tweaks to the remote txn resolution logic
* Add logic to delete a specific record in remote_txn table by GID
* Allow heal logic to move on to other cleanup if one specific GID
fails
* Do not rely on ongoing txns while cleaning up entries from remote_txn
table

Includes test case changes to try out various failure scenarios in the
healing function.

Fixes #3219

### DIFF
--- a/tsl/src/remote/txn.c
+++ b/tsl/src/remote/txn.c
@@ -6,6 +6,7 @@
 #include "libpq-fe.h"
 #include <postgres.h>
 #include <access/xact.h>
+#include <storage/procarray.h>
 #include <utils/builtins.h>
 #include <utils/snapmgr.h>
 #include <libpq-fe.h>
@@ -152,13 +153,17 @@ remote_txn_begin(RemoteTxn *entry, int curlevel)
 	}
 }
 
+/*
+ * Check if the access node transaction which is driving the 2PC on the datanodes is
+ * still in progress.
+ */
 bool
-remote_txn_is_still_in_progress(TransactionId access_node_xid)
+remote_txn_is_still_in_progress_on_access_node(TransactionId access_node_xid)
 {
 	if (TransactionIdIsCurrentTransactionId(access_node_xid))
 		elog(ERROR, "checking if a commit is still in progress on same txn");
 
-	return XidInMVCCSnapshot(access_node_xid, GetTransactionSnapshot());
+	return TransactionIdIsInProgress(access_node_xid);
 }
 
 size_t
@@ -731,27 +736,43 @@ persistent_record_tuple_delete(TupleInfo *ti, void *data)
 	return SCAN_CONTINUE;
 }
 
+/* If gid is NULL, then delete all entries belonging to the provided datanode.  */
 int
-remote_txn_persistent_record_delete_for_data_node(Oid foreign_server_oid)
+remote_txn_persistent_record_delete_for_data_node(Oid foreign_server_oid, const char *gid)
 {
 	Catalog *catalog = ts_catalog_get();
 	ScanKeyData scankey[1];
 	ScannerCtx scanctx;
+	int scanidx;
 	ForeignServer *server = GetForeignServer(foreign_server_oid);
 
-	ScanKeyInit(&scankey[0],
-				Anum_remote_txn_data_node_name_idx_data_node_name,
-				BTEqualStrategyNumber,
-				F_NAMEEQ,
-				CStringGetDatum(server->servername));
+	if (gid == NULL)
+	{
+		ScanKeyInit(&scankey[0],
+					Anum_remote_txn_data_node_name_idx_data_node_name,
+					BTEqualStrategyNumber,
+					F_NAMEEQ,
+					CStringGetDatum(server->servername));
+		scanidx = REMOTE_TXN_DATA_NODE_NAME_IDX;
+	}
+	else
+	{
+		ScanKeyInit(&scankey[0],
+					Anum_remote_txn_pkey_idx_remote_transaction_id,
+					BTEqualStrategyNumber,
+					F_TEXTEQ,
+					CStringGetTextDatum(gid));
+		scanidx = REMOTE_TXN_PKEY_IDX;
+	}
 
 	scanctx = (ScannerCtx){
 		.table = catalog->tables[REMOTE_TXN].id,
-		.index = catalog_get_index(catalog, REMOTE_TXN, REMOTE_TXN_DATA_NODE_NAME_IDX),
+		.index = catalog_get_index(catalog, REMOTE_TXN, scanidx),
 		.nkeys = 1,
 		.scankey = scankey,
 		.tuple_found = persistent_record_tuple_delete,
 		.lockmode = RowExclusiveLock,
+		.snapshot = GetTransactionSnapshot(),
 		.scandirection = ForwardScanDirection,
 	};
 
--- a/tsl/src/remote/txn_resolve.c
+++ b/tsl/src/remote/txn_resolve.c
@@ -5,6 +5,7 @@
  */
 #include <postgres.h>
 #include <utils/fmgrprotos.h>
+#include <utils/guc.h>
 #include <utils/snapmgr.h>
 #include <utils/fmgroids.h>
 #include <access/xact.h>
@@ -18,13 +19,24 @@
 RemoteTxnResolution
 remote_txn_resolution(Oid foreign_server, const RemoteTxnId *transaction_id)
 {
-	if (remote_txn_is_still_in_progress(transaction_id->xid))
-		/* transaction still ongoing; don't know it's state */
-		return REMOTE_TXN_RESOLUTION_UNKNOWN;
+	if (remote_txn_is_still_in_progress_on_access_node(transaction_id->xid))
+		/* transaction still ongoing; don't know its state */
+		return REMOTE_TXN_RESOLUTION_IN_PROGRESS;
 
+	/*
+	 * If an entry exists in the "remote_txn" table and is visible then it means
+	 * that the transaction committed on the AN
+	 */
 	if (remote_txn_persistent_record_exists(transaction_id))
-		return REMOTE_TXN_RESOLUTION_COMMT;
+		return REMOTE_TXN_RESOLUTION_COMMIT;
 
+	/*
+	 * If the txn is not in progress and is not committed as per the "remote_txn"
+	 * table then it's presumed to be aborted.
+	 *
+	 * We could ask PG machinery to confirm the abort but as long as we are sticking
+	 * to one uniform behavior consistently it should be ok for now.
+	 */
 	return REMOTE_TXN_RESOLUTION_ABORT;
 }
 
@@ -58,8 +70,11 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
 	 */
 	PGresult *res;
 	int row;
-	List *unknown_txn_gid = NIL;
-	int non_ts_txns = 0;
+	List *in_progress_txn_gids = NIL, *healed_txn_gids = NIL;
+	int non_ts_txns = 0, ntuples;
+#ifdef TS_DEBUG
+	int n_gid_errors = 0; /* how many errors to induce? */
+#endif
 
 	/*
 	 * This function cannot be called inside a transaction block since effects
@@ -70,9 +85,10 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
 	res = remote_connection_query_ok(conn, GET_PREPARED_XACT_SQL);
 
 	Assert(1 == PQnfields(res));
-	for (row = 0; row < PQntuples(res); row++)
+	ntuples = PQntuples(res);
+	for (row = 0; row < ntuples; row++)
 	{
-		const char *id_string = PQgetvalue(res, row, 0);
+		char *id_string = PQgetvalue(res, row, 0);
 		RemoteTxnId *tpc_gid;
 		RemoteTxnResolution resolution;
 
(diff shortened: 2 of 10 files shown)


tsl/src/remote/txn_resolve.c
```diff
@@ -22,10 +22,21 @@ remote_txn_resolution(Oid foreign_server, const RemoteTxnId *transaction_id)
 		/* transaction still ongoing; don't know it's state */
```
akuzm, line 22: GitHub doesn't let me comment on the above line...
`remote_txn_is_still_in_progress` looks weird -- it actually checks whether the access node transaction id (`RemoteTxnId.xid`) is in current MVCC snapshot. I'm not sure what this even means. I remember we discussed removing this snapshot check, is it still needed here?

tsl/src/remote/txn_resolve.c
```diff
@@ -22,10 +22,21 @@ remote_txn_resolution(Oid foreign_server, const RemoteTxnId *transaction_id)
...
 		return REMOTE_TXN_RESOLUTION_COMMT;
 
-	return REMOTE_TXN_RESOLUTION_ABORT;
+	/*
+	 * We need to check the Postgres machinery to ensure that this txn has indeed
+	 * aborted before declaring so.
+	 */
+	if (TransactionIdDidAbort(transaction_id->xid))
+		return REMOTE_TXN_RESOLUTION_ABORT;
+	else /* something's still going on with this transaction */
```
akuzm, line 38: Do we know some scenario where this happens? It could be helpful to describe it in this comment.

tsl/src/remote/txn_resolve.c
```diff
@@ -108,9 +142,22 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
...
+	 *
+	 * In a heavily loaded system there's a possibility of ongoing transactions always being
+	 * present in which case we will never get a chance to clean up entries in "remote_txn"
+	 * table. So, we track healed gids in a list and delete those specific rows to keep the
+	 * "remote_txn" table from growing up indefinitely.
 	 */
-	if (list_length(unknown_txn_gid) == 0)
-		remote_txn_persistent_record_delete_for_data_node(foreign_server_oid);
+	if (list_length(unknown_txn_gids) == 0)
+		remote_txn_persistent_record_delete_for_data_node(foreign_server_oid, NULL);
```
erimatnor, line 152: I think this wildcard removal might be broken because the internal Scanner uses a "SnapshotSelf" by default, which means the delete scan here can see new entries at this instant that weren't present in the scan of gids above. We should probably modify the scan functions to use a common snapshot set in the beginning of the function (or use the TransactionSnapshot) so that they have the same view of the metadata.

tsl/src/remote/txn_resolve.c
erimatnor: If it is in the current MVCC snapshot, it means the other transaction is still ongoing (not committed) so no decision on commit or abort has been made at this point. Thus, the healing function should ignore that GID.

tsl/src/remote/txn_resolve.c
erimatnor: It is a concurrent and still ongoing transaction where there is no decisions on COMMIT or ABORT yet. Thus no healing or cleanup decision can be made.

I agree that the comment can be more explanatory though.

tsl/src/remote/txn_resolve.c
nikkhils: @akuzm agreed. I have renamed the function and added a comment on top as well.

tsl/src/remote/txn_resolve.c
nikkhils: Ok, good catch. Using `GetTransactionSnapshot()` now.

tsl/src/remote/txn_resolve.c
```diff
@@ -108,9 +142,22 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
...
+	else if (resolved)
+	{
+		ListCell *lc;
+		Assert(healed_txn_gids != NIL);
+
+		foreach (lc, healed_txn_gids)
+			remote_txn_persistent_record_delete_for_data_node(foreign_server_oid, lfirst(lc));
```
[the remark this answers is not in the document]
nikkhils: I don't know of any scan which takes multiple inputs like `IN (1,2,3)`

tsl/src/remote/txn_resolve.c
```diff
@@ -85,18 +96,41 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
...
 
+		/*
+		 * We don't expect these commands to fail, but if they do, continue and move on to
+		 * healing up the next GID in the list. The ones that failed will get retried if
+		 * they are still around on the datanodes the next time over.
+		 */
 		switch (resolution)
 		{
 			case REMOTE_TXN_RESOLUTION_COMMT:
-				remote_connection_cmd_ok(conn, remote_txn_id_commit_prepared_sql(tpc_gid));
```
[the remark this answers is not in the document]
nikkhils: Fixed.

tsl/src/remote/txn.c
```diff
@@ -152,13 +153,17 @@ remote_txn_begin(RemoteTxn *entry, int curlevel)
...
+ */
 bool
-remote_txn_is_still_in_progress(TransactionId access_node_xid)
+remote_txn_is_still_in_progress_on_access_node(TransactionId access_node_xid)
 {
 	if (TransactionIdIsCurrentTransactionId(access_node_xid))
 		elog(ERROR, "checking if a commit is still in progress on same txn");
 
-	return XidInMVCCSnapshot(access_node_xid, GetTransactionSnapshot());
+	return TransactionIdIsInProgress(access_node_xid);
```
erimatnor, line 166: While this seems to be the right function to use, what are the differences to how things were before with `XidInMVCCSnapshot`?

tsl/src/remote/txn_resolve.c
```diff
@@ -85,18 +99,80 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
...
+						remote_connection_exec(conn, remote_txn_id_commit_prepared_sql(tpc_gid))) ==
+					PGRES_COMMAND_OK)
+				{
+					healed_txn_gids = lappend(healed_txn_gids, id_string);
+					resolved++;
+				}
+				else
+					ereport(WARNING,
+							(errmsg("\"COMMIT PREPARED\" failed. Will be retried later"),
+							 errhint("gid: %s", id_string)));
```
erimatnor, line 159: This error doesn't give enough context for users when they see it in the log. It is not clear what the prepared transaction is for, which data node is involved, or that his anything to do with multi-node. Also, I would recommend to use `errdetail` for general information and hints for for giving information to the user on how to fix the error.

If we say that this will be retried later it might be a bit  misleading since it sounds as it will happen automatically. This is not always true, since a healing policy must be running. I think we can be more clear, something like:

```suggestion
					ereport(WARNING,
							(errmsg("could not commit prepared transaction %s on data node \"%s\"", id_string, remote_connection_node_name(conn)),
							 errhint("To retry, manually run \"COMMIT PREPARED %s\" on the data node or run the healing function again.", id_string));
```

I would also avoid mentioning GID, as this is an internal name and generally the PG documentation just calls this `transaction_id`.

Btw, what is the behavior of running `COMMIT PREPARED` yourself on the transaction ID? I guess that would work too?


Finally, since this is called in a loop, I am wondering if we risk flooding the output/log with WARNINGs in case of a data node having problems. There might be many prepared transactions to heal on a node. 

One idea is to collect the IDs in the loop and then print only one warning outside the loop that lists all IDs (unless there are a huge amount of transactions that failed).

tsl/src/remote/txn_resolve.c
```diff
@@ -85,18 +99,80 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
...
+														  remote_txn_id_rollback_prepared_sql(
+															  tpc_gid))) == PGRES_COMMAND_OK)
+				{
+					healed_txn_gids = lappend(healed_txn_gids, id_string);
+					resolved++;
+				}
+				else
+					ereport(WARNING,
+							(errmsg("\"ROLLBACK PREPARED\" failed. Will be retried later"),
+							 errhint("gid: %s", id_string)));
```
erimatnor, line 172: Same here as above:

```suggestion
					ereport(WARNING,
							(errmsg("could not roll back prepared transaction %s on data node \"%s\"", id_string, remote_connection_node_name(conn)),
							 errhint("To retry, manually run \"ROLLBACK PREPARED %s\" on the data node or run the healing function again.", id_string));
```

tsl/src/remote/txn_resolve.c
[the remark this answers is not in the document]
erimatnor: Nevermind then.

tsl/src/remote/txn_resolve.c
```diff
@@ -107,10 +183,24 @@ remote_txn_heal_data_node(PG_FUNCTION_ARGS)
...
-		remote_txn_persistent_record_delete_for_data_node(foreign_server_oid);
+	if (list_length(unknown_txn_gids) == 0 && resolved == ntuples)
+		remote_txn_persistent_record_delete_for_data_node(foreign_server_oid, NULL);
+	else if (resolved)
+	{
+		ListCell *lc;
+		Assert(healed_txn_gids != NIL);
+
+		foreach (lc, healed_txn_gids)
+			remote_txn_persistent_record_delete_for_data_node(foreign_server_oid, lfirst(lc));
```
erimatnor, line 202: Any way we can hit this in tests?

tsl/test/expected/remote_txn_resolve.out
```diff
@@ -137,6 +137,146 @@ SELECT count(*) FROM _timescaledb_catalog.remote_txn;
...
+SELECT * FROM table_modified_by_txns;
+           describes            
+--------------------------------
+ committed
+ prepared not comitted
+ committed with concurrent heal
+ committed
+(4 rows)
+
+SELECT count(*) FROM pg_prepared_xacts;
```
erimatnor, line 165: Would be more interesting to se the prepared transactions on the data node(s).

tsl/test/expected/remote_txn_resolve.out
```diff
@@ -137,6 +137,146 @@ SELECT count(*) FROM _timescaledb_catalog.remote_txn;
...
+SELECT create_records();
+ create_records 
+----------------
+ 
+(1 row)
+
+--inject errors in the GID and test "commit" resolution for it
+SET timescaledb.debug_inject_gid_error TO 'commit';
+--heal should error out and the prepared transaction should still be visible
+SELECT _timescaledb_internal.remote_txn_heal_data_node((SELECT OID FROM pg_foreign_server WHERE srvname = 'loopback2'));
```
erimatnor, line 149: It isn't really clear what we are testing here. It seems as if we are just failing all the healings we are trying, which means we are actually just not executing the healing code. 

Wouldn't it be more interesting to test a failure in healing one transaction out of several transaction, but seeing that the rest succeeds? Or is that what is happening, although it is not clear to me?

erimatnor (commented):
I think the code is probably OK, but I have some suggestions for improving a couple of things. I am also not sure I understand exactly what we are testing.

tsl/src/remote/txn.c
nikkhils: As @akuzm pointed out, `TransactionIdIsInProgress` reports the liveness of the XID across the entire set of ongoing txns whereas `XidInMVCCSnapshot` just checks against our current transaction snapshot here. The former might be a little bit more safer.

tsl/src/remote/txn_resolve.c
nikkhils: I didn't want to print the `id_string` in the main ereport to get a static report in the test output which is used/visible in the test runs. Will modify the errhint as per your suggestion.

Yes, if you run `COMMIT PREPARED` manually then that will work as well.
You cannot have more than `max_prepared_transactions` in the system which is typically not a very large value. Also, typically there are NOT very many pending prepared transactions in case a datanode is behaving sanely. So I don't expect a lot of entries being obtained by the healing function.

tsl/src/remote/txn_resolve.c
nikkhils: Ok, will rework the errhint

tsl/src/remote/txn_resolve.c
nikkhils: I already added a test for this in `test_basic_persistent_record` function.

tsl/test/expected/remote_txn_resolve.out
nikkhils: Well, that output is not going to be the same everytime we run the test, right?

tsl/test/expected/remote_txn_resolve.out
nikkhils: The workings of the healing code is tested earlier on many times. Here we are trying to exercise the code paths which will fail very rarely via this GUC. Yeah, all heals will fail due to the GUC if there are more than one entries returned from the remote side. But I am not sure that extra effort to test that is worth it.

tsl/src/remote/txn_resolve.c
```diff
@@ -18,14 +19,26 @@
...
 {
-	if (remote_txn_is_still_in_progress(transaction_id->xid))
-		/* transaction still ongoing; don't know it's state */
+	if (remote_txn_is_still_in_progress_on_access_node(transaction_id->xid))
+		/* transaction still ongoing; don't know its state */
 		return REMOTE_TXN_RESOLUTION_UNKNOWN;
 
+	/*
+	 * If an entry exists in the "remote_txn" table and is visible then it means
+	 * that the transaction committed on the AN
```
pmwkaa, line 28: That is a bit confusing what is actually commit means here. Is it `PREPARE TRANSACTION`? That does not mean, that the transaction committed actually, it means that the all nodes in the cluster prepared the transaction and it is ready to be commited in the next stage, which will be `COMMIT PREPARED`.

pmwkaa (approved):
Changes looks good overall, however it is not clear what was changed that fixes the original failure in the issue. For eg, will the healing continue to operate if it fails on one of the transactions?

It would be great to test healing functionality using data nodes at least and by recreating real life scenarios according to the transaction workflow and it's life cycle. Not necessary from the implementation perspective, but from expected functionality and outcome to the user.

In order to do that, we need to introduce wait stages during transaction execution on the data node specifically and deterministically run the healing function in between to make sure that all system component are in correct state.

Not sure if we should do it in this PR, maybe add a new issue at least for a more extensive testing?

tsl/src/remote/txn_resolve.c
nikkhils: The comment says committed in the AN, so that means all stages including PREPARE TRANSACTION and COMMIT PREPARED have already completed on all the datanodes.

nikkhils:
> Changes looks good overall, however it is not clear what was changed that fixes the original failure in the issue. For eg, will the healing continue to operate if it fails on one of the transactions?
> 

Yes, the healing will continue with the healing of the next prepared transaction in case the earlier fails.

> It would be great to test healing functionality using data nodes at least and by recreating real life scenarios according to the transaction workflow and it's life cycle. Not necessary from the implementation perspective, but from expected functionality and outcome to the user.
> 

There are already existing test cases for this in `remote_txn_resolve.sql` test file. So the regular healing mechanism is ALREADY covered properly.

> In order to do that, we need to introduce wait stages during transaction execution on the data node specifically and deterministically run the healing function in between to make sure that all system component are in correct state.
> 
> Not sure if we should do it in this PR, maybe add a new issue at least for a more extensive testing?

IMHO, the existing testing in `remote_txn_resolve.sql` is good enough for now and we don't need any additional complicated mechanism.

tsl/src/remote/txn.c
erimatnor: I just want to make sure that we cover the case when the other transaction completes before the end of the "current" transaction, but other transaction is not visible in the snapshot.

So, let's say T_a starts before T_b but T_b commits before T_a. Then there's time between T_b commit and T_a commit where this function returns false. Do we handle that situation correctly?

tsl/src/remote/txn_resolve.c
```diff
@@ -18,14 +19,26 @@
...
-		return REMOTE_TXN_RESOLUTION_COMMT;
+		return REMOTE_TXN_RESOLUTION_COMMIT;
 
-	return REMOTE_TXN_RESOLUTION_ABORT;
+	/*
+	 * We need to check the Postgres machinery to ensure that this txn has indeed
+	 * aborted before declaring so. If it's not aborted and not committed as per
+	 * above then it has to be a concurrent ongoing transaction.
+	 */
+	if (TransactionIdDidAbort(transaction_id->xid))
```
erimatnor, line 38: I think this only works if the transaction is still on record in the clog. I am wondering if there is a situation where a data node has been down for a long time and vacuum has removed the record of the XID on the access node, thus this might return `REMOTE_TXN_RESOLUTION_UNKNOWN` and the record will never be cleaned up because it is believed to still be ongoing?


Therefore, I think it is a bit strange to check for an explicit abort when we implement "presumed abort". Given that we implement "presumed abort" it should be the default unless we have an explicit record of it either being in process or committed.

tsl/src/remote/txn.c
nikkhils: Returns `false` for `T_b`? No, if we use `TransactionIdIsInProgress` then it will not return `false` for `T_b` AIUI

tsl/src/remote/txn_resolve.c
nikkhils: Went back to the earlier logic of declaring abort.

tsl/test/expected/remote_txn_resolve.out
erimatnor: Couldn't one add a counter or boolean in the debug code to fail just the first prepared transaction instead of all of them?

erimatnor (approved):
Just one comment in one of the tests.

tsl/test/expected/remote_txn_resolve.out
nikkhils: I added the above code. Now the first prepared transaction fails and the other one gets healed up as desired.

PR #3873, opened 2021-11-30
Enable compression on continuous aggregates

gayyappan:
Disable-check: commit-count


tsl/src/bgw_policy/compression_api.c
```diff
@@ -151,13 +155,35 @@ policy_recompression_proc(PG_FUNCTION_ARGS)
...
+}
+
 Datum
 policy_compression_add(PG_FUNCTION_ARGS)
 {
 	NameData application_name;
 	NameData proc_name, proc_schema, owner;
 	int32 job_id;
-	Oid ht_oid = PG_GETARG_OID(0);
+	Oid user_oid = PG_GETARG_OID(0);
```
mfundul, line 186: You should add some comments about what this argument can or cannot be.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -340,3 +373,76 @@ policy_compression_remove(PG_FUNCTION_ARGS)
...
+{
+	ContinuousAggHypertableStatus status;
+	Hypertable *ht = ts_hypertable_cache_get_entry(hcache, user_htoid, true /* missing_ok */);
+	*is_cagg = false;
+
+	if (ht != NULL)
+	{
+		if (!TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht))
+		{
+			ts_cache_release(hcache);
```
mfundul, line 394: Why do we need all those `ts_cache_release()` statements? Don't the exceptions clean up all the process state?

sql/updates/reverse-dev.sql
```diff
@@ -0,0 +1,2 @@
+-- revert changes to continuous aggregates view definition
+DROP VIEW IF EXISTS timescaledb_information.continuous_aggregates;
```
mfundul, line 2: I think in this case we cannot downgrade. If there are compressed materialized hypertables and we downgrade then we break the installation.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -276,6 +295,20 @@ policy_compression_add(PG_FUNCTION_ARGS)
 							CONFIG_KEY_COMPRESS_AFTER,
 							format_type_be(compress_after_type))));
 	}
+	/* If this is a compression policy for a cagg, verify that
+	 * compress_after > refresh_start of cagg policy. We do not want
+	 * to compress regions that can eb refreshed by the cagg policy.
```
mfundul, line 300:
```suggestion
	 * to compress regions that can be refreshed by the cagg policy.
```

tsl/src/bgw_policy/compression_api.c
```diff
@@ -276,6 +295,20 @@ policy_compression_add(PG_FUNCTION_ARGS)
...
+	 */
+	if (is_cagg && !policy_refresh_cagg_refresh_start_lt(hypertable->fd.id,
```
mfundul, line 302: What about changing the CAGG policy? Shouldn't there be a comparison against `compress_after` there as well for the same reason?

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -132,6 +133,82 @@ policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config)
...
+		bool found;
+		Assert(IS_INTEGER_TYPE(cmp_type));
+		int64 cmpval = ts_interval_value_to_internal(cmp_interval, cmp_type);
+		int64 refresh_start =
+			ts_jsonb_get_int64_field(cagg_config, CONFIG_KEY_START_OFFSET, &found);
+		if (!found) /*this is a null value */
+			return false;
+		Datum res =
+			DirectFunctionCall2(int8lt, Int64GetDatum(refresh_start), Int64GetDatum(cmpval));
+		ret = DatumGetBool(res);
```
mfundul, line 197: Why can't we do this here?
```
ret = refresh_start < cmpval;
```

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +66,163 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
 	ts_scan_iterator_close(&iterator);
 }
 
+/*
+ * Retrieve the cagg view query and find the groupby clause and
+ * time_bucket clause.Map them to the column names(of mat.hypertable)
```
mfundul, line 71:
```suggestion
 * time_bucket clause. Map them to the column names(of mat.hypertable)
```

src/chunk.c
```diff
@@ -3946,9 +3946,10 @@ list_return_srf(FunctionCallInfo fcinfo)
  *
  * Will error if relid is not a hypertable or view (or cannot be found) or if
  * it is a materialized hypertable.
+ * allow_matht permits materialized hypertables and will not error out
```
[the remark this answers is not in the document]
gayyappan: modified comment.

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +66,163 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+										 "compress_orderby",
+										 (Node *) makeString((char *) mat_ht_timecolname),
+										 DEFELEM_UNSPEC,
+										 -1);
+	defelems = lappend(defelems, ordby);
+	List *grp_colnames = cagg_find_groupingcols(agg, mat_ht);
+	if (grp_colnames)
+	{
+		ListCell *lc;
+		/* we have column names. they are guaranteed to be atmost
```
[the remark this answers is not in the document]
gayyappan: fixed.

tsl/src/bgw_policy/compression_api.c
gayyappan: fixed.

tsl/src/continuous_aggs/options.c
gayyappan: fixed.

tsl/src/bgw_policy/compression_api.c
gayyappan: added

tsl/src/bgw_policy/compression_api.c
gayyappan: Yes, the supported apis are to add/remove cagg policies.

tsl/src/bgw_policy/continuous_aggregate_api.c
gayyappan: int64 representation differences on platforms

mfundul: It's also in `int8l` implementation

src/process_utility.c
```diff
@@ -187,6 +187,7 @@ check_continuous_agg_alter_table_allowed(Hypertable *ht, AlterTableStmt *stmt)
 		{
 			case AT_AddIndex:
 			case AT_ReAddIndex:
+			case AT_SetRelOptions:
```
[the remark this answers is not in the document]
gayyappan: done.

sql/updates/reverse-dev.sql
```diff
@@ -75,3 +75,29 @@ BEGIN
...
+  cagg_name NAME;
+  matht_name NAME;
+  matht_schema NAME;
+BEGIN
+    FOR cagg_name , matht_schema, matht_name IN
+        SELECT view_name, 
+               materialization_hypertable_schema ,
+               materialization_hypertable_name   
+        FROM timescaledb_information.continuous_aggregates
+        WHERE compression_enabled is TRUE
```
fabriziomello, line 94: IMHO we can save all caggs with compression in an array and then show to the user the complete list, something that we did here already: https://github.com/timescale/timescaledb/commit/958040699c4986a98fab788bafd599d24281f1d8#diff-328ad2a463009f63aded2e997441ab13d9f7e725b14cc68e68fb807a78984e7bR10

sql/views.sql
```diff
@@ -113,6 +113,10 @@ SELECT ht.schema_name AS hypertable_schema,
   cagg.user_view_name AS view_name,
   viewinfo.viewowner AS view_owner,
   cagg.materialized_only,
+  CASE WHEN mat_ht.compressed_hypertable_id is NOT NULL
```
fabriziomello, line 116:
```suggestion
  CASE WHEN mat_ht.compressed_hypertable_id IS NOT NULL 
```

sql/views.sql
```diff
@@ -113,6 +113,10 @@ SELECT ht.schema_name AS hypertable_schema,
   cagg.user_view_name AS view_name,
   viewinfo.viewowner AS view_owner,
   cagg.materialized_only,
+  CASE WHEN mat_ht.compressed_hypertable_id is NOT NULL 
+       THEN TRUE 
+       ELSE FALSE 
+  END as compression_enabled,
```
fabriziomello, line 119:
```suggestion
  END AS compression_enabled,
```

tsl/src/bgw_policy/compression_api.c
fabriziomello: I also got a bit confused because we register the XactCallback here: https://github.com/timescale/timescaledb/blob/master/src/cache.c#L362

sql/updates/reverse-dev.sql
gayyappan: modified to block downgrade for this case.

sql/updates/reverse-dev.sql
```diff
@@ -75,3 +75,28 @@ BEGIN
...
+    FOR cagg_name IN
+        SELECT view_name, 
+               materialization_hypertable_schema ,
+               materialization_hypertable_name   
+        FROM timescaledb_information.continuous_aggregates
+        WHERE compression_enabled is TRUE
+    LOOP
+        RAISE NOTICE 'Compression is enabled for continuous aggregate: %', cagg_name; 
+    END LOOP;
+    RAISE EXCEPTION 'Cannot downgrade as compression is enabled for continuous aggregates'
```
fabriziomello, line 96: I think this new logic is broken because you always issue an exception after the loop... actually you should raise the exception only it some cagg with compression is found on the previous loop, no?

scripts/test_updates_pg12.sh
```diff
@@ -20,3 +20,5 @@ run_tests "$@" -v7 \
           2.0.0-rc2-pg12 2.0.0-rc3-pg12 2.0.0-rc4-pg12 2.0.0-pg12 2.0.1-pg12 2.0.2-pg12 2.1.0-pg12 \
           2.1.1-pg12 2.2.0-pg12 2.2.1-pg12 2.3.0-pg12 2.3.1-pg12 2.4.0-pg12 2.4.1-pg12 2.4.2-pg12 \
           2.5.0-pg12 2.5.1-pg12
```
svenklemm, line 22:
```suggestion
```
v8 should be a superset of v7 and if it works starting at 2.5.0 then  2.5+ should be removed from v7 here.

scripts/test_updates_pg13.sh
```diff
@@ -10,3 +10,6 @@ source ${SCRIPT_DIR}/test_functions.inc
 run_tests -v7 \
           2.1.0-pg13 2.1.1-pg13 2.2.0-pg13 2.2.1-pg13 2.3.0-pg13 2.3.1-pg13 \
           2.4.0-pg13 2.4.1-pg13 2.4.2-pg13 2.5.0-pg13 2.5.1-pg13
```
svenklemm, line 12:
```suggestion
          2.4.0-pg13 2.4.1-pg13 2.4.2-pg13
```
v8 should be a superset of v7 and if it works starting at 2.5.0 then  2.5+ should be removed from v7 here.

sql/updates/reverse-dev.sql
gayyappan: You are too quick with the reviews :-) Just caught that with my local testing and modified.

scripts/test_updates_pg14.sh
```diff
@@ -9,3 +9,5 @@ source ${SCRIPT_DIR}/test_functions.inc
 
 run_tests -v7 \
           2.5.0-pg14 2.5.1-pg14
```
svenklemm, line 11:
```suggestion
run_tests -v7 \
          2.5.0-pg14 2.5.1-pg14
```
v8 should be a superset of v7 and if it works starting at 2.5.0 then  2.5+ should be removed from v7 here.

scripts/test_updates_pg12.sh
svenklemm, line 22:
```suggestion
          2.1.1-pg12 2.2.0-pg12 2.2.1-pg12 2.3.0-pg12 2.3.1-pg12 2.4.0-pg12 2.4.1-pg12 2.4.2-pg12
```
v8 should be a superset of v7 and if it works starting at 2.5.0 then  2.5+ should be removed from v7 here.

scripts/test_updates_pg12.sh
```diff
@@ -20,3 +20,5 @@ run_tests "$@" -v7 \
...
+run_tests "$@" -v8 \
+          2.5.0-pg12 2.5.1-pg12
```
svenklemm, line 24: Since you are adding a test for enabling compression on cagg which is only possible after this PR is merged the earliest v8 can run is 2.6.0

tsl/src/bgw_policy/compression_api.c
```diff
@@ -297,17 +331,38 @@ policy_compression_add(PG_FUNCTION_ARGS)
...
 
 	TS_PREVENT_FUNC_IF_READ_ONLY();
 
-	ht = ts_hypertable_cache_get_cache_and_entry(hypertable_oid, CACHE_FLAG_NONE, &hcache);
+	ht = ts_hypertable_cache_get_cache_and_entry(user_oid, CACHE_FLAG_MISSING_OK, &hcache);
+	if (!ht)
+	{
+		const char *view_name = get_rel_name(user_oid);
+
+		if (!view_name)
```
afiskon, line 350: If possible, it would be nice to add a test case to cover this code path.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -297,17 +331,38 @@ policy_compression_add(PG_FUNCTION_ARGS)
...
+			ereport(ERROR,
```
afiskon, line 351: If possible, it would be nice to add a test case to cover this code path.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -297,17 +331,38 @@ policy_compression_add(PG_FUNCTION_ARGS)
...
+					(errcode(ERRCODE_UNDEFINED_OBJECT),
+					 errmsg("relation is not a hypertable or continuous aggregate")));
+		else
+		{
+			ContinuousAgg *ca = ts_continuous_agg_find_by_relid(user_oid);
+
+			if (!ca)
+				ereport(ERROR,
```
afiskon, line 359: If possible, it would be nice to add a test case to cover this code path.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -297,17 +331,38 @@ policy_compression_add(PG_FUNCTION_ARGS)
 	PG_RETURN_INT32(job_id);
 }
 
+/* remove compression policy from ht or cagg */
 Datum
 policy_compression_remove(PG_FUNCTION_ARGS)
 {
-	Oid hypertable_oid = PG_GETARG_OID(0);
+	Oid user_oid = PG_GETARG_OID(0);
```
afiskon, line 338: I find this variable name very confusing. This doesn't seem to be an OID of a user. Could we come up with a better name?

tsl/src/bgw_policy/compression_api.c
```diff
@@ -340,3 +395,76 @@ policy_compression_remove(PG_FUNCTION_ARGS)
...
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("compression not enabled on hypertable \"%s\"",
+							get_rel_name(user_htoid)),
+					 errhint("Enable compression before adding a compression policy.")));
+		}
+		status = ts_continuous_agg_hypertable_status(ht->fd.id);
+		if ((status == HypertableIsMaterialization || status == HypertableIsMaterializationAndRaw))
+		{
+			ts_cache_release(hcache);
+			ereport(ERROR,
```
afiskon, line 427: The code path needs a corresponding test case, if it's possible to add one.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -340,3 +395,76 @@ policy_compression_remove(PG_FUNCTION_ARGS)
...
+		/*check if this is a cont aggregate view */
+		int32 mat_id;
+		bool found;
+
+		ContinuousAgg *cagg = ts_continuous_agg_find_by_relid(user_htoid);
+
+		if (cagg == NULL)
+		{
+			ts_cache_release(hcache);
+			ereport(ERROR,
```
afiskon, line 446: The code path needs a corresponding test case, if it's possible to add one.

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -132,6 +133,80 @@ policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config)
...
 }
 
+/* returns false if a policy could not be found */
+bool
+policy_refresh_cagg_exists(int32 materialization_id)
+{
+	Hypertable *mat_ht = ts_hypertable_get_by_id(materialization_id);
+
+	if (!mat_ht)
+		ereport(ERROR,
```
afiskon, line 143: The code path needs a corresponding test case, if it's possible to add one.

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -132,6 +133,80 @@ policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config)
...
+ * of cagg policy.
+ */
+bool
+policy_refresh_cagg_refresh_start_lt(int32 materialization_id, Oid cmp_type, Datum cmp_interval)
+{
+	Hypertable *mat_ht = ts_hypertable_get_by_id(materialization_id);
+	bool ret = false;
+
+	if (!mat_ht)
+		ereport(ERROR,
```
afiskon, line 169: The code path needs a corresponding test case, if it's possible to add one.

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -132,6 +133,80 @@ policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config)
...
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("configuration materialization hypertable id %d not found",
+						materialization_id)));
+
+	List *jobs = ts_bgw_job_find_by_proc_and_hypertable_id(POLICY_REFRESH_CAGG_PROC_NAME,
+														   INTERNAL_SCHEMA_NAME,
+														   materialization_id);
+	if (jobs == NIL)
+		return false;
```
afiskon, line 178: The code path needs a corresponding test case, if it's possible to add one. Otherwise, please add a comment that it should never happen in practice and is needed only for debugging purposes.

tsl/src/bgw_policy/continuous_aggregate_api.c
```diff
@@ -132,6 +133,80 @@ policy_refresh_cagg_get_refresh_end(const Dimension *dim, const Jsonb *config)
...
+	Oid dim_type = ts_dimension_get_partition_type(open_dim);
+	if (IS_INTEGER_TYPE(dim_type))
+	{
+		bool found;
+		Assert(IS_INTEGER_TYPE(cmp_type));
+		int64 cmpval = ts_interval_value_to_internal(cmp_interval, cmp_type);
+		int64 refresh_start =
+			ts_jsonb_get_int64_field(cagg_config, CONFIG_KEY_START_OFFSET, &found);
+		if (!found) /*this is a null value */
+			return false;
```
afiskon, line 194: The code path needs a corresponding test case, if it's possible to add one. Otherwise, please add a comment that it should never happen in practice and is needed only for debugging purposes.

tsl/src/bgw_policy/policy_utils.c
```diff
@@ -100,3 +102,27 @@ subtract_interval_from_now(Interval *lag, Oid time_dim_type)
...
+	{
+		/* if this a materialization hypertable related to cont agg
+		 * then need to get the right dimension which has
+		 * integer_now function
+		 */
+
+		open_dim = ts_continuous_agg_find_integer_now_func_by_materialization_id(mat_id);
+		if (open_dim == NULL)
+		{
+			elog(ERROR,
```
afiskon, line 122: The code path needs a corresponding test case, if it's possible to add one. Otherwise, please add a comment that it should never happen in practice and is needed only for debugging purposes.

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +66,163 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+	ListCell *lc;
+	Oid cagg_view_oid =
+		get_relname_relid(NameStr(agg->data.user_view_name),
+						  get_namespace_oid(NameStr(agg->data.user_view_schema), false));
+	Relation cagg_view_rel = table_open(cagg_view_oid, AccessShareLock);
+	RuleLock *cagg_view_rules = cagg_view_rel->rd_rules;
+	Assert(cagg_view_rules && cagg_view_rules->numLocks == 1);
+	RewriteRule *rule = cagg_view_rules->rules[0];
+	if (rule->event != CMD_SELECT)
+		elog(ERROR, "rule missing for view ");
```
afiskon, line 87: The code path needs a corresponding test case, if it's possible to add one. Otherwise, please add a comment that it should never happen in practice and is needed only for debugging purposes.

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +66,163 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+	/* the view rule has dummy old and new range table entries as the 1st and 2nd entries
+	 */
+	Assert(list_length(cagg_view_query->rtable) >= 2);
+	if (cagg_view_query->setOperations)
+	{
+		/* This corresponds to the union view.
+		 * the 3rd RTE entry has the SELECT 1 query from the union view. */
+		RangeTblEntry *orig_query_rte = lthird(cagg_view_query->rtable);
+		if (orig_query_rte->rtekind != RTE_SUBQUERY)
+			elog(ERROR, "unexpected view rule action ");
```
afiskon, line 102: The code path needs a corresponding test case, if it's possible to add one. Otherwise, please add a comment that it should never happen in practice and is needed only for debugging purposes.

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +66,163 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+			}
+			collen = strlen(grpcol);
+			if (seglen - segidx > collen)
+			{
+				strlcpy(segmentby + segidx, grpcol, collen + 1);
+				segidx = segidx + collen;
+			}
+			else
+			{
+				ereport(ERROR,
```
afiskon, line 164: The code path needs a corresponding test case, if it's possible to add one. Otherwise, please add a comment that it should never happen in practice and is needed only for debugging purposes.

tsl/test/sql/include/continuous_aggs_ddl_common.sql
```diff
@@ -1107,3 +1107,98 @@ AS
 
 ALTER MATERIALIZED VIEW i3696_cagg2 SET (timescaledb.materialized_only = 'true');
 
+--TEST test with multiple settings on continuous aggregates --
+-- test for materialized_only + compress combinations (real time aggs enabled initially)
+CREATE TABLE test_setting(time timestamptz not null, val numeric);
+\if :IS_DISTRIBUTED
+SELECT create_distributed_hypertable('test_setting', 'time', replication_factor => 2);
```
afiskon, line 1114: Since https://github.com/timescale/timescaledb/pull/3753 was merged before this patch I'm afraid we have to add a couple of more tests:
- Compression of CAGGs with variable-sized buckets
- Compression of CAGGs with variable-sized buckets on top of distributed HT

If it doesn't work it's fine just to make sure that the user will see an adequate error message and add corresponding support to the backlog.

afiskon (changes requested):
Hey, @gayyappan excellent work! Please see my **preliminary** code review. My main concern so far is that the feature doesn't seem to be well test-covered. Please see my comments. I don't insist on covering the ereport() code paths that can't happen in practice, but in this case please add a corresponding comment.

scripts/test_updates_pg12.sh
gayyappan: The test is in post.continuous_aggs - checks if we can enable compression on an existing  cagg , after upgrading to the latest version.Removed -v7 test from 2.5.0 and 2.5.1 since v8 is a superset of v7.

tsl/test/sql/include/continuous_aggs_ddl_common.sql
gayyappan: I added a basic test to the new test file. But could you elaborate why we need a special test case for caggs with variable sized buckets? Compression is applied on the underlying storage i.e. on the materialized hypertable. time_bucket_ng is just another function used in the definition of the view. All the code in this PR is syntactic sugar so that users don't have to enable compression directly on the materialized hypertable.

tsl/src/continuous_aggs/options.c
gayyappan: Isn't that obvious from the error message that this is unexpected type of query?

gayyappan: used different error code to make it even more obvious.

tsl/src/continuous_aggs/options.c
gayyappan: modified error to use unexpected error code to make it obvious that it is not possible.

tsl/test/sql/include/continuous_aggs_ddl_common.sql
afiskon: @gayyappan thanks! If we have a feature (compression of CAGGs) that works with N other features (CAGGs with variable buckets, CAGGs with fixed buckets, CAGGs on top of distributed HT) we have to make sure that they work together as expected (on all platforms, with different compile flags, etc). Otherwise, the code coverage is simply inadequate, although maybe it looks OK in terms of LoC. This allows to discover bugs early (before the support team discovers them) and simplifies the maintenance of the project in the long term.

tsl/src/bgw_policy/compression_api.c
mfundul: Agreed, it's also used in `policy_compression_add()`. Maybe `rel_oid` or `user_rel_oid`?

tsl/src/bgw_policy/policy_utils.c
mfundul: To be fair this code was always there, just moved around.

tsl/test/sql/include/continuous_aggs_ddl_common.sql
gayyappan: The point is caggs with variable size buckets or fixed size buckets are internals specific to the time_bucket function implementation and should affect only bucket computation logic (or places where bucket computation logic is needed). It does not affect anything related to how the materialized data is stored. As I stated in my previous comment, I added a test for sanity checking.

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +67,166 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+		if (orig_query_rte->rtekind != RTE_SUBQUERY)
+			ereport(ERROR,
+					(errcode(ERRCODE_TS_UNEXPECTED),
+					 errmsg("unexpected rte type for view %d", orig_query_rte->rtekind)));
+
+		orig_query = orig_query_rte->subquery;
+	}
+	else
+	{
+		orig_query = cagg_view_query;
```
mfundul, line 111: `orig_query` is confusing me because it reminds me of the direct view. Maybe rephrase?

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +67,166 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+		{
+			char *grpcol = (char *) lfirst(lc);
+			/* skip time dimension col if it appears in group-by list */
+			if (namestrcmp((Name) & (mat_ht_dim->fd.column_name), grpcol) == 0)
+				continue;
+			int collen = 1;
+			if (segidx > 0 && (seglen - segidx) > collen)
+			{
+				strlcpy(segmentby + segidx, ",", collen + 1);
+				segidx = segidx + 1;
```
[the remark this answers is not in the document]
gayyappan: I prefer 1. I think of collen as length of col and find it misleading.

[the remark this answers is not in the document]
mfundul: Then you should remove:
```
 			int collen = 1;
```
and do:
```
if (segidx > 0 && (seglen - segidx) > 1)
```
```
	strlcpy(segmentby + segidx, ",", 1 + 1);
```
cause you use it twice here.

tsl/src/continuous_aggs/options.c
```diff
@@ -81,7 +240,18 @@ continuous_agg_update_options(ContinuousAgg *agg, WithClauseResult *with_clause_
 		update_materialized_only(agg, agg->data.materialized_only);
 		ts_cache_release(hcache);
 	}
+	if (!with_clause_options[ContinuousViewOptionCompress].is_default)
+	{
+		bool compress_enable =
+			DatumGetBool(with_clause_options[ContinuousViewOptionCompress].parsed);
```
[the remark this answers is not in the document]
gayyappan: The compression code handles this. You can enable/disable as long as there are no compressed chunks.

tsl/src/continuous_aggs/options.c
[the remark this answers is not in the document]
gayyappan: used 1 instead of collen for this block.

tsl/src/continuous_aggs/options.c
gayyappan: renamed as finalize_query

gayyappan:
> Hey, @gayyappan excellent work! Please see my **preliminary** code review. My main concern so far is that the feature doesn't seem to be well test-covered. Please see my comments. I don't insist on covering the ereport() code paths that can't happen in practice, but in this case please add a corresponding comment.

Added code coverage tests based on code cov report. Used ERRCODE_TS_UNEXPECTED_ERROR for unexpected errors.

tsl/src/bgw_policy/compression_api.c
gayyappan: renamed as user_rel_oid. I used user to indicate that this is user supplied oid.  I do see  that it might be interpreted as user id.

tsl/src/bgw_policy/continuous_aggregate_api.c
gayyappan: This is an API function - it is not possible to reach this path from the current callers. But possible if it is called incorrectly.

tsl/src/bgw_policy/compression_api.c
mfundul: Can you rename this one to `user_rel_oid` as well?

tsl/src/bgw_policy/compression_api.c
mfundul: I don't see the rename. Can you rename this for `policy_compression_add()` as well?

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +67,170 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+		int seglen = ((NAMEDATALEN + 1) * list_length(grp_colnames)) + 1;
+		char *segmentby = (char *) palloc(seglen);
+		int segidx = 0;
+		foreach (lc, grp_colnames)
+		{
+			char *grpcol = (char *) lfirst(lc);
+			/* skip time dimension col if it appears in group-by list */
+			if (namestrcmp((Name) & (mat_ht_dim->fd.column_name), grpcol) == 0)
+				continue;
+			int collen = 1;
```
mfundul, line 158: This variable doesn't need an initialization anymore. Maybe move the definition to line 164?

mfundul (approved):
Just some nits about code style cleanup pending.

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +67,170 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+						  get_namespace_oid(NameStr(agg->data.user_view_schema), false));
+	Relation cagg_view_rel = table_open(cagg_view_oid, AccessShareLock);
+	RuleLock *cagg_view_rules = cagg_view_rel->rd_rules;
+	Assert(cagg_view_rules && cagg_view_rules->numLocks == 1);
+	RewriteRule *rule = cagg_view_rules->rules[0];
+	if (rule->event != CMD_SELECT)
+		ereport(ERROR, (errcode(ERRCODE_TS_UNEXPECTED), errmsg("unexpected rule event for view")));
+
+	Query *cagg_view_query = copyObject(linitial(rule->actions));
+	table_close(cagg_view_rel, NoLock); // lock with be released at end of txn
```
afiskon, line 95: Oops, I don't think we are using single-line comments in this project

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +67,170 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+		Assert(IsA(cagg_tle->expr, Var));
+		Var *mat_var = castNode(Var, cagg_tle->expr);
+		char *mat_colname = get_attname(mat_relid, mat_var->varattno, false);
+		retlist = lappend(retlist, mat_colname);
+	}
+	return retlist;
+}
+
+static List *
+cagg_get_compression_params(ContinuousAgg *agg, Hypertable *mat_ht)
```
afiskon, line 131: Adding a short comment would be nice.

tsl/src/continuous_aggs/options.c
afiskon: Since the CI is not complaining anymore, I assume it was fixed.

tsl/src/bgw_policy/compression_api.c
gayyappan: was not pushed.

tsl/src/bgw_policy/compression_api.c
gayyappan: done.

tsl/test/sql/include/continuous_aggs_ddl_common.sql
afiskon: The point is, unless we have at least the simplest tests, it's not proven that named functionalities are working together. Also, we will not catch it in the future if they stop working together.

afiskon (dismissed):

sql/updates/latest-dev.sql
```diff
@@ -24,3 +23,5 @@ SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs
 -- Adding overloaded versions of invalidation_process_hypertable_log() and invalidation_process_cagg_log()
 -- with bucket_functions argument is done in cagg_utils.sql. Note that this file is included when building
 -- the update scripts, so we don't have to do it here.
+
+DROP VIEW IF EXISTS timescaledb_information.continuous_aggregates;
```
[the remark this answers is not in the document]
gayyappan: because we are changing the defn. It gets recreated as part of extension update.

sql/updates/reverse-dev.sql
```diff
@@ -75,3 +75,32 @@ BEGIN
...
+    END LOOP;
+    IF cnt > 0 THEN
+       RAISE EXCEPTION 'Cannot downgrade as compression is enabled for continuous aggregates' 
+            USING DETAIL = 'Please disable compression on all continuous aggregates before downgrading',
+            HINT = ('To disable compression, call decompress_chunk to decompress chunks, then drop any existing compression policy on the continuous aggregate, and finally run ALTER MATERIALIZED VIEW % SET timescaledb.compress = ''false'' ');
+    END IF;
+END $$;
+
+-- revert changes to continuous aggregates view definition
+DROP VIEW IF EXISTS timescaledb_information.continuous_aggregates;
```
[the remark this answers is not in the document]
gayyappan: same reason as changes to latest-dev.sql

tsl/src/bgw_policy/compression_api.c
fabriziomello: IMHO the better we can do here is just remove those `ts_release_cache` lines

sql/updates/reverse-dev.sql
```diff
@@ -75,3 +75,32 @@ BEGIN
...
+               materialization_hypertable_schema ,
+               materialization_hypertable_name   
+        FROM timescaledb_information.continuous_aggregates
+        WHERE compression_enabled is TRUE
+    LOOP
+        RAISE NOTICE 'Compression is enabled for continuous aggregate: %', cagg_name;
+        cnt := cnt + 1; 
+    END LOOP;
+    IF cnt > 0 THEN
+       RAISE EXCEPTION 'Cannot downgrade as compression is enabled for continuous aggregates'
```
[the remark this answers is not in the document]
gayyappan: The downgrade test runs as a cron job as a github action. Its not clear to me what versions we are testing against and if a downgrade test can be added for this feature before 2.6. is released. I'll defer this until Mats is back - he added the downgrade testing script. I don't want to risk breaking master.

tsl/src/continuous_aggs/options.c
```diff
@@ -62,12 +67,170 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
...
+	defelems = lappend(defelems, ordby);
+	List *grp_colnames = cagg_find_groupingcols(agg, mat_ht);
+	if (grp_colnames)
+	{
+		ListCell *lc;
+		/* we have column names. they are guaranteed to be at most
+		 * NAMEDATALEN
+		 */
+		int seglen = ((NAMEDATALEN + 1) * list_length(grp_colnames)) + 1;
+		char *segmentby = (char *) palloc(seglen);
```
fabriziomello, line 150: I'm curious why do not use `StringInfo` instead of C string manipulation???

tsl/test/expected/continuous_aggs_policy.out
```diff
@@ -450,9 +450,6 @@ WARNING:  continuous aggregate policy already exists for "mat_smallint"
 (1 row)
 
 \set ON_ERROR_STOP 1
-DROP MATERIALIZED VIEW mat_smallint;
-NOTICE:  drop cascades to 3 other objects
-DROP TABLE smallint_tab CASCADE;
 --bigint table
```
fabriziomello, line 453: We have tests for `SMALLINT (int2)` and `BIGINT (int8)`, shouldn't also have for `INTEGER (int4)` ??

fabriziomello (approved):
Despite minor comments I've made LGTM

tsl/src/continuous_aggs/options.c
gayyappan: StringInfo is overkill for this. There is no buffer resizing needed. Also, at the end we need the char* representation - the char * buffer gets passed to makeString for creating the DefElem.

tsl/test/expected/continuous_aggs_policy.out
gayyappan: These tests are for code coverage only (policy_refresh_cagg_refresh_start_lt).  IIRC with the current code, one of the integer types is sufficient. Compression policies should be tested with both integer based and timestamp based materialized views. We have both kinds.

sql/views.sql
gayyappan: done.

sql/views.sql
gayyappan: done.

tsl/src/bgw_policy/compression_api.c
gayyappan: We have this usage all over the code base. Don't see a reason to add a special comment here

afiskon:
@gayyappan @fabriziomello  @mfundul  I dismissed my own review since the PR already got two approvals. This being said, I checked `git fetch timescale && git diff timescale/master timescale/cagg_compression` again. I looked for `time_bucket_ng` and didn't find any use of it in the PR. Thus it looks like the PR still doesn't check if CAGG compression works with all other CAGGs features, OR that the user gets appropriate error messages if it doesn't work.

In other words, some of my review comments were ignored, which I find frustrating. I will not block the PR if the team consensus is not to add corresponding tests, but personally, I find this practice concerning and inefficient in the long term.

gayyappan:
> @gayyappan @fabriziomello @mfundul I dismissed my own review since the PR already got two approvals. This being said, I checked `git fetch timescale && git diff timescale/master timescale/cagg_compression` again. I looked for `time_bucket_ng` and didn't find any use of it in the PR. Thus it looks like the PR still doesn't check if CAGG compression works with all other CAGGs features (monthly buckets in this case; didn't check for CAGGs on top of distributed HT, etc), OR that the user gets appropriate error messages if it doesn't work.
> 
> I will not block the PR if the team consensus is not to add corresponding tests, but personally, I find this practice concerning and inefficient in the long term.

The sanity check test is in tsl/test/sql/continuous_aggs_variable_size_buckets.sql

afiskon:
@gayyappan You are right 😅 Thanks!

afiskon (approved):

sql/updates/reverse-dev.sql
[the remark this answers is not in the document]
mkindahl: The downgrade tests are quite simplistic at this point. They basically run the following sequence:

1. Create the lower version X
2. Run the post scripts and dump them.
2. Upgrade to higher version Y
3. Immediately downgrade to version X again.
4. Run the post scripts and make sure that the results match what was there before.

To run this kind of test, we need two things:

- Add the ability to run some things after the upgrade. This was not added before since we did not intend to support anything but an immediate downgrade. If you make changes, to the installation after that, we cannot guarantee that a downgrade will work, so we are not testing this.
- Add the ability to check for the failure case.
Both of these are possible, but it is quite a lot of work in the current state.

PR #3875, opened 2021-12-01
Fix #3874 invalidate cache in parallel worker to avoid data corruption with VACUUM FULL

### LINKED ISSUE #3874: [Bug]: TimescaleDB corrupt catalog tables
### What type of bug is this?

Data corruption

Data node

### What happened?

#2680 report the bug of catalog table corruption and [3a8f396bc7d014a2722b8113236b34b0eda73eec](https://github.com/timescale/timescaledb/commit/3a8f396bc7d014a2722b8113236b34b0eda73eec) solved it. But it only discuss the situation with preload timescaledb.
When I create extension timescaledb and running the following script, the catalog table would be corrupted again.
We believe the corruption happens in the ParallelWorker situation. 
The function call is like this:

> ParallelWorker

    -> InvalidateSystemCaches
        -> RelationCacheInvalidate
            -> cache_invalidate_callback
                -> ts_catalog_get
                    -> get_namespace_oid
                        -> GetSysCacheOid1 (this step reproduce system cache)
                        -> other functions
If we call VACUUM FULL, parallel workers will actually do the things. If the above code reproduce cache, parallel workers will delete catalog because of the incorrect cache.

2.5.0

### PostgreSQL version used

13.5

RedHat 7u2 3.10.0-327

Source

On prem/Self-hosted

_No response_

```bash
#!/bin/sh
set +e
exitcode=0
datapath=invalidcache_data
logpath=invalidcache_log

export PGDATABASE=bug2680
export PGPORT=12345
rm -rf $datapath $logpath
initdb -D $datapath
cp test/postgresql.conf $datapath/
pg_ctl -D $datapath -l $logpath start

columnnum=1000
partnum=100
tablenum=10

for j in $(seq 0 1)
do
  psql -c "drop database if exists $PGDATABASE" postgres
  psql -c "create database $PGDATABASE" postgres
  if [[ "$j" = "1" ]]; then
    psql -c "create extension if not exists timescaledb"
  fi
  columndef=""
  for i in $(seq 0 $[$columnnum-1])
  do
    columndef="$columndef,c$i int"
  done

  echo "create tables"
  for i in $(seq 0 $[$tablenum-1])
  do
    psql -c "drop table if exists userinfo$i" > /dev/null 2>&1
    psql -c "CREATE TABLE userinfo$i(id int$columndef) PARTITION BY HASH(id);" > /dev/null 2>&1
    for j in $(seq 0 $[$partnum-1])
    do
      psql -c "CREATE TABLE userinfo${i}_$j PARTITION of userinfo$i FOR VALUES WITH (MODULUS $partnum, REMAINDER $j);" > /dev/null 2>&1
    done
    wait `(jobs -p)`
  done

  for i in $(seq 0 1)
  do
    psql -c "vacuum full analyze pg_attribute;" && psql -c "vacuum full;"
    if [[ "$?" -ne "0" ]];then
      exitcode=1
      break
    fi
  done

  if [[ "$exitcode" = "1" ]]; then
    break
  fi
done

if [[ "$1" = "check" ]];then
  pg_ctl -D $datapath stop
fi

exit $exitcode
```

markrui3:
Fix #3874, add extra RelationCacheInvalidate in parallel worker, to avoid data corruption with VACUUM FULL.

### DIFF
--- a/src/cache_invalidate.c
+++ b/src/cache_invalidate.c
@@ -4,6 +4,7 @@
  * LICENSE-APACHE for a copy of the license.
  */
 #include <postgres.h>
+#include <access/parallel.h>
 #include <access/xact.h>
 #include <utils/lsyscache.h>
 #include <utils/inval.h>
@@ -64,7 +65,7 @@ cache_invalidate_relcache_all(void)
  * Should route the invalidation to the correct cache.
  */
 static void
-cache_invalidate_callback(Datum arg, Oid relid)
+cache_invalidate_callback_internal(Datum arg, Oid relid)
 {
 	static bool in_recursion = false;
 
@@ -99,6 +100,21 @@ cache_invalidate_callback(Datum arg, Oid relid)
 		if (relid == InvalidOid)
 			cache_invalidate_relcache_all();
 	}
+}
+
+static void
+cache_invalidate_callback(Datum arg, Oid relid)
+{
+	/* original callback */
+	cache_invalidate_callback_internal(arg, relid);
+
+	/* invalidate cache in parallel worker */
+	if (IsParallelWorker())
+#if PG_VERSION_NUM >= 120009
+		RelationCacheInvalidate(false);
+#else
+		RelationCacheInvalidate();
+#endif
 }
 
 /* Registration for given cache ids happens in non-TSL code when the extension


afiskon:
I can confirm that the proposed patch fixes #3874, and seems to pass `make installcheck` both locally and on CI.

afiskon (approved):

mfundul:
I'm sceptical about calling `RelationCacheInvalidate()` in a invalidation callback context. This kind of recursion has created other problems in the past.

mfundul (changes requested):
https://github.com/timescale/timescaledb/pull/3875#issuecomment-984677177

I think we need a solution that does not issue cache invalidations from within a cache invalidation callback context, since we've seen this causing database corruption before. We need to make sure the fix does not do more damage than what it fixes.

markrui3:
> [#3875 (comment)](https://github.com/timescale/timescaledb/pull/3875#issuecomment-984677177)
> 
> I think we need a solution that does not issue cache invalidations from within a cache invalidation callback context, since we've seen this causing database corruption before. We need to make sure the fix does not do more damage than what it fixes.

Could you show the issue that caused database corruption?
In origin code logic, 

```
void
InvalidateSystemCaches(void)
{
	int			i;

	InvalidateCatalogSnapshot();
	ResetCatalogCaches();
	RelationCacheInvalidate();	/* gets smgr and relmap too */

	for (i = 0; i < syscache_callback_count; i++)
	{
		struct SYSCACHECALLBACK *ccitem = syscache_callback_list + i;

		ccitem->function(ccitem->arg, ccitem->id, 0);
	}

	for (i = 0; i < relcache_callback_count; i++)
	{
		struct RELCACHECALLBACK *ccitem = relcache_callback_list + i;

		ccitem->function(ccitem->arg, InvalidOid);
	}
}
```
In my opinion, ccitem->function cannot produce relcahe again, but now timescaledb actually does.

I think the only problem you care is that, alfter I clean relcache again, some varibles that point to the cache may be null pointer. I check the varibles and don't get this problem.

Maybe there is another solution? Like deleting callback as you did in [3a8f396bc7d014a2722b8113236b34b0eda73eec](https://github.com/timescale/timescaledb/commit/3a8f396bc7d014a2722b8113236b34b0eda73eec). Maybe you can give me more details.

mfundul:
@markrui3 what about if you tried this:
```
--- a/src/cache_invalidate.c
+++ b/src/cache_invalidate.c
@@ -68,6 +68,9 @@ cache_invalidate_callback(Datum arg, Oid relid)
 {
        static bool in_recursion = false;
 
+       if (IsParallelWorker())
+               return;
+
        if (ts_extension_invalidate(relid))
        {
                cache_invalidate_relcache_all();
```
Does this fix the problem?

nikkhils:
> @markrui3 what about if you tried this:

@mfundul your code snippet does seem to fix the errors/crashes! I will dig in a little bit more to see how and why it does so.

src/cache_invalidate.c
```diff
@@ -101,6 +102,21 @@ cache_invalidate_callback(Datum arg, Oid relid)
 	}
 }
 
+static void
+cache_invalidate_callback(Datum arg, Oid relid)
+{
+	/* original callback */
+	cache_invalidate_callback_internal(arg, relid);
```
nikkhils, line 109: We have just renamed this function and are still calling it unconditionally. So I fail to see how this is solving the problem at hand.

src/cache_invalidate.c
```diff
@@ -101,6 +102,21 @@ cache_invalidate_callback(Datum arg, Oid relid)
...
+
+	/* invalidate cache in parallel worker */
+	if (IsParallelWorker())
```
nikkhils, line 112: No need to call `RelationCacheInvalidate` again. It will happen via the PG backend invalidation processing automatically.

afiskon:
@markrui3 sorry it took look. The team reached the consensus to decline this particular PR because we believe this is not the best solution and will cause new problems eventually. @mfundul is going to work on solving the problem.

PR #3878, opened 2021-12-01
Use standard random() instead of pg_lrand48()

afiskon:
pg_*rand48() functions family was removed in the upstream and the code that
uses it will not compile against PG15. This patch replaces pg_lrand48() with
standard random() function from stdlib.h. Also the patch rewrites the
bgw_db_scheduler test so that the output doesn't contain any pseudo-random
data.

### DIFF
--- a/src/bgw/job_stat.c
+++ b/src/bgw/job_stat.c
@@ -7,6 +7,7 @@
 #include <access/xact.h>
 #include <utils/fmgrprotos.h>
 
+#include <stdlib.h>
 #include <math.h>
 
 #include "job_stat.h"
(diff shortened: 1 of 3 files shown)


afiskon:
This PR is an alternative version of https://github.com/timescale/timescaledb/pull/3866

akuzm (approved):
So, does this switch from predictable random values to unpredictable ones? I guess that's why you'd want to exclude them from the reference. It would be good to have some way to have predictable random values here (e.g. reset the seed, print it, and allow to reset it externally), so that the test failures are easier to reproduce.

afiskon:
@akuzm I'm more concerned that random() is implemented differently depending on the environment. I proposed to have predictable random values in the alternative PR, but the team doesn't seem to like the idea. In fact, I agree that ideally the PRNG should be covered in the unit tests and mocked in integration tests. Here we are dealing with system tests, that shouldn't test such kind of behavior at all.

Regardless, my primary goal is to support the upcoming PG15 release. The way we test TimescaleDB deserves a separate large discussion.

afiskon:
@akuzm BTW feel free to review the alternative PR. I would appreciate your feedback https://github.com/timescale/timescaledb/pull/3866

mkindahl (approved):
LGTM

PR #3881, opened 2021-12-02
Fix SkipScan crash due to pruned unique path

erimatnor:
When SkipScan paths are added it can result in a crash if the new
SkipScan dominates the PostgreSQL planned unique path. In that case,
the previous unique path is pruned and freed, leading to a potential
memory corruption since a pointer to the original unique path is used
when looping over all subpaths to try to apply SkipScan. Thus, if the
original unique path is pruned, and there are multiple subpaths, a
crash will occur in the next iteration over the subpaths.

The crash is fixed by making a shallow copy of the original unique
path while iterating the subpaths.

### DIFF
--- a/tsl/src/nodes/skip_scan/planner.c
+++ b/tsl/src/nodes/skip_scan/planner.c
@@ -211,11 +211,21 @@ tsl_skip_scan_paths_add(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo *ou
 			break;
 		}
 	}
+
 	/* no UniquePath found so this query might not be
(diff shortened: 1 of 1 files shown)


erimatnor:
I've tried to add a test case but it is very difficult to create on for the regression suite. Basically, these conditions must be fulfilled:

* There must be multiple paths considered for scanning the base relation
* The SkipScan path must dominate a previous unique path so that it is pruned
* The pruned path cannot be an indexpath (they're never pfreed in `add_path`)

pmwkaa (approved):

akuzm (approved):

PR #3894, opened 2021-12-07
Use compat.h version of FirstBootstrapObjectId

afiskon:
The macro FirstBootstrapObjectId was renamed in the upstream. This patch adds
a corresponding FirstBootstrapObjectIdCompat macro to compat.h and uses it
instead of FirstBootstrapObjectId. This will allow us to compile agains PG15.

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=a49d0812


mkindahl (approved):
LGTM

mfundul (approved):

PR #3899, opened 2021-12-08
Fix segfault in Continuous Aggregates

### LINKED ISSUE #3898: [Bug]: Segfault on a select query over a Continuous Aggregate
### What type of bug is this?

Crash, Incorrect result

Continuous aggregate

### What happened?

Doing a select over a Continuous Aggregate crashes even when there are no data in the tables. Depending on the Continuous Aggregate definition there may be no crash but incorrect results.

2.5.1

### PostgreSQL version used

13.5

Ubuntu 20.04

Source

On prem/Self-hosted

```bash
frame #0: 0x00007fff6d35133a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff6d40de60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff6d2d8808 libsystem_c.dylib`abort + 120
    frame #3: 0x000000010b49c33a postgres`ExceptionalCondition(conditionName="attno >= rel->min_attr && attno <= rel->max_attr", errorType="FailedAssertion", fileName="initsplan.c", lineNumber=253) at assert.c:67:2
    frame #4: 0x000000010b16453b postgres`add_vars_to_targetlist(root=0x00007f8d70053838, vars=0x00007f8d70046718, where_needed=0x00007f8d70046770, create_new_ph=true) at initsplan.c:253:4
  * frame #5: 0x000000010b16433c postgres`build_base_rel_tlists(root=0x00007f8d70053838, final_tlist=0x00007f8d700431c0) at initsplan.c:196:3
    frame #6: 0x000000010b169bf4 postgres`query_planner(root=0x00007f8d70053838, qp_callback=(postgres`standard_qp_callback at planner.c:3595), qp_extra=0x00007ffee4ed8c70) at planmain.c:178:2
    frame #7: 0x000000010b16e545 postgres`grouping_planner(root=0x00007f8d70053838, inheritance_update=false, tuple_fraction=5) at planner.c:2059:17
    frame #8: 0x000000010b16b98a postgres`subquery_planner(glob=0x00007f8d7004f240, parse=0x00007f8d7004efa8, parent_root=0x00007f8d70050190, hasRecursion=false, tuple_fraction=5) at planner.c:1015:3
    frame #9: 0x000000010b12c10d postgres`set_subquery_pathlist(root=0x00007f8d70050190, rel=0x00007f8d7004ed90, rti=1, rte=0x00007f8d700112b0) at allpaths.c:2306:17
    frame #10: 0x000000010b12b1bc postgres`set_rel_size(root=0x00007f8d70050190, rel=0x00007f8d7004ed90, rti=1, rte=0x00007f8d700112b0) at allpaths.c:422:5
    frame #11: 0x000000010b127e60 postgres`set_base_rel_sizes(root=0x00007f8d70050190) at allpaths.c:323:3
    frame #12: 0x000000010b127b30 postgres`make_one_rel(root=0x00007f8d70050190, joinlist=0x00007f8d70051348) at allpaths.c:185:2
```

Run the following commands.

```
CREATE TABLE IF NOT EXISTS public.transactions
(
    "time" timestamp with time zone NOT NULL,
    type integer,
    coin integer,
    user_fiat integer,
    fiat_value integer,
    amount integer
);

SELECT create_hypertable('transactions', 'time');

CREATE materialized view cashflows(
    bucket,
    amount,
    cashflow
) WITH (
    timescaledb.continuous,
    timescaledb.materialized_only = true
) AS
SELECT time_bucket ('1 day', time) AS bucket,
    amount,
    amount + sum(fiat_value) AS cashflow
FROM transactions
GROUP BY bucket,
    amount;

SELECT * FROM cashflows;
```

mfundul:
Add the missing variables to the finalization view of Continuous Aggregates and the corresponding columns to the materialization table. Cover the case of targets that contain Aggref nodes and Var nodes that are outside of the Aggref nodes at the same time.
    
Stop rebuilding the Continuous Aggregate view with ALTER MATERIALIZED VIEW. Attempt to repair the view at post-update time instead, and fail gracefully if it is not possible to do so without raw hypertable schema or data modifications.
    
Stop rebuilding the Continuous Aggregate view when switching realtime aggregation on and off. Instead, manipulate the User View by either:
  1. removing the `UNION ALL` right-hand side and the `WHERE` clause when disabling realtime aggregation
  2. adding the Direct View to the right of a `UNION ALL` operator and defining `WHERE` clauses with the relevant watermark checks when enabling realtime aggregation
    
Fixes #3898

### DIFF
--- a/sql/updates/post-update.sql
+++ b/sql/updates/post-update.sql
@@ -6,7 +6,11 @@ DECLARE
  ts_version TEXT;
 BEGIN
     SELECT extversion INTO ts_version FROM pg_extension WHERE extname = 'timescaledb';
-
+    IF ts_version >= '2.7.0' THEN
+            CREATE OR REPLACE PROCEDURE @extschema@.post_update_cagg_try_repair(
+                cagg_view REGCLASS
+            ) AS '@MODULE_PATHNAME@', 'ts_cagg_try_repair' LANGUAGE C;
+    END IF;
     FOR vname, materialized_only IN select format('%I.%I', cagg.user_view_schema, cagg.user_view_name)::regclass, cagg.materialized_only from _timescaledb_catalog.continuous_agg cagg
     LOOP
         -- the cast from oid to text returns
@@ -18,10 +22,16 @@ BEGIN
         -- this case.
         IF ts_version < '2.0.0' THEN
             EXECUTE format('ALTER VIEW %s SET (timescaledb.materialized_only=%L) ', vname::text, materialized_only);
-        ELSE
+        ELSIF ts_version < '2.7.0' THEN
             EXECUTE format('ALTER MATERIALIZED VIEW %s SET (timescaledb.materialized_only=%L) ', vname::text, materialized_only);
+        ELSE
+            SET log_error_verbosity TO VERBOSE;
+            CALL @extschema@.post_update_cagg_try_repair(vname);
         END IF;
     END LOOP;
+    IF ts_version >= '2.7.0' THEN
+            DROP PROCEDURE IF EXISTS @extschema@.post_update_cagg_try_repair;
+    END IF;
     EXCEPTION WHEN OTHERS THEN RAISE;
 END
 $$;
--- a/src/utils.c
+++ b/src/utils.c
@@ -1092,3 +1092,24 @@ ts_get_node_name(Node *node)
 			return psprintf("Node (%d)", nodeTag(node));
 	}
 }
+
+/*
+ * Implementation marked unused in PostgreSQL lsyscache.c
+ */
+int
+ts_get_relnatts(Oid relid)
+{
+	HeapTuple tp;
+	Form_pg_class reltup;
+	int result;
+
+	tp = SearchSysCache1(RELOID, ObjectIdGetDatum(relid));
+	if (!HeapTupleIsValid(tp))
+		return InvalidAttrNumber;
+
+	reltup = (Form_pg_class) GETSTRUCT(tp);
+	result = reltup->relnatts;
+
+	ReleaseSysCache(tp);
+	return result;
+}
--- a/tsl/src/continuous_aggs/create.c
+++ b/tsl/src/continuous_aggs/create.c
@@ -174,7 +174,7 @@ typedef struct CAggTimebucketInfo
 {
 	int32 htid;				/* hypertable id */
 	Oid htoid;				/* hypertable oid */
-	AttrNumber htpartcolno; /*primary partitioning column */
+	AttrNumber htpartcolno; /* primary partitioning column of raw hypertable */
 							/* This should also be the column used by time_bucket */
 	Oid htpartcoltype;
 	int64 htpartcol_interval_len; /* interval length setting for primary partitioning column */
@@ -193,9 +193,20 @@ typedef struct CAggTimebucketInfo
 typedef struct AggPartCxt
 {
 	struct MatTableColumnInfo *mattblinfo;
-	bool addcol;
+	bool added_aggref_col;
+	bool var_outside_of_aggref; /* Set to true when you come across a Var that is not inside an
+								   Aggref node */
 	Oid ignore_aggoid;
 	int original_query_resno;
+	/*
+	 * "Original variables" are the Var nodes of the target list of the original
+	 * CREATE MATERIALIZED VIEW query. "Mapped variables" are the Var nodes of the materialization
+	 * table columns. The partialization query is the one that populates those columns. The
+	 * finalization query should use the "mapped variables" to populate the user view.
+	 */
+	List *orig_vars; /* List of Var nodes that have been mapped to materialization table columns */
+	List *mapped_vars; /* List of Var nodes of the corresponding materialization table columns */
+					   /* orig_vars and mapped_vars lists are mapped 1 to 1 */
 } AggPartCxt;
 
 /* STATIC functions defined on the structs above */
@@ -227,8 +238,9 @@ static bool function_allowed_in_cagg_definition(Oid funcid);
 static Const *cagg_boundary_make_lower_bound(Oid type);
 static Oid cagg_get_boundary_converter_funcoid(Oid typoid);
 static Oid relation_oid(NameData schema, NameData name);
-static Query *build_union_query(CAggTimebucketInfo *tbinfo, MatTableColumnInfo *mattblinfo,
-								Query *q1, Query *q2, int materialize_htid);
+static Query *build_union_query(CAggTimebucketInfo *tbinfo, int matpartcolno, Query *q1, Query *q2,
+								int materialize_htid);
+static Query *destroy_union_query(Query *q);
 
 /* create a entry for the materialization table in table CONTINUOUS_AGGS */
 static void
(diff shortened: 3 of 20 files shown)


tsl/src/continuous_aggs/create.c
```diff
@@ -186,9 +186,15 @@ typedef struct CAggTimebucketInfo
...
 {
 	struct MatTableColumnInfo *mattblinfo;
-	bool addcol;
+	bool added_aggref_col;
+	bool inside_aggref;			/* Set to true when traversing within an Aggref node */
+	bool var_outside_of_aggref; /* Set to true when you come across a Var that is not inside an
+								   Aggref node */
 	Oid ignore_aggoid;
 	int original_query_resno;
+	List *orig_vars; /* List of Var nodes that have been mapped to materialization table columns */
```
nikkhils, line 195: These comments don't seem to make sense to me. Can we clarify them further?

tsl/src/continuous_aggs/create.c
```diff
@@ -1443,9 +1467,64 @@ add_partialize_column(Aggref *agg_to_partialize, AggPartCxt *cxt)
...
+var_already_mapped(Var *var, AggPartCxt *cxt)
+{
+	ListCell *lc_old, *lc_new;
+
+	forboth (lc_old, cxt->orig_vars, lc_new, cxt->mapped_vars)
+	{
+		Var *orig_var = (Var *) lfirst_node(Var, lc_old);
+		Var *mapped_var = (Var *) lfirst_node(Var, lc_new);
+
+		/* the should be no subqueries so varlevelsup should not be a problem here */
```
nikkhils, line 1489: typo, `the` -> `there`

tsl/src/continuous_aggs/create.c
```diff
@@ -1460,10 +1539,21 @@ add_aggregate_partialize_mutator(Node *node, AggPartCxt *cxt)
 		if (cxt->ignore_aggoid == ((Aggref *) node)->aggfnoid)
 			return node; /*don't process this further */
 
+		bool old_inside_aggref = cxt->inside_aggref;
+		if (!old_inside_aggref)
+			cxt->inside_aggref = true;
+
 		Aggref *newagg = add_partialize_column((Aggref *) node, cxt);
```
nikkhils, line 1546: Not a fan of declaring variables in the middle lines like this. Even though this is existing code, we could clean this up. Also, shouldn't the compiler have barfed about this?

mfundul: Our codebase is full with this kind of declarations. This is because we follow C99 standards. See our [style guide](https://github.com/timescale/timescaledb/blob/master/docs/StyleGuide.md#declarations-before-code).

tsl/src/continuous_aggs/create.c
mfundul: Added clarifications.

nikkhils (approved):
LGTM!

tsl/src/continuous_aggs/create.c
```diff
@@ -1334,6 +1346,24 @@ mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input, int original_q
 			}
 		}
 		break;
+		case T_Var:
+		{
+			PRINT_MATCOLNAME(colbuf, "var", original_query_resno, matcolno);
```
gayyappan, line 1351: We should maintain 2 invariants:
1. every aggref gets its own column in the materialized hypertable
2. every grouby element gets its own column in the materialized hypertable.

Prior to this PR, the code adds a column for any target entry exprs like var1+var2 as well. We do not need this additioanl column in the materialized hypertable (Note due SQL grammar rules var1 and var2 are group by columns or var1+var2 is a group by column and already exists in the materialized hypertable). But we do not want to modify exprs of this form ( that involve only vars) as it would break compatibility with the existing materialized hypertable defn. (We should do this eventually and provide an upgrade path for materialized hypertable schema changes).

The current code has the following problem with exprs that involve a combination of aggrefs + grouping columns: 
SELECT  expr( aggref, group by column) -----> problematic expr

The issue is : the target entry is not fixed up for vars, the group-by column reference (i.e. does not refer to the correct column in the materialized hypertable) . The expr is fixed up only for the aggrefs. The correct solution is to fix up the var references to point to the correct attnos from the materialized hypertable instead of ADDING a new column for it.
This will ensure that we are compatibile with existing materialized hypertables and will allow us to fix up the finalize view definition for caggs with this issue(instead of having to drop and add the cagg).
e.g.
create materialized view c4 with (timescaledb.continuous) as select time_bucket( '1 day' ::interval, "Time") bkt, sum(b) + (i +b) + b from test1 group by  bkt, b, (i+b) ;

mfundul: The motivation behind @gayyappan 's comment is the potential to do data recovery during an update (or post-update with some tools) on the materialized table data, since the existing issue leads to crashes or data loss/corruption.

The current state of this PR does not repair any preexisting data, it only makes sure that the new CAGG creations do not have the issue.

afiskon (approved):

sql/updates/post-update.sql
```diff
@@ -18,10 +23,19 @@ BEGIN
...
             EXECUTE format('ALTER VIEW %s SET (timescaledb.materialized_only=%L) ', vname::text, materialized_only);
-        ELSE
+        ELSIF ts_version < '2.7.0' THEN
             EXECUTE format('ALTER MATERIALIZED VIEW %s SET (timescaledb.materialized_only=%L) ', vname::text, materialized_only);
+        ELSE
+            SET log_error_verbosity TO VERBOSE;
+            RAISE WARNING 'find the name of CAGG before the call %', vname::text;
+            CALL @extschema@.post_update_cagg_try_repair(vname);
+            RAISE WARNING 'find the name of CAGG after the call %', vname::text;
         END IF;
```
fabriziomello, line 33: Not sure if this logic is correct because it means that if I'm upgrading to 2.7.0 we'll not execute the `ALTER MATERIALIZED VIEW` anymore... and as far I know it's necessary for users upgrading from versions before 2.0.0

mfundul: `cagg_rebuild_view_definition()` will do the same thing as `ALTER MATERIALIZED VIEW` for any older CAGG views. It will skip any view without partials and it will print warnings on buggy views.

tsl/src/continuous_aggs/create.c
```diff
@@ -1663,9 +1748,19 @@ add_aggregate_partialize_mutator(Node *node, AggPartCxt *cxt)
 		if (cxt->ignore_aggoid == ((Aggref *) node)->aggfnoid)
 			return node; /*don't process this further */
 
+		bool old_inside_aggref = cxt->inside_aggref;
+		if (!old_inside_aggref)
+			cxt->inside_aggref = true;
```
konskov, line 1753: It looks like `cxt->inside_aggref` is not used in `add_partialize_column` so it's probably unnecessary to set it and then unset it.

mfundul: I probably created them initially during recursion, but at this point it seems like the code doesn't recurse into anything anymore, so you are right, they should be removed.

fabriziomello (approved):

konskov (approved):

src/utils.c
```diff
@@ -1092,3 +1092,24 @@ ts_get_node_name(Node *node)
 			return psprintf("Node (%d)", nodeTag(node));
 	}
 }
+
+/*
+ * Implementation marked unused in PostgresQL lsyscache.c
```
svenklemm, line 1097:
```suggestion
 * Implementation marked unused in PostgreSQL lsyscache.c
```

svenklemm (approved):

PR #3903, opened 2021-12-09
Support UPPERREL_PARTIAL_DISTINCT stage in fdw_create_upper_paths()

afiskon:
UpperRelationKind enum was extended in the upstream and now can be
UPPERREL_PARTIAL_DISTINCT as well. This patch rewrites the switch-case
expression in the fdw_create_upper_paths() procedure accordingly.

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=22c4e88e

### DIFF
--- a/tsl/src/fdw/scan_plan.c
+++ b/tsl/src/fdw/scan_plan.c
@@ -967,6 +967,9 @@ fdw_create_upper_paths(TsFdwRelInfo *input_fpinfo, PlannerInfo *root, UpperRelat
 		case UPPERREL_SETOP:
 		case UPPERREL_WINDOW:
 		case UPPERREL_FINAL:
+#if PG15_GE
+		case UPPERREL_PARTIAL_DISTINCT:
+#endif
 			break;
 	}
 }


tsl/src/fdw/scan_plan.c
```diff
@@ -962,11 +962,7 @@ fdw_create_upper_paths(TsFdwRelInfo *input_fpinfo, PlannerInfo *root, UpperRelat
 									   create_path);
 			break;
 			/* Currently not handled (or received) */
-		case UPPERREL_DISTINCT:
-		case UPPERREL_ORDERED:
-		case UPPERREL_SETOP:
-		case UPPERREL_WINDOW:
-		case UPPERREL_FINAL:
+		default:
```
erimatnor, line 965: I am not 100% convinced about this change. A wildcard capture like this isn't always great because it doesn't make clear when there is something changing, like now happened, and in some cases it is actually better to handle the new case than letting it go through the default path without consideration.

Could we just add the new case value to the switch clause instead? 

In either case, we should probably update the comment above as well, because now we are clearly receiving some of these cases that we don't handle.

afiskon: > Could we just add the new case value to the switch clause instead?
With a couple of #if's that would check the PostgreSQL version, sure. To my knowledge, we don't normally do it right in the code, this is why I choose the `default` clause. However, personally, I'm fine with either solution.

afiskon: @erimatnor  So, to clarify, do you insist on this change? I see the pros and cons of either approach and don't have a strong opinion here.

akuzm (approved):
Let's make an explicit list of cases, it feels more reassuring because we won't have to think "ohh which other uperrels are there again".

tsl/src/fdw/scan_plan.c
fabriziomello: I'm with @erimatnor in favor of explicitly deal with all enum values adding the properly #ifdefs.

Today we don't need to take care of this new enum value but in the future a new enum value can be added and maybe we should take care of it in our side so IMHO is better to raise a compiler warning/error so we can pay attention with this specific change.

afiskon: OK since @akuzm seconds this, here is an updated patch.

afiskon: @fabriziomello OK, OK :)

fabriziomello (approved):

PR #3890, opened 2021-12-06
Estimate chunk sizes faster

akuzm:
just testing something

PR #3912, opened 2021-12-10
Stop early when checking if the chunk is historical

akuzm:
We don't have to enumerate all the subsequent chunks to find out that
the given chunk is not historical. Stop doing this useless work to
improve performance.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -2640,10 +2640,12 @@ ts_chunk_get_by_id(int32 id, bool fail_if_not_found)
 
 /*
  * Number of chunks created after given chunk.
- * If chunk2.id > chunk1.id then chunk2 is created after chunk1
+ * If chunk2.id > chunk1.id then chunk2 is created after chunk1.
+ * Can accept a limit for cases where we only need to check that there is more
+ * than a certain number of chunks after the given one.
  */
 int
-ts_chunk_num_of_chunks_created_after(const Chunk *chunk)
+ts_chunk_num_of_chunks_created_after(const Chunk *chunk, int limit)
 {
 	ScanKeyData scankey[1];
 
@@ -2662,7 +2664,7 @@ ts_chunk_num_of_chunks_created_after(const Chunk *chunk)
 							   NULL,
 							   NULL,
 							   NULL,
-							   0,
+							   limit,
 							   ForwardScanDirection,
 							   AccessShareLock,
 							   CurrentMemoryContext);
--- a/tsl/src/fdw/relinfo.c
+++ b/tsl/src/fdw/relinfo.c
@@ -184,8 +184,6 @@ estimate_chunk_fillfactor(Chunk *chunk, Hyperspace *space)
 	const Dimension *time_dim = hyperspace_get_open_dimension(space, 0);
 	const DimensionSlice *time_slice = get_chunk_time_slice(chunk, space);
 	Oid time_dim_type = ts_dimension_get_partition_type(time_dim);
-	int num_created_after = ts_chunk_num_of_chunks_created_after(chunk);
-	int total_slices = get_total_number_of_slices(space);
 
 	if (IS_TIMESTAMP_TYPE(time_dim_type))
 	{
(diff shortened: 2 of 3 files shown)


akuzm:
> test bgw_db_scheduler             ... FAILED     5078 ms

Flaky as usual.

src/chunk.c
```diff
@@ -2640,10 +2640,12 @@ ts_chunk_get_by_id(int32 id, bool fail_if_not_found)
 
 /*
  * Number of chunks created after given chunk.
- * If chunk2.id > chunk1.id then chunk2 is created after chunk1
+ * If chunk2.id > chunk1.id then chunk2 is created after chunk1.
```
svenklemm, line 2643: I did not check all the surrounding code here but this logic seems flawed. Do we really care about chunk ids here or do we actually care about relative position on the time dimension. I think we are looking for the latter and in that case chunk id being greater does not mean that a chunk time is after a chunk with a lower id as chunks might be created by backfilling.

akuzm: I'm not sure what we have to consider 'historical/current' to estimate the fill factor -- current in 'timeseries time' or current in 'real time' (chunk id). In the usual case where we are inserting the current time series data, these two are the same. If we are backfilling, maybe 'current in real time' makes more sense because these are the chunks we created last, and they are likely to have smaller fill factor. @erimatnor maybe you remember the logic behind this choice? Anyway, I think this PR doesn't have to change this logic. Whatever chunk order we use, we don't have to go through all of them.

erimatnor: AFAIK, the logic is not flawed. The way this is used is to find a "stable" chunk, i.e., its size is more or less stable and won't see a lot of writes. So, we care about how long ago the chunk was created in logical time, assuming that the more chunks have been created after this one, the more stable it is.

Usually, logical time correlates well with the real time dimension as well, but we also have to account for things like multi-dimensional partitioning, so we could have many chunks with the same or similar time ranges. I think the number of space partitions should be used to set the limit of how far we look beyond the current chunk.

So, if we have 4 space partitions, then one can assume that we should at least look 5 chunks back in logical time (including the current chunk) to find a reasonably stable chunk. I don't think this has to be exact, it is just used for estimation.

tsl/src/fdw/relinfo.c
```diff
@@ -203,6 +201,9 @@ estimate_chunk_fillfactor(Chunk *chunk, Hyperspace *space)
 		/* if we are beyond end range then chunk can possibly be totally filled */
 		if (time_slice->fd.range_end <= now_internal_time)
 		{
+			int total_slices = get_total_number_of_slices(space);
+			int num_created_after = ts_chunk_num_of_chunks_created_after(chunk, total_slices);
```
erimatnor, line 205: Maybe a simpler way of doing this is just to get the max(chunk_id) and compare it to the chunk_id of the current chunk, assuming that it is always increasing.

num_created_after = max_chunk_id - current_chunk_id;

src/chunk.c
erimatnor: Note that the number of space partitions could have changed, so there might have been a different number of space partitions when the chunk was created. Not sure how to easily account for this though.

akuzm: I realized another problem with chunk ids -- they are global, for all hypertables. This function looks them up w/o any filter by hypertable id, so all these comparisons with numbers of dimensions etc don't really make sense unless there is only one hypertable in the database. We don't have an index on (hypertable_id, chunk_id), so there is no fast way to look up the chunks for a particular hypertable.

Considering that this logic is a fallback for missing ANALYZE data, all this looks unnecessarily complicated. Couldn't we just treat all the chunks w/o statistics as recent ones, and estimate a fill factor of 0.5 for them, or something?

tsl/src/fdw/relinfo.c
akuzm: There's no way to get the max chunk id for a given hypertable, unfortunately.

PR #3913, opened 2021-12-10
Use function oids instead of names when transforming the sort keys

akuzm:
This is more explicit, and looking up the function name each time has a
performance impact.


akuzm:
The 'query' test seems to be flaky, let's restart.

svenklemm (approved):

nikkhils (approved):
LGTM

PR #3918, opened 2021-12-10
Fix DataNodeScan plans with One-Time Filter

### LINKED ISSUE #3658: distributed hypertable: unexpected child node of Append or MergeAppend: 60
```
 timescaledb │ 2.5.0-dev │ public     │ Enables scalable inserts and complex queries for time-series data
 PostgreSQL 13.4 on x86_64-pc-linux-gnu, compiled by Ubuntu clang version 13.0.0-++20210924062554+d7b669b3a303-1~exp1~20210924163351.114, 64-bit


create table ttt(time timestamptz, series int, value float);
select create_distributed_hypertable('ttt', 'time', partitioning_column => 'series');
insert into ttt select now() + (interval '1 minute') * (random() * 10 + x), x, random() * x from generate_series(1, 1000) x;
explain select * from ttt where now() < now();
ERROR:  unexpected child node of Append or MergeAppend: 60
```

AsyncAppend can't handle a Result node that is generated on the remote side for a constant true/false qual.

Somewhat related to this one: https://github.com/timescale/timescaledb/issues/3598

svenklemm:
When a query has a filter that only needs to be evaluated once per
query it will be represented as a Result node with the filter
condition on the Result node and the actual query as child of the
result node. find_data_node_scan_state_child did not consider
Result node as valid node to contain a DataNodeScan node leading
to a `unexpected child node of Append or MergeAppend: 62` for
queries that had one-time filter with a subquery.

Fixes #3658

### DIFF
--- a/tsl/test/shared/expected/dist_queries.out
+++ b/tsl/test/shared/expected/dist_queries.out
@@ -1,0 +1,20 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+-- Test DataNodeScan with subquery with one-time filter
+SELECT
+  id
+FROM
+  insert_test
+WHERE
+  NULL::int2 >= NULL::int2 OR
+  EXISTS (SELECT 1 from dist_chunk_copy WHERE insert_test.id IS NOT NULL)
+ORDER BY id;
+ id 
+----
+  1
+  2
+  3
+  4
+(4 rows)
+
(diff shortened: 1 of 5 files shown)


nikkhils (approved):
LGTM

tsl/test/shared/expected/dist_queries.out
```diff
@@ -0,0 +1,19 @@
+-- This file and its contents are licensed under the Timescale License.
```
nikkhils, line 1: Why not add this to existing `dist_query.sql`?

akuzm (approved):

tsl/test/shared/expected/dist_queries.out
svenklemm: Having tests as shared tests is preferable to the non-shared variant as a shared test does not incur the database creation overhead a normal test would.

PR #3923, opened 2021-12-13
Allow changing a data node via alter server

### LINKED ISSUE #3915: [Enhancement]: Allow altering a data node on the access node

User experience, API improvement

Node management

It might be necessary to update the configuration of a data node if, e.g., the port, host, or database name changes. This can be done by either enabling `ALTER SERVER`, adding an `alter_data_node` command or both. 

### Implementation challenges

Adding an `alter_data_node` function might require further consideration to get the API right.

erimatnor:
Allow changing the name and configuration of a data node via `ALTER
SERVER` since there is no `alter_data_node` command.

The functions `add_data_node` and `delete_data_node` are wrappers
around `CREATE SERVER` and `DROP SERVER`, respectively. They block the
PostgreSQL commmands since they do additional things, like data node
bootstrapping.

However, there's currently no way to change a data node's
configuration, and no additional functionality is required over
standard `ALTER SERVER`, so we might as well allow it until a
`alter_data_node` is implemented.

Fixes #3915

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -474,10 +474,13 @@ process_alter_foreign_server(ProcessUtilityArgs *args)
 {
 	AlterForeignServerStmt *stmt = (AlterForeignServerStmt *) args->parsetree;
 
-	if (block_on_foreign_server(stmt->servername))
+	if (stmt->has_version)
 		ereport(ERROR,
 				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-				 errmsg("alter server not supported on a TimescaleDB data node")));
+				 errmsg("operation not supported"),
+				 errdetail("It is not possible to set a version on the data node configuration.")));
+
+	/* Other options are validated by the FDW */
 
 	return DDL_CONTINUE;
 }
@@ -2091,13 +2094,6 @@ process_rename(ProcessUtilityArgs *args)
 		 * stmt->relation never be NULL unless we are renaming a schema or
 		 * other objects, like foreign server
 		 */
-		if ((stmt->renameType == OBJECT_FOREIGN_SERVER) &&
-			block_on_foreign_server(strVal(stmt->object)))
-		{
-			ereport(ERROR,
-					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-					 errmsg("rename not supported on a TimescaleDB data node")));
-		}
 		if (stmt->renameType != OBJECT_SCHEMA)
 			return DDL_CONTINUE;
 	}
--- a/tsl/test/sql/data_node.sql
+++ b/tsl/test/sql/data_node.sql
@@ -55,11 +55,20 @@ SELECT * FROM add_data_node('data_node_2', host => 'localhost', database => :'DN
 
 SELECT * FROM add_data_node('data_node_3', host => 'localhost', database => :'DN_DBNAME_3');
 
--- Test altering server command is blocked
+
+-- Altering the host, dbname, and port should work via ALTER SERVER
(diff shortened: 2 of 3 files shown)


nikkhils (approved):
I guess we have to change documentation along with this to mention the use of standard ALTER SERVER for changing parameters? 
Looks good otherwise.

erimatnor:
@nikkhils Yes, we need to update docs.

In fact, I realized there were some inconsistencies when working on the docs, so I made this PR. For example, we  allow changing owner via `ALTER SERVER` but nothing else, which is a bit strange.

src/process_utility.c
```diff
@@ -472,10 +472,13 @@ process_alter_foreign_server(ProcessUtilityArgs *args)
 {
 	AlterForeignServerStmt *stmt = (AlterForeignServerStmt *) args->parsetree;
 
-	if (block_on_foreign_server(stmt->servername))
+	if (stmt->has_version)
 		ereport(ERROR,
 				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-				 errmsg("alter server not supported on a TimescaleDB data node")));
+				 errmsg("operation not supported"),
+				 errdetail("It is not possible to set a version on a data node.")));
```
fabriziomello, line 479: Reading this message seems like I'm executing the `ALTER SERVER` on the data node and it's not true because actually we're executing it in the access node to change a data node server configuration, what about rephrasing it a bit?

```suggestion
				 errdetail("It is not possible to set a version to a data node server configuration.")));
```

As I'm not a native speaker not sure if my suggestion is good enough. Otherwise looks good to me.

fabriziomello (approved):

src/process_utility.c
```diff
@@ -472,10 +472,13 @@ process_alter_foreign_server(ProcessUtilityArgs *args)
 {
 	AlterForeignServerStmt *stmt = (AlterForeignServerStmt *) args->parsetree;
 
-	if (block_on_foreign_server(stmt->servername))
+	if (stmt->has_version)
```
pmwkaa, line 475: I wonder if we should move foreign server related code and blocking logic to the dist ddl instead

tsl/test/sql/data_node.sql
```diff
@@ -55,11 +55,20 @@ SELECT * FROM add_data_node('data_node_2', host => 'localhost', database => :'DN
 
 SELECT * FROM add_data_node('data_node_3', host => 'localhost', database => :'DN_DBNAME_3');
 
--- Test altering server command is blocked
+
+-- Altering the host, dbname, and port should work via ALTER SERVER
+BEGIN;
+ALTER SERVER data_node_3 OPTIONS (SET host 'data_node_3', SET dbname 'new_db_name', SET port '9999');
```
pmwkaa, line 61: Since we are changing host name here, I believe it would make sense to test how we (and if we) clean up our connection cache here. What if we openned connection to the previous host and then change the host name?

pmwkaa (approved):
Just a couple of notes, otherwise looks good

tsl/test/sql/data_node.sql
erimatnor: There's already a callback to invalidate a connection when the foreign server changes (see the connection cache code). That should already be tested.

src/process_utility.c
erimatnor: I thought about it but didn't see a need for it at this point.

src/process_utility.c
erimatnor: Updated the message.

PR #3928, opened 2021-12-14
Use moving average to estimate chunk size

akuzm:
When no size statistics is available for a chunk because it haven't
been ANALYZEd, we have to resort to some estimates. A good way to
estimate the size of the chunk is to look at the sizes of the nearby
chunk.  Before this commit, we would look up the ten nearby chunk for
each chunk w/o statistics, which was slow. This commit adds the code
to maintain exponential moving average of the recent chunk sizes in
the parent RelOptInfo as we go through them, avoiding the additional
lookups.

Part of #3890

### DIFF
--- a/tsl/src/fdw/relinfo.c
+++ b/tsl/src/fdw/relinfo.c
@@ -230,74 +230,11 @@ estimate_chunk_fillfactor(Chunk *chunk, Hyperspace *space)
 	}
 }
 
-typedef struct RelEstimates
+static void
+estimate_tuples_and_pages_using_shared_buffers(PlannerInfo *root, Hypertable *ht, RelOptInfo *rel)
 {
-	double tuples;
-	BlockNumber pages;
-} RelEstimates;
-
-/*
- * The idea is to look into number of tuples and pages for N previous chunks
- * and calculate an average. Ideally we could add weights to this calculation
- * and give more importance to newer chunks but a ballpark estimate should be
- * just fine.
- */
-static RelEstimates *
-estimate_tuples_and_pages_using_prev_chunks(PlannerInfo *root, Hyperspace *space,
-											Chunk *current_chunk)
-{
-	RelEstimates *estimates = palloc0(sizeof(RelEstimates));
-	ListCell *lc;
-	float4 total_tuples = 0;
-	int32 total_pages = 0;
-	int non_zero_reltuples_cnt = 0;
-	int non_zero_relpages_cnt = 0;
-	const DimensionSlice *time_slice = get_chunk_time_slice(current_chunk, space);
-	List *prev_chunks = ts_chunk_get_window(time_slice->fd.dimension_id,
-											time_slice->fd.range_start,
-											DEFAULT_CHUNK_LOOKBACK_WINDOW,
-											CurrentMemoryContext);
-
-	foreach (lc, prev_chunks)
-	{
-		Chunk *pc = lfirst(lc);
-		HeapTuple rel_tuple = SearchSysCache1(RELOID, ObjectIdGetDatum(pc->table_id));
-		Form_pg_class rel_form;
-
-		if (!HeapTupleIsValid(rel_tuple))
-			ereport(ERROR,
-					(errcode(ERRCODE_TS_INTERNAL_ERROR),
-					 errmsg("cache lookup failed for chunk relation %u", pc->fd.id),
-					 errdetail("Failed to estimate number of tuples and pages for chunk %d.",
-							   pc->table_id)));
-
-		rel_form = (Form_pg_class) GETSTRUCT(rel_tuple);
-		if (rel_form->reltuples > 0)
-		{
-			total_tuples += rel_form->reltuples;
-			non_zero_reltuples_cnt++;
-		}
-		if (rel_form->relpages > 0)
-		{
-			total_pages += rel_form->relpages;
-			non_zero_relpages_cnt++;
-		}
-		ReleaseSysCache(rel_tuple);
-	}
-
-	if (non_zero_reltuples_cnt > 0)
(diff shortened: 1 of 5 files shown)


nikkhils (commented):
We have another work item to fetch analyze stats in the background for distributed hypertables. Once that's in place we will always have relevant stats for them. Do we need these changes in the light of that?

akuzm:
> We have another work item to fetch analyze stats in the background for distributed hypertables. Once that's in place we will always have relevant stats for them. Do we need these changes in the light of that?

This is a fallback for the case when we still don't have the stats, e.g. when the remote analyze didn't run. So probably it makes sense to keep it. Not sure that the "moving average" stuff is important, maybe just a constant estimate "by shared buffers" would be enough as a fallback.

tsl/src/fdw/relinfo.c
```diff
@@ -308,74 +245,97 @@ estimate_tuples_and_pages_using_shared_buffers(PlannerInfo *root, Hypertable *ht
...
+	TsFdwRelInfo *parent_private = fdw_relinfo_get(parent_info);
+	RangeTblEntry *parent_rte = planner_rt_fetch(parent_relid, root);
+	Cache *hcache = ts_hypertable_cache_pin();
+	Hypertable *ht = ts_hypertable_cache_get_entry(hcache, parent_rte->relid, CACHE_FLAG_NONE);
+	Hyperspace *hyperspace = ht->space;
+	RangeTblEntry *chunk_rte = planner_rt_fetch(chunk_rel->relid, root);
+	Chunk *chunk = ts_chunk_get_by_relid(chunk_rte->relid, true /* fail_if_not_found */);
+
+	const double fillfactor = estimate_chunk_fillfactor(chunk, hyperspace);
+	const bool have_chunk_statistics = chunk_rel->pages != 0 || chunk_rel->tuples > 0;
```
erimatnor, line 285: Should we check that we have both pages and tuples set? Or assert it? (Can you have tuples stored without a page?)

```suggestion
	const bool have_chunk_statistics = chunk_rel->pages != 0 && chunk_rel->tuples > 0;
```

tsl/src/fdw/relinfo.c
```diff
@@ -308,74 +245,97 @@ estimate_tuples_and_pages_using_shared_buffers(PlannerInfo *root, Hypertable *ht
...
+	{
+		/*
+		 * We have the moving average of chunk sizes and a good estimate
+		 * of this chunk size from ANALYZE. Update the moving average.
+		 */
+		const double f = 0.1;
+		parent_private->average_chunk_pages =
+			(1 - f) * parent_private->average_chunk_pages + f * chunk_rel->pages / fillfactor;
+		parent_private->average_chunk_tuples =
+			(1 - f) * parent_private->average_chunk_tuples + f * chunk_rel->tuples / fillfactor;
```
erimatnor, line 330: It seems as if we update the moving average on every query that involves the chunk instead of when we have new stats. Won't it make the moving average a bit weird if we are, e.g., querying the same chunk many times?

In other words, a chunk's effect on the moving average depends on how many times it is queried rather than just once.

Ideally, one would expect to add a chunk to the moving average only once, or at least only when it is re-analyzed.

Does it make more sense to update this moving average when we run ANALYZE instead of at query time?

We also have to account for the fact that a chunk is sometimes "incomplete", but i guess that is what the fillfactor is supposed to account for.

erimatnor (commented):
I think the basic idea is good, but wondering if we shouldn't update the moving average on ANALYZE instead of on every query? Also, wondering if we really need a moving average in that case, instead of a standard average (assuming that ANALYZE will update the stats for all chunks).

Doing it on every query seems to have the effect that we  repeatedly add the same chunk to the moving average, which seems weird.

akuzm:
> I think the basic idea is good, but wondering if we shouldn't update the moving average on ANALYZE instead of on every query? Also, wondering if we really need a moving average in that case, instead of a standard average (assuming that ANALYZE will update the stats for all chunks).

If we ran ANALYZE, we are going to have statistics for each chunks and don't need this estimate all. This code is a fallback for the case where the ANALYZE didn't run.

> Doing it on every query seems to have the effect that we repeatedly add the same chunk to the moving average, which seems weird.

This moving average is per query, per hypertable. It only looks at the chunks participating in the query, and is discarded after the query is planned. So there is no accumulation with repeated queries.

afiskon (approved):
👍

nikkhils (approved):
LGTM! DO we need more tests?

akuzm:
Flaky test: `dist_api_calls               ... FAILED     4651 ms`
<details>
```
diff -u /home/runner/work/timescaledb/timescaledb/tsl/test/expected/dist_api_calls.out /home/runner/work/timescaledb/timescaledb/build/tsl/test/results/dist_api_calls.out
--- /home/runner/work/timescaledb/timescaledb/tsl/test/expected/dist_api_calls.out	2021-12-16 17:17:43.268621419 +0000
+++ /home/runner/work/timescaledb/timescaledb/build/tsl/test/results/dist_api_calls.out	2021-12-16 17:20:53.600124238 +0000
@@ -380,13 +380,4 @@
 (1 row)
 
 ANALYZE disttable_repl;
-SELECT approximate_row_count('disttable_repl');
- approximate_row_count 
------------------------
-                     8
-(1 row)
-
-DROP TABLE disttable_repl;
-DROP DATABASE :DN_DBNAME_1;
-DROP DATABASE :DN_DBNAME_2;
-DROP DATABASE :DN_DBNAME_3;
+ERROR:  [data_node_1]: relation "disttable_repl" does not exist
```
</details>

akuzm:
> LGTM! DO we need more tests?

I think the two tests we have for chunks w/o statistics are good enough. They cover the use of this moving average and the estimation of the fill factor.

PR #3929, opened 2021-12-14
Add option to drop database when deleting data node

### LINKED ISSUE #3876: [Enhancement]: Add option to drop data node DBs when deleting a data node

User experience

Multi-node, Data node, Node management

When deleting a data node, the database on the node is left intact. This is to ensure that `delete_data_node` is a local operation on the access node that can succeed even if the data node is not responding.

However, if the data node is available, it makes sense to have an option to cleanup the data node, e.g., delete the database on it. Then it is easier to add the data node again.

### Implementation challenges

The deletion of the data node's database should be optional and probably not the default.

erimatnor:
When deleting a data node, it is often convenient to be able to also
drop the database on the data node so that the node can be added again
using the same database name. However, dropping the database is
optional since it should be possible to delete a data node even if it
is no longer responding.

With the new functionality, a data node's database can be dropped as
follows:

```sql
SELECT delete_data_node('dn1', drop_database=>true);
```

Note that the default behavior is still to not drop the database in
order to be compatible with the old behavior. Enabling the option also
makes the function non-transactional, since dropping a database is not
transactional. Therefore, it is not possible to use this option in a
transaction block.

Closes #3876

### DIFF
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -1320,6 +1320,132 @@ data_node_detach(PG_FUNCTION_ARGS)
 	PG_RETURN_INT32(removed);
 }
 
+/*
+ * Drop a data node's database.
+ *
+ * To drop the database on the data node, a connection must be made to another
+ * database since one cannot drop the database currently connected
+ * to. Therefore, we bypass the connection cache and use a "raw" connection to
+ * a standard database (e.g., template0 or postgres), similar to how
+ * bootstrapping does it.
+ *
+ * Note that no password is provided on the command line as is done when
+ * bootstrapping. Instead, it is assumed that the current user already has a
+ * method to authenticate with the remote data node (e.g., via a password
+ * file, certificate, or user mapping). This should normally be the case or
+ * otherwise the user wouldn't have been able to use the data node.
+ *
+ * Note that the user that deletes a data node also must be the database owner
+ * on the data node. The database will only be dropped if there are no other
+ * concurrent connections so all connections must be closed before being able
+ * to drop the database.
+ */
+static void
+drop_data_node_database(const ForeignServer *server)
+{
+	ListCell *lc;
+	TSConnection *conn;
+	Oid userid = GetUserId();
+	TSConnectionId connid = {
+		.server_id = server->serverid,
+		.user_id = userid,
+	};
+	/* Make a copy of the node name since the server pointer will be
+	 * updated */
+	char *nodename = pstrdup(server->servername);
+	char *dbname = NULL;
+	char *err = NULL;
+	int i;
+
+	/* Figure out the name of the database that should be dropped */
+	foreach (lc, server->options)
+	{
+		DefElem *d = lfirst(lc);
+
+		if (strcmp(d->defname, "dbname") == 0)
+		{
+			dbname = defGetString(d);
+			break;
+		}
+	}
+
+	if (NULL == dbname)
+	{
+		/* This should not happen unless the configuration is messed up */
+		ereport(ERROR,
+				(errcode(ERRCODE_TS_DATA_NODE_INVALID_CONFIG),
+				 errmsg("could not drop the database on data node \"%s\"", nodename),
+				 errdetail("The data node configuration lacks the \"dbname\" option.")));
+		pg_unreachable();
+		return;
+	}
+
+	/* Clear potentially cached connection to the data node for the current
+	 * session so that it won't block dropping the database */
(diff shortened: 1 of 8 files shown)


erimatnor:
Realized this might be useful when writing documentation for multi-node. It has also been a requested feature and felt like low-hanging fruit. Let me know if this is something we should do now or hold off on.

RafiaSabih:
This looks good, wanted to clarify if such an option will also be useful in detach_data_node....?

tsl/src/data_node.c
```diff
@@ -1320,13 +1320,140 @@ data_node_detach(PG_FUNCTION_ARGS)
...
+		AlterForeignServer(&stmt);
+
+		/* Make changes to foreign server database visible */
+		CommandCounterIncrement();
+
+		/* Get the updated server definition */
+		server = data_node_get_foreign_server(nodename, ACL_USAGE, true, false);
+		/* Open a connection to the bootstrap database using the new server options */
+		conn_options = remote_connection_prepare_auth_options(server, userid);
+		conn = remote_connection_open_with_options_nothrow(nodename, conn_options, &err);
```
pmwkaa, line 1425: nit: might be slighter easier on eyes to do this pattern instead:

```
conn = NULL
for (i = 0; i < lengthof(bootstrap_databases); i++)
{
   ...
   If (conn)
      break;
}

if (conn == NULL)
   ereport(ERROR, ...)

remote_connection_cmdf_ok(conn, "DROP DATABASE %s", quote_identifier(dbname));
remote_connection_close(conn);
```

tsl/src/data_node.c
```diff
@@ -1320,13 +1320,140 @@ data_node_detach(PG_FUNCTION_ARGS)
...
+				(errcode(ERRCODE_TS_DATA_NODE_INVALID_CONFIG),
+				 errmsg("could not drop the database on data node \"%s\"", nodename),
+				 errdetail("The data node configuration lacks the \"dbname\" option.")));
+		pg_unreachable();
+		return;
+	}
+
+	/* Clear potentially cached connection to the data node for the current
+	 * session so that it won't block dropping the database */
+	remote_connection_cache_remove(connid);
```
pmwkaa, line 1385: Cached connections being dropped as well after the execution of this function in the `delete_data_node()`, maybe we should invalidate them first before executing `drop_data_node_database()` then?

tsl/test/sql/data_node_bootstrap.sql
```diff
@@ -79,8 +79,46 @@ DELETE FROM _timescaledb_catalog.metadata WHERE key = 'dist_uuid';
 SELECT * FROM add_data_node('bootstrap_test', host => 'localhost',
                             database => 'bootstrap_test', bootstrap => false);
 
-SELECT * FROM delete_data_node('bootstrap_test');
+\set ON_ERROR_STOP 0
+-- Dropping the database with delete_data_node should not work in a
+-- transaction block since it is non-transactional.
+BEGIN;
+SELECT * FROM delete_data_node('bootstrap_test', drop_database => true);
```
pmwkaa, line 86: Might be good to have a case, when someone is still using the database and that prevents the drop database execution, yet not sure how to hold the connection here, maybe there is a similar way to do achieve this

pmwkaa (approved):
Looks good, just a couple of notes

RafiaSabih (approved):

tsl/src/data_node.c
erimatnor: Invalidating a connection doesn't close it or remove it from the cache. It only marks it as invalid so that it is recreated on the next cache lookup.

We need to purge/close any of our own cached connections to the data node's database, or else it will block DROP DATABASE.  The purpose of this call is the close the connection and remove it from the cache. There's only one connection in the cache for each data node, so there's no point in removing other connections.

tsl/test/sql/data_node_bootstrap.sql
erimatnor: Yeah, I thought about that, but it probably requires an isolation test to spawn multiple concurrent processes.

I did verify that it will fail if the current session has an open connection, and that's why I added the code to purge the cache of open connections. Still, I think we are at that point just testing that DROP DATABASE fails the way it should, which I think we should be able to trust that it does.

erimatnor:
> 

So, `detach_data_node` only removes the data node from one hypertable, but keeps it on other hypertables and in the pool of available data nodes. So, "detach" has different semantics compared to "delete" and shouldn't delete the database since it is still in use.

PR #3933, opened 2021-12-15
Refactoring: get rid of max_bucket_width

afiskon:
Our code occasionally mentions max_bucket_width. However, in practice, there is
no such thing. For fixed-sized buckets, bucket_width and max_bucket_width are
always the same, while for variable-sized buckets bucket_width is not used at
all (except the fact that it equals -1 to indicate that the bucket size is
variable).

This patch removes any use of max_bucket_width, except for arguments of:

- _timescaledb_internal.invalidation_process_hypertable_log()
- _timescaledb_internal.invalidation_process_cagg_log()

The signatures of these functions were not changed for backward compatibility
between access and data nodes, which can run different versions of TimescaleDB.

### DIFF
--- a/sql/cagg_utils.sql
+++ b/sql/cagg_utils.sql
@@ -51,8 +51,8 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.drop_dist_ht_invalidation_trigg
 --                      Access Node that belong to 'raw_hypertable_id'
 -- bucket_widths - The array of time bucket widths for all the CAGGs that belong to
 --                 'raw_hypertable_id'
--- max_bucket_widths - The array of the maximum time bucket widths for all the CAGGs that belong
---                     to 'raw_hypertable_id'
+-- max_bucket_widths - (Deprecated) This argument is ignored and is present only
+--                     for backward compatibility.
 -- bucket_functions - (Optional) The array of serialized information about bucket functions
 CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_hypertable_log(
     mat_hypertable_id INTEGER,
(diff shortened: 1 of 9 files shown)


src/continuous_agg.c
```diff
@@ -528,19 +516,15 @@ ts_populate_caggs_info_from_arrays(ArrayType *mat_hypertable_ids, ArrayType *buc
...
-			bucket_function_deserialize(TextDatumGetCString(array_datum4));
+			bucket_function_deserialize(TextDatumGetCString(array_datum3));
 		/* bucket_function is cast to non-const type to make Visual Studio happy */
 		all_caggs->bucket_functions =
 			lappend(all_caggs->bucket_functions, (ContinuousAggsBucketFunction *) bucket_function);
 	}
 	array_free_iterator(it_htids);
 	array_free_iterator(it_widths);
-	array_free_iterator(it_maxes);
+	array_free_iterator(it_bfs);
```
mfundul, line 527: nice!

mfundul (approved):

src/continuous_agg.c
afiskon: Yep :D

sql/cagg_utils.sql
```diff
@@ -51,8 +51,8 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.drop_dist_ht_invalidation_trigg
 --                      Access Node that belong to 'raw_hypertable_id'
 -- bucket_widths - The array of time bucket widths for all the CAGGs that belong to
 --                 'raw_hypertable_id'
--- max_bucket_widths - The array of the maximum time bucket widths for all the CAGGs that belong
---                     to 'raw_hypertable_id'
+-- max_bucket_widths - (Deprecated) This argument is ignored and is present only
```
pmwkaa, line 54: I am curious if we should remove this argument as well, it is a bit unclear how this argument would be used for backward compatibility otherwise

afiskon: It will be removed eventually, the entire signature of the function will be quite different. Please see "Variable-Sized Buckets and Timezones Support for Continuous Aggregates" RFC.

mfundul: We can't remove it, because the AN can be of an older version than the DN. The order of upgrades is, upgrade DN first, then AN last.

PR #3943, opened 2021-12-19
Optimize first/last

### LINKED ISSUE #3935: [Bug]: OOM with first/last aggregate functions
### What type of bug is this?

Crash

Query executor

### What happened?

With reproducing example below, the last query leaks memory while the transaction is running. Likely something is allocated on a per-transaction context rather than on per-tuple context. If query finishes, memory is released, but if query takes long enough it just leads to an OOM. Memory usage increases pretty quickly.
Haven't tested on 2.5.x but I don't see anything in the release notes related to this.

2.4.1

### PostgreSQL version used

12

centos

Source

On prem/Self-hosted

_No response_

```bash
CREATE TABLE test (
    ts BIGINT NOT NULL,
    a TEXT NOT NULL,
    b TEXT NOT NULL,
    c TEXT NOT NULL,
    d BOOLEAN
);

SELECT create_hypertable(
    'test',
    'ts',
    chunk_time_interval => (7 * 24 * 3600)::bigint,
    create_default_indexes => FALSE
);
create index on test (a, b, ts desc);

insert into test
   select ts, a::text, b::text, 'abc', false from generate_series(1, 7*24*3600*7, 4) ts, generate_series(1, 5) a, generate_series(3, 7) b;

-- this increases memory usage massively while its running - leading to OOM if query last long enough
select first(ts, ts), last(ts, ts) from test;
```

svenklemm:
This patch optimizes how first()/last() initialize the compare
function. Previously the compare function would be looked up for
every transition function call but since polymorphic types are
resolved at parse time for a specific aggregate instance the compare
function will not change during its lifetime. Additionally this
patch also fixes a memory leak when using first()/last() on
pointer types.
These changes lead to a roughly 2x speed improvement for first()/
last() and make the memory usage near-constant.

image

Fixes #3935

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,17 @@ accidentally triggering the load of a previous DB version.**
 **Features**
 * #3768 Allow ALTER TABLE ADD COLUMN with DEFAULT on compressed hypertable
 * #3769 Allow ALTER TABLE DROP COLUMN on compressed hypertable
+* #3943 Optimize first/last
 
 **Bugfixes**
 * #3808 Properly handle `max_retries` option
 * #3918 Fix DataNodeScan plans with one-time filter
+* #3938 Fix subtract_integer_from_now on 32-bit platforms and improve error handling
 * #3939 Fix projection handling in time_bucket_gapfill
 
 **Thanks**
 * @erikhh for reporting an issue with time_bucket_gapfill
+* @fvannee for reporting a first/last memory leak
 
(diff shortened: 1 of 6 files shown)


erimatnor (commented):
Suggest adding changelog entry and thanks to bug reporter.

akuzm (approved):

fabriziomello (approved):
Awesome... just add an entry to CHANGELOG

PR #3946, opened 2021-12-21
Don't display an error message when creating an empty CAGG

afiskon:
This is a follow-up to 95804069, "Monthly buckets support in CAGGs". In that
patch we didn't check the creation of a CAGG on top of an empty hypertable.
It works, but displays "ERROR: origin must be before the given date", which is
a bit misleading.

This happens because 95804069 didn't modify invalidation_threshold_compute()
in the same way as tsl_process_continuous_agg_viewstmt(). As a result,
compute_circumscribed_bucketed_refresh_window_for_months() calls
time_bucket_ng() for ts_time_get_min() argument. It doesn't work because
the minimum time is 4714-11-24 BC which is before any reasonable origin,
and the user sees corresponding error message.

This patch fixes the problem in the same way as 95804069 did and adds
a corresponding test.

### DIFF
--- a/tsl/test/sql/continuous_aggs_variable_size_buckets.sql
+++ b/tsl/test/sql/continuous_aggs_variable_size_buckets.sql
@@ -491,3 +491,54 @@ DROP TABLE conditions_dist CASCADE;
 DROP DATABASE :DATA_NODE_1;
 DROP DATABASE :DATA_NODE_2;
 DROP DATABASE :DATA_NODE_3;
+
+-- Test the specific code path of creating a CAGG on top of empty hypertable.
+
+CREATE TABLE conditions_empty(
+  day DATE NOT NULL,
+  city text NOT NULL,
+  temperature INT NOT NULL);
+
+SELECT create_hypertable(
+  'conditions_empty', 'day',
+  chunk_time_interval => INTERVAL '1 day'
+);
+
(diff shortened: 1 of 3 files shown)


fabriziomello (approved):

tsl/test/sql/continuous_aggs_variable_size_buckets.sql
```diff
@@ -491,3 +491,24 @@ DROP TABLE conditions_dist CASCADE;
...
+);
+
+CREATE MATERIALIZED VIEW conditions_summary_empty
+WITH (timescaledb.continuous, timescaledb.materialized_only=true) AS
+SELECT city,
+   timescaledb_experimental.time_bucket_ng('1 month', day) AS bucket,
+   MIN(temperature),
+   MAX(temperature)
+FROM conditions_empty
+GROUP BY city, bucket;
```
gayyappan, line 514: you should add a test that shows that refresh works after you add data to the table: 
1) add data to the hypertable.
2) call refresh_continuous_aggregate.
My primary concern is that this code changes the record that gets added to the invalidation log for an empty table and does this have any unintended side-effects?

afiskon: Thanks, @gayyappan I added the test 🙏

pmwkaa (approved):

PR #3947, opened 2021-12-21
Use the chunk index directly to determine the number of chunks created after the given one

akuzm:
test another approach to https://github.com/timescale/timescaledb/pull/3912


akuzm:
Checking max chunk id doesn't make much sense because they are global across hypertables https://github.com/timescale/timescaledb/pull/3912#discussion_r773914824

PR #3948, opened 2021-12-21
Avoid double PQclear() in data fetchers

erimatnor:
The macro `remote_result_elog` is used in the cursor and row-by-row
fetchers and clears the result value using the function `PQclear()`
before rethrowing its error. However, this error is later captured to
clear the wrapping result value that holds a reference to the original
result. To avoid this double-clearing, set the response to NULL before
the error is thrown.

### DIFF
--- a/tsl/src/remote/cursor_fetcher.c
+++ b/tsl/src/remote/cursor_fetcher.c
@@ -12,6 +12,7 @@
 #include <postgres.h>
 #include <lib/stringinfo.h>
 #include <utils/rel.h>
+#include <utils/guc.h>
 
 #include "utils.h"
 #include "async.h"
@@ -305,7 +306,13 @@ cursor_fetcher_fetch_data_complete(CursorFetcher *cursor)
 
 		/* On error, report the original query, not the FETCH. */
 		if (PQresultStatus(res) != PGRES_TUPLES_OK)
+		{
+			/* remote_result_elog will call PQclear() on the result, so need
+			 * to mark the response as NULL to avoid double PQclear() */
+			pfree(response);
+			response = NULL;
 			remote_result_elog(res, ERROR);
+		}
 
 		/* Convert the data into HeapTuples */
 		numrows = PQntuples(res);
--- a/tsl/src/remote/row_by_row_fetcher.c
+++ b/tsl/src/remote/row_by_row_fetcher.c
@@ -174,7 +174,14 @@ row_by_row_fetcher_complete(RowByRowFetcher *fetcher)
 
 			if (!(PQresultStatus(res) == PGRES_SINGLE_TUPLE ||
 				  PQresultStatus(res) == PGRES_TUPLES_OK))
+			{
+				/* remote_result_elog will call PQclear() on the result, so
+				 * need to mark the response as NULL to avoid double
+				 * PQclear() */
+				pfree(response);
+				response = NULL;
 				remote_result_elog(res, ERROR);
+			}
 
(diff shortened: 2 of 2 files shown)


tsl/src/remote/cursor_fetcher.c
```diff
@@ -303,9 +304,22 @@ cursor_fetcher_fetch_data_complete(CursorFetcher *cursor)
 
 		MemoryContextSwitchTo(cursor->state.batch_mctx);
 
+		const char *opt = GetConfigOption("timescaledb.generate_error", true, false);
```
erimatnor, line 307: Added this config option as a temporary way to inject an error at the right place in the code. Should not be merged with the fix.

To reproduce the crash, first remote the `response = NUL`, then set this session variable:

```
SET timescaledb.generate_error=true;
```

And run an update query against a distributed hypertable.

fabriziomello (approved):

tsl/src/remote/row_by_row_fetcher.c
```diff
@@ -174,7 +174,13 @@ row_by_row_fetcher_complete(RowByRowFetcher *fetcher)
 
 			if (!(PQresultStatus(res) == PGRES_SINGLE_TUPLE ||
 				  PQresultStatus(res) == PGRES_TUPLES_OK))
+			{
+				/* remote_result_elog will call PQclear() on the result, so
```
akuzm, line 178: What about `pfree(AsyncResponseResult)`, is it not needed here? 

Maybe it would be simpler to remove this PQClear from remote_result_elog, if we wrap it in another try/catch anyway.

akuzm (approved):

tsl/src/remote/row_by_row_fetcher.c
erimatnor: AsyncResponseResult is manged by a PostgreSQL memory context, so isn't strictly needed, although we can  of course do that too.

The libpq library, which PQclear belongs to, is not using a memory context and is instead malloc'd so it is much more important to free that, although we have added extra functionality to also auto-free that memory on connection close if not freed earlier. Still, we'd like to PQclear results as soon as possible to not accumulate too much memory on long-lived connections.

I considered removing the PQclear from remote_result_elog but didn't want to do that now because I am unsure of the consequences for other places that use remote_result_elog.

I think the whole async library is in need of a refactoring, so would rather leave that for later and focus on fixing the immediate issue.

erimatnor:
Been working on an isolation test to trigger a serialization error (conflict) on a data node, but the isolation test suite doesn't work well for the multi-node scenario since it doesn't realize a session is blocked (on e.g., a lock) when the lock happens on a data node session and not the monitored access node session managed by the test suite. Will see if adding a test later is a possibility.

tsl/src/remote/row_by_row_fetcher.c
erimatnor: added explicit pfrees

mfundul:
> Been working on an isolation test to trigger a serialization error (conflict) on a data node, but the isolation test suite doesn't work well for the multi-node scenario since it doesn't realize a session is blocked (on e.g., a lock) when the lock happens on a data node session and not the monitored access node session managed by the test suite. Will see if adding a test later is a possibility.

The isolation tester looks a the `pg_locks` view in the Access Node. I wonder if it's possible to hack this somehow, and propagate Data Node locks back to the Access Nodes with unique identifiers. It doesn't sound easy though.

Issue #3909, opened 2021-12-10
[Enhancement]: Distribute ALTER and DROP SCHEMA to data nodes


erimatnor:
### What type of enhancement is this?

User experience

### What subsystems and features will be improved?

Multi-node, Node management

### What does the enhancement do?

We currently distribute CREATE SCHEMA to data nodes, but not the corresponding ALTER and DROP SCHEMA commands.

erimatnor:
DROP SCHEMA currently works if the schema contains a distributed hypertable. This should change.

PR #3949, opened 2021-12-22
Add support for DROP SCHEMA on multi-node

pmwkaa:
This PR changes logic and allows DROP SCHEMA command to be executed
on each data node, even if the schema being dropped has no distributed
hypertables in it.

Fix: #3909

### DIFF
--- a/tsl/src/remote/dist_ddl.c
+++ b/tsl/src/remote/dist_ddl.c
@@ -368,19 +368,24 @@ dist_ddl_process_drop(const ProcessUtilityArgs *args)
 {
 	DropStmt *stmt = castNode(DropStmt, args->parsetree);
 
+	/* For DROP TABLE and DROP SCHEMA operations hypertable_list will be empty */
 	if (list_length(args->hypertable_list) == 0)
 	{
-		/*
-		 * CASCADE operations of DROP TABLE and DROP SCHEMA are handled in
-		 * sql_drop trigger.
-		 */
-
-		/* For DROP TABLE and DROP SCHEMA operations hypertable_list will be
-		 * empty. Wait for sql_drop events.
-		 */
-		if (stmt->removeType == OBJECT_TABLE || stmt->removeType == OBJECT_SCHEMA)
-			dist_ddl_state_schedule(DIST_DDL_EXEC_ON_END, args);
-
+		switch (stmt->removeType)
+		{
+			case OBJECT_TABLE:
+				/* Wait for further sql_drop events  */
+				dist_ddl_state_schedule(DIST_DDL_EXEC_ON_END, args);
+				break;
+			case OBJECT_SCHEMA:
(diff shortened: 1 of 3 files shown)


tsl/src/remote/dist_ddl.c
```diff
@@ -368,19 +368,24 @@ dist_ddl_process_drop(const ProcessUtilityArgs *args)
 {
 	DropStmt *stmt = castNode(DropStmt, args->parsetree);
 
+	/* For DROP TABLE and DROP SCHEMA operations hypertable_list will be empty */
 	if (list_length(args->hypertable_list) == 0)
 	{
```
akuzm, line 373: If the `hypertable_list` is not empty, and `removeType == OBJECT_TABLE`, what would that mean?

tsl/src/remote/dist_ddl.c
```diff
@@ -368,19 +368,24 @@ dist_ddl_process_drop(const ProcessUtilityArgs *args)
...
-
+		switch (stmt->removeType)
+		{
+			case OBJECT_TABLE:
+				/* Wait for further sql_drop events  */
+				dist_ddl_state_schedule(DIST_DDL_EXEC_ON_END, args);
+				break;
+			case OBJECT_SCHEMA:
+				/* Forward DROP SCHEMA command to all data nodes, following
+				 * sql_drop events will be ignored */
```
akuzm, line 382: A completely clueless question: why do we have to execute DROP SCHEMA immediately, in contrast to DROP TABLE?

akuzm (approved):
I don't really understand the code but the tests make sense.

tsl/src/remote/dist_ddl.c
akuzm: Or what does the default branch in the following switch/case means... looks like this needs more assertions/comments.

pmwkaa: I believe hypertable can't be added into the hypertable list, because drop operation is already happened by PostgreSQL standard hook (src/process_utility.c) at this point and it can't be resolved

tsl/src/remote/dist_ddl.c
pmwkaa: Previously we did not support `CREATE SCHEMA` explicitly on data nodes and when`DROP SCHEMA CASCADE` was executed, we only shipped it by analyzing distributed hypertables which we part of that schema. No it is much simpler, we assume schema exists on all data nodes and we can drop it right away.

tsl/src/remote/dist_ddl.c
akuzm: Should it be an `Assert` then, with this explanation as a comment?

afiskon (approved):

erimatnor (approved):

PR #3961, opened 2022-01-06
Fix crash in EXPLAIN VERBOSE on dist table

### LINKED ISSUE #3960: [Bug]: crash on EXPLAIN VERBOSE UPDATE on a distributed hypertable
### What type of bug is this?

Crash

Query planner

### What happened?

The following set of simple queries are causing a crash when using `EXPLAIN VERBOSE`

```
CREATE TABLE example (a timestamp, b integer, c integer);  
SELECT create_distributed_hypertable('example', 'a'); 
INSERT into example values (now(), 1, 2);   
EXPLAIN VERBOSE UPDATE example  set c = c + 0.1;
```

The crash is occurring in `fdw_scan_explain` function because the `fsstate` passed in is NULL and we try to access the `fetcher_type` field from it in VERBOSE mode.

When we are doing EXPLAIN, we short circuit the init foreign scan calls and that's why `fsstate` is not filled up

2.5.1

### PostgreSQL version used

13.2

Mac OSC

Source

Not applicable

```bash
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   timescaledb-tsl-2.6.0-dev.so  	0x000000010feaaba3 fdw_scan_explain + 179 (scan_exec.c:454)
1   timescaledb-tsl-2.6.0-dev.so  	0x000000010fea5ce2 explain_foreign_scan + 66 (fdw.c:304)
2   postgres                      	0x000000010f025264 show_foreignscan_info + 132
3   postgres                      	0x000000010f021905 ExplainNode + 7109 (explain.c:1867)
4   postgres                      	0x000000010f026bde ExplainMemberNodes + 78 (explain.c:3815)
5   postgres                      	0x000000010f02236a ExplainNode + 9770 (explain.c:2083)
6   postgres                      	0x000000010f01eebb ExplainPrintPlan + 283 (explain.c:755)
7   postgres                      	0x000000010f01e99e ExplainOnePlan + 638 (explain.c:586)
8   postgres                      	0x000000010f01df73 ExplainOneQuery + 611 (explain.c:407)
9   postgres                      	0x000000010f01da44 ExplainQuery + 1876 (explain.c:280)
10  postgres                      	0x000000010f3660b5 standard_ProcessUtility + 2325 (utility.c:847)
```
```


```
CREATE TABLE example (a timestamp, b integer, c integer);  
SELECT create_distributed_hypertable('example', 'a'); 
INSERT into example values (now(), 1, 2);   
EXPLAIN VERBOSE UPDATE example  set c = c + 0.1;
```

### PR DESCRIPTION
**Pull request by nikkhils**
The crash is occurring in "fdw_scan_explain" function because the
fsstate passed in is NULL and we try to access the "fetcher_type" field
from it in VERBOSE mode.

When we are doing EXPLAIN, we short circuit the init foreign scan calls
and that's why fsstate is not filled up.

Fixes #3960

### DIFF
--- a/tsl/src/fdw/scan_exec.c
+++ b/tsl/src/fdw/scan_exec.c
@@ -449,7 +449,10 @@ fdw_scan_explain(ScanState *ss, List *fdw_private, ExplainState *es, TsFdwScanSt
 		char *sql;
 
 		ExplainPropertyText("Data node", server->servername, es);
-		ExplainPropertyText("Fetcher Type", explain_fetcher_type(fsstate->fetcher_type), es);
+
+		/* fsstate can be NULL, so check that first */
+		if (fsstate)
+			ExplainPropertyText("Fetcher Type", explain_fetcher_type(fsstate->fetcher_type), es);
 
 		if (chunk_oids != NIL)
 		{
@@ -474,7 +477,8 @@ fdw_scan_explain(ScanState *ss, List *fdw_private, ExplainState *es, TsFdwScanSt
 
 		ExplainPropertyText("Remote SQL", sql, es);
 
-		if (ts_guc_enable_remote_explain)
+		/* fsstate should be set up but better check again to avoid crashes */
+		if (ts_guc_enable_remote_explain && fsstate)
 		{
 			const char *data_node_explain =
 				get_data_node_explain(fsstate->query, fsstate->conn, es);
--- a/tsl/test/expected/dist_hypertable-12.out
+++ b/tsl/test/expected/dist_hypertable-12.out
@@ -229,6 +229,102 @@ INSERT INTO disttable VALUES
        ('2018-07-01 08:01', 29, 1.5),
        ('2018-07-01 09:21', 90, 2.8),
        ('2018-07-01 08:21', 29, 1.2);
+-- EXPLAIN some updates/deletes to see what plans and explain output for
+-- remote operations look like
+EXPLAIN (VERBOSE, COSTS FALSE)
+UPDATE disttable SET temp = 3.7 WHERE device = 1;
+                                                            QUERY PLAN                                                            
+----------------------------------------------------------------------------------------------------------------------------------
+ Update on public.disttable
+   Update on public.disttable
+   Foreign Update on _timescaledb_internal._dist_hyper_1_1_chunk
+     Remote SQL: UPDATE _timescaledb_internal._dist_hyper_1_1_chunk SET temp = $2 WHERE ctid = $1
+   Foreign Update on _timescaledb_internal._dist_hyper_1_2_chunk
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: tsl/src/fdw/scan_exec.c [hunk 1]
```diff
@@ -476,6 +479,7 @@ fdw_scan_explain(ScanState *ss, List *fdw_private, ExplainState *es, TsFdwScanSt
 
 		if (ts_guc_enable_remote_explain)
 		{
+			Assert(fsstate);
```
svenklemm, line 482: since fsstate can be NULL according to comment on line 453 this should be part of the if check otherwise the next line will segfault, asserts that should actually be proper check is a very common source of bugs

nikkhils: If `ts_guc_enable_remote_explain` is enabled then even for EXPLAIN cases, we should be setting up `fsstate`. So asserting that it's present is needed, no? @svenklemm

svenklemm, line 482:
```suggestion
		if (ts_guc_enable_remote_explain && fsstate)
		{
```
My suggestion was to move the assert into the if not drop it completely

nikkhils: I got that. My point is/was that if `ts_guc_enable_remote_explain` is true then `fsstate` HAS to be non-NULL always

nikkhils: I'd rather get an assert or a crash which we can fix later in such a case @svenklemm

svenklemm: I  disagree the code has a lot of crash bugs already so not introducing new possible crashes should be a priority. Also a crash affects all running sessions and is very disruptive so handling errors as graceful as possible seems preferable.

tsl/src/fdw/scan_exec.c
```diff
@@ -449,7 +449,10 @@ fdw_scan_explain(ScanState *ss, List *fdw_private, ExplainState *es, TsFdwScanSt
 		char *sql;
 
 		ExplainPropertyText("Data node", server->servername, es);
-		ExplainPropertyText("Fetcher Type", explain_fetcher_type(fsstate->fetcher_type), es);
+
+		/* fsstate can be NULL, so check that first */
```
akuzm, line 453:
```suggestion
		/*
		 * For the EXPLAIN queries, the fsstate is not set up unless the remote EXPLAIN is
		 * requested with the ts_guc_enable_remote_explain option.
		 */
```

akuzm (approved):

tsl/src/fdw/scan_exec.c
nikkhils: I am not too convinced.. but making the change as above for now..

pmwkaa (approved):

PR #3975, opened 2022-01-11
Fix the build against REL_14_STABLE

afiskon:
The number of arguments of pg_md5_hash() has changed in REL_14_STABLE.
This patch adds a corresponding compat.h version of the procedure.

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=3a0cced8

### DIFF
--- a/src/compat/compat.h
+++ b/src/compat/compat.h
@@ -213,6 +213,72 @@ get_vacuum_options(const VacuumStmt *stmt)
 		   (analyze ? VACOPT_ANALYZE : 0);
 }
 
+/*
+ * The number of arguments of pg_md5_hash() has changed in PG >= 14.2. At the
+ * moment of writing this used to be an upcoming release, thus PG_VERSION_NUM
+ * couldn't be used to distinguish between REL_14_1 and REL_14_STABLE.
+ *
+ * https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=3a0cced8
+ */
+#if PG14
+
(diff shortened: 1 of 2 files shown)


svenklemm (commented):
When adding compatibility macros we make the compatibility functions match the new API so we can easily remove the macros once all supported versions have the new API. Your suggested change does the opposite and doesn't conform to how we normally handle compatibility macros.

afiskon:
@svenklemm 👍 fixed

mkindahl (approved):
LGTM

src/compat/compat.h
```diff
@@ -213,6 +213,72 @@ get_vacuum_options(const VacuumStmt *stmt)
 		   (analyze ? VACOPT_ANALYZE : 0);
 }
 
+/*
+ * The number of arguments of pg_md5_hash() has changed in PG >= 14.2. At the
+ * moment of writing this used to be an upcoming release, thus PG_VERSION_NUM
+ * couldn't be used to distinguish between REL_14_1 and REL_14_STABLE.
```
akuzm, line 219: Does this mean we're going to remove the copy of the function when the 14.2 is released?

akuzm (approved):

src/compat/compat.h
afiskon: Sure, we can do it if we want to. Although so far we never used #if's for minor versions. If that's not some sort of policy, I see no reason to keep a copy of the code.

PR #3977, opened 2022-01-11
Timezones support in CAGGs

afiskon:
This patch allows using timezones in CAGGs, for instance:

time_bucket_ng(bucket_size, ts, 'Europe/Moscow')

Both months/years and days/hours/minutes can be used in bucket_size.
The bucket size is considered variable-sized in all these cases
because DST affects days/hours/minutes as well.

CAGGs on top of distributed hypertables, compressed hypertables and
compressed distributed hypertables are supported as well.

### DIFF
--- a/src/timezones.c
+++ b/src/timezones.c
@@ -1,0 +1,50 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+
+#include <postgres.h>
+#include <port.h>
+#include <pgtime.h>
+#include <datatype/timestamp.h>
+#include <utils/timestamp.h>
+#include <access/xact.h>
+#include "timezones.h"
+
+/* Checks if the given TZ name is valid. */
+bool
+ts_is_valid_timezone_name(const char *tz_name)
+{
+	pg_tz *tz;
+	int tzoff;
+	struct pg_tm tm;
+	fsec_t fsec;
+	const char *abbrev;
+	bool found = false;
+	TimestampTz now = GetCurrentTransactionStartTimestamp();
+	pg_tzenum *tzenum = pg_tzenumerate_start();
+
+	while (true)
+	{
+		tz = pg_tzenumerate_next(tzenum);
+		if (!tz)
+			break;
+
+		/*
+		 * Convert now() to time in this TZ and skip if conversion fails.
+		 * This check is the same that pg_timezone_names() does.
+		 */
+		if (timestamp2tm(now, &tzoff, &tm, &fsec, &abbrev, tz) != 0)
+			continue;
+
+		if ((!strcmp(tz_name, pg_get_timezone_name(tz))) || (abbrev && !strcmp(tz_name, abbrev)))
+		{
+			found = true;
+			break;
+		}
+	}
+
+	pg_tzenumerate_end(tzenum);
+	return found;
+}
--- a/src/ts_catalog/continuous_agg.c
+++ b/src/ts_catalog/continuous_agg.c
@@ -281,6 +281,7 @@ continuous_agg_fill_bucket_function(int32 mat_hypertable_id, ContinuousAggsBucke
 	init_scan_cagg_bucket_function_by_mat_hypertable_id(&iterator, mat_hypertable_id);
 	ts_scanner_foreach(&iterator)
 	{
+		const char *bucket_width_str;
 		Datum values[Natts_continuous_aggs_bucket_function];
 		bool isnull[Natts_continuous_aggs_bucket_function];
 
@@ -300,9 +301,17 @@ continuous_agg_fill_bucket_function(int32 mat_hypertable_id, ContinuousAggsBucke
 		Assert(!isnull[Anum_continuous_aggs_bucket_function_name - 1]);
 		bf->name = TextDatumGetCString(values[Anum_continuous_aggs_bucket_function_name - 1]);
 
+		/*
+		 * So far bucket_width is stored as TEXT for flexibility, but it's type
+		 * most likely is going to change to Interval when the variable-sized
+		 * buckets feature will stabilize.
+		 */
 		Assert(!isnull[Anum_continuous_aggs_bucket_function_bucket_width - 1]);
-		bf->bucket_width =
+		bucket_width_str =
 			TextDatumGetCString(values[Anum_continuous_aggs_bucket_function_bucket_width - 1]);
+		Assert(strlen(bucket_width_str) > 0);
+		bf->bucket_width = DatumGetIntervalP(
+			DirectFunctionCall3(interval_in, CStringGetDatum(bucket_width_str), InvalidOid, -1));
 
 		Assert(!isnull[Anum_continuous_aggs_bucket_function_origin - 1]);
 		bf->origin = TextDatumGetCString(values[Anum_continuous_aggs_bucket_function_origin - 1]);
@@ -407,6 +416,7 @@ ts_continuous_agg_get_all_caggs_info(int32 raw_hypertable_id)
 static const char *
 bucket_function_serialize(const ContinuousAggsBucketFunction *bf)
 {
+	char *bucket_width_str;
 	StringInfo str;
 
 	if (NULL == bf)
@@ -415,14 +425,15 @@ bucket_function_serialize(const ContinuousAggsBucketFunction *bf)
 	str = makeStringInfo();
 
 	/* We are pretty sure that user can't place ';' character in these fields */
-	Assert(strstr(bf->bucket_width, ";") == NULL);
 	Assert(strstr(bf->origin, ";") == NULL);
 	Assert(strstr(bf->timezone, ";") == NULL);
 
+	bucket_width_str =
+		DatumGetCString(DirectFunctionCall1(interval_out, IntervalPGetDatum(bf->bucket_width)));
 	appendStringInfo(str,
 					 "%d;%s;%s;%s;",
 					 BUCKET_FUNCTION_SERIALIZE_VERSION,
-					 bf->bucket_width,
+					 bucket_width_str,
 					 bf->origin,
 					 bf->timezone);
 
@@ -477,7 +488,9 @@ bucket_function_deserialize(const char *str)
 	bf = palloc(sizeof(ContinuousAggsBucketFunction));
 	bf->experimental = true;
 	bf->name = "time_bucket_ng";
-	bf->bucket_width = strings[1];
+	Assert(strlen(strings[1]) > 0);
+	bf->bucket_width = DatumGetIntervalP(
+		DirectFunctionCall3(interval_in, CStringGetDatum(strings[1]), InvalidOid, -1));
 	bf->origin = strings[2];
 	bf->timezone = strings[3];
 	return bf;
@@ -1391,12 +1404,11 @@ watermark_create(const ContinuousAgg *cagg, MemoryContext top_mctx)
 		{
 			/*
 			 * Since `value` is already bucketed, `bucketed = true` flag can
-			 * be added to ts_time_bucket_and_add_months() as an optimization,
-			 * if necessary.
+			 * be added to ts_compute_beginning_of_the_next_bucket_variable() as
+			 * an optimization, if necessary.
 			 */
-			w->value = ts_time_bucket_and_add_months(value,
-													 ts_bucket_function_to_bucket_width_in_months(
-														 cagg->bucket_function));
+			w->value =
+				ts_compute_beginning_of_the_next_bucket_variable(value, cagg->bucket_function);
 		}
 		else
 		{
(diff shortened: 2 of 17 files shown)


pfried:
@afiskon Saw this PR with an exciting title, is there any documentation / plans around supporting timezones in caggs?

afiskon:
@pfried sure, the documentation is going to follow this PR after it gets merged. You can use the tests as a reference for now.

afiskon:
@fabriziomello @mfundul There is one corner case left that needs to be investigated (see the comments). Other than that the patch is ready for review.

src/timezones.c
```diff
@@ -0,0 +1,83 @@
...
+	{
+		Oid argtypes[] = { TEXTOID };
+		savedPlanNormalizeTimezone = SPI_prepare("SELECT name FROM pg_catalog.pg_timezone_names "
+												 "WHERE name = $1 OR abbrev = $1 "
+												 "ORDER BY name LIMIT 1",
+												 1,
+												 argtypes);
+
+		if (savedPlanNormalizeTimezone == NULL)
+			elog(ERROR, "Error preparing query");
```
erimatnor, line 46: Non-conforming error. Also, I think this requires a bit more context as it would be hard to understand what this means.

```suggestion
			elog(ERROR, "error preparing query");
```

src/timezones.c
```diff
@@ -0,0 +1,83 @@
...
+
+		if (SPI_keepplan(savedPlanNormalizeTimezone))
+			elog(ERROR, "Error keeping query plan");
```
erimatnor, line 49: Same here about giving context:

```suggestion
			elog(ERROR, "error keeping query plan");
```

src/timezones.c
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	res = SPI_execute_plan(savedPlanNormalizeTimezone,
+						   qvalues,
+						   NULL,
+						   true /* read only */,
+						   1 /* count */);
+
+	if (res < 0)
+		elog(ERROR, "Failed to execute query plan");
```
erimatnor, line 59:
```suggestion
		elog(ERROR, "failed to execute query plan");
```

src/timezones.c
```diff
@@ -0,0 +1,83 @@
...
+	char *buffer;
+	Datum qvalues[] = { CStringGetTextDatum(tz_name) };
+
+	if (SPI_connect() != SPI_OK_CONNECT)
+		elog(ERROR, "could not connect to SPI");
+
+	if (savedPlanNormalizeTimezone == NULL)
+	{
+		Oid argtypes[] = { TEXTOID };
+		savedPlanNormalizeTimezone = SPI_prepare("SELECT name FROM pg_catalog.pg_timezone_names "
```
erimatnor, line 39: Can we do this with an internal catalog scan like we normally do for native code or use a utility function?

If you look at the view `pg_timezone_names` it is actually just generated from calling the same named function, which in turn enumerates the timezones with `pg_tzenumerate_start()`, which you could call directly in native code in a similar way.

There's also a PostgreSQL function called `FetchDynamicTimeZone` in `datetime.c`  that shows how you can use `pg_tzset` to lookup timezones. It is using an internal cache.

tsl/src/continuous_aggs/create.c
```diff
@@ -722,6 +726,39 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+					Assert(tz->consttype == TEXTOID);
+					const char *orig_tz_name = TextDatumGetCString(tz->constvalue);
+					const char *tz_name = ts_validate_and_normalize_timezone_name(orig_tz_name);
+
+					if (!tz_name)
+					{
+						ereport(ERROR,
+								(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+								 errmsg("invalid timezone name"),
+								 errhint("%s", orig_tz_name)));
```
erimatnor, line 748: I hint should suggest a way to fix/avoid the error. Prefer just folding the timezone name into the error.

```suggestion
								 errmsg("invalid timezone name \"%s\"", orig_tz_name)));
```

src/timezones.c
afiskon: Also, I doubt that doing a manual scan of the table is necessarily better than executing a cached SPI plan.

src/ts_catalog/continuous_agg.c
```diff
@@ -1486,35 +1498,114 @@ ts_continuous_agg_bucket_width(const ContinuousAgg *agg)
...
+								   IntervalPGetDatum(bf->bucket_width),
+								   timestamp,
+								   CStringGetTextDatum(bf->timezone));
+	}
+	else
+	{
+		return DirectFunctionCall2(ts_time_bucket_ng_timestamp,
+								   IntervalPGetDatum(bf->bucket_width),
+								   timestamp);
+	}
```
fabriziomello, line 1517:
```suggestion

	return DirectFunctionCall2(ts_time_bucket_ng_timestamp,
								   IntervalPGetDatum(bf->bucket_width),
								   timestamp);
```

Minor simplification!

src/ts_catalog/continuous_agg.c
```diff
@@ -1486,35 +1498,114 @@ ts_continuous_agg_bucket_width(const ContinuousAgg *agg)
...
+	{
+		// AALEKSEEV TODO: should we account for the timezone here? (e.g. gaps in the time)
+		start_new = DirectFunctionCall2(timestamp_pl_interval,
+										start_new,
+										IntervalPGetDatum(bf->bucket_width));
+	}
+
+	*start = ts_time_value_to_internal(start_new, TIMESTAMPOID);
+	*end = ts_time_value_to_internal(end_new, TIMESTAMPOID);
+}
```
fabriziomello, line 1554: The logic of this function is pretty similar of the circumscribed function below. What about simplify it creating an internal function passing an parameter indicating to calculate `start_new` or `end_new`???

src/timezones.c
fabriziomello: @afiskon I think the main @erimatnor's concern is because as far I understood since I joined the team is we try to avoid use SPI as far we can.

afiskon: @fabriziomello because.... ?

src/ts_catalog/continuous_agg.c
afiskon: That's usually not such a good idea. You save 5-10 lines of code for sure, but also create a **dependency** of two similar but unrelated procedures. This has a tendency to end up as a single procedure with 10 if-else's inside in the long run, not to mention the fact that the comments explaining what are the bool arguments for turn into a mess.

src/timezones.c
mfundul: I don't see the benefit of SPI here either. It's not faster, the code stack is deeper, more things can go wrong and it's not cleaner to look at as far as code maintainability/readability is concerned.

src/ts_catalog/continuous_agg.c
```diff
@@ -1487,34 +1499,137 @@ ts_continuous_agg_bucket_width(const ContinuousAgg *agg)
 }
 
 /*
- * Determines bucket size in months for given bucketing function.
+ * Calls one of time_bukcket_ng() versions depending on the arguments. This is
```
mfundul, line 1502:
```suggestion
 * Calls one of time_bucket_ng() versions depending on the arguments. This is
```

src/timezones.c
erimatnor: `pg_tzenumerate_start()` is basically the thing that gets called underneath the hood in the SPI query, so might as well just call that directly for cleaner code.

I am also wondering, however, if there isn't a way to normalize time zones with existing built-in API functions? The query here essentially reads all the timezone files from disk, which seems a bit inefficient. Isn't there a place where this information is cached?

src/ts_catalog/continuous_agg.c
erimatnor, line 1502:
```suggestion
 * Calls one of time_bucket_ng() versions depending on the arguments. This is
```

src/ts_catalog/continuous_agg.c
```diff
@@ -1487,34 +1499,137 @@ ts_continuous_agg_bucket_width(const ContinuousAgg *agg)
...
+ * a common procedure used by ts_compute_* below.
+ */
+static Datum
+generic_time_bucket_ng(const ContinuousAggsBucketFunction *bf, Datum timestamp)
+{
+	if (strlen(bf->timezone) > 0)
```
erimatnor, line 1508: Does this account for NULL as opposed to empty string (`""`)? Should we also have a NULL check here?

src/ts_catalog/continuous_agg.c
```diff
@@ -1487,34 +1499,137 @@ ts_continuous_agg_bucket_width(const ContinuousAgg *agg)
...
+ *     start = time_bucket(bucket_size, start) + interval bucket_size
+ *
+ */
+void
+ts_compute_inscribed_bucketed_refresh_window_variable(int64 *start, int64 *end,
+													  const ContinuousAggsBucketFunction *bf)
+{
+	Datum start_old, end_old, start_new, end_new;
+
+	start_old = ts_internal_to_time_value(*start, TIMESTAMPOID);
```
erimatnor, line 1573: It is non-obvious why hard-coding `TIMESTAMPOID` is OK here. There's nothing in the function interface that lets you know that this is an operation on start and end as a timestamp.


So, my question is if it is OK to implicitly assume `TIMESTAMP` type here? Should we pass the type as an argument to the function instead? At the very least, I think the function name should convey that we are computing this under the assumption that start and end is in a timestamp range.

src/ts_catalog/continuous_agg.c
```diff
@@ -1487,34 +1499,137 @@ ts_continuous_agg_bucket_width(const ContinuousAgg *agg)
...
+ * if(end != time_bucket(bucket_size, end))
+ *     end = time_bucket(bucket_size, end) + interval bucket_size
+ */
+void
+ts_compute_circumscribed_bucketed_refresh_window_variable(int64 *start, int64 *end,
+														  const ContinuousAggsBucketFunction *bf)
+{
+	Datum start_old, end_old, start_new, end_new;
+
+	start_old = ts_internal_to_time_value(*start, TIMESTAMPOID);
```
erimatnor, line 1604: Same question here as above about hard-coding timestamp type.

src/timezones.c
afiskon: @erimatnor there are less than 600 rows in the table and the query takes about 60 ms because the table is cached in memory. It reads nothing from the disk.

@mfundul This is an odd argument considering the fact that the benefits of the alternative approach were not stated so far either.

> I am also wondering, however, if there isn't a way to normalize time zones with existing built-in API functions?

That's in fact a good question. I will investigate.

src/ts_catalog/continuous_agg.c
afiskon: Thanks for the question. bf->timezone can't be NULL. I will add a corresponding comment and an Assert.

src/ts_catalog/continuous_agg.c
afiskon: Yes, it's OK to use TIMESTAMPOID. I will add a comment.

src/timezones.c
mfundul: I see benefits in not calling and not error checking the following functions:
`SPI_connect()`, `SPI_prepare()`, `SPI_keepplan()`, `SPI_getbinval()`, `SPI_palloc()`, `SPI_finish()` or using OIDs such as `argtypes`.
As far as readability is concerned, it doesn't look good. Something like the following would read better in my opinion:
```
	for (tzenum = pg_tzenumerate_start(), tz = pg_tzenumerate_next(tzenum) ; tz; tz = pg_tzenumerate_next(tzenum))
	{
		if (timestamp2tm(GetCurrentTransactionStartTimestamp(),
						 &tzoff, &tm, &fsec, &tzn, tz) != 0)
			continue;

        name = pg_get_timezone_name(tz);
        abbrev = tzn ? tzn : "";

		if (!strcmp(tz_name, name) || !strcmp(tz_name, abbrev))
		{
			found = true;
			tz_name = CStringGetTextDatum(name);
			break;
		}
	}
```

afiskon: Thanks for all the great feedback! Here is a new version of the patch based on pg_tzenumerate_* as @mfundul suggested.

src/ts_catalog/continuous_agg.c
afiskon: Here is the updated patch with added comments.

src/ts_catalog/continuous_agg.c
afiskon: Here is the updated patch with added comments.

src/timezones.c
```diff
@@ -0,0 +1,51 @@
...
+			break;
+
+		/* Convert now() to time in this TZ and skip if conversion fails. */
+		if (timestamp2tm(now, &tzoff, &tm, &fsec, &abbrev, tz) != 0)
+			continue;
+
+		/* Skip long abbreviations the same way pg_timezone_names() does. */
+		if (abbrev && strlen(abbrev) > 31)
+			continue;
+
```
mfundul, line 41: I think we can remove those lines completely since we already do `strcmp(tz_name, abbrev)`

afiskon: I don't think we can do it. This will make the behavior different from the PostgreSQL one. If there is a long abbreviation in the database (which can be the internal PostgreSQL one, or the system TZif file, depending on the configure flags) we will allow specifying it, while PostgreSQL will reject it. The rest of the TZ support in the CAGGs will break.

afiskon: On second thought the previous line checks that the TZ will not be rejected, so I think you are right.

tsl/test/expected/cagg_with_timezone.out
```diff
@@ -0,0 +1,957 @@
...
+GROUP BY city, bucket
+WITH NO DATA;
+SELECT mat_hypertable_id AS cagg_id_tz
+FROM _timescaledb_catalog.continuous_agg
+WHERE user_view_name = 'conditions_summary_tz'
+\gset
+SELECT raw_hypertable_id AS ht_id
+FROM _timescaledb_catalog.continuous_agg
+WHERE user_view_name = 'conditions_summary_tz'
+\gset
```
fabriziomello, line 113:
```suggestion
SELECT mat_hypertable_id AS cagg_id_tz, raw_hypertable_id AS ht_id
FROM _timescaledb_catalog.continuous_agg
WHERE user_view_name = 'conditions_summary_tz'
\gset
```

tsl/test/expected/cagg_with_timezone.out
```diff
@@ -0,0 +1,946 @@
...
+  city  |        month        | min | max 
+--------+---------------------+-----+-----
+ Moscow | 2021-06-01 00:00:00 |  22 |  34
+(1 row)
+
+-- Add some dummy data for two more months (no invalidations test case)
+INSERT INTO conditions_tz (day, city, temperature)
+SELECT ts, city, row_number() OVER ()
+FROM generate_series('2021-07-01 MSK' :: timestamptz, '2021-08-31 MSK', '1 day') as ts,
+     unnest(array['Moscow', 'Berlin']) as city;
```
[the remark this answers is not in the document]
afiskon: 👍 I will change it to generate_series(..., '3 day')

tsl/src/continuous_aggs/create.c
```diff
@@ -722,6 +726,37 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
 							 "time bucket function must reference a hypertable dimension column")));
 
+			if (list_length(fe->args) == 3)
+			{
+				/*
+				 * Currently the third argument of the bucketing function can be
+				 * only a timezone. Only immutable expressions can be specified.
+				 */
+				tz_arg = eval_const_expressions(NULL, lthird(fe->args));
+				if (IsA(tz_arg, Const))
```
mfundul, line 736: Just some code style nit pick. If you did:
```
if (!IsA(tz_arg, Const))
	ereport(ERROR,
		 errmsg("only immutable expressions allowed in time bucket function"),
		 errhint("Use an immutable expression as third argument"
			 " to the time bucket function.")));
```
then you save one level of identation.

afiskon: Good point. I will fix it, thanks

tsl/src/continuous_aggs/create.c
```diff
@@ -722,6 +726,37 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+				{
+					Const *tz = castNode(Const, tz_arg);
+					Assert(tz->consttype == TEXTOID);
+					const char *tz_name = TextDatumGetCString(tz->constvalue);
```
mfundul, line 740: What if this is not `TEXTOID` in a production build? Won't it crash by calling `TextDatumGetCString()` on a different data type?

tsl/src/continuous_aggs/create.c
```diff
@@ -722,6 +726,35 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+				if (!IsA(tz_arg, Const))
+					ereport(ERROR,
+							(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							 errmsg("only immutable expressions allowed in time bucket function"),
+							 errhint("Use an immutable expression as third argument"
+									 " to the time bucket function.")));
+
+				Const *tz = castNode(Const, tz_arg);
+				Assert(tz->consttype == TEXTOID);
+				const char *tz_name = TextDatumGetCString(tz->constvalue);
```
mfundul, line 745: What if this is not TEXTOID in a production build? Won't it crash by calling TextDatumGetCString() on a different data type?

afiskon: It shouldn't happen, is_valid_bucketing_function() will take care of it. I will add a corresponding comment.

mfundul: But `ts_is_valid_timezone_name()` is called after `TextDatumGetCString()`.

mfundul: Whoops, ok, wrong function.

tsl/src/continuous_aggs/create.c
afiskon: This can't happen, is_valid_bucketing_function() will take care of it. I added the corresponding comment to clarify this.

tsl/src/continuous_aggs/create.c
```diff
@@ -722,6 +726,36 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
 						 errmsg(
 							 "time bucket function must reference a hypertable dimension column")));
 
+			if (list_length(fe->args) == 3)
+			{
+				/*
+				 * Currently the third argument of the bucketing function can be
+				 * only a timezone. Only immutable expressions can be specified.
+				 */
```
mfundul, line 734: Isn't it possible that the third argument is not a timezone? E.g.

```

CREATE OR REPLACE FUNCTION timescaledb_experimental.time_bucket_ng(bucket_width INTERVAL, ts DATE, origin DATE) RETURNS DATE
    AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng_date' LANGUAGE C IMMUTABLE PARALLEL SAFE STRICT;

CREATE OR REPLACE FUNCTION timescaledb_experimental.time_bucket_ng(bucket_width INTERVAL, ts TIMESTAMP, origin TIMESTAMP) RETURNS TIMESTAMP
    AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng_timestamp' LANGUAGE C IMMUTABLE PARALLEL SAFE STRICT;

```

afiskon: Currently - no. Origin will be supported in the next PR.

mfundul (approved):

PR #3978, opened 2022-01-11
Don't guess historical/current chunk by ids

akuzm:
For the chunks that don't have the ANALYZE stats, we have to estimate
the number of pages and tuples in some other ways. One thing we factor
in this estimation is the expected fill factor of the chunk. There are
two ways we calculate the fill factor: 
1. For time dimensions, by
looking at whether the now() belongs to the chunks [begin, end)
interval and calculating the fill factor accordingly.  
2. For space
dimensions, by looking at whether the chunk is one of the last chunks
in the hypertable.

To check (2), we used to compare the chunk ids. Turns out it didn't
work correctly, because the chunk ids are global for all hypertables,
and looking at them doesn't tell us anything if there are many
hypertables.  The code that did that was also very slow.

This commit just removes this logic and considers all the chunks on
space dimensions w/o the ANALYZE stats as recent chunks with fill
factor of 0.5.

Part of #3890

### DIFF
--- a/tsl/src/fdw/relinfo.c
+++ b/tsl/src/fdw/relinfo.c
@@ -157,27 +157,34 @@ get_total_number_of_slices(Hyperspace *space)
 }
 
 /*
- * Fillfactor values are between 0 and 1. It's an indication of how much data is in the chunk.
+ * Estimate fill factor for the chunks that don't have ANALYZE statistics.
+ * Fill factor values are between 0 and 1. It's an indication of how much data is
+ * in the chunk, expressed as a fraction of its estimated final size.
  *
- * Two major drivers for estimation is current time and number of chunks created after.
- *
- * Fill factor estimation assumes that data written is 'recent' in regards to time dimension (eg.
- * almost real-time). For the case when writing historical data, given estimates might be more off
- * as we assume that historical chunks have fill factor 1 unless the number of chunks created after
- * is smaller then total number of slices. Even for writing historical data we might not be totally
+ * Fill factor estimation assumes that data written is 'recent' in regards to
+ * time dimension (eg. almost real-time). For the case when writing historical
+ * data, given estimates might be more off as we assume that historical chunks
+ * have fill factor 1. Even for writing historical data we might not be totally
  * wrong since most probably data has monotonically increasing time.
  *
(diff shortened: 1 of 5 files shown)


akuzm:
`[00:05:40] test bgw_launcher                 ... FAILED     9018 ms` -- flaky

tsl/src/fdw/relinfo.c
```diff
@@ -157,77 +157,92 @@ get_total_number_of_slices(Hyperspace *space)
 }
 
 /*
- * Fillfactor values are between 0 and 1. It's an indication of how much data is in the chunk.
+ * Estimate fill factor for the chunks that don't have ANALYZE statistics.
+ * Fill factor values are between 0 and 1. It's an indication of how much data is
+ * in the chunk, expressed as a fraction of its maximum size.
```
erimatnor, line 162: "maximum size" is a bit misleading here as there is really no maximum.

```suggestion
 * in the chunk, expressed as a fraction of its estimated final size.
```

tsl/src/fdw/relinfo.c
```diff
@@ -157,77 +157,92 @@ get_total_number_of_slices(Hyperspace *space)
...
+ * If hypertable uses timestamp type to partition data then there are three
+ * possible scenarios here: we are beyond chunk end time (historical chunk), we
+ * are somewhere in between chunk time boundaries (current chunk) or chunk start
+ * time is in the future (highly unlikely, also treated as current chunk).
  *
- * To explain how number of chunks created after the chunk affects estimation
- * let's imagine that table is space partitioned with one dimension and having 3 partitions. If data
- * is equaliy distributed amount partitions then there will be 3 current chunks. If there are two
- * new chunks created after chunk X then chunk X is the current chunk.
+ * For integer type we assume that all chunks w/o ANALYZE stats are current.
```
erimatnor, line 179: For integer time types, the user needs to specify a time function so we should be able to say when "current time" falls within the chunk boundaries.

erimatnor (approved):

tsl/src/fdw/relinfo.c
akuzm: We didn't do this before, and I'm not sure how important this is, since this logic is a fallback. Let's keep it as is in this PR, and consider adding this logic later. I've updated the comment.

PR #3979, opened 2022-01-11
Update syntactic attribute number

RafiaSabih:
When adjusting attribute numbers of chunk index, also adjust
varattnosyn which is used at the time of reconstructing index
definition.

### DIFF
--- a/src/chunk_index.c
+++ b/src/chunk_index.c
@@ -107,7 +107,11 @@ adjust_expr_attnos(Oid ht_relid, IndexInfo *ii, Relation chunkrel)
 
 		char *attname = get_attname(ht_relid, var->varattno, false);
 		var->varattno = get_attnum(chunkrel->rd_id, attname);
-
+#if PG13_GE
+		var->varattnosyn = var->varattno;
+#else
+		var->varoattno = var->varattno;
+#endif
 		if (var->varattno == InvalidAttrNumber)
 			elog(ERROR, "index attribute %s not found in chunk", attname);
 	}
--- a/test/expected/alter.out
+++ b/test/expected/alter.out
@@ -734,6 +734,34 @@ SELECT hypertable_constraint_name, constraint_name from _timescaledb_catalog.chu
 
 INSERT INTO t_hypertable AS h VALUES ( 1, '2020-01-01 00:01:00', 3.2) ON CONFLICT (id, time) DO UPDATE SET value = h.value + EXCLUDED.value;
 ROLLBACK;
+-- predicate reconstruction when attnos are different in hypertable and chunk
+CREATE TABLE p_hypertable (a integer not null, b integer, c integer);
+SELECT create_hypertable('p_hypertable', 'a', chunk_time_interval => int '3');
+     create_hypertable      
+----------------------------
+ (14,public,p_hypertable,t)
+(1 row)
+
+BEGIN;
+ALTER TABLE p_hypertable DROP COLUMN b, ADD COLUMN d boolean;
+CREATE INDEX idx_ht ON p_hypertable(a, c) WHERE d = FALSE;
+END;
+INSERT INTO p_hypertable(a, c, d) VALUES (1, 1, FALSE);
+\d+ _timescaledb_internal._hyper_14_28_chunk
+                     Table "_timescaledb_internal._hyper_14_28_chunk"
+ Column |  Type   | Collation | Nullable | Default | Storage | Stats target | Description 
+--------+---------+-----------+----------+---------+---------+--------------+-------------
+ a      | integer |           | not null |         | plain   |              | 
+ c      | integer |           |          |         | plain   |              | 
+ d      | boolean |           |          |         | plain   |              | 
+Indexes:
+    "_hyper_14_28_chunk_idx_ht" btree (a, c) WHERE NOT d
+    "_hyper_14_28_chunk_p_hypertable_a_idx" btree (a DESC)
+Check constraints:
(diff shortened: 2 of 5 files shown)


src/chunk_index.c
```diff
@@ -107,6 +107,7 @@ adjust_expr_attnos(Oid ht_relid, IndexInfo *ii, Relation chunkrel)
 
 		char *attname = get_attname(ht_relid, var->varattno, false);
 		var->varattno = get_attnum(chunkrel->rd_id, attname);
+		var->varattnosyn = get_attnum(chunkrel->rd_id, attname);
```
svenklemm, line 110:
```suggestion
#if PG13_GE
		var->varattnosyn = var->varattno;
#else
		var->varoattno = var->varattno;
#endif
```
varattnosyn was named varoattno before PG13

akuzm (approved):
lgtm but needs a test

nikkhils (commented):
@RafiaSabih let's also add a test case for this predicate based index usages using the customer reported repro scenario

test/expected/alter.out
```diff
@@ -734,6 +734,35 @@ SELECT hypertable_constraint_name, constraint_name from _timescaledb_catalog.chu
...
+ Column |            Type             | Collation | Nullable | Default | Storage | Stats target | Description 
+--------+-----------------------------+-----------+----------+---------+---------+--------------+-------------
+ a      | timestamp without time zone |           | not null |         | plain   |              | 
+ c      | integer                     |           |          |         | plain   |              | 
+ d      | boolean                     |           |          |         | plain   |              | 
+Indexes:
+    "_hyper_14_28_chunk_idx_ht" btree (a, c) WHERE NOT d
+    "_hyper_14_28_chunk_p_hypertable_a_idx" btree (a DESC)
+Check constraints:
+    "constraint_34" CHECK (a >= 'Mon Jan 10 00:00:00 2022'::timestamp without time zone AND a < 'Thu Jan 13 00:00:00 2022'::timestamp without time zone)
```
nikkhils, line 762: I believe this output will keep on changing with time. "Jan 10 to Jan 13". Better to use integer `chunk_time_interval` value like `chunk_time_interval => 100000` as an example.

test/expected/alter.out
```diff
@@ -734,6 +734,34 @@ SELECT hypertable_constraint_name, constraint_name from _timescaledb_catalog.chu
...
+----------------------------
+ (14,public,p_hypertable,t)
+(1 row)
+
+BEGIN;
+ALTER TABLE p_hypertable DROP COLUMN b, ADD COLUMN d boolean;
+CREATE INDEX idx_ht ON p_hypertable(a, c) WHERE d = FALSE;
+END;
+INSERT INTO p_hypertable(a, c, d) VALUES (1, 1, FALSE);
+\d+ _timescaledb_internal._hyper_14_28_chunk
```
fabriziomello, line 750: Isn't a consensus to use `test/sql/utils/testsupport.sql` functions to show metadata instead of psql commands?? I.e.: `test.show_columns`, `test.show_indexes`

mfundul: That's good to know, `\d+` is still used in some tests, e.g. when displaying view queries.

svenklemm: hardcoded chunk names should be avoided too if possible

nikkhils: Hmmm. We need to actually describe the chunk to ensure that it's partial index shows up properly here.

RafiaSabih: > Isn't a consensus to use `test/sql/utils/testsupport.sql` functions to show metadata instead of psql commands?? I.e.: `test.show_columns`, `test.show_indexes`

For this test case we require a path which reconstructs the predicate to include deparsing of column 'd'. As per my understanding suggested functions are not covering this.

nikkhils: AIUI, `test.show_indexes` function has an `Expr` column output which should show the predicate expression.

RafiaSabih: > AIUI, `test.show_indexes` function has an `Expr` column output which should show the predicate expression.

```SELECT * FROM test.show_indexes('_timescaledb_internal._hyper_14_28_chunk');
                            Index                            | Columns | Expr | Unique | Primary | Exclusion | Tablespace 
-------------------------------------------------------------+---------+------+--------+---------+-----------+------------
 _timescaledb_internal._hyper_14_28_chunk_idx_ht             | {a,c}   |      | f      | f       | f         | 
 _timescaledb_internal._hyper_14_28_chunk_p_hypertable_a_idx | {a}     |      | f      | f       | f         | 
(2 rows)```
Doesn't give any information of column 'd' here.

Comment (nikkhils) [reply test/expected/alter.out:750]: Ah, ok. Can stick to `\d+` for now in that case.

**Review by nikkhils — APPROVED** [review]
LGTM with all the revisions. Might make sense to add an entry in the commit message saying:

Fixes #3794

File: test/expected/alter.out (hunk 3 above)
Comment (fabriziomello) [reply test/expected/alter.out:750]: @RafiaSabih @nikkhils sorry I missed your comments here, anyway just FYI there is a `test.show_constraints` there:

```sql
select * from test.show_constraints('_timescaledb_internal._hyper_1_1_chunk'::regclass);
  Constraint  | Type | Columns | Index |          Expr          | Deferrable | Deferred | Validated 
--------------+------+---------+-------+------------------------+------------+----------+-----------
 constraint_1 | c    | {a}     | -     | ((a >= 0) AND (a < 3)) | f          | f        | t
(1 row)
```

PR #3984, opened 2022-01-12
Support continuous agg trigger on copy into compressed chunks

gayyappan:
AFTER ROW triggers are not supported on compressed chunks.
Directly call the continuous aggregate trigger function for copies.

This fix is similar to PR 3764 that handles cagg triggers for
inserts into compressed chunks.

Disable-check: commit-count

### DIFF
--- a/src/copy.c
+++ b/src/copy.c
@@ -348,8 +348,8 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, void (*call
 		estate->es_result_relation_info = resultRelInfo;
 #endif
 
-		if (cis->compress_state != NULL)
-			check_resultRelInfo = cis->orig_result_relation_info;
+		if (cis->compress_info != NULL)
+			check_resultRelInfo = cis->compress_info->orig_result_relation_info;
 		else
 			check_resultRelInfo = resultRelInfo;
 
@@ -388,14 +388,14 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, void (*call
 				ExecConstraints(check_resultRelInfo, myslot, estate);
 			}
 
-			if (cis->compress_state)
+			if (cis->compress_info)
 			{
 				TupleTableSlot *compress_slot =
-					ts_cm_functions->compress_row_exec(cis->compress_state, myslot);
+					ts_cm_functions->compress_row_exec(cis->compress_info->compress_state, myslot);
 				/* After Row triggers do not work with compressed chunks. So
 				 * explicitly call cagg trigger here
 				 */
-				if (cis->has_cagg_trigger)
+				if (cis->compress_info->has_cagg_trigger)
 				{
 					Assert(ts_cm_functions->continuous_agg_call_invalidation_trigger);
 					HeapTupleTableSlot *hslot = (HeapTupleTableSlot *) myslot;
--- a/src/copy.c
+++ b/src/copy.c
@@ -392,6 +392,30 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, void (*call
 			{
 				TupleTableSlot *compress_slot =
 					ts_cm_functions->compress_row_exec(cis->compress_state, myslot);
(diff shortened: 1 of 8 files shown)


mfundul:
`compression_insert-12.out` and `compression_insert-13.out` are missing.

svenklemm (commented):
Can we get rid of cagg trigger completely and always do it the way done in this PR?

gayyappan:
> Can we get rid of cagg trigger completely and always do it the way done in this PR?

That is certainly a possibility. This approach does not work yet with distributed hypertables. I hope to fix the distributed hypertables triggers in a follow-up PR. We can consider this after that is fixed.

svenklemm (approved):

mfundul:
> Can we get rid of cagg trigger completely and always do it the way done in this PR?

This is not possible without catalog changes for distributed hypertables. The cagg trigger for Data Nodes needs information that is located at the Access Node (e.g. raw hypertable ID of AN, materialization hypertable ID of AN).

Since we have not modified CAGG metadata yet, they live in the AN database. What we do, and can do, today is get the arguments from the trigger definition that is pushed down to the data nodes during CAGG creation.

So, in effect, the trigger function definition itself works like a catalog and provides us with the required information.

src/copy.c
```diff
@@ -388,10 +388,34 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, void (*call
...
+
+					ts_cm_functions->continuous_agg_call_invalidation_trigger(
+						ht->fd.id,
+						cis->rel,
+						hslot->tuple,
+						NULL /* chunk_newtuple */,
+						false /* update */,
+						false /* is_distributed_hypertable
+							   */
+						,
```
afiskon, line 415: That's strange formatting :) Maybe it would be a bit better to use a couple of const bool's instead?

src/nodes/chunk_dispatch_state.c
```diff
@@ -127,43 +127,44 @@ chunk_dispatch_exec(CustomScanState *node)
...
+			skip_tuple =
+				!ExecBRInsertTriggers(estate, cis->compress_info->orig_result_relation_info, slot);
 
 			if (skip_tuple)
 				return NULL;
 		}
 
 		if (cis->rel->rd_att->constr && cis->rel->rd_att->constr->has_generated_stored)
-			ExecComputeStoredGeneratedCompat(cis->orig_result_relation_info,
+			ExecComputeStoredGeneratedCompat(cis->compress_info->orig_result_relation_info,
```
afiskon, line 149: Codecov says that this line is never triggered by tests. Would it be difficult to fix it?

src/nodes/chunk_insert_state.h
```diff
@@ -15,6 +15,14 @@
 #include "cache.h"
 #include "cross_module_fn.h"
 
+typedef struct CompressChunkInsertState
+{
+	Relation compress_rel;
+	ResultRelInfo *orig_result_relation_info;
+	CompressSingleRowState *compress_state;
+	bool has_cagg_trigger;
+} CompressChunkInsertState;
```
afiskon, line 24: Maybe a few comments would be appropriate, especially considering the fact that this is a public declaration.

afiskon (approved):
Here is my ✅ but please take a look at the comments I left. They are not critical but if this is something that is easy to fix I suggest doing so.

mfundul:
What about distributed hypertable triggers? Is this going to be implemented in a different PR?

src/nodes/chunk_insert_state.h
gayyappan: added.

gayyappan:
> What about distributed hypertable triggers? Is this going to be implemented in a different PR?

Yes. https://github.com/timescale/timescaledb/pull/3984#issuecomment-1013147819

src/nodes/chunk_dispatch_state.c
gayyappan: This PR only makes a syntactic change. I'll look into adding a test case in another PR.

PR #3988, opened 2022-01-13
Add variable-sized buckets RFC to the docs/ directory

afiskon:
We used to store all our RFCs in Google Docs. This approach has several
disadvantages. Most importantly, the RFCs are not available to potential
third-party contributors. Also, the RFCs are prone to "rot" and/or describe
something that was in fact never implemented. You can always have a call with
a colleague who knows the particular part of the system, but this approach
doesn't scale.

The team agreed to start publishing the RFCs on GitHub. This PR adds the first
RFC that describes variable-sized buckets.

### DIFF
--- a/docs/VariableSizedBuckets.md
+++ b/docs/VariableSizedBuckets.md
@@ -1,0 +1,258 @@
+# Variable-Sized Buckets and Timezones Support for Continuous Aggregates
+
+## Goals
+
+Our [survey][survey] showed that:
+
+* Users need the support of buckets for continuous aggregates (caggs) that are
+  N-month and/or N-years in size;
+* There is also a demand for timezones (including daylight saving time) support;
+
+So we have two goals. At the moment of writing, it is the most upvoted
+functionality in our Community Slack.
+
+
+## Background
+
+See [TimescaleDocs / Continuous Aggregates][docs] for the general information
+about caggs.
+
+Currently, we support only buckets that are fixed in size, like 3 hours or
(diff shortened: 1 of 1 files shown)


svenklemm:
When we agreed to include documentation for variable sized buckets we talked about detailed description of the implementation not about describing processes, open issues or future developments. I don't think this kind of information belongs into the repository.

fabriziomello:
Hey @afiskon, maybe there was a misunderstanding about our agreement in add tech docs to the repo and I agree with @svenklemm.

As far I understood for RFCs is fine to use GDocs because it's part of the development process, but at the end we'll result with an implementation plus documentation (user and tech), so the idea is add into the repo a tech description of what we implemented and it's not necessarily all the designs, proposals, discussions we had during the development.

What I understood about the output of our work is:
- code
- tech doc (if applicable)
- user doc (if applicable)

afiskon:
@fabriziomello the team reached a consensus to give it a try on the Retrospective yesterday 🤷‍♂️

afiskon:
For the record - OK so apparently we had a little misunderstanding here because I proposed to publish the RFC, but in the poll, we used "Documentation" and people thought they are voting for publishing design documents which is not the same as RFC. As a result, this PR does not what part of the team thought we agreed on. We agreed to vote explicitly on publishing RFC next Monday when the entire team will gather.

afiskon:
The consensus was reached by the team not to publish RFCs, thus I'm withdrawing this PR.

PR #3997, opened 2022-01-17
Ignore memoize test in pg14 snapshot

svenklemm:
Upstream changed the explain output for the memoize node to include
an additional `Cache Mode` line. While we could adjust our test
runner to always ignore that line this would prevent us from testing
the cache mode in future tests.

https://github.com/postgres/postgres/commit/6c32c0977783fae217b5eaa1d22d26c96e5b0085

### DIFF
--- a/scripts/gh_matrix_builder.py
+++ b/scripts/gh_matrix_builder.py
@@ -155,7 +155,7 @@ if event_type != "pull_request":
   # the stable branches of supported PG releases
   m["include"].append(build_debug_config({"pg":12,"snapshot":"snapshot"}))
   m["include"].append(build_debug_config({"pg":13,"snapshot":"snapshot"}))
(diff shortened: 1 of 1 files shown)


svenklemm:
Since the cache mode can have an influence on the resultset it is better not to ignore it.

mkindahl (approved):

afiskon:
@svenklemm Maybe we should adjust the test for the recent version and ignore on <= PG13 instead?

svenklemm:
> @svenklemm Maybe we should adjust the test for the recent version and ignore on <= PG13 instead?

The test does not run on PG < 14 because there is no Memoize node in those postgres versions. I think our tests should pass on latest released version which they do. Once 14.2 is released we should switch the logic around to have our tests include the cache mode line and be ignored on versions that don't have them. 
Currently there is no released version that has the cache mode line.

fabriziomello (approved):

Issue #3951, opened 2021-12-31
Segmentation fault when execute add_data_node


daydayup863:
### What type of bug is this?

Crash

### What subsystems and features are affected?

Access node

### What happened?

when execute select add_data_node, postgres Segmentation fault, and the database system is in recovery mode.

### TimescaleDB version affected

2.5.1

### PostgreSQL version used

14.1

### What operating system did you use?

CentOS7.6.1810

### What installation method did you use?

Source, RPM

### What platform did you run on?

On prem/Self-hosted

### Relevant log output and stack trace

```bash
[postgres postgres psql [local] 2021-12-31 14:54:00.139 CST 5687 61cea906.1637 3 6/5 0 0]STATEMENT:  select add_data_node(node_name=>'testpgdb3', host=>'testpgdb3', database=>'test_beta', port=>6432, if_not_exists=>true, password=>'729cc20e-e249-4255-858d-08eeaf2b5a');
[    2021-12-31 14:55:41.363 CST 13200 61ce9ee3.3390 13  0 0]LOG:  server process (PID 5779) was terminated by signal 11: Segmentation fault
[    2021-12-31 14:55:41.363 CST 13200 61ce9ee3.3390 14  0 0]DETAIL:  Failed process was running: select add_data_node(node_name=>'testpgdb3', host=>'testpgdb3', database=>'test_beta', port=>6432, if_not_exists=>true, password=>'729cc20e-e249-4255-858d-08eeaf2b5a');
[    2021-12-31 14:55:41.363 CST 13200 61ce9ee3.3390 15  0 0]LOG:  terminating any other active server processes
[test_beta test_beta [unknown] [local] 2021-12-31 14:55:41.364 CST 6728 61cea96d.1a48 1  0 0]FATAL:  the database system is in recovery mode
[    2021-12-31 14:55:41.370 CST 13200 61ce9ee3.3390 16  0 0]LOG:  all server processes terminated; reinitializing
[    2021-12-31 14:55:52.074 CST 6774 61cea978.1a76 1  0 0]LOG:  database system was interrupted; last known up at 2021-12-31 14:50:40 CST
[    2021-12-31 14:55:52.096 CST 6774 61cea978.1a76 2  0 0]LOG:  database system was not properly shut down; automatic recovery in progress
[    2021-12-31 14:55:52.098 CST 6774 61cea978.1a76 3  0 0]LOG:  redo starts at 13/9C003490
```

### How can we reproduce the bug?

```bash
[...]

PR #4007, opened 2022-01-20
Fix crash when execute add_data_node

pmwkaa:
On disconnect libpq will create a result object without
creating event data, which is usually done for a regular errors.

This fix handles this case and forces the result object to always
have have associated connection meta data.

Fix: #3951

### DIFF
--- a/tsl/src/remote/connection.c
+++ b/tsl/src/remote/connection.c
@@ -288,7 +288,16 @@ fill_result_error(TSConnectionError *err, int errcode, const char *errmsg, const
 	const char *sqlstate;
 
 	if (NULL == err || NULL == res || NULL == entry)
+	{
+		if (err)
+		{
+			MemSet(err, 0, sizeof(*err));
+			err->errcode = errcode;
+			err->msg = errmsg;
+			err->nodename = "";
+		}
 		return false;
+	}
 
 	Assert(entry->conn);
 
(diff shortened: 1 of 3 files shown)


tsl/src/remote/connection.c
```diff
@@ -288,7 +288,16 @@ fill_result_error(TSConnectionError *err, int errcode, const char *errmsg, const
...
 
 	if (NULL == err || NULL == res || NULL == entry)
+	{
+		if (err)
+		{
+			MemSet(err, 0, sizeof(*err));
+			err->errcode = errcode;
+			err->msg = errmsg;
+			err->nodename = "";
+		}
```
akuzm, line 298: Why isn't this part covered by the tests, can we add something for it as well?

tsl/src/remote/connection.c
```diff
@@ -867,14 +878,36 @@ remote_connection_get_result_error(const PGresult *res, TSConnectionError *err)
...
-	return PQexec(conn->pg_conn, cmd);
+	res = PQexec(conn->pg_conn, cmd);
+
+	/*
+	 * Workaroud for the libpq disconnect case.
+	 *
+	 * libpq disconnect will create an result object without creating
+	 * events, which is usually done for a regular errors.
+	 *
+	 * In order to be comptible with our errors handling code, force
```
akuzm, line 898:
```suggestion
	 * In order to be compatible with our errors handling code, force
```

akuzm (approved):

tsl/src/remote/connection.c
svenklemm, line 898:
```suggestion
	 * In order to be compatible with our error handling code, force
```

tsl/src/remote/connection.c
```diff
@@ -867,14 +878,36 @@ remote_connection_get_result_error(const PGresult *res, TSConnectionError *err)
...
+	 * create result event, unless the result object does not have
```
svenklemm, line 899:
```suggestion
	 * create result event, if the result object does not have
```

svenklemm (approved):

tsl/src/remote/connection.c
pmwkaa: This is done mostly as precaution, because otherwise error object will be left uninitialized in some calls

PR #4012, opened 2022-01-23
Route delete on Hypertables through CustomScan

svenklemm:
This patch changes DELETE handling for hypertables to have the
postgres ModifyTable node be wrapped in a custom HypertableModify
node. By itself this does not change DELETE handling for hypertables
but instead enables subsequent patches to implement e.g. chunk
exclusion for DELETE or DELETE on compressed chunks.
Since PG 14 codepath for INSERT is different from previous versions
this PR will only change the plan for PG14+.

### DIFF
--- a/test/expected/delete-14.out
+++ b/test/expected/delete-14.out
@@ -111,37 +111,38 @@ WHERE series_1 IN (SELECT series_1 FROM "two_Partitions" WHERE series_1 > series
 EXPLAIN (costs off)
 DELETE FROM "two_Partitions"
 WHERE series_1 IN (SELECT series_1 FROM "two_Partitions" WHERE series_1 > series_val());
-                                                                    QUERY PLAN                                                                    
---------------------------------------------------------------------------------------------------------------------------------------------------
- Delete on "two_Partitions"
-   Delete on "two_Partitions" "two_Partitions_2"
-   Delete on _hyper_1_1_chunk "two_Partitions_3"
-   Delete on _hyper_1_2_chunk "two_Partitions_4"
-   Delete on _hyper_1_3_chunk "two_Partitions_5"
-   Delete on _hyper_1_4_chunk "two_Partitions_6"
-   ->  Hash Join
-         Hash Cond: ("two_Partitions".series_1 = "two_Partitions_1".series_1)
-         ->  Append
-               ->  Seq Scan on "two_Partitions" "two_Partitions_2"
-               ->  Seq Scan on _hyper_1_1_chunk "two_Partitions_3"
-               ->  Seq Scan on _hyper_1_2_chunk "two_Partitions_4"
-               ->  Seq Scan on _hyper_1_3_chunk "two_Partitions_5"
-               ->  Seq Scan on _hyper_1_4_chunk "two_Partitions_6"
-         ->  Hash
-               ->  HashAggregate
-                     Group Key: "two_Partitions_1".series_1
-                     ->  Append
-                           ->  Seq Scan on "two_Partitions" "two_Partitions_7"
-                                 Filter: (series_1 > (series_val())::double precision)
-                           ->  Index Scan using "_hyper_1_1_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_1_chunk "two_Partitions_8"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-                           ->  Index Scan using "_hyper_1_2_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_2_chunk "two_Partitions_9"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-                           ->  Index Scan using "_hyper_1_3_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_3_chunk "two_Partitions_10"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-                           ->  Index Scan using "_hyper_1_4_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_4_chunk "two_Partitions_11"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-(28 rows)
+                                                                       QUERY PLAN                                                                       
+--------------------------------------------------------------------------------------------------------------------------------------------------------
+ Custom Scan (HypertableModify)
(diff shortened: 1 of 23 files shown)


src/nodes/hypertable_insert.c
```diff
@@ -469,6 +478,32 @@ hypertable_insert_plan_create(PlannerInfo *root, RelOptInfo *rel, CustomPath *be
...
+				Var *var = castNode(Var, tle->expr);
+				tle->expr = (Expr *) makeNullConst(var->vartype, var->vartypmod, var->varcollid);
+			}
+		}
+#else
+		/* We only route DELETEs through our CustomNode for PG 14+ because
+		 * the codepath for earlier versions is different. */
+		Assert(false);
+#endif
+	}
```
fabriziomello, line 506:
```suggestion
#if PG14_GE
	/*
	 * For DELETE processed_tlist will have ROWID_VAR. We need to remove
	 * those because set_customscan_references will bail if it sees
	 * ROWID_VAR entries in the targetlist.
	 */
	if (mt->operation == CMD_DELETE)
	{
		ListCell *lc;
		foreach (lc, cscan->scan.plan.targetlist)
		{
			TargetEntry *tle = lfirst_node(TargetEntry, lc);

			if (IsA(tle->expr, Var) && castNode(Var, tle->expr)->varno == ROWID_VAR)
			{
				Var *var = castNode(Var, tle->expr);
				tle->expr = (Expr *) makeNullConst(var->vartype, var->vartypmod, var->varcollid);
			}
		}
	}
#endif
```

Isn't this way simpler?

src/nodes/hypertable_insert.c
```diff
@@ -685,14 +755,86 @@ ExecModifyTable(PlanState *pstate)
...
+
+			if (relkind == RELKIND_RELATION || relkind == RELKIND_MATVIEW ||
+				relkind == RELKIND_PARTITIONED_TABLE)
+			{
+				/* ri_RowIdAttNo refers to a ctid attribute */
+				Assert(AttributeNumberIsValid(resultRelInfo->ri_RowIdAttNo));
+				datum = ExecGetJunkAttribute(slot, resultRelInfo->ri_RowIdAttNo, &isNull);
+				/* shouldn't ever get a null result... */
+				if (isNull)
+					elog(ERROR, "ctid is NULL");
```
fabriziomello, line 783: Turn into Assert?

src/nodes/hypertable_insert.c
```diff
@@ -685,14 +755,86 @@ ExecModifyTable(PlanState *pstate)
...
+			 *
+			 * Other relevant relkinds, currently limited to views, always
+			 * have a wholerow attribute.
+			 */
+			else if (AttributeNumberIsValid(resultRelInfo->ri_RowIdAttNo))
+			{
+				datum = ExecGetJunkAttribute(slot, resultRelInfo->ri_RowIdAttNo, &isNull);
+				/* shouldn't ever get a null result... */
+				if (isNull)
+					elog(ERROR, "wholerow is NULL");
```
fabriziomello, line 811: Turn into Assert?

src/nodes/hypertable_insert.c
svenklemm: i wanted to catch any non-pg14 deletes ending up here

src/nodes/hypertable_insert.c
```diff
@@ -660,6 +709,27 @@ ExecModifyTable(PlanState *pstate)
...
+		 */
+		if (AttributeNumberIsValid(node->mt_resultOidAttno))
+		{
+			Datum datum;
+			bool isNull;
+			Oid resultoid;
+
+			datum = ExecGetJunkAttribute(planSlot, node->mt_resultOidAttno, &isNull);
+			if (isNull)
+				elog(ERROR, "tableoid is NULL");
```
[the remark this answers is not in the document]
svenklemm: this is copied from postgres, also turning ifs into assert can be dangerous and has been the source of many bugs.

fabriziomello (approved):
I did some minor suggestions but is up to you accept or not... otherwise LGTM

test/expected/delete-14.out
```diff
@@ -111,37 +111,38 @@ WHERE series_1 IN (SELECT series_1 FROM "two_Partitions" WHERE series_1 > series
...
-                           ->  Index Scan using "_hyper_1_1_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_1_chunk "two_Partitions_8"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-                           ->  Index Scan using "_hyper_1_2_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_2_chunk "two_Partitions_9"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-                           ->  Index Scan using "_hyper_1_3_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_3_chunk "two_Partitions_10"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-                           ->  Index Scan using "_hyper_1_4_chunk_two_Partitions_timeCustom_series_1_idx" on _hyper_1_4_chunk "two_Partitions_11"
-                                 Index Cond: (series_1 > (series_val())::double precision)
-(28 rows)
+                                                                       QUERY PLAN
```
gayyappan, line 114: Do we have tests for these cases?
1. Triggers for delete (before/after row and stmt)
2. Delete with join on the table  (from PG docs below) 
DELETE FROM films USING producers
  WHERE producer_id = producers.id AND producers.name = 'foo';
3. DELETE ... with RETURNING *

gayyappan (commented):
Would be good to note in the commit message that deletes for distr. hypertables are not yet routed through this node.

Requested some additional tests.

gayyappan (approved):
Code changes look good. Please add the additional tests, if we don't have coverage for these cases.

src/nodes/hypertable_insert.c
svenklemm: this is copied from postgres, also turning ifs into assert can be dangerous and has been the source of many bugs.

test/expected/delete-14.out
svenklemm: ok added test for DELETE RETURNING, there were already tests for the other stuff

PR #4021, opened 2022-01-25
Add successful_duration counters to job_stats

jamessewell:
Sometimes it is useful to know the min, max, mean and total run durations for (only) successful jobs.

This adds that.

I haven't added tests because it wasn't immediately obvious how the tests worked - happy to help there too.

### DIFF
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -237,7 +237,10 @@ CREATE TABLE IF NOT EXISTS _timescaledb_internal.bgw_job_stat (
   total_failures bigint NOT NULL,
   total_crashes bigint NOT NULL,
   consecutive_failures int NOT NULL,
-  consecutive_crashes int NOT NULL
+  consecutive_crashes int NOT NULL,
+  max_successful_duration interval NOT NULL,
+  min_successful_duration interval NOT NULL,
+  total_successful_duration interval NOT NULL
 );
 
 --The job_stat table is not dumped by pg_dump on purpose because
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -76,7 +76,27 @@ SELECT ht.schema_name AS hypertable_schema,
   END AS next_start,
   js.total_runs,
   js.total_successes,
-  js.total_failures
+  js.total_failures,
+  CASE WHEN js.total_successes = 0 THEN
+	NULL
+  ELSE 
+  	js.total_successful_duration
+  END AS total_successful_duration,
+  CASE WHEN js.total_successes = 0 THEN
+	NULL
+  ELSE 
+  	js.max_successful_duration
+  END AS max_successful_duration,
+  CASE WHEN js.total_successes = 0 THEN
+	NULL
+  ELSE 
+  	js.min_successful_duration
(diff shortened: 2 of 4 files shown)


svenklemm:
Hello, this is missing migrations scripts for the changed fields in `sql/updates/latest-dev.sql` and downgrade scripts in `sql/updates/reverse-dev.sql`

sql/views.sql
```diff
@@ -76,7 +76,27 @@ SELECT ht.schema_name AS hypertable_schema,
   END AS next_start,
   js.total_runs,
   js.total_successes,
-  js.total_failures
+  js.total_failures,
+  CASE WHEN js.total_successes = 0 THEN
+	NULL
+  ELSE 
+  	js.total_successful_duration
+  END AS total_successful_duration,
```
fabriziomello, line 84:
```suggestion
  NULLIF(js.total_successes, 0) AS total_successful_duration,
```

For better readability

fabriziomello:
Also you should check and change some regression tests, for example:

```bash
➜ git grep 'SELECT \* FROM timescaledb_information.job_stats'
bgw_custom.sql:SELECT * FROM timescaledb_information.job_stats WHERE job_id > 1000;
bgw_db_scheduler.sql:SELECT * FROM timescaledb_information.job_stats;
bgw_db_scheduler.sql:SELECT * FROM timescaledb_information.job_stats;
bgw_reorder_drop_chunks.sql:SELECT * FROM timescaledb_information.job_stats;
bgw_reorder_drop_chunks.sql:SELECT * FROM timescaledb_information.job_stats;
bgw_reorder_drop_chunks.sql:SELECT * FROM timescaledb_information.job_stats;
include/cagg_bgw_common.sql:SELECT * FROM timescaledb_information.job_stats;
```

It's failing on the CI: https://github.com/timescale/timescaledb/runs/4932458868?check_suite_focus=true

sql/views.sql
jamessewell: This won't work, it would pass through `js.total_successes`  when `=0`, but I'm passing through `js.total_successful_duration`

jamessewell: I mean I could do `NULLIF(js.total_successful_duration, js.total_successful_duration + js.total_successes)` - but that feels a bit wonky

jamessewell: I think it's probably fine just to start them at zero and not worry about NULLing?

fabriziomello: Ahh you're correct

sql/pre_install/tables.sql
```diff
@@ -237,7 +237,10 @@ CREATE TABLE IF NOT EXISTS _timescaledb_internal.bgw_job_stat (
   total_failures bigint NOT NULL,
   total_crashes bigint NOT NULL,
   consecutive_failures int NOT NULL,
-  consecutive_crashes int NOT NULL
+  consecutive_crashes int NOT NULL,
```
fabriziomello, line 240: is this something left behind? cause it's used nowhere

fabriziomello:
Hey @jamessewell, are u planning to keep working on it?

PR #4024, opened 2022-01-25
Fix premature cache release call

gayyappan:
The cache entry for a hypertable is created by calling
ts_hypertable_from_tupleinfo. This sets up the ht->space
structure, which in turn sets up the dimension information.
These structures are allocated in the cache's memory context.
The dimension information is accessed after the cache is
released in ts_subtract_integer_from_now. This PR fixes this
by releasing the cache before returning from the function.

Fixes #4014

PR #4026, opened 2022-01-26
Add coccinelle scripts for detecting use-after-free

svenklemm:
This patch adds coccinelle scripts for detecting use-after-free
bugs in relation to ts_cache_release.

This will find code using the following pattern:
Use of the hypertable from cache after releasing the cache
and use of the dimension of the hypertable from a cache
after the cache was released.

This will detect bugs similar to #4024, #3106, #3104

### DIFF
--- a/.github/workflows/coccinelle.yaml
+++ b/.github/workflows/coccinelle.yaml
@@ -6,15 +6,16 @@ jobs:
   coccinelle:
     name: Coccinelle
     runs-on: ubuntu-latest
+    container:
+      # coccinelle version in ubuntu-latest (20.04) is too old so we run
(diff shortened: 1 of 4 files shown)


akuzm:
```
Run ./scripts/coccinelle.sh
minus: parse error: 
  File "./scripts/../coccinelle/hypertable_cache.cocci", line 37, column 0, charpos = 728
  around = '@',
  whole content = @ safelist2 depends on cache_get && !safelist1 @
 ```
 
Looks like the check doesn't fail if parsing a script failed.

gayyappan (approved):

fabriziomello (approved):

Issue #4001, opened 2022-01-18
[Feature]: a guc to disable forwarding of DDL commands to data nodes


pmwkaa:
### What problem does the new feature solve?

Upcoming 2.6 release will have multi-node support for `CREATE SCHEMA` forwarding, we need a way to disable this functionality if it will cause a problems for a user

### What does the feature do?

Disables DDL forwarding to data nodes. So user will have to use `distributed_exec` or manually create necessary distributed objects

erimatnor:
DDL forwarding should also be disabled any time you alter or install an extension by checking the internal variable `creating_extension`.

PR #4034, opened 2022-01-28
Add a way to disable dist ddl on a global objects

pmwkaa:
This change adds `timescaledb.enable_ddl_on_global_objects` GUC
which allows to enable or disable distributed DDL operations on
a global objects such as SCHEMA, DATABASE or GRANT command.

Basically everything that does not involve particular distributed
hypertable associated with the command can be disabled.

Affected commands:

```
CREATE/DROP SCHEMA
ALTER SCHEMA
ALTER DATABASE
REASSING OWNER
DROP OWNED
ALTER DEFAULT PRIVILEGES
GRANT ON DATABASE
GRANT IN SCHEMA
```
This change also set this GUC to false by default, which disables
forwarding DDL commands on global objects to data nodes by default.

Issue: #4001


tsl/src/remote/dist_ddl.c
```diff
@@ -1070,6 +1066,70 @@ dist_ddl_get_analyze_stats(const ProcessUtilityArgs *args)
...
+		case T_DropOwnedStmt:
+		case T_ReassignOwnedStmt:
+		case T_AlterDefaultPrivilegesStmt:
+			return false;
+
+		case T_DropStmt:
+		{
+			const DropStmt *stmt = castNode(DropStmt, args->parsetree);
+			if (stmt->removeType == OBJECT_SCHEMA)
+				return false;
```
akuzm, line 1096: So this kind of duplicates the logic in `dist_ddl_drop`? I wonder if we should somehow keep it to one place to avoid confusion.

tsl/src/remote/dist_ddl.c
```diff
@@ -980,10 +980,6 @@ dist_ddl_process(const ProcessUtilityArgs *args)
 			dist_ddl_process_truncate(args);
 			break;
 
-		case T_CopyStmt:
-			/* Nothing to do for COPY. */
-			break;
-
```
afiskon, line 986: Why does adding a new GUC need this change?

pmwkaa: That is a leftover from previous work, this case is never reached because it is checked at the beginning of the ddl start function

tsl/src/remote/dist_ddl.c
pmwkaa: I don't think this is related to the `dist_ddl_drop` because it is needed to be scheduled first which is done by the preprocess function first.

But yes, it is a bit of logic duplication in a way that it could be checked separately for each function. But in such case it slightly complicates overall logic, because each function would require separately to check this guc and would require another way to see if the query been skipped for testing.

Anyway, if people are feel strongly about it, I can blend this check to the processing functions.

akuzm (approved):

afiskon (approved):

src/guc.c
```diff
@@ -283,6 +284,17 @@ _guc_init(void)
 							 NULL,
 							 NULL);
 
+	DefineCustomBoolVariable("timescaledb.enable_ddl_on_global_objects",
+							 "Enable distributed DDL operations on global objects",
+							 "Forward DDL operations on global objects to data nodes",
```
erimatnor, line 289: I think it would be good to elaborate in the long description on what we mean with a global object. I don't think it is clear for a user what a "global object" is.

erimatnor (commented):
LGTM, but I have concerns around the GUC name, especially the idea of "global objects", which is not a terminology users might be familiar with.

While I understand the idea of "global objects", it is also a bit troublesome since some commands create the global objects and some are "ddl on them" once they exist. It sounds from the GUC that this blocks only the latter commands.

The other issue is that once we have the option to enable or disable these DDLs, a user could end up with a mix of global and non-global objects that we cannot distinguish between. For instance, take `CREATE SCHEMA` as an example. If the GUC is off, then the schema created will not be global, but if we later turn the GUC on we will still distribute commands (e.g., ALTER SCHEMA) that involve this object, leading to a failure.

I guess these issues are alleviated by having this turned off by default so that we kind of punt on the issue and leave it up to the user to decide. But ultimately, I am wondering if we don't need a way to explicitly create "global objects" and then track which objects are global and non-global.

PR #4037, opened 2022-01-28
Fix status for dropped chunks that have catalog entries

gayyappan:
Chunks that are dropped but preserve the catalog entries
    have an incorrect status when they are marked as dropped.
    This happens if the chunk was previously compressed and then
    gets dropped - the status in the catalog tuple reflects the
    compression status. This should be reset since the data is now
    dropped


svenklemm (approved):

mkindahl (approved):
LGTM

PR #4040, opened 2022-01-31
Bugfix: make add_continuous_aggregate_policy() work with variable-sized buckets

afiskon:
Variable-sized buckets don't work with CAGG policies:

```
SELECT add_continuous_aggregate_policy('conditions_summary_policy',
    start_offset => INTERVAL '65 days',
    end_offset => INTERVAL '1 day',
    schedule_interval => INTERVAL '1 hour');
ERROR:  bucket width is not defined for a variable bucket
```

This patch fixes it and adds corresponding tests.

Reported by Miranda Auhl <miranda@timescale.com>


mfundul (approved):
Some food for thought. Shouldn't we support `INTERVALOID` with custom `get_current_time()`-type functions similar to `get_integer_now()`?

The use-cases would be time-series data that are either historical or future prediction models.

akuzm (approved):

PR #4045, opened 2022-02-01
Custom origin's support in CAGGs

afiskon:
This patch allows using custom origin's in CAGGs, for instance:

time_bucket_ng('7 days', day, '2000-01-03' :: date) AS bucket

For weekly buckets this allows the user to choose what should be considered
the beginning of the week - Sunday or Monday. Also by shifting the origin
one second forward or backward user can tweak the inclusiveness of the
buckets.

This works for date's, timestamp's and timestamptz's. The bucket size is
considered variable-sized in all these cases. CAGGs on top of distributed
hypertables, compressed hypertables and compressed distributed hypertables
are supported as well.

Additionally, this patch does several small refactorings. Firstly, it makes
sure that experimental features of CAGGs will be tested in both Debug and
Release builds. This was previously overlooked. Secondly, it renames the tests
so that a person who is working on experimental features in CAGGs will be able
to easily execute all the related tests:  `TESTS='exp_cagg_*' make installcheck`

Last but not least the patch refactors is_valid_bucketing_function() and
renames it to function_allowed_in_cagg_definition(). The reason to do it in
this patch is that otherwise, the logic of the function gets rather confusing
which complicates code review.


afiskon:
data_node test is flaky https://github.com/timescale/timescaledb/issues/4047#issuecomment-1040012883

mfundul:
@afiskon can you also fix a comment that is obsolete?
`src/ts_catalog/catalog.h`:line 25
```
sql/common/tables.sql.
```
-->
```
sql/pre_install/tables.sql.
```

afiskon:
@mfundul done 👍

tsl/src/continuous_aggs/create.c
```diff
@@ -690,10 +693,16 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
 	bool found = false;
 
 	/*
-	 * Make sure bucket_width was initialized by caggtimebucketinfo_init().
-	 * This assumption is used below.
+	 * Make sure tbinfo is initialized. This assumption is used below.
+	 *
+	 * It's not guaranteed that caggtimebucketinfo_init() will be called before
```
mfundul, line 698: If it's not guaranteed, doesn't this mean that `tbinfo->origin` may be overwritten by `caggtimebucketinfo_init()` later?

afiskon: To be honest I'm a bit confused by the way we call caggtimebucketinfo_init(). Here is how cagg_validate_query() is implemented:

```
...
if (rte->relkind == RELKIND_RELATION) {
   ...
   caggtimebucketinfo_init(...)
   ...
}

...

caggtimebucket_validate(...)
```

So currently we can call validate() without init() but not wise versa. Also (1) the name "_init()" seems to be wrong, it just fills certain fields (2) I used Asserts() before I discovered the fact stated above and it seems that IN PRACTICE init() is always called before validate().

So, do you suggest we use Asserts() here or keep the assignment? Refactoring cagg_validate_query() is another option but I think we better do this in a separate pull request.

tsl/src/continuous_aggs/create.c
```diff
@@ -690,10 +693,16 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+	 * caggtimebucket_validate() - see cagg_validate_query() implementation.
+	 * For this reason we can't safaly use Assert() here. If you are looking
```
fabriziomello, line 700:
```suggestion
	 * For this reason we can't safely use Assert() here. If you are looking
```

src/ts_catalog/continuous_agg.h
```diff
@@ -65,8 +65,8 @@ typedef struct ContinuousAggsBucketFunction
 	char *name;
 	/* `bucket_width` argument of the function */
 	Interval *bucket_width;
-	/* `origin` argument of the function provided by the user */
-	char *origin;
+	/* `origin` argument of the function. If not specified, stores infinity */
+	Timestamp origin;
```
mfundul, line 69: Shouldn't `origin` be a `union` (or something else) and include `TimestampTz` as well? The origin is not always `Timestamp`. At the moment it's the same implementation, but that might change in the future and it's not very clear in the code that this might be a `TimestampTz` after all.

Also, what about `ts_time_bucket_ng_date()`? Don't they also have origin? Do we use those?

afiskon: Timestamp and TimestampTz are exactly the same in PostgreSQL, thus we cat get away with this. I am OK with using union's but I think this will make the code unnecessarily complicated and a bit slower (because we end up checking if timestamp or timestamptz is stored). Although I agree there should be a comment about this. I will add one.

DATEs can be converted to TIMESTAMPs and vice versa without any loss of data. I will add a comment about this as well.

tsl/src/continuous_aggs/create.c
afiskon: Thanks, fixed! 👍

tsl/src/continuous_aggs/create.c
```diff
@@ -1326,7 +1441,11 @@ is_valid_bucketing_function(Oid funcid)
 	is_timezone = (finfo->nargs == 3) && (finfo->arg_types[0] == INTERVALOID) &&
 				  (finfo->arg_types[1] == TIMESTAMPTZOID) && (finfo->arg_types[2] == TEXTOID);
 
-	return is_timescale && ((finfo->nargs == 2) || is_timezone);
+	is_custom_origin = ((finfo->nargs == 3) || (finfo->nargs == 4)) &&
+					   (finfo->arg_types[0] == INTERVALOID) &&
+					   (finfo->arg_types[1] == finfo->arg_types[2]);
```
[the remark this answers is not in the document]
afiskon: OK, this procedure is a bit silly. We already have `static FuncInfo funcinfo[]` with the exact schema, name, and arguments of the function. So we could just add a `bool is_valid_bucketing_function` to the FuncInfo structure and  `is_valid_bucketing_function()` would become just `return finfo->is_valid_bucketing_function;`. I will add this as an item to the Epic task for now, but will not touch this in this PR. I will submit a refactoring right after we merge this PR, it is a trivial change.

src/ts_catalog/continuous_agg.h
afiskon: Hopefully, it's better now. Feel free to re-open this discussion or create a new one if necessary.

tsl/src/continuous_aggs/create.c
afiskon: OK, after thinking a bit more about it I returned Assert()'s here as it is currently in `master`. I think the code is more clear this way. Please feel free to re-open the discussion if necessary.

tsl/src/continuous_aggs/create.c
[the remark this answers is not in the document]
afiskon: I'm marking this as resolved for now. Please feel free to re-open the thread if necessary.

src/ts_catalog/continuous_agg.c
```diff
@@ -1510,15 +1533,40 @@ generic_time_bucket_ng(const ContinuousAggsBucketFunction *bf, Datum timestamp)
...
+									   timestamp,
+									   CStringGetTextDatum(bf->timezone));
+		}
+		else
+		{
+			/* custom origin specified */
+			return DirectFunctionCall4(ts_time_bucket_ng_timezone_origin,
+									   IntervalPGetDatum(bf->bucket_width),
+									   timestamp,
+									   TimestampGetDatum(bf->origin),
```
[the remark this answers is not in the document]
fabriziomello: @mfundul Why casting if we can just do it?

```suggestion
									   TimestampTzGetDatum(bf->origin),
```

At the end will be the same `Int64GetDatum` behind the scenes but make sense since we're using here `TIMESTAMP WITH TIMEZONE`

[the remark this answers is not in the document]
mfundul: Because this is already a hack, and we need to remind ourselves that:
1. it's not a `Timestamp`
2. the function does not take a `TIMESTAMP` datum argument
3. we are relying on an implementation detail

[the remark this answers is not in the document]
afiskon: I have to agree with @fabriziomello - there is no need in this casting here. The fact that both TimestampTz and Timestamp are stored as uint64_t UTC time in microseconds is very well-known, it's been this way for years and is very unlikely to change. What we actually have here is not two types, but two macros with different names that do exactly the same. There is no reason to prefer one or another. I just used Timestamp everywhere for consistancy.

tsl/src/continuous_aggs/create.c
```diff
@@ -786,6 +871,37 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+					 * If a timezone was specified the check should be done in this timezone.
+					 */
+					Timestamp origin = tbinfo->origin;
+					if (tbinfo->timezone != NULL)
+					{
+						/* The code is equal to 'timestamptz AT TIME ZONE tzname'. */
+						origin = DatumGetTimestamp(
+							DirectFunctionCall2(timestamptz_zone,
+												CStringGetTextDatum(tbinfo->timezone),
+												TimestampGetDatum(origin)));
```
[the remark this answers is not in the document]
afiskon: We seem to have a consensus that the cast is not needed, please see https://github.com/timescale/timescaledb/pull/4045/files#r808797617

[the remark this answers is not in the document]
mfundul: I don't think we have a consensus yet. Moreover, Fabrizio also suggested to use `TimestampTzGetDatum()` instead of `TimestampGetDatum()`.

src/ts_catalog/continuous_agg.c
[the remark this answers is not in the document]
mfundul: We can discuss about the cast, but there is also the `TimestampTzGetDatum()` function that should be used.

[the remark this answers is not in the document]
afiskon: Oh, OK. @fabriziomello do you suggest using `TimestampTzGetDatum` or `TimestampGetDatum` here? Personally, I'm fine with any option 🤷‍♂️

tsl/src/continuous_aggs/create.c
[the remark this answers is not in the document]
afiskon: OK, OK. @fabriziomello same question - do you suggest using `TimestampTzGetDatum` or `TimestampGetDatum` here? Personally, I'm fine with any option.

src/ts_catalog/continuous_agg.h
```diff
@@ -65,8 +65,15 @@ typedef struct ContinuousAggsBucketFunction
...
-	char *origin;
+	/*
+	 * Custom origin value. If not specified, stores infinity.
+	 *
+	 * Actually, it can store Timestamp, TimestampTz or Date. Timestamp and
+	 * TimestampTz are the same in PostgreSQL so we can get away with this.
+	 * Date representation is different but it can be converted to Timestamp
+	 * with date_timestamp() without any loss of information.
+	 */
+	Timestamp origin;
```
mkindahl, line 76: It does not make sense to store `origin` as a `Timestamp` type when it can actually be any of three types and a `Datum` is more appropriate. PostgreSQL is using type-punning quite heavily but it offers the guarantee that you can convert a type (e.g., `Timestamp`) to a `Datum` and back, but it offers no guarantees that you can convert between types.  Use that guarantee instead to ensure that the conversion will always work, even if PostgreSQL makes changes. It will even properly handle conversions from PostgreSQL types to references and back, if that turns out to be the case for some platform.

```suggestion
	Datum origin;
```

tsl/src/continuous_aggs/create.c
mkindahl, line 891:
```suggestion
						origin = DirectFunctionCall2(timestamptz_zone,
												CStringGetTextDatum(tbinfo->timezone),
												origin);
```

tsl/src/continuous_aggs/create.c
```diff
@@ -786,6 +871,37 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+			if ((tbinfo->bucket_width == BUCKET_WIDTH_VARIABLE) && (tbinfo->interval->month != 0))
+			{
+				/* Monthly buckets case */
+				if (!TIMESTAMP_NOT_FINITE(tbinfo->origin))
+				{
+					/*
+					 * Origin was specified - make sure it's the first day of the month.
+					 * If a timezone was specified the check should be done in this timezone.
+					 */
+					Timestamp origin = tbinfo->origin;
```
mkindahl, line 884:
```suggestion
					Datum origin = tbinfo->origin;
```

tsl/src/continuous_aggs/create.c
```diff
@@ -786,6 +871,37 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+						/* The code is equal to 'timestamptz AT TIME ZONE tzname'. */
+						origin = DatumGetTimestamp(
+							DirectFunctionCall2(timestamptz_zone,
+												CStringGetTextDatum(tbinfo->timezone),
+												TimestampGetDatum(origin)));
+					}
+
+					const char *day =
+						TextDatumGetCString(DirectFunctionCall2(timestamp_to_char,
+																TimestampGetDatum(origin),
```
mkindahl, line 896:
```suggestion
																origin,
```

mkindahl (commented):
Here are some quick comments. I think it is dangerous to assume that implementation types will match and it is not sufficient that we can get away with it because they match right now. Since PostgreSQL offers the guarantee that you can convert to a `Datum` and back to the same type, I think you can store the origin as a `Datum` and convert as necessary. It also simplifies the code significantly.

src/ts_catalog/continuous_agg.c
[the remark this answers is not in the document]
afiskon: Interestingly, in PostgreSQL there is TIMESTAMP_NOT_FINITE macro but there is no matching TZ-version of it.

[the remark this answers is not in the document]
mfundul: They still use `TimestampTz` where appropriate even if `TIMESTAMP_NOT_FINITE()` works for both: https://github.com/postgres/postgres/blob/REL_14_2/src/backend/utils/adt/timestamp.c#L5622

src/ts_catalog/continuous_agg.h
afiskon: That's actually a very good point, thanks. I'll try to refactor the code accordingly 👍

afiskon: On second thought, I just confused everyone with a wrong comment. What origin **actually** stores is always Timestamp in UTC, regardless of whether a CAGG works with Dates, Timestamp, or TimestampTZs. The actual problem is that there are places where I treated it as TimestampTz. I will update the comment and refactor the code according to @mfundul and @fabriziomello notes.

src/ts_catalog/continuous_agg.c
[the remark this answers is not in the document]
afiskon: After thinking a bit more about it I believe @mfundul is right. Here we should cast Timestamp to TimestampTz and use TimestampTzGetDatum(), because ts_time_bucket_ng_timezone_origin expects TimestampTz. I will update the pull request accordingly.

tsl/src/continuous_aggs/create.c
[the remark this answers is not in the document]
afiskon: After thinking about it I realized @mfundul is right - origin stores Timestamp but timestamptz_zone expects TimestampTz, so we better explicitly cast Timestamp to TimestampTz and use TimestampTzGetDatum(). I will update the pull request accordingly.

tsl/src/continuous_aggs/create.c
```diff
@@ -182,6 +182,16 @@ typedef struct CAggTimebucketInfo
...
+
+	/*
+	 * Custom origin value. If not specified, stores infinity.
+	 *
+	 * Actually, it can store Timestamp, TimestampTz or Date. Timestamp and
+	 * TimestampTz are the same in PostgreSQL so we can get away with this.
+	 * Date representation is different but it can be converted to Timestamp
+	 * with date_timestamp() without any loss of information.
+	 */
+	Timestamp origin;
```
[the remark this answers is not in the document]
afiskon: See https://github.com/timescale/timescaledb/pull/4045#discussion_r808946294

tsl/src/continuous_aggs/create.c
```diff
@@ -756,6 +768,79 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+					}
+
+					switch (tz->consttype)
+					{
+						case DATEOID:
+							tbinfo->origin = DatumGetTimestamp(
+								DirectFunctionCall1(date_timestamp, tz->constvalue));
+							break;
+						case TIMESTAMPOID:
+							tbinfo->origin = DatumGetTimestamp(tz->constvalue);
```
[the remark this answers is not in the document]
afiskon: See https://github.com/timescale/timescaledb/pull/4045#discussion_r808946294

tsl/src/continuous_aggs/create.c
```diff
@@ -756,6 +768,79 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+						ereport(ERROR,
+								(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+								 errmsg("invalid origin value: null")));
+					}
+
+					switch (tz->consttype)
+					{
+						case DATEOID:
+							tbinfo->origin = DatumGetTimestamp(
+								DirectFunctionCall1(date_timestamp, tz->constvalue));
```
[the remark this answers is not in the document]
afiskon: See https://github.com/timescale/timescaledb/pull/4045#discussion_r808946294

tsl/src/continuous_aggs/create.c
```diff
@@ -756,6 +768,79 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+							break;
+						case TIMESTAMPOID:
+							tbinfo->origin = DatumGetTimestamp(tz->constvalue);
+							break;
+						case TIMESTAMPTZOID:
+							tbinfo->origin = DatumGetTimestampTz(tz->constvalue);
```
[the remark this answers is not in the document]
afiskon: See https://github.com/timescale/timescaledb/pull/4045#discussion_r808946294

tsl/src/continuous_aggs/create.c
afiskon: See https://github.com/timescale/timescaledb/pull/4045#discussion_r808946294

tsl/src/continuous_aggs/create.c
afiskon: See https://github.com/timescale/timescaledb/pull/4045#discussion_r808946294

tsl/src/continuous_aggs/create.c
afiskon: See https://github.com/timescale/timescaledb/pull/4045#discussion_r808946294

src/ts_catalog/continuous_agg.h
mfundul: I am still confused about this. I've been looking at this definition:
```
CREATE FUNCTION timescaledb_experimental.time_bucket_ng(bucket_width INTERVAL, ts TIMESTAMPTZ, origin TIMESTAMPTZ, timezone TEXT) RETURNS TIMESTAMPTZ
    AS '@MODULE_PATHNAME@', 'ts_time_bucket_ng_timezone_origin' LANGUAGE C IMMUTABLE PARALLEL SAFE STRICT;
```
In this case don't you have to push down to the data nodes the `TIMESTAMPTZ` origin instead of a `TIMESTAMP` origin? I imagine the data node will execute `ts_time_bucket_ng_timezone_origin()` with a `TIMESTAMPTZ` origin argument.

afiskon: Internally `origin` is always stored as Timestamp in UTC (Note that the previous comment I added to the code was wrong, check out the new comment). Dates and TimestampTz can both be converted to Timestamp in UTC without any data loss. The timezone is stored separately. So between data nodes `origin` can be serialized/deserialized using simply timestamp_in / timestamp_out.  See bucket_function_serialize() https://github.com/timescale/timescaledb/pull/4045/files#diff-559a60cc746cb2fe273d872b89dc6f398ae14c3a9775f38a46406fedac0aa930R425

tsl/src/continuous_aggs/create.c
```diff
@@ -756,6 +764,79 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+					switch (tz->consttype)
+					{
+						case DATEOID:
+							tbinfo->origin = DatumGetTimestamp(
+								DirectFunctionCall1(date_timestamp, tz->constvalue));
+							break;
+						case TIMESTAMPOID:
+							tbinfo->origin = DatumGetTimestamp(tz->constvalue);
+							break;
+						case TIMESTAMPTZOID:
```
mfundul, line 817: It could be the case that 3rd arg is `TIMESTAMPOID` and 4th arg is `TEXTOID`. This would be allowed by `is_valid_bucketing_function()` as long as the 2nd argument is also `TIMESTAMPOID`. But this `time_bucket_ng` overloaded function is not defined and does not exist.

afiskon: No, actually it will not be allowed. `is_valid_bucketing_function()` takes allowed functions from the func cache, and there is no function with such signature there. This is ensured by `initialize_func_info()` which places only functions listed in `FuncInfo funcinfo[]` there. Basically, is_valid_bucketing_function simply calculates a boolean value that could be stored in funcinfo[] instead.

But I agree all this may be pretty confusing. I think I will just include a refactoring we discussed previously to this PR.

tsl/src/continuous_aggs/create.c
```diff
@@ -786,6 +867,37 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
...
+					}
+
+					const char *day =
+						TextDatumGetCString(DirectFunctionCall2(timestamp_to_char,
+																TimestampGetDatum(origin),
+																CStringGetTextDatum("DD")));
+					if (strcmp(day, "01") != 0)
+						ereport(ERROR,
+								(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+								 errmsg("for monthly buckets origin must be the first day of the "
```
mfundul, line 897: Why is that?

mfundul (approved):

tsl/src/continuous_aggs/create.c
afiskon: For instance, if you specify origin = '2022-01-29' adding 1 month will result in '2022-02-29', but there is no such date. It's unclear how to handle such cases, thus currently we explicitly forbid doing this. See also the tests for time_bucket_ng().

tsl/src/continuous_aggs/create.c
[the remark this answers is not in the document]
afiskon: I refactored the function in the updated version of the PR 👍

tsl/src/continuous_aggs/create.c
afiskon: I refactored the function in the updated version of the PR 👍

tsl/src/continuous_aggs/create.c
mfundul: Ok I guess, but I wonder if users expect to do every 15th of the month with origin for example.

afiskon: Sure. We just didn't want to implement this in MVP thus currently it's not supported by time_bucket_ng() 🤷‍♂️ If there will be a demand from the users it will be not difficult to deliver.

fabriziomello (approved):

PR #4051, opened 2022-02-03
Improve telemetry with stats broken down into relation kind

### LINKED ISSUE #3932: [Enhancement]: Extend telemetry to report table categories and deeper stats

Other

Telemetry

Break down telemetry into the different kinds of relations:

* Regular tables
* Hypertables 
   - compressed
   - uncompressed
* Distributed hypertables
   - compressed
   - uncompressed
* Chunks
   - compressed 
   - uncompressed
* Materialized views
* Continuous aggregates

And, for each of these categories, report

* Approximate number of rows
* Data size
* Data size if compression would not have been enabled (pre-compression size)

### Implementation challenges

All the stats should be quick to gather, or otherwise we shouldn't include them.

erimatnor:
Refactor the telemetry function and format to include stats broken
down on common relation types. The types include:

- Tables
- Partitioned tables
- Hypertables
- Distributed hypertables
- Continuous aggregates
- Materialized views
- Views

and for each of these types report (when applicable):

- Total number of relations
- Total number of children/chunks
- Total data volume (broken into heap, toast, and indexes).
- Compression stats
- PG stats, like reltuples

The telemetry function has also been refactored to return `jsonb`
instead of `text`. This makes it easier to query and manipulate the
resulting JSON format, and also gives cleaner output.

Closes #3932

Disable-Check: Commit-Count


erimatnor:
Still need to add more tests. There's also a problem with an old test that runs out of background workers for some reason.

src/utils.c
```diff
@@ -903,3 +903,30 @@ ts_subtract_integer_from_now(PG_FUNCTION_ARGS)
...
+RelationSize
+ts_relation_size(Oid relid)
+{
+	int64 tot_size;
+	int i = 0;
+	RelationSize ret;
+	Datum oid = ObjectIdGetDatum(relid);
+	static const char *filtyp[] = { "main", "init", "fsm", "vm" };
+
+	/* for heap get size from fsm, vm, init and main as this is included in
```
nikkhils, line 916: Nit picking: Most `multi-line` comments are starting on the same line as `\*` whereas they should start on the next line.

src/telemetry/telemetry.c
```diff
@@ -581,26 +674,12 @@ ts_telemetry_main(const char *host, const char *path, const char *service)
 	return false;
 }
 
-TS_FUNCTION_INFO_V1(ts_get_telemetry_report);
```
nikkhils, line 584: isn't this function `ts_get_telemetry_report ` outward visible? Should we be changing it?

nikkhils (approved):
Overall LGTM. Have one comment and one nit-pick.

pmwkaa (approved):
Looks good to me.

I am thinking maybe it makes sense to move the classify logic somewhere outside of the telemetry code, for instance we have a `HypertableType` already, I guess we could also want to have `ChunkType` as well. For the sake of time I do not propose to do it now, maybe we want to do it in the future.

src/telemetry/stats.c
```diff
@@ -0,0 +1,579 @@
...
+			}
+			default:
+				Assert(ht->fd.replication_factor >= 1);
+				return RELTYPE_DISTRIBUTED_HYPERTABLE;
+		}
+	}
+}
+
+static StatsRelType
+classify_chunk(Cache *htcache, const Hypertable **ht, const Chunk *chunk)
```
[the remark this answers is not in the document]
erimatnor: Good point. We do classify things in the planner (`src/planner.c`). That code is not as general as this one, but it might be worth looking into how to unify those code paths into something more general. I think we can leave that for follow-up work though.

src/telemetry/stats.c
```diff
@@ -0,0 +1,579 @@
...
+	TelemetryStats *stats;
+	ScanIterator compressed_chunk_stats_iterator;
+	bool iterator_valid;
+} StatsContext;
+
+/*
+ * Determine the type of a hypertable.
+ */
+static StatsRelType
+classify_hypertable(const Hypertable *ht)
```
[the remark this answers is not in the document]
erimatnor: HyperTable type is used in this function. `ts_hypertable_get_type()`  doesn't work in this case because it only returns a subset of the types we use here. We might be able to unify this in the future into something more general.

src/telemetry/telemetry.c
```diff
@@ -307,20 +254,134 @@ format_iso8601(Datum value)
...
+#define REQ_RELKIND_COMPRESSED_ROWCOUNT "compressed_row_count"
+
+static JsonbValue *
+add_compression_stats_object(JsonbParseState *parse_state, StatsRelType reltype,
+							 const HyperStats *hs)
+{
+	JsonbValue name = {
+		.type = jbvString,
+		.val.string.val = pstrdup("compression"),
+		.val.string.len = strlen("compression"),
```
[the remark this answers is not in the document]
erimatnor: Not sure I understand what you are trying to say. What do you mean with "wonder if compiler will understand this call"?

src/utils.c
erimatnor: I just moved this code from another file, but will update the comment.

erimatnor: We should try to update the clang formatter to ensure multi-line comments are formatted correctly.

Issue #4043, opened 2022-02-01
Don't allow to inherit from hypertables


akuzm:
We don't support it in any meaningful way, should throw an error when creating a table like this.

An example of poorly defined behavior:

```
test=# create table t_hyper(ts timestamp, value text);
test=# select create_hypertable('t_hyper', 'ts');
test=# create table t_child() inherits (t_hyper);

test=# insert into t_child values ('2021-01-01 00:00:00', 'inheritance child');
test=# insert into t_hyper values ('2021-01-01 00:00:01', 'hypertable');


test=# with t as materialized (select * from t_hyper) select * from t where ts >= '2021-01-01';
         ts          │       value       
─────────────────────┼───────────────────
 2021-01-01 00:00:00 │ inheritance child
 2021-01-01 00:00:01 │ hypertable
(2 rows)

test=# with t as (select * from t_hyper) select * from t where ts >= '2021-01-01';
         ts          │   value    
─────────────────────┼────────────
 2021-01-01 00:00:01 │ hypertable
(1 row)
```

PR #4052, opened 2022-02-03
Cache chunk data when performing chunk exclusion

akuzm:
We cache the Chunk structs in RelOptInfo private data. They are later
used to estimate the chunk sizes, check which data nodes they belong
to, et cetera. Looking up the chunks is expensive, so this change
speeds up the planning.

### DIFF
--- a/src/plan_expand_hypertable.c
+++ b/src/plan_expand_hypertable.c
@@ -28,18 +28,18 @@
 #include "chunk.h"
 #include "compat/compat.h"
 #include "cross_module_fn.h"
-#include "guc.h"
 #include "extension.h"
 #include "extension_constants.h"
+#include "guc.h"
 #include "hypertable.h"
 #include "hypertable_restrict_info.h"
+#include "import/planner.h"
+#include "nodes/chunk_append/chunk_append.h"
 #include "partitioning.h"
 #include "plan_expand_hypertable.h"
 #include "plan_partialize.h"
 #include "planner.h"
 #include "time_utils.h"
-#include "nodes/chunk_append/chunk_append.h"
-#include "import/planner.h"
 
 typedef struct CollectQualCtx
 {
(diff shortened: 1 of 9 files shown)


src/plan_expand_hypertable.c
```diff
@@ -846,23 +846,59 @@ collect_quals_walker(Node *node, CollectQualCtx *ctx)
...
+find_children_chunks(HypertableRestrictInfo *hri, Hypertable *ht, LOCKMODE lockmode,
+					 unsigned int *num_chunks)
+{
+	if (TS_HYPERTABLE_IS_INTERNAL_COMPRESSION_TABLE(ht))
+	{
+		/*
+		 * FIXME somehow chunk lookup doesn't work for internal compression
+		 * tables
+		 */
+		List *chunk_oids = find_inheritance_children(ht->main_table_relid, lockmode);
```
akuzm, line 865: Not sure why the normal chunk lookup e.g. `ts_chunk_find_all` doesn't work in this case. I'll have to ask @gayyappan or other compression experts.

test/sql/include/join_query.sql
```diff
@@ -487,14 +487,6 @@ SELECT * FROM t3;
...
 
--- Test join against inheritance tree
-
-create temp table t2a () inherits (t2);
-
-insert into t2a values (200, 2001);
-
-select * from t1 left join t2 on (t1.a = t2.a);
-
 -- Test matching of column name with wrong alias
```
akuzm, line 490: This doesn't work in any meaningful way, so the test is removed. See https://github.com/timescale/timescaledb/issues/4043

tsl/test/expected/compression_errors.out
```diff
@@ -494,25 +494,3 @@ SELECT config FROM _timescaledb_config.bgw_job WHERE id = :compressjob_id;
...
-(0 rows)
-
-CREATE TABLE extras (more_magic bool) INHERITS (i165);
-INSERT INTO i165 (time) VALUES
-       (generate_series(TIMESTAMP '2019-08-01', TIMESTAMP '2019-08-10', INTERVAL '10 minutes'));
-\set VERBOSITY default
-SELECT * FROM i165;
-ERROR:  chunk not found
-DETAIL:  schema_name: public, table_name: extras
-\set VERBOSITY terse
```
akuzm, line 518: We should disable inheritance from hypertables, no matter if they are compressed or not. See https://github.com/timescale/timescaledb/issues/4043
I am removing this test, because the error it produced was an artifact of some planning optimization. We should have a more principled check instead.

afiskon (approved):

src/plan_expand_hypertable.c
```diff
@@ -1198,14 +1234,14 @@ ts_plan_expand_timebucket_annotate(PlannerInfo *root, RelOptInfo *rel)
...
- * a hypertable chunks into an append relationship */
+ * a hypertable chunks into an append relation. */
 void
 ts_plan_expand_hypertable_chunks(Hypertable *ht, PlannerInfo *root, RelOptInfo *rel)
 {
 	TimescaleDBPrivate *priv = rel->fdw_private;
 	RangeTblEntry *rte = rt_fetch(rel->relid, root->parse->rtable);
 	Oid parent_oid = rte->relid;
-	List *inh_oids;
+	List *inh_oids = NULL;
```
pmwkaa, line 1244: nit: I guess it should be `NIL`

src/plan_expand_hypertable.c
```diff
@@ -846,23 +846,59 @@ collect_quals_walker(Node *node, CollectQualCtx *ctx)
...
+		if (chunk_oids == NIL)
+		{
+			*num_chunks = 0;
+			return 0;
+		}
+
+		*num_chunks = list_length(chunk_oids);
+		Chunk **chunks = (Chunk **) palloc(sizeof(Chunk *) * *num_chunks);
 
+		for (int i = 0; i < *num_chunks; i++)
```
pmwkaa, line 875: nit: not sure if we are allowed to use C99 type definition inside `for` statement :)

pmwkaa (approved):
Looks good to me, sorry for the late review. Only nitpicking, you can totally ignore them

src/plan_expand_hypertable.c
akuzm: Let's use it, it looks better. I think even Postgres allows it now.

src/plan_expand_hypertable.c
akuzm:
```suggestion
	List *inh_oids = NIL;
```

akuzm:
```
....
dist_grant-13                ... FAILED    12175 ms

...
+ERROR:  relation "public." does not exist
```

Ehhh what's that... seems to be flaky.

PR #4064, opened 2022-02-07
time_bucket_ng() bugfix

afiskon:
time_bucket_ng() may return wrong results for monthly buckets with overwritten
origin. As an example:

time_bucket_ng('1 month', '2001-01-01' :: date, origin => '2000-06-01')

... returns 2000-01-01 instead of 2001-01-01. In fact, our tests showed this
but we overlooked the fact that the results are incorrect. The reason is that
the month of the origin is not accounted for when calculating the resulting
year.

This patch fixes this.

Also it updates linux-32bit-build-and-test.yaml workflow due to todays
14.2/13.6/12.10 releases. "Memoize" test doesn't pass on 14.2 thus the patch
adds it to the list of skipped tests. The patch doesn't rename the checks
because we list "PG14.1 Debug linux-i386" etc as required checks on GitHub.
This will be addressed in the follow-up patches.


mkindahl (approved):
LGTM

svenklemm (approved):

PR #4069, opened 2022-02-09
Fix riinfo NULL handling in ANY construct

nikkhils:
If the ANY construct contains a singleton NULL then the logic in
"dimension_values_create_from_array" barfs causing a crash. Fix it
appropriately in the caller "hypertable_restrict_info_add_expr"
function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,12 +20,14 @@ accidentally triggering the load of a previous DB version.**
 * #4015 Eliminate float rounding instabilities in interpolate
 * #4020 Fix ALTER TABLE EventTrigger initialization
 * #4024 Fix premature cache release call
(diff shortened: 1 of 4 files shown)


svenklemm (approved):

svenklemm:
Dont forget CHANGELOG entry

nikkhils:
> Dont forget CHANGELOG entry

Done!

mkindahl (changes requested):
We do not reference support issues in the commit message or pull requests since they are internal. You can add a reference to this issue from the support issue instead.

nikkhils:
> We do not reference support issues in the commit message or pull requests since they are internal. You can add a reference to this issue from the support issue instead.

Ok, removed the reference

mkindahl (approved):

PR #4075, opened 2022-02-09
Fix openssl detection for PG14 on windows

svenklemm:
For some reason the configure parameter for openssl on windows is
--with-ssl=openssl instead of --with-openssl on PG14. This patch adjusts
the check to detect both versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -568,7 +568,8 @@ execute_process(
   COMMAND ${PG_CONFIG} --configure
   OUTPUT_VARIABLE PG_CONFIGURE_FLAGS
   OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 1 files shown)


svenklemm:
This issue pops up every release on our windows vm where we build the windows packages.

fabriziomello:
@svenklemm this commit https://github.com/postgres/postgres/commit/fe61df7f82 introduce this configure option for PG14. I'm wonder because for previous versions if we build Postgres we'll get those warnings during the configure:

```bash
13 on  REL_13_STABLE 
➜ ./configure --with-ssl=openssl
configure: WARNING: unrecognized options: --with-ssl
...
configure: WARNING: unrecognized options: --with-ssl
```

gayyappan (approved):

PR #4076, opened 2022-02-10
Fix memcpy in force_group_by_push_down

mfundul:


mkindahl (approved):
They should be the same anyway, so not really necessary, but it does look more consistent.

mfundul:
> They should be the same anyway, so not really necessary, but it does look more consistent.

It's actually a heap buffer overflow, so this fix is needed.

PR #4081, opened 2022-02-10
Add log message for retention policy

konskov:


tsl/src/bgw_policy/job.c
```diff
@@ -250,9 +250,50 @@ bool
...
+			break;
+		case DATEOID:
+			elog(LOG,
+				 "%s: \"%s\" dropping data older than %s",
+				 message,
+				 relname,
+				 DatumGetCString(DirectFunctionCall1(date_out, boundary)));
+			break;
+		default:
+			/* this should never happen as otherwise hypertable has unsupported time type */
```
mfundul, line 293: If it's truely impossible, you should add an `Assert()` here. Otherwise, you could add a print or throw an exception.

tsl/src/bgw_policy/job.c
```diff
@@ -250,9 +250,50 @@ bool
...
+	if (IS_INTEGER_TYPE(policy_data.boundary_type))
+	{
+		elog(LOG,
+			 "%s: %s dropping data older than %s",
+			 message,
+			 relname,
+			 DatumGetCString(DirectFunctionCall1(int8out, boundary)));
+	}
+	switch (policy_data.boundary_type)
+	{
```
mfundul, line 270: This whole `if`, `switch()` block should be in a separate function that returns a boundary string. Also, the `if` and the `switch` should not both be executed. Maybe add an `else` to the `switch`.

mfundul:
You should probably use something like:
```
	elog(elevel,
		 "%s \"%s\" in window [ %s, %s ]",
		 msg,
		 NameStr(cagg->data.user_view_name),
		 DatumGetCString(OidFunctionCall1(outfuncid, start_ts)),
		 DatumGetCString(OidFunctionCall1(outfuncid, end_ts)));
```
from `tsl/src/continuous_aggs/refresh.c` for the prints.

konskov:
Okay! What kind of window makes sense for a retention policy?
Perhaps a message like` “Retaining data in window [boundary, now]"` is the equivalent for the retention policy? 
Where now is the time we get from `ts_time_get_end_or_max`

tsl/src/bgw_policy/job.c
konskov: How about an `ereport` with `ERRCODE_INVALID_PARAMETER_VALUE` ?

mfundul:
> Okay! What kind of window makes sense for a retention policy?
> Perhaps a message like` “Retaining data in window [boundary, now]"` is the equivalent for the retention policy?
> Where now is the time we get from `ts_time_get_end_or_max`

I didn't mean to print a window. Just that you could avoid using `if`,`else`,`switch` by using `outfuncid` in similar fashion. You could take a look at `refresh.c` to get some ideas.

tsl/src/bgw_policy/job.c
```diff
@@ -253,6 +276,8 @@ policy_retention_execute(int32 job_id, Jsonb *config)
 
 	policy_retention_read_and_validate_config(config, &policy_data);
 
+	log_retention_window(LOG, &policy_data, "applying retention policy to hypertable");
```
svenklemm, line 279: this needs to be conditional and only happen when the job is set to verbose

svenklemm: you wanna do something similiar like we have for compression job:
tsl/src/bgw_policy/compression_api.c
look for CONFIG_KEY_VERBOSE_LOG

fabriziomello: @svenklemm interesting, seems the function [`policy_compression_get_verbose_log`](https://github.com/timescale/timescaledb/blob/master/tsl/src/bgw_policy/compression_api.c#L69) is dead code because is not used anywhere

fabriziomello: Seems we left it behind when @gayyappan change the policy_compression_execute to be a plpgsql procedure in this commit fffd6c2350f5b3237486f3d49d7167105e72a55b

PR #4099, opened 2022-02-15
Remove unused code from compression_api

konskov:
Function policy_compression_get_verbose_log was still
defined in compression_api.c but never called anywhere in the code.
Also update test compression_bgw

### DIFF
--- a/tsl/src/bgw_policy/compression_api.c
+++ b/tsl/src/bgw_policy/compression_api.c
@@ -65,15 +65,6 @@ policy_compression_get_maxchunks_per_job(const Jsonb *config)
 	return (found && maxchunks > 0) ? maxchunks : 0;
 }
 
-bool
-policy_compression_get_verbose_log(const Jsonb *config)
-{
-	bool found;
-	bool verbose_log = ts_jsonb_get_bool_field(config, CONFIG_KEY_VERBOSE_LOG, &found);
-
-	return found ? verbose_log : false;
-}
-
 int32
 policy_compression_get_hypertable_id(const Jsonb *config)
 {
--- a/tsl/src/bgw_policy/job.h
+++ b/tsl/src/bgw_policy/job.h
@@ -16,7 +16,6 @@
 #include "bgw_policy/chunk_stats.h"
 
 /* Add config keys common across job types here */
-#define CONFIG_KEY_VERBOSE_LOG "verbose_log" /*used only by compression now*/
 
 typedef struct PolicyReorderData
 {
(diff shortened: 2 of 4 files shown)


svenklemm (changes requested):
When verbose_log was introduced we decided that eventually every job should use this mechanism, not sure why you are trying to remove it and what the alternative to achieve something similar is.

gayyappan (commented):
It is not clear why this code is being removed.
Ideally, all jobs would follow a similar approach to enable/disable verbose logging.

This approach solves the following problems:
Users often want to add logging to a specific job for debugging. If you output messages at DEBUG level, then you need to enable it at the server level, otherwise they will not get written to pg logs. 
If you enable it at LOG level, then it is too verbose during normal operations. 
So you need a way to conditionally turn on logging for jobs and write with error level LOG.

The alter_job api provides a convenient way of doing this at a per job level, as needed.

mkindahl (approved):

gayyappan (commented):
ok, i see the config is parsed by the .sql file, so the C code is no longer used.
`111   verbose_log         := COALESCE(jsonb_object_field_text(config, 'verbo    se_log')::BOOLEAN, FALSE);`

tsl/test/sql/compression_bgw.sql
```diff
@@ -205,11 +205,7 @@ SELECT add_compression_policy AS job_id
 -- job compresses only 1 chunk at a time --
 SELECT alter_job(id,config:=jsonb_set(config,'{maxchunks_to_compress}', '1'))
  FROM _timescaledb_config.bgw_job WHERE id = :job_id;
-SELECT alter_job(id,config:=jsonb_set(config,'{verbose_log}', 'true'))
- FROM _timescaledb_config.bgw_job WHERE id = :job_id;
-set client_min_messages TO LOG;
 CALL run_job(:job_id);
```
gayyappan, line 208: We want to verify that the message is still logged when verbose_log is set in the path. The changes to .sql and .out files are not needed. Please remove them.

tsl/src/bgw_policy/compression_api.c
```diff
@@ -65,15 +65,6 @@ policy_compression_get_maxchunks_per_job(const Jsonb *config)
 	return (found && maxchunks > 0) ? maxchunks : 0;
 }
 
-bool
-policy_compression_get_verbose_log(const Jsonb *config)
-{
```
gayyappan, line 70: Since you are adding this change (i.e. verbose_log in config ) for retention policy etc, it makes more sense to move this function to utils.c, instead of removing this code.
Rename the function to something more generic: policy_get_verbose_log when you move it to utils.c

tsl/src/bgw_policy/job.h
```diff
@@ -16,7 +16,6 @@
 #include "bgw_policy/chunk_stats.h"
 
 /* Add config keys common across job types here */
-#define CONFIG_KEY_VERBOSE_LOG "verbose_log" /*used only by compression now*/
```
gayyappan, line 19: Don't remove this. Use this for the retention policy logging PR.

gayyappan (commented):
Suggestions on how to move this code for reuse by other policies.
.sql file and .out changes should be reverted.

PR #4102, opened 2022-02-15
Set search_path again after COMMIT

svenklemm:
SET LOCAL is only active until end of transaction so we set search_path
again after COMMIT in functions that do transaction control. While we
could use SET at the start of the function we do not want to bleed out
search_path to caller.

### DIFF
--- a/sql/policy_internal.sql
+++ b/sql/policy_internal.sql
@@ -64,9 +64,18 @@ BEGIN
     ELSIF chunk_rec.status = 3 AND recompress_enabled IS TRUE THEN
        PERFORM @extschema@.decompress_chunk(chunk_rec.oid, if_compressed => true);
        COMMIT;
+       -- SET LOCAL is only active until end of transaction.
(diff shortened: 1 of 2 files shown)


sql/policy_internal.sql
```diff
@@ -67,6 +67,11 @@ BEGIN
        PERFORM @extschema@.compress_chunk(chunk_rec.oid);
     END IF;
     COMMIT;
+    -- SET LOCAL is only active until end of transaction.
+    -- While we could use SET at the start of the function we do not
+    -- want to bleed out search_path to caller, so we do SET LOCAL
+    -- again after COMMIT
+    SET LOCAL search_path TO pg_catalog;
```
gayyappan, line 74: Don't you need  to set this at line 67 (before compress_chunk) too?

gayyappan (approved):

fabriziomello (approved):

PR #4108, opened 2022-02-17
Post release 2.6.0

akuzm:


svenklemm (changes requested):
Your changes to the test_updates_pg files are not correct. You should append them to the v8 list instead of starting v9 list

svenklemm (approved):

fabriziomello (approved):

PR #4111, opened 2022-02-18
Improve chunk scan performance

### LINKED ISSUE #2173: Allow Scanner to do scans on already open relations
The Scanner currently opens a relation, starts a scan, ends the scan, and the closes the relation. This is convenient, but not very flexible.

Sometimes it make sense to keep an open relation and do multiple scans or operations on it before closing it. Therefore it would make sense to allow the scanner to work on already open relation, i.e., only do start and end scan, not open and close the relation.

This would allow code to keep relations open for multiple processing steps, and then only close when done with it.

erimatnor:
[](https://github.com/timescale/timescaledb/commits?author=erimatnor)Chunk scan performance during querying is improved by avoiding
repeated open and close of relations and indexes when joining chunk
information from different metadata tables.

When executing a query on a hypertable, it is expanded to include all
its children chunks. However, during the expansion, the chunks that
don't match the query constraints should also be excluded. The
following changes are made to make the scanning and exclusion more
efficient:

* Ensure metadata relations and indexes are only opened once even
  though metadata for multiple chunks are scanned. This avoids doing
  repeated open and close of tables and indexes for each chunk
  scanned.
* Avoid interleaving scans of different relations, ensuring better
  data locality, and having, e.g., indexes warm in cache.
* Avoid unnecessary scans that repeat work already done.
* Ensure chunks are locked in a consistent order (based on Oid).

To enable the above changes, some refactoring was necessary. The chunk
scans that happen during constraint exclusion are moved into separate
source files (`chunk_scan.c`) for better structure and readability.

Some test outputs are affected due to the new ordering of chunks in
append relations.

The scanner interface has also been updated  (separate commits) to
allow more control over how relations are opened and closed during
scans in order to avoid unnecessary open and close operations
when, e.g., scanning for multiple chunks at once.

Closes #2173

Disable-Check: Commit-Count


test/expected/ddl-13.out
```diff
@@ -487,10 +487,10 @@ SELECT * FROM alter_test WHERE time_us > '2017-05-20T10:00:01';
                                      QUERY PLAN                                      
 -------------------------------------------------------------------------------------
  Append
-   ->  Seq Scan on _hyper_9_4_chunk
-         Filter: (time_us > 'Sat May 20 10:00:01 2017'::timestamp without time zone)
    ->  Seq Scan on _hyper_9_5_chunk
          Filter: (time_us > 'Sat May 20 10:00:01 2017'::timestamp without time zone)
+   ->  Seq Scan on _hyper_9_4_chunk
```
akuzm, line 492: I see a lot of changes like this where the plan branches change places. I was wondering what's the reason?

erimatnor: Yeah, so the order of the children in an append is governed by the order in which chunks are discovered during the parent expansion. Since we change how we scan for the chunks that should be included, the order also changed. We could sort the chunks according to, e.g., relation ID or similar to always have the same order, but that is an unnecessary sorting step that will only slow things down.

akuzm:
The performance in release mode looks good, 3.84 -> 2.69 ms, 1.42 times faster. Test query:
```
create table t(ts timestamp, name text, value numeric);

select create_distributed_hypertable('t', 'ts', partitioning_column => 'name');

with dummy as (select '2021-01-01'::timestamp as start),
	times as (
		select start + generate_series(0, 60 * 24 * 365) * (interval '1 minute') ts
		from dummy
		order by random()),
	series(name, period1, period2) as (values
		('green', interval '1 week', interval '2 month'),
		('blue', interval '4 month', interval '1 hour'),
		('red', interval '1 year', interval '3 week'),
		('yellow', interval '3 day', interval '5 month'),
		('black', interval '7 hour', interval '11 week'))
insert into t
select ts, name, 10 * sin(2 * pi() * extract(epoch from (ts - start)) / extract(epoch from period1))
	+ 5 * sin(2 * pi() * extract(epoch from (ts - start)) / extract(epoch from period2)) as value
from times, series, dummy

-- benchmark:
explain select distinct name from t where ts > '2021-06-01';
```

akuzm:
I'd expect an even more significant setup from the next part -- move the constraints & data node lookup out of `chunk_tuple_found` into separate loops over all chunks, first for constraints, then for data nodes. This will greatly improve the data access locality, and speed it up another couple of times.

erimatnor:
> I'd expect an even more significant setup from the next part -- move the constraints & data node lookup out of `chunk_tuple_found` into separate loops over all chunks, first for constraints, then for data nodes. This will greatly improve the data access locality, and speed it up another couple of times.

@akuzm Moving the constraints and data node lookups out of `chunk_tuple_found` is what this PR does for the scan that happens during hypertable expansion. The `chunk_tuple_found` function is used for other scans as well, but they are not on the same critical path. Still, I hope to refactor those scans later as well.

erimatnor:
Updated the PR with the following:

* Ensure chunks returned by hypertable expansion are always in chunk OID order in order to avoid deadlocks when locking the chunks. This is to be consistent with standard PG table expansion, which we sometimes fall back to.

src/chunk.c
```diff
@@ -4267,3 +4118,66 @@ ts_chunk_status(PG_FUNCTION_ARGS)
...
+ * Lock the chunk if the lockmode demands it.
+ *
+ * Also check that the chunk relation actually exists after the lock is
+ * acquired. Return true if no locking is necessary or the chunk relation
+ * exists and the lock was successfully acquired. Otherwise return false.
+ */
+bool
+ts_chunk_lock_if_exists(Oid chunk_oid, LOCKMODE chunk_lockmode)
+{
+	/* No lock is requests, so assume relation exists */
```
nikkhils, line 4132: requests -> requested

tsl/test/shared/expected/ordered_append_join-12.out
```diff
@@ -1733,8 +1733,8 @@ FROM :TEST_TABLE o1
 WHERE o1.device_id = 1
 ORDER BY o1.time
 LIMIT 100;
-                                                                     QUERY PLAN                                                                     
-----------------------------------------------------------------------------------------------------------------------------------------------------
```
nikkhils, line 1737: Interesting change this.. if the output hasn't changed..

akuzm:
> @akuzm Moving the constraints and data node lookups out of `chunk_tuple_found` is what this PR does for the scan that happens during hypertable expansion.

Cool, I see this now.

src/scanner.c
```diff
@@ -231,73 +214,141 @@ ts_scanner_start_scan(ScannerCtx *ctx, InternalScannerCtx *ictx)
...
+	Scanner *scanner;
+	TupleDesc tuple_desc;
+
+	if (ictx->started)
+	{
+		Assert(!ictx->ended);
+		Assert(ctx->tablerel);
+		Assert(OidIsValid(ctx->table));
+		return;
+	}
```
akuzm, line 250: I see lots of uncovered lines here and there, are we missing some tests?

erimatnor: This one I might be able to do something about, but most others are either:
* Freeing materialized tuples in scans (this seldom happens and is not up to us):
```
if (should_free)
						heap_freetuple(tuple);
```
* Or, a chunk was concurrently removed before got a lock during scans. We could try to add an isolation test for that case, but it is quite difficult to test for.

src/chunk.c
```diff
@@ -4267,3 +4118,66 @@ ts_chunk_status(PG_FUNCTION_ARGS)
...
+	if (chunk_lockmode == NoLock)
+		return true;
```
erimatnor, line 4134: We never call this function with NoLock in our code, but the line should probably be here for completeness. (This is also based on corresponding locking done in PG's `find_inheritance_children()`

src/scanner.c
erimatnor: Added a new isolation test and scanner native test to improve the coverage.

scripts/test_updates_pg12.sh
```diff
@@ -3,6 +3,7 @@
 set -e
 
 SCRIPT_DIR=$(dirname $0)
+export PG_VERSION=12
```
svenklemm, line 6: If you set this unconditionally here you should remove PG_VERSION env from caller in .github/workflows/update-test.yaml. Also this is unrelated to your change so i'd prefer to see this in separate commit.

scripts/test_updates_pg13.sh
```diff
@@ -3,6 +3,7 @@
 set -e
 
 SCRIPT_DIR=$(dirname $0)
+export PG_VERSION=13
```
svenklemm, line 6: If you set this unconditionally here you should remove PG_VERSION env from caller in .github/workflows/update-test.yaml. Also this is unrelated to your change so i'd prefer to see this in separate commit.

scripts/test_updates_pg14.sh
```diff
@@ -3,6 +3,7 @@
 set -e
 
 SCRIPT_DIR=$(dirname $0)
+export PG_VERSION=14
```
svenklemm, line 6: If you set this unconditionally here you should remove PG_VERSION env from caller in .github/workflows/update-test.yaml. Also this is unrelated to your change so i'd prefer to see this in separate commit.

src/chunk_constraint.c
```diff
@@ -782,7 +803,8 @@ ts_chunk_constraint_delete_by_dimension_slice_id(int32 dimension_slice_id)
 		ts_scan_iterator_create(CHUNK_CONSTRAINT, RowExclusiveLock, CurrentMemoryContext);
 	int count = 0;
 
-	init_scan_by_dimension_slice_id(&iterator, dimension_slice_id);
+	ts_chunk_constraint_scan_iterator_set_slice_id(&iterator, dimension_slice_id);
+
```
akuzm, line 807: Codecov complains here, does this mean this function is never used? May be a good time to delete it while we've noticed.

akuzm (approved):

src/chunk.c
akuzm: I'd use `Assert(false)` for the stuff that should never happen, but maybe it's fine this way too.

nikkhils (approved):
LGTM overall!

svenklemm (approved):

src/chunk_constraint.c
erimatnor: Yeah, I think that might be the case. Probably related to the change a while back where chunks are tombstoned (with the dropped=true) instead of deleted. 

I'd prefer to leave further refactoring out of this PR, however.

scripts/test_updates_pg13.sh
erimatnor: I don't think it make sense to have a script named `xyz_pg13.sh` and  then run it with, e.g., PG12. So, yes, we should probably make this unconditional or create a generic script that can run any postgres version. I'll leave it for a separate PR, though.

src/chunk.c
erimatnor: I inverted the check instead, which makes the logic equivalent but codecov should stop complaining.

We only use AccessShareLock currently, so other lockmodes not tested either.

PR #4119, opened 2022-02-22
Update remote plan costs to PG14 code base

nikkhils:
The remote DataNodeScan and FDW costing model was based on the
underlying PG11 code base. A lot of changes have occurred in PG since
then. So, update the costing model to the latest code base. We will
make further modifications on top of this for our specific use cases
soon.

Includes updates to the plans which now push order by clauses to the
datanodes in more cases.

### DIFF
--- a/tsl/src/fdw/estimate.c
+++ b/tsl/src/fdw/estimate.c
@@ -8,6 +8,7 @@
 #include <nodes/nodeFuncs.h>
 #include <optimizer/cost.h>
 #include <optimizer/clauses.h>
+#include <optimizer/pathnode.h>
 #include <optimizer/prep.h>
 #include <optimizer/tlist.h>
 #include <optimizer/paths.h>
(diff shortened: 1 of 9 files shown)


tsl/src/fdw/estimate.c
```diff
@@ -285,53 +136,312 @@ adjust_foreign_grouping_path_cost(PlannerInfo *root, List *pathkeys, double retr
...
+		TsFdwRelInfo *fpinfo_i;
+		TsFdwRelInfo *fpinfo_o;
+		QualCost join_cost;
+		QualCost remote_conds_cost;
+		double nrows;
+
+		/* Use rows/width estimates made by the core code. */
+		rows = foreignrel->rows;
+		width = foreignrel->reltarget->width;
+
```
akuzm, line 227: Lots of stuff not covered, is this because we don't support join pushdown? May be better to remove it then.

nikkhils: Why not leave it around for the future when we do support it? @akuzm

akuzm: We could, but I'm afraid that after being dead code for many months, it might end up broken in some subtle ways. And it will also confuse the readers who might think we somehow support the joins. On the other hand, deleting it now and copying it back will also be a pain, so no strong opinion here.

tsl/src/fdw/estimate.c
```diff
@@ -285,53 +136,312 @@ adjust_foreign_grouping_path_cost(PlannerInfo *root, List *pathkeys, double retr
...
+		nrows = clamp_row_est(nrows * fpinfo->joinclause_sel);
+		run_cost += nrows * remote_conds_cost.per_tuple;
+		run_cost += fpinfo->local_conds_cost.per_tuple * retrieved_rows;
+
+		/* Add in tlist eval cost for each output row */
+		startup_cost += foreignrel->reltarget->cost.startup;
+		run_cost += foreignrel->reltarget->cost.per_tuple * rows;
+	}
+	else if (IS_UPPER_REL(foreignrel))
+	{
```
erimatnor, line 296: I would prefer that these changes are applied to `get_upper_rel_estimate()`? 

Now it is impossible to see the diff and whether changes we've made are lost or not. I don't think the goal here should be to 100% adopt PG's costing code as we've made changes and probably want to make further changes to adapt the code to account for async execution across multiple nodes.

tsl/src/fdw/scan_plan.c
```diff
@@ -942,6 +1117,415 @@ add_foreign_grouping_paths(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo
...
 
+/*
+ * add_foreign_ordered_paths
+ *		Add foreign paths for performing the final sort remotely.
+ *
+ * Given input_rel contains the source-data Paths.  The paths are added to the
+ * given ordered_rel.
+ */
+static void
+add_foreign_ordered_paths(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo *ordered_rel,
```
erimatnor, line 1128: This whole function is never called according to code coverage, so we should not add it (now).

tsl/src/fdw/scan_plan.c
```diff
@@ -942,6 +1117,415 @@ add_foreign_grouping_paths(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo
...
+	 */
+	if (!parse->rowMarks && !extra->limit_needed)
+		return;
+
+	/* We don't support cases where there are any SRFs in the targetlist */
+	if (parse->hasTargetSRFs)
+		return;
+
+	/* Save the input_rel as outerrel in fpinfo */
+	fpinfo->outerrel = input_rel;
```
erimatnor, line 1292: Looks like, from here on, this code is never executed. We should not include the code in that case.

tsl/test/expected/dist_query.out
```diff
@@ -5106,19 +5106,19 @@ LIMIT 10
                            Data node: data_node_1
                            Fetcher Type: Row by row
                            Chunks: _dist_hyper_1_1_chunk, _dist_hyper_1_4_chunk, _dist_hyper_1_8_chunk, _dist_hyper_1_12_chunk, _dist_hyper_1_13_chunk, _dist_hyper_1_15_chunk, _dist_hyper_1_17_chunk
-                           Remote SQL: SELECT "time", device FROM public.hyper WHERE _timescaledb_internal.chunks_in(public.hyper.*, ARRAY[1, 2, 3, 4, 5, 6, 7])
+                           Remote SQL: SELECT "time", device FROM public.hyper WHERE _timescaledb_internal.chunks_in(public.hyper.*, ARRAY[1, 2, 3, 4, 5, 6, 7]) ORDER BY device ASC NULLS LAST
```
erimatnor, line 5109: This seems like an unnecessary ORDER BY

erimatnor (changes requested):
This patch is very difficult to review and I am not convinced that this is the right approach.

* It is impossible to see from the diff what has actually changed because the code refactors (or throws away our refactoring) at the same time as new functionality is imported.
* It is hard to see if our own changes are lost or not in the process
* A lot of unused code (e.g., for joins) is imported as evident by code coverage. We should not include code that is never executed. It makes it harder to understand the code that matters and we don't know if the dead code actually works.

Apart from the issues above, it seems the plan changes we see are actually adding ORDER BY when it is not needed. So, those would be worse plans, right?

tsl/src/fdw/estimate.c
nikkhils: Ok, can remove the JOIN related code path.

tsl/src/fdw/scan_plan.c
nikkhils: @erimatnor this function and others will be called in my follow on patch for upper rel handling that I am stabilizing currently

tsl/src/fdw/scan_plan.c
nikkhils: @erimatnor this function and others will be called in my follow on patch for upper rel handling that I am stabilizing currently

tsl/src/fdw/scan_plan.c
```diff
@@ -942,6 +1117,415 @@ add_foreign_grouping_paths(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo
...
+	/* and add it to the final_rel */
+	add_path(final_rel, (Path *) final_path);
+}
+
+/*
+ * Find an equivalence class member expression to be computed as a sort column
+ * in the given target.
+ */
+Expr *
+find_em_expr_for_input_target(PlannerInfo *root, EquivalenceClass *ec, PathTarget *target)
```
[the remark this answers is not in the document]
nikkhils: @erimatnor this function and others will be called in my follow on patch for upper rel handling that I am stabilizing currently

tsl/src/fdw/scan_plan.c
```diff
@@ -963,25 +1547,37 @@ fdw_create_upper_paths(TsFdwRelInfo *input_fpinfo, PlannerInfo *root, UpperRelat
...
 			add_foreign_grouping_paths(root,
 									   input_rel,
 									   output_rel,
 									   (GroupPathExtraData *) extra,
 									   create_path);
 			break;
-			/* Currently not handled (or received) */
-		case UPPERREL_DISTINCT:
 		case UPPERREL_ORDERED:
+			add_foreign_ordered_paths(root, input_rel, output_rel, create_path);
```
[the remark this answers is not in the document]
nikkhils: @erimatnor this function and others will be called in my follow on patch for upper rel handling that I am stabilizing currently

tsl/test/expected/dist_query.out
```diff
@@ -6199,19 +6199,19 @@ LIMIT 10
                            Data node: data_node_1
                            Fetcher Type: Row by row
                            Chunks: _dist_hyper_2_20_chunk
-                           Remote SQL: SELECT "time", device FROM public.hyper1d WHERE _timescaledb_internal.chunks_in(public.hyper1d.*, ARRAY[8])
+                           Remote SQL: SELECT "time", device FROM public.hyper1d WHERE _timescaledb_internal.chunks_in(public.hyper1d.*, ARRAY[8]) ORDER BY device ASC NULLS LAST
```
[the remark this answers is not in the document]
nikkhils: The upper rel optimization and cost related optimizations should remove the sort step on the AN so then it will be a good change right?

PR #4120, opened 2022-02-22
Add logging for retention policy

konskov:
Also remove unused code from compression_api. The function
policy_compression_get_verbose_log was unused. Moved it to
policy_utils and renamed to policy_get_verbose_log so that it can
be used by all policies

### DIFF
--- a/tsl/src/bgw_policy/job.c
+++ b/tsl/src/bgw_policy/job.c
@@ -56,6 +56,27 @@
 #define REORDER_SKIP_RECENT_DIM_SLICES_N 3
 
 static void
+log_retention_boundary(int elevel, PolicyRetentionData *policy_data, const char *message)
+{
+	char *relname;
+	Datum boundary;
+	Oid outfuncid = InvalidOid;
+	bool isvarlena;
+
+	getTypeOutputInfo(policy_data->boundary_type, &outfuncid, &isvarlena);
+
+	relname = get_rel_name(policy_data->object_relid);
+	boundary = policy_data->boundary;
+
+	if (OidIsValid(outfuncid))
+		elog(elevel,
+			 "%s \"%s\": dropping data older than %s",
+			 message,
+			 relname,
+			 DatumGetCString(OidFunctionCall1(outfuncid, boundary)));
+}
+
+static void
 enable_fast_restart(int32 job_id, const char *job_name)
 {
 	BgwJobStat *job_stat = ts_bgw_job_stat_find(job_id);
@@ -250,8 +271,13 @@ bool
 policy_retention_execute(int32 job_id, Jsonb *config)
 {
 	PolicyRetentionData policy_data;
+	bool verbose_log;
 
 	policy_retention_read_and_validate_config(config, &policy_data);
+
+	verbose_log = policy_get_verbose_log(config);
+	if (verbose_log)
+		log_retention_boundary(LOG, &policy_data, "applying retention policy to hypertable");
 
 	chunk_invoke_drop_chunks(policy_data.object_relid,
 							 policy_data.boundary,
@@ -552,6 +578,13 @@ job_execute(BgwJob *job)
 	StringInfo query;
 	Portal portal = ActivePortal;
 
+	if (job->fd.config)
+		elog(DEBUG1,
+			 "Executing %s with parameters %s",
+			 NameStr(job->fd.proc_name),
+			 DatumGetCString(DirectFunctionCall1(jsonb_out, JsonbPGetDatum(job->fd.config))));
+	else
+		elog(DEBUG1, "Executing %s with no parameters", NameStr(job->fd.proc_name));
 	/* Create a portal if there's no active */
 	if (!PortalIsValid(portal))
 	{
--- a/tsl/src/bgw_policy/policy_utils.c
+++ b/tsl/src/bgw_policy/policy_utils.c
@@ -132,3 +132,12 @@ get_open_dimension_for_hypertable(const Hypertable *ht)
 	}
 	return open_dim;
 }
+
+bool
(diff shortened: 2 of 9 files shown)


tsl/src/bgw_policy/policy_utils.c
```diff
@@ -132,3 +132,12 @@ get_open_dimension_for_hypertable(const Hypertable *ht)
...
 }
+
+bool
+policy_get_verbose_log(const Jsonb *config)
+{
+	bool found;
+	bool verbose_log = ts_jsonb_get_bool_field(config, CONFIG_KEY_VERBOSE_LOG, &found);
+
+	return found ? verbose_log : false;
+}
```
mfundul, line 143:
```suggestion
}

```

tsl/src/bgw_policy/job.c
```diff
@@ -250,9 +273,15 @@ bool
 policy_retention_execute(int32 job_id, Jsonb *config)
 {
 	PolicyRetentionData policy_data;
+	bool verbose_log;
+	int elevel;
 
 	policy_retention_read_and_validate_config(config, &policy_data);
 
+	verbose_log = policy_get_verbose_log(config);
+	elevel = (verbose_log) ? LOG : DEBUG1;
```
mfundul, line 282: I think the original idea was not to have `DEBUG` vs `LOG` but simply the following kind of thing:
```
		if (verbose_log)
			elog(LOG,
```

tsl/src/bgw_policy/job.c
```diff
@@ -55,6 +55,29 @@
...
+	boundary = policy_data->boundary;
+
+	if (outfuncid != InvalidOid)
+		elog(elevel,
+			 "%s \"%s\": dropping data older than %s",
+			 message,
+			 relname,
+			 DatumGetCString(OidFunctionCall1(outfuncid, boundary)));
+	else
+		ereport(ERROR, errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("Invalid boundary type"));
```
mfundul, line 78: If this function is eventually conditionally called then we can't be throwing exceptions here depending on the `verbose_log` level. An error logging function should never influence the result of execution.

tsl/src/bgw_policy/job.c
konskov: I did it this way because it's similar to what's done in `continuous_agg_refresh_internal`:
```
log_refresh_window(callctx == CAGG_REFRESH_POLICY ? LOG : DEBUG1, 
cagg, 
&refresh_window, 
"refreshing continuous aggregate");
```
Do you think the DEBUG should be completely removed?

tsl/src/bgw_policy/job.c
```diff
@@ -55,6 +55,29 @@
...
+	Datum boundary;
+	Oid outfuncid = InvalidOid;
+	bool isvarlena;
+
+	getTypeOutputInfo(policy_data->boundary_type, &outfuncid, &isvarlena);
+
+	relname = get_rel_name(policy_data->object_relid);
+	boundary = policy_data->boundary;
+
+	if (outfuncid != InvalidOid)
```
fabriziomello, line 71:
```suggestion
	if (OidIsValid(outfuncid))
```

Just for better code readability

tsl/src/bgw_policy/job.c
fabriziomello: I think we don't need to complicate things here and do the same what we already does with `log_refresh_window`

mfundul: @gayyappan and @svenklemm 's proposal was to not do that though, and only use `LOG` instead, and this is why I suggested it. You can look how it was done in https://github.com/timescale/timescaledb/pull/3520/files

fabriziomello: Ahh now I got what u meant... I was looking into our `master` branch and didn't found this `verbose_log` usage because after the PR you mentioned we changed the `policy_compression_execute` re-implementing it as a PL/pgSQL procedure instead of C code and we loose it in our code base.

Anyway the usage on the procedure is similar what you said: https://github.com/timescale/timescaledb/blob/master/sql/policy_internal.sql#L79

fabriziomello:
@konskov Plesa rebase it so we can do another round of reviews.

fabriziomello (approved):

tsl/test/sql/bgw_reorder_drop_chunks.sql
```diff
@@ -285,3 +285,55 @@ SELECT show_chunks('test_drop_chunks_table');
...
+SELECT alter_job(id,config:=jsonb_set(config,'{verbose_log}', 'true'))
+ FROM _timescaledb_config.bgw_job WHERE id = :drop_chunks_tsntz_job_id;
+
+CALL run_job(:drop_chunks_date_job_id);
+SELECT ts_bgw_db_scheduler_test_run_and_wait_for_scheduler_finish(1000);
+
+CALL run_job(:drop_chunks_tsntz_job_id);
+SELECT ts_bgw_db_scheduler_test_run_and_wait_for_scheduler_finish(1000);
+
+SELECT * FROM sorted_bgw_log;
```
mfundul, line 339: You should add a newline character here.

mfundul:
Is the idea that logging will be off by default? It could be desirable to have this kind of logging on by default if we want, maybe add a GUC?

fabriziomello:
> Is the idea that logging will be off by default? It could be desirable to have this kind of logging on by default if we want, maybe add a GUC?

As far I understood this verbose option is off by default (we already do it for compression policies) and maybe adding a new GUC for it should be material for another issue.

svenklemm (approved):
Please add CHANGELOG entry

mfundul (approved):

PR #4123, opened 2022-02-23
Refactor tsl_debug_append_path

svenklemm:
This patch splits the node name logic from the child path logic
to allow getting a string representation for any postgres node.
This adds a new function:
const char * get_node_name(Path *path)

This patch doesn't add any new callers to the function but it will
be used in subsequent patches to produce more user friendly error
messages when unexpected node types are encountered during planning.


fabriziomello (approved):

src/utils.c
```diff
@@ -931,3 +931,69 @@ ts_relation_size(Oid relid)
...
+				default:
+					return psprintf("Path (%d)", path->pathtype);
+			}
+			NODE_CASE(IndexPath);
+			NODE_CASE(BitmapHeapPath);
+			NODE_CASE(BitmapAndPath);
+			NODE_CASE(BitmapOrPath);
+			NODE_CASE(TidPath);
+			NODE_CASE(SubqueryScanPath);
+			NODE_CASE(ForeignPath);
```
fabriziomello, line 969: Shouldn't it be indented with the `case T_Path:`?

src/utils.c
```diff
@@ -931,3 +931,69 @@ ts_relation_size(Oid relid)
...
+			NODE_CASE(UpperUniquePath);
+			NODE_CASE(AggPath);
+			NODE_CASE(GroupingSetsPath);
+			NODE_CASE(MinMaxAggPath);
+			NODE_CASE(WindowAggPath);
+			NODE_CASE(SetOpPath);
+			NODE_CASE(RecursiveUnionPath);
+			NODE_CASE(LockRowsPath);
+			NODE_CASE(ModifyTablePath);
+			NODE_CASE(LimitPath);
```
fabriziomello, line 995: Shouldn't it be indented with the `case T_CustomPath:`?

src/utils.c
fabriziomello, line 969: Shouldn't it be indented with the `case T_Path:`?

svenklemm: fixed

src/utils.c
```diff
@@ -931,3 +931,69 @@ ts_relation_size(Oid relid)
...
+#define NODE_CASE(name)                                                                            \
+	case T_##name:                                                                                 \
+		return STR_VALUE(name)
+
+/*
+ * Return a string with the name of the node.
+ *
+ */
+const char *
+ts_get_node_name(Path *path)
```
gayyappan, line 945: suggest renaming as ts_get_path_node_name.

svenklemm: changed the signature to take Node* instead

PR #4141, opened 2022-03-03
Improve restriction scanning during hypertable expansion

erimatnor:
Improve the performance of metadata scanning during hypertable
expansion.

When a hypertable is expanded to include all children chunks, only the
chunks that match the query restrictions are included. To find the
matching chunks, the planner first scans for all matching dimension
slices. The chunks that reference those slices are the chunks to
include in the expansion.

This change optimizes the scanning for slices by avoiding repeated
open/close of the dimension slice metadata table and index.

At the same time, related dimension slice scanning functions have been
refactored along the same line.

An index on the chunk constraint metadata table is also changed to
allow scanning on dimension_slice_id. Previously, dimension_slice_id
was the second key in the index, which made scans on this key less
efficient.

### DIFF
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -165,9 +165,8 @@ CREATE TABLE _timescaledb_catalog.chunk_constraint (
   UNIQUE (chunk_id, constraint_name)
 );
 
+CREATE INDEX chunk_constraint_dimension_slice_id_idx ON _timescaledb_catalog.chunk_constraint (dimension_slice_id);
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.chunk_constraint', '');
-
-CREATE INDEX chunk_constraint_chunk_id_dimension_slice_id_idx ON _timescaledb_catalog.chunk_constraint (chunk_id, dimension_slice_id);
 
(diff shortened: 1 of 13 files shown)


akuzm (approved):
The perf improvement on my queries is small (1%), but this change makes sense anyway as a continuation of removing excessive open/closes in scans.

erimatnor:
> The perf improvement on my queries is small (1%), but this change makes sense anyway as a continuation of removing excessive open/closes in scans.

@akuzm I fixed the index issue by adding a dedicated dimension_slice index and removing the <chunk_id, dimension_slice> index (which wasn't really used anyway). Could you re-run your test to see if that also helps further?

akuzm:
> @akuzm I fixed the index issue by adding a dedicated dimension_slice index and removing the <chunk_id, dimension_slice> index (which wasn't really used anyway). Could you re-run your test to see if that also helps further?

The resulting query time still doesn't change much (maybe 5%, can't reliably tell it from the noise), but I do see in the flamegraphs that this part now takes less CPU time. I double-checked how we use other indexes and they seem to be OK.

nikkhils (approved):
LGTM overall.

PR #4152, opened 2022-03-08
Add initial support for upper rels

nikkhils:
ORDER BYs and LIMITs clauses (and others) are represented via UPPERREL
entries during the planning process. Our code didn't handle all these
upper rel entries properly. Add initial support for that so that these
could be pushed down to the datanodes. To start off we are supporting
queries that have one datanode involved.

We will need costing changes in a follow on patch to get the full
benefits of supporting these upperrels.

### DIFF
--- a/tsl/src/fdw/scan_plan.c
+++ b/tsl/src/fdw/scan_plan.c
@@ -870,6 +870,64 @@ foreign_grouping_ok(PlannerInfo *root, RelOptInfo *grouped_rel, GroupPathExtraDa
 }
 
 /*
+ * Find an equivalence class member expression to be computed as a sort column
+ * in the given target.
+ */
+static Expr *
+find_em_expr_for_input_target(PlannerInfo *root, EquivalenceClass *ec, PathTarget *target)
+{
+	ListCell *lc1;
+	int i;
+
+	i = 0;
+	foreach (lc1, target->exprs)
+	{
+		Expr *expr = (Expr *) lfirst(lc1);
+		Index sgref = get_pathtarget_sortgroupref(target, i);
+		ListCell *lc2;
+
+		/* Ignore non-sort expressions */
+		if (sgref == 0 || get_sortgroupref_clause_noerr(sgref, root->parse->sortClause) == NULL)
+		{
+			i++;
+			continue;
+		}
+
+		/* We ignore binary-compatible relabeling on both ends */
+		while (expr && IsA(expr, RelabelType))
+			expr = ((RelabelType *) expr)->arg;
+
+		/* Locate an EquivalenceClass member matching this expr, if any */
+		foreach (lc2, ec->ec_members)
+		{
+			EquivalenceMember *em = (EquivalenceMember *) lfirst(lc2);
+			Expr *em_expr;
+
+			/* Don't match constants */
+			if (em->em_is_const)
+				continue;
+
+			/* Ignore child members */
+			if (em->em_is_child)
+				continue;
+
+			/* Match if same expression (after stripping relabel) */
+			em_expr = em->em_expr;
+			while (em_expr && IsA(em_expr, RelabelType))
+				em_expr = ((RelabelType *) em_expr)->arg;
+
+			if (equal(em_expr, expr))
+				return em->em_expr;
+		}
+
+		i++;
+	}
+
+	elog(ERROR, "could not find pathkey item to sort");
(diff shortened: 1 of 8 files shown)


erimatnor (commented):
There's a lot of code paths that are never executed by our current tests. Can we add new tests to execute some of this code, especially the "negative" conditions when the final or sorting is not possible?

I would have expected to see some new tests to show what the new code does. But I don't see any new tests and almost no changes to existing tests (apart from 1 added output) .

nikkhils:
> There's a lot of code paths that are never executed by our current tests. Can we add new tests to execute some of this code, especially the "negative" conditions when the final or sorting is not possible?
> 
> I would have expected to see some new tests to show what the new code does. But I don't see any new tests and almost no changes to existing tests (apart from 1 added output) .

@erimatnor the debug code paths are coming up in the codecov reports. 

Other than that, the existing dist_query.sql touches all these new functions that we have added. Otherwise they would have shown up on the codecov, right?

The reason there's no change in the test output is again because of the costs. The new paths that we add are removed because they are deemed to be costlier than existing ones.

tsl/src/fdw/scan_plan.c
```diff
@@ -869,6 +869,64 @@ foreign_grouping_ok(PlannerInfo *root, RelOptInfo *grouped_rel, GroupPathExtraDa
...
+			if (em->em_is_child)
+				continue;
+
+			/* Match if same expression (after stripping relabel) */
+			em_expr = em->em_expr;
+			while (em_expr && IsA(em_expr, RelabelType))
+				em_expr = ((RelabelType *) em_expr)->arg;
+
+			if (equal(em_expr, expr))
+				return em->em_expr;
```
erimatnor, line 920: It looks like the code always exits here on first iteration with current tests, as line 923 is never executed.

tsl/src/fdw/scan_plan.c
```diff
@@ -942,46 +1000,505 @@ add_foreign_grouping_paths(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo
...
+		 * join relation.
+		 */
+		Assert(input_rel->reloptkind == RELOPT_BASEREL || input_rel->reloptkind == RELOPT_JOINREL ||
+			   (input_rel->reloptkind == RELOPT_UPPER_REL && ifpinfo->stage == UPPERREL_ORDERED &&
+				(ifpinfo->outerrel->reloptkind == RELOPT_BASEREL ||
+				 ifpinfo->outerrel->reloptkind == RELOPT_JOINREL)));
+
+		foreach (lc, input_rel->pathlist)
+		{
+			Path *path = (Path *) lfirst(lc);
```
erimatnor, line 1211: This loop is never executed.

tsl/src/fdw/scan_plan.c
```diff
@@ -942,46 +1000,505 @@ add_foreign_grouping_paths(PlannerInfo *root, RelOptInfo *input_rel, RelOptInfo
...
+	 */
+	if (parse->commandType != CMD_SELECT)
+		return;
+
+	/*
+	 * No work if there is no FOR UPDATE/SHARE clause and if there is no need
+	 * to add a LIMIT node
+	 */
+	if (!parse->rowMarks && !extra->limit_needed)
+		return;
```
erimatnor, line 1168: Looks like this function always exits here or statement above. The code below is never executed.

erimatnor:
> > There's a lot of code paths that are never executed by our current tests. Can we add new tests to execute some of this code, especially the "negative" conditions when the final or sorting is not possible?
> > I would have expected to see some new tests to show what the new code does. But I don't see any new tests and almost no changes to existing tests (apart from 1 added output) .
> 
> @erimatnor the debug code paths are coming up in the codecov reports.
> 
> Other than that, the existing dist_query.sql touches all these new functions that we have added. Otherwise they would have shown up on the codecov, right?
> 

In some cases the functions you added exits early, so most of the code is never executed. If you look at the codecov, there are a lot of code in the added functions that is not covered by codecov. Is there anything we can do about that? Also, I am wondering why there are no changes to existing plans. What is the code actually doing if there is no visible effect? 

> The reason there's no change in the test output is again because of the costs. The new paths that we add are removed because they are deemed to be costlier than existing ones.

If you look more closely at code coverage, some of the new functions exits in the first few statements, so a lot of the added code is never executed.

akuzm:
Multinode not supported anymore.

PR #4169, opened 2022-03-15
Add chunk exclusion for DELETE

svenklemm:
Currently only IMMUTABLE constraints will exclude chunks from a DELETE plan,
with this patch STABLE expressions will be used to exclude chunks as well.
This is a big performance improvement as chunks not matching partitioning
column constraints don't have to be scanned for DELETEs.
Additionally this improves usability of DELETEs on hypertables with some
chunks compressed. Previously you weren't able to do DELETE on those hypertables
which had non-IMMUTABLE constraints.

With this patch the plan for DELETE on hypertables looks like this:

 Custom Scan (HypertableModify) (actual rows=0 loops=1)
   ->  Delete on metrics (never executed)
         Delete on metrics metrics_1
         Delete on _hyper_5_8_chunk metrics
         Delete on _hyper_5_11_chunk metrics
         Delete on _hyper_5_12_chunk metrics
         Delete on _hyper_5_13_chunk metrics
         Delete on _hyper_5_14_chunk metrics_2
         ->  Custom Scan (ChunkAppend) on metrics (actual rows=1 loops=1)
               Chunks excluded during startup: 4
               ->  Seq Scan on metrics metrics_1 (actual rows=0 loops=1)
                     Filter: ("time" > (now() - '3 years'::interval))
               ->  Bitmap Heap Scan on _hyper_5_14_chunk metrics_2 (actual rows=1 loops=1)
                     Recheck Cond: ("time" > (now() - '3 years'::interval))
                     Heap Blocks: exact=1
                     ->  Bitmap Index Scan on _hyper_5_14_chunk_metrics_time_idx (actual rows=1 loops=1)
                           Index Cond: ("time" > (now() - '3 years'::interval))


src/nodes/hypertable_modify.c
```diff
@@ -524,8 +571,48 @@ static CustomPathMethods hypertable_modify_path_methods = {
...
+					if (contain_mutable_functions((Node *) rinfo->clause) ||
+						ts_contain_param((Node *) rinfo->clause))
+						return true;
+				}
+				return false;
+				break;
+			}
+		default:
+			return false;
+	}
```
fabriziomello, line 609:
```suggestion
	if (nodeTag(path) == T_AppendPath)
	{
		/*
		 * If there are clauses that have mutable functions, or clauses that reference
		 * Params this Path might benefit from startup or runtime exclusion
		 */
		AppendPath *append = castNode(AppendPath, path);
		ListCell *lc;

		/* Don't create ChunkAppend with no children */
		if (list_length(append->subpaths) == 0)
			return false;

		foreach (lc, rel->baserestrictinfo)
		{
			RestrictInfo *rinfo = (RestrictInfo *) lfirst(lc);

			if (contain_mutable_functions((Node *) rinfo->clause) ||
				ts_contain_param((Node *) rinfo->clause))
				return true;
		}
	}
	return false;
```

Do we really need a `switch...case` here?

tsl/test/expected/delete_exclusion.out
```diff
@@ -0,0 +1,476 @@
...
+               ->  Seq Scan on metrics_timestamptz metrics_timestamptz_1 (actual rows=0 loops=1)
+                     Filter: ("time" = 'Sat Jan 01 00:00:00 2000 PST'::timestamp with time zone)
+               ->  Index Scan using _hyper_6_6_chunk_metrics_timestamptz_time_idx on _hyper_6_6_chunk metrics_timestamptz_2 (actual rows=1 loops=1)
+                     Index Cond: ("time" = 'Sat Jan 01 00:00:00 2000 PST'::timestamp with time zone)
+(9 rows)
+
+ROLLBACK;
+-- stable constraints
+-- should have ChunkAppend since constraint is stable
+-- should only hit 1 chunk and base table
```
akuzm, line 319: For stable constraints, if we prepare a statement and it uses a generic plan, is it going to work correctly? In this case, the stable functions should be re-evaluated and chunk exclusion should be done again.

svenklemm: For stable constraints this happens at executor startup not during planning, so every execution will reevaluate constraints.

fabriziomello: Should we need to add some tests here?

src/nodes/hypertable_modify.c
svenklemm: if we ever want to support more then append we need the switch

fabriziomello: So "if" and "when" it happen then we turn it into a `switch...case` IMHO

src/nodes/hypertable_modify.c
```diff
@@ -524,8 +571,48 @@ static CustomPathMethods hypertable_modify_path_methods = {
 	.PlanCustomPath = hypertable_modify_plan_create,
 };
 
+#if PG14_GE
+static bool
+should_chunk_append(Hypertable *ht, PlannerInfo *root, RelOptInfo *rel, Path *path)
```
akuzm, line 576: This name was somewhat confusing when I first read the code, I'd call it 'use_chunk_append_path'

akuzm (approved):
I think I understand what this patch does -- when we have mutable constraints, puts ChunkAppend path under ModifyHypertable, and ChunkAppend can do runtime chunk exclusion to make use of these constraints. Makes sense.

src/nodes/hypertable_modify.c
```diff
@@ -401,6 +422,35 @@ ts_hypertable_modify_fixup_tlist(Plan *plan)
 	}
 }
 
+/* ROWID_VAR only exists in PG14+ */
+#if PG14_GE
+List *
+ts_replace_rowid_vars(PlannerInfo *root, List *tlist, int varno)
```
mfundul, line 428:
```suggestion
ts_replace_rowid_vars(PlannerInfo *root, List *tlist, Index rt_index)
```
I don't really understand the arguments of this function. Maybe rename them some way? Or add some comment?

svenklemm: Those are very common names in the planner PlannerInfo is almost always named root and targetlists are commonly abbreviated as tlist.

gayyappan (commented):
Open question related chunk exclusion support for deletes in the presence of joins.
Other comments are minor.

test/expected/insert-14.out
```diff
@@ -317,7 +317,7 @@ SELECT 1 \g | grep -v "Planning" | grep -v "Execution"
  Result (actual rows=1 loops=1)
    CTE insert_cte
      ->  Custom Scan (HypertableModify) (actual rows=0 loops=1)
-           ->  Insert on one_space_test (never executed)
+           ->  Insert on one_space_test (actual rows=0 loops=1)
```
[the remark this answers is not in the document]
svenklemm: Due to the way we hijack ModifyTable on PG14+ instrumentation for that node was not getting updated, so the `never executed` was actually wrong as you can see from child nodes being executed.

src/nodes/hypertable_modify.c
```diff
@@ -524,8 +571,48 @@ static CustomPathMethods hypertable_modify_path_methods = {
...
+				/* Don't create ChunkAppend with no children */
+				if (list_length(append->subpaths) == 0)
+					return false;
+
+				foreach (lc, rel->baserestrictinfo)
+				{
+					RestrictInfo *rinfo = (RestrictInfo *) lfirst(lc);
+
+					if (contain_mutable_functions((Node *) rinfo->clause) ||
+						ts_contain_param((Node *) rinfo->clause))
```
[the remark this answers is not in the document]
svenklemm: i'll definitely add more tests, DELETE is severely undertested atm.

fabriziomello (approved):

src/planner.c
```diff
@@ -1115,15 +1114,9 @@ involves_hypertable(PlannerInfo *root, RelOptInfo *rel)
  *		  |
  *	  [ subplan ]
  *
- * PG11 adds a value to the create_upper_paths_hook for FDW support. (See:
- * https://github.com/postgres/postgres/commit/7e0d64c7a57e28fbcf093b6da9310a38367c1d75).
- * Additionally, it calls the hook in a different place, once for each
- * RelOptInfo (see:
- * https://github.com/postgres/postgres/commit/c596fadbfe20ff50a8e5f4bc4b4ff5b7c302ecc0),
- * we do not change our behavior yet, but might choose to in the future.
  */
```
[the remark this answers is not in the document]
svenklemm: done

PR #4186, opened 2022-03-21
Fix owner change for distributed hypertable

### LINKED ISSUE #4180: [Bug]: ALTER TABLE OWNER TO does not work for distributed hypertable
### What type of bug is this?

Unexpected error

Access node

### What happened?

This command should be supported, there is no technical reason to block it. Currently it produces an error:

```
```

2.6.0

### PostgreSQL version used

14.2

Debian

Source

On prem/Self-hosted

_No response_

_No response_

pmwkaa:
Allow ALTER TABLE OWNER TO command to be used with distributed
hypertable.

Fix #4180

### DIFF
--- a/tsl/test/expected/dist_ddl.out
+++ b/tsl/test/expected/dist_ddl.out
@@ -462,6 +462,52 @@ ALTER TABLE some_schema.disttable SET SCHEMA public;
 ALTER TABLE disttable SET SCHEMA some_unexist_schema;
 ERROR:  schema "some_unexist_schema" does not exist
 \set ON_ERROR_STOP 1
+-- OWNER TO
+RESET ROLE;
+ALTER TABLE disttable OWNER TO :ROLE_2;
+SELECT tableowner FROM pg_tables WHERE tablename = 'disttable';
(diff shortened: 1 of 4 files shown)


tsl/test/expected/dist_ddl.out
```diff
@@ -462,6 +462,8 @@ ALTER TABLE some_schema.disttable SET SCHEMA public;
 ALTER TABLE disttable SET SCHEMA some_unexist_schema;
 ERROR:  schema "some_unexist_schema" does not exist
 \set ON_ERROR_STOP 1
+-- OWNER TO
+ALTER TABLE disttable OWNER TO :ROLE_1;
```
nikkhils, line 466: Don't we need to check that the command has be carried out properly across the entire cluster? Maybe setting to `ROLE_1` and doing a SELECT from disttable will be good enough to check that?

svenklemm (approved):

svenklemm:
Please add CHANGELOG entry

nikkhils (approved):
LGTM

PR #4198, opened 2022-03-29
Remove `chunk_id` from materialization hypertable

fabriziomello:
Avoid the re-aggregation in Continuous Aggregates user view removing
the `chunk_id` from the materialization hypertable.

Also added new metadata column named `finalized` to `continuous_cagg`
catalog table in order to store information about the new following
finalized version of Continuous Aggregates that will not need the
partials anymore. This flag is important to maintain backward
compatibility with previous Continuous Aggregate implementation that
requires the `chunk_id` to refresh data properly.


sql/updates/reverse-dev.sql
```diff
@@ -2,3 +2,88 @@ DROP VIEW _timescaledb_internal.hypertable_chunk_local_size;
...
+    END IF;
+END;
+$$
+LANGUAGE 'plpgsql';
+
+DROP VIEW IF EXISTS timescaledb_information.hypertables;
+DROP VIEW IF EXISTS timescaledb_information.continuous_aggregates;
+
+ALTER TABLE _timescaledb_catalog.continuous_agg
+    DROP COLUMN finalized;
```
svenklemm, line 35: need to rebuild the table

fabriziomello: but this is exactly what I'm doing below

svenklemm: yes  i saw that not sure why you are dropping the column here when you rebuild the table below

sql/updates/reverse-dev.sql
```diff
@@ -2,3 +2,88 @@ DROP VIEW _timescaledb_internal.hypertable_chunk_local_size;
...
+-- will mess with `pg_attribute.attnum` and we will end up with issues when trying
+-- to update data in those catalog tables.
+--
+ALTER EXTENSION timescaledb
+    DROP TABLE _timescaledb_catalog.continuous_agg;
+
+ALTER TABLE _timescaledb_catalog.continuous_aggs_materialization_invalidation_log
+    DROP CONSTRAINT continuous_aggs_materialization_invalid_materialization_id_fkey;
+
+CREATE TEMP TABLE __tmp_ts_catalog_continuous_agg ON COMMIT DROP
```
svenklemm, line 54: this won't work in certain environments, you can't create temp tables when run under pgextwlist

tsl/src/continuous_aggs/create.c
```diff
@@ -1550,83 +1548,6 @@ mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input, int original_q
 	return var;
 }
 
-/*add internal columns for the materialization table */
```
konskov, line 1553: This function is currently used by `cagg_create` and by `cagg_rebuild_view_definition`. So I don't think it should be removed.

fabriziomello: This is the main purpose of this PR, remove the internal column named `chunk_id` so we'll not need it anymore in the following PR for finals form.

konskov: Confusion points for me:

I thought that the purpose of `cagg_rebuild_view_definition` is to check for errors in an existing (old) cagg. That is, a cagg that still includes the `chunk_id` in the materialization hypertable. And the check for errors is performed by checking if there is a mismatch between the number of columns of the existing (old) cagg’s materialization table and the materialization table that would be generated by the current code (after [this fix](https://github.com/timescale/timescaledb/pull/3899#event-6447088872)), which also includes the `chunk_id` in the materialization hypertable. 

Currently this check uses `mattablecolumninfo_addinternal`, and I haven’t understood how it can still be performed if this function is removed. Won’t it always detect a mismatch and fail in that case?
 
 Apologies if I'm horribly wrong about this 🥲

fabriziomello: @konskov I made confusion here... you're correct... working to fix it.

svenklemm (approved):

tsl/test/expected/cagg_drop_chunks.out
```diff
@@ -145,184 +145,3 @@ NOTICE:  drop cascades to 32 other objects
...
--- This is the simplest case, when the updated bucket is inside a chunk, so it is expected
--- that the update is refreshed.
-UPDATE conditions SET value = 4.00 WHERE time_int = 2;
--- This case updates values in the bucket, which affects two different partials in
--- two different chunks.
-UPDATE conditions SET value = 4.00 WHERE time_int = 9;
-UPDATE conditions SET value = 4.00 WHERE time_int = 11;
-SELECT timescaledb_experimental.refresh_continuous_aggregate('conditions_7', show_chunks('conditions', 20));
- refresh_continuous_aggregate 
-------------------------------
```
mfundul, line 202: All those chunk-versus-timebucket cases are not covered by other tests, so we shouldn't just remove them. Maybe figure out which ranges to use with the non-experimental `refresh_continuous_aggregate()`?

sql/views.sql
```diff
@@ -111,9 +111,10 @@ SELECT ht.schema_name AS hypertable_schema,
   cagg.user_view_name AS view_name,
   viewinfo.viewowner AS view_owner,
   cagg.materialized_only,
-  CASE WHEN mat_ht.compressed_hypertable_id IS NOT NULL 
-       THEN TRUE 
-       ELSE FALSE 
+  cagg.finalized,
```
mfundul, line 114: I fear we may be breaking some user API here. Maybe try and keep the ordering the same by putting `finalized` last? Don't know.

test/sql/updates/post.continuous_aggs.sql
```diff
@@ -3,7 +3,48 @@
...
+    stddev_temp,
+    round(stddev_pop) AS stddev_pop,
+    stddev_samp,
+    round(variance) AS variance,
+    round(var_pop) AS var_pop,
+    round(var_samp) AS var_samp,
+    last_temp,
+    last_hl,
+    first_hl,
+    histogram
```
mfundul, line 43: Why do we need these? Are the tests flaky?

tsl/test/sql/continuous_aggs_deprecated.sql
```diff
@@ -0,0 +1,1232 @@
...
+
+CREATE OR REPLACE FUNCTION test.continuous_aggs_find_view(cagg REGCLASS) RETURNS VOID
+AS :TSL_MODULE_PATHNAME, 'ts_test_continuous_agg_find_by_view_name' LANGUAGE C;
+
+\set WAIT_ON_JOB 0
+\set IMMEDIATELY_SET_UNTIL 1
+\set WAIT_FOR_OTHER_TO_ADVANCE 2
+
+-- remove any default jobs, e.g., telemetry so bgw_job isn't polluted
+DELETE FROM _timescaledb_config.bgw_job WHERE TRUE;
```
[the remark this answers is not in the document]
fabriziomello: Hmmm... didn't pay attention on that because I just made a copy of the current `continuous_agg.sql` regression tests... will remove because is useless

test/sql/updates/post.continuous_aggs.sql
fabriziomello: Exactly

sql/views.sql
fabriziomello: It will break just for users that for some weird reason use `SELECT *...` in their applications... but for those users there are a special place reserved in the hell... heheheheheh... just kidding :-)

In fact we can't prevent users to shoot their own foot here... when we introduce the `materialized_only` column it was not added as the last column of the VIEW. For catalog tables is another history because the `attnums` we use in C code.

tsl/src/continuous_aggs/create.c
```diff
@@ -2098,29 +2099,37 @@ cagg_create(const CreateTableAsStmt *create_stmt, ViewStmt *stmt, Query *panquer
...
-	usertbl_rte = list_nth(panquery->rtable, 0);
-	mattablecolumninfo_addinternal(&mattblinfo, usertbl_rte, origquery_ht->htid);
+	/*
+	 * Step 0: add any internal columns needed for materialization based
+	 *         on the user query's table
+	 */
+	if (!finalized)
+	{
+		RangeTblEntry *usertbl_rte;
+		usertbl_rte = list_nth(panquery->rtable, 0);
```
mfundul, line 2130:
```suggestion
		RangeTblEntry *usertbl_rte = list_nth(panquery->rtable, 0);
```

tsl/src/continuous_aggs/create.c
```diff
@@ -2409,8 +2388,11 @@ cagg_rebuild_view_definition(ContinuousAgg *agg, Hypertable *mat_ht)
 
 	mattablecolumninfo_init(&mattblinfo, NIL, NIL, copyObject(direct_query->groupClause));
 	finalizequery_init(&fqi, direct_query, &mattblinfo);
-	RangeTblEntry *usertbl_rte = list_nth(direct_query->rtable, 0);
-	mattablecolumninfo_addinternal(&mattblinfo, usertbl_rte, 0);
+	if (!agg->data.finalized)
```
mfundul, line 2391: We don't need the `if (!agg->data.finalized)` here because it already `return`ed above with `if (agg->data.finalized)`.

fabriziomello: Fixed!

tsl/src/continuous_aggs/create.c
fabriziomello: Fixed

tsl/test/expected/cagg_drop_chunks.out
fabriziomello: I removed this test because removing the `chunk_id` from the materialization hypertable is not possible to it anymore, I mean refresh data from only one specific chunk. It will work just when we have the raw table data chunks aligned with the materialization hypertable chunks but it does not work for all cases.

mfundul: Can we rework those tests to work with ranges?

sql/updates/reverse-dev.sql
```diff
@@ -2,3 +2,101 @@ DROP VIEW _timescaledb_internal.hypertable_chunk_local_size;
...
+
+ALTER TABLE _timescaledb_catalog._tmp_continuous_agg
+    RENAME TO continuous_agg;
+
+ALTER TABLE _timescaledb_catalog.continuous_agg
+    ADD CONSTRAINT continuous_agg_pkey PRIMARY KEY (mat_hypertable_id),
+    ADD CONSTRAINT continuous_agg_partial_view_schema_partial_view_name_key UNIQUE (partial_view_schema, partial_view_name),
+    ADD CONSTRAINT continuous_agg_user_view_schema_user_view_name_key UNIQUE (user_view_schema, user_view_name),
+    ADD CONSTRAINT continuous_agg_mat_hypertable_id_fkey FOREIGN KEY (mat_hypertable_id) REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE,
+    ADD CONSTRAINT continuous_agg_raw_hypertable_id_fkey FOREIGN KEY (raw_hypertable_id) REFERENCES _timescaledb_catalog.hypertable(id) ON DELETE CASCADE;
```
mfundul, line 91: Why didn't we do this during `CREATE TABLE`? Is this for performance reasons?

fabriziomello: Because the `pg_class` entries created by the PRIMARY/UNIQUE indexes with the same name of the original `continuous_agg` table. In `pg_class` we have a unique index by `relname, relnamespace` so it will forbid two objects (in this case indexes) with the same name (constraint name) in the same tablespace.

mkindahl (commented):
Here are some comments. In addition, I think that you should only add necessary indexes and allow the user to add any additional indexes separately. In particular, any columns that are not in the `GROUP BY` clause does not need to be added to an index.

sql/updates/reverse-dev.sql
```diff
@@ -2,3 +2,101 @@ DROP VIEW _timescaledb_internal.hypertable_chunk_local_size;
...
+    raw_hypertable_id integer NOT NULL,
+    user_view_schema name NOT NULL,
+    user_view_name name NOT NULL,
+    partial_view_schema name NOT NULL,
+    partial_view_name name NOT NULL,
+    bucket_width bigint NOT NULL,
+    direct_view_schema name NOT NULL,
+    direct_view_name name NOT NULL,
+    materialized_only bool NOT NULL DEFAULT FALSE
+);
```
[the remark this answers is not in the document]
fabriziomello: > Here are some comments. In addition, I think that you should only add necessary indexes and allow the user to add any additional indexes separately. In particular, any columns that are not in the `GROUP BY` clause does not need to be added to an index.

I've fixed it yesterday!

sql/views.sql
fabriziomello: Ok, you both convinced me... fixed!

test/sql/updates/setup.continuous_aggs.sql
```diff
@@ -7,15 +7,15 @@ CREATE TYPE custom_type AS (high int, low int);
...
-    );
+      temperature DOUBLE PRECISION,
+      humidity    DOUBLE PRECISION,
+      lowp        DOUBLE PRECISION,
+      highp       DOUBLE PRECISION,
+      allnull     DOUBLE PRECISION,
+      highlow     custom_type,
+      bit_int     SMALLINT,
+      good_life   BOOLEAN
+);
```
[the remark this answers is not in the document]
fabriziomello: Make sense. Fixed!

tsl/src/continuous_aggs/materialize.c
```diff
@@ -59,7 +59,8 @@ continuous_agg_update_materialization(SchemaAndName partial_view,
 	if (res != SPI_OK_CONNECT)
 		elog(ERROR, "could not connect to SPI in materializer");
 
-	/* pin the start of new_materialization to the end of new_materialization,
+	/*
+	 * Pin the start of new_materialization to the end of new_materialization,
```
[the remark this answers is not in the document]
fabriziomello: Fixed

mkindahl (approved):

tsl/test/sql/cagg_drop_chunks.sql
```diff
@@ -57,71 +57,3 @@ DROP MATERIALIZED VIEW records_monthly;
...
--- This is the simplest case, when the updated bucket is inside a chunk, so it is expected
--- that the update is refreshed.
-UPDATE conditions SET value = 4.00 WHERE time_int = 2;
-
--- This case updates values in the bucket, which affects two different partials in
--- two different chunks.
-UPDATE conditions SET value = 4.00 WHERE time_int = 9;
-UPDATE conditions SET value = 4.00 WHERE time_int = 11;
-
-SELECT timescaledb_experimental.refresh_continuous_aggregate('conditions_7', show_chunks('conditions', 20));
```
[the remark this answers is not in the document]
fabriziomello: Added some more tests!

tsl/test/expected/cagg_drop_chunks.out
fabriziomello: Added some more tests!

test/sql/updates/post.continuous_aggs.sql
fabriziomello: You're correct... the better place to call round is in the CAgg definition. Fixed it!

PR #4204, opened 2022-03-30
try to always use binary when possible

akuzm:
see what breaks


akuzm:
Everything works, only found a problem with error handling when reading the data from the data nodes. Turns out PG12 errors out if the type oid is different. On later PG versions this is ignored for user-defined data types. I guess we should leave this GUC as is as a fallback at least until we stop supporting PG12.

PR #4209, opened 2022-04-03
Add chunk exclusion for UPDATE for PG14

svenklemm:
Currently only IMMUTABLE constraints will exclude chunks from an UPDATE plan,
with this patch STABLE expressions will be used to exclude chunks as well.
This is a big performance improvement as chunks not matching partitioning
column constraints don't have to be scanned for UPDATEs.
Since the codepath for UPDATE is different for PG < 14 this patch only adds
the optimization for PG14.

With this patch the plan for UPDATE on hypertables looks like this:
```
 Custom Scan (HypertableModify) (actual rows=0 loops=1)
   ->  Update on public.metrics_int2 (actual rows=0 loops=1)
         Update on public.metrics_int2 metrics_int2_1
         Update on _timescaledb_internal._hyper_1_1_chunk metrics_int2
         Update on _timescaledb_internal._hyper_1_2_chunk metrics_int2
         Update on _timescaledb_internal._hyper_1_3_chunk metrics_int2
         ->  Custom Scan (ChunkAppend) on public.metrics_int2 (actual rows=0 loops=1)
               Output: '123'::text, metrics_int2.tableoid, metrics_int2.ctid
               Startup Exclusion: true
               Runtime Exclusion: false
               Chunks excluded during startup: 3
               ->  Seq Scan on public.metrics_int2 metrics_int2_1 (actual rows=0 loops=1)
                     Output: metrics_int2_1.tableoid, metrics_int2_1.ctid
                     Filter: (metrics_int2_1."time" = length(version()))
```


tsl/test/sql/update_exclusion.sql
```diff
@@ -0,0 +1,167 @@
...
+-- should only hit 1 chunk and base table
+BEGIN;
+:PREFIX UPDATE metrics_int2 SET value = 0.3 WHERE time = length(substring(version(),1,23));
+:PREFIX UPDATE metrics_int4 SET value = 0.3 WHERE time = length(substring(version(),1,23));
+:PREFIX UPDATE metrics_int8 SET value = 0.3 WHERE time = length(substring(version(),1,23));
+ROLLBACK;
+
+-- should have ChunkAppend since constraint is stable
+-- should only hit 1 chunk and base table, toplevel rows should be 1
+BEGIN;
```
gayyappan, line 93: To be exhaustive, I would also add a test with multiple column updates 
UPDATE <hypertable> SET col1 =xxx,  col2 = yyy, col3 = zzz
WHERE ....
RETURNING ...

gayyappan (approved):
Changes look good to me. Minor comment.

tsl/test/sql/update_exclusion.sql
fabriziomello: Will it work also with writable CTEs?

svenklemm: added a couple more tests

PR #4211, opened 2022-04-04
Move the check to a more appropriate place

akuzm:
Before, we would complain that we don't support fetching the system
columns with per-data node queries enabled, but still execute the code
that fetches it. Don't do this and complain earlier.


mkindahl (approved):
LGTM

svenklemm (approved):
Please make commit subject a bit more descriptive

Issue #4222, opened 2022-04-07
[Bug]: Very high planner times with decompressed nodes b/c of JIT


cevian:
### What type of bug is this?

Performance issue

### What subsystems and features are affected?

Query planner

### What happened?

Planner times are very high on a quite simple query over compressed data. Turning off JIT and/or removing the ORDER BY makes the issue go away. Explain plans included below. 

Note: PG 13.6 not sure if reoccurs on 14

### TimescaleDB version affected

2.6.0

### PostgreSQL version used

13.6

### What operating system did you use?

Cloud

### What installation method did you use?

Docker

### What platform did you run on?

Timescale Cloud

### Relevant log output and stack trace

```bash
tsdb=> set jit=0;
SET
Time: 20.364 ms
tsdb=> explain SELECT series_id, time, "value" as value FROM "prom_data"."my_hypertable" metric WHERE time >= '2022-03-05T14:45:06.47Z' AND time <= '2022-03-25T10:00:00.062Z' ORDER BY series_id, time;
                                                                                    QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Merge Append  (cost=12.45..600588.56 rows=18619019 width=24)
   Sort Key: metric_1.series_id, metric_1."time"
   ->  Custom Scan (DecompressChunk) on _hyper_5542_173024_chunk metric_1  (cost=0.76..45.02 rows=59000 width=24)
         Filter: (("time" >= '2022-03-05 14:45:06.47+00'::timestamp with time zone) AND ("time" <= '2022-03-25 10:00:00.062+00'::timestamp with time zone))
         ->  Index Scan using compress_hyper_5545_178227_chun__compressed_hypertable_5545_ser on compress_hyper_5545_178227_chunk  (cost=0.28..45.02 rows=59 width=112)
               Filter: ((_ts_meta_max_1 >= '2022-03-05 14:45:06.47+00'::timestamp with time zone) AND (_ts_meta_min_1 <= '2022-03-25 10:00:00.062+00'::timestamp with time zone))
   ->  Custom Scan (DecompressChunk) on _hyper_5542_235085_chunk metric_2  (cost=0.15..80.22 rows=532000 width=24)
         Filter: (("time" >= '2022-03-05 14:45:06.47+00'::timestamp with time zone) AND ("time" <= '2022-03-25 10:00:00.062+00'::timestamp with time zone))
[...]

Issue #4195, opened 2022-03-25
[Bug]: unused columns retrieved from compressed chunks through view


phemmer:
### What type of bug is this?

Performance issue

### What subsystems and features are affected?

Compression

### What happened?

When using a view, data retrieval from compressed chunks returns all columns regardless of whether they are used in the query. This causes significant performance impact.

### TimescaleDB version affected

2.6.0

### PostgreSQL version used

14.2

### What operating system did you use?

docker image timescale/timescaledb:2.6.0-pg14

### What installation method did you use?

Docker

### What platform did you run on?

On prem/Self-hosted

### How can we reproduce the bug?

```sql
CREATE TABLE test (time TIMESTAMP, a INT, b INT, c INT);
SELECT create_hypertable('test', 'time', chunk_time_interval => INTERVAL '1h');
ALTER TABLE test SET (timescaledb.compress);
INSERT INTO test VALUES
    ('2022-03-25T11:00:00Z',1,1,1),
    ('2022-03-25T12:00:00Z',0,0,0)
;
CREATE VIEW testview AS SELECT * FROM test;
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS) SELECT a FROM testview;
```
```
Append  (cost=0.00..71.00 rows=3400 width=4) (actual time=0.004..0.018 rows=2 loops=1)
  Buffers: shared hit=2
  ->  Seq Scan on _timescaledb_internal._hyper_14_18_chunk  (cost=0.00..27.00 rows=1700 width=4) (actual time=0.004..0.004 rows=1 loops=1)
        Output: _hyper_14_18_chunk.a
        Buffers: shared hit=1
  ->  Seq Scan on _timescaledb_internal._hyper_14_19_chunk  (cost=0.00..27.00 rows=1700 width=4) (actual time=0.013..0.013 rows=1 loops=1)
        Output: _hyper_14_19_chunk.a
        Buffers: shared hit=1
Planning:
  Buffers: shared hit=96 read=1
Planning Time: 0.602 ms
Execution Time: 0.031 ms
```
Note the `Output:` line for each chunk, and how it's only outputting column `a`. This is the correct behavior.

Compress a chunk and try again:
```sql
SELECT compress_chunk(chunk_schema || '.' || chunk_name)
    FROM timescaledb_information.chunks
    WHERE hypertable_name='test' AND range_start='2022-03-25T11:00:00Z';
[...]

PR #4234, opened 2022-04-13
Avoid extra projection when scanning a compressed chunk

akuzm:
Namely, we don't have to project the result of the underlying scan of DecompressChunk, because the latter can selectively decompress the resulting columns anyway. The projection slows down the things especially when we use JIT, when the projection has to be compiled for each chunk.

Fixes https://github.com/timescale/timescaledb/issues/4222
Fixes https://github.com/timescale/timescaledb/issues/4195

### Commit message
Avoid extra projection when scanning a compressed chunk

We don't have to project the result of the underlying scan of DecompressChunk, because the latter can selectively decompress the resulting columns anyway. The projection slows down the things especially when we use JIT, when the projection has to be compiled for each chunk.

To avoid this unnecessary work, use the physical targetlist for compressed scan if we can. This commit rewrites the function that determines what we have to decompress, and makes it use the targetlist generated by the planner, instead of building it from scratch. Then we can give it the physical targetlist when possible.

### Implementation details
1. Teach DecompressChunks to skip input columns. #4265
1. Don't put whole-row vars into compressed scan tlist, this makes no sense #4260
1. Instead of building the compressed scan targetlist anew, trust the targetlist generated by the planner. Note that up to this point we still didn't introduce any functional changes.
1. Use the physical targetlist for compressed scan if we can. This is the actual functional change and it's like five lines.

### Narrow table

```
┌─speedup─┬─time_ms─┬─query─────────────────────┐
│    2.11 │  169.82 │ compressed-index-jit.sql  │
│    1.16 │   73.14 │ compressed-index.sql      │
│    1.45 │  586.62 │ jit-compressed.sql        │
│    1.28 │  593.66 │ nojit-compressed-view.sql │
│    1.19 │  601.77 │ nojit-compressed.sql      │
└─────────┴─────────┴───────────────────────────┘
```

The queries are below. The "index" queries use an index scan filtering by the segmentby column. The "view" queries use a simple "select *" view to test #4195.
```
==> compressed-index-jit.sql <==
set jit_above_cost = 1;
set jit_inline_above_cost = 1;
set jit_optimize_above_cost = 1;
set jit = on;
copy (select time + interval '1 second' from live_steps_originals_lll where user_id = 1) to '/dev/null' with (format binary);

==> compressed-index.sql <==
set jit = off;
copy (select time + interval '1 second' from live_steps_originals_lll where user_id = 1) to '/dev/null' with (format binary);

==> jit-compressed.sql <==
set jit = on;
set jit_above_cost = 1;
set jit_inline_above_cost = 1;
set jit_optimize_above_cost = 1;
copy (select time + interval '1 second' from live_steps_originals_lll) to '/dev/null' with (format binary);

==> nojit-compressed-view.sql <==
set jit = off;
copy (select time + interval '1 second' from vvv) to '/dev/null' with (format binary);

==> nojit-compressed.sql <==
set jit = off;
copy (select time + interval '1 second' from live_steps_originals_lll) to '/dev/null' with (format binary);
```

Table:
```
test=# \d live_steps_originals_lll
                Table "public.live_steps_originals_lll"
 Column  │            Type             │ Collation │ Nullable │ Default 
─────────┼─────────────────────────────┼───────────┼──────────┼─────────
 user_id │ bigint                      │           │ not null │ 
 time    │ timestamp without time zone │           │ not null │ 
 steps   │ smallint                    │           │          │ 
Indexes:
    "live_steps_originals_lll_time_idx" btree ("time" DESC)
Triggers:
    ts_insert_blocker BEFORE INSERT ON live_steps_originals_lll FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker()
Number of child tables: 6 (Use \d+ to list them.)


test=# SELECT * FROM timescaledb_information.compression_settings where hypertable_name = 'live_steps_originals_lll';
 hypertable_schema │     hypertable_name      │ attname │ segmentby_column_index │ orderby_column_index │ orderby_asc │ orderby_nullsfirst 
───────────────────┼──────────────────────────┼─────────┼────────────────────────┼──────────────────────┼─────────────┼────────────────────
 public            │ live_steps_originals_lll │ user_id │                      1 │                      │             │ 
 public            │ live_steps_originals_lll │ time    │                        │                    1 │ t           │ f
(2 rows)


test=# select count(*) from live_steps_originals_lll
  count  
─────────
 2592001


CREATE VIEW vvv AS SELECT * FROM live_steps_originals_lll;
```

### Wide table
`live_steps_originals_wide100` is the same as the narrow table, but with 100 steps columns.

```
┌─speedup─┬─time_ms─┬─query─────────────────────┐
│    1.32 │  499.98 │ compressed-wide-100th.sql │
│    1.35 │  489.83 │ compressed-wide-1st.sql   │
│    1.39 │  488.11 │ compressed-wide-50th.sql  │
└─────────┴─────────┴───────────────────────────┘


==> compressed-wide-100th.sql <==
copy (select steps100 + 1 from live_steps_originals_wide100) to '/dev/null' with (format binary);

==> compressed-wide-1st.sql <==
copy (select steps1 + 1 from live_steps_originals_wide100) to '/dev/null' with (format binary);

==> compressed-wide-50th.sql <==
copy (select steps50 + 1 from live_steps_originals_wide100) to '/dev/null' with (format binary);
````

### DIFF
--- a/tsl/src/nodes/decompress_chunk/decompress_chunk.c
+++ b/tsl/src/nodes/decompress_chunk/decompress_chunk.c
@@ -491,12 +491,24 @@ ts_decompress_chunk_generate_paths(PlannerInfo *root, RelOptInfo *chunk_rel, Hyp
 	compressed_rel->reloptkind = RELOPT_DEADREL;
 }
 
+/*
+ * Add a var for a particular column to the reltarget. attrs_used is a bitmap
+ * of which columns we already have in reltarget. We do not add the columns that
+ * are already there, and update it after adding something.
+ */
 static void
 compressed_reltarget_add_var_for_column(RelOptInfo *compressed_rel, Oid compressed_relid,
 										const char *column_name, Bitmapset **attrs_used)
 {
 	AttrNumber attnum = get_attnum(compressed_relid, column_name);
 	Assert(attnum > 0);
+
+	if (bms_is_member(attnum, *attrs_used))
+	{
+		/* This column is already in reltarget, we don't need duplicates. */
+		return;
+	}
+
 	*attrs_used = bms_add_member(*attrs_used, attnum);
 
 	Oid typid, collid;
@@ -518,8 +530,30 @@ compressed_rel_setup_reltarget(RelOptInfo *compressed_rel, CompressionInfo *info
 	Bitmapset *attrs_used = NULL;
 
 	Oid compressed_relid = info->compressed_rte->relid;
+
+	/*
+	 * We have to decompress three kinds of columns:
+	 * 1) output targetlist of the relation,
+	 * 2) columns required for the quals (WHERE),
+	 * 3) columns required for joins.
+	 */
+	List *exprs = list_copy(info->chunk_rel->reltarget->exprs);
 	ListCell *lc;
-	foreach (lc, info->chunk_rel->reltarget->exprs)
+	foreach (lc, info->chunk_rel->baserestrictinfo)
+	{
+		exprs = lappend(exprs, ((RestrictInfo *) lfirst(lc))->clause);
+	}
+	foreach (lc, info->chunk_rel->joininfo)
+	{
+		exprs = lappend(exprs, ((RestrictInfo *) lfirst(lc))->clause);
+	}
+
+	/*
+	 * Now go over the required expressions we prepared above, and add the
+	 * required columns to the compressed reltarget.
+	 */
+	info->compressed_rel->reltarget->exprs = NIL;
+	foreach (lc, exprs)
 	{
 		ListCell *lc2;
 		List *chunk_vars = pull_var_clause(lfirst(lc), PVC_RECURSE_PLACEHOLDERS);
@@ -531,9 +565,12 @@ compressed_rel_setup_reltarget(RelOptInfo *compressed_rel, CompressionInfo *info
 
 			/* skip vars that aren't from the uncompressed chunk */
 			if (chunk_var->varno != info->chunk_rel->relid)
+			{
 				continue;
+			}
 
-			/* if there's a system column or whole-row reference, add a whole-
+			/*
+			 * If there's a system column or whole-row reference, add a whole-
 			 * row reference, and we're done.
 			 */
 			if (chunk_var->varattno <= 0)
--- a/tsl/src/nodes/decompress_chunk/exec.c
+++ b/tsl/src/nodes/decompress_chunk/exec.c
@@ -144,7 +144,11 @@ initialize_column_state(DecompressChunkState *state)
 		next_compressed_scan_attno++;
 
 		AttrNumber output_attno = lfirst_int(lc);
-		Assert(output_attno != 0);
+		if (output_attno == 0)
+		{
+			/* We are asked not to decompress this column, skip it. */
+			continue;
+		}
 
 		DecompressChunkColumnState *column = &state->columns[state->num_columns];
 		state->num_columns++;
--- a/tsl/src/nodes/decompress_chunk/planner.c
+++ b/tsl/src/nodes/decompress_chunk/planner.c
@@ -14,6 +14,7 @@
 #include <nodes/nodeFuncs.h>
 #include <optimizer/optimizer.h>
 #include <optimizer/paths.h>
+#include <optimizer/plancat.h>
 #include <optimizer/restrictinfo.h>
 #include <optimizer/tlist.h>
 #include <parser/parsetree.h>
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

**Comment by akuzm** [comment]
```
     dist_gapfill_pushdown-14     ... FAILED     5863 ms
```
flaky

**Review by mkindahl — COMMENTED** [review]
For the message on the pull request, I think you should focus on describing what the pull request actually does and add all the comments about comparisons to the issue. That way, the Git log is clear on what the change does, and the issue contains information about background and other information relevant to people that run into this issue.

**Comment by akuzm** [comment]
> That way, the Git log is clear on what the change does

I'm not going to make this PR description into a commit message, that's just some context for the reviewers. For the commit message body, I guess I'll use something along the lines of points 3 and 4 from the description, because this is the actual change.

**Comment by mkindahl** [comment]
> > That way, the Git log is clear on what the change does
> 
> I'm not going to make this PR description into a commit message, that's just some context for the reviewers. For the commit message body, I guess I'll use something along the lines of points 3 and 4 from the description, because this is the actual change.

Ok, but part of doing the review is reviewing the commit message, so given that this is not the actual commit message, this makes that part harder. This is why I suggest to put context information with the issue, so that the reviewer can look at the issue and understand more about the context while doing the review and also review the actual commit message.

**Review by svenklemm — APPROVED** [review]

File: tsl/src/nodes/decompress_chunk/exec.c [hunk 1]
```diff
@@ -144,7 +144,11 @@ initialize_column_state(DecompressChunkState *state)
 		next_compressed_scan_attno++;
 
 		AttrNumber output_attno = lfirst_int(lc);
-		Assert(output_attno != 0);
+		if (output_attno == 0)
+		{
+			/* skip this column from compressed scan targetlist, do nothing */
+			continue;
+		}
```
mkindahl, line 151: Add a comment on why this is OK.

tsl/src/nodes/decompress_chunk/planner.c
```diff
@@ -40,169 +41,219 @@ _decompress_chunk_init(void)
...
-												COMPRESSION_COLUMN_METADATA_COUNT_NAME,
-												DECOMPRESS_CHUNK_COUNT_ID,
-												list_length(scan_tlist) + 1);
-	scan_tlist = lappend(scan_tlist, tle);
+		if (compressed_attno == InvalidAttrNumber)
+		{
+			elog(ERROR,
+				 "column '%s' not found in the compressed chunk '%s'",
+				 NameStr(fd->attname),
+				 get_rel_name(path->info->compressed_rte->relid));
```
mkindahl, line 111: Would be good to have a test for this case, just to make sure that it works as expected.

tsl/src/nodes/decompress_chunk/planner.c
```diff
@@ -40,169 +41,219 @@ _decompress_chunk_init(void)
...
+	foreach (lc, scan_tlist)
 	{
-		/* we support tableoid so skip that */
-		if (bit == TableOidAttributeNumber - FirstLowInvalidHeapAttributeNumber)
-			bit = bms_next_member(attrs_used, bit);
+		TargetEntry *target = (TargetEntry *) lfirst(lc);
+		if (!IsA(target->expr, Var))
+		{
+			elog(ERROR, "compressed scan targetlist entries must be Vars");
+		}
```
mkindahl, line 128: Can you create a test case for this? Not sure if it is possible.

tsl/src/nodes/decompress_chunk/planner.c
```diff
@@ -40,169 +41,219 @@ _decompress_chunk_init(void)
...
+			Assert(chunk_attno != InvalidAttrNumber);
+
+			// The versions older than this commit didn't set up the proper
+			// collation and typmod for segmentby columns in compressed chunks,
+			// so we have to determine them from the main hypertable.
+			// Additionally, we have to set the proper type for the compressed
+			// columns. It would be cool to get rid of this code someday and
+			// just use the types from the compressed chunk, but the problem is
+			// that we have to support the chunks created by the older versions
+			// of TimescaleDB.
```
mkindahl, line 169: We use C comments, not C++ comments.

tsl/src/nodes/decompress_chunk/planner.c
```diff
@@ -40,169 +41,219 @@ _decompress_chunk_init(void)
...
+			 * custom scan.
+			 * See compressed_rel_setup_reltarget().
+			 */
+			elog(ERROR, "compressed scan targetlist must not have whole-row vars");
+		}
 
-		foreach (lc, path->info->ht_rte->eref->colnames)
+		char *column_name = get_attname(path->info->compressed_rte->relid,
+										compressed_attno,
+										/* missing_ok = */ false);
```
mkindahl, line 147: Nit: const-correctness

```suggestion
		const char *column_name = get_attname(path->info->compressed_rte->relid,
										compressed_attno,
										/* missing_ok = */ false);
```

tsl/src/nodes/decompress_chunk/planner.c
```diff
@@ -40,169 +41,219 @@ _decompress_chunk_init(void)
...
-			/* bits are offset by FirstLowInvalidHeapAttributeNumber */
-			AttrNumber ht_attno = bit + FirstLowInvalidHeapAttributeNumber;
+		elog(ERROR,
+			 "column '%s' (%d) not found in the scan targetlist for compressed chunk '%s'",
+			 get_attname(path->info->chunk_rte->relid,
+						 bit + FirstLowInvalidHeapAttributeNumber,
+						 /* missing_ok = */ true),
+			 bit + FirstLowInvalidHeapAttributeNumber,
+			 get_rel_name(path->info->compressed_rte->relid));
+	}
```
mkindahl, line 246: Hmm.... suspect it is not easy to create a test for this, but checking. Can you?

mkindahl (commented):
Looks good, but I think it would be good with some comments on some of your choices since it is not obvious, and some tests that seems like they are missing.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -497,6 +497,12 @@ compressed_reltarget_add_var_for_column(RelOptInfo *compressed_rel, Oid compress
 {
 	AttrNumber attnum = get_attnum(compressed_relid, column_name);
 	Assert(attnum > 0);
+
+	if (bms_is_member(attnum, *attrs_used))
+	{
+		return;
+	}
+
```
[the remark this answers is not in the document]
akuzm: Added.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -518,8 +524,24 @@ compressed_rel_setup_reltarget(RelOptInfo *compressed_rel, CompressionInfo *info
...
 	ListCell *lc;
-	foreach (lc, info->chunk_rel->reltarget->exprs)
+	foreach (lc, info->chunk_rel->baserestrictinfo)
+	{
+		exprs = lappend(exprs, ((RestrictInfo *) lfirst(lc))->clause);
+	}
+	foreach (lc, info->chunk_rel->joininfo)
+	{
+		exprs = lappend(exprs, ((RestrictInfo *) lfirst(lc))->clause);
+	}
```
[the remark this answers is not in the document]
akuzm: I look at `joininfo` in line 538, did you mean something else? I extended the comment above.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -531,9 +553,12 @@ compressed_rel_setup_reltarget(RelOptInfo *compressed_rel, CompressionInfo *info
 
 			/* skip vars that aren't from the uncompressed chunk */
 			if (chunk_var->varno != info->chunk_rel->relid)
+			{
 				continue;
+			}
 
-			/* if there's a system column or whole-row reference, add a whole-
+			/*
+			 * If there's a system column or whole-row reference, add a whole-
```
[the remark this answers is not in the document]
akuzm: That's probably a remnant of when I added something else to this branch. That's why I don't like skipping braces for single-statement blocks, makes it harder/error-prone to modify. I can keep the old formatting here.

tsl/src/nodes/decompress_chunk/exec.c
akuzm: That's the intended behavior, output_attno == 0 means 'skip this column'. I reworded the comment.

tsl/src/nodes/decompress_chunk/planner.c
akuzm: Not sure how valuable this is, this would be the user tinkering with the compressed chunk table manually.

tsl/src/nodes/decompress_chunk/planner.c
akuzm: No, this would be a logic error.

tsl/src/nodes/decompress_chunk/planner.c
akuzm: Fixed.

tsl/src/nodes/decompress_chunk/planner.c
akuzm: Fixed. By the way, the return value isn't const, it's an pstrdup'd string. If you mean something like 'static single assignment is generally good', I agree, but this also applies to another ten places in this PR.

tsl/src/nodes/decompress_chunk/planner.c
akuzm: This would be a logic error, so hopefully it's not possible to trigger :) The targetlist for the compressed chunk should have all the required columns.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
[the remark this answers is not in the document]
akuzm: Although given that I'm touching a lot of code in this function, it can be a good time to fix style of the comments and so on...

akuzm:
> Ok, but part of doing the review is reviewing the commit message, so given that this is not the actual commit message, this makes that part harder.

I added the commit message to the top of the PR description.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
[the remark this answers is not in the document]
mkindahl: It's not critical to not change comments or do other non-functional changes. Sometimes it is good to do it, but if we do it all the time, it will be tricky to trace down bugs and other issues later.

tsl/src/nodes/decompress_chunk/planner.c
mkindahl: It's a nit, so you decide how valuable you think it is, but imagine that `fd` or `path` happens be be NULL here. Then we have a crashing bug rather than an error message.

tsl/src/nodes/decompress_chunk/planner.c
mkindahl: Ok

tsl/src/nodes/decompress_chunk/planner.c
mkindahl: No, `const` data is not really constant, but it avoids mistakes like trying to modify values that are logically immutable in the function. It just serves as a signal to the compiler that the value is not intended to be modified. Might have avoided the issue we had with modifying partials in https://github.com/timescale/timescaledb/pull/4219, for example.

tsl/src/nodes/decompress_chunk/planner.c
mkindahl: Ok.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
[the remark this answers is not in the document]
mkindahl: > > Ok, but part of doing the review is reviewing the commit message, so given that this is not the actual commit message, this makes that part harder.
> 
> I added the commit message to the top of the PR description.

Thanks!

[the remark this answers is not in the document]
mkindahl: Sorry, I probably confused myself, Thanks for adding the comment.

mkindahl (approved):
LGTM.

tsl/src/nodes/decompress_chunk/planner.c
akuzm: I agree, recently had to fix a segfault in error handling not covered by the tests. But it was more complicated. Here we access both variables a couple of lines before, so I think this particular case is OK.

akuzm: Shouldn't we have at least a share lock on the RTEs?

akuzm: For compressed chunks we take the lock in `decompress_chunk_make_rte`. In general I think it is reasonable to assume for all RTEs we have in a query, otherwise it would be difficult to examine these tables in a concurrent environment.

PR #4235, opened 2022-04-14
Fix Coverity defect 376602

mfundul:
Fix the potential NULL pointer dereference in
ts_dimension_slice_scan_iterator_get_by_id().

### DIFF
--- a/src/dimension_slice.c
+++ b/src/dimension_slice.c
@@ -734,8 +734,8 @@ ts_dimension_slice_scan_iterator_get_by_id(ScanIterator *it, int32 slice_id,
 	ts_scan_iterator_start_or_restart_scan(it);
 	ti = ts_scan_iterator_next(it);
 	Assert(ti);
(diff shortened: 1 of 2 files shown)


src/dimension_slice.c
```diff
@@ -734,8 +734,8 @@ ts_dimension_slice_scan_iterator_get_by_id(ScanIterator *it, int32 slice_id,
 	ts_scan_iterator_start_or_restart_scan(it);
 	ti = ts_scan_iterator_next(it);
 	Assert(ti);
```
akuzm, line 736: So Coverity doesn't understand the meaning of this assert? I was thinking of marking this as a false positive...

mfundul: It's because Coverity runs against the release build.

konskov (approved):

fabriziomello (approved):

svenklemm (approved):

PR #4236, opened 2022-04-14
Set the correct collation for the compressed chunk columns

akuzm:
We don't do this currently, so some queries return the wrong ordering
of rows if there is an index on the compressed chunk.

### DIFF
--- a/tsl/src/compression/create.c
+++ b/tsl/src/compression/create.c
@@ -270,6 +270,9 @@ compresscolinfo_init(CompressColInfo *cc, Oid srctbl_relid, List *segmentby_cols
 	for (attno = 0; attno < tupdesc->natts; attno++)
 	{
 		Oid attroid = InvalidOid;
+		int32 typmod = -1;
+		Oid collid = 0;
+
 		Form_pg_attribute attr = TupleDescAttr(tupdesc, attno);
 		ColumnDef *coldef;
 		if (attr->attisdropped)
@@ -287,6 +290,8 @@ compresscolinfo_init(CompressColInfo *cc, Oid srctbl_relid, List *segmentby_cols
 			if (segorder_colindex[attno] <= seg_attnolen)
 			{
 				attroid = attr->atttypid; /*segment by columns have original type */
+				typmod = attr->atttypmod;
+				collid = attr->attcollation;
 				cc->col_meta[colno].segmentby_column_index = segorder_colindex[attno];
 			}
 			else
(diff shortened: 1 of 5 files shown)


pmwkaa (approved):

fabriziomello (approved):

tsl/src/compression/create.c
```diff
@@ -307,7 +312,7 @@ compresscolinfo_init(CompressColInfo *cc, Oid srctbl_relid, List *segmentby_cols
 		{
 			cc->col_meta[colno].algo_id = 0; // invalid algo number
 		}
-		coldef = makeColumnDef(NameStr(attr->attname), attroid, -1 /*typmod*/, 0 /*collation*/);
+		coldef = makeColumnDef(NameStr(attr->attname), attroid, typmod, collid);
```
svenklemm, line 315: wouldnt this change require a migration otherwise definitions of compressed hypertables differ before and after the change?

akuzm: They do change. I think the migration is not feasible and can't be automated (e.g. you'll have to recreate the indexes if you have any). The rest of the code should still work with the old compressed chunk definitions. So the bug is only fixed for the newly created chunks.

akuzm: And the old chunks will continue working, but they'll still have the bug.

mfundul:
Don't you need to do the same for `compresscolinfo_init_singlecolumn()`?

akuzm:
> Don't you need to do the same for `compresscolinfo_init_singlecolumn()`?

I don't know... It's used if we add a column to a chunk that is compressed already, but we can't modify it anyway. Probably I'll add it for some future-proofing, but not sure I can construct a test case for this now.

akuzm:
> > Don't you need to do the same for `compresscolinfo_init_singlecolumn()`?
> 
> I don't know... It's used if we add a column to a chunk that is compressed already, but we can't modify it anyway. Probably I'll add it for some future-proofing, but not sure I can construct a test case for this now.

Actually no, that function is only for compressed columns which have a different type and don't have a collation. The problem is with locale for the segmentby columns.

akuzm:
Not sure why codecov complains, every line of the diff should be covered.

akuzm:
`     dist_gapfill                 ... FAILED     7545 ms`
dist_gapfill is known flaky

akuzm:
`test bgw_launcher                 ... FAILED     3239 ms`

Flaky

akuzm:
`     bgw_custom                   ... FAILED   204692 ms`
Also flaky....

akuzm:
```
     dist_hypertable-14           ... FAILED     1614 ms

+ERROR:  could not close pipe to external command: Broken pipe
```
Flaky

PR #4241, opened 2022-04-17
Don't export local reorder function

svenklemm:
Change reorder_rel function to not be exported and also removes the
timescale_ function name prefix from local functions.

Disable-check: commit-count

### DIFF
--- a/tsl/src/reorder.c
+++ b/tsl/src/reorder.c
@@ -69,13 +69,13 @@
 #include "indexing.h"
 #include "reorder.h"
 
-extern void timescale_reorder_rel(Oid tableOid, Oid indexOid, bool verbose, Oid wait_id,
-								  Oid destination_tablespace, Oid index_tablespace);
+static void reorder_rel(Oid tableOid, Oid indexOid, bool verbose, Oid wait_id,
+						Oid destination_tablespace, Oid index_tablespace);
 
 #define REORDER_ACCESS_EXCLUSIVE_DEADLOCK_TIMEOUT "101000"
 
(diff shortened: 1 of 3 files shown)


konskov (approved):

tsl/src/reorder.c
```diff
@@ -419,17 +419,17 @@ chunk_get_reorder_index(Hypertable *ht, Chunk *chunk, Oid index_relid, ChunkInde
...
  * swapping the relfilenodes of the new table and the old table, so
  * the OID of the original table is preserved.
  *
  * Indexes are rebuilt in the same manner.
  */
-void
-timescale_reorder_rel(Oid tableOid, Oid indexOid, bool verbose, Oid wait_id,
-					  Oid destination_tablespace, Oid index_tablespace)
+static void
+reorder_rel(Oid tableOid, Oid indexOid, bool verbose, Oid wait_id, Oid destination_tablespace,
```
akuzm, line 431: I find myself adding prefixes instead, when there's a postgres function with the same name. I always confuse them in backtraces, switch to the wrong one in "go to definition", change the wrong one and don't understand why it doesn't work, it's really annoying :)

mfundul (approved):

tsl/src/reorder.c
svenklemm: Hmm i'm fine with either here but timescale_ is certainly the wrong prefix.

akuzm: Is it for public exports? IIRC I added `ts_` to some stuff from /import

svenklemm: we use ts_ for exported stuff, we don't really have rules for local prefixes but timescale is the company name so if we wanted to use a long prefix it should have been timescaledb_

PR #4243, opened 2022-04-18
Mark hypertable parent as dummy rel

svenklemm:
When postgres expands an inheritance tree it also adds the
parent hypertable as child relation. Since for a hypertable the
parent will never have any data we can mark this relation as
dummy relation so it gets ignored in later steps. This is only
relevant for code paths that use the postgres inheritance code
as we don't include the hypertable as child when expanding the
hypertable ourself.

Disable-check: commit-count


src/planner.c
```diff
@@ -1043,6 +1043,8 @@ timescaledb_set_rel_pathlist(PlannerInfo *root, RelOptInfo *rel, Index rti, Rang
 	{
 		case TS_REL_HYPERTABLE_CHILD:
 			/* empty child is not of interest */
+			if (root->parse->commandType == CMD_DELETE)
```
akuzm, line 1046: Should we mark it as empty for updates as well? Also I think the entire PR description should go as a comment here.

akuzm (approved):

src/planner.c
svenklemm: Yes we should treat update the same, but i got test failures when doing this for updates, so i will do it in separate step.

fabriziomello (approved):

PR #4244, opened 2022-04-18
Function telemetry

JLockerman:
This PR extends our telemetry system with function call telemetry. It gathers function call-counts from all queries, and send back counts for those functions that are built in or from our related extensions.

See [the RFC](https://docs.google.com/document/d/1wy_ma4RzCNceCDkR6HAC1NPZffdHJTOWTy4QhGPsXsY/edit#heading=h.6bq5eesegmp8) (internal link) for more details.

### DIFF
--- a/test/sql/telemetry.sql
+++ b/test/sql/telemetry.sql
@@ -210,6 +210,43 @@ SELECT :'installed_time' = :'installed_time2' AS equal, length(:'installed_time'
 
 RESET datestyle;
 
+-- test function call telemetry
+CREATE FUNCTION not_visible_in_telemetry() RETURNS INT AS $$
+    SELECT 1;
+$$ LANGUAGE SQL;
+
+-- drain old function call telemetry so we have fixed out put;
+SELECT FROM get_telemetry_report();
+
+-- call some arbirary functions
+SELECT 1 + 1, not_visible_in_telemetry(), 1 + 1, abs(-1), not_visible_in_telemetry()
+WHERE 1 + 1 = 2;
+
+-- call some aggregates
+SELECT min(not_visible_in_telemetry()), sum(not_visible_in_telemetry());
+
+-- check that we can record from a prepared statement
+PREPARE record_from_prepared AS SELECT 1 - 1;
+
+-- execute 10 times to make sure turning it into generic plan works
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+
+DEALLOCATE record_from_prepared;
+
+SELECT get_telemetry_report()->'functions_used';
+
+-- check the report again to see if resetting works
+SELECT get_telemetry_report()->'functions_used';
+
 \c :TEST_DBNAME :ROLE_SUPERUSER
 TRUNCATE _timescaledb_catalog.metadata;
 
@@ -223,3 +260,7 @@ SELECT * FROM _timescaledb_internal.test_privacy();
 
 -- To make sure nothing was sent, we check the UUID table to make sure no exported UUID row was created
 SELECT key from _timescaledb_catalog.metadata;
+
+\set ON_ERROR_STOP 0
+-- test that the telemetry gathering code doesn't break nonexistent statements
+EXECUTE noexistent_statement;
--- a/src/guc.c
+++ b/src/guc.c
@@ -19,22 +19,32 @@
 typedef enum TelemetryLevel
 {
 	TELEMETRY_OFF,
+	TELEMETRY_NO_FUNCTIONS,
 	TELEMETRY_BASIC,
 } TelemetryLevel;
 
 /* Define which level means on. We use this object to have at least one object
  * of type TelemetryLevel in the code, otherwise pgindent won't work for the
  * type */
-static const TelemetryLevel on_level = TELEMETRY_BASIC;
+static const TelemetryLevel on_level = TELEMETRY_NO_FUNCTIONS;
 
 bool
 ts_telemetry_on()
 {
-	return ts_guc_telemetry_level == on_level;
+	return ts_guc_telemetry_level >= on_level;
+}
+
+bool
+ts_function_telemetry_on()
+{
+	return ts_guc_telemetry_level > TELEMETRY_NO_FUNCTIONS;
 }
 
 static const struct config_enum_entry telemetry_level_options[] = {
-	{ "off", TELEMETRY_OFF, false }, { "basic", TELEMETRY_BASIC, false }, { NULL, 0, false }
+	{ "off", TELEMETRY_OFF, false },
+	{ "no_functions", TELEMETRY_NO_FUNCTIONS, false },
+	{ "basic", TELEMETRY_BASIC, false },
+	{ NULL, 0, false }
 };
 #endif
 
--- a/src/guc.h
+++ b/src/guc.h
@@ -12,6 +12,7 @@
 
 #ifdef USE_TELEMETRY
 extern bool ts_telemetry_on(void);
+extern bool ts_function_telemetry_on(void);
 #endif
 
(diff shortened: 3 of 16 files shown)


svenklemm (commented):
Why are you not using the infrastructure provided by postgres with `pg_stat_user_functions`. Also not a fan of putting this in the loader, changes to the loader require postgres restart.

JLockerman:
(ideally we'd use [dshash](https://github.com/postgres/postgres/blob/a62bff74b135b191de0417d1cc8ec52f144c8fe7/src/backend/lib/dshash.c) but it's not released yet)

fabriziomello:
> @svenklemm as I understand it, `pg_stat_user_functions` records the wrong thing for us (runtime, instead of calls), and imposes a decent bit more overhead to get that info, so it isn't something we'd want to mandate. While this changes the loader, it does so in a backwards compatible way: if the telemetry-gatherer see that its shared memory isn't allocated it'll simply skip gathering telemetry, and everything will keep working. The sacrifice we make is that we won't gather the new telemetry until DB restart, but this on its own shouldn't force one.

Not sure where you saw that information but Postgres already does it for us in [`src/backend/utils/adt/pgstatfuncs.c`](https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/pgstatfuncs.c). (have a look into the docs https://www.postgresql.org/docs/14/monitoring-stats.html#PG-STAT-USER-FUNCTIONS-VIEW).

```sql
fabrizio=# \d+ pg_stat_user_functions 
                        View "pg_catalog.pg_stat_user_functions"
   Column   |       Type       | Collation | Nullable | Default | Storage | Description 
------------+------------------+-----------+----------+---------+---------+-------------
 funcid     | oid              |           |          |         | plain   | 
 schemaname | name             |           |          |         | plain   | 
 funcname   | name             |           |          |         | plain   | 
 calls      | bigint           |           |          |         | plain   | 
 total_time | double precision |           |          |         | plain   | 
 self_time  | double precision |           |          |         | plain   | 
View definition:
 SELECT p.oid AS funcid,
    n.nspname AS schemaname,
    p.proname AS funcname,
    pg_stat_get_function_calls(p.oid) AS calls,
    pg_stat_get_function_total_time(p.oid) AS total_time,
    pg_stat_get_function_self_time(p.oid) AS self_time
   FROM pg_proc p
     LEFT JOIN pg_namespace n ON n.oid = p.pronamespace
  WHERE p.prolang <> 12::oid AND pg_stat_get_function_calls(p.oid) IS NOT NULL;
```

We just need to config the [`track_functions`](https://postgresqlco.nf/doc/en/param/track_functions/) GUC. And pgstat subsystem have a dedicated process named `stats collector` so I think what we're doing here is adding more overhead than the current Postgres implementation, no?

fabriziomello:
As @JLockerman Postgres 15 have a complete new implementation for old "Stats Collector" that now doesn't exist anymore (See commit https://github.com/postgres/postgres/commit/5891c7a8ed8f2d3d577e7eea34dacff12d7b6bbd).

Anyway not sure if we'll gain too much here inventing our own infrastructure to something that Postgres already does.

JLockerman:
@fabriziomello 

> Postgres already does it for us...

`track_functions` doesn't do _quite_ the same thing this does; the `calls` in `pg_stat_user_functions` is the number of times the function is _executed_ at runtime, while this PR records the number of times that the function is written by the user. An (artificial) illustrative example
```SQL
=> SELECT lots(i) FROM generate_series(1, 10000000) a(i);
<output elided>

=> SELECT little(i) FROM generate_series(1, 10000000) a(i);
<output elided>

=> select * from pg_stat_user_functions;
       funcname       |  calls   
----------------------+----------
 lots                 | 10000000

=> select get_telemetry_report()->'functions_used';
             ?column?                                                                                                                                    
------------------------------------
 {"lots(...)": 1, "little(...)": 1}

```
In other words `pg_stat_users_functions` records info about execution plans and datasets, while this PR records info about user intent. (There's also some weirdness around what things the stats collector is willing to collect I haven't bothered to dig into yet)

> And pgstat subsystem have a dedicated process named `stats collector` so I think what we're doing here is adding more overhead than the current Postgres implementation, no?

So these systems are doing different enough tasks that I'm not sure a direct performance comparison is well-founded, that said:

This PR records stats at plan time, and has overhead proportional to the length of the query string. The Postgres stats collector records info at function-execution time (in the [expression evaluator](https://github.com/postgres/postgres/blob/dec8ad367e46180f826d5b6dc820fbecba1b71d2/src/backend/executor/execExprInterp.c#L2394) among [other places](https://grep.app/search?q=pgstat_end_function_usage&filter[repo][0]=postgres/postgres)) giving it an overhead proportional to the data queries. In addition to that, the Postgres stats collector does a bunch of extra work to maintain on-disk copies of the stats so they persist after restart (the separate stats collection process was removed in the commit you linked).

My completely unscientific benchmarks seem to confirm this, with the Postgres stat collector adding something like 10x the overhead when compared to a regular function call, and this PR adding nothing to the execution time.

fabriziomello:
@JLockerman 

> `track_functions` doesn't do _quite_ the same thing this does; the `calls` in `pg_stat_user_functions` is the number of times the function is _executed_ at runtime, while this PR records the number of times that the function is written by the user. An (artificial) illustrative example
> 
> ```sql
> => SELECT lots(i) FROM generate_series(1, 10000000) a(i);
> <output elided>
> 
> => SELECT little(i) FROM generate_series(1, 10000000) a(i);
> <output elided>
> 
> => select * from pg_stat_user_functions;
>        funcname       |  calls   
> ----------------------+----------
>  lots                 | 10000000
> 
> => select get_telemetry_report()->'functions_used';
>              ?column?                                                                                                                                    
> ------------------------------------
>  {"lots(...)": 1, "little(...)": 1}
> ```
> 

Now I got your point ...

> 
> This PR records stats at plan time, and has overhead proportional to the length of the query string. The Postgres stats collector records info at function-execution time (in the [expression evaluator](https://github.com/postgres/postgres/blob/dec8ad367e46180f826d5b6dc820fbecba1b71d2/src/backend/executor/execExprInterp.c#L2394) among [other places](https://grep.app/search?q=pgstat_end_function_usage&filter%5Brepo%5D%5B0%5D=postgres/postgres)) giving it an overhead proportional to the data queries. In addition to that, the Postgres stats collector does a bunch of extra work to maintain on-disk copies of the stats so they persist after restart (the separate stats collection process was removed in the commit you linked).
> 
> My completely unscientific benchmarks seem to confirm this, with the Postgres stat collector adding something like 10x the overhead when compared to a regular function call, and this PR adding nothing to the execution time.

You're correct ... what you want from telemetry perspective is how many times a given function are used in queries and not how many times postgres call the function.

Didn't you try to get this info from `pg_stat_statements` by parsing the SQL statement?? Doing that you'll offload this overhead from our planning, no? (just another idea)

JLockerman:
> Didn't you try to get this info from pg_stat_statements by parsing the SQL statement?? Doing that you'll offload this overhead from our planning, no? (just another idea)

That's interesting, but getting the functions and types out is a rather involved process, and `pg_stat_statements` add strictly more overhead than this even if it's not gathering per-function telemetry, so I think it'd be more work for a worse user experience. (This PR is very heavily inspired by what `pg_stat_statements` does, except it strips out everything we don't need for this particular task)

svenklemm (approved):
Hmm I think the counter reset should not happen in get_telemetry_report but instead when we actually send the data to our telemetry server.  When playing around with it i was confused why this would reset the counter.

JLockerman:
@svenklemm ok, I can update it

svenklemm (approved):

fabriziomello (approved):

test/sql/telemetry.sql
```diff
@@ -210,6 +210,40 @@ SELECT :'installed_time' = :'installed_time2' AS equal, length(:'installed_time'
...
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+EXECUTE record_from_prepared;
+
+DEALLOCATE record_from_prepared;
+
+SELECT get_telemetry_report()->'functions_used';
```
jnidzwetzki, line 245: A second subsequent call of the `functions_used` function could be useful. This would test whether the function counters are reset properly.

PR #4255, opened 2022-04-20
Fix option "timescaledb.create_group_indexes"

### LINKED ISSUE #4249: [Bug]: create_group_indexes option in caggs does not work properly
### What type of bug is this?

Incorrect result

Continuous aggregate

### What happened?

Using the create_group_indexes option in caggs gives the same result whether you set the option to true or false. In both cases, it does not create group indexes.

More information in this docs issue: https://github.com/timescale/docs/issues/807

2.6.0

### PostgreSQL version used

14

whatever the default timescaledb-latest image is based on

Docker

On prem/Self-hosted

_No response_

```bash
create table test (
time timestamptz,
symbol int,
value numeric
);

select create_hypertable('test', 'time');

insert into test
(select generate_series(
	'2020-01-01',
    '2022-01-25', INTERVAL '30 sec'
  ),
  round(random()*10),
  random()*5);

 
create materialized view cagg_index_default
with (timescaledb.continuous) as 
select
	time_bucket('1 day', "time") as bucket,
	sum(value),
	symbol 
from test 
group by bucket, symbol 

create materialized view cagg_index_false
with (timescaledb.continuous, timescaledb.create_group_indexes = false) as 
select
	time_bucket('1 day', "time") as bucket,
	sum(value),
	symbol 
from test 
group by bucket, symbol 

create materialized view cagg_index_true
with (timescaledb.continuous, timescaledb.create_group_indexes = true) as 
select
	time_bucket('1 day', "time") as bucket,
	sum(value),
	symbol 
from test 
group by bucket, symbol 


select i.*
from pg_indexes i
join pg_class c
    on schemaname = relnamespace::regnamespace::text 
    and tablename = relname
where tablename like '_materialized_hypertable%'
order by tablename
    
select * from timescaledb_information.continuous_aggregates ca
```

konskov:
Previously this option was ignored when creating a
continuous aggregate, even when explicitly set to true.

Fixes #4249

### DIFF
--- a/tsl/test/expected/continuous_aggs.out
+++ b/tsl/test/expected/continuous_aggs.out
@@ -1738,3 +1738,71 @@ SELECT
 --
 (0 rows)
 
+-- test that option create_group_indexes is taken into account
+SET client_min_messages = ERROR;
+CREATE TABLE test_group_idx (
+time timestamptz,
+symbol int,
+value numeric
+);
+select create_hypertable('test_group_idx', 'time');
+      create_hypertable       
+------------------------------
+ (48,public,test_group_idx,t)
+(1 row)
+
+insert into test_group_idx
+select t, round(random()*10), random()*5
+from generate_series('2020-01-01', '2020-02-25', INTERVAL '12 hours') t;
+create materialized view cagg_index_true
+with (timescaledb.continuous, timescaledb.create_group_indexes = true) as 
+select
+	time_bucket('1 day', "time") as bucket,
+	sum(value),
+	symbol 
+from test_group_idx 
+group by bucket, symbol;
+create materialized view cagg_index_false
+with (timescaledb.continuous, timescaledb.create_group_indexes = false) as 
+select
+	time_bucket('1 day', "time") as bucket,
+	sum(value),
+	symbol 
+from test_group_idx 
+group by bucket, symbol;
+create materialized view cagg_index_default
+with (timescaledb.continuous) as 
+select
+	time_bucket('1 day', "time") as bucket,
+	sum(value),
+	symbol 
+from test_group_idx 
+group by bucket, symbol;
+-- see corresponding materialization_hypertables 
+select view_name, materialization_hypertable_name from timescaledb_information.continuous_aggregates ca
+where view_name like 'cagg_index_%';
+     view_name      | materialization_hypertable_name 
+--------------------+---------------------------------
+ cagg_index_default | _materialized_hypertable_51
+ cagg_index_false   | _materialized_hypertable_50
+ cagg_index_true    | _materialized_hypertable_49
+(3 rows)
+
+-- now make sure a group index has been created when explicitly asked for
+select i.*
+from pg_indexes i
+join pg_class c
+    on schemaname = relnamespace::regnamespace::text
+    and tablename = relname
+where tablename in (select materialization_hypertable_name from timescaledb_information.continuous_aggregates
+where view_name like 'cagg_index_%')
+order by tablename;
+      schemaname       |          tablename          |                   indexname                   | tablespace |                                                                     indexdef                                                                      
+-----------------------+-----------------------------+-----------------------------------------------+------------+---------------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


svenklemm (commented):
Dont forget to add CHANGELOG entry

tsl/test/sql/continuous_aggs.sql
```diff
@@ -1226,3 +1226,72 @@ FROM issue3248 GROUP BY 1,2;
...
+
+\gset
+
+\d+ :mat_hyper_cagg_index_true
+
+SELECT format('%I.%I', materialization_hypertable_schema,
+materialization_hypertable_name) AS mat_hyper_cagg_index_false
+FROM timescaledb_information.continuous_aggregates where view_name like 'cagg_index_false';
+
+\gset
```
mfundul, line 1289:
```suggestion
SELECT format('%I.%I', materialization_hypertable_schema,
materialization_hypertable_name) AS mat_hyper_cagg_index_false
FROM timescaledb_information.continuous_aggregates where view_name like 'cagg_index_false' \gset
```
To speed up the queries, instead of executing them twice.

tsl/test/expected/continuous_aggs.out
```diff
@@ -1738,3 +1738,107 @@ SELECT
...
+ (48,public,test_group_idx,t)
+(1 row)
+
+insert into test_group_idx
+(select generate_series(
+	'2020-01-01',
+    '2020-02-25', INTERVAL '30 sec'
+  ),
+  round(random()*10),
+  random()*5);
```
fabriziomello, line 1760:
```suggestion
insert into test_group_idx
select t, round(random()*10), random()*5
from generate_series('2020-01-01', '2020-02-25', INTERVAL '12 hours') t;
```
Why too many rows in the tests?? Do we really need it??? What about use `12 hours` so it will produce two rows for each time bucket.

tsl/test/expected/continuous_aggs.out
```diff
@@ -1738,3 +1738,107 @@ SELECT
...
+ agg_2_2  | bytea                    |           |          |         | extended |              | 
+ symbol   | integer                  |           |          |         | plain    |              | 
+ chunk_id | integer                  |           |          |         | plain    |              | 
+Indexes:
+    "_materialized_hypertable_49_bucket_idx" btree (bucket DESC)
+    "_materialized_hypertable_49_symbol_bucket_idx" btree (symbol, bucket DESC)
+Triggers:
+    ts_insert_blocker BEFORE INSERT ON _timescaledb_internal._materialized_hypertable_49 FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker()
+Child tables: _timescaledb_internal._hyper_49_106_chunk,
+              _timescaledb_internal._hyper_49_107_chunk
```
fabriziomello, line 1804: I'm always wondering about it and why we don't define a standard because there are several places using test support functions like `test.show_columns` and `test.show_columnsp` (test/sql/utils/testsupport.sql).

mfundul:
You should fix the commit message to show `Fixes #4249`. Also please update the title of the PR and the body to reflect the new commit message.

mfundul:
I think we need an extra test case. In the last test case, you create a CAGG with default parameters. We should make sure that `ALTER MATERIALIZED VIEW` also works. E.g. default => true => false

konskov:
Hmm I'm not sure I understand how to test that. According to [our documentation](https://docs.timescale.com/api/latest/continuous-aggregates/alter_materialized_view/#alter-materialized-view-continuous-aggregate) (and  implementation) setting the option `create_group_indexes` is not possible with `ALTER MATERIALIZED VIEW`, it's an option that can be set only upon cagg creation. If I try and set this option with `ALTER` I will get the error: `ERROR:  cannot alter create_group_indexes option for continuous aggregates`. So I didn't understand what the extra test case is meant to check?

mfundul (approved):

svenklemm (approved):

fabriziomello (approved):

charislam:
Thanks @konskov! 💖 Quick clarification so I can improve the docs: is the default behavior `true`?

konskov:
@charislam yes, the default behavior is `true`.

PR #4259, opened 2022-04-21
Fix logic bug in init_privs query

svenklemm:
The query to get the list of saved privileges during extension
upgrade had a bug and only applying the classoid restriction
for a subset of the entries when it should have been applied to
all returned rows leading to a failure during extension update
when init privileges for other classoids existed on any of the
relevant objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ accidentally triggering the load of a previous DB version.**
 * #4209 Add support for chunk exclusion on UPDATE to PG14
 
 **Bugfixes**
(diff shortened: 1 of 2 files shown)


akuzm (approved):
What changed? I see you moved a condition from WHERE to ON, but for inner joins it shouldn't matter.

Probably we can construct a test for this.

svenklemm:
> What changed? I see you moved a condition from WHERE to ON, but for inner joins it shouldn't matter.
> 
> Probably we can construct a test for this.

It moves the classoid to the ANDed join conditions, previously it was ORed and wasn't always applied

jnidzwetzki (approved):

PR #4261, opened 2022-04-21
Add pgspot to CI

svenklemm:
Add pgspot to check our installation scripts for following best
practices.


.github/workflows/pgspot.yaml
```diff
@@ -0,0 +1,49 @@
+name: pgspot
+on:
+  pull_request:
+
```
jnidzwetzki, line 4: It might also be useful to run the workflow as well when changes are made to the `prerelease_test` branch (e.g., like the `linux-build-and-test.yaml` workflow)

```suggestion
  push:
    branches:
      - prerelease_test
```

fabriziomello (approved):

jnidzwetzki (approved):

PR #4269, opened 2022-04-22
Continuous Aggregates finals form

### LINKED ISSUE #4233: [Bug]: Bad Continuous Aggregate View Definition
### What type of bug is this?

Data corruption

Continuous aggregate

### What happened?

Creating a Continuous Aggregate where the `time_bucket()` function is not the top level expression of the `SELECT` target generates a bad view which disallows the timescaledb upgrades to take place.

2.6.1

### PostgreSQL version used

14.1

Ubuntu 20.04

Source

On prem/Self-hosted

_No response_

_Originally posted by @hardikm10 :

I think I was able to reproduce the error and tried to do an upgrade (even from 11 to 12) and it resulted in the same error. Here are the instructions to generate a similar C-agg and then on upgrade, error is thrown:

```sql
drop table sensor_data cascade;

create table sensor_data(
time timestamp not null,
sensor_id integer not null,
cpu double precision null,

--on 2.x
select from create_hypertable(relation=>'sensor_data', time_column_name=> 'time');

INSERT INTO sensor_data
SELECT
sensor_id,
random() AS cpu,
random()* 100 AS temperature
FROM
generate_series(now() - INTERVAL '1 months', now() - INTERVAL '1 week', INTERVAL '100 minute') AS g1(time),
ORDER BY
time;

\d+ sensor_data

DROP MATERIALIZED VIEW sensor_summary_daily CASCADE;

--2.x
create materialized view sensor_summary_daily with (timescaledb.continuous)            
as select max(cpu) as max_cpu, max(temperature) as max_temp, 
            sensor_id,     time_bucket(INTERVAL '1 minute', time) + '1 minute' AS time
 from sensor_data group by sensor_id, time_bucket(INTERVAL '1 minute', time) ;

defaultdb=> \d+ sensor_summary_daily
                                View "public.sensor_summary_daily"
  Column   |            Type             | Collation | Nullable | Default | Storage | Description
-----------+-----------------------------+-----------+----------+---------+---------+-------------
 max_cpu   | double precision            |           |          |         | plain   |
 max_temp  | double precision            |           |          |         | plain   |
 sensor_id | integer                     |           |          |         | plain   |
 time      | timestamp without time zone |           |          |         | plain   |
View definition:
 SELECT _timescaledb_internal.finalize_agg('pg_catalog.max(double precision)'::text, NULL::name, NULL::name, '{{pg_catalog,float8}}'::name[], _materialized_hypertable_30.agg_1_1, NULL::double precision) AS max_cpu,
    _timescaledb_internal.finalize_agg('pg_catalog.max(double precision)'::text, NULL::name, NULL::name, '{{pg_catalog,float8}}'::name[], _materialized_hypertable_30.agg_2_2, NULL::double precision) AS max_temp,
    _materialized_hypertable_30.sensor_id,
    _materialized_hypertable_30."time"
   FROM _timescaledb_internal._materialized_hypertable_30
  WHERE _materialized_hypertable_30.time_partition_col < COALESCE(_timescaledb_internal.to_timestamp_without_timezone(_timescaledb_internal.cagg_watermark(30)), '-infinity'::timestamp without time zone)
  GROUP BY _materialized_hypertable_30.sensor_id, _materialized_hypertable_30.time_partition_col
UNION ALL
 SELECT max(sensor_data.cpu) AS max_cpu,
    max(sensor_data.temperature) AS max_temp,
    sensor_data.sensor_id,
    time_bucket('00:01:00'::interval, sensor_data."time") + '00:01:00'::interval AS "time"
   FROM sensor_data
  WHERE sensor_data."time" >= COALESCE(_timescaledb_internal.to_timestamp_without_timezone(_timescaledb_internal.cagg_watermark(30)), '-infinity'::timestamp without time zone)
  GROUP BY sensor_data.sensor_id, (time_bucket('00:01:00'::interval, sensor_data."time"));

```

Notice the `time_bucket(INTERVAL '1 minute', time) + '1 minute' AS time` in the C-agg defintion. It is to be noted that I don't have the definition of the c-agg yet, so they could be running into this via something different.

fabriziomello:
Following work started by https://github.com/timescale/timescaledb/pull/4294 to improve performance of Continuous
Aggregates by removing the re-aggregation in the user view.

This PR get rid of `partialize_agg` and `finalize_agg` aggregate
functions and store the finalized aggregated (plain) data in the
materialization hypertable.

Because we're not storing partials anymore and removed the
re-aggregation, now is be possible to create indexes on aggregated
columns in the materialization hypertable in order to improve the
performance even more.

Also removed restrictions on types of aggregates users can perform
with Continuous Aggregates:
* aggregates with DISTINCT
* aggregates with FILTER
* aggregates with FILTER in HAVING clause
* aggregates without combine function
* ordered-set aggregates
* hypothetical-set aggregates

By default new Continuous Aggregates will be created using this new
format, but the previous version (with partials) will be supported.

Users can create the previous style by setting to `false` the storage
paramater named `timescaledb.finalized` during the creation of the
Continuous Aggregate.

Fixes #4233

### DIFF
--- a/test/sql/updates/setup.continuous_aggs.v2.sql
+++ b/test/sql/updates/setup.continuous_aggs.v2.sql
@@ -12,7 +12,8 @@ SELECT
 	extversion < '2.0.0' AS has_ignore_invalidations_older_than,
 	extversion < '2.0.0' AS has_max_interval_per_job,
 	extversion >= '2.0.0' AS has_create_mat_view,
-	extversion >= '2.0.0' AS has_continuous_aggs_policy
+	extversion >= '2.0.0' AS has_continuous_aggs_policy,
+	extversion >= '2.7.0' AS has_continuous_aggs_finals_form
   FROM pg_extension
  WHERE extname = 'timescaledb' \gset
 
@@ -54,7 +55,11 @@ SELECT generate_series('2018-11-01 00:00'::timestamp, '2018-12-15 00:00'::timest
     -- we had a bug related to that and need to verify if compression can be
     -- enabled on such a view
     CREATE MATERIALIZED VIEW rename_cols
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=false, timescaledb.finalized=false) AS
+    \else
     WITH (timescaledb.continuous, timescaledb.materialized_only = false) AS
+    \endif
 \endif
     SELECT time_bucket('1 week', timec) AS bucket, location, round(avg(humidity)) AS humidity
     FROM conditions_before
@@ -71,7 +76,11 @@ SELECT generate_series('2018-11-01 00:00'::timestamp, '2018-12-15 00:00'::timest
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.refresh_lag='-30 day', timescaledb.max_interval_per_job ='1000 day')
 \else
     CREATE MATERIALIZED VIEW IF NOT EXISTS mat_before
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.finalized=false)
+    \else
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true)
+    \endif
 \endif
     AS
       SELECT time_bucket('1week', timec) as bucket,
@@ -147,7 +156,11 @@ CREATE SCHEMA cagg;
     WITH ( timescaledb.continuous, timescaledb.materialized_only=false, timescaledb.refresh_lag='-30 day', timescaledb.max_interval_per_job ='1000 day')
 \else
     CREATE MATERIALIZED VIEW IF NOT EXISTS cagg.realtime_mat
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=false, timescaledb.finalized=false)
+    \else
     WITH ( timescaledb.continuous, timescaledb.materialized_only=false)
+    \endif
 \endif
     AS
       SELECT time_bucket('1week', timec) as bucket,
@@ -215,7 +228,11 @@ CALL refresh_continuous_aggregate('cagg.realtime_mat',NULL,NULL);
            timescaledb.max_interval_per_job = '100000 days')
 \else
     CREATE MATERIALIZED VIEW IF NOT EXISTS  mat_ignoreinval
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.finalized=false)
+    \else
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true)
+    \endif
 \endif
     AS
       SELECT time_bucket('1 week', timec) as bucket,
@@ -252,7 +269,11 @@ SELECT generate_series('2018-12-01 00:00'::timestamp, '2018-12-20 00:00'::timest
            timescaledb.max_interval_per_job='100000 days' )
 \else
     CREATE MATERIALIZED VIEW mat_inval
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.finalized=false)
+    \else
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true )
+    \endif
 \endif
     AS
       SELECT time_bucket('10 minute', time) as bucket, location, min(temperature) as min_temp,
@@ -296,7 +317,11 @@ INSERT INTO int_time_test VALUES
            timescaledb.refresh_interval='12 hours')
 \else
     CREATE MATERIALIZED VIEW mat_inttime
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.finalized=false)
+    \else
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true )
+    \endif
 \endif
     AS
       SELECT time_bucket( 2, timeval), COUNT(col1)
@@ -314,7 +339,11 @@ INSERT INTO int_time_test VALUES
            timescaledb.refresh_interval='12 hours')
 \else
     CREATE MATERIALIZED VIEW mat_inttime2
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.finalized=false)
+    \else
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true )
+    \endif
 \endif
     AS
       SELECT time_bucket( 2, timeval), COUNT(col1)
@@ -348,7 +377,11 @@ SELECT create_hypertable('conflict_test', 'time', chunk_time_interval => INTERVA
            timescaledb.refresh_interval='12 hours' )
 \else
     CREATE MATERIALIZED VIEW mat_conflict
+    \if :has_continuous_aggs_finals_form
+    WITH (timescaledb.continuous, timescaledb.materialized_only=true, timescaledb.finalized=false)
+    \else
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true )
+    \endif
 \endif
     AS
       SELECT time_bucket('10 minute', time) as bucket, location, min(temperature) as min_temp,
@@ -407,6 +440,9 @@ WITH (
 \if :has_max_interval_per_job
      timescaledb.refresh_lag='-30 day',
      timescaledb.max_interval_per_job ='1000 day',
+\endif
+\if :has_continuous_aggs_finals_form
+     timescaledb.finalized = false,
 \endif
      timescaledb.continuous
 ) AS
--- a/tsl/src/continuous_aggs/create.c
+++ b/tsl/src/continuous_aggs/create.c
@@ -165,9 +165,10 @@ typedef struct MatTableColumnInfo
 
 typedef struct FinalizeQueryInfo
 {
-	List *final_seltlist;   /*select target list for finalize query */
-	Node *final_havingqual; /*having qual for finalize query */
+	List *final_seltlist;   /* select target list for finalize query */
+	Node *final_havingqual; /* having qual for finalize query */
 	Query *final_userquery; /* user query used to compute the finalize_query */
+	bool finalized;			/* finalized form? */
 } FinalizeQueryInfo;
 
 typedef struct CAggTimebucketInfo
@@ -210,10 +211,10 @@ typedef struct AggPartCxt
 } AggPartCxt;
 
 /* STATIC functions defined on the structs above */
-static void mattablecolumninfo_init(MatTableColumnInfo *matcolinfo, List *collist, List *tlist,
-									List *grouplist);
+static void mattablecolumninfo_init(MatTableColumnInfo *matcolinfo, List *grouplist);
 static Var *mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input,
-										int original_query_resno);
+										int original_query_resno, bool finalized,
+										bool *skip_adding);
 static void mattablecolumninfo_addinternal(MatTableColumnInfo *matcolinfo,
 										   RangeTblEntry *usertbl_rte, int32 usertbl_htid);
 static int32 mattablecolumninfo_create_materialization_table(
@@ -221,7 +222,7 @@ static int32 mattablecolumninfo_create_materialization_table(
 	CAggTimebucketInfo *origquery_tblinfo, bool create_addl_index, char *tablespacename,
 	char *table_access_method, ObjectAddress *mataddress);
 static Query *mattablecolumninfo_get_partial_select_query(MatTableColumnInfo *matcolinfo,
-														  Query *userview_query);
+														  Query *userview_query, bool finalized);
 
 static void caggtimebucketinfo_init(CAggTimebucketInfo *src, int32 hypertable_id,
 									Oid hypertable_oid, AttrNumber hypertable_partition_colno,
@@ -620,16 +621,29 @@ mattablecolumninfo_create_materialization_table(MatTableColumnInfo *matcolinfo, 
  * the materialization columns and remove HAVING clause and ORDER BY
  */
 static Query *
-mattablecolumninfo_get_partial_select_query(MatTableColumnInfo *mattblinfo, Query *userview_query)
+mattablecolumninfo_get_partial_select_query(MatTableColumnInfo *mattblinfo, Query *userview_query,
+											bool finalized)
 {
 	Query *partial_selquery;
+
 	CAGG_MAKEQUERY(partial_selquery, userview_query);
 	partial_selquery->rtable = copyObject(userview_query->rtable);
 	partial_selquery->jointree = copyObject(userview_query->jointree);
+
 	partial_selquery->targetList = mattblinfo->partial_seltlist;
 	partial_selquery->groupClause = mattblinfo->partial_grouplist;
-	partial_selquery->havingQual = NULL;
-	partial_selquery->sortClause = NULL;
+
+	if (finalized)
+	{
+		partial_selquery->havingQual = copyObject(userview_query->havingQual);
+		partial_selquery->sortClause = copyObject(userview_query->sortClause);
+	}
+	else
+	{
+		partial_selquery->havingQual = NULL;
+		partial_selquery->sortClause = NULL;
+	}
+
 	return partial_selquery;
 }
 
@@ -1088,7 +1102,7 @@ cagg_query_supported(Query *query, StringInfo hint, StringInfo detail)
 }
 
 static CAggTimebucketInfo
-cagg_validate_query(Query *query)
+cagg_validate_query(Query *query, bool finalized)
 {
 	CAggTimebucketInfo ret;
 	Cache *hcache;
@@ -1108,9 +1122,13 @@ cagg_validate_query(Query *query)
 				 detail->len > 0 ? errdetail("%s", detail->data) : 0));
 	}
 
-	/*validate aggregates allowed */
-	cagg_agg_validate((Node *) query->targetList, NULL);
-	cagg_agg_validate((Node *) query->havingQual, NULL);
+	/* finalized cagg doesn't have those restrictions anymore */
+	if (!finalized)
+	{
+		/* validate aggregates allowed */
+		cagg_agg_validate((Node *) query->targetList, NULL);
+		cagg_agg_validate((Node *) query->havingQual, NULL);
+	}
 
 	fromList = query->jointree->fromlist;
 	if (list_length(fromList) != 1 || !IsA(linitial(fromList), RangeTblRef))
@@ -1444,12 +1462,14 @@ function_allowed_in_cagg_definition(Oid funcid)
 	return finfo->allowed_in_cagg_definition;
 }
 
-/*initialize MatTableColumnInfo */
+/*
+ * Initialize MatTableColumnInfo
+ */
 static void
-mattablecolumninfo_init(MatTableColumnInfo *matcolinfo, List *collist, List *tlist, List *grouplist)
+mattablecolumninfo_init(MatTableColumnInfo *matcolinfo, List *grouplist)
 {
-	matcolinfo->matcollist = collist;
-	matcolinfo->partial_seltlist = tlist;
+	matcolinfo->matcollist = NIL;
+	matcolinfo->partial_seltlist = NIL;
 	matcolinfo->partial_grouplist = grouplist;
 	matcolinfo->mat_groupcolname_list = NIL;
 	matcolinfo->matpartcolno = -1;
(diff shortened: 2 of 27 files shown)


mkindahl (commented):
Please remove the backquotes from the title line. I do not think it is a big issue, but we decided to not use those in the title line, while using in the body is fine.

fabriziomello:
> Please remove the backquotes from the title line. I do not think it is a big issue, but we decided to not use those in the title line, while using in the body is fine.

Removed, thanks!

mfundul:
As discussed, we should remember to add a CAGG version field to the telemetry data.

tsl/src/continuous_aggs/create.c
```diff
@@ -1529,27 +1560,41 @@ mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input, int original_q
...
+			/* keep original resjunk if finalized or not time bucket */
+			if (!finalized || timebkt_chk)
+			{
+				/*
+				 * Need to project all the partial entries so that
+				 * materialization table is filled
+				 */
+				part_te->resjunk = false;
+			}
+
```
mfundul, line 1583: This check of the time-bucketing functions above:
```
			if (IsA(tle->expr, FuncExpr))
timebkt_chk = function_allowed_in_cagg_definition(((FuncExpr *) tle->expr)->funcid);
```
is not correct and has never been correct.

Some things wrong with it:
  - `time_partition_col` is the default name given to those columns. This assumes it's only one column using a time-bucketing function in the target list, which may not be the case.
  - it only checks top-level function expressions, so if the function is hidden within an expression it is never checked and no special handling will occur.

E.g. this statement potentially breaks the code in various places:

```
CREATE MATERIALIZED VIEW conditions_summary
    WITH (timescaledb.continuous, timescaledb.materialized_only=true)
    AS SELECT
        time_bucket('1 days', time) AT TIME ZONE 'UTC',
        time_bucket('1 days', time) AT TIME ZONE 'UTC' + '1 day',
        avg(temperature)
        FROM conditions
        GROUP BY device_id, time_bucket('1 days', time)  WITH NO DATA;
```

mfundul (changes requested):

mkindahl (commented):
First batch of comment: I'll continue later but thought that you might want to look at these.

Nit: You are making a lot of white-space and formatting changes, making reviewing the patch a lot harder since it is hard to filter out significant changes from non-significant changes. Please avoid this (no need to change anything in this pull request, I'll review it as it is).

Apart from the comments in-line, I'm curious about these:

- We have a very low coverage in `cagg_rename_view_columns`, `tsl_cagg_try_repair`, `var_already_mapped`, and `cagg_rebuild_view_definition`. Was this the case before, or is it a result of the changed format? If we have the code there, it would make sense to have tests that it is working as expected since we can assume we do not know if code that is not tested actually works.
-

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -44,287 +45,320 @@ SELECT generate_series('2018-11-01 00:00'::timestamp, '2018-12-31 00:00'::timest
...
+      covar_samp(temperature, humidity),
+      regr_avgx(temperature, humidity),
+      regr_avgy(temperature, humidity),
+      regr_count(temperature, humidity),
+      regr_intercept(temperature, humidity),
+      regr_r2(temperature, humidity),
+      regr_slope(temperature, humidity),
+      regr_sxx(temperature, humidity),
+      regr_sxy(temperature, humidity),
+      round(regr_syy(temperature, humidity)) as regr_syy,
```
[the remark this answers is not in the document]
fabriziomello: You're correct. Added round to all of them.

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -44,287 +45,320 @@ SELECT generate_series('2018-11-01 00:00'::timestamp, '2018-12-31 00:00'::timest
...
 \if :has_refresh_mat_view
-REFRESH MATERIALIZED VIEW rename_cols;
-REFRESH MATERIALIZED VIEW mat_before;
+  REFRESH MATERIALIZED VIEW rename_cols;
+  REFRESH MATERIALIZED VIEW mat_before;
 \else
-CALL refresh_continuous_aggregate('rename_cols',NULL,NULL);
-CALL refresh_continuous_aggregate('mat_before',NULL,NULL);
+  CALL refresh_continuous_aggregate('rename_cols',NULL,NULL);
+  CALL refresh_continuous_aggregate('mat_before',NULL,NULL);
```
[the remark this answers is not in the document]
fabriziomello: Fixed

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -44,287 +45,320 @@ SELECT generate_series('2018-11-01 00:00'::timestamp, '2018-12-31 00:00'::timest
...
+      regr_count(temperature, humidity),
+      regr_intercept(temperature, humidity),
+      regr_r2(temperature, humidity),
+      regr_slope(temperature, humidity),
+      regr_sxx(temperature, humidity),
+      regr_sxy(temperature, humidity),
+      round(regr_syy(temperature, humidity)) as regr_syy,
+      stddev(temperature) as stddev_temp,
+      round(stddev_pop(temperature)) as stddev_pop,
+      stddev_samp(temperature),
```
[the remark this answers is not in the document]
fabriziomello: Fixed

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -44,287 +45,320 @@ SELECT generate_series('2018-11-01 00:00'::timestamp, '2018-12-31 00:00'::timest
...
-DO LANGUAGE PLPGSQL $$
-DECLARE
-  ts_version TEXT;
-BEGIN
-  SELECT extversion INTO ts_version FROM pg_extension WHERE extname = 'timescaledb';
-  IF ts_version < '2.0.0' THEN
-    CREATE VIEW rename_cols
-    WITH (timescaledb.continuous, timescaledb.materialized_only = false, timescaledb.refresh_lag='14 days') AS
-    SELECT time_bucket('1 week', timec) AS bucket, location, avg(humidity)
+\if :has_create_mat_view
```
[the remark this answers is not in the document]
fabriziomello: It's necessary because I introduced a new psql variable named `has_continuous_aggs_finals_form` and it don't work properly inside anonymous blocks.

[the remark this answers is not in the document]
mkindahl: It does not matter to me, unless you're planning to make more changes to the code before pushing.

tsl/test/sql/continuous_aggs_deprecated.sql
```diff
@@ -1230,3 +1230,59 @@ FROM issue3248 GROUP BY 1,2;
...
+from generate_series('2020-01-01', '2020-02-25', INTERVAL '12 hours') t;
+
+create materialized view cagg_index_true
+with (timescaledb.continuous, timescaledb.create_group_indexes=true, timescaledb.finalized=false) as
+select
+	time_bucket('1 day', "time") as bucket,
+	sum(value),
+	symbol
+from test_group_idx
+group by bucket, symbol;
```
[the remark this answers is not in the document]
fabriziomello: I'm already doing that cause what I did is basically copied the `continuous_aggs.sql` to `continuous_aggs_deprecated.sql` and added the `timescaledb.finalized=false` to caggs definition, so in practice we're testing both cases already.

tsl/test/sql/continuous_aggs_deprecated.sql
```diff
@@ -1230,3 +1230,59 @@ FROM issue3248 GROUP BY 1,2;
...
+
+create materialized view cagg_index_default
+with (timescaledb.continuous, timescaledb.finalized=false) as
+select
+	time_bucket('1 day', "time") as bucket,
+	sum(value),
+	symbol
+from test_group_idx
+group by bucket, symbol;
```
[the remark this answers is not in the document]
fabriziomello: This test was introduced by https://github.com/timescale/timescaledb/commit/687e7c7233d898c08aa3b1a461b0b7296beb0a3e. Should I change it?

tsl/test/sql/continuous_aggs_deprecated.sql
```diff
@@ -1230,3 +1230,59 @@ FROM issue3248 GROUP BY 1,2;
 SELECT
   FROM issue3248 AS m,
        LATERAL(SELECT m FROM issue3248_cagg WHERE avg IS NULL LIMIT 1) AS lat;
+
+-- test that option create_group_indexes is taken into account
+SET client_min_messages = ERROR;
```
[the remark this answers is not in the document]
fabriziomello: This test was introduced by https://github.com/timescale/timescaledb/commit/687e7c7233d898c08aa3b1a461b0b7296beb0a3e. Should I change it?

tsl/test/sql/continuous_aggs_deprecated.sql
```diff
@@ -1230,3 +1230,59 @@ FROM issue3248 GROUP BY 1,2;
...
+
+-- now make sure a group index has been created when explicitly asked for
+select i.*
+from pg_indexes i
+join pg_class c
+    on schemaname = relnamespace::regnamespace::text
+    and tablename = relname
+where tablename in (select materialization_hypertable_name from timescaledb_information.continuous_aggregates
+where view_name like 'cagg_index_%')
+order by tablename;
```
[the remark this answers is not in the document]
fabriziomello: This test was introduced by https://github.com/timescale/timescaledb/commit/687e7c7233d898c08aa3b1a461b0b7296beb0a3e. Should I change it?

tsl/test/expected/continuous_aggs.out
```diff
@@ -68,7 +68,7 @@ INNER JOIN _timescaledb_catalog.hypertable h ON(h.id = ca.mat_hypertable_id)
 WHERE user_view_name = 'mat_m1'
 \gset
 insert into :"MAT_SCHEMA_NAME".:"MAT_TABLE_NAME"
-select a, _timescaledb_internal.partialize_agg(count(b)),
+select a, count(b),
```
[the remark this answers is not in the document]
fabriziomello: It's already tested in `continuous_aggs_deprecated.sql`

tsl/test/sql/continuous_aggs_deprecated.sql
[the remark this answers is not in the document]
mfundul: Mats meant test them at the same time, concurrently. Like having both new and old style aggregates co-existing in space and time.

mkindahl (commented):
A few minor comments. Otherwise, I think it looks fine. Please fix these:

- The assert issue, so that we do not have strange errors in release builds.
- Have a test that uses both the new and old format at the same time and both should work without problems.

tsl/src/continuous_aggs/create.c
mfundul: @fabriziomello let me take care of it in another PR.

test/sql/updates/setup.continuous_aggs.v2.sql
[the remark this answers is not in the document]
fabriziomello: Fixed by https://github.com/timescale/timescaledb/pull/4324

tsl/src/continuous_aggs/create.c
fabriziomello: @mfundul Ok. let's create another issue and then discuss more there

tsl/src/continuous_aggs/create.c
```diff
@@ -1568,15 +1613,26 @@ mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input, int original_q
...
 			elog(ERROR, "invalid node type %d", nodeTag(input));
 			break;
 	}
-	Assert(list_length(out->matcollist) == list_length(out->partial_seltlist));
+
+	if (!finalized)
+		Assert(list_length(out->matcollist) == list_length(out->partial_seltlist));
+	else
+		Assert(list_length(out->matcollist) <= list_length(out->partial_seltlist));
 	Assert(col != NULL);
```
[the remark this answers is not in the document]
fabriziomello: Fixed!

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -291,35 +220,31 @@ CALL refresh_continuous_aggregate('cagg.realtime_mat',NULL,NULL);
...
     WITH ( timescaledb.continuous, timescaledb.materialized_only=true)
+    \endif
+\endif
     AS
       SELECT time_bucket('1 week', timec) as bucket,
     max(temperature) as maxtemp
       FROM conditions_before
+\if :has_refresh_mat_view
+      GROUP BY bucket;
+\else
```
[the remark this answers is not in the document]
fabriziomello: This is for the old version of caggs using `CREATE VIEW` instead of `CREATE MATERIALIZED VIEW`

test/sql/updates/setup.continuous_aggs.v2.sql
```diff
@@ -336,36 +261,31 @@ SELECT generate_series('2018-12-01 00:00'::timestamp, '2018-12-20 00:00'::timest
...
+    \endif
+\endif
     AS
       SELECT time_bucket('10 minute', time) as bucket, location, min(temperature) as min_temp,
-        max(temperature) as max_temp, avg(temperature) as avg_temp
+        max(temperature) as max_temp, round(avg(temperature)) as avg_temp
       FROM inval_test
+\if :has_refresh_mat_view
+      GROUP BY bucket, location;
+\else
```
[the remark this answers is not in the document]
fabriziomello: This is for the old version of caggs using `CREATE VIEW` instead of `CREATE MATERIALIZED VIEW`

tsl/test/sql/continuous_aggs_deprecated.sql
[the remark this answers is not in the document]
fabriziomello: You're correct it's not necessary. Removed!

tsl/test/sql/continuous_aggs_deprecated.sql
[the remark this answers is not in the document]
fabriziomello: Added new tests for this at the end of the `continuous_caggs.sql`.

tsl/test/sql/continuous_aggs_deprecated.sql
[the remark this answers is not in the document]
fabriziomello: Added new tests for this at the end of the continuous_caggs.sql.

mkindahl (approved):
LGTM

tsl/src/continuous_aggs/create.c
```diff
@@ -1518,6 +1543,12 @@ mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input, int original_q
 				{
 					PRINT_MATCOLNAME(colbuf, "grp", original_query_resno, matcolno);
 					colname = colbuf;
+
+					/* for finalized form we skip those columns */
+					if (finalized)
+					{
+						*skip_adding = true;
+					}
 				}
```
mfundul, line 1552: What's the assumption here? Is it :
```
Assert(tle->resname == NULL && tle->ressortgroupref > 0 && tle->resjunk == true);
```
Are we only aiming at group-by targets here?

tsl/src/continuous_aggs/create.c
```diff
@@ -1529,27 +1560,41 @@ mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input, int original_q
 			}
 			else
 			{
-				out->mat_groupcolname_list = lappend(out->mat_groupcolname_list, pstrdup(colname));
+				if (!*skip_adding && ((TargetEntry *) input)->ressortgroupref > 0)
+					out->mat_groupcolname_list =
+						lappend(out->mat_groupcolname_list, pstrdup(colname));
```
mfundul, line 1565: Does this do anything for the finalized form, other than adding the time_bucket column?

fabriziomello: The `mat_groupcolname_list` list is used for adding indexes to the materialization hypertable.

tsl/src/continuous_aggs/create.c
fabriziomello: The finals form does not require those extra-group-by columns in the materialization hypertable

mfundul: I think I understand more now, there are some cases which we should document, and see if we're missing something if anything.

- `resname != NULL` means it's projected in our case
- `ressortgroupref > 0` means part of GROUP BY, which can be projected or not, depending on whether the column was deduplicated in the target list
- `resjunk == true` applies for GROUP BY columns that are not projected

So, in this scenario:
```
CREATE MATERIALIZED VIEW conditions_summary
    WITH (timescaledb.continuous, timescaledb.materialized_only=true)
AS SELECT
    time_bucket('1 days', time) AT TIME ZONE 'UTC',
    device_id,
    avg(temperature)
FROM conditions
GROUP BY device_id, time_bucket('1 days', time)  WITH NO DATA;
```
we get an index on `time_bucket()` regardless, and we get an index on `device_id` because it's deduplicated in the target-list and merged, as a member of both the GROUP BY and the projection.

However in this case:
```
CREATE MATERIALIZED VIEW conditions_summary
    WITH (timescaledb.continuous, timescaledb.materialized_only=true)
AS SELECT
    time_bucket('1 days', time) AT TIME ZONE 'UTC',
    device_id + 1,
    avg(temperature)
FROM conditions
GROUP BY device_id, time_bucket('1 days', time)  WITH NO DATA;
```
We never get the index on `device_id`.

fabriziomello: Awesome @mfundul ... your explanation here summarize everything.

tsl/test/expected/cagg_errors.out
```diff
@@ -158,33 +129,14 @@ from conditions
...
-   initcond1 = '{0,0}'
-);
-CREATE MATERIALIZED VIEW mat_m1 WITH ( timescaledb.continuous)
-AS
-Select sum(humidity), newavg(temperature::int4)
-from conditions
- group by time_bucket('1week', timec) , location WITH NO DATA;
-ERROR:  aggregates which are not parallelizable are not supported
+ERROR:  only immutable functions supported in continuous aggregate view
+HINT:  Make sure the function includes only immutable expressions, e.g., time_bucket('1 hour', time AT TIME ZONE 'GMT').
```
davidkohn88, line 139: Is this testing what we want it to test or is the time_bucket the problem?

davidkohn88: Maybe we can have a version at the top of the file that demonstrates that this basic view definition does work, ie just count(*) or the like in this file, so that we know the normal stuff is valid and the time_bucket is valid etc in this file?

davidkohn88 (commented):
Took a cursory look, really excited about this and can't wait to have it. One test / error message stood out as slightly confusing in my cursory glance, might just mean we should make the error message more clear or maybe distinguish between the case where the group by has a non-immutable function vs the agg itself.

tsl/test/expected/cagg_errors.out
fabriziomello: > Is this testing what we want it to test or is the time_bucket the problem?

This is exact what we want to test... `json_agg` is stable:

```sql
fabrizio=# \df+ json_agg
List of functions
-[ RECORD 1 ]-------+--------------------------
Schema              | pg_catalog
Name                | json_agg
Result data type    | json
Argument data types | anyelement
Type                | agg
Volatility          | stable
Parallel            | safe
Owner               | fabrizio
Security            | invoker
Access privileges   | 
Language            | internal
Source code         | aggregate_dummy
Description         | aggregate input into json
```

davidkohn88: Ah. I think my confusion is because of the error message - we should, if possible, distinguish between stable things in the group by ie `time_bucket` and stable things as aggregates, so we can have different error messages for each so as not to confuse users. Or we could add a line to the error message with another example so folks aren't drawn only to the time_bucket. 

I think my problem is really with the hint.

HINT:  Make sure the function includes only immutable expressions, e.g., time_bucket('1 hour', time AT TIME ZONE 'GMT').

alternative: Make sure the view definition contains only IMMUTABLE expressions for aggregates and GROUP BYs.

davidkohn88: The current one makes it seem like I need to add a time zone to my time bucket and I think that's just confusing.
we might want to point them at the function volatility stuff. I just updated mine and I think it's pretty good? But I dunno.

davidkohn88: Or perhaps we could add a hint to run \df+ [function_name] in psql and look at the "volatility" field?

fabriziomello: > Or perhaps we could add a hint to run \df+ [function_name] in psql and look at the "volatility" field?

I think is not good show psql commands hints for users because they can use another tool to connect to postgres

fabriziomello: > alternative: Make sure the view definition contains no functions marked STABLE or VOLATILE in aggregates and GROUP BYs. Note that functions may be IMMUTABLE for one data type, but only STABLE for another.

@davidkohn88 what about?

Make sure the view definition includes only IMMUTABLE expressions. Note that functions or expressions may be IMMUTABLE for one data type, but only STABLE for another.

davidkohn88: Slight revision: Make sure the continuous aggregate definition includes only IMMUTABLE expressions. Note that functions or expressions may be IMMUTABLE for one data type, but STABLE  or VOLATILE for another.

davidkohn88: I will say that the first sentence feels redundant given the error message: only immutable functions supported in continuous aggregate view


Maybe something like: 
ERROR:  only immutable functions supported in continuous aggregate view
HINT: Make sure all functions in the continuous aggregate definition have IMMUTABLE volatility. Note that functions or expressions may be IMMUTABLE for one data type, but STABLE or VOLATILE for another.

That at least points them on how to look up the docs for volatility?

mfundul:
We should probably add https://github.com/timescale/timescaledb/issues/4233 to the changelog, and add a `#Fixes 4233` in the commit message.

tsl/src/continuous_aggs/create.c
```diff
@@ -1529,27 +1564,46 @@ mattablecolumninfo_addentry(MatTableColumnInfo *out, Node *input, int original_q
...
 			if (part_te->resname == NULL)
 			{
 				part_te->resname = pstrdup(colname);
 			}
 		}
 		break;
+
 		case T_Var:
 		{
 			PRINT_MATCOLNAME(colbuf, "var", original_query_resno, matcolno);
```
mfundul, line 1609: I think at this point we should remove this line completely:
```
1612    out->mat_groupcolname_list = lappend(out->mat_groupcolname_list, pstrdup(colname));
```

tsl/test/expected/cagg_errors.out
fabriziomello: Done!

fabriziomello:
> We should probably add this PR to the changelog as a fix of #4233, and add a `#Fixes 4233` in the commit message.

Done!

tsl/src/continuous_aggs/create.c
fabriziomello: Why? This is part of the old view code, no?

mfundul: This code was introduced between releases, it's not technically part of the old code. Just so that we don't create the index, no other reason.

mfundul (approved):

PR #4273, opened 2022-04-25
Use binary format in row-by-row fetcher

akuzm:
This gives some minor speedup and prepares for use of COPY protocol in
row-by-row-fetcher. We can only use the binary format if all the data
types support binary serialization. This is checked at execution time
when creating a TupleFactory. The fetcher interface is refactored to
allow for external creation of TupleFactory.

When a data type doesn't have binary serialization, we revert to using Cursor fetcher.

Part of #4188

### DIFF
--- a/tsl/src/fdw/scan_exec.c
+++ b/tsl/src/fdw/scan_exec.c
@@ -22,6 +22,7 @@
 #include "remote/row_by_row_fetcher.h"
 #include "remote/cursor_fetcher.h"
 #include "guc.h"
+#include "planner.h"
 
 /*
  * Indexes of FDW-private information stored in fdw_private lists.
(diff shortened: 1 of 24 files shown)


tsl/src/remote/tuplefactory.c
```diff
@@ -90,22 +90,25 @@ conversion_error_callback(void *arg)
...
 		ScanState *ss = errpos->ss;
-		ForeignScan *fsplan;
+		List *tlist;
 		EState *estate = ss->ps.state;
 		TargetEntry *tle;
 
 		if (IsA(ss->ps.plan, ForeignScan))
-			fsplan = (ForeignScan *) ss->ps.plan;
+		{
+			ForeignScan *fsplan = (ForeignScan *) ss->ps.plan;
```
akuzm, line 99: This error reporting code used to segfault, I fixed it and added a test to dist_hypertable.

mkindahl (commented):
It's a little hard to understand what this pull request does from the description. You say that you're preparing for following changes, but the description does not outline what _this_ commit changes, so a better description would be suitable here.

You mention a problem with types that cannot be binary serialized, but it is unclear what was present before and what this commit changes, so something like along the lines "If the automatic remote data selection was picked by the user and the RowByRow protocol was picked but some data types do not support binary serialization, the current code will fail [does it?]. This commit fixes this by reverting to the Cursor fetcher type to read the rows."

tsl/test/expected/dist_hypertable-12.out
```diff
@@ -535,22 +535,19 @@ SELECT * FROM disttable;
          ->  Custom Scan (DataNodeScan) on public.disttable disttable_1
                Output: disttable_1."time", disttable_1.device, disttable_1.temp
                Data node: db_dist_hypertable_1
-               Fetcher Type: Row by row
```
svenklemm, line 538: Why did you remove fetcher type from explain output?

akuzm: Now the fetcher type is ultimately determined at the execution stage, so we can show it only in EXPLAIN ANALYZE. I'm not sure, maybe we should do this at planning stage? It is relatively heavy and involves looking up the send/receive functions for all participating columns, so I kept it at execution stage for now.

akuzm:
> It's a little hard to understand what this pull request does from the description.

The general idea is to have two types of fetcher: "fast" and "general purpose". I want to make the row-by-row fetcher into the "fast one". In this PR, I'm removing support of text protocol in this fetcher, because it's only needed for some niche types that don't have a binary serialization, and is also slower than binary one. Because the row-by-row fetcher now only understands binary protocol, we must check that the binary serialization is actually available for the participating data types. These types are currently looked up at execution time, so I basically added several lines of code that allow us to re-check this at execution time and revert to using Cursor fetcher if possible. The rest is removing the text protocol support from row-by-row, plus EXPLAIN changes (we don't know the fetcher type at the planning stage anymore, so not showing it).

tsl/src/fdw/scan_exec.c
```diff
@@ -128,26 +129,41 @@ create_data_fetcher(ScanState *ss, TsFdwScanState *fsstate)
...
+			 * case we can revert to cursor fetcher which supports text
+			 * serialization.
+			 */
+			fsstate->planned_fetcher_type = CursorFetcherType;
+		}
+		else
+		{
+			ereport(ERROR,
+					(errmsg("cannot use row-by-row fetcher because some of the column types do not "
+							"have binary serialization")));
```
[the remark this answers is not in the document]
akuzm: This is a user error: row-by-row fetcher is requested by the GUC, and some data types do not have binary serialization. There was some data type like this in our tests, I'll add a test case for this error.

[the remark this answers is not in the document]
akuzm: Added a test case.

svenklemm (approved):

mkindahl (approved):

PR #4281, opened 2022-04-28
Use one connection per data node for distributed copy insert

akuzm:
Currently we use one connection per chunk, and they all run the same
query `COPY destination_hypertable FROM STDIN`. Not sure what's the
point.

Part of https://github.com/timescale/timescaledb/pull/4285

### DIFF
--- a/tsl/src/remote/dist_copy.c
+++ b/tsl/src/remote/dist_copy.c
@@ -32,14 +32,6 @@
 #define DEFAULT_PG_DELIMITER '\t'
 #define DEFAULT_PG_NULL_VALUE "\\N"
 
-/* This will maintain a list of connections associated with a given chunk so we don't have to keep
- * looking them up every time.
- */
-typedef struct ChunkConnectionList
-{
-	int32 chunk_id;
-	List *connections;
-} ChunkConnectionList;
 
 /* This contains the information needed to parse a dimension attribute out of a row of text copy
  * data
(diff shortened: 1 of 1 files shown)


tsl/src/remote/dist_copy.c
```diff
@@ -54,13 +45,32 @@ typedef struct CopyDimensionInfo
...
-	List *cached_connections;
+	/*
+	 * Cached connections to data nodes.
+	 * Why do we need another layer of caching, when there is dist_txn layer
+	 * already? The API it provides is one function that "does everything
+	 * automatically", namely it's going to stop the COPY each time we request
+	 * the connection. This is not something we want to do for each row when
+	 * we're trying to do bulk copy.
+	 */
+	List *data_node_connections;
```
erimatnor, line 67: We might not need this list if we instead use the standard connection cache (see comment below).

erimatnor: It's ok to not refactor to use the connection cache in this PR, although I think this is what we should aim for as to not have multiple ways of dealing with connections.

akuzm: The problem with using `remote_connection_cache` directly is that the remote chunk creation (`chunk_api_create_on_data_nodes`) uses the `dist_txn` layer.

akuzm:
Merged as part of another PR.

PR #4284, opened 2022-04-28
Fix unqualified type reference in time_bucket

svenklemm:
This was not caught earlier and is not currently caught by CI
because the check for unqualified casts is currently only in main
branch of pgspot and not yet in the tagged version we use as
part of PR checks.


fabriziomello (approved):

mkindahl (approved):

svenklemm:
> Is there no other way? This is really unreadable code.

Normally the correct thing to do would be to set search_path on the function body but that would prevent inling so thats why we have to jump through hoops and make all object references schema qualified. Ideally we would get rid of the SQL implementation and move these to C which would have other benefits as well. At the moment this functions will expand to their inner sql so they are not actually seen in the extension code and instead an expression is seen with time_bucket not at toplevel.

PR #4286, opened 2022-04-29
Ignore flaky "dist_gapfill_pushdown" test

konskov:
This is a *temporary* change to our workflows to ignore the test
because it is flaky and requires many jobs to be restarted before a
commit can be merged.  Does not address the issue that causes the
flakiness, which needs to be investigated.


svenklemm (approved):
Can you add in commit message that this is temporary because the test is currently flaky.

mfundul:
We're leaving code in `main` that produces wrong results and just disable the test that detects the problem. I think we should disable the gapfill pushdown together with disabling the test themselves, otherwise we're keeping bad code in `main` and removing the only evidence that we have bad code.

PR #4287, opened 2022-04-29
Support foreign table as hypertable chunk

gayyappan:


sql/chunk.sql
```diff
@@ -90,3 +90,9 @@ RETURNS BOOL AS '@MODULE_PATHNAME@', 'ts_chunk_freeze_chunk' LANGUAGE C VOLATILE
 CREATE OR REPLACE FUNCTION _timescaledb_internal.drop_chunk(
    chunk REGCLASS)
 RETURNS BOOL AS '@MODULE_PATHNAME@', 'ts_chunk_drop_single_chunk' LANGUAGE C VOLATILE;
+
+-- internal API used by OSM extension to attach a table as a chunk of the hypertable
+CREATE OR REPLACE FUNCTION _timescaledb_internal.attach_osm_table_chunk(
```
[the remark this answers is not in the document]
gayyappan: I had a more generic name , attach_foreign_table_chunk, for my first pass.  I decided to give it a more specific name so that are no misconceptions - e.g. any foreign table can be attached to the hypertable.  For instance, this would not work for the multinode case. There are  constraints on the foreign table chunk that are specific to OSM.

src/chunk.c
```diff
@@ -4359,3 +4366,115 @@ ts_chunk_scan_iterator_set_chunk_id(ScanIterator *it, int32 chunk_id)
...
+	/* insert dimension slices if they do not exist.
+	 * Then add dimension constraints for the chunk
+	 */
+	ts_dimension_slice_insert_multi(chunk->cube->slices, chunk->cube->num_slices);
+	ts_chunk_constraints_add_dimension_constraints(chunk->constraints, chunk->fd.id, chunk->cube);
+	/* check constraints are not automatically created for foreign tables.
+	 * See: ts_chunk_constraints_add_dimension_constraints.
+	 */
+	ts_chunk_constraints_insert_metadata(chunk->constraints);
+	// return chunk;
```
fabriziomello, line 4451:
```suggestion
```

tsl/test/expected/chunk_utils_internal.out
```diff
@@ -280,3 +283,71 @@ SELECT * FROM hyper1_cagg ORDER BY 1;
...
+GRANT ALL PRIVILEGES ON DATABASE postgres_fdw_db TO :ROLE_4;
+\c postgres_fdw_db :ROLE_4
+CREATE TABLE fdw_table( timec timestamptz NOT NULL , acq_id bigint, value bigint);
+INSERT INTO fdw_table VALUES( '2020-01-01 01:00', 100, 1000);
+--create foreign server and user mappings as superuser
+\c :TEST_DBNAME :ROLE_SUPERUSER
+SELECT current_setting('port') as "PORTNO" \gset
+CREATE EXTENSION postgres_fdw;
+CREATE SERVER s3_server FOREIGN DATA WRAPPER postgres_fdw 
+OPTIONS ( host 'localhost', dbname 'postgres_fdw_db', port :'PORTNO');
```
fabriziomello, line 299:
```suggestion
OPTIONS (dbname 'postgres_fdw_db', port :'PORTNO');
```
You don't need to set the `host` to connect in the same postgres instance

src/chunk.c
```diff
@@ -4359,3 +4366,113 @@ ts_chunk_scan_iterator_set_chunk_id(ScanIterator *it, int32 chunk_id)
...
+	Assert(hs->num_dimensions == 1); // does not work with partitioned range
+	for (int i = 0; i < hs->num_dimensions; i++)
+	{
+		const Dimension *dim = &hs->dimensions[i];
+		Assert(dim->type == DIMENSION_TYPE_OPEN);
+		Oid dimtype = ts_dimension_get_partition_type(dim);
+		// Datum val = ts_time_datum_get_nobegin_or_min(dimtype);
+		Datum val = ts_time_get_min(dimtype);
+		p->coordinates[p->num_coords++] = ts_time_value_to_internal(val, dimtype);
+		cube->slices[i] = ts_dimension_calculate_default_slice(dim, p->coordinates[i]);
```
akuzm, line 4385: So this chunk gets the "earliest" time slice, right? This means you can't create two such chunks, because chunks don't intersect? I'd expect that `dimension_slice` for the chunk to be a sensible range corresponding to the chunk data.

gayyappan: This is a dummy dimension slice - it is not based on the actual data. The foreign table manages its own dimension/slice data. But we cannot a chunk to a hypertable without creating a dimension for it.

sql/chunk.sql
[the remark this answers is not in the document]
gayyappan: This interface does not work for a generic foreign table - it is targeted to work only with the foreign tables defined in the OSM extension.

src/chunk.c
akuzm: So the normal chunk exclusion is not going to work for hypertables with these foreign chunks, and it's performed separately by the OSM extension? Would be interesting to read a design doc for this, if there's one.

akuzm (approved):

PR #4288, opened 2022-05-02
Append pg_temp to search_path

svenklemm:
Postgres will prepend pg_temp to the effective search_path if it
is not present in the search_path. While pg_temp will never be
used to look up functions or operators unless explicitly requested
pg_temp will be used to look up relations. Putting pg_temp in
search_path makes sure objects in pg_temp will be considered last
and pg_temp cannot be used to mask existing objects.

### DIFF
--- a/sql/hypertable_constraint.sql
+++ b/sql/hypertable_constraint.sql
@@ -30,7 +30,7 @@ BEGIN
     IF def IS NOT NULL THEN
         -- to allow for custom types with operators outside of pg_catalog
         -- we set search_path to @extschema@
-        SET LOCAL search_path TO @extschema@;
+        SET LOCAL search_path TO @extschema@, pg_temp;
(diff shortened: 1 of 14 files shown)


sql/hypertable_constraint.sql
```diff
@@ -38,4 +38,4 @@ BEGIN
     END IF;
```
jnidzwetzki, line 38: Also in line 33 of this file, the `search_path` is adjusted. Adding `pg_temp` seems not to be necessary at the moment since no relations are looked up in the function. However, this might change in the future and the developer might forget to adjust the `search_path`.

svenklemm: good catch

jnidzwetzki (approved):

gayyappan (approved):

PR #4289, opened 2022-05-02
Support moving compressed chunks between data nodes

pmwkaa:
This change allows to copy or move compressed chunks between data nodes by including compressed chunk into the chunk copy command stages.

### DIFF
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -410,6 +410,7 @@ CREATE TABLE _timescaledb_catalog.chunk_copy_operation (
   completed_stage name NOT NULL, -- the completed stage/step
   time_start timestamptz NOT NULL DEFAULT NOW(), -- start time of the activity
   chunk_id integer NOT NULL REFERENCES _timescaledb_catalog.chunk (id) ON DELETE CASCADE,
+  compress_chunk_name name NOT NULL,
   source_node_name name NOT NULL,
   dest_node_name name NOT NULL,
   delete_on_source_node bool NOT NULL -- is a move or copy activity
--- a/sql/updates/reverse-dev.sql
+++ b/sql/updates/reverse-dev.sql
@@ -105,4 +105,83 @@ ANALYZE _timescaledb_catalog.continuous_agg;
 
 DROP PROCEDURE timescaledb_experimental.move_chunk(REGCLASS, NAME, NAME, NAME);
 DROP PROCEDURE timescaledb_experimental.copy_chunk(REGCLASS, NAME, NAME, NAME);
+
 DROP FUNCTION IF EXISTS timescaledb_experimental.subscription_exec(TEXT);
+
+DROP FUNCTION _timescaledb_internal.create_compressed_chunk(REGCLASS, REGCLASS,
+	BIGINT, BIGINT, BIGINT, BIGINT, BIGINT, BIGINT, BIGINT, BIGINT);
+
+--
+-- Rebuild the catalog table `_timescaledb_catalog.chunk_copy_operation`
+--
+-- We need to recreate the catalog from scratch because when we drop a column
+-- Postgres mark the `pg_attribute.attisdropped=TRUE` instead of removing it from
+-- the `pg_catalog.pg_attribute` table.
+--
+-- If we downgrade and upgrade the extension without rebuild the catalog table it
+-- will mess with `pg_attribute.attnum` and we will end up with issues when trying
+-- to update data in those catalog tables.
+
+ALTER TABLE _timescaledb_catalog.chunk_copy_operation
+	DROP COLUMN compress_chunk_name;
+
+CREATE TABLE _timescaledb_catalog._tmp_chunk_copy_operation (
+    LIKE _timescaledb_catalog.chunk_copy_operation
+    INCLUDING ALL
+    EXCLUDING INDEXES
+    EXCLUDING CONSTRAINTS
+);
+
+INSERT INTO _timescaledb_catalog._tmp_chunk_copy_operation
+    SELECT
+        operation_id,
+        backend_pid,
+        completed_stage,
+        time_start,
+        chunk_id,
+        source_node_name,
+        dest_node_name,
+        delete_on_source_node
+    FROM
+        _timescaledb_catalog.chunk_copy_operation
+    ORDER BY
+        operation_id;
+
+ALTER EXTENSION timescaledb
+    DROP TABLE _timescaledb_catalog.chunk_copy_operation;
+
+DROP TABLE _timescaledb_catalog.chunk_copy_operation;
+
+CREATE TABLE _timescaledb_catalog.chunk_copy_operation (
+    LIKE _timescaledb_catalog._tmp_chunk_copy_operation
+    INCLUDING ALL
+    EXCLUDING INDEXES
+    EXCLUDING CONSTRAINTS
+);
+
+-- Create a new table to void doing rename operation on the tmp table
+--
+INSERT INTO _timescaledb_catalog.chunk_copy_operation
+    SELECT
+        operation_id,
+        backend_pid,
+        completed_stage,
+        time_start,
+        chunk_id,
+        source_node_name,
+        dest_node_name,
+        delete_on_source_node
+    FROM
+        _timescaledb_catalog._tmp_chunk_copy_operation
+    ORDER BY
+        operation_id;
+
+DROP TABLE _timescaledb_catalog._tmp_chunk_copy_operation;
+
+ALTER TABLE _timescaledb_catalog.chunk_copy_operation
+    ADD CONSTRAINT chunk_copy_operation_pkey PRIMARY KEY (operation_id),
+    ADD CONSTRAINT chunk_copy_operation_chunk_id_fkey FOREIGN KEY (chunk_id) REFERENCES _timescaledb_catalog.chunk(id) ON DELETE CASCADE;
+
+GRANT SELECT ON TABLE _timescaledb_catalog.chunk_copy_operation TO PUBLIC;
+
+ANALYZE _timescaledb_catalog.chunk_copy_operation;
--- a/src/cross_module_fn.c
+++ b/src/cross_module_fn.c
@@ -11,6 +11,7 @@
 #include "export.h"
 #include "cross_module_fn.h"
 #include "guc.h"
+#include "license_guc.h"
 #include "bgw/job.h"
 #ifdef USE_TELEMETRY
 #include "telemetry/telemetry.h"
@@ -69,6 +70,7 @@ CROSSMODULE_WRAPPER(dictionary_compressor_append);
 CROSSMODULE_WRAPPER(dictionary_compressor_finish);
 CROSSMODULE_WRAPPER(array_compressor_append);
 CROSSMODULE_WRAPPER(array_compressor_finish);
+CROSSMODULE_WRAPPER(create_compressed_chunk);
 CROSSMODULE_WRAPPER(compress_chunk);
 CROSSMODULE_WRAPPER(decompress_chunk);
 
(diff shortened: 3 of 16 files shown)


tsl/test/expected/dist_move_chunk.out
```diff
@@ -335,6 +321,208 @@ FROM timescaledb_experimental.chunk_replication_status;
...
+
+-- Make sure same compressed chunk hash been created on the destination data node
+\c :DN_DBNAME_2 :ROLE_CLUSTER_SUPERUSER;
+-- Chunk created on data node has different id but the same name, make sure
+-- compressed_chunk_id is correctly set
+SELECT c2.table_name
+FROM _timescaledb_catalog.chunk c1
+JOIN _timescaledb_catalog.chunk c2 ON (c1.compressed_chunk_id = c2.id)
+WHERE c1.table_name = 'compress_hyper_3_12_chunk';
+ table_name
```
nikkhils, line 441: This should return a compressed chunk name, right?

tsl/src/chunk_copy.c
```diff
@@ -420,16 +426,120 @@ chunk_copy_stage_create_empty_chunk_cleanup(ChunkCopy *cc)
...
+	dist_res =
+		ts_dist_cmd_invoke_on_data_nodes(cmd, list_make1(NameStr(cc->fd.source_node_name)), true);
+	res = ts_dist_cmd_get_result_by_node_name(dist_res, NameStr(cc->fd.source_node_name));
+	if (PQresultStatus(res) != PGRES_TUPLES_OK)
+		ereport(ERROR,
+				(errcode(ERRCODE_CONNECTION_EXCEPTION), errmsg("%s", PQresultErrorMessage(res))));
+
+	/* Set compressed chunk name of the source data node */
+	if (PQntuples(res) != 0)
+		snprintf(cc->fd.compressed_chunk_name.data,
```
nikkhils, line 452: Can assert that we got ONE result from the other side.

pmwkaa: Not sure that Assert is a good way to deal here, but I think we should generate an error at least

tsl/test/expected/dist_move_chunk.out
pmwkaa: Good catch, this should be `_dist_hyper_3_12_chunk` but query should return `compress_hyper_3_6_chunk`

tsl/src/compression/compress_utils.c
```diff
@@ -436,6 +437,80 @@ decompress_remote_chunk(FunctionCallInfo fcinfo, const Chunk *chunk, bool if_com
 	return success;
 }
 
+/*
+ * Create a new compressed chunk using existing table with compressed data.
+ *
```
akuzm, line 442: Sounds like "attach", not "create"

tsl/src/compression/create.c
```diff
@@ -585,7 +585,7 @@ create_compression_table(Oid owner, CompressColInfo *compress_cols)
  * Caller is expected to do this explicitly.
  */
 Chunk *
-create_compress_chunk_table(Hypertable *compress_ht, Chunk *src_chunk)
+create_compress_chunk_table(Hypertable *compress_ht, Chunk *src_chunk, Oid table_id)
```
akuzm, line 588: Not sure it's feasible, but would be more straightforward if this was two functions, create the table + attach it as a compressed chunks. When moving the chunk, you replace the "create the table" part with "fetch it from other node".

src/cross_module_fn.c
```diff
@@ -398,9 +415,10 @@ TSDLLEXPORT CrossModuleFunctions ts_cm_functions_default = {
 	/* compression */
 	.compressed_data_send = error_no_default_fn_pg_community,
 	.compressed_data_recv = error_no_default_fn_pg_community,
-	.compressed_data_in = error_no_default_fn_pg_community,
+	.compressed_data_in = process_compressed_data_in,
```
akuzm, line 418: I can't find a comment anywhere about what "compressed_data_in" is supposed to do.

This change is a little unclear, so you want to still have some wrapper for this function even in an unlicensed version, because the replication worker doesn't load TSL by default?

tsl/src/compression/compress_utils.c
```diff
@@ -436,6 +437,80 @@ decompress_remote_chunk(FunctionCallInfo fcinfo, const Chunk *chunk, bool if_com
...
+	LockRelationOid(cxt.compress_ht->main_table_relid, AccessShareLock);
+	LockRelationOid(cxt.srcht_chunk->table_id, ShareLock);
+
+	/* Aquire locks on catalog tables to keep till end of txn */
+	LockRelationOid(catalog_get_table_id(ts_catalog_get(), HYPERTABLE_COMPRESSION),
+					AccessShareLock);
+	LockRelationOid(catalog_get_table_id(ts_catalog_get(), CHUNK), RowExclusiveLock);
+
+	/* Create compressed chunk DDL using existing table */
+	compress_ht_chunk = create_compress_chunk_table(cxt.compress_ht, cxt.srcht_chunk, chunk_table);
```
akuzm, line 484: So the function `create_compress_chunk_table` actually uses the existing compressed table, but creates something else? Sounds confusing :)

tsl/src/compression/create.c
akuzm: I mean, now the function "something-something-create-table" has an argument that means "but don't actually create the table", and this might be confusing.

akuzm (approved):
I somewhat understand what this PR does, seems reasonable.

nikkhils (approved):
LGTM. As akuzm pointed out, a few more comments for the confusing functions makes sense.

tsl/src/compression/compress_utils.c
pmwkaa: Well, it is not actually attach, since chunk not exists yet, basically no meta-data about it. Additionally this and further API changes are mimicking what we already had for non-compressed chunks

tsl/src/compression/compress_utils.c
pmwkaa: I did not want to copy-paste or change compression related functions here. The only difference here is that now attached table consist compressed data, before that it was empty. If you take a look at the `compress_chunk` API

src/cross_module_fn.c
pmwkaa: This will work only in licensed version, unlicensed still will produce an error.

The problem was, that even in licensed version TSL module did not loaded properly from replication worker, so it produced misleading error which says that your license is "timescale" and you should upgrade to "timescale". I did not found a better place to check if module is loaded.

`compressed_data_in` function called when replication worker started to insert data to table created on dest data node. I think there is a type exists for compressed data `_timescaledb_internal.compressed_data`

tsl/src/compression/create.c
akuzm: I mean, can we move the `ts_chunk_create_table` out of  `create_compress_chunk_table`, and always pass `table_oid`? And call `create_compress_chunk_table` something like `create_compressed_chunk_metadata`. If not, makes sense to at least rename it somehow, because now it reads like `void create_table(bool do_not_create_table)` .

tsl/src/compression/compress_utils.c
akuzm: As far as I understand, there are two parts of creating a compressed chunk: creating the compressed chunk table and creating the metadata. If you already have the table and want to create the respective metadata, I'd call this operation "attaching a compressed chunk". Do we have other kind of "attach"?

src/cross_module_fn.c
akuzm: Ah, OK, I think I understand better now.
I'd write something like this:
```
/*
 * TSL library is not loaded by the replication worker for some reason,
 * so a call to `compressed_data_in` function would produce a misleading
 * error saying that your license is "apache" and you should upgrade to
 * "timescale" license, even if you have already upgraded.
 * As a workaround, we try to load the TSL module it in this function. It will still
 * error out in the "apache" version
 */
 ```
 
 And I'd also replace the function pointer check with `ts_is_license_apache`, it was kinda confusing for me.

File: tsl/src/compression/compress_utils.c (hunk 3 above)
Comment (pmwkaa) [reply tsl/src/compression/compress_utils.c:442]: It is basically the same as if you `select create_hypertable(table)` when table has data, it still reads as create not attach, right? I guess this is a matter of perspective

File: tsl/src/compression/create.c (hunk 4 above)
Comment (pmwkaa) [reply tsl/src/compression/create.c:588]: This function is used in other places, like `compress_chunk`. I am not sure it is a good idea to touch them here and do bigger refactoring.

I think the problem is the function name, how about name it just `create_compress_chunk` instead?

File: tsl/src/compression/compress_utils.c (hunk 3 above)
Comment (akuzm) [reply tsl/src/compression/compress_utils.c:442]: Yep, but it creates different tables and moves data there. If it was just modifying metadata, I'd also think that it is an "attach" operation. Anyway, don't want to slow you down too much with these terminological discussions :)

File: tsl/src/compression/create.c (hunk 4 above)
Comment (pmwkaa) [reply tsl/src/compression/create.c]: I've renamed it, hope it looks better now

Comment (akuzm) [reply tsl/src/compression/create.c]: Yes, this name is better

**Review by svenklemm — CHANGES_REQUESTED** [review]
Since operation_id is NAME you should quote it with quote_identifier before using it in sql queries. Should also add tests accordingly.

File: sql/updates/reverse-dev.sql [hunk 7]
```diff
@@ -105,3 +105,9 @@ ANALYZE _timescaledb_catalog.continuous_agg;
 
 DROP PROCEDURE timescaledb_experimental.move_chunk(REGCLASS, NAME, NAME, NAME);
 DROP PROCEDURE timescaledb_experimental.copy_chunk(REGCLASS, NAME, NAME, NAME);
+
+ALTER TABLE _timescaledb_catalog.chunk_copy_operation
+	DROP COLUMN compress_chunk_name;
```
svenklemm, line 110: dropping a column needs  to completely rewrite the table otherwise you create a hazard for future changes to that table.

pmwkaa: Can you elaborate this one please?

svenklemm: You must not use DROP COLUMN in a upgrade/downgrade script as DROP COLUMN only marks the column dropped in the postgres catalog but leaves metadata behind resulting in a later ADD COLUMN behaving differently between fresh install/updated/downgraded install as the assigned attnum will be different. To get completely rid of the column you need to completely rebuild the table. copy current data into temp table, drop current table, recreate table with correct schema, there should be many examples of this in previous upgrade/downgrade scripts.

pmwkaa:
@svenklemm During testing `operation_id` I've found out that it actually should be compatible with slot naming conventions, which are more strict then regular names. I've used `ReplicationSlotValidateName()` function in case if the `operation_id` argument is being used. I don't see a need to quote it everywhere, since the naming convention allows to use it without quoting (basically as the auto-generated names for copy move chunk operation we used before).

I've also updated the upgrade scripts as suggested.

tsl/src/chunk_copy.c
```diff
@@ -732,6 +900,50 @@ chunk_copy_stage_attach_chunk(ChunkCopy *cc)
...
+
+	if (!ts_chunk_is_compressed(cc->chunk))
+		return;
+
+	chunk_name = psprintf("%s.%s",
+						  quote_identifier(cc->chunk->fd.schema_name.data),
+						  quote_identifier(cc->chunk->fd.table_name.data));
+
+	compressed_chunk_name = psprintf("%s.%s",
+									 quote_identifier(INTERNAL_SCHEMA_NAME),
```
svenklemm, line 920: hmm shouldnt we use the actual schema of the chunk here in case it isnt in internal schema

svenklemm (approved):

sql/pre_install/tables.sql
```diff
@@ -410,6 +410,7 @@ CREATE TABLE _timescaledb_catalog.chunk_copy_operation (
   completed_stage name NOT NULL, -- the completed stage/step
   time_start timestamptz NOT NULL DEFAULT NOW(), -- start time of the activity
   chunk_id integer NOT NULL REFERENCES _timescaledb_catalog.chunk (id) ON DELETE CASCADE,
+  compress_chunk_name name NOT NULL,
```
fabriziomello, line 413: Why add it in the middle of the catalog table instead of at the end?

PR #4294, opened 2022-05-02
Remove `chunk_id` from materialization hypertable

fabriziomello:
First step to remove the re-aggregation for Continuous Aggregates
is to remove the `chunk_id` from the materialization hypertable.

Also added new metadata column named `finalized` to `continuous_cagg`
catalog table in order to store information about the new following
finalized version of Continuous Aggregates that will not need the
partials anymore. This flag is important to maintain backward
compatibility with previous Continuous Aggregate implementation that
requires the `chunk_id` to refresh data properly.


tsl/test/expected/cagg_drop_chunks.out
```diff
@@ -145,184 +145,152 @@ NOTICE:  drop cascades to 32 other objects
...
+CREATE TABLE conditions(time_int INT NOT NULL, value FLOAT);
+SELECT create_hypertable('conditions', 'time_int', chunk_time_interval => 4);
     create_hypertable    
 -------------------------
  (3,public,conditions,t)
 (1 row)
 
 INSERT INTO conditions
-SELECT time_val, time_val % 4, 3.14 FROM generate_series(0,100,1) AS time_val;
+SELECT time_val, 1 FROM generate_series(0, 19, 1) AS time_val;
```
mfundul, line 183: Why can we not copy the original data? Is there some hidden limitation?

tsl/test/expected/cagg_drop_chunks.out
```diff
@@ -145,184 +145,152 @@ NOTICE:  drop cascades to 32 other objects
...
-(11 rows)
-
--- This is the simplest case, when the updated bucket is inside a chunk, so it is expected
--- that the update is refreshed.
-UPDATE conditions SET value = 4.00 WHERE time_int = 2;
--- This case updates values in the bucket, which affects two different partials in
--- two different chunks.
-UPDATE conditions SET value = 4.00 WHERE time_int = 9;
-UPDATE conditions SET value = 4.00 WHERE time_int = 11;
-SELECT timescaledb_experimental.refresh_continuous_aggregate('conditions_7', show_chunks('conditions', 20));
```
mfundul, line 200: Why do we remove those comments? Don't we test the same cases?

tsl/test/expected/cagg_drop_chunks.out
fabriziomello: Just did it to make review life easier

tsl/test/expected/cagg_drop_chunks.out
fabriziomello: It's not exactly the same because with the new implementation we can't test this in this way because the raw hypertable chunks are not aligned anymore with data stored into the materialization hypertable (we removed the chunk_id)

mfundul (approved):

tsl/test/sql/include/cagg_ddl_common.sql
```diff
@@ -288,15 +288,9 @@ SELECT * FROM truncate_partitioned;
 -- ALTER TABLE tests
 \set ON_ERROR_STOP 0
 -- test a variety of ALTER TABLE statements
-ALTER TABLE :drop_chunks_mat_table_u RENAME chunk_id TO bad_name;
-ALTER TABLE :drop_chunks_mat_table_u ADD UNIQUE(chunk_id);
```
svenklemm, line 292: why are you removing tests here? can't you do these tests on a different column instead of removing them?

fabriziomello: fixed!

svenklemm (approved):

Issue #4306, opened 2022-05-05
[Enhancement]: Parallel Execution Plans on Data Nodes


TroyDey:
### What type of enhancement is this?

Performance

### What subsystems and features will be improved?

Distributed hypertable

### What does the enhancement do?

I have pretty good luck getting Postgres to choose parallel execution plans for regular hypertables, but it is nearly impossible to get Postgres to choose a parallel execution plan for remote queries executed on the data nodes.  I have been experimenting with this using the timescaledb.enable_remote_explain = on setting and simply executing the remote SQL directly on the data node.

One barrier I found was that the partialize_agg function completely prevents a parallel plan from being chosen even when force_parallel_mode is set to on.

When I set parallel_setup_cost and parallel_tuple_cost = 1 and force_parallel_mode = on, and change partialize_agg to a regular aggregation, such as avg, I can generate a parallel plan but it only ever plans one worker even though max_worker_processes = 59, max_parallel_workers = 48 and max_parallel_workers_per_gather = 24, and the table contains 25 million rows.  This has become part of another discussion [here](https://www.timescale.com/forum/t/distributed-hypertable-poor-performance-for-simple-bucketed-aggregation/498).

I realize this is open ended so feel free to close, I mainly wanted to raise the issue of partialize_agg blocking parallel plans.

PR #4307, opened 2022-05-06
Mark partialize_agg as parallel safe

akuzm:
Just an experiment to see what happens.

UPD: the experiment turned out to be a success. Seems that we can mark the partialize_agg function as parallel safe always, and the parallel plans are generated depending on the underlying aggregate function. Postgres does the full expression tree traversal to check for parallel unsafe functions, as expected.

Fixes https://github.com/timescale/timescaledb/issues/4306

### Commit message
Mark partialize_agg as parallel-safe

This function is used to push down aggregation to data nodes, and making it parallel-safe allows to use parallel plans on data nodes, if the underlying aggregate function is parallel-safe as well. Add negative and positive tests for this.

### DIFF
--- a/test/runner_shared.sh
+++ b/test/runner_shared.sh
@@ -81,6 +81,7 @@ ${PSQL} -U ${TEST_PGUSER} \
           sed  -e '/<exclude_from_test>/,/<\/exclude_from_test>/d' \
                -e 's!_[0-9]\{1,\}_[0-9]\{1,\}_chunk!_X_X_chunk!g' \
                -e 's!^ \{1,\}QUERY PLAN \{1,\}$!QUERY PLAN!' \
+               -e 's!:  actual rows!: actual rows!' \
                -e '/^-\{1,\}$/d' \
                -e 's!\(_timescaledb_internal.chunks_in([^,]\{1,\}, ARRAY\[\)[^]]\{1,\}\]!\1..]!' \
                -e 's! Memory: [0-9]\{1,\}kB!!' \
--- a/sql/partialize_finalize.sql
+++ b/sql/partialize_finalize.sql
@@ -2,24 +2,30 @@
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-APACHE for a copy of the license.
 
+-- These wrapper functions are used to push down aggregation to data nodes.
+-- They can be marked parallel-safe, and the parallel plan will be chosen
+-- depending on whether the underlying aggregate function itself is
+-- parallel-safe.
+
 CREATE OR REPLACE FUNCTION _timescaledb_internal.partialize_agg(arg ANYELEMENT)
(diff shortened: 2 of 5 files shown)


sql/updates/latest-dev.sql
```diff
@@ -0,0 +1,30 @@
...
+AS '@MODULE_PATHNAME@', 'ts_finalize_agg_ffunc'
+LANGUAGE C IMMUTABLE PARALLEL SAFE;
+
+CREATE OR REPLACE AGGREGATE _timescaledb_internal.finalize_agg(agg_name TEXT,  inner_agg_collation_schema NAME,  inner_agg_collation_name NAME, inner_agg_input_types NAME[][], inner_agg_serialized_state BYTEA, return_type_dummy_val anyelement) (
+    SFUNC = _timescaledb_internal.finalize_agg_sfunc,
+    STYPE = internal,
+    FINALFUNC = _timescaledb_internal.finalize_agg_ffunc,
+    FINALFUNC_EXTRA,
+    PARALLEL = SAFE
+);
```
svenklemm, line 30:
```suggestion
```
This is not needed as CREATE OR REPLACE for all functions is appended to any update script we generate. You only need  this if you change signature for example.

test/runner_shared.sh
```diff
@@ -81,6 +81,7 @@ ${PSQL} -U ${TEST_PGUSER} \
           sed  -e '/<exclude_from_test>/,/<\/exclude_from_test>/d' \
                -e 's!_[0-9]\{1,\}_[0-9]\{1,\}_chunk!_X_X_chunk!g' \
                -e 's!^ \{1,\}QUERY PLAN \{1,\}$!QUERY PLAN!' \
+               -e 's!:  actual rows!: actual rows!' \
```
svenklemm, line 84: Hmm not a fan of this. Can you get a stable test output without this change as the number of rows is useful for other tests so I would rather not have this removed globally.

akuzm: I'm keeping the number, just replacing double space for single. The formatting was changed between postgres versions, so I'm normalizing it to avoid having separate references.

sql/updates/latest-dev.sql
akuzm: Ah, right, removed.

svenklemm (approved):

erimatnor (approved):
Nice. Could you expand in the PR description (commit message) on why it is OK to make the function parallel safe.

I assume it is because partial aggregates actually exists in PG to do parallel computation, so they are parallel safe by definition? But it would be nice to make that clear if this is the reason.

akuzm:
> I assume it is because partial aggregates actually exists in PG to do parallel computation, so they are parallel safe by definition? But it would be nice to make that clear if this is the reason.

Yes, Postgres knows whether a given aggregate is parallel-safe and paralellizes based on that. We don't need to artificially disable parallel plans with our partial aggregation wrapper, because this wrapper is a pure function that produces the unfinished aggregation state as the result, and doesn't change parallel safety. I'll try to describe this in more detail...

PR #4311, opened 2022-05-09
Support non-superuser move chunk operations

nikkhils:
The non-superuser needs to have REPLICATION privileges atleast. A
new function "subscription_cmd" has been added to allow running
subscription related commands on datanodes. This function implicitly
upgrades to the bootstrapped superuser and then performs subscription
creation/alteration/deletion commands. It only accepts subscriptions
related commands and errors out otherwise.

### DIFF
--- a/tsl/src/chunk_copy.c
+++ b/tsl/src/chunk_copy.c
@@ -5,6 +5,7 @@
  */
 #include <postgres.h>
 #include <foreign/foreign.h>
+#include <catalog/pg_authid.h>
 #include <catalog/pg_foreign_server.h>
 #include <catalog/pg_foreign_table.h>
 #include <catalog/dependency.h>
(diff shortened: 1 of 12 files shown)


tsl/src/chunk_copy.c
```diff
@@ -635,21 +643,27 @@ chunk_copy_stage_sync(ChunkCopy *cc)
 static void
 chunk_copy_stage_drop_subscription(ChunkCopy *cc)
 {
-	char *cmd;
+	char *cmd1, *cmd;
 
 	/* Stop data transfer on the destination node */
-	cmd = psprintf("ALTER SUBSCRIPTION %s DISABLE", NameStr(cc->fd.operation_id));
+	cmd1 = psprintf("ALTER SUBSCRIPTION %s DISABLE", NameStr(cc->fd.operation_id));
+	cmd = psprintf("SELECT timescaledb_experimental.subscription_cmd($sql$%s$sql$)", cmd1);
```
pmwkaa, line 650: Since those command are repeated, I am suggesting to create a function which executes it remote as a wrapper:
```

static void
chunk_copy_execute_subscription_command(ChunkCopy *cc, const char *command, data nodes ...)
{
// include command in the argument to subscription_cmd()
// execute on data nodes
}

or

static void
chunk_copy_subscription_exec(...)
{
}
```

pmwkaa (approved):
Looks good overall, I suggest to create a separate function which wraps a call to the new function instead of repeating it.

Maybe a better name for `subscription_cmd`, maybe `subscription_exec` just to be in sync with `distributed_exec` for example

svenklemm (dismissed):
Since operation_id is NAME you should quote it with quote_identifier before using it in sql queries. Should also add tests accordingly.

tsl/src/chunk_copy.c
fabriziomello: +1 for @pmwkaa suggestion

nikkhils: @pmwkaa @fabriziomello done the above

nikkhils:
> Since operation_id is NAME you should quote it with quote_identifier before using it in sql queries. Should also add tests accordingly.

@svenklemm As mentioned by @pmwkaa he is fixing this in his PR. If you don't have any other feedback can you please approve this PR?

svenklemm (approved):

PR #4312, opened 2022-05-09
Update dump_meta_data.sql

lservini:
Added `c.dropped is false ` for table _timescaledb_catalog.chunk to skip dropped chunks in the size calculation.


fabriziomello:
@lservini thanks for your PR. Some comments:

1. You need to fix the regression tests failures as pointed out in CI
2. Add a failing test case for your scenario
3. Also would be good to use our own `_timescaledb_internal.relation_size` function instead of Postgres size functions

gayyappan (approved):

svenklemm (approved):

PR #4313, opened 2022-05-09
Telemetry Stats for CAggs finals form

fabriziomello:
Introduced by #4294 and #4269 PRs the default implementation of
Continuous Aggregates get rid of `chunk_id` in the materialization
hypertable and `partialize_agg`/`finalize_agg` aggregate functions.

A new counter named `num_caggs_finalized` was added to telemetry report
in this PR to count the number of Continuos Aggregates created in this
new format.

### DIFF
--- a/src/ts_catalog/continuous_agg.h
+++ b/src/ts_catalog/continuous_agg.h
@@ -92,6 +92,12 @@ typedef struct ContinuousAgg
 	Oid partition_type;
 } ContinuousAgg;
 
+static inline bool
+ContinuousAggIsFinalized(const ContinuousAgg *cagg)
+{
+	return (cagg->data.finalized == true);
+}
+
 typedef enum ContinuousAggHypertableStatus
 {
 	HypertableIsNotContinuousAgg = 0,
--- a/src/telemetry/stats.c
+++ b/src/telemetry/stats.c
@@ -266,6 +266,9 @@ process_continuous_agg(CaggStats *cs, Form_pg_class class, const ContinuousAgg *
 
 	if (!cagg->data.materialized_only)
 		cs->uses_real_time_aggregation_count++;
+
+	if (ContinuousAggIsFinalized(cagg))
+		cs->finalized++;
 }
 
 static void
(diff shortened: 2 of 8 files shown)


gayyappan (approved):

src/ts_catalog/continuous_agg.h
```diff
@@ -26,6 +26,8 @@ typedef enum ContinuousAggViewOption
 	ContinuousViewOptionFinalized,
 } ContinuousAggViewOption;
 
+#define ContinuousAggIsFinalized(cagg) (cagg->data.finalized == true)
```
mfundul, line 29:
```suggestion
static inline bool ContinuousAggIsFinalized(ContinuousAgg *agg)
{
    return (cagg->data.finalized == true);
}
```
I think we should avoid macros if possible. This way we get type checking and other checks, readable code and identical performance.

mfundul: Macros are [unsafe](https://wiki.sei.cmu.edu/confluence/display/c/PRE00-C.+Prefer+inline+or+static+functions+to+function-like+macros), especially since it's dereferencing arbitrary text, and it's just as fast, see [gcc documentation](https://gcc.gnu.org/onlinedocs/gcc-3.1.1/gcc/Inline.html).

Using `static inline` in header files is a very common pattern, e.g. see the linux kernel:
https://github.com/torvalds/linux/blob/9be9ed2612b5aedb52a2c240edb1630b6b743cb6/include/linux/list.h#L261
and postgresql:
https://github.com/postgres/postgres/blob/fe20afaee8aac7838ed6e4a76baa83e547629582/src/include/lib/ilist.h#L278

It's not a problem being used in different modules or in any kind of usage as a library.

fabriziomello: Well Postgres also use macros for the same purpose: https://github.com/postgres/postgres/blob/master/src/include/c.h#L710

Anyway you convinced me and I've changed it.

mfundul (approved):

PR #4319, opened 2022-05-11
More tests for errors when fetching from data nodes

akuzm:
Add a special function that allows to inject these errors.

### DIFF
--- a/test/src/test_utils.c
+++ b/test/src/test_utils.c
@@ -7,12 +7,15 @@
 #include <fmgr.h>
 #include <utils/builtins.h>
 #include <storage/procarray.h>
+#include <storage/proc.h>
 #include <commands/dbcommands.h>
 
 #include "test_utils.h"
 #include "debug_point.h"
 
 TS_FUNCTION_INFO_V1(ts_test_error_injection);
+TS_FUNCTION_INFO_V1(ts_debug_shippable_error_after_n_rows);
+TS_FUNCTION_INFO_V1(ts_debug_shippable_fatal_after_n_rows);
 
 /*
  * Test assertion macros.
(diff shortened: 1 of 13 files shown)


akuzm:
Looks like Coverity is broken by the merge commits. Anyway, this should be better than before.

tsl/src/fdw/deparse.c
```diff
@@ -327,6 +327,15 @@ contain_mutable_functions_checker(Oid func_id, void *context)
 	if (function_is_whitelisted(func_id))
 		return false;
 
+#ifndef NDEBUG
+	/* Special debug functions that we want to ship to data nodes. */
+	const char debug_func_prefix[] = "ts_debug_";
```
erimatnor, line 332: Maybe the prefix should be something like `ts_debug_remote_` or `ts_debug_shippable_` since `ts_debug_` is quite generic and developers could easily create other debug functions with that prefix that they do not intend to get pushed-down to data nodes (causing failures if they haven't created the function on the data nodes).

tsl/test/expected/dist_hypertable-14.out
```diff
@@ -5140,6 +5140,29 @@ SELECT count(*) FROM disttable;
 -- ignore this oid mismatch for non-builtin types.
 SET timescaledb.enable_connection_binary_data=true;
 \set ON_ERROR_STOP 0
+SET timescaledb.remote_data_fetcher = 'rowbyrow';
+SELECT * FROM disttable ORDER BY 1;
```
erimatnor, line 5144: Nit: would be nice if we could compare the output of rowbyrow and cursor programmatically instead of having to do it visually.

test/src/test_utils.c
```diff
@@ -75,6 +78,44 @@ ts_test_error_injection(PG_FUNCTION_ARGS)
 	PG_RETURN_VOID();
 }
 
+static int
+throw_every_nth_row(int max_rows, int severity)
```
erimatnor, line 82: I am a little confused about the semantics here. It seems to me that it errors out _after_ N rows rather then every nth row. I mean, once it errors the first time it won't continue.

erimatnor (commented):
Looks fine, although I am not sure about the semantics of the function (see separate inline comment).

tsl/test/expected/dist_hypertable-14.out
akuzm: They shouldn't necessarily be the same. In general, I don't care about output in these tests since it can be influenced by some minor changes in libpq between releases and so on. I'd disable the error output altogether if I could. The main purpose is to get the error handling code some coverage -- check that it doesn't segfault and doesn't break the state of the session somehow.

tsl/src/fdw/deparse.c
akuzm: Will rename.

test/src/test_utils.c
akuzm: Yep, makes sense, I should call it `throw_after_n_rows`.

tsl/test/expected/dist_hypertable-14.out
akuzm: Oops, I was thinking about a different test. For this one, its purpose is to fail in PG <= 13 to test the error handling code. So the comparison of results in 14 is not intended.

test/src/test_utils.c
pmwkaa: It also reads as if literally throw each nth row :black_cat:

pmwkaa (approved):

test/src/test_utils.c
akuzm: > It also reads as if literally throw each nth row

You mean, the new name `throw_after_n_rows` as well? I'm out of ideas, but if you suggest something, can rename.

PR #4324, opened 2022-05-12
Refactor upgrade/downgrade setup sql test scripts

fabriziomello:
The setup scripts for upgrade/downgrade tests of Continuous Aggregates
has too many duplicated code for pre-2.0 tests. Refactor it a bit
removing the duplicated code by using `\if \else \endif` psql
meta-commands.

Also added a properly `round` function to all functions that returns
`float8` in SQL scripts  because in rare cases it lead to flaky tests.

This is part of #4269.

### DIFF
--- a/test/sql/updates/setup.continuous_aggs.sql
+++ b/test/sql/updates/setup.continuous_aggs.sql
@@ -2,6 +2,15 @@
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-APACHE for a copy of the license.
 
+-- Collect information about different features so that we can pick
+-- the right usage. Some of these are changed in the same version, but
+-- we keep them separate anyway so that we can do additional checking
+-- if necessary.
+SELECT
+	extversion < '2.0.0' AS has_refresh_mat_view
+  FROM pg_extension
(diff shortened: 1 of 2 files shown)


akuzm (approved):

svenklemm (approved):

test/sql/updates/setup.continuous_aggs.sql
```diff
@@ -26,22 +35,21 @@ SELECT generate_series('2018-11-01 00:00'::timestamp, '2018-12-31 00:00'::timest
...
-	min(allnull) as min_allnull,
-	max(temperature) as max_temp,
-	sum(temperature)+sum(humidity) as agg_sum_expr,
-	avg(humidity) AS avg_humidity,
-	stddev(humidity),
+	round(min(allnull)) as min_allnull,
+	round(max(temperature)) as max_temp,
+	round(sum(temperature)+sum(humidity)) as agg_sum_expr,
+	round(avg(humidity)) AS avg_humidity,
+	round(stddev(humidity)) as stddev_humidity,
```
svenklemm, line 52: Why are you adding the round?

PR #4325, opened 2022-05-12

svenklemm:
Bump postgres versions used in CI to 14.3, 13.7 and 12.11.

Disable-check: commit-count


akuzm (approved):

fabriziomello (approved):

fabriziomello:
@svenklemm wouldn't better send those refactoring in another PR??? Because we approved it based on the $SUBJECT and $DESCRIPTION

svenklemm:
> @svenklemm wouldn't better send those refactorings in another PR??? Because we approved it based on the $SUBJECT and $DESCRIPTION

These are related to subject and will cut done the number of places we have to adjust when doing another bump for the next postgres version in a couple months. So I dont think we need separate PRs here.

fabriziomello:
> > @svenklemm wouldn't better send those refactorings in another PR??? Because we approved it based on the $SUBJECT and $DESCRIPTION
> 
> These are related to subject and will cut done the number of places we have to adjust when doing another bump for the next postgres version in a couple months. So I dont think we need separate PRs here.

One idea for the future can be get the versions from here: https://www.postgresql.org/versions.rss

svenklemm:
> > > @svenklemm wouldn't better send those refactorings in another PR??? Because we approved it based on the $SUBJECT and $DESCRIPTION
> > 
> > 
> > These are related to subject and will cut done the number of places we have to adjust when doing another bump for the next postgres version in a couple months. So I dont think we need separate PRs here.
> 
> One idea for the future can be get the versions from here: https://www.postgresql.org/versions.rss

Unfortunately release of a new postgres version does not mean it is packaged everywhere. Atm this PR is blocked because docker images for the new version are still not yet available after almost a week of tagging 14.3.

PR #4330, opened 2022-05-13
Add GUC "bgw_launcher_poll_time"

### LINKED ISSUE #4217: [Bug]: Compression policy job delaying to start
### What type of bug is this?

Configuration

Background worker, Compression, Configuration, Policy

### What happened?

Hello all,

In the company I work for, we are currently using TimescaleDB version 2.5 with a Docker environment for testing purposes (latest-pg12 tag).

We have a test environment that runs a `JDBCDatabaseContainer` from `testcontainers` and we perform some integration tests on it.

Some of our tests add a compression policy using the process described [here](https://docs.timescale.com/api/latest/compression/add_compression_policy/#add-compression-policy).

As we have a hypertable with a `bigint` timestamp, our test also set the `integer_now_func`, as described in the docs.

Briefly, our test sets the compression policy on a given hypertable, and try to verify the `timescaledb_information.job_stats` view, in order to check the status of the job. Basically, if the job status is Success, then the test checks for compressed chunks and finishes.

The strange behavior comes because in this test scenario, as soon as we call `add_compression_policy`, the job can be found at 
`timescaledb_information.jobs` but has a null value on `next_start`. Also, it does not show up on `timescaledb_information.job_stats`. Exactly after 1 minute, the `next_start` value gets set correctly and the job runs, making it possible to find it on `timescaledb_information.job_stats`.

We are trying to understand where this "1 minute delay" comes from. We have already tried to increase the number of background workers / parallel jobs, with no success.

We have also tried to use `alter_job()` to set a specific `next_start` value, but nothing happens. The job runs only after 1 minute of having set the `add_compression_policy()`. Same way, we have tried to call `run_job`, passing the job_id found on `timescaledb_information.jobs`. The jobs actually gets executed, but nothing changes on those views.

The big question is: Is this behavior expected? If so, is there a way we can change this "delay" (by setting an environment variable, for example)?

One interesting fact: If we call the `add_compression_policy` on an environment with docker/timescale that already has a database created, in our local machine, the job gets run instantly and those views get updated. This "1 minute delay" happens only on a container built with testcontainers for testing purposes.

Important to say that before each test is run, we create a new database / hypertable and enable the compression. This led us to guess there might be some "minimal" time after the database creation which the background jobs must wait before running, but we are not sure. As said above, this delay does not happen on a docker environment where the database / hypertables are already created.

We tried to find more information about how the background workers / scheduler work in the timescale docs, but no luck atm.

Any help on this matter would be much appreciated. If there is any information / log we can send to help understanding this issue, please let me know.

Example of [JDBCDatabaseContainer](https://javadoc.io/doc/org.testcontainers/postgresql/1.11.1/org/testcontainers/containers/PostgreSQLContainer.html) (currently we implemented our own TimescaleDBContainer based on this)

[JDBC Testcontainers](https://www.testcontainers.org/modules/databases/jdbc/)

2.5.0

### PostgreSQL version used

12

Ubuntu 21.10

Docker

On prem/Self-hosted

_No response_

_No response_

konskov:
This GUC permits configuring the TIMEOUT parameter
for the background worker launcher in the loader.

Fixes #4217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ If you use compression with a non-default collation on a segmentby-column you mi
 * #4169 Add support for chunk exclusion on DELETE to PG14
 * #4209 Add support for chunk exclusion on UPDATE to PG14
 * #4301 Add support for bulk inserts in COPY operator
+* #4330 Add GUC "bgw_launcher_poll_time"
 
 **Bugfixes**
 * #3899 Fix segfault in Continuous Aggregates
@@ -21,6 +22,7 @@ If you use compression with a non-default collation on a segmentby-column you mi
 * #4236 Fix potential wrong order of results for compressed hypertable with a non-default collation
 * #4255 Fix option "timescaledb.create_group_indexes"
 * #4300 Fix refresh window cap for cagg refresh policy
+* #4330 Add GUC "bgw_launcher_poll_time"
 
 **Thanks**
 * @jsoref for fixing various misspellings in code, comments and documentation
--- a/src/loader/bgw_launcher.c
+++ b/src/loader/bgw_launcher.c
@@ -82,13 +82,6 @@ typedef enum SchedulerState
 #define BGW_LAUNCHER_RESTART_TIME_S 60
 #endif
 
-/* WaitLatch expects a long */
-#ifdef TS_DEBUG
-#define BGW_LAUNCHER_POLL_TIME_MS 10L
-#else
-#define BGW_LAUNCHER_POLL_TIME_MS 60000L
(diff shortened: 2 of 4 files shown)


mfundul (commented):
You should probably use more verbose names than "launcher poll time", since it is actually the "background worker launcher".

mfundul (approved):

mfundul:
> You should probably use more verbose names than "launcher poll time", since it is actually the "background worker launcher".

Please also fix the commit message and add this to the changelog.

fabriziomello (approved):

PR #4340, opened 2022-05-16
Enable now() usage in plan-time chunk exclusion

svenklemm:
This implements an optimization to allow now() expression to be
used during plan time chunk exclusions. Since now() is stable it
would not normally be considered for plan time chunk exclusion.
To enable this behaviour we convert `column > now()` expressions
into `column > const AND column > now()`. Assuming that time
always moves forward this is save even for prepared statements.
This optimization works for SELECT, UPDATE and DELETE.
On hypertables with many chunks this can lead to a considerable
speedup for certain queries.

The following expressions are supported:
- column > now()
- column >= now()
- column > now() - Interval
- column > now() + Interval
- column >= now() - Interval
- column >= now() + Interval

Interval must not have a day or month component as those depend
on timezone settings.

Some microbenchmark to show the improvements, I did best of five
for all of the queries.

-- hypertable with 1k chunks
-- with optimization
select * from metrics1k where time > now() - '5m'::interval;
Time: 3.090 ms

-- without optimization
select * from metrics1k where time > now() - '5m'::interval;
Time: 145.640 ms

-- hypertable with 5k chunks
-- with optimization
select * from metrics5k where time > now() - '5m'::interval;
Time: 4.317 ms

-- without optimization
select * from metrics5k where time > now() - '5m'::interval;
Time: 775.259 ms

-- hypertable with 10k chunks
-- with optimization
select * from metrics10k where time > now() - '5m'::interval;
Time: 4.853 ms

-- without optimization
select * from metrics10k where time > now() - '5m'::interval;
Time: 1766.319 ms (00:01.766)

-- hypertable with 20k chunks
-- with optimization
select * from metrics20k where time > now() - '5m'::interval;
Time: 6.141 ms

-- without optimization
select * from metrics20k where time > now() - '5m'::interval;
Time: 3321.968 ms (00:03.322)

Speedup with 1k chunks: 47x
Speedup with 5k chunks: 179x
Speedup with 10k chunks: 363x
Speedup with 20k chunks: 540x

### DIFF
--- a/src/planner/add_hashagg.c
+++ b/src/planner/add_hashagg.c
@@ -27,7 +27,7 @@
 #include "estimate.h"
 
 /* This optimization adds a HashAggregate plan to many group by queries.
- * In plain postgres, many time-series queries will not use a the hash aggregate
+ * In plain postgres, many time-series queries will not use a hash aggregate
  * because the planner will incorrectly assume that the number of rows is much larger than
  * it actually is and will use the less efficient GroupAggregate instead of a HashAggregate
  * to prevent running out of memory.
--- a/src/planner/constify_now.c
+++ b/src/planner/constify_now.c
@@ -1,0 +1,213 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+
+#include <postgres.h>
+#include <access/xact.h>
+#include <datatype/timestamp.h>
+#include <nodes/makefuncs.h>
+#include <optimizer/optimizer.h>
+#include <utils/fmgroids.h>
+
+#include "cache.h"
+#include "dimension.h"
+#include "hypertable.h"
+#include "hypertable_cache.h"
+#include "planner.h"
+
+/*
+ * This implements an optimization to allow now() expression to be
+ * used during plan time chunk exclusions. Since now() is stable it
+ * would not normally be considered for plan time chunk exclusion.
+ * To enable this behaviour we convert `column > now()` expressions
+ * into `column > const AND column > now()`. Assuming that times
+ * always moves forward this is safe even for prepared statements.
+ *
+ * We consider the following expressions valid for this optimization:
+ * - Var > now()
+ * - Var >= now()
+ * - Var > now() - Interval
+ * - Var > now() + Interval
+ * - Var >= now() - Interval
+ * - Var >= now() + Interval
+ *
+ * Additionally Interval needs to be Const and not contain day or month
+ * components as those would be affected by timezone, which can change
+ * between executions of a prepared statement.
+ */
+static const Dimension *
+get_hypertable_dimension(Oid relid)
+{
+	Hypertable *ht = ts_planner_get_hypertable(relid, CACHE_FLAG_CHECK);
+	if (!ht)
+		return NULL;
+	return hyperspace_get_open_dimension(ht->space, 0);
+}
+
+static bool
+is_valid_now_expr(OpExpr *op, List *rtable)
+{
+	/* Var > or Var >= */
+	if ((op->opfuncid != F_TIMESTAMPTZ_GT && op->opfuncid != F_TIMESTAMPTZ_GE) ||
+		!IsA(linitial(op->args), Var))
+		return false;
+
+	/*
+	 * Check that the constraint is actually on a partitioning
+	 * column. We only check for match on first open dimension
+	 * because that will be the time column.
+	 */
+	Var *var = linitial_node(Var, op->args);
+	Assert(var->varno <= list_length(rtable));
+	RangeTblEntry *rte = list_nth(rtable, var->varno - 1);
+
+	const Dimension *dim = get_hypertable_dimension(rte->relid);
+	if (!dim || dim->fd.column_type != TIMESTAMPTZOID || dim->column_attno != var->varattno)
+		return false;
+
+	/* Var > now() or Var >= now() */
+	if (IsA(lsecond(op->args), FuncExpr) && lsecond_node(FuncExpr, op->args)->funcid == F_NOW)
+		return true;
+
+	if (!IsA(lsecond(op->args), OpExpr))
+		return false;
+
+	/* Var >|>= now() +|- Const */
+	OpExpr *op_inner = lsecond_node(OpExpr, op->args);
+	if ((op_inner->opfuncid != F_TIMESTAMPTZ_MI_INTERVAL &&
+		 op_inner->opfuncid != F_TIMESTAMPTZ_PL_INTERVAL) ||
+		!IsA(linitial(op_inner->args), FuncExpr) ||
+		linitial_node(FuncExpr, op_inner->args)->funcid != F_NOW ||
+		!IsA(lsecond(op_inner->args), Const))
+		return false;
+
+	/*
+	 * The consttype check should not be necessary since the
+	 * operators we whitelist above already mandates it.
+	 */
+	Const *c = lsecond_node(Const, op_inner->args);
+	Assert(c->consttype == INTERVALOID);
+	if (c->constisnull || c->consttype != INTERVALOID)
+		return false;
+
+	Interval *offset = DatumGetIntervalP(c->constvalue);
+	/*
+	 * We don't consider day or month intervals safe here as
+	 * they are affected by timezones and therefore not
+	 * safe to evaluate during planning.
+	 */
+	if (offset->day != 0 || offset->month != 0)
+		return false;
+
+	return true;
+}
+
+static Const *
+make_now_const()
+{
+	return makeConst(TIMESTAMPTZOID,
+					 -1,
+					 InvalidOid,
+					 sizeof(TimestampTz),
+#ifdef TS_DEBUG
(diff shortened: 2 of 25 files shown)


src/planner/planner.c
```diff
@@ -282,18 +291,30 @@ typedef struct
...
 	if (node == NULL)
 		return false;
 
-	if (IsA(node, Query))
+	if (ts_guc_enable_now_constify && IsA(node, FromExpr))
+	{
+		FromExpr *from = castNode(FromExpr, node);
+		if (from->quals)
+		{
+			from->quals = ts_constify_now(context->root, from->quals);
```
akuzm, line 306: We already have a more generic variant here: https://github.com/timescale/timescaledb/blob/main/tsl/src/fdw/scan_plan.c#L552-L563
We use it to evaluate stable constraints at the access node before sending them to the data nodes. Can we just use it here?

akuzm: That code evaluates many timestamp-related stable functions, not just `now()`: https://github.com/timescale/timescaledb/blob/main/tsl/src/fdw/deparse.c#L118-L180

akuzm: By the way, this is for local hypertables, right? This means we don't need a whitelist and can evaluate all stable functions before performing constraint exclusion. Although I'm not sure if `preprocess_queries` is the right place to do this, because this happens at planning time, and the stable functions should be evaluated at execution time.

svenklemm: This is still a draft so i'll expand the commit message but you cannot just constify now() like that at plan time that would not be safe. What this patch does is transform `column > now() - '5m'` into `column > '2022-05-16 14:49' AND column > now() - '5min'`. This allows the constified value to be used during plan time chunk exclusion while still returning correct results even for prepared statements.

akuzm: Oh, nice idea!

src/planner/add_hashagg.c
```diff
@@ -27,7 +27,7 @@
 #include "estimate.h"
 
 /* This optimization adds a HashAggregate plan to many group by queries.
- * In plain postgres, many time-series queries will not use a the hash aggregate
```
nikkhils, line 30: The typo is still not fixed :-)

src/planner/constify_now.c
```diff
@@ -0,0 +1,202 @@
...
+#include "hypertable_cache.h"
+#include "planner.h"
+
+/*
+ * This implements an optimization to allow now() expression to be
+ * used during plan time chunk exclusions. Since now() is stable it
+ * would not normally be considered for plan time chunk exclusion.
+ * To enable this behaviour we convert `column > now()` expressions
+ * into `column > const AND column > now()`. Assuming that times
+ * always moves forward this is save even for prepared statements.
```
nikkhils, line 26: save -> safe

src/planner/constify_now.c
```diff
@@ -0,0 +1,202 @@
...
+make_now_const()
+{
+	return makeConst(TIMESTAMPTZOID,
+					 -1,
+					 InvalidOid,
+					 sizeof(TimestampTz),
+#ifdef TS_DEBUG
+					 ts_get_mock_time_or_current_time(),
+#else
+					 TimestampTzGetDatum(GetCurrentTransactionStartTimestamp()),
```
nikkhils, line 116: Hmm, we use transaction start time here?

src/planner/constify_now.c
```diff
@@ -0,0 +1,202 @@
...
+ * op will be OpExpr with Var > now() - Expr
+ */
+static OpExpr *
+constify_now_expr(PlannerInfo *root, OpExpr *op)
+{
+	OpExpr *constified = copyObject(op);
+	if (IsA(lsecond(op->args), FuncExpr))
+	{
+		/*
+		 * Sanity check that this is a supported expression. We should never
```
nikkhils, line 133: Will this handle the case correctly for below:

`col1 > now() OR col2 > 1234`

Will this get properly converted into:

`(col1 > timestamp AND col1 > now()) OR col2 > 1234`?

tsl/test/shared/expected/constify_now.out
```diff
@@ -0,0 +1,462 @@
...
+-- all of these should have a constified value as filter
+-- none of these initial tests will actually exclude chunks
+-- because we want to see the constified now expression in
+-- EXPLAIN output
+:PREFIX SELECT FROM metrics WHERE time > now();
+                                                   QUERY PLAN                                                   
+----------------------------------------------------------------------------------------------------------------
+ Append
+   ->  Index Only Scan using _hyper_X_X_chunk_metrics_time_idx on _hyper_X_X_chunk
+         Index Cond: (("time" > now()) AND ("time" > 'Mon Jan 01 00:00:00 1990 PST'::timestamp with time zone))
```
nikkhils, line 18: Do we gain anything if we add `("time" > 'Mon Jan 01 00:00:00 1990 PST'::timestamp with time zone)` before the `now()` part?

src/planner/constify_now.c
svenklemm: Thats what now uses:
```
Datum
now(PG_FUNCTION_ARGS)
{
  PG_RETURN_TIMESTAMPTZ(GetCurrentTransactionStartTimestamp());
}
```

src/planner/constify_now.c
svenklemm: this PR only optimizes top-level ANDed constraints

tsl/test/shared/expected/constify_now.out
```diff
@@ -0,0 +1,462 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+SET timescaledb.enable_chunk_append TO false;
+SET timescaledb.enable_constraint_aware_append TO false;
+SET timescaledb.current_timestamp_mock TO '1990-01-01';
```
akuzm, line 6: We also have the `tsl_override_current_timestamptz` thing, need to merge them sometime I guess...

src/planner/add_hashagg.c
svenklemm: It is for me are you looking at stale version?

src/planner/constify_now.c
```diff
@@ -0,0 +1,202 @@
...
+ * We consider the following expressions valid for this optimization:
+ * - Var > now()
+ * - Var >= now()
+ * - Var > now() - Interval
+ * - Var > now() + Interval
+ * - Var >= now() - Interval
+ * - Var >= now() + Interval
+ *
+ * Additionally Interval needs to be Const and not contain day or month
+ * components as those would be affected by timezone.
```
akuzm, line 37:
```suggestion
 * Additionally Interval needs to be Const and not contain day or month
 * components as those would be affected by timezone, which can change
 * between executions of a prepared statement.
```

I'd add something like this.

src/planner/constify_now.c
```diff
@@ -0,0 +1,202 @@
...
+#include "cache.h"
+#include "dimension.h"
+#include "hypertable.h"
+#include "hypertable_cache.h"
+#include "planner.h"
+
+/*
+ * This implements an optimization to allow now() expression to be
+ * used during plan time chunk exclusions. Since now() is stable it
+ * would not normally be considered for plan time chunk exclusion.
```
akuzm, line 23: Is it currently used for runtime chunk exclusion? But for some reason it's much less effective than plan time chunk exclusion, so that's why this PR gives huge speedup?

svenklemm: correct

tsl/test/shared/expected/constify_now.out
svenklemm: doesnt change much for the microbenchmarks i did but current variant should be better optimized for the list operation.

akuzm (approved):
Tested locally, gives some impressive speedup.

tsl/test/shared/expected/constify_now-12.out
```diff
@@ -0,0 +1,462 @@
...
+-- all of these should have a constified value as filter
+-- none of these initial tests will actually exclude chunks
+-- because we want to see the constified now expression in
+-- EXPLAIN output
+:PREFIX SELECT FROM metrics WHERE time > now();
+                                                   QUERY PLAN                                                   
+----------------------------------------------------------------------------------------------------------------
+ Append
+   ->  Index Only Scan using _hyper_X_X_chunk_metrics_time_idx on _hyper_X_X_chunk
+         Index Cond: (("time" > now()) AND ("time" > 'Mon Jan 01 00:00:00 1990 PST'::timestamp with time zone))
```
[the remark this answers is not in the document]
svenklemm: Not sure what you are referring to but at the level the optimization is happening there are no scans yet. We need this all to happen way before anything scan related is there because we want to get rid of the planning overhead of excluded chunks.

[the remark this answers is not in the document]
svenklemm: Oh i think I understand what you mean but that would limit the usefulness of the optimization. This optimization works with both our own hypertable expansion and postgres inheritance expansion. That means it applies to SELECT/UPDATE/DELETE. If we would do this only in our own hypertable expansion it would no longer work  for UPDATE/DELETE

gayyappan (approved):

Issue #4266, opened 2022-04-22
[Bug]: Periodic background worker crash and DB restart


abrownsword:
### What type of bug is this?

Crash

### What subsystems and features are affected?

Background worker

### What happened?

2-3 times per day my database logs this and then goes into recovery mode and restarts:

```
2022-04-15 15:15:24.112 UTC [11443] WARNING:  failed to launch job 1005 "Refresh Continuous Aggregate Policy [1005]": out of background workers
2022-04-15 15:15:24.112 UTC [11443] WARNING:  failed to launch job 1007 "Refresh Continuous Aggregate Policy [1007]": out of background workers
2022-04-15 15:15:24.117 UTC [11443] WARNING:  failed to launch job 1007 "Refresh Continuous Aggregate Policy [1007]": out of background workers
2022-04-15 15:15:24.122 UTC [30121] LOG:  refreshing continuous aggregate "environ_summary_daily" in window [ 2022-03-16 00:00:00+00, 2022-04-14 00:00:00+00 ]
2022-04-15 15:15:24.131 UTC [30124] LOG:  refreshing continuous aggregate "vsensor_summary_daily" in window [ 2022-03-16 00:00:00+00, 2022-04-14 00:00:00+00 ]
2022-04-15 15:15:24.131 UTC [30122] LOG:  refreshing continuous aggregate "monitor_summary_daily" in window [ 2022-03-16 00:00:00+00, 2022-04-14 00:00:00+00 ]
2022-04-15 15:15:24.132 UTC [30123] LOG:  refreshing continuous aggregate "state_summary_daily" in window [ 2022-03-16 00:00:00+00, 2022-04-14 00:00:00+00 ]
2022-04-15 15:15:24.527 UTC [30125] LOG:  refreshing continuous aggregate "power_summary_daily" in window [ 2022-03-16 00:00:00+00, 2022-04-14 00:00:00+00 ]
2022-04-15 15:42:02.743 UTC [2442] LOG:  received fast shutdown request
2022-04-15 15:42:02.746 UTC [2442] LOG:  aborting any active transactions
terminating TimescaleDB job scheduler due to administrator command
2022-04-15 15:42:02.747 UTC [11443] FATAL:  terminating background worker "TimescaleDB Background Worker Scheduler" due to administrator command
2022-04-15 15:42:02.747 UTC [30248] grafana@grafana FATAL:  terminating connection due to administrator command
[...]

PR #4358, opened 2022-05-19
Fix crash and other issues in telemetry reporter 

erimatnor:
Make the following changes to the telemetry reporter background
worker:

- Add a read lock to the current relation that the reporter collects
  stats for. This lock protects against concurrent deletion of the
  relation, which could lead to errors that would prevent the reporter
  from completing its report.
- Set an active snapshot in the telemetry background process for use
  when scanning a relation for stats collection.

- Reopen the scan iterator when collecting chunk compression stats for
  a relation instead of keeping it open and restarting the scan. The
  previous approach seems to cause crashes due to memory corruption of
  the scan state. Unfortunately, the exact cause has not been
  identified, but the change has been verified to work on a live
  running instance (thanks to @abrownsword for the help with
  reproducing the crash and testing fixes).

Fixes https://github.com/timescale/timescaledb/issues/4266

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,11 @@ If you use compression with a non-default collation on a segmentby-column you mi
 * #4269 Fix bad Continuous Aggregate view definition reported in #4233
 * #4300 Fix refresh window cap for cagg refresh policy
 * #4330 Add GUC "bgw_launcher_poll_time"
+* #4358 Fix crash and other issues in telemetry reporter
 
 **Thanks**
 * @jsoref for fixing various misspellings in code, comments and documentation
+* @abrownsword for reporting a bug in the telemetry reporter and testing the fix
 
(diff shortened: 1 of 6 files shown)


erimatnor:
This PR depends on #4349, which should be merged first.

svenklemm (approved):

fabriziomello (approved):

tsl/src/compression/compress_utils.c
```diff
@@ -12,6 +12,7 @@
 #include <catalog/dependency.h>
 #include <commands/tablecmds.h>
 #include <commands/trigger.h>
+#include <commands/event_trigger.h>
```
mfundul, line 15: Does this do anything here?

erimatnor: Nope, left-over from a change I reverted. Will remove.

PR #4359, opened 2022-05-19
Create composite index on segmentby columns

### LINKED ISSUE #4314: [Enhancement]: Optimize index creation for compressed hypertables

Performance

Compression

Instead of creating an index for every segmentby column we should create a composite index all the segmentby columns. This would improve space efficiency and query performance. 

### Implementation challenges

_No response_

konskov:
Previously we created one index per segmentby column, of the form
`(col, _ts_meta_sequence_num)`. 
Compressed data is ordered by segmentby, then by orderby within
the segments and lastly by the sequence number of the batch. So if
segmentby columns are missing from the index, that index cannot be
used to produce ordered output, requiring an extra sort step.

A composite index containing all segmentby columns removes the 
additional sort step and gives better plans. 

Fixes #4314

### DIFF
--- a/tsl/src/compression/create.c
+++ b/tsl/src/compression/create.c
@@ -424,42 +424,67 @@ create_compressed_table_indexes(Oid compresstable_relid, CompressColInfo *compre
 		.name = COMPRESSION_COLUMN_METADATA_SEQUENCE_NUM_NAME,
 	};
 	int i;
+	NameData index_name;
+	ObjectAddress index_addr;
+	HeapTuple index_tuple;
+	List *indexcols = NIL;
+
+	StringInfo buf = makeStringInfo();
+
 	for (i = 0; i < compress_cols->numcols; i++)
 	{
-		NameData index_name;
-		ObjectAddress index_addr;
-		HeapTuple index_tuple;
 		FormData_hypertable_compression *col = &compress_cols->col_meta[i];
-		IndexElem segment_elem = { .type = T_IndexElem, .name = NameStr(col->attname) };
+
+		IndexElem *segment_elem = makeNode(IndexElem);
 
 		if (col->segmentby_column_index <= 0)
 			continue;
 
-		stmt.indexParams = list_make2(&segment_elem, &sequence_num_elem);
-		index_addr = DefineIndex(ht->main_table_relid,
-								 &stmt,
-								 InvalidOid, /* IndexRelationId */
-								 InvalidOid, /* parentIndexId */
-								 InvalidOid, /* parentConstraintId */
-								 false,		 /* is_alter_table */
-								 false,		 /* check_rights */
-								 false,		 /* check_not_in_use */
-								 false,		 /* skip_build */
-								 false);	 /* quiet */
-		index_tuple = SearchSysCache1(RELOID, ObjectIdGetDatum(index_addr.objectId));
+		segment_elem->name = pstrdup(NameStr(col->attname));
 
-		if (!HeapTupleIsValid(index_tuple))
-			elog(ERROR, "cache lookup failed for index relid %u", index_addr.objectId);
-		index_name = ((Form_pg_class) GETSTRUCT(index_tuple))->relname;
-		elog(DEBUG1,
-			 "adding index %s ON %s.%s USING BTREE(%s, %s)",
-			 NameStr(index_name),
-			 NameStr(ht->fd.schema_name),
-			 NameStr(ht->fd.table_name),
-			 NameStr(col->attname),
-			 COMPRESSION_COLUMN_METADATA_SEQUENCE_NUM_NAME);
-		ReleaseSysCache(index_tuple);
+		/* if this isn't the first element, add a ',' before appending it */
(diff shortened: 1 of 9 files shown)


akuzm:
Why is this beneficial? If we have two segmentby columns, and there's a filter only on second one, this index can't be used. Wouldn't that degrade performance, compared to having a separate index for each column?

konskov:
Indeed, it would not be beneficial in terms of query performance then. It would improve space efficiency, but there are tradeoffs and I'm not sure what's preferable. Let me link the related issue and some very basic testing I've done

konskov:
Tested with two datasets, one good for compression (follows our rule of thumb of at least 100 rows per chunk per segment), one that doesn't.

## Good dataset
```
create table metrics2 (
    time timestamp with time zone not null, 
    device_id int,
    metric_id int,
    value double precision
);

select create_hypertable('metrics2', 'time');

alter table metrics2 set (timescaledb.compress=true, timescaledb.compress_segmentby='device_id, metric_id');

-- 25 million records, approx. 300 rows per segment/chunk. this will give a bitmap heap scan in the plan
insert into metrics2(time,device_id, metric_id) select time, device_id, metric_id from  
(SELECT floor(random()*90) as device_id, floor(random()*20 + 1) as metric_id, 
generate_series(now() - interval '1 month', now(), interval '100 milliseconds') as time) q1;
 
update metrics set (value) = (select ceil(random()*metric_id*50) );

select compress_chunk(show_chunks('metrics'));

```
Results
```
-- current code - single index per segmentby column
 index_after | index_before | table_after | table_before | total_after | total_before | toast_after | toast_before 
-------------+--------------+-------------+--------------+-------------+--------------+-------------+--------------
 1632 kB     | 998 MB       | 6152 kB     | 1290 MB      | 168 MB      | 2288 MB      | 161 MB      | 0 bytes
 
-- composite index
 index_after | index_before | table_after | table_before | total_after | total_before | toast_after | toast_before 
-------------+--------------+-------------+--------------+-------------+--------------+-------------+--------------
 1104 kB     | 998 MB       | 6152 kB     | 1290 MB      | 168 MB      | 2288 MB      | 161 MB      | 0 bytes

```

konskov:
Closing this PR since I can't tell how useful/beneficial it would be, if at all.

svenklemm:
Currently we create on index (segment_by_column, _ts_meta_sequence_num). This index is only useful if there is a single segment by column. For multiple segment by columns a combined index would be much more useful. In addition to that having the _ts_meta_sequence_num in the index when multiple columns are present is just wasted space as the index could never be used for ordering since that depends on all segmentby columns being present. While it is true that for queries on non-leading segmentby columns the index might not be used this is still a better default and if users want to optimize those queries they can add additional indexes.

tsl/src/compression/create.c
```diff
@@ -424,43 +424,69 @@ create_compressed_table_indexes(Oid compresstable_relid, CompressColInfo *compre
...
-			 NameStr(ht->fd.schema_name),
-			 NameStr(ht->fd.table_name),
-			 NameStr(col->attname),
-			 COMPRESSION_COLUMN_METADATA_SEQUENCE_NUM_NAME);
-		ReleaseSysCache(index_tuple);
+		segment_elem->name = pstrdup(NameStr(col->attname));
+
+		/* if this isn't the first element, add a ',' before appending it */
+		if (list_length(indexcols) > 0)
+			appendStringInfoString(buf, sep);
```
mfundul, line 448: nit:
```suggestion
			appendStringInfoString(buf, ", ");
```
Our codebase is full of using string literals as separators instead of a variable, so that would make it more consistent with the rest of the code.

Moreover it is more readable, since you don't have to guess what value `sep` may contain.

Finally, string literals don't occupy more memory when they have identical contents because they are optimized by the compiler, so we're not saving anything here.

tsl/src/compression/create.c
```diff
@@ -424,43 +424,69 @@ create_compressed_table_indexes(Oid compresstable_relid, CompressColInfo *compre
...
+		indexcols = lappend(indexcols, segment_elem);
 	}
 
+	if (list_length(indexcols) == 0)
+	{
+		ts_cache_release(hcache);
+		return;
+	}
+
+	appendStringInfoString(buf, sep);
```
mfundul, line 460:
```suggestion
	appendStringInfoString(buf, ", ");
```

tsl/src/compression/create.c
```diff
@@ -424,43 +424,69 @@ create_compressed_table_indexes(Oid compresstable_relid, CompressColInfo *compre
 		.name = COMPRESSION_COLUMN_METADATA_SEQUENCE_NUM_NAME,
 	};
 	int i;
+	NameData index_name;
+	ObjectAddress index_addr;
+	HeapTuple index_tuple;
+	List *indexcols = NIL;
+
+	StringInfo buf = makeStringInfo();
+	char *sep = ", ";
```
mfundul, line 433:
```suggestion
```

mfundul (approved):
The commit message should be a little more verbose.

PR #4367, opened 2022-05-20
Improved buffer management in the copy operator

jnidzwetzki:
The multi-buffer copy optimization creates a multi-insert buffer per
`ChunkInsertState`. However, chunks can be closed. When
`ts_chunk_dispatch_get_chunk_insert_state` is called for a closed chunk
again, a new `ChunkInsertState` is returned. So far, also a new
multi-insert buffer has been created. Therefore, multiple batch
operations could be executed per chunk, which reduces the efficiency
of the optimization.

This patch introduces an `HTAB` that maps from the `chunk_id` to the
multi-insert buffer. Even when a chunk is closed, all tuples for
a chunk are stored in the same buffer.

### DIFF
--- a/src/copy.c
+++ b/src/copy.c
@@ -50,6 +50,7 @@
 #include <utils/builtins.h>
 #include <utils/elog.h>
 #include <utils/guc.h>
+#include <utils/hsearch.h>
 #include <utils/lsyscache.h>
 #include <utils/rel.h>
 #include <utils/rls.h>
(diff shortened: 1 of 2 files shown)


src/copy.c
```diff
@@ -102,7 +103,8 @@ typedef struct TSCopyMultiInsertBuffer
  */
 typedef struct TSCopyMultiInsertInfo
```
gayyappan, line 104: Could you add the comment here that explains why it is useful to maintain chunk-id to buffer mapping - especially since the code will no longer have the comment at line 897.

jnidzwetzki: @gayyappan Thank you for pointing this out. I improved the comment.

svenklemm (approved):

src/copy.c
```diff
@@ -406,54 +458,40 @@ TSCopyMultiInsertBufferCleanup(TSCopyMultiInsertInfo *miinfo, TSCopyMultiInsertB
...
 		TSCopyMultiInsertBufferCleanup(miinfo, buffer);
 	}
 
-	list_free(miinfo->multiInsertBuffers);
-	miinfo->multiInsertBuffers = NIL;
+	/* All existing buffers are flushed and the multi-insert states
+	 * are freed. So, delete old hash map and create a new one for further
+	 * inserts.
+	 */
+	hash_destroy(miinfo->multiInsertBuffers);
```
akuzm, line 489: TSCopyMultiInsertInfoDestroy? TBH I'd delete that function, it's one line.

akuzm (approved):
LGTM.

Another thought for furhter improvements -- it makes sense to make the size of both caches the same, i.e. check for `(list_length(miinfo->multiInsertBuffers) >= ts_guc_max_open_chunks_per_insert` in `TSCopyMultiInsertInfoIsFull`. I've seen cache thrashing in some benchmarks with the current code: when the rows are small, we accumulate a lot of multi-insert buffers, more than the chunk insert state cache.  And then when we flush we have to set up each chunk insert state again because they are not in cache anymore.

src/copy.c
jnidzwetzki: I removed the function

PR #4368, opened 2022-05-20
Skip tablespace test on PG12.0

svenklemm:
The tablespace test causes a segfault on PG12.0 due to an upstream
bug in the event trigger handling.

### DIFF
--- a/scripts/gh_matrix_builder.py
+++ b/scripts/gh_matrix_builder.py
@@ -133,10 +133,10 @@ if event_type != "pull_request":
 
   # add debug test for first supported PG12 version
   # most of the IGNORES are the isolation tests because the output format has changed between versions
-  # chunk_utils is skipped because of a use after free bug in postgres 12.0 which one of our tests hit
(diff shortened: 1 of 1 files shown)


mfundul (approved):

konskov (approved):

gayyappan (approved):

scripts/gh_matrix_builder.py
```diff
@@ -133,10 +133,10 @@ def macos_config(overrides):
 
   # add debug test for first supported PG12 version
   # most of the IGNORES are the isolation tests because the output format has changed between versions
-  # chunk_utils is skipped because of a use after free bug in postgres 12.0 which one of our tests hit
+  # chunk_utils and tablespace is skipped because of use after free bugs in postgres 12.0 which those tests hit
```
erimatnor, line 136:
```suggestion
  # chunk_utils and tablespace are skipped because of use after free bugs in postgres 12.0 which those tests hit
```

PR #4369, opened 2022-05-20
Show warning for caggs with numeric

svenklemm:
Postgres changes the internal state format for numeric aggregates
which we materialize in caggs in PG14. This will invalidate the
affected columns when upgrading from PG13 to PG14. This patch
adds a warning to the update script when we encounter this
configuration.

### DIFF
--- a/sql/updates/post-update.sql
+++ b/sql/updates/post-update.sql
@@ -96,3 +96,30 @@ UPDATE pg_class cl SET relacl = tmpacl
   FROM to_update WHERE cl.oid = to_update.oid;
 
 DROP TABLE _timescaledb_internal.saved_privs;
+
+-- warn about partial storage format change for numeric
+DO $$
+DECLARE
+  cagg_name text;
+  cagg_column text;
+  cnt int := 0;
+BEGIN
+  IF current_setting('server_version_num')::int <  140000 THEN
+    FOR cagg_name, cagg_column IN
+      SELECT
+        attrelid::regclass::text,
+        att.attname
+      FROM _timescaledb_catalog.continuous_agg cagg
+      INNER JOIN pg_attribute att ON (
+        att.attrelid = format('%I.%I',cagg.user_view_schema,cagg.user_view_name)::regclass AND
+        atttypid = 'numeric'::regtype)
+      WHERE cagg.finalized = false
+    LOOP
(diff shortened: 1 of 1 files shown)


mfundul:
Let's wait for me to reproduce this first before we merge this PR, to understand and verify the nature of the problem.

mfundul:
> This is the error you will get an affected caggs:
> 
> ```
> sven@cagg_num[3448944]=# select * from cagg ;
> ERROR:  08P01: insufficient data left in message
> LOCATION:  pq_copymsgbytes, pqformat.c:533
> Time: 11.277 ms
> ```

You had to do it first didn't you 😄

sql/updates/post-update.sql
```diff
@@ -96,3 +96,30 @@ UPDATE pg_class cl SET relacl = tmpacl
...
+  IF current_setting('server_version_num')::int <  140000 THEN
+    FOR cagg_name, cagg_column IN
+      SELECT
+        attrelid::regclass::text,
+        att.attname
+      FROM _timescaledb_catalog.continuous_agg cagg
+      INNER JOIN pg_attribute att ON (
+        att.attrelid = format('%I.%I',cagg.user_view_schema,cagg.user_view_name)::regclass AND
+        atttypid = 'numeric'::regtype)
+      WHERE cagg.finalized = false
```
mfundul, line 116: Is this safe to do? What happens if we downgrade and upgrade again?

mfundul (approved):

mkindahl (commented):
The change here looks fine, but given the change in format, I am concerned about the case that the user upgrades from PG13 to PG14 and as a result uses the old numeric format. The `finalize_agg` function will, however, expect the new format (since it was built with PG14). If the user does not heed the warning (because it wasn't noticed, or for other reasons) and run the query anyway, Isn't there a risk of a segmentation fault in this case? I see a few alternative ways to prevent this, but not sure what is most feasible:

1. Block updating to PG14 if not using the finalized format. Not sure how easy this is to block.
2. Make sure that upgrading does not cause the queries to read the old format with the new version `finalize_agg` by, e.g., having different functions for different versions of PG.

mkindahl (approved):
After our discussions we concluded that we can provide the necessary functions to fix the format after the release so we will be able to unblock a user that accidentally upgrade to PG14 and for some reason cannot change to use finalized format. Issue #4376 is added to track this work.

fabriziomello (approved):

PR #4375, opened 2022-05-23
Don't ask for orderby column if default already set

### LINKED ISSUE #4331: [Bug]: alter table for compression has inconsistent/misleading error messages
### What type of bug is this?

Configuration

Compression, Configuration

### What happened?

We witnessed inconsistent behavior regarding the default value of timescaledb.compress_orderby, when the same alter table statement to turn on compression and set compression options are executed more than once. 

To reproduce the error: 

1. Create hypertable as below: 

```
CREATE TABLE events.abc_evt(   
 timestamp BIGINT NOT NULL,    
 live_timescale_event_id BIGSERIAL NOT NULL,    
 event jsonb NOT NULL,
segment jsonb NOT NULL);

SELECT create_hypertable ('events.abc_evt', 'timestamp', chunk_time_interval => 604800000);
```
2. Set the following compression settings: 

`ALTER TABLE events."abc_evt" set (timescaledb.compress=true , timescaledb.compress_segmentby = '"segment"');
`

3. Set the same compression setting again: 

`ALTER TABLE events."abc_evt" set (timescaledb.compress=true , timescaledb.compress_segmentby = '"segment"');
`

The last command above throws the following error: 

```
ERROR:  must specify a column to order by
DETAIL:  The timescaledb.compress_orderby option was previously set and must also be specified in the updated configuration.
SQL state: 0A000
```

Note that in both statements we did not specified column to order by. Just to be clear, the column to segmentby has nothing do with the error, as it also happens when we executed the following statement twice: 

```
ALTER TABLE events."abc_evt" set (timescaledb.compress=true) 
```

The documentation states `timescaledb.compress_orderby` is an optional parameter and "The default is the descending order of the hypertable's time column." However, when we explicit specified the column to orderby in the statement, but using its default value, the error does not happen: 

```
ALTER TABLE events."abc_evt" set (timescaledb.compress=true, timescaledb.compress_orderby = 'timestamp DESC', timescaledb.compress_segmentby = '"segment"' );
```

In all cases above (with or without timescaledb.compress_orderby), the compression options/setting were the same. We checked using the statement and its result is show in the image below:  
`select * from timescaledb_information.compression_settings;`

image

Trying to understand the error, we specified the time column as the column to segment by, by executing the following: 

```
ALTER TABLE events."abc_evt" set (timescaledb.compress=false);

ALTER TABLE events."abc_evt" set (timescaledb.compress=true, timescaledb.compress_segmentby = '"timestamp"');
```								  
Even without the orderby parameter, executing the last statement twice, the error does not occur. In addition, the compression settings shows a different result: 

image

In our understanding, as the column to order by is not specified in any execution, the settings should have been the same: for timestamp column, the column orderby_asc should be false and orderby_nullsfirst should be true, as this is default option. In addition, executing the compression setting without `timescaledb.compress_orderby` should have not thrown the stated error (as it does not occur when the default value is explicit used).  

2.5.1

### PostgreSQL version used

12.10

Ubuntu 20.04.4 LTS

Docker

On prem/Self-hosted

ERROR:  must specify a column to order by
DETAIL:  The timescaledb.compress_orderby option was previously set and must also be specified in the updated configuration.
SQL state: 0A000

Described with details above, but to resume: 

1. Create hypertable as below: 

```
CREATE TABLE events.abc_evt(   
 timestamp BIGINT NOT NULL,    
 live_timescale_event_id BIGSERIAL NOT NULL,    
 event jsonb NOT NULL,
segment jsonb NOT NULL);

SELECT create_hypertable ('events.abc_evt', 'timestamp', chunk_time_interval => 604800000);
```
2. Set the following compression settings: 

`ALTER TABLE events."abc_evt" set (timescaledb.compress=true , timescaledb.compress_segmentby = '"segment"');`

3. Set the same compression setting again: 

`ALTER TABLE events."abc_evt" set (timescaledb.compress=true , timescaledb.compress_segmentby = '"segment"');`

konskov:
When enabling compression on a hypertable, the orderby option can be
omitted, which will set the default value of "time DESC". 
However previously, when executing the same command twice not setting
orderby, the second time we would get an error that orderby was
previously set and must be specified.
For example executing
`alter table set (timescaledb.compress, timescaledb.segmentby = '..')`
 
The reason for that error was that it's unclear if no orderby means
leave as is, or if it means set the default value. 
But in the case where orderby is already set to the default value, 
there is no ambiguity and both cases are equivalent, so the default
value can be reset without giving an error. 

Fixes #4331

### DIFF
--- a/tsl/src/compression/create.c
+++ b/tsl/src/compression/create.c
@@ -847,7 +847,8 @@ validate_existing_constraints(Hypertable *ht, CompressColInfo *colinfo)
 }
 
 static void
-check_modify_compression_options(Hypertable *ht, WithClauseResult *with_clause_options)
+check_modify_compression_options(Hypertable *ht, WithClauseResult *with_clause_options,
+								 List *parsed_orderby_cols)
 {
 	bool compress_enable = DatumGetBool(with_clause_options[CompressEnabled].parsed);
 	bool compressed_chunks_exist;
@@ -864,9 +865,13 @@ check_modify_compression_options(Hypertable *ht, WithClauseResult *with_clause_o
 
 	/* Require both order by and segment by when altering if they were previously set because
 	 * otherwise it's not clear what the default value means: does it mean leave as-is or is it an
-	 * empty list. */
+	 * empty list.
+	 * In the case where orderby is already set to the default value (time DESC),
+	 * both interpretations should lead to orderby being set to the default value,
+	 * so we allow skipping orderby if the default is the already set value. */
 	if (compress_enable && compression_already_enabled)
 	{
+		List *current_orderby_cols = NIL;
 		List *info = ts_hypertable_compression_get(ht->fd.id);
 		ListCell *lc;
 		bool segment_by_set = false;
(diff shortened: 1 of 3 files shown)


svenklemm:
Hmm is this the behaviour we want? When not supplying orderby but i previously set orderby i would expect the previously set value to not change. Currently with this PR the value would get reset to default.

svenklemm (approved):

mfundul (commented):
You should probably mention something about compression in the commit description.

tsl/src/compression/create.c
```diff
@@ -867,6 +868,7 @@ check_modify_compression_options(Hypertable *ht, WithClauseResult *with_clause_o
 	 * empty list. */
```
mfundul, line 868: You should probably update this comment, which is not entirely accurate any more.

tsl/src/compression/create.c
```diff
@@ -878,16 +880,46 @@ check_modify_compression_options(Hypertable *ht, WithClauseResult *with_clause_o
...
+					  list_length(parsed_orderby_cols) == 1 &&
+					  (namestrcmp(&colname1, NameStr(colname2)) == 0) &&
+					  orderby_time_default_matches))
+					ereport(ERROR,
+							(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+							 errmsg("must specify a column to order by"),
+							 errdetail("The timescaledb.compress_orderby option was"
+									   " previously set and must also be specified"
+									   " in the updated configuration.")));
+			}
```
mfundul, line 921: You can deduplicate this `ereport()` call here.

E.g.
```
if (list_length(parsed_orderby_cols) == 1)
{
  ... do stuff
}

if (!(list_length(current_orderby_cols) == 1 &&
    list_length(parsed_orderby_cols) == 1 &&
    (namestrcmp(&colname1, NameStr(colname2)) == 0) &&
    orderby_time_default_matches))
        ereport(ERROR,
            errmsg("must specify a column to order by"),
            errdetail("The timescaledb.compress_orderby option was"
                      " previously set and must also be specified"
                      " in the updated configuration.")));
```

konskov: Thank you. Fixed it

tsl/src/compression/create.c
```diff
@@ -878,16 +883,42 @@ check_modify_compression_options(Hypertable *ht, WithClauseResult *with_clause_o
...
+				elem2 = list_nth_cell(parsed_orderby_cols, 0);
+				cpc_elem2 = lfirst(elem2);
+				colname2 = cpc_elem2->colname;
+				orderby_time_default_matches = (fd_elem1->orderby_asc == cpc_elem2->asc);
+			}
 
+			// this is okay only if the orderby that's already set is only the time column
+			// check for the same attribute name and same order (desc)
+			if (!(list_length(current_orderby_cols) == 1 && list_length(parsed_orderby_cols) == 1 &&
+				  (namestrcmp(&colname1, NameStr(colname2)) == 0) && orderby_time_default_matches))
```
mfundul, line 914:
```suggestion
				  (namestrcmp(NameStr(colname1), NameStr(colname2)) == 0) && orderby_time_default_matches))
```

mfundul (approved):

PR #4380, opened 2022-05-24
Add chunk exclusion for space dimensions to UPDATE/DELETE

svenklemm:
This patch adds support for chunk exclusion to UPDATE/DELETE on
space dimensions of hypertables.

### DIFF
--- a/src/planner/space_constraint.c
+++ b/src/planner/space_constraint.c
@@ -1,0 +1,314 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+
+#include <postgres.h>
+#include <access/xact.h>
+#include <datatype/timestamp.h>
+#include <nodes/makefuncs.h>
+#include <nodes/pg_list.h>
+#include <optimizer/optimizer.h>
+#include <parser/parse_func.h>
+#include <utils/fmgroids.h>
+
+#include "cache.h"
+#include "dimension.h"
+#include "hypertable.h"
+#include "hypertable_cache.h"
+#include "partitioning.h"
+#include "planner.h"
+
+/*
+ * Returns space dimension for a specific column. Returns NULL
+ * if the column is not a space dimension.
+ */
+static Dimension *
+get_space_dimension(Oid relid, Index varattno)
+{
+	Hypertable *ht = ts_planner_get_hypertable(relid, CACHE_FLAG_CHECK);
+	if (!ht)
+		return NULL;
+
+	for (uint16 i = 0; i < ht->space->num_dimensions; i++)
+	{
+		Dimension *dim = &ht->space->dimensions[i];
+		if (dim->type == DIMENSION_TYPE_CLOSED && dim->column_attno == varattno)
+		{
+			return dim;
+		}
+	}
+	return NULL;
+}
+
+/*
+ * Valid constraints are: Var = Const
+ * Var has to refer to a space partitioning column
+ */
+static bool
+is_valid_space_constraint(OpExpr *op, List *rtable)
+{
+	Assert(IsA(op, OpExpr));
+	if (!IsA(linitial(op->args), Var) || !IsA(lsecond(op->args), Const))
+		return false;
+
+	Var *var = linitial_node(Var, op->args);
+	TypeCacheEntry *tce = lookup_type_cache(var->vartype, TYPECACHE_EQ_OPR);
+
+	if (op->opno != tce->eq_opr || var->varlevelsup != 0)
+		return false;
+
+	/*
+	 * Check that the constraint is actually on a partitioning column.
+	 */
+	Assert(var->varno <= list_length(rtable));
+	RangeTblEntry *rte = list_nth(rtable, var->varno - 1);
+	Dimension *dim = get_space_dimension(rte->relid, var->varattno);
+
+	if (!dim)
+		return false;
+
+	return true;
+}
+
+/*
+ * Valid constraints are:
+ *   Var = ANY(ARRAY[Const,Const])
+ *   Var IN (Const,Const)
+ * Var has to refer to a space partitioning column
+ */
+static bool
+is_valid_scalar_space_constraint(ScalarArrayOpExpr *op, List *rtable)
+{
+	Assert(IsA(op, ScalarArrayOpExpr));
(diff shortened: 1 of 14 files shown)


svenklemm:
#4374 should be merged before this

fabriziomello:
Seems you added the generated test sql files insted of the related output files for each supported postgres version

src/planner/space_constraint.c
```diff
@@ -0,0 +1,283 @@
...
+	partcall->args = list_make1(copyObject(var));
+
+	OpExpr *ret = (OpExpr *) make_opclause(tce->eq_opr /* opno */,
+										   BOOLOID /*opresulttype */,
+										   false /* opretset */,
+										   (Expr *) partcall /* left */,
+										   (Expr *) part_value /* right */,
+										   InvalidOid /* opcollid */,
+										   InvalidOid /* inputcollid */);
+	ret->location = -29811;
```
akuzm, line 150: Should be a macro I guess

src/planner/space_constraint.c
```diff
@@ -0,0 +1,283 @@
...
+	op2->args = list_make2(partcall, arr2);
+	op2->inputcollid = InvalidOid;
+	op2->useOr = true;
+	op2->location = PLANNER_LOCATION_MAGIC;
+
+	return op2;
+}
+
+Node *
+ts_amend_space_constraints(PlannerInfo *root, List *rtable, Node *node)
```
akuzm, line 205: Let's add a top-level description of the algorithm here. If we have an equality condition on a space partitioning column, we also add a corresponding condition on `get_partition_hash` on this column. These conditions match the constraints on chunks, so Postgres' constraint exclusion is able to use them and exclude the chunks. Did I get it right?

`amend` sounds slightly imprecise, makes me think that we modify something, but here we just add a different form that matches the constraints on chunks. Maybe `ts_add_space_constraints`? Anyway, not so important.

tsl/test/expected/dist_hypertable-12.out
```diff
@@ -239,91 +239,43 @@ UPDATE disttable SET temp = 3.7 WHERE device = 1;
...
    Foreign Update on _timescaledb_internal._dist_hyper_1_4_chunk
      Remote SQL: UPDATE _timescaledb_internal._dist_hyper_1_4_chunk SET temp = $2 WHERE ctid = $1
-   Foreign Update on _timescaledb_internal._dist_hyper_1_5_chunk
-     Remote SQL: UPDATE _timescaledb_internal._dist_hyper_1_5_chunk SET temp = $2 WHERE ctid = $1
-   Foreign Update on _timescaledb_internal._dist_hyper_1_6_chunk
-     Remote SQL: UPDATE _timescaledb_internal._dist_hyper_1_6_chunk SET temp = $2 WHERE ctid = $1
    ->  Seq Scan on public.disttable
          Output: disttable."time", disttable.device, NULL::integer, '3.7'::double precision, disttable.ctid
-         Filter: (disttable.device = 1)
+         Filter: ((disttable.device = 1) AND (_timescaledb_internal.get_partition_hash(disttable.device) = 242423622))
```
akuzm, line 246: Do you want to delete them back as well as you did with constraints for `now()`?

akuzm (approved):
lgtm

src/planner/space_constraint.c
akuzm: TBH I would expect the Postgres planner to be smart enough to prove that, if you have an immutable function in a constraint and a constant condition in WHERE...

tsl/test/expected/dist_hypertable-12.out
svenklemm: yes they shouldnt end up in the explain output, will fix this

src/planner/space_constraint.c
akuzm: `operator_predicate_proof` doesn't evaluate sides of the constraint and only matches costraint & clause verbatim. And `eval_const_expression` we do on the relation constraints can't use equivalence classes to evaluate Vars to const. So no, it's not smart enough :(

gayyappan (approved):
Minor comments.
Changes look good to me.

src/planner/space_constraint.c
```diff
@@ -0,0 +1,283 @@
...
+	arr2->multidims = false;
+	arr2->location = -1;
+	arr2->elements = part_values;
+
+	ScalarArrayOpExpr *op2 = makeNode(ScalarArrayOpExpr);
+	op2->opno = tce->eq_opr;
+	op2->args = list_make2(partcall, arr2);
+	op2->inputcollid = InvalidOid;
+	op2->useOr = true;
+	op2->location = PLANNER_LOCATION_MAGIC;
```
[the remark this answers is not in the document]
svenklemm: This is so we can easily spot which nodes we added here so we can easily remove them once they are no longer needed.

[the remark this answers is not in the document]
svenklemm: relevant code is in src/planner/constraint_cleanup.c

src/planner/space_constraint.c
svenklemm: done

tsl/test/expected/dist_hypertable-12.out
svenklemm: we could make the constraint removal a guc but i'm somewhat hesitant to add a GUC like that

src/planner/space_constraint.c
```diff
@@ -0,0 +1,283 @@
...
+	{
+		Dimension *dim = &ht->space->dimensions[i];
+		if (dim->type == DIMENSION_TYPE_CLOSED && dim->column_attno == varattno)
+		{
+			return dim;
+		}
+	}
+	return NULL;
+}
+
```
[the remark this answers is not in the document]
svenklemm: done

src/planner/space_constraint.c
```diff
@@ -0,0 +1,283 @@
...
+{
+	/* build FuncExpr to use in eval_const_expressions */
+	return makeFuncExpr(partfunc.func_fmgr.fn_oid /* funcid */,
+						partfunc.rettype /* rettype */,
+						args /* args */,
+						InvalidOid /* funccollid */,
+						inputcollid /* inputcollid */,
+						COERCE_EXPLICIT_CALL /* fformat */);
+}
+
```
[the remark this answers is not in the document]
svenklemm: done

PR #4381, opened 2022-05-24
Post-Release 2.7.0

svenklemm:
Adjust version.config and add 2.7.0 to update/downgrade test.


mfundul (approved):
We should probably standardize the post release message a little bit. It's always diffierent in git history.

PR #4386, opened 2022-05-25
Fix PG13.2 nightly CI failures

svenklemm:
The 001_extension test uses the background_psql function which is not
present in the 13.2 PostgresNode module. This function is only
available in PG 13.5+.

https://github.com/postgres/postgres/commit/a9d0a540

Disable-check: commit-count


akuzm (approved):

jnidzwetzki (approved):
LGTM

PR #4388, opened 2022-05-25
Set statement_timeout for test instances

svenklemm:
This patch sets a 10 second statement timeout so any statement
taking longer than that will get canceled by postgres. For our
tests no single statement should take longer than that anyway
but quite regularly bgw tests get stuck and run into much longer
5 minute timeout imposed elsewhere.

### DIFF
--- a/test/postgresql.conf.in
+++ b/test/postgresql.conf.in
@@ -6,6 +6,7 @@ shared_preload_libraries=timescaledb
 max_worker_processes=24
 autovacuum=false
 random_page_cost=1.0
+statement_timeout=10s
 timescaledb.license='apache'
 @TELEMETRY_DEFAULT_SETTING@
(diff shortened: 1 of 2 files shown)


akuzm (approved):
I think we're going to have a lot of spurious failures because of this timeout, because the test run on some unknown infrastructure susceptible to sudden load spikes, disk hiccups and so on. Let's observe.

mkindahl (commented):
Seems strange to change the timeout to 10s. It will make tests that are already destined to fail, fail faster (but why do you want to do that?), but also make tests fail more often with flaky results. It will not speed up tests that are going to succeed anyway.

Even if you want to lower the timeout, 10s seems like a very low timeout.

fabriziomello:
> Even if you want to lower the timeout, 10s seems like a very low timeout.

Agree with @mkindahl 10s is very low timeout.

PR #4285, opened 2022-04-29
testing some stuff around distributed inserts

akuzm:
### Changes
1. Use nonblocking mode on COPY connections.
2. Group the rows into batches on access node before sending them to data nodes.

### Benchmarks

I'm using the UK Price Paid data set because that's what I had at hand: https://www.gov.uk/government/statistical-data-sets/price-paid-data-downloads

I'm benchmarking inserts of 10k, 100k and 1M rows, ordered either randomly or by `time` & `postcode2` which is used for space partitioning. The '-1' and '-2' versions of files are just different random subsets of data, to test inserting into new chunks or existing chunks. I'm tesing distributed hypertable ordered by `time`, either with no space partitioning or partitioned by `postcode2`:

```
# \d uk_price_paid
             Table "public.uk_price_paid"
  Column   │   Type   │ Collation │ Nullable │ Default 
───────────┼──────────┼───────────┼──────────┼─────────
 price     │ integer  │           │          │ 
 date      │ date     │           │ not null │ 
 postcode1 │ text     │           │          │ 
 postcode2 │ text     │           │          │ 
 type      │ smallint │           │          │ 
 is_new    │ boolean  │           │          │ 
 duration  │ smallint │           │          │ 
 addr1     │ text     │           │          │ 
 addr2     │ text     │           │          │ 
 street    │ text     │           │          │ 
 locality  │ text     │           │          │ 
 town      │ text     │           │          │ 
 district  │ text     │           │          │ 
 country   │ text     │           │          │ 
 category  │ smallint │           │          │ 
Indexes:
    "uk_price_paid_date_idx" btree (date DESC)
    "uk_price_paid_postcode2_date_idx" btree (postcode2, date DESC)
Triggers:
    ts_insert_blocker BEFORE INSERT ON uk_price_paid FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker()

# select create_hypertable('uk_price_paid_l', 'date');
# select create_distributed_hypertable('uk_price_paid_ns', 'date');
# select create_distributed_hypertable('uk_price_paid', 'date', 'postcode2');
```

I'm using the following, pretty generic, queries:
```
==> insert-binary.sql <==
begin;
\copy rides1 from '/home/akuzm/nyc_data_rides.pgbinary' with (format binary);
rollback;

==> insert-prices-100k-ordered.sql <==
begin;
copy uk_price_paid from '/home/akuzm/uk-price-paid/prices-100k-ordered-1.tsv';
rollback;

[...]

PR #4390, opened 2022-05-26
Faster lookup of chunks by point

akuzm:
Don't keep the chunk constraints while searching. The number of
candidate chunks can be very large, so keeping these constraints is a
lot of work and uses a lot of memory. For finding the matching chunk,
it is enough to track the number of dimensions that matched a given
chunk id. After finding the chunk id, we can look up only the matching
chunk data with the usual function.

This saves some work when doing INSERTs.

Gives 10%-30% speedup on my COPY tests that touch a lot of chunks.

Part of #4285

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -126,13 +126,14 @@ static int chunk_scan_ctx_foreach_chunk_stub(ChunkScanCtx *ctx, on_chunk_stub_fu
 											 uint16 limit);
 static Datum chunks_return_srf(FunctionCallInfo fcinfo);
 static int chunk_cmp(const void *ch1, const void *ch2);
-static Chunk *chunk_find(const Hypertable *ht, const Point *p, bool resurrect, bool lock_slices);
+static int chunk_point_find_chunk_id(const Hypertable *ht, const Point *p);
 static void init_scan_by_qualified_table_name(ScanIterator *iterator, const char *schema_name,
 											  const char *table_name);
 static Hypertable *find_hypertable_from_table_or_cagg(Cache *hcache, Oid relid, bool allow_matht);
 static Chunk *get_chunks_in_time_range(Hypertable *ht, int64 older_than, int64 newer_than,
 									   const char *caller_name, MemoryContext mctx,
 									   uint64 *num_chunks_returned, ScanTupLock *tuplock);
+static Chunk *chunk_resurrect(const Hypertable *ht, int chunk_id);
 
 /*
  * The chunk status field values are persisted in the database and must never be changed.
@@ -1435,15 +1436,31 @@ ts_chunk_find_or_create_without_cuts(const Hypertable *ht, Hypercube *hc, const 
 }
 
 /*
+ * Find the chunk containing the given point, locking all its dimension slices
+ * for share. NULL if not found.
+ */
+Chunk *
+ts_chunk_find_for_point(const Hypertable *ht, const Point *p)
+{
+	int chunk_id = chunk_point_find_chunk_id(ht, p);
+	if (chunk_id == 0)
+	{
+		return NULL;
+	}
+
+	/* The chunk might be dropped, so we don't fail if we haven't found it. */
+	return ts_chunk_get_by_id(chunk_id, /* fail_if_not_found = */ false);
+}
+
+/*
  * Create a chunk through insertion of a tuple at a given point.
  */
 Chunk *
-ts_chunk_create_from_point(const Hypertable *ht, const Point *p, const char *schema,
-						   const char *prefix)
+ts_chunk_create_for_point(const Hypertable *ht, const Point *p, const char *schema,
+						  const char *prefix)
 {
-	Chunk *chunk;
-
 	/*
+	 * We're going to have to resurrect or create the chunk.
 	 * Serialize chunk creation around a lock on the "main table" to avoid
 	 * multiple processes trying to create the same chunk. We use a
 	 * ShareUpdateExclusiveLock, which is the weakest lock possible that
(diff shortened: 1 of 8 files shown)


src/chunk_constraint.c
```diff
@@ -444,6 +444,63 @@ typedef struct ChunkConstraintScanData
 	DimensionSlice *slice;
 } ChunkConstraintScanData;
 
+/*
+ * Scan for all chunk constraints that match the given slice ID. The chunk
+ * constraints are not saved, only the matching chunk id is returned.
```
erimatnor, line 449: Suggest expanding on motivation why only chunk ID is returned here.

src/chunk_constraint.c
```diff
@@ -444,6 +444,63 @@ typedef struct ChunkConstraintScanData
...
+
+	ScanIterator iterator = ts_scan_iterator_create(CHUNK_CONSTRAINT, AccessShareLock, mctx);
+	ts_chunk_constraint_scan_iterator_set_slice_id(&iterator, slice->fd.id);
+
+	ts_scanner_foreach(&iterator)
+	{
+		ChunkScanEntry *entry;
+		bool found;
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		Datum datum = slot_getattr(ti->slot, Anum_chunk_constraint_chunk_id, &found);
```
erimatnor, line 465: Misnamed boolean `found`. The `slot_getattr` boolean indicates nullness of the returned value, not whether it is "found". I suggest creating a new boolean named `isnull` for use here.

```suggestion
		Datum datum = slot_getattr(ti->slot, Anum_chunk_constraint_chunk_id, &isnull);
		Assert(!isnull);
```

src/chunk_constraint.c
```diff
@@ -444,6 +444,63 @@ typedef struct ChunkConstraintScanData
...
+		int32 chunk_id = DatumGetInt32(datum);
+
+		if (slot_attisnull(ts_scan_iterator_slot(&iterator),
+						   Anum_chunk_constraint_dimension_slice_id))
+			continue;
```
erimatnor, line 470: I suggest moving this check about `slot_getattr` above, since there is no need to get the `chunk_id` if this chunk isn't true.

src/chunk_constraint.c
```diff
@@ -444,6 +444,63 @@ typedef struct ChunkConstraintScanData
...
+		{
+			entry->stub = NULL;
+			entry->num_dimension_constraints = 0;
+		}
+
+		/*
+		 * We have only the dimension constraints here, because we're searching
+		 * by dimension slice id.
+		 */
+		Assert(!slot_attisnull(ti->slot, Anum_chunk_constraint_dimension_slice_id));
```
erimatnor, line 484: Can remove this assert since there is already a null check for this dimension slice ID above.

src/chunk.c
```diff
@@ -1980,57 +1956,42 @@ chunk_resurrect(const Hypertable *ht, const ChunkStub *stub)
...
-		 * the chunk is marked as dropped. */
-		chunk = chunk_create_from_stub(&stubctx);
+	/* Find constraints matching dimension slices. */
+	ListCell *lc;
+	foreach (lc, all_slices)
+	{
+		chunk_id =
+			ts_chunk_constraint_scan_by_dimension_slice_chunk_id((DimensionSlice *) lfirst(lc),
+																 &ctx,
+																 CurrentMemoryContext);
```
erimatnor, line 1984: Note that this will open and close the chunk_constraint table for every iteration in the loop. Previously, we found that this was often costly, so it is better to open the table outside the foreach loop and then only restarting the scan within the loop.

src/chunk_constraint.c
```diff
@@ -444,6 +444,63 @@ typedef struct ChunkConstraintScanData
...
+ * Scan for all chunk constraints that match the given slice ID. The chunk
+ * constraints are not saved, only the matching chunk id is returned.
+ */
+int
+ts_chunk_constraint_scan_by_dimension_slice_chunk_id(const DimensionSlice *slice, ChunkScanCtx *ctx,
+													 MemoryContext mctx)
+{
+	int result = 0;
+
+	ScanIterator iterator = ts_scan_iterator_create(CHUNK_CONSTRAINT, AccessShareLock, mctx);
```
erimatnor, line 457: Consider moving this outside this function and, e.g., passing in the iterator to avoid repeated open/close.

erimatnor (commented):
Nice improvement if the numbers are accurate. Added a few comments.

src/chunk_constraint.c
akuzm: I removed this function entirely (embedded into caller) and applied changes based on your comments below.

src/chunk_constraint.c
akuzm: Done.

src/chunk_constraint.c
akuzm: I removed the null check instead.

src/chunk_constraint.c
akuzm: Removed this check.

src/chunk_constraint.c
akuzm: Renamed.

erimatnor (approved):

src/nodes/chunk_dispatch.c
```diff
@@ -131,7 +131,11 @@ ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
 	{
 		Chunk *new_chunk;
 
-		new_chunk = ts_hypertable_get_or_create_chunk(dispatch->hypertable, point);
+		new_chunk = ts_hypertable_find_chunk_for_point(dispatch->hypertable, point);
+		if (new_chunk == NULL)
+		{
+			new_chunk = ts_hypertable_create_chunk_for_point(dispatch->hypertable, point);
+		}
```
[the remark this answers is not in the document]
akuzm: Unfortunately the second check is required, it's done after locking, and before that the chunk can be created by a concurrent transaction.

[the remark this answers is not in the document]
mkindahl: I was actually referring to the first check: is that really needed? There is already a check inside `ts_hypertable_create_chunk_for_point` that will return the chunk if it already exists.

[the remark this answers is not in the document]
akuzm: The check in `ts_hypertable_create_chunk_for_point` is performed under a `ShareUpdateExclusiveLock` on the hypertable, to serialize concurrent creation of the chunks. We don't want to take this lock unless we have to create the chunk, that's why we first search for the chunk w/o the lock in `ts_hypertable_find_chunk_for_point`. So the first check serves as a "fast path" for the "usual" case when the chunk already exists. This fast path is normally taken for the overwhelming majority of the rows inserted in the chunk -- that is, all except the first one.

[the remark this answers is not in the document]
akuzm: I added a comment about this.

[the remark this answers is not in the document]
mkindahl: So this means that to use the API correctly it is necessary to first check for the existence of a chunk with `ts_hypertable_find_chunk_for_point` before calling `ts_hypertable_create_chunk_for_point`? Looks like a very cumbersome interface to use and it sounds like this check should be incorporated into the `ts_hypertable_create_chunk_for_point` instead to avoid making mistakes when executing a fetch-or-create.

It would also allow you to avoid the second check since you can upgrade the lock instead of releasing the read lock and re-take a shared update exclusive lock.

However, I note that `get_target_chunk` in `dist_copy.c` actually want to take special action when the chunk does not exist, which would also have to be incorporated into the function instead of putting this requirement on the caller.

However, this is out of scope for the PR, so I leave it to you to decide if you want to do this refactoring.

mkindahl (approved):

src/nodes/chunk_dispatch.c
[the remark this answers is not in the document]
mkindahl: Also note that the "fast path" when the chunk already exists is not affected by the `SHARE UPDATE EXCLUSIVE` lock: it will only block other threads that try to create the chunk at the same time, not reads of the table.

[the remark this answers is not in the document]
akuzm: Yes, the reason I wanted to keep these functions separate is that some callers need to make a decision based on whether it is a newly created chunk or the old one. Good point about upgrading the lock, I didn't think about that. I think that with #4285, the requirements to this API might change, so I'll revisit it.

PR #4397, opened 2022-05-28
Support intervals with month component when constifying now()

svenklemm:
When dealing with Intervals with month component timezone changes
can result in multiple day differences in the outcome of these
calculations due to different month lengths. When dealing with
months we add a 7 day safety buffer.
For all these calculations it is fine if we exclude less chunks
than strictly required for the operation, additional exclusion
with exact values will happen in the executor. But under no
circumstances must we exclude too much cause there would be
no way for the executor to get those chunks back.

### DIFF
--- a/src/planner/constify_now.c
+++ b/src/planner/constify_now.c
@@ -33,8 +33,7 @@
  * - Var >= now() - Interval
  * - Var >= now() + Interval
  *
- * Additionally Interval needs to be Const and not contain a month
- * component.
+ * Interval needs to be Const in those expressions.
  */
 static const Dimension *
 get_hypertable_dimension(Oid relid)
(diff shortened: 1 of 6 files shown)


src/planner/constify_now.c
```diff
@@ -163,13 +154,25 @@ constify_now_expr(PlannerInfo *root, OpExpr *op)
 		 * will be repeated with exact values during execution.
 		 * Since dst switches seem to range between -1 and 2 hours we set
 		 * the safety buffer to 4 hours.
+		 * When dealing with Intervals with month component timezone changes
+		 * can result in multiple day differences in the outcome of these
+		 * calculations due to different month lengths. When dealing with
+		 * months we add a 7 day safety buffer.
```
akuzm, line 160: Is there any rationale for the particular number 7?

akuzm (approved):

src/planner/constify_now.c
svenklemm: so the worst diff i can think of will be when february is involved leading up to 3 day difference due to month length. So I doubled that and rounded up to full week.

fabriziomello (approved):

jfjoly:
@svenklemm I'm moving this change to 2.8.0 since this is a new feature.

Issue #1094, opened 2019-03-06
Timescale hypertables don't support ON CONFLICT ON CONSTRAINT upsert statement (breaking Hasura compatibility)


waterdrop01:
**Relevant system information:**
 - OS: Ubuntu 18.04
 - PostgreSQL version: PostgreSQL 11.1
 - TimescaleDB version: 1.2.1
 - Installation method: docker `timescale/timescaledb:1.2.1-pg11`

**Describe the bug**
Here is my schema:
```sql
CREATE TABLE data (
    "timestamp" timestamp with time zone NOT NULL,
    "uuid" uuid NOT NULL
);

ALTER TABLE ONLY data
    ADD CONSTRAINT data_pkey PRIMARY KEY ("uuid", "timestamp");

SELECT create_hypertable('data', 'timestamp');
```

**To Reproduce**
Run this SQL
```sql
INSERT INTO "data" ( "timestamp", "uuid" ) VALUES ( "2019-03-06T16:46:46.988Z", "9db6dbe1-cd14-1d17-38c0-4d4a36679fd6" ) ON CONFLICT ON CONSTRAINT "data_pkey" DO NOTHING RETURNING *
```

**Expected behavior**
Expected to correctly insert/upsert data without error, as it is doing with **PostgreSQL 10.6** and timescale 1.2.1 (docker `timescale/timescaledb:1.2.1-pg10`).

**Actual behavior**
Postgres throws the following error: 
```
ERROR:  hypertables do not support ON CONFLICT statements that reference constraints
HINT:  Use column names to infer indexes instead.
```

Note: mention of my error in this file: https://github.com/timescale/timescaledb/blob/master/test/expected/upsert.out

waterdrop01:
Note: Just had a feedback from the folks over at Hasura: https://github.com/hasura/graphql-engine/issues/1711#issuecomment-470429671

waterdrop01:
note: It may not actually come from PG version, but only from Timescaledb version.

waterdrop01:
Not being able to reference a constraint by its name is problematic. Is it long term design, or short term workaround?

having a feedback from Timescaledb team on this issue would be very valuable,

Thank you! :)

waterdrop01:
Hi Timescaledb team,

I understand you may have other priorities, but any chance to get some feedback on this?

Thank you!

j0h:
[...]

PR #4405, opened 2022-06-01
Try to enable ON CONFLICT ON CONSTRAINT

akuzm:
Looks like it might work since we started to rely on Postgres' arbiter
index inference.

Fixes https://github.com/timescale/timescaledb/issues/1094


svenklemm:
I think this needs more tests.

akuzm:
I modified the existing `upsert` test to use constraints instead of indexes in some cases. This way we can test both ways of arbiter index inference on the same table. So now we have also test ON CONSTRAINT for multi-column constraints and several constraints on one table. We also check that all variants of index inference don't work on distributed hypertables. Let me know if you have other cases in mind.

svenklemm (approved):

fabriziomello (approved):

PR #4411, opened 2022-06-02
Add shmem_request_hook

svenklemm:
This patch consolidates all shared memory requests in a
shmem_request_hook. While there are no strict requirements when
to request shared memory for PG < 15 in PG 15 it has to happen
in the shmem_request_hook otherwise the request will be blocked.

https://github.com/postgres/postgres/commit/4f2400cb

### DIFF
--- a/src/loader/loader.c
+++ b/src/loader/loader.c
@@ -114,6 +114,9 @@ int ts_guc_bgw_launcher_poll_time = BGW_LAUNCHER_POLL_TIME_MS;
 /* This is the hook that existed before the loader was installed */
 static post_parse_analyze_hook_type prev_post_parse_analyze_hook;
 static shmem_startup_hook_type prev_shmem_startup_hook;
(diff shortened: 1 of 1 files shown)


src/loader/loader.c
```diff
@@ -672,9 +694,11 @@ _PG_init(void)
 	prev_post_parse_analyze_hook = post_parse_analyze_hook;
 	/* register shmem startup hook for the background worker stuff */
 	prev_shmem_startup_hook = shmem_startup_hook;
+	prev_shmem_request_hook = shmem_request_hook;
```
jnidzwetzki, line 697: Shouldn't these hooks only be registered with PG >= 15 (other versions seems to use the function call in Line 654)?

svenklemm: fixed

fabriziomello (approved):

PR #4413, opened 2022-06-03
Don't redefine PGDLLEXPORT

afiskon:
This code was introduced in 027b7b29 but unfortunatelly the commit message
doesn't clarify the need to redefine PGDLLEXORT. It should be defined correctly
by Postgres and we shouldn't touch it.

This was not a problem until recently. This code is not going to work with
the upcoming support of Meson build system in the upstream, and as long as
this code is present the extension can't be tested for compatibility with
the upstream.

Per discussion:
https://www.postgresql.org/message-id/20220602152257.szbstfitf3lwdssp%40alap3.anarazel.de

### DIFF
--- a/src/export.h
+++ b/src/export.h
@@ -10,32 +10,6 @@
 
 #include "config.h"
 
-/* Definitions for symbol exports */
-
-#define TS_CAT(x, y) x##y
-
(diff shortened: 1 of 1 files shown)


afiskon:
Nope, it didn't go well. Apparently this should be addressed differently.

PR #4417, opened 2022-06-03
Add workflow to check for CHANGELOG update

konskov:


fabriziomello:
Just some comments here:

1. Isn't easier just checkout the PR and check if there are some CHANGELOG.md changes running the following git command:
```sh
$ git diff --name-status main | grep CHANGELOG.md
```
2. IMHO we should add a tag to explicit disable this check like we do for commit count, something like `Disable-check: force-changelog-changed`

akuzm:
Do we expect to start updating CHANGELOG for most PRs? Each of them is going to have conflicts then, right? Maybe switching CHANGELOG.md to union merge can alleviate this...
It can be enabled by adding something like this to `.gitattributes`:
```
# to avoid frequent conflicts
CHANGELOG.md text merge=union
```

svenklemm (approved):

fabriziomello (approved):

PR #4430, opened 2022-06-08
Create indexes on Continuous Aggregates

fabriziomello:
that allows users efectivelly create and use indexes in the
materialization hypertable. The boring part of it is that users should
discover what is the associated materialization hypertable to issue a
`CREATE INDEX` statement.
    
Improved it by allowing users to easily create indexes in the
materialization hypertable by simple executing a `CREATE INDEX` direct
in the Continuous Aggregate.
    
Example:
  `CREATE INDEX name_of_the_index ON continuous_agregate (column);`

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -2568,8 +2568,38 @@ process_index_start(ProcessUtilityArgs *args)
 
 	if (NULL == ht)
 	{
-		ts_cache_release(hcache);
-		return DDL_CONTINUE;
+		/* Check if the relation is a Continuous Aggregate */
+		ContinuousAgg *cagg = ts_continuous_agg_find_by_rv(stmt->relation);
+
+		if (cagg)
+		{
+			/* If the relation is a CAgg and it is finalized */
+			if (ContinuousAggIsFinalized(cagg))
+				ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
+			else
+			{
+				ts_cache_release(hcache);
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("operation not supported on continuous aggreates that are not "
+								"finalized"),
+						 errhint("Recreate the continuous aggregate to allow index creation.")));
+			}
+		}
+
+		if (NULL == ht)
+		{
+			ts_cache_release(hcache);
+			return DDL_CONTINUE;
+		}
+
+		if (stmt->unique)
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("continuous aggregates do not support UNIQUE indexes")));
+
+		/* Make the RangeVar for the underlying materialization hypertable */
+		stmt->relation = makeRangeVar(NameStr(ht->fd.schema_name), NameStr(ht->fd.table_name), -1);
 	}
 	else if (TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht))
 	{
--- a/test/sql/utils/testsupport.sql
+++ b/test/sql/utils/testsupport.sql
@@ -82,6 +82,29 @@ $BODY$
     ORDER BY c.relname;
 $BODY$;
 
+CREATE OR REPLACE FUNCTION test.show_indexespred(rel regclass)
+RETURNS TABLE("Index" regclass,
+              "Columns" name[],
+              "Expr" text,
+              "Pred" text,
+              "Unique" boolean,
+              "Primary" boolean,
+              "Exclusion" boolean,
+              "Tablespace" name) LANGUAGE SQL STABLE AS
+$BODY$
+    SELECT c.oid::regclass,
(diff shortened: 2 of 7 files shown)


svenklemm (commented):
Creating unique indexes should probably blocked as you can mess up your cagg that way, also needs tests with unique indexes

mkindahl (commented):
I think you're missing tests for the following cases:
- Using `ONLY`. It should default to the same situation as for hypertables, but better check.
- Using `UNIQUE`. Adding `UNIQUE` can cause the refresh job to start failing, so not sure how useful it is, but not convinced that is a good reason to now allow it.
- Using `CONCURRENTLY`. We don't support it on hypertables right now, but better check that it gives an error.
- Test for partial indexes.
- Using `INCLUDE`. This can be quite useful, so good to add a check that it actually works so that we do not make a mistake causing it to not work in the future.

src/process_utility.c
```diff
@@ -2568,8 +2568,34 @@ process_index_start(ProcessUtilityArgs *args)
...
+		if (NULL == ht)
+		{
+			ts_cache_release(hcache);
+			return DDL_CONTINUE;
+		}
+
+		/* Make the RangeVar for the underlying materialization hypertable */
+		stmt->relation = makeRangeVar((char *) NameStr(ht->fd.schema_name),
+									  (char *) NameStr(ht->fd.table_name),
+									  0);
```
mfundul, line 2598:
```suggestion
									  -1);
```
Shouldn't token location be unknown?

src/process_utility.c
```diff
@@ -2568,8 +2568,34 @@ process_index_start(ProcessUtilityArgs *args)
...
+
+		if (NULL == ht)
+		{
+			ts_cache_release(hcache);
+			return DDL_CONTINUE;
+		}
+
+		/* Make the RangeVar for the underlying materialization hypertable */
+		stmt->relation = makeRangeVar((char *) NameStr(ht->fd.schema_name),
+									  (char *) NameStr(ht->fd.table_name),
```
mfundul, line 2597:
```suggestion
		stmt->relation = makeRangeVar(NameStr(ht->fd.schema_name),
									  NameStr(ht->fd.table_name),
```

src/process_utility.c
```diff
@@ -2568,8 +2568,34 @@ process_index_start(ProcessUtilityArgs *args)
...
+			}
+		}
+
+		if (NULL == ht)
+		{
+			ts_cache_release(hcache);
+			return DDL_CONTINUE;
+		}
+
+		/* Make the RangeVar for the underlying materialization hypertable */
```
mfundul, line 2595: In general, I think this `stmt->relation = makeRangeVar` should only happen for the new case of CAGGs, and not for all hypertables. It could be harmless, but still, it looks like an unnecessary change for the common codepath to me.

src/process_utility.c
```diff
@@ -2568,8 +2568,34 @@ process_index_start(ProcessUtilityArgs *args)
...
+			/* If the relation is a CAgg and it is finalized */
+			if (ContinuousAggIsFinalized(cagg))
+				ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
+			else
+			{
+				ts_cache_release(hcache);
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("operation not supported on continuous aggreates that are not "
+								"finalized")));
```
mfundul, line 2585: Maybe add a hint that the user needs to recreate the CAGG, since they may not understand what finalized is.

fabriziomello:
> Creating unique indexes should probably blocked as you can mess up your cagg that way, also needs tests with unique indexes

Fixed!

fabriziomello:
> I think you're missing tests for the following cases:
> 
> * Using `ONLY`. It should default to the same situation as for hypertables, but better check.
> * Using `UNIQUE`. Adding `UNIQUE` can cause the refresh job to start failing, so not sure how useful it is, but not convinced that is a good reason to now allow it.
> * Using `CONCURRENTLY`. We don't support it on hypertables right now, but better check that it gives an error.
> * Test for partial indexes.
> * Using `INCLUDE`. This can be quite useful, so good to add a check that it actually works so that we do not make a mistake causing it to not work in the future.

All fixed!

src/process_utility.c
fabriziomello: Fixed!

src/process_utility.c
fabriziomello: Fixed!

src/process_utility.c
fabriziomello: Added a hint!

src/process_utility.c
fabriziomello: But it will happen only for the new CAGG. I've added a test before and raise an error if the CAGG is not finalized!

mfundul: Yes, I've misread the code, my bad.

mkindahl (approved):

svenklemm (approved):

PR #4434, opened 2022-06-10
Add more tests for errors on data nodes

akuzm:
Use a data type with faulty send/recv functions to test various error
handling paths.

### DIFF
--- a/test/src/test_utils.c
+++ b/test/src/test_utils.c
@@ -97,7 +97,7 @@ throw_after_n_rows(int max_rows, int severity)
 	if (max_rows <= rows_seen)
 	{
 		ereport(severity,
-				(errmsg("debug point: requested to error out every %d rows, %d rows seen",
+				(errmsg("debug point: requested to error out after %d rows, %d rows seen",
 						max_rows,
 						rows_seen)));
 	}
@@ -115,6 +115,33 @@ Datum
 ts_debug_shippable_fatal_after_n_rows(PG_FUNCTION_ARGS)
 {
 	PG_RETURN_INT32(throw_after_n_rows(PG_GETARG_INT32(0), FATAL));
+}
+
+/*
+ * Broken send/receive functions for int4 that throw after an (arbitrarily
+ * chosen prime) number of rows.
+ * Use ERROR, not FATAL, because PG versions < 14 are unable to report a FATAL
+ * error to the access node before closing the connection, so the test results
+ * would be different.
+ */
+#define ARBITRARY_PRIME_NUMBER 7103
+
+TS_FUNCTION_INFO_V1(ts_debug_broken_int4recv);
+
+Datum
+ts_debug_broken_int4recv(PG_FUNCTION_ARGS)
+{
+	(void) throw_after_n_rows(ARBITRARY_PRIME_NUMBER, ERROR);
+	return int4recv(fcinfo);
+}
+
+TS_FUNCTION_INFO_V1(ts_debug_broken_int4send);
+
+Datum
+ts_debug_broken_int4send(PG_FUNCTION_ARGS)
+{
+	(void) throw_after_n_rows(ARBITRARY_PRIME_NUMBER, ERROR);
+	return int4send(fcinfo);
 }
 
 TS_FUNCTION_INFO_V1(ts_bgw_wait);
--- a/tsl/src/remote/async.c
+++ b/tsl/src/remote/async.c
@@ -723,27 +723,49 @@ wait_to_consume_data(AsyncRequestSet *set, TimestampTz end_time)
 
 		CHECK_FOR_INTERRUPTS();
 
+		if (event.events & ~(WL_SOCKET_READABLE | WL_LATCH_SET))
+		{
+			/*
+			 * Sanity check on the wait result: we haven't requested anything
+			 * other than my latch or the socket becoming readable.
+			 */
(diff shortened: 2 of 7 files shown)


test/src/test_utils.c
```diff
@@ -117,6 +117,68 @@ ts_debug_shippable_fatal_after_n_rows(PG_FUNCTION_ARGS)
...
+}
+
+TS_FUNCTION_INFO_V1(ts_debug_int4out);
+
+Datum
+ts_debug_int4out(PG_FUNCTION_ARGS)
+{
+	return int4out(fcinfo);
+}
+
```
svenklemm, line 181:
```suggestion

```

You don't need this you can use postgres internal functions like so:

```
CREATE or REPLACE function foo() returns bytea language internal as 'int4send';
```

tsl/src/remote/connection.c
```diff
@@ -1222,6 +1222,16 @@ set_ssl_options(const char *user_name, const char **keywords, const char **value
...
+	ssl_enabled = GetConfigOption("timescaledb.debug_enable_ssl", true, false);
+	if (ssl_enabled && strcmp(ssl_enabled, "on") != 0)
+	{
+		keywords[option_pos] = "sslmode";
+		values[option_pos] = "disable";
+		option_pos++;
+		*option_start = option_pos;
+		return;
+	}
+
```
svenklemm, line 1234: Hmm why do you need this? Can we only do this in debug builds?

tsl/src/remote/async.c
```diff
@@ -724,25 +724,44 @@ wait_to_consume_data(AsyncRequestSet *set, TimestampTz end_time)
...
 		}
-		else
+
+		if (event.events & ~(WL_SOCKET_READABLE | WL_LATCH_SET))
 		{
-			result = async_response_error_create("unexpected event");
+			/* Not my latch and not readable socket. */
+			result = async_response_error_create(
+				psprintf("Unexpected event 0x%X while waiting for async request result",
+						 event.events));
```
mfundul, line 764: Should this check be before `PQisBusy()`?

tsl/src/remote/connection.c
akuzm: To disable SSL connections to data nodes, because using SSL leads to some version-specifica changes in error messages. I'll make it debug-only.

tsl/src/remote/async.c
akuzm: This is a sanity check on result returned by `WaitEventSetWait`, we shouldn't see any flags besides these two. I'll move it to the very beginning.

test/src/test_utils.c
```diff
@@ -117,6 +117,31 @@ ts_debug_shippable_fatal_after_n_rows(PG_FUNCTION_ARGS)
...
+ * error to the access node before closing the connection, so the test results
+ * would be different.
+ */
+
+TS_FUNCTION_INFO_V1(ts_debug_broken_int4recv);
+
+Datum
+ts_debug_broken_int4recv(PG_FUNCTION_ARGS)
+{
+	(void) throw_after_n_rows(7103 /* an arbitrary prime */, ERROR);
```
mfundul, line 132:
```suggestion
	(void) throw_after_n_rows(ARBITRARY_PRIME_NUMBER, ERROR);
```

test/src/test_utils.c
```diff
@@ -117,6 +117,31 @@ ts_debug_shippable_fatal_after_n_rows(PG_FUNCTION_ARGS)
...
+	return int4recv(fcinfo);
+}
+
+TS_FUNCTION_INFO_V1(ts_debug_broken_int4send);
+
+Datum
+ts_debug_broken_int4send(PG_FUNCTION_ARGS)
+{
+	(void) throw_after_n_rows(7103 /* an arbitrary prime */, ERROR);
```
mfundul, line 141:
```suggestion
	(void) throw_after_n_rows(ARBITRARY_PRIME_NUMBER, ERROR);
```

tsl/test/shared/sql/dist_remote_error.sql
```diff
@@ -35,43 +44,90 @@ set client_min_messages to ERROR;
...
-select 1 from metrics_dist_remote_error where ts_debug_shippable_error_after_n_rows(10000000, device_id)::int != 0;
+select 1 from metrics_dist_remote_error where ts_debug_shippable_error_after_n_rows(701, device_id)::int != 0;
 
 explain (analyze, verbose, costs off, timing off, summary off)
-select 1 from metrics_dist_remote_error where ts_debug_shippable_fatal_after_n_rows(1, device_id)::int != 0;
+select 1 from metrics_dist_remote_error where ts_debug_shippable_error_after_n_rows(10000, device_id)::int != 0;
 
 explain (analyze, verbose, costs off, timing off, summary off)
-select 1 from metrics_dist_remote_error where ts_debug_shippable_fatal_after_n_rows(10000, device_id)::int != 0;
+select 1 from metrics_dist_remote_error where ts_debug_shippable_error_after_n_rows(16348, device_id)::int != 0;
```
mfundul, line 63:
```suggestion
select 1 from metrics_dist_remote_error where ts_debug_shippable_error_after_n_rows(16384, device_id)::int != 0;
```
Consider my OCD triggered!

svenklemm (approved):

PR #4437, opened 2022-06-12
Fix segfault in subscription_exec

svenklemm:
Add a check for NULL input to subscription_exec as the function
currently segfaults on NULL input. Found by sqlsmith.

### DIFF
--- a/tsl/src/reorder.c
+++ b/tsl/src/reorder.c
@@ -276,6 +276,9 @@ tsl_subscription_exec(PG_FUNCTION_ARGS)
 	List *parsetree_list;
 	ListCell *parsetree_item;
 
+	if (!subscription_cmd)
+		PG_RETURN_VOID();
+
 	/*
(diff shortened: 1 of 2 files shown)


fabriziomello (approved):

tsl/src/reorder.c
```diff
@@ -276,6 +276,9 @@ tsl_subscription_exec(PG_FUNCTION_ARGS)
 	List *parsetree_list;
 	ListCell *parsetree_item;
 
+	if (!subscription_cmd)
+		PG_RETURN_VOID();
```
pmwkaa, line 280: I suggest to error out in such case. I believe this function is for internal use only (used by copy/move chunk), ideally it should not accept NULL values

svenklemm: I disagree we don't error on empty string so not erroring on NULL would be consistent with that.

gayyappan (approved):

PR #4442, opened 2022-06-14
Fix perms in copy/move chunk

nikkhils:
We mandate that a superuser or a user with REPLICATION privileges can
invoke copy_chunk or move_chunk procedures. Internally, many stages
are carried out to complete the activity and different stages need
different user permissions. To keep things uniform we now switch to the
bootstrap superuser in each stage. Care is taken to ensure that the
original hypertable ownership is retained on the new chunk post the
move operation.

### DIFF
--- a/tsl/src/chunk_api.c
+++ b/tsl/src/chunk_api.c
@@ -1665,6 +1665,8 @@ chunk_create_empty_table(PG_FUNCTION_ARGS)
 	Cache *const hcache = ts_hypertable_cache_pin();
 	Hypertable *ht;
 	Hypercube *hc;
+	Oid uid, saved_uid;
+	int sec_ctx;
 
 	GETARG_NOTNULL_OID(hypertable_relid, 0, "hypertable");
 	GETARG_NOTNULL_NULLABLE(slices, 1, "slices", JSONB_P);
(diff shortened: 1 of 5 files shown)


tsl/src/chunk_copy.c
```diff
@@ -1057,6 +1080,8 @@ chunk_copy_execute(ChunkCopy *cc)
 
 		DEBUG_ERROR_INJECTION(stage->name);
 
+		if (BOOTSTRAP_SUPERUSERID != saved_uid)
+			SetUserIdAndSecContext(saved_uid, sec_ctx);
```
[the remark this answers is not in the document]
nikkhils: Yes, it gets correctly restored. The aborting of the transaction clears off the uid.

tsl/src/chunk_copy.c
```diff
@@ -300,18 +300,20 @@ chunk_copy_setup(ChunkCopy *cc, Oid chunk_relid, const char *src_node, const cha
...
+	 * need to check if the user has "ownership" on the hypertable now.
+	 */
 	if (!hypertable_is_distributed(ht))
 		ereport(ERROR,
 				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 				 errmsg("hypertable \"%s\" is not distributed",
 						get_rel_name(ht->main_table_relid))));
 
-	cc->src_server = data_node_get_foreign_server(src_node, ACL_USAGE, true, false);
+	cc->src_server = data_node_get_foreign_server(src_node, ACL_NO_CHECK, false, false);
```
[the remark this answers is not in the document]
nikkhils: The activity will be done as superuser whereas the caller with REPLICATION perms might not have foreign server privileges. So, we only want the src_server structure to be filled up here

erimatnor (approved):

tsl/src/chunk_api.c
```diff
@@ -1673,13 +1675,37 @@ chunk_create_empty_table(PG_FUNCTION_ARGS)
...
+	 */
+	if (strcmp(schema_name, INTERNAL_SCHEMA_NAME) == 0)
+		uid = ts_catalog_database_info_get()->owner_uid;
+	else
+	{
+		Relation rel;
+
+		rel = table_open(ht->main_table_relid, AccessShareLock);
+		uid = rel->rd_rel->relowner;
+		table_close(rel, AccessShareLock);
```
erimatnor, line 1691: Looks like the new code isn't hit in a test.  I guess the new chunks need to be created in non-internal schema for this code path to be evaluated. I think this can be done by setting `associated_schema_name` when creating the hypertable.

tsl/src/chunk_copy.c
```diff
@@ -1206,6 +1248,8 @@ chunk_copy_cleanup_internal(ChunkCopy *cc, int stage_idx)
 		else
 			first = false;
 
+		if (BOOTSTRAP_SUPERUSERID != saved_uid)
+			SetUserIdAndSecContext(saved_uid, sec_ctx);
```
erimatnor, line 1252: Again, this new code path is never tested.

tsl/src/chunk_copy.c
```diff
@@ -1194,7 +1219,24 @@ chunk_copy_cleanup_internal(ChunkCopy *cc, int stage_idx)
...
+		 * Replication slots management needs superuser or REPLICATION privs
+		 *
+		 * The cleanup function can only be invoked by superuser
+		 * or REPLICATION users. To keep things manageable, we switch to the
+		 * bootstrap superuser for each stage of the execution.
+		 */
 		SPI_start_transaction();
+		GetUserIdAndSecContext(&saved_uid, &sec_ctx);
+		if (BOOTSTRAP_SUPERUSERID != saved_uid)
+			SetUserIdAndSecContext(BOOTSTRAP_SUPERUSERID, sec_ctx | SECURITY_LOCAL_USERID_CHANGE);
```
erimatnor, line 1239: Also not tested.

tsl/src/chunk_copy.c
nikkhils: Added test coverage

Issue #4125, opened 2022-02-24
[Enhancement]: Over partitioning of chunks to make future node balancing optimal


nikkhils:
### What type of enhancement is this?

User experience

### What subsystems and features will be improved?

Access node, Multi-node, Node management, Partitioning

### What does the enhancement do?

To provide multi-node elasticity a data node will be added to an existing multinode cluster. We expect to do chunk rebalancing to allow us to use this newly added data node effectively. 

We need to investigate if over-partitioning of the chunks on existing datanodes will be useful to allow us to do effective rebalancing to new added data nodes. 

We could also consider a max limit of data nodes that we want to support if that helps in doing the over partitioning effectively.

### Implementation challenges

We expect changes in TimescaleDB and thus will be dependant on a release. 

Related issues:

- https://github.com/timescale/timescaledb/issues/3287
- https://github.com/timescale/timescaledb/issues/1899

erimatnor:
Here's a write up of current state of partitioning with a discussion about necessary changes to support re-balancing: https://docs.google.com/document/d/1kWemwjy9J15lOrmj0mjPZsmhrJ3YO6NzmOxz5qAFAL8/edit?usp=sharing

erimatnor:
Transferred from duplicate/overlapping #4254: 

Space partitioning is commonly used in multi-node to assign specific space partitions to data nodes so that partitionwise aggregation can be used. However, when the number of space partitions is changed (e.g., when adding additional data nodes), the space dimension is re-divided among all data nodes, breaking the previous partitioning scheme and preventing partitionwise aggregation.

One way to deal with this is to over-partition initially (i.e., creating more space partitions than there are data nodes) so that an entire space partition (or several) can be given to new data nodes without having to "repartition". However, the current chunk assignment in TimescaleDB is not sticky along the space dimension, so new chunks might not end up on the node that is responsible for the space partition the chunk belongs to.


To fix this issue, the chunk-data node assignment needs to be updated to make it "sticky". In other words, if a space partition is moved to a new data node, then also new chunks in that partition should be created on the right data node. While it might be possible to do this computationally without keeping state, a stateful approach might be preferred.; such an approach is easy to introspect and users might even be able to manually reassign responsibility for space partitions to specific data nodes.

### Implementation challenges

A stateful apprach likely requires a new metadata table that looks something like this:


| partition id   |      range start      |  range end  | data nodes
| --------------- |------------------------|-----------------|
[...]

PR #4443, opened 2022-06-14
Add stateful partition mappings

erimatnor:
Add a new metadata table `dimension_partition` which explicitly and
statefully details how a space dimension is split into partitions, and
(in the case of multi-node) which data nodes are responsible for
storing chunks in each partition. Previously, partition and data nodes
were assigned dynamically based on the current state when creating a
chunk.

This is the first in a series of changes that will add more advanced
functionality over time. For now, the metadata table simply writes out
what was previously computed dynamically in code. Future code changes
will alter the behavior to do smarter updates to the partitions when,
e.g., adding and removing data nodes.

The idea of the `dimension_partition` table is to minimize changes in
the partition to data node mappings across various events, such as
changes in the number of data nodes, number of partitions, or the
replication factor, which affect the mappings. For example, increasing
the number of partitions from 3 to 4 currently leads to redefining all
partition ranges and data node mappings to account for the new
partition. Complete repartitioning can be disruptive to multi-node
deployments. With stateful mappings, it is possible to split an
existing partition without affecting the other partitions (similar to
partitioning using consistent hashing).

Note that the dimension partition table expresses the current state of
space partitions; i.e., the space-dimension constraints and data nodes
to be assigned to new chunks. Existing chunks are not affected by
changes in the dimension partition table, although an external job
could rewrite, move, or copy chunks as desired to comply with the
current dimension partition state. As such, the dimension partition
table represents the "desired" space partitioning state.

Part of #4125

### DIFF
--- a/src/ts_catalog/dimension_partition.c
+++ b/src/ts_catalog/dimension_partition.c
@@ -1,0 +1,415 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+#include <postgres.h>
+#include <access/heapam.h>
+#include <access/xact.h>
+#include <catalog/catalog.h>
+#include <commands/tablecmds.h>
+#include <nodes/parsenodes.h>
+#include <utils/array.h>
+#include <utils/palloc.h>
+#include <utils/rel.h>
+#include <utils/builtins.h>
+#include <utils/snapmgr.h>
+
+#include "ts_catalog/catalog.h"
+#include "dimension.h"
+#include "dimension_slice.h"
+#include "dimension_partition.h"
+#include "hypertable_cache.h"
+#include "scanner.h"
+#include "config.h"
+
+#include "compat/compat.h"
+
+static ScanIterator
+ts_dimension_partition_scan_iterator_create(LOCKMODE lockmode)
+{
+	ScanIterator it = ts_scan_iterator_create(DIMENSION_PARTITION, lockmode, CurrentMemoryContext);
+	it.ctx.flags |= SCANNER_F_NOEND_AND_NOCLOSE;
+
+	return it;
+}
+
+static void
+ts_dimension_partition_scan_iterator_set_dimension_id(ScanIterator *it, int32 dimension_id,
+													  const ScanTupLock *tuplock)
+{
+	it->ctx.index = catalog_get_index(ts_catalog_get(),
+									  DIMENSION_PARTITION,
+									  DIMENSION_PARTITION_DIMENSION_ID_RANGE_START_IDX);
+	ts_scan_iterator_scan_key_reset(it);
+	ts_scan_iterator_scan_key_init(
+		it,
+		Anum_dimension_partition_dimension_id_range_start_idx_dimension_id,
+		BTEqualStrategyNumber,
+		F_INT4EQ,
+		Int32GetDatum(dimension_id));
+	it->ctx.tuplock = tuplock;
+}
+
+/*
+ * Comparison function for dimension partitions.
+ *
+ * Used to sort arrays of partitions on range_start and to lookup a partition
+ * using binary search.
+ */
+static int
+dimpart_cmp(const void *key, const void *node)
+{
+	const DimensionPartition *dp_key = *((const DimensionPartition **) key);
+	const DimensionPartition *dp_node = *((const DimensionPartition **) node);
+
+	/* If the key's range is enclosed (or exactly matches) the partition,
+	 * there's a match */
+	if (dp_key->range_start >= dp_node->range_start && dp_key->range_end < dp_node->range_end)
+		return 0;
+	else if (dp_key->range_start < dp_node->range_start)
+		return -1;
+	else
+	{
+		Assert(dp_key->range_start > dp_node->range_start);
+		return 1;
+	}
+}
+
+static DimensionPartition *
+dimpart_create(const HeapTuple tup, const TupleDesc tupdesc)
+{
+	DimensionPartition *dp = palloc0(sizeof(DimensionPartition));
+	Datum values[Natts_dimension_partition];
+	bool isnull[Natts_dimension_partition] = { false };
+
+	heap_deform_tuple(tup, tupdesc, values, isnull);
+
+	dp->dimension_id =
+		DatumGetInt32(values[AttrNumberGetAttrOffset(Anum_dimension_partition_dimension_id)]);
+	dp->range_start =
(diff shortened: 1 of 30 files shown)


sql/ddl_internal.sql
```diff
@@ -50,3 +50,4 @@ BEGIN
     RAISE 'subscription sync wait timedout';
 END
 $BODY$ SET search_path TO pg_catalog, pg_temp;
+
```
nikkhils, line 53: unintentional change?

pmwkaa (approved):
Looks good to me, mostly comments and observations. I hope I didn't miss anything important.

src/ts_catalog/dimension_partition.h
```diff
@@ -0,0 +1,37 @@
...
+#include <lib/rbtree.h>
+
+#include "scan_iterator.h"
+
+typedef struct DimensionPartition
+{
+	RBTNode rbtnode;
+	int32 dimension_id;
+	int64 range_start;
+	int64 range_end;
```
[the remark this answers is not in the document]
erimatnor: Yeah, we cache it here since we need it to create chunks after we look up the dimension partition. We don't want to have to look up two partitions during chunk creation.

The cached information is thrown away every time we update the dimension_partition information, so it will be recalculated automatically.

I opted to not embed the Form since I am converting the data nodes from array form to list form

src/ts_catalog/dimension_partition.c
```diff
@@ -0,0 +1,432 @@
...
+ * Used also to lookup the matching partition for a particular point, which
+ * will be disguised as a partition with range_start==range_end.
+ *
+ * The "key" is the user-passed argument that we are matching against the
+ * tree.
+ *
+ * The "node" is the node in the tree that we are comparing against.
+ */
+static int
+dimpart_cmp(const RBTNode *key, const RBTNode *node, void *arg)
```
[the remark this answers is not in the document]
erimatnor: Sure, array and binary search should work the same, just different implementation. I can look quickly on whether it would simplify the code or not but otherwise probably not worthwhile making the change since it won't make a huge difference.

src/ts_catalog/dimension_partition.c
```diff
@@ -0,0 +1,432 @@
...
+Datum
+ts_dimension_partition_update(PG_FUNCTION_ARGS)
+{
+	Oid hypertable_relid = PG_GETARG_OID(0);
+	Cache *hcache;
+	const Hypertable *ht =
+		ts_hypertable_cache_get_cache_and_entry(hypertable_relid, CACHE_FLAG_NONE, &hcache);
+	ts_hypertable_update_dimension_partitions(ht);
+	ts_cache_release(hcache);
+	PG_RETURN_VOID();
```
nikkhils, line 431: In general, this file has a lot of warnings about less code coverage. We can try to increase tests which cover atleast some of the warnings

tsl/test/sql/data_node.sql
```diff
@@ -16,6 +16,16 @@
 \set DN_DBNAME_5 :TEST_DBNAME _5
 \set DN_DBNAME_6 :TEST_DBNAME _6
 
+-- View to see dimension partitions. Note RIGHT JOIN to see that
+-- dimension partitions are cleaned up (deleted) properly.
+CREATE VIEW hypertable_partitions AS
+SELECT table_name, dimension_id, range_start, data_nodes
+FROM _timescaledb_catalog.hypertable h
```
nikkhils, line 23: Just wondering if it makes sense to have this view available via our catalogs? Might be a good way for users/devs to check the contents.

src/ts_catalog/dimension_partition.c
```diff
@@ -0,0 +1,432 @@
...
+	Assert(num_partitions > 0);
+	Assert(data_nodes == NIL || replication_factor > 0);
+
+	/* Delete all partitions for the dimension */
+	dimension_partition_info_delete(dimension_id, SCANNER_F_KEEPLOCK, RowExclusiveLock);
+
+	/* Lock already held */
+	rel = table_open(relid, NoLock);
+
+	tree = rbt_create(sizeof(DimensionPartition),
```
nikkhils, line 365: From where are we getting this rbt_create API?

nikkhils (approved):
Overall LGTM! This is a large patch and I have added a few comments/questions.

src/ts_catalog/dimension_partition.c
erimatnor: It's standard PG library

src/ts_catalog/dimension_partition.c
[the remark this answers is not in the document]
erimatnor: Switched to array+binary search. Made the code a bit cleaner, so thanks for the suggestion.

src/ts_catalog/dimension_partition.c
erimatnor: Switched to array+binary search, so no longer using rb-tree.

src/ts_catalog/dimension_partition.c
erimatnor: Switched to array+binary search and this should improve test coverage as there will be fever unused rb-tree functions.

src/ts_catalog/dimension_partition.c
```diff
@@ -0,0 +1,414 @@
...
+	 * array expansion below */
+	unsigned int max_partitions = 2;
+#else
+	unsigned int max_partitions = 20;
+#endif
+
+	it = ts_dimension_partition_scan_iterator_create(AccessShareLock);
+	ts_dimension_partition_scan_iterator_set_dimension_id(&it, dimension_id, NULL);
+	/* Create a temporary array for partitions, which is likely too big */
+	partitions = palloc0(sizeof(DimensionPartition *) * max_partitions);
```
pmwkaa, line 143: nit: maybe there is a simple way to get a number of rows in the dimension partition table?

pmwkaa (approved):

src/ts_catalog/dimension_partition.c
```diff
@@ -0,0 +1,414 @@
...
+		return NULL;
+	}
+	else if (count > 1)
+	{
+		partitions[count - 2]->range_end = partitions[count - 1]->range_start;
+	}
+
+	dpi = palloc0(sizeof(DimensionPartitionInfo));
+	dpi->num_partitions = count;
+	dpi->partitions = palloc0(sizeof(DimensionPartition *) * count);
```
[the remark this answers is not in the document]
erimatnor: Not sure I follow. A memory block matching the actual number of partitions is allocated here as opposed to the "temporary" block used before we know `count`. The temporary memory block is freed below.

PR #4449, opened 2022-06-16
Do not allow distributed ddl during create extension

pmwkaa:
This change fixes logic for the extension loading check, by
moving it before the distributed_ddl guc check


nikkhils (approved):
LGTM.

PR #4450, opened 2022-06-16
Retain hypertable ownership on attach_data_node

### LINKED ISSUE #4433: Set original permissions for created tables on data nodes
When we invoke the `attach_data_node` function, the current user will be used for the creation of all tables on the attached data node.
That is undesired, as it leads to permission conflicts like that:

```
TimescaleDB: 2022-06-10 07:25:04 UTC [152]: [62a2f1d0.98-1] multipostgres@tsdb,app=timescaledb [00000] LOG:  execute <unnamed>: CREATE TABLE "public"."dtable1" ("time" timestamp with time zone NOT NULL, "device" integer, "temp" double precision) USING "heap" ;
TimescaleDB: 2022-06-10 07:25:04 UTC [152]: [62a2f1d0.98-2] multipostgres@tsdb,app=timescaledb [00000] LOG:  execute <unnamed>: CREATE INDEX dtable1_time_idx ON public.dtable1 USING btree ("time" DESC);

TimescaleDB: 2022-06-10 09:55:46 UTC [9616]: [62a31522.2590-1] tsdbadmin@tsdb,app=timescaledb [42501] ERROR:  permission denied for table "dtable1"
TimescaleDB: 2022-06-10 09:55:46 UTC [9616]: [62a31522.2590-2] tsdbadmin@tsdb,app=timescaledb [42501] DETAIL:  Insert privileges required on "dtable1" to create chunks.
```

What happens is here the following:
1. Create a table as `tsdbadmin`, then convert it to a DHT.
2. A new data node is added.
3. A superuser `multipostgres` invokes the `attach_data_node` function.
4. TimescaleDB extension creates the tables and indices accordingly but uses the `multipostgres` user instead of the original `tsdbadmin`.
5. Operations for `tsdbadmin` on the `dtable1` now fail on that data node.

It's possible that a multinode cluster will host many users with different privileges, so to correctly attach a new data node to all the tables that exist in the cluster we either need to use a superuser (as in the case above) or check to whom does the table belong and assume their role before invoking the function. I believe that the timescaledb extension should reinstate the original permissions so they are the same on all nodes.

nikkhils:
If a superuser is used to invoke attach_data_node on a hypertable then
we need to ensure that the object created on this data node has the
same original ownership permissions.
    
Fixes #4433


tsl/src/data_node.c
```diff
@@ -899,6 +902,20 @@ data_node_attach(PG_FUNCTION_ARGS)
...
 
+	/*
+	 * Change to the hypertable owner so that the same permissions will be set up on the
+	 * datanode being attached to as well. We need to do this explicitly because the
+	 * caller of this function could be a superuser and we definitely don't want to create
+	 * this hypertable with superuser ownership on the datanode being attached to!
+	 */
+	rel = table_open(ht->main_table_relid, AccessShareLock);
+	uid = rel->rd_rel->relowner;
+	table_close(rel, AccessShareLock);
```
erimatnor, line 913: Maybe we want to keep this lock until the end of the transaction?

I am thinking of the case where there's a concurrent change of owner and maybe there's a race condition. For example, we release the lock here, then another process changes the owner (which only applies to the nodes attached before this command completes), then this code proceeds with creating the hypertable on the data node with the old owner.



```suggestion
	table_close(rel, NoLock);
```

erimatnor: Maybe it is even possible to create an isolation test for that case?

erimatnor: Maybe the lock type should also be `AccessExclusiveLock` to block other operations while we attach the node, e.g., owner change.

nikkhils: The `owner change` takes an `AccessExclusiveLock` so it should be ok enough to close the table with `NoLock` here. Done with that change.

erimatnor: Ok, that seems reasonable.

erimatnor (approved):

pmwkaa (approved):

PR #4459, opened 2022-06-20
Allow database owner to do pg_dump/restore

konskov:
First attempt. Don't know if we need to call GRANT in another place.
Maybe would be better to call it from our C code.


sql/pre_install/tables.sql
```diff
@@ -497,3 +497,21 @@ GRANT SELECT ON ALL SEQUENCES IN SCHEMA _timescaledb_catalog TO PUBLIC;
...
+BEGIN
+    SELECT pg_catalog.pg_get_userbyid(d.datdba) as db_owner
+    FROM pg_catalog.pg_database d
+    WHERE d.datname = 'test'
+    ORDER BY db_owner limit 1 into db_owner;
+
+    -- raise LOG 'db_owner is %', db_owner;
+    -- in tests the owner is NULL
+    -- can that be the case in other scenarios as well?
+    IF coalesce(db_owner, '') <> '' then
```
fabriziomello, line 514:
```suggestion
    IF db_owner IS NOT NULL then
```

src/loader/bgw_interface.c
```diff
@@ -69,24 +70,30 @@ ts_bgw_db_workers_start(PG_FUNCTION_ARGS)
...
+/*
+ * allow superusers OR database owners to stop and restart background workers
+ * database owners must be allowed to do this in order to do pg_dump/restore
+ */
 Datum
 ts_bgw_db_workers_stop(PG_FUNCTION_ARGS)
 {
-	if (!superuser())
+	bool db_owner = pg_database_ownercheck(MyDatabaseId, GetUserId());
+	if (!superuser() && !db_owner)
```
fabriziomello, line 81:
```suggestion
	if (!pg_database_ownercheck(MyDatabaseId, GetUserId()))
```
Isn't more simpler in that way? The `pg_database_ownercheck` also check for `superuser`. I mean will return true (bypass) the check for super users.

src/loader/bgw_interface.c
```diff
@@ -69,24 +70,30 @@ ts_bgw_db_workers_start(PG_FUNCTION_ARGS)
...
 
 	PG_RETURN_BOOL(ts_bgw_message_send_and_wait(STOP, MyDatabaseId));
 }
 
 Datum
 ts_bgw_db_workers_restart(PG_FUNCTION_ARGS)
 {
-	if (!superuser())
+	bool db_owner = pg_database_ownercheck(MyDatabaseId, GetUserId());
+	if (!superuser() && !db_owner)
```
fabriziomello, line 93:
```suggestion
	if (!pg_database_ownercheck(MyDatabaseId, GetUserId()))
```
Same as above.

sql/pre_install/tables.sql
```diff
@@ -497,3 +497,19 @@ GRANT SELECT ON ALL SEQUENCES IN SCHEMA _timescaledb_catalog TO PUBLIC;
...
+    SELECT pg_catalog.pg_get_userbyid(d.datdba) as db_owner
+    FROM pg_catalog.pg_database d
+    WHERE d.datname = 'test'
+    ORDER BY db_owner limit 1 into db_owner;
+
+    -- in tests the owner is NULL
+    IF db_owner IS NOT NULL then
+      EXECUTE format('GRANT SELECT, UPDATE, INSERT, DELETE ON _timescaledb_catalog.metadata TO %I', db_owner);
+    end if;
+END $$;
```
fabriziomello, line 515:
```suggestion
SELECT
  format('GRANT SELECT, UPDATE, INSERT, DELETE ON _timescaledb_catalog.metadata TO %I', pg_catalog.pg_get_userbyid(d.datdba))
FROM
  pg_catalog.pg_database d
WHERE
  d.datname = current_database()
\gexec
```

fabriziomello:
Now is time to add a TAP test to really execute a pg_dump/pg_restore and also would be good to describe better what you want to achieve with this issue in the description

PR #4460, opened 2022-06-20
Inspect pg_config.h for SSL support

jchampio:
We parse apart the output of `pg_config --configure` to see whether or not OpenSSL is enabled. This is a bit brittle; it has broken in the past with the change from `--with-openssl` to `--with-ssl`, and upstream is currently testing a change to the build system (Meson, which doesn't rely on autotools during configuration) that is likely to further interfere with this approach.

As an alternative, we can just look at the header files to get this information. `USE_OPENSSL` is not defined in pg_config.h if SSL support is not compiled in.

Tested with PG15, both autotools and Meson, with and without OpenSSL support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.10)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
 include(CheckCCompilerFlag)
+include(CheckSymbolExists)
 include(GitCommands)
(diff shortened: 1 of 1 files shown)


jchampio:
Rebased to squash the `cmake-format` fix.

mkindahl (commented):
LGTM.

mkindahl (approved):

akuzm (approved):

svenklemm (approved):

jchampio:
Force-push to try to bump Codecov out of its fugue.

PR #4461, opened 2022-06-20
Update commit-msg hook to Python 3 and tweak column counting

jchampio:
I ran CMake on the project and suddenly I could no longer commit; my machine doesn't have Python 2, so the commit-msg hook failed, taking my commit messages with it. First commit updates the hook to Python 3 with a run through `2to3` and some manual tweaks to string interpolation.

After the update to Python 3, the commit-msg hook warned that I had overlong lines in the commit message for the fix. I thought that was odd since my editor enforces 50/72 for commits; turns out the logic was counting the line feed at the end of the line. Second commit adjusts this with an `rstrip('\n')`.

Disable-check: commit-count

### DIFF
--- a/scripts/githooks/commit_msg.py
+++ b/scripts/githooks/commit_msg.py
@@ -60,7 +60,7 @@ class GitCommitMessage:
 
     def check_subject_limit(self):
         'Rule 2: Limit the subject line to 50 characters'
-        return len(self.subject) <= 50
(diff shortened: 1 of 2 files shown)


mkindahl (approved):
LGTM

fabriziomello (approved):

svenklemm (approved):

PR #4464, opened 2022-06-21
Add api to associate hypertable with custom jobs

gayyappan:
This PR introduces a new SQL function to associate a
hyperatable or continuous agg with a custom job. If
this dependency is setup, the job is automatically
deleted when the hypertable/cagg is dropped.

Reviewers: Need this functionality  for policy functions in OSM  extension.

### DIFF
--- a/sql/job_api.sql
+++ b/sql/job_api.sql
@@ -28,3 +28,9 @@ CREATE OR REPLACE FUNCTION @extschema@.alter_job(
 RETURNS TABLE (job_id INTEGER, schedule_interval INTERVAL, max_runtime INTERVAL, max_retries INTEGER, retry_period INTERVAL, scheduled BOOL, config JSONB, next_start TIMESTAMPTZ)
 AS '@MODULE_PATHNAME@', 'ts_job_alter'
 LANGUAGE C VOLATILE;
+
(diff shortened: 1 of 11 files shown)


svenklemm:
Since hypertable_id is NULLABLE calling it with NULL hypertable should be valid and set hypertable_id accordingly.  We probably want a permission check that job owner and hypertable owner match, otherwise this would allow people to delete arbitrary jobs.

konskov (commented):
I also think it would be good to have a check for ownership match between hypertable/cagg and job, as Sven suggested.

gayyappan:
> Since hypertable_id is NULLABLE calling it with NULL hypertable should be valid and set hypertable_id accordingly. We probably want a permission check that job owner and hypertable owner match, otherwise this would allow people to delete arbitrary jobs.

Fixed.

konskov:
Just a comment, though since this is an internal api, I don’t know if it is relevant/important.
When no hypertable is associated with the job, maybe we could return null, or another value, instead of returning the job_id, to indicate that no association was made?

fabriziomello (approved):

svenklemm (approved):

konskov (approved):

PR #4470, opened 2022-06-24
Block drop chunk if chunk is in frozen state

gayyappan:
A chunk in frozen state cannot be dropped.
drop_chunks will skip over frozen chunks without erroring.
Internal api , drop_chunk will error if you attempt to  drop
a chunk without unfreezing it.

This PR also adds a new internal API to unfreeze a chunk.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -71,6 +71,7 @@ TS_FUNCTION_INFO_V1(ts_chunk_drop_chunks);
 TS_FUNCTION_INFO_V1(ts_chunk_drop_single_chunk);
 TS_FUNCTION_INFO_V1(ts_chunk_attach_osm_table_chunk);
 TS_FUNCTION_INFO_V1(ts_chunk_freeze_chunk);
+TS_FUNCTION_INFO_V1(ts_chunk_unfreeze_chunk);
 TS_FUNCTION_INFO_V1(ts_chunks_in);
 TS_FUNCTION_INFO_V1(ts_chunk_id_from_relid);
 TS_FUNCTION_INFO_V1(ts_chunk_show);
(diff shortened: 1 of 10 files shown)


src/chunk.c
```diff
@@ -4003,6 +4050,29 @@ ts_chunk_freeze_chunk(PG_FUNCTION_ARGS)
...
+	Oid chunk_relid = PG_ARGISNULL(0) ? InvalidOid : PG_GETARG_OID(0);
+	TS_PREVENT_FUNC_IF_READ_ONLY();
+	Chunk *chunk = ts_chunk_get_by_relid(chunk_relid, true);
+	Assert(chunk != NULL);
+	if (chunk->relkind == RELKIND_FOREIGN_TABLE)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("operation not supported on distributed chunk or foreign table \"%s\"",
+						get_rel_name(chunk_relid))));
```
fabriziomello, line 4065: Add some test for this should be good

src/chunk.h
```diff
@@ -186,6 +186,7 @@ extern void ts_chunks_rename_schema_name(char *old_schema, char *new_schema);
 
 extern TSDLLEXPORT bool ts_chunk_set_unordered(Chunk *chunk);
 extern TSDLLEXPORT bool ts_chunk_set_frozen(Chunk *chunk);
+extern TSDLLEXPORT bool ts_chunk_undo_frozen(Chunk *chunk);
```
fabriziomello, line 189: Nit. The name is a bit weird, what about something like `ts_chunk_unset_frozen` or `ts_chunk_set_unfrozen`

gayyappan: renamed.

fabriziomello (approved):

akuzm (approved):

PR #4483, opened 2022-06-30
Add more tests for distributed INSERT and COPY

akuzm:
More interleavings of INSERT/COPY, and test with slow recv() to check
waiting.

Also fix some flakiness while we're at it.

Fixes https://github.com/timescale/timescaledb/issues/4485

### DIFF
--- a/tsl/test/shared/expected/dist_remote_error.out
+++ b/tsl/test/shared/expected/dist_remote_error.out
@@ -5,7 +5,6 @@
 \unset ECHO
 -- Disable SSL to get stable error output across versions. SSL adds some output
 -- that changed in PG 14.
-\c -reuse-previous=on sslmode=disable
 set timescaledb.debug_enable_ssl to off;
 set client_min_messages to error;
 -- A relatively big table on one data node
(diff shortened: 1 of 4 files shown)


mkindahl (approved):
You are missing a few newlines at the end of files here and there, otherwise looks good.

tsl/test/shared/expected/dist_remote_error.out
```diff
@@ -111,15 +110,42 @@ select table_name from create_distributed_hypertable('metrics_dist_br',
...
+   table_name    
+ metrics_dist_bl
+(1 row)
+
+-- Test INSERT and COPY with slow data node.
+\copy metrics_dist_bl from 'dist_remote_error.text' with (format text);
+insert into metrics_dist_bl select * from metrics_dist_remote_error;
+select count(*) from metrics_dist_bl;
+ count 
+ 40000
```
pmwkaa, line 147: That would be 40 second test to run, right? I hope that's not too long

pmwkaa (approved):

tsl/test/shared/expected/dist_remote_error.out
akuzm: The entire test runs about 20 s, because only about 5k rows go to the third data node where the recv() is sleepy.

PR #4484, opened 2022-06-30
 Use non-blocking data node connections for COPY

akuzm:
This PR switches the postgres connections to data nodes to non-blocking mode for the duration of the COPY. Other operations on these connections are unaffected.

It serves as a preparation for working with multiple connections simultaneously. No functional changes are introduced, but it's done as a separate PR to simplify review and debugging/bisect in case we introduce an error.

Part of https://github.com/timescale/timescaledb/pull/4285

### DIFF
--- a/tsl/src/remote/connection.c
+++ b/tsl/src/remote/connection.c
@@ -749,7 +749,7 @@ remote_connection_create(PGconn *pg_conn, bool processing, const char *node_name
 	conn->ln.next = conn->ln.prev = NULL;
 	conn->pg_conn = pg_conn;
 	conn->closing_guard = false;
-	conn->status = processing ? CONN_PROCESSING : CONN_IDLE;
+	remote_connection_set_status(conn, processing ? CONN_PROCESSING : CONN_IDLE);
 	namestrcpy(&conn->node_name, node_name);
 	conn->tz_name = NULL;
 	conn->autoclose = true;
@@ -1834,7 +1834,7 @@ remote_connection_drain(TSConnection *conn, TimestampTz endtime, PGresult **resu
 			if (res == NULL)
 			{
 				/* query is complete */
-				conn->status = CONN_IDLE;
+				remote_connection_set_status(conn, CONN_IDLE);
 				connresult = CONN_OK;
 				break;
 			}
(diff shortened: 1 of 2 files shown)


tsl/src/remote/connection.c
```diff
@@ -2176,8 +2176,17 @@ remote_connection_begin_copy(TSConnection *conn, const char *copycmd, bool binar
 	if (binary && !send_binary_copy_header(conn, err))
 		goto err_end_copy;
 
+	if (PQsetnonblocking(pg_conn, 1))
```
erimatnor, line 2179: What's the actual return value here for success?

When reading this code, it seems like we are generating an error on success, but I guess it returns `false` on success? If we look at `PQisnonblocking()` below, it is the other way around. That looks like an inconsistency.

If the return value is `0` on success, I think we should explicitly check for that to make things clear, e.g., if `(PQsetnonblocking(pg_conn, 1) != 0)`

tsl/src/remote/connection.c
```diff
@@ -2214,18 +2223,96 @@ send_end_binary_copy_data(const TSConnection *conn, TSConnectionError *err)
...
+				/* Error. */
+				return fill_simple_error(err,
+										 ERRCODE_CONNECTION_EXCEPTION,
+										 "failed to flush the COPY connection",
+										 conn);
+			}
+		}
+
+		/* Switch the connection into blocking mode. */
+		if (PQsetnonblocking(conn->pg_conn, 0))
```
erimatnor, line 2284: Again, the success check for this command seems inconsistent with `PQisnonblocking()`. I would recommend explicit return value check.

erimatnor (commented):
Not entirely sure what this does right now. Is it actually using connections in non-blocking mode? How does it affect other parts of the code that might expect blocking sockets?

akuzm:
>Not entirely sure what this does right now. Is it actually using connections in non-blocking mode? How does it affect other parts of the code that might expect blocking sockets?

I updated the description. It doesn't do much by itself, and doesn't affect operations other than COPY, but I still want to merge it separately to simplify review/debugging. The next PR will use this to send the COPY data to the data nodes concurrently.

svenklemm (approved):

mkindahl (approved):

PR #4492, opened 2022-07-05
Speed up chunk search by restriction clauses

akuzm:
We don't have to look up the dimension slices for dimensions for which
we don't have restrictions.

This also fixes a planning time regression introduced in 2.7. It was introduced in this commit https://github.com/timescale/timescaledb/commit/37190e8a8#diff-e488f45c83647e5d9415d438f2e79eb2780639b7888c7533890b6c577e02d03dL852-L857
We stopped using the fast path for case where there are no restrictions, and the normal path turned out to be slower. This commit speeds it up to the previous level.

The lookup logic this PR introduces is basically the same as the one introduced for `chunk_point_find_chunk_id` by https://github.com/timescale/timescaledb/pull/4390 . The difference is that here the conditions may be underspecified, so multiple chunks match.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -4,67 +4,68 @@
  * LICENSE-APACHE for a copy of the license.
  */
 #include <postgres.h>
-#include <catalog/pg_class.h>
-#include <catalog/namespace.h>
-#include <catalog/pg_trigger.h>
-#include <catalog/indexing.h>
-#include <catalog/pg_inherits.h>
-#include <catalog/toasting.h>
-#include <commands/trigger.h>
-#include <commands/tablecmds.h>
-#include <commands/defrem.h>
-#include <tcop/tcopprot.h>
+
 #include <access/htup.h>
 #include <access/htup_details.h>
-#include <access/xact.h>
 #include <access/reloptions.h>
+#include <access/tupdesc.h>
+#include <access/xact.h>
+#include <catalog/indexing.h>
+#include <catalog/namespace.h>
+#include <catalog/pg_class.h>
+#include <catalog/pg_inherits.h>
+#include <catalog/pg_trigger.h>
+#include <catalog/pg_type.h>
+#include <catalog/toasting.h>
+#include <commands/defrem.h>
+#include <commands/tablecmds.h>
+#include <commands/trigger.h>
+#include <executor/executor.h>
+#include <fmgr.h>
+#include <funcapi.h>
+#include <miscadmin.h>
+#include <nodes/execnodes.h>
 #include <nodes/makefuncs.h>
+#include <storage/lmgr.h>
+#include <tcop/tcopprot.h>
+#include <utils/acl.h>
 #include <utils/builtins.h>
+#include <utils/datum.h>
+#include <utils/hsearch.h>
 #include <utils/lsyscache.h>
 #include <utils/syscache.h>
-#include <utils/hsearch.h>
-#include <storage/lmgr.h>
-#include <miscadmin.h>
-#include <funcapi.h>
-#include <fmgr.h>
-#include <utils/datum.h>
-#include <catalog/pg_type.h>
-#include <utils/acl.h>
 #include <utils/timestamp.h>
-#include <nodes/execnodes.h>
-#include <executor/executor.h>
-#include <access/tupdesc.h>
 
-#include "export.h"
-#include "debug_point.h"
 #include "chunk.h"
+
+#include "bgw_policy/chunk_stats.h"
+#include "cache.h"
 #include "chunk_index.h"
-#include "ts_catalog/chunk_data_node.h"
+#include "chunk_scan.h"
+#include "compat/compat.h"
 #include "cross_module_fn.h"
-#include "ts_catalog/catalog.h"
-#include "ts_catalog/continuous_agg.h"
-#include "cross_module_fn.h"
+#include "debug_point.h"
 #include "dimension.h"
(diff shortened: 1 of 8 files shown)


erimatnor:
For reviewing purposes, it would be good to understand what change introduced the regression.

akuzm:
https://grafana.ops.savannah-dev.timescale.com/d/NdmLnOk4z/compare-benchmark-runs?orgId=1&var-branch=main&var-branch=release&var-branch=pr&var-postgres=14&var-run1=379&var-run2=397&var-threshold=0.02

Looks like this does improve some queries, although not all of them.

src/hypertable_restrict_info.c
```diff
@@ -524,6 +549,18 @@ gather_restriction_dimension_vectors(const HypertableRestrictInfo *hri)
 														   open->lower_strategy,
 														   open->lower_bound);
 
+				/*
+				 * If we have a nontrivial condition for the second index column
```
svenklemm, line 553: What does nontrivial mean exactly?

src/chunk.c
```diff
@@ -1536,6 +1537,84 @@ ts_chunk_create_for_point(const Hypertable *ht, const Point *p, const char *sche
...
+	ScanIterator iterator = ts_chunk_constraint_scan_iterator_create(CurrentMemoryContext);
+
+	ListCell *lc;
+	foreach (lc, dimension_vecs)
+	{
+		const DimensionVec *vec = lfirst(lc);
+		/*
+		 * We shouldn't see a dimension with zero matching dimension slices.
+		 * That would mean that no chunks match at all, this should have been
+		 * handled earlier by gater_restriction_dimension_vectors().
```
svenklemm, line 1563:
```suggestion
		 * handled earlier by gather_restriction_dimension_vectors().
```

src/hypertable_restrict_info.c
svenklemm: Hmm that is weird why do you need backwards scan for constraints on the 2nd index column scan direction shouldnt matter for that.

src/hypertable_restrict_info.c
```diff
@@ -524,6 +549,18 @@ gather_restriction_dimension_vectors(const HypertableRestrictInfo *hri)
...
+				 * range_start, use a backward scan direction, so that the index
+				 * is able to use the second column as well to choose the
+				 * starting point.
+				 * If not, prefer forward direction, because backwards scan is
+				 * slightly slower for some reason.
+				 * Ideally we need some other index type than btree for this.
```
svenklemm, line 559: Why do you need non-btree index here, what would we gain by other index type.

src/hypertable_restrict_info.c
akuzm: If you have `<=` for the 2nd column, and scan direction is forward, you can't use the 2nd column to find the starting point for the scan. So you use only the first column, and go through the rest of values sequentially. For backwards, you know where the first record is that matches both the conditions on the 1st and the 2nd columns.

More about this in postgres code: https://github.com/postgres/postgres/blob/master/src/backend/access/nbtree/nbtsearch.c#L928

We have this optimization in several other places, so I decided to add it here as well.

src/hypertable_restrict_info.c
akuzm: The btree index is not so suitable for queries like `find interval that contains the given point` and `find intervals that intersect with the given interval`. Idk what it should be, we also have this dimension id column on which we should filter first.

akuzm: I mean, in all queries we make the third column (range_end) is only compared in a sequential scan over index entries, not used to find the starting/ending point.

svenklemm (approved):

src/hypertable_restrict_info.c
```diff
@@ -595,11 +624,64 @@ gather_restriction_dimension_vectors(const HypertableRestrictInfo *hri)
...
 ts_hypertable_restrict_info_get_chunks(HypertableRestrictInfo *hri, Hypertable *ht,
-									   LOCKMODE lockmode, unsigned int *num_chunks)
+									   unsigned int *num_chunks)
 {
-	List *dimension_vecs = gather_restriction_dimension_vectors(hri);
-	Assert(hri->num_dimensions == ht->space->num_dimensions);
-	return ts_chunk_scan_by_constraints(ht->space, dimension_vecs, lockmode, num_chunks);
+	/*
+	 * Filter out the dimensions for which we don't have a nontrivial
+	 * restriction.
```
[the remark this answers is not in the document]
erimatnor: Btw, do we need to include "trivial" restrictions in `HypertableRestrictInfo` to begin with? Why not exclude them from the beginning so that we need not filter them out later?

erimatnor (approved):
Looks ok, but I left some comments you might want to review.

src/hypertable_restrict_info.c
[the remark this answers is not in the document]
akuzm: > Btw, do we need to include "trivial" restrictions in HypertableRestrictInfo to begin with? Why not exclude them from the beginning so that we need not filter them out later?

Originally I just didn't want to change the algorithm too much, because it would require different code for initializing the dimension restrict infos on demand.

But now I think this approach also has value for queries when there are multiple conditions on a dimension, that end up adding to the entire dimension. This isn't likely to arise in simple hand-written queries, but more likely in auto-generated queries or more complex queries that use views. So we can consider this approach of "removing empty ones in the end" to be an additional optimization.

Issue #2818, opened 2021-01-13
Support adding dimension when data exists


erimatnor:
We should allow users to add dimensions when data already exists. The new dimension would have a single partition/slice for any existing chunks. Metadata for those chunks would have to be updated, including constraints on chunk tables.

Useful specifically for migrating a regular hypertable to a distributed hypertable.

cevian:
Also useful for adding space partitions to a distributed hypertable. Use case: we have a lot of metric hypertables. Most of them should not be space partitioned to begin with. That way, different metrics go on different nodes in a round-robin fashion but a single metric, for a single time, is on one node. But, if one metric gets too big, it may become useful to partition that metric. The straightforward way to do that would be to add a space dimension to that particular metric.

NunoFilipeSantos:
Thanks for the enhancement proposal @erimatnor !

yaitskov:
+1. I don't see any value in add_dimension function until it can work with nonempty tables. If table should be empty then it is easy to drop table and create it from scratch with new dimension set.

Is there workaround?

richardARPANET:
@yaitskov Will `add_dimension` work if I TRUNCATE the data, run `add_dimension` and re-populate the table?

yaitskov:
@richardARPANET , It was working. I don't know about current bugs.
I haven't worked with timescale db for months.
As far as remember I use aliases - a meta table keeping relation between alias and real table.
Once table structure changes I create new table and reassign the alias.
You can play Hanoi game with a table if you want to keep physical name untouched ;)

yaitskov:
As for me updating physical structure for adding/removing columns in a table with zillion rows is a dead end. An atomic table rename function (or swap names for 2 tables) would suit this use case better. It requires minimum work on system internals.
Background live synchronization is already doable on application side plus triggers.

guikubivan:
Looks like this is now possible, but there's no mention of this functionality in https://docs.timescale.com/api/latest/hypertable/add_dimension/#add-dimension. Can we use this the same and assume it will work?

PR #4495, opened 2022-07-05
Support add_dimension() with existing data

pmwkaa:
This change allows to create new dimensions even with
existing chunks.

It does not modify any existing data or do migration,
instead it creates full-range (-inf/inf) dimension slice for
existing chunks in order to be compatible with newly created
dimension.

All new chunks created after this will follow logic of the new
dimension and its partitioning.

Issue: https://github.com/timescale/timescaledb/issues/2818

### DIFF
--- a/src/dimension.c
+++ b/src/dimension.c
@@ -15,6 +15,7 @@
 #include <funcapi.h>
 #include <miscadmin.h>
 #include <nodes/makefuncs.h>
+#include <storage/lmgr.h>
 
 #include "ts_catalog/catalog.h"
 #include "compat/compat.h"
@@ -30,6 +31,7 @@
 #include "time_utils.h"
 #include "utils.h"
 #include "errors.h"
+#include "debug_point.h"
 
 /* add_dimension record attribute numbers */
 enum Anum_add_dimension
@@ -1504,13 +1506,14 @@ ts_dimension_add(PG_FUNCTION_ARGS)
 	 * means, that when this function is called from create_hypertable()
 	 * instead of directly, num_dimension is already set to one. We therefore
 	 * need to lock the hypertable tuple here so that we can set the correct
-	 * number of dimensions once we've added the new dimension
+	 * number of dimensions once we've added the new dimension.
+	 *
+	 * This lock is also used to serialize access from concurrent add_dimension()
+	 * call and a chunk creation.
 	 */
-	if (!ts_hypertable_lock_tuple_simple(info.table_relid))
-		ereport(ERROR,
-				(errcode(ERRCODE_LOCK_NOT_AVAILABLE),
-				 errmsg("could not lock hypertable \"%s\" for update",
-						get_rel_name(info.table_relid))));
+	LockRelationOid(info.table_relid, ShareUpdateExclusiveLock);
+
+	DEBUG_WAITPOINT("add_dimension_ht_lock");
 
 	info.ht = ts_hypertable_cache_get_cache_and_entry(info.table_relid, CACHE_FLAG_NONE, &hcache);
 
@@ -1529,14 +1532,6 @@ ts_dimension_add(PG_FUNCTION_ARGS)
 	if (!info.skip)
 	{
 		int32 dimension_id;
-
-		if (ts_hypertable_has_chunks(info.table_relid, AccessShareLock))
-			ereport(ERROR,
-					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-					 errmsg("hypertable \"%s\" has data or empty chunks",
-							get_rel_name(info.table_relid)),
-					 errdetail("It is not possible to add dimensions to a hypertable that has "
-							   "chunks. Please truncate the table.")));
 
 		/*
 		 * Note that space->num_dimensions reflects the actual number of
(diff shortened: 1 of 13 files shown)


src/dimension.c
```diff
@@ -1558,6 +1550,38 @@ ts_dimension_add(PG_FUNCTION_ARGS)
...
+			slice = ts_dimension_slice_create(dimension_id,
+											  DIMENSION_SLICE_MINVALUE,
+											  DIMENSION_SLICE_MAXVALUE);
+			ts_dimension_slice_insert_multi(&slice, 1);
+
+			foreach (lc, chunk_id_list)
+			{
+				int chunk_id = lfirst_int(lc);
+				Chunk *chunk = ts_chunk_get_by_id(chunk_id, true);
+				ChunkConstraint *cc = ts_chunk_constraints_add(chunk->constraints,
```
erimatnor, line 1577: What's the locking semantics here? Do we need a lock on the chunk since we're making modifications?

Often the semantics are:

Check for chunk -> get lock -> check again that chunk exists after locking

erimatnor: Would be good to have an isolation test checking for locking/concurrency issues, for example:

* Running concurrent add_dimension() on the same table in two different session
* Running add_dimension() concurrently with creating new chunk (obviously the new chunk should also have the dimension)

This can perhaps be achieved by adding a waitlock in this code, after retrieving the "Chunk *" but before adding the constraints. Then running an insert to create a new chunk (and/or, calling explicit chunk creation function used by copy/move chunks)

test/expected/create_hypertable.out
```diff
@@ -235,14 +235,169 @@ ERROR:  column "location" is already a dimension
...
+ chunk_id | hypertable_id | dimension_id | dimension_slice_id |     chunk_schema      |   chunk_table    |     range_start      |      range_end      
+----------+---------------+--------------+--------------------+-----------------------+------------------+----------------------+---------------------
+        5 |             6 |           14 |                  8 | _timescaledb_internal | _hyper_6_5_chunk |     1097366400000000 |    1097452800000000
+        5 |             6 |           15 |                  9 | _timescaledb_internal | _hyper_6_5_chunk | -9223372036854775808 |          1073741823
+        6 |             6 |           14 |                 10 | _timescaledb_internal | _hyper_6_6_chunk |     1098230400000000 |    1098316800000000
+        6 |             6 |           15 |                 11 | _timescaledb_internal | _hyper_6_6_chunk |           1073741823 | 9223372036854775807
+(4 rows)
+
+-- new dimension slice will cover full range on existing chunks
+SELECT add_dimension('dim_test', 'data', 1);
```
erimatnor, line 359: I'd like to see this test also for distributed hypertables. We'd like to see that `add_dimension` is also called on data nodes, and they get the dimension added to their local chunks.

erimatnor (commented):
Looks good, but I have questions about locking and concurrency.

I would like to see:

* Test on distributed hypertables to  ensure that the dimension is also added on data nodes
* Test concurrent add_dimension() with create chunk on same table 
* Test concurrent add_dimension() with other add_dimension() on same table

See inline comments for more details.

tsl/test/expected/dist_hypertable-14.out
```diff
@@ -2235,27 +2235,60 @@ NOTICE:  adding not-null constraint to column "time"
...
+       11 |             5 |            9 |                 10 | _timescaledb_internal | _dist_hyper_5_11_chunk | -9223372036854775808 | 9223372036854775807
+       11 |             5 |           10 |                 11 | _timescaledb_internal | _dist_hyper_5_11_chunk | -9223372036854775808 | 9223372036854775807
+(4 rows)
+
+SELECT * FROM dimented_table ORDER BY time;
+             time             | column1 |           column2            | column3 
+------------------------------+---------+------------------------------+---------
+ Sun Jan 01 06:01:00 2017 PST |       1 | Sun Jan 01 08:01:00 2017 PST |       1
+(1 row)
+
```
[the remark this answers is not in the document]
pmwkaa: We are not adding full range constraints as far as I understood

src/dimension.c
pmwkaa: Good point, will do

test/expected/create_hypertable.out
```diff
@@ -235,14 +235,169 @@ ERROR:  column "location" is already a dimension
...
+SELECT add_dimension('dim_test', 'device', 2);
+         add_dimension         
+-------------------------------
+ (13,public,dim_test,device,t)
+(1 row)
+
+SELECT * FROM test.show_constraints('_timescaledb_internal._hyper_5_2_chunk');
+  Constraint  | Type | Columns | Index |                                                                      Expr                                                                      | Deferrable | Deferred | Validated 
+--------------+------+---------+-------+------------------------------------------------------------------------------------------------------------------------------------------------+------------+----------+-----------
+ constraint_3 | c    | {time}  | -     | (("time" >= 'Sat Oct 09 17:00:00 2004 PDT'::timestamp with time zone) AND ("time" < 'Sun Oct 10 17:00:00 2004 PDT'::timestamp with time zone)) | f          | f        | t
```
erimatnor, line 288: There is no CHECK constraint for the `device` dimension. I would expect one to exist for each chunk after adding a new dimension, at least for new chunks. Should probably create a new chunk after add_dimension() to show that this is the case.

erimatnor (approved):

test/expected/create_hypertable.out
pmwkaa: This part of test does not show them for previous chunks, we did not add full range CHECK constraints before. This parts shows change in the `dimension_slice` table for old chunks.

What you are asking for is down below basically, when we add a new chunk after that (starting from line 306):

Well basically it shows them down below starting from here: https://github.com/timescale/timescaledb/pull/4495/files/63aed2332487bb4a07911438f169c72fb67eec4b#diff-7a9fb9a68eb195be632aa06cc899ef25fa305fb55643dfc9c2d0727624aa3a25R306

nikkhils (approved):
Have all Erik's comments been addressed? LGTM otherwise!

PR #4505, opened 2022-07-08
Enable assertions in appveyor build

akuzm:


akuzm:
It worked, but:

```
[00:07:49]  [3160]  LOG:  server process (PID 4980) was terminated by exception 0xC0000409
```

0xc0000409 means STATUS_STACK_BUFFER_OVERRUN. Not sure what's going on there...

PR #4519, opened 2022-07-18
Throw better error message on incompatible row fetcher settings

### LINKED ISSUE #4518: [Bug]: unexpected PQresult status 7 when starting COPY mode
### What type of bug is this?

Unexpected error

Multi-node

### What happened?

Got the following error in a join on distributed hypertable:
```
ERROR:  XX000: unexpected PQresult status 7 when starting COPY mode
LOCATION:  row_by_row_fetcher_send_fetch_request, row_by_row_fetcher.c:125
```

main

### PostgreSQL version used

14.4

Archlinux 2022.07.01 x64

Source

On prem/Self-hosted

```bash
ERROR:  XX000: unexpected PQresult status 7 when starting COPY mode
LOCATION:  row_by_row_fetcher_send_fetch_request, row_by_row_fetcher.c:125
```

```
select * from
  conditions_dist1 ref_0
where EXISTS (
  select from
            distinct_on_distributed as ref_1,
            lateral (select * from metrics as ref_2) as subq_3
    where
        (select device_id from metrics_compressed limit 1 offset 3) >= ref_0.device
);
```

svenklemm:
When a query has multiple distributed hypertables the row-by-by
fetcher cannot be used. This patch adjusts the fetcher type
selection to always use cursor fetcher when multiple distributed
hypertables are involved.

Fixes #4518

### DIFF
--- a/src/planner/planner.c
+++ b/src/planner/planner.c
@@ -515,20 +515,29 @@ timescaledb_planner(Query *parse, int cursor_opts, ParamListInfo bound_params)
 			{
 				reset_fetcher_type = true;
 
-				if (ts_guc_remote_data_fetcher == AutoFetcherType)
+				if (context.num_distributed_tables >= 2)
 				{
-					if (context.num_distributed_tables >= 2)
+					if (ts_guc_remote_data_fetcher == RowByRowFetcherType)
 					{
-						ts_data_node_fetcher_scan_type = CursorFetcherType;
+						ereport(ERROR,
+								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+								 errmsg("row-by-row fetcher not supported"),
+								 errhint("Row-by-row fetching of data is not supported in "
+										 "queries with multiple distributed hypertables."
+										 " Use cursor fetcher instead.")));
 					}
-					else
-					{
-						ts_data_node_fetcher_scan_type = RowByRowFetcherType;
-					}
+					ts_data_node_fetcher_scan_type = CursorFetcherType;
 				}
 				else
 				{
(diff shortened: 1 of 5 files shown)


svenklemm:
I'll add tests once the other row-by-row PR is merged #4516

src/planner/planner.c
```diff
@@ -482,21 +482,21 @@ timescaledb_planner(Query *parse, int cursor_opts, ParamListInfo bound_params)
 			{
 				reset_fetcher_type = true;
 
-				if (ts_guc_remote_data_fetcher == AutoFetcherType)
+				if (context.num_distributed_tables >= 2)
```
akuzm, line 485: The idea was for manual `ts_guc_remote_data_fetcher` to have preference over everything else. So if you set it manually to row-by-row when it can't work, it is expected to fail. I think this is more straightforward to use, i.e. it doesn't quietly ignore the explicit setting in some cases.

The default is `auto`, so why does it fail in your case, did you change it to row-by-row?

akuzm: The error message should be less cryptic though

svenklemm: Hmm yes this is due to the guc setting, similar planner gucs in postgres will fall back to working configuration to prevent running in a dead-end. e.g. disable_seqscan will still do seq scan if nothing else works. I think this behaviour is preferable over the current behaviour. If we encounter a situation that is known to not work we should not produce weird errors but instead fall back to working configuration.

svenklemm: Changed it to throw error instead

akuzm (approved):

RafiaSabih (approved):

PR #4525, opened 2022-07-19
Add config file and workflow for pull-review

konskov:
This commit adds a workflow for running pull-review in
the serverless mode, using docker container CLI.
Currently, it only does a dry run and logs the output.

### DIFF
--- a/.github/workflows/pull-review.yaml
+++ b/.github/workflows/pull-review.yaml
@@ -1,0 +1,19 @@
+name: Run pull-review with docker
+on: 
+  pull_request:
+jobs:
+  ask-review:
+    name: Run pull-review
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run shell script
+        env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            OWNER: ${{ github.repository_owner }}
+            REPO: ${{ github.event.repository.name }}
+            PULL_REQUEST_NUMBER : ${{ github.event.pull_request.number }}
+        shell: bash
+        run: |
+          PR_URL="https://github.com/${OWNER}/${REPO}/pull/${PULL_REQUEST_NUMBER}"
+          bash scripts/pull-review.sh "${GITHUB_TOKEN}" "${PR_URL}"
--- a/.pull-review
+++ b/.pull-review
@@ -1,0 +1,63 @@
+---
+# pull-review config version (required)
+version: 1
+
+# use review requests instead of assignees to assign reviewers to pull requests
+use_review_requests: true
+
+# maximum number of files to evaluate per pull request (set to 0 for no limit)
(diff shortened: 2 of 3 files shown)


.github/workflows/pull-review.yaml
```diff
@@ -0,0 +1,21 @@
...
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            OWNER: ${{ github.repository_owner }}
+            REPO: ${{ github.event.repository.name }}
+            PULL_REQUEST_NUMBER : ${{ github.event.pull_request.number }}
+        shell: bash
+        run: |
+          PR_URL="https://github.com/${OWNER}/${REPO}/pull/${PULL_REQUEST_NUMBER}"
+          bash scripts/pull-review.sh "${GITHUB_TOKEN}" "${PR_URL}"
+
+
```
mfundul, line 21: You can add the newline characters to the end of your files to stop github from complaining.

.pull-review
```diff
@@ -0,0 +1,63 @@
...
+use_review_requests: true
+
+# maximum number of files to evaluate per pull request (set to 0 for no limit)
+max_files: 0 # evaluate all files of the PR
+
+# minimum number of reviewers to assign and notify for a pull request
+min_reviewers: 2
+
+# maximum number of reviewers to assign and notify for a pull request
+max_reviewers: 3
```
mfundul, line 15: I would argue that 2 should be enough. I also am not sure if it assigns 2 extra reviewers, or if it is 2 total.

scripts/pull-review.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+#
+# This script runs pull-review on a pull request
+# to intelligently assign reviewers to it.
+#
+PULL_REVIEW_GITHUB_TOKEN=$1
+PULL_REQUEST_URL=$2
+
+docker pull ghcr.io/imsky/pull-review
```
mfundul, line 9: I am not sure this line is needed, I think `docker run` will pull the image regardless.

konskov: Of course, you are right. Removed it!

.pull-review
```diff
@@ -0,0 +1,63 @@
+---
+# pull-review config version (required)
+version: 1
```
jnidzwetzki, line 3: Does this version number have to be incremented each time the configuration is changed or does it refer to the version of the software?

jnidzwetzki (approved):

.pull-review
konskov: It refers to the version of pull-review. So we don't have to change it when modifying the configuration file.

.pull-review
konskov: Ok, 2 it is then! Pull-review will check how many reviewers are already assigned and take them into account, so that's 2 total, not extra.

mfundul (approved):

PR #4526, opened 2022-07-19
Fix gapfill group comparison for toasted values

svenklemm:
The gapfill mechanism to detect an aggregation group change was
using datumIsEqual to compare the group values. datumIsEqual does
not detoast values so when one value is toasted and the other value
is not it will not return the correct result. This patch changes
the gapfill code to use the correct equal operator for the type
of the group column instead of datumIsEqual.


akuzm:
Would be good to have a test for this.

gayyappan (approved):

PR #4534, opened 2022-07-20
Use explicit memory context with hash_create

svenklemm:
This changes the hash_create calls in the function cache and in the
cagg code to use explicit memory context. Without this hash_create
will create the hash table in TopMemoryContext potentially leading
to memory leaks.

### DIFF
--- a/src/func_cache.c
+++ b/src/func_cache.c
@@ -464,7 +464,10 @@ initialize_func_info()
 	Relation rel;
 	int i;
 
-	func_hash = hash_create("func_cache", _MAX_CACHE_FUNCTIONS, &hashctl, HASH_ELEM | HASH_BLOBS);
+	func_hash = hash_create("func_cache",
+							_MAX_CACHE_FUNCTIONS,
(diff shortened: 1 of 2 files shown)


jnidzwetzki (approved):
In general, LGTM. Two comments on the PR:

* In my opinion, this PR (along with #4532) should be mentioned in the changelog as a bug fix.

* I was wondering if it would be possible to write a _coccinelle_ test that automatically detects the missing memory contexts and `HASH_CONTEXT` flags?

mkindahl (approved):

svenklemm:
> In general, LGTM. Two comments on the PR:
> 
> * In my opinion, this PR (along with #4532) should be mentioned in the changelog as a bug fix.
> * I was wondering if it would be possible to write a _coccinelle_ test that automatically detects the missing memory contexts and `HASH_CONTEXT` flags?

They should definitely be in changelog, i'll add them in release PR #4535.
coccinelle script is outside of scope of this PR but would be useful to have

Issue #4486, opened 2022-07-01
[Bug]: Adding boolean column with default value doesn't work on compressed table


janko:
### What type of bug is this?

Unexpected error

### What subsystems and features are affected?

Compression

### What happened?

When I try to add a boolean column with a default value to an existing table that has compression enabled, the alter table command fails with an error from Timescale, saying it cannot add a column with "non-constant default expression" on a compressed hypertable.

When I look at the [Modify a schema](https://docs.timescale.com/timescaledb/latest/how-to-guides/compression/modify-a-schema/#add-a-nullable-column) page in documentation, it appeared this should be supported. I understand why a non-constant expression wouldn't work, but I used the default of `false`, which *is* a constant expression 🤔 

### TimescaleDB version affected

2.7.0

### PostgreSQL version used

14.4

### What operating system did you use?

Mac OS X 11.6 M1

### What installation method did you use?

Homebrew

### What platform did you run on?

On prem/Self-hosted

### Relevant log output and stack trace

```
ERROR:  cannot add column with non-constant default expression to a hypertable that has compression enabled
```

### How can we reproduce the bug?

```sql
CREATE TABLE records (time timestamp NOT NULL);
SELECT create_hypertable('records', 'time');
ALTER TABLE records SET (timescaledb.compress = true);
ALTER TABLE records ADD COLUMN col boolean DEFAULT false NOT NULL;
```

RafiaSabih:
https://github.com/timescale/timescaledb/pull/4537

PR #4537, opened 2022-07-21
Fix alter column for compressed table

RafiaSabih:
Enables adding a boolean column with default value to a compressed table.
This limitation was occurring due to the internal representation of default 
boolean values like 'True' or 'False', hence more checks are added for this.

Fixes https://github.com/timescale/timescaledb/issues/4486

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -273,6 +273,7 @@ check_altertable_add_column_for_compressed(Hypertable *ht, ColumnDef *col)
 		bool has_default = false;
 		bool has_notnull = col->is_not_null;
 		ListCell *lc;
+		bool is_bool = false;
 		foreach (lc, col->constraints)
 		{
 			Constraint *constraint = lfirst_node(Constraint, lc);
@@ -282,14 +283,30 @@ check_altertable_add_column_for_compressed(Hypertable *ht, ColumnDef *col)
 					has_notnull = true;
 					continue;
 				case CONSTR_DEFAULT:
-					/* since default expressions might trigger a table rewrite we
-					 * only allow Const here for now */
+					/*
+					 * Since default expressions might trigger a table rewrite we
+					 * only allow Const here for now.
+					 */
 					if (!IsA(constraint->raw_expr, A_Const))
-						ereport(ERROR,
-								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-								 errmsg("cannot add column with non-constant default expression "
-										"to a hypertable that has compression enabled")));
-
+					{
+						if (IsA(constraint->raw_expr, TypeCast) &&
+							IsA(castNode(TypeCast, constraint->raw_expr)->arg, A_Const))
+						{
+							/*
+							 * Ignore error only for boolean column, as values like
+							 * 'True' or 'False' are treated as TypeCast.
+							 */
+							char *name =
+								strVal(llast(((TypeCast *) constraint->raw_expr)->typeName->names));
+							is_bool = strstr(name, "bool") ? true : false;
+						}
+						if (!is_bool)
+							ereport(ERROR,
+									(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+									 errmsg(
+										 "cannot add column with non-constant default expression "
+										 "to a hypertable that has compression enabled")));
+					}
 					has_default = true;
 					continue;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ accidentally triggering the load of a previous DB version.**
 * #4393 Support intervals with day component when constifying now()
 * #4397 Support intervals with month component when constifying now()
 
+**Bugfixes**
+* #4486 Adding boolean column with default value doesn't work on compressed table
+
+**Thanks**
+@janko for reporting
+
 ## 2.7.2 (2022-07-26)
 
 This release is a patch release. We recommend that you upgrade at the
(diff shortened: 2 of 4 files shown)


svenklemm (approved):

src/process_utility.c
```diff
@@ -282,9 +283,20 @@ check_altertable_add_column_for_compressed(Hypertable *ht, ColumnDef *col)
...
+					 * only allow Const here for now.
+					 */
+					if (IsA(constraint->raw_expr, TypeCast))
+					{
+					   /*
+						* Ignore error for boolean column as values like
+						* 'True' or 'False' are treated as TypeCast.
+						*/
+						char *name = strVal(llast(((TypeCast *) constraint->raw_expr)->typeName->names));
+						is_bool = strcasestr(name, "bool") ? true : false;
```
akuzm, line 297: This feels too specific -- e.g. can I spell it as 'boolean'? What about casts for other types? etc

Maybe instead we could call `eval_const_expressions` before performing the old check?

svenklemm: This is actually added with that exact spelling by parser. If you look at the tests you can see that the column definition actually uses a different spelling for the column type. This is special casing because the parser treats `'f'` different from `false` or `true` so if you do `ADD COLUMN foo DEFAULT false` it will fail the Const check but `ADD COLUMN foo DEFAULT 'f'` it will be represented as Const. This is specific to bool handling and all the other datatypes should be covered by the Const-check.

RafiaSabih: Well it doesn't matter what is given in the query, because this value is always bool. TypeName->names is a list of schema and type which is 'pg_catalog' and 'bool' respectively.
Also, so far we only know about this case where we want to allow.

mfundul:
I don't quite understand the commit description. Can you rephrase it somehow?

src/process_utility.c
```diff
@@ -282,9 +283,20 @@ check_altertable_add_column_for_compressed(Hypertable *ht, ColumnDef *col)
 					has_notnull = true;
 					continue;
 				case CONSTR_DEFAULT:
-					/* since default expressions might trigger a table rewrite we
-					 * only allow Const here for now */
-					if (!IsA(constraint->raw_expr, A_Const))
+					/*
+					 * Since default expressions might trigger a table rewrite we
+					 * only allow Const here for now.
```
mfundul, line 288: is this comment about `Const` valid still?

src/process_utility.c
```diff
@@ -282,9 +283,20 @@ check_altertable_add_column_for_compressed(Hypertable *ht, ColumnDef *col)
...
+					if (IsA(constraint->raw_expr, TypeCast))
+					{
+					   /*
+						* Ignore error for boolean column as values like
+						* 'True' or 'False' are treated as TypeCast.
+						*/
+						char *name = strVal(llast(((TypeCast *) constraint->raw_expr)->typeName->names));
+						is_bool = strcasestr(name, "bool") ? true : false;
+					}
+					else if (!is_bool && (!IsA(constraint->raw_expr, A_Const)))
```
mfundul, line 299:
```suggestion
					else if (!is_bool && (!IsA(constraint->raw_expr, TypeCast)))
```
maybe? Not sure.

mfundul, line 299: Isn't `is_bool` always `false` since it's an `else if` so `is_bool` is never set?

src/process_utility.c
svenklemm: To make it bulletproof we should also check that the typecast-arg is a Const, we dont want function calls or anything else here.

src/process_utility.c
svenklemm: yes looks like the else shouldnt be there

src/process_utility.c
svenklemm: While using `eval_const_expressions` would be the most generic solution it cannot be used here because it only works on planner nodes and the nodes we are dealing with here are parser nodes.

src/process_utility.c
```diff
@@ -282,14 +283,29 @@ check_altertable_add_column_for_compressed(Hypertable *ht, ColumnDef *col)
...
+					 */
 					if (!IsA(constraint->raw_expr, A_Const))
-						ereport(ERROR,
-								(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-								 errmsg("cannot add column with non-constant default expression "
-										"to a hypertable that has compression enabled")));
-
+					{
+						if (IsA(constraint->raw_expr, TypeCast) &&
+							IsA(((TypeCast *) constraint->raw_expr)->arg, Const))
```
svenklemm, line 293:
```suggestion
							IsA((castNode(TypeCast,constraint)->raw_expr)->arg, A_Const))
```

src/process_utility.c
```diff
@@ -282,14 +283,29 @@ check_altertable_add_column_for_compressed(Hypertable *ht, ColumnDef *col)
...
+						{
+						   /*
+							* Ignore error only for boolean column, as values like
+							* 'True' or 'False' are treated as TypeCast.
+							*/
+							char *name =
+									strVal(llast(((TypeCast *) constraint->raw_expr)->typeName->names));
```
svenklemm, line 300:
```suggestion
									strVal(llast((castNode(TypeCast,constraint)->raw_expr)->typeName->names));
```

src/process_utility.c
svenklemm: yes

mfundul (approved):

PR #4545, opened 2022-07-25
Add hypertable distributed argument and defaults 

pmwkaa:
This PR introduces a new `distributed` argument to the
create_hypertable() function as well as two new GUC's to
control its default behaviour: timescaledb.hypertable_distributed_default
and timescaledb.hypertable_replication_factor_default.

The main idea of this change is to allow automatic creation
of the distributed hypertables by default.

### DIFF
--- a/sql/ddl_api.sql
+++ b/sql/ddl_api.sql
@@ -20,8 +20,9 @@
 -- chunk_target_size - (Optional) The target size for chunks (e.g., '1000MB', 'estimate', or 'off')
 -- chunk_sizing_func - (Optional) A function to calculate the chunk time interval for new chunks
 -- time_partitioning_func - (Optional) The partitioning function to use for "time" partitioning
--- replication_factor - (Optional) A value of 1 or greater makes this hypertable distributed
+-- replication_factor - (Optional) Set replication_factor to use with the new hypertable
 -- data_nodes - (Optional) The specific data nodes to distribute this hypertable across
+-- distributed - (Optional) Create distributed hypertable
 CREATE OR REPLACE FUNCTION @extschema@.create_hypertable(
     relation                REGCLASS,
     time_column_name        NAME,
@@ -38,10 +39,10 @@ CREATE OR REPLACE FUNCTION @extschema@.create_hypertable(
     chunk_sizing_func       REGPROC = '_timescaledb_internal.calculate_chunk_interval'::regproc,
     time_partitioning_func  REGPROC = NULL,
     replication_factor      INTEGER = NULL,
-    data_nodes              NAME[] = NULL
+    data_nodes              NAME[] = NULL,
+    distributed             BOOLEAN = NULL
 ) RETURNS TABLE(hypertable_id INT, schema_name NAME, table_name NAME, created BOOL) AS '@MODULE_PATHNAME@', 'ts_hypertable_create' LANGUAGE C VOLATILE;
 
--- Same functionality as create_hypertable, only must have a replication factor > 0 (defaults to 1)
 CREATE OR REPLACE FUNCTION @extschema@.create_distributed_hypertable(
     relation                REGCLASS,
     time_column_name        NAME,
@@ -57,7 +58,7 @@ CREATE OR REPLACE FUNCTION @extschema@.create_distributed_hypertable(
     chunk_target_size       TEXT = NULL,
     chunk_sizing_func       REGPROC = '_timescaledb_internal.calculate_chunk_interval'::regproc,
     time_partitioning_func  REGPROC = NULL,
-    replication_factor      INTEGER = 1,
+    replication_factor      INTEGER = NULL,
     data_nodes              NAME[] = NULL
 ) RETURNS TABLE(hypertable_id INT, schema_name NAME, table_name NAME, created BOOL) AS '@MODULE_PATHNAME@', 'ts_hypertable_distributed_create' LANGUAGE C VOLATILE;
 
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -105,3 +105,47 @@ SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_agg_
 GRANT SELECT ON _timescaledb_catalog.continuous_agg_migrate_plan TO PUBLIC;
 GRANT SELECT ON _timescaledb_catalog.continuous_agg_migrate_plan_step TO PUBLIC;
 GRANT SELECT ON _timescaledb_catalog.continuous_agg_migrate_plan_step_step_id_seq TO PUBLIC;
+
+-- add distributed argument
+DROP FUNCTION IF EXISTS @extschema@.create_hypertable;
+DROP FUNCTION IF EXISTS @extschema@.create_distributed_hypertable;
+
+CREATE FUNCTION @extschema@.create_hypertable(
+    relation                REGCLASS,
+    time_column_name        NAME,
+    partitioning_column     NAME = NULL,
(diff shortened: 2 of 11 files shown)


fabriziomello (approved):

src/guc.c
```diff
@@ -486,6 +487,24 @@ _guc_init(void)
 							   /* assign_hook= */ NULL,
 							   /* show_hook= */ NULL);
 #endif
+
+	DefineCustomIntVariable("timescaledb.default_replication_factor",
```
erimatnor, line 491: I suggest changing to the following to make clear that this is a hypertable setting.

```suggestion
	DefineCustomIntVariable("timescaledb.hypertable_replication_factor_default",
```

Then we'd also have a similar GUC, as discussed:

```sql
timescaledb.hypertable_distributed_default
```

src/guc.c
```diff
@@ -486,6 +487,24 @@ _guc_init(void)
...
+							"Default replication factor value for distributed database",
```
erimatnor, line 492:
```suggestion
							"Default replication factor value for distributed hypertables",
```

src/guc.c
```diff
@@ -486,6 +487,24 @@ _guc_init(void)
...
+
+							"When acting as a access node, automatically create distributed "
+							"hypertables"
+							"instead of regular hypertables by setting default_replication_factor"
```
erimatnor, line 496:
```suggestion
							"instead of regular hypertables by setting timescaledb.hypertable_replication_factor_default"
```

erimatnor (commented):
Just added some comments. I guess this PR is being updated.

sql/ddl_api.sql
```diff
@@ -57,8 +59,9 @@ CREATE OR REPLACE FUNCTION @extschema@.create_distributed_hypertable(
     chunk_target_size       TEXT = NULL,
     chunk_sizing_func       REGPROC = '_timescaledb_internal.calculate_chunk_interval'::regproc,
     time_partitioning_func  REGPROC = NULL,
-    replication_factor      INTEGER = 1,
-    data_nodes              NAME[] = NULL
+    replication_factor      INTEGER = NULL,
+    data_nodes              NAME[] = NULL,
+    distributed             BOOLEAN = true
```
erimatnor, line 64: I think we should elide this argument for `create_distributed_hypertable`. It doesn't make sense to have it here IMO. It is always "true" and cannot be changed.

```suggestion
   
```

sql/updates/latest-dev.sql
```diff
@@ -70,3 +70,48 @@ WHERE proc_schema = '_timescaledb_internal'
...
+    create_default_indexes  BOOLEAN = TRUE,
+    if_not_exists           BOOLEAN = FALSE,
+    partitioning_func       REGPROC = NULL,
+    migrate_data            BOOLEAN = FALSE,
+    chunk_target_size       TEXT = NULL,
+    chunk_sizing_func       REGPROC = '_timescaledb_internal.calculate_chunk_interval'::regproc,
+    time_partitioning_func  REGPROC = NULL,
+    replication_factor      INTEGER = NULL,
+    data_nodes              NAME[] = NULL,
+    distributed             BOOLEAN = true
```
erimatnor, line 116:
```suggestion
```

src/guc.c
```diff
@@ -55,6 +55,13 @@ static const struct config_enum_entry remote_data_fetchers[] = {
 	{ NULL, 0, false }
 };
 
+static const struct config_enum_entry hypertable_distributed_types[] = {
+	{ "auto", HYPERTABLE_DIST_AUTO, false },
```
erimatnor, line 59: What does "auto" do? And how is it different from NULL?

src/hypertable.c
```diff
@@ -1712,11 +1686,109 @@ ts_validate_replication_factor(int32 replication_factor, bool is_null, bool is_d
...
+				distributed = true;
+				break;
+		}
+	}
+
+	if (distributed_local_error)
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("operation not supported"),
+				 errdetail("Local hypertables cannot set replication_factor.")));
```
erimatnor, line 1765: Just use the main message, it is more informative.
```suggestion
				 errmsg("local hypertables cannot set a replication_factor")));
```

tsl/test/expected/dist_ddl.out
```diff
@@ -2533,6 +2533,246 @@ NOTICE:  adding not-null constraint to column "time"
...
+SET timescaledb.hypertable_distributed_default TO 'auto';
+-- create regular hypertable by default
+CREATE TABLE drf_test(time TIMESTAMPTZ NOT NULL);
+SELECT create_hypertable('drf_test', 'time', distributed=>false);
+   create_hypertable    
+------------------------
+ (22,public,drf_test,t)
+(1 row)
+
+SELECT h.table_name, h.replication_factor FROM _timescaledb_catalog.hypertable h WHERE h.table_name = 'drf_test';
```
erimatnor, line 2582: Maybe use the `timescaledb_information.hypertables` view for this where there's a "distributed" bool column.

src/guc.c
pmwkaa: Is there a NULL for guc value? I am not aware of it

sql/ddl_api.sql
pmwkaa: I can do it , but I was thinking to just make this function an exact alias to the `create_hypertable()`, otherwise internally we will have to rely on the `dist_call` argument to see if the function was actually been called when we check for the call arguments. Also there would be gaps in argument enumeration, when a new argument will be added after the `distributed` in the future

svenklemm (approved):

src/guc.c
pmwkaa: I think this can be possible only for session variables

erimatnor: Sorry, I wasn't clear. I am wondering what "auto" does. Is it similar to what we'd expect NULL (unset) to be?

sql/ddl_api.sql
erimatnor: I understand the desire to make the signatures the same, but IMO, it doesn't make sense to have this particular parameter for the "distributed" function. So, I don't think we should add parameters that don't make sense just to keep the signatures the same.

src/guc.c
```diff
@@ -55,6 +55,13 @@ static const struct config_enum_entry remote_data_fetchers[] = {
...
+	{ "local", HYPERTABLE_DIST_LOCAL, false },
+	{ "distributed", HYPERTABLE_DIST_DISTRIBUTED, false },
```
erimatnor, line 61: I am wondering if these should be "true" and "false" since this GUC is the default for the "distributed" boolean parameter. I think "true" and "false" also makes more sense given the GUC name, e.g., `hypertable_distributed_default=true;` rather than setting `hypertable_distributed_default=local` seems a bit counter intuitive. 


```suggestion
	{ "false", HYPERTABLE_DIST_LOCAL, false },
	{ "true", HYPERTABLE_DIST_DISTRIBUTED, false },
```

src/guc.c
pmwkaa: Yes, this is current behaviour. We can name it to something else

src/guc.c
pmwkaa: This can't be boolean and having booleanish values as strings is confusing when there is an actual boolean type exists.

Boolean value is either true of false (there are no notion of undefined in gucs), but having false by default will break compatability and will force `create_hypertable(replication_factor)` to emit error.

PR #4548, opened 2022-07-26
Add a generic memory leak test to CI

svenklemm:
Unlike the existing memory leak test which tests a single INSERT
of 100.000.000 rows in a single transaction, this runs 1000000
SELECT, 10K UPDATE and 10K INSERT queries in the same backend.

### DIFF
--- a/scripts/test_memory_spikes.py
+++ b/scripts/test_memory_spikes.py
@@ -4,8 +4,7 @@
 #  Please see the included NOTICE for copyright information and
 #  LICENSE-APACHE for a copy of the license.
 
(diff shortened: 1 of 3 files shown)


scripts/test_memory_spikes.py
```diff
@@ -13,7 +13,7 @@
```
jnidzwetzki, line 13: The comments in lines 8, 98, and 141 still refer to _random inserts_. It might be good to update them and add a more general description of the workload.

jnidzwetzki (approved):

konskov (approved):

PR #4556, opened 2022-07-29
Add runtime exclusion for ANY using subselect

cevian:
Add runtime exclusion for statements such as :

   WHERE col @> ANY(subselect)

In this case, if the subselect returns no rows, the
result is always false. This is independent of whether
or not the partitioning columns is referenced.

Also fix a bug (or missed optimization) if the subselect
hasn't been executed by the time the runtime exclusion
executes.

### DIFF
--- a/src/nodes/chunk_append/chunk_append.c
+++ b/src/nodes/chunk_append/chunk_append.c
@@ -114,10 +114,25 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
 		{
 			ListCell *lc_var;
 
-			/*
-			 * check the param references a partitioning column of the hypertable
-			 * otherwise we skip runtime exclusion
+			/* We have two types of exclusion:
+			 *
+			 * Parent exclusion fires if the entire hypertable can be excluded.
+			 * This happens if doing things like joining against a parameter
+			 * value that is an empty array or NULL. It doesn't happen often,
+			 * but when it does, it speeds up the query immensely. It's also cheap
+			 * to check for this condition as you check this once per hypertable
+			 * at runtime.
+			 *
+			 * Child exclusion works by seeing if there is a contradiction between
+			 * the chunks constraints and the expression on parameter values. For example,
+			 * it can evaluate whether a time parameter from a subquery falls outside
+			 * the range of the chunk. It is more widely applicable than the parent
+			 * exclusion but is also more expensive to evaluate since you have to perform
+			 * the check on every chunk. Child exclusion can only apply if one of the quals
+			 * involves a partitioning column.
+			 *
 			 */
+			path->runtime_exclusion_parent = true;
 			foreach (lc_var, pull_var_clause((Node *) rinfo->clause, 0))
 			{
 				Var *var = lfirst(lc_var);
@@ -130,12 +145,22 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
 				if (var->varno == rel->relid && var->varattno > 0 &&
 					ts_is_partitioning_column(ht, var->varattno))
 				{
-					path->runtime_exclusion = true;
+					path->runtime_exclusion_children = true;
 					break;
 				}
 			}
 		}
 	}
+	/*
+	 * Our strategy is to use child exclusion if possible (if a partitioning
+	 * column is used) and fall back to parent exclusion if we can't use child
+	 * exclusion. Please note: there is no point to using both child and parent
+	 * exclusion at the same time since child exclusion would always exclude
+	 * the same chunks that parent exclusion would.
+	 */
+
+	if (path->runtime_exclusion_parent && path->runtime_exclusion_children)
+		path->runtime_exclusion_parent = false;
 
 	/*
 	 * Make sure our subpath is either an Append or MergeAppend node
@@ -253,7 +278,8 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht, 
 		if (!has_scan_childs)
 		{
 			path->startup_exclusion = false;
-			path->runtime_exclusion = false;
+			path->runtime_exclusion_parent = false;
+			path->runtime_exclusion_children = false;
 		}
 
 		path->cpath.custom_paths = nested_children;
--- a/src/nodes/chunk_append/chunk_append.h
+++ b/src/nodes/chunk_append/chunk_append.h
@@ -15,7 +15,8 @@ typedef struct ChunkAppendPath
 {
 	CustomPath cpath;
 	bool startup_exclusion;
-	bool runtime_exclusion;
+	bool runtime_exclusion_parent;
+	bool runtime_exclusion_children;
 	bool pushdown_limit;
 	int limit_tuples;
 	int first_partial_path;
--- a/src/nodes/chunk_append/exec.c
+++ b/src/nodes/chunk_append/exec.c
@@ -52,7 +52,8 @@ typedef struct ChunkAppendState
 
 	Oid ht_reloid;
 	bool startup_exclusion;
-	bool runtime_exclusion;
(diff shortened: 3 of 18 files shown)


cevian:
I'll add tests if we agree to go forward with this approach.

svenklemm:
I think you should not tie it to scalar expressions as this could potentially be useful in other situations too. Do you know if any our current tests hit that path?

src/nodes/chunk_append/exec.c
```diff
@@ -345,49 +372,67 @@ initialize_runtime_exclusion(ChunkAppendState *state)
...
+		 * inference
+		 */
+		if (can_exclude_constraints_using_clauses(state,
+												  list_make1(makeBoolConst(true, false)),
+												  state->initial_parent_clauses,
+												  root,
+												  &state->csstate.ss.ps))
+		{
+			state->runtime_initialized = true;
+			state->runtime_number_exclusions = state->num_subplans;
```
svenklemm, line 397:
```suggestion
			state->runtime_number_exclusions += state->num_subplans;
```

svenklemm:
I think it would be useful to track hypertable vs chunk runtime exclusion separately.

src/nodes/chunk_append/exec.c
```diff
@@ -786,6 +831,8 @@ constify_param_mutator(Node *node, void *context)
 			{
 				ExprContext *econtext = GetPerTupleExprContext(estate);
 				ExecSetParamPlan(prm.execPlan, econtext);
+				// reload prm as it may have changed
```
akuzm, line 834: When does it change? That would be more than one loops for chunk append plan, so I guess when it's on the inner side of a nested loop join?

src/nodes/chunk_append/exec.c
```diff
@@ -345,49 +372,67 @@ initialize_runtime_exclusion(ChunkAppendState *state)
...
+
+	if (state->initial_parent_clauses != NIL)
+	{
+		/* try to exclude all the chunks using the parents clauses.
+		 * here, all constraints are true but exclusion can still
+		 * happen because of things like ANY(empty set), and NULL
+		 * inference
+		 */
+		if (can_exclude_constraints_using_clauses(state,
+												  list_make1(makeBoolConst(true, false)),
```
akuzm, line 391: So here we're passing it a list of constraints that consists of a single true value, to effectively make it check whether the `initial_parent_clauses` evaluate to true, given the plan parameter?

src/nodes/chunk_append/planner.c
```diff
@@ -272,10 +273,21 @@ ts_chunk_append_plan_create(PlannerInfo *root, RelOptInfo *rel, CustomPath *path
 				chunk_rt_indexes = lappend_oid(chunk_rt_indexes, scan->scanrelid);
 			}
 		}
+
 		Assert(list_length(cscan->custom_plans) == list_length(chunk_ri_clauses));
 		Assert(list_length(chunk_ri_clauses) == list_length(chunk_rt_indexes));
 	}
 
+	/* only do parent exclusion if not doing chunk-by-chunk exclusion */
```
akuzm, line 281: It would be good to have a comment describing what the parent exclusion is, as opposed to chunk-by-chunk exclusion.

src/nodes/chunk_append/chunk_append.h
```diff
@@ -16,6 +16,7 @@ typedef struct ChunkAppendPath
 	CustomPath cpath;
 	bool startup_exclusion;
 	bool runtime_exclusion;
```
akuzm, line 18: So this variable is now a flag for "parent exclusion", and it used to be "chunk-by-chunk exclusion", which is now `runtime_exclude_chunk_constraints`?

Let's add comments about this, and rename the old variable to be specific as well, e.g. `runtime_exclude_by_parent_clauses`.  Otherwise it's easy to miss some usages that relied on the old meaning.

src/nodes/chunk_append/exec.c
cevian: Its changed by the ExecSetParamPlan call above. I updated the column to be more clear.

src/nodes/chunk_append/exec.c
cevian: Yes, except that we are checking whether initial_parent_clauses are always false (not true) for exclusion. But yes this is simply checking the parent clauses for internal contradiction given the evaluated parameter values

src/nodes/chunk_append/planner.c
cevian: Put a big comment in chunk_append.c to explain things

src/nodes/chunk_append/chunk_append.h
cevian: 👍 refactored stuff to hopefully be more clear

test/expected/append-14.out
```diff
@@ -2260,6 +2266,123 @@ ANALYZE i3030;
...
+--parent runtime exclusion tests:
+--optimization works with ANY (array)
+:PREFIX
+SELECT *
+FROM append_test a
+WHERE a.attr @> ANY((SELECT coalesce(array_agg(attr), array[]::jsonb[]) FROM join_test_plain WHERE temp > 100)::jsonb[]);
+                             QUERY PLAN                             
+--------------------------------------------------------------------
+ Custom Scan (ChunkAppend) on append_test a (actual rows=0 loops=1)
+   Runtime excluded entire hypertable: true
```
svenklemm, line 2278: This should be a counter since it is a per loop decission. Also not a fan of the boolean in EXPLAIN output i would prefer two distinct messages for runtime exclusion variant. Exclusion information should only be visible with EXPLAIN (verbose).
eg:
Chunks excluded during runtime: 123
and
Hypertables excluded during runtime: 4

src/nodes/chunk_append/exec.c
akuzm: Does this mean that parent clauses should evaluate to constant false after `constify_restrictinfo_params`? I wonder if we can wrirte the check explicitly like this, `constify_restrictinfo_params` and then check that every element is a false Const.

I mean, I just spent half an hour at it, and I still can't tell what `predicate_refuted_by` does if you pass clause = False, predicate = Null, and weak = true. And this is a fallback used when there are more than one predicate, which apparently isn't covered by the new tests.

akuzm (approved):

src/nodes/chunk_append/chunk_append.c
```diff
@@ -114,10 +114,25 @@ ts_chunk_append_path_create(PlannerInfo *root, RelOptInfo *rel, Hypertable *ht,
...
+			 * but when it does, it speeds up the query immensely. It's also cheap
+			 * to check for this condition as you check this once per hypertable
+			 * at runtime.
+			 *
+			 * Child exclusion works by seeing if there is a contradiction between
+			 * the chunks constraints and the expression on parameter values. For example,
+			 * it can evaluate whether a time parameter from a subquery falls outside
+			 * the range of the chunk. It is more widely applicable than the parent
+			 * exclusion but is also more expensive to evaluate since you have to perform
+			 * the check on every chunk. Child exclusion can only apply of one of the quals
```
svenklemm, line 131:
```suggestion
			 * the check on every chunk. Child exclusion can only apply if one of the quals
```

tsl/test/expected/dist_api_calls.out
```diff
@@ -380,13 +380,4 @@ SELECT approximate_row_count('disttable_repl');
...
- approximate_row_count 
------------------------
-                     8
-(1 row)
-
-DROP TABLE disttable_repl;
-DROP DATABASE :DN_DBNAME_1;
-DROP DATABASE :DN_DBNAME_2;
-DROP DATABASE :DN_DBNAME_3;
+ERROR:  [data_node_1]: relation "disttable_repl" does not exist
```
svenklemm, line 383: Hmm that doesnt look right, and shouldnt be part of your PR. I think this is just result of flaky multinode test you accidentally added.

svenklemm (approved):

PR #4567, opened 2022-08-02
Exponentially backoff when out of background workers

### LINKED ISSUE #4562: [Enhancement]: Job scheduler - exponential backoffs

User experience

Background worker

When a job fails, the next attempt is scheduled with an exponential backoff.
For instance, the first attempt is executed after 10 second, the next attempts are executed after 30 seconds, 2 minutes, 10 minutes...

### Implementation challenges

_No response_

konskov:
The scheduler detects the following three types of job failures:

1.Jobs that fail to launch (due to shortage of background workers)
2.Jobs that throw a runtime error
3.Jobs that crash due to a process crashing

In cases 2 and 3, additive backoff is applied in calculating the next
start time of a failed job.
In case 1 we previously retried to launch all jobs that failed to launch
simultaneously.

This commit introduces exponential backoff in case 1,
randomly selecting a wait time in [2, 2 + 2^f] seconds at microsecond granularity.
The aim is to reduce the collision probability for jobs that compete
for a background worker. The maximum backoff value is 1 minute.
It does not change the behavior for cases 2 and 3.

Fixes #4562

### DIFF
--- a/src/bgw/job_stat.c
+++ b/src/bgw/job_stat.c
@@ -184,9 +184,10 @@ calculate_jitter_percent()
 	return ldexp((double) (16 - (int) (percent % 32)), -7);
 }
 
-/* For failures we have additive backoff based on consecutive failures
- * along with a ceiling at schedule_interval * MAX_INTERVALS_BACKOFF
- * We also limit the additive backoff in case of consecutive failures as we don't
+/* For failures we have backoff based on consecutive failures
+ * along with a ceiling at schedule_interval * MAX_INTERVALS_BACKOFF / 1 minute
+ * for jobs failing at runtime / for jobs failing to launch.
+ * We also limit the backoff in case of consecutive failures as we don't
  * want to pass in input that leads to out of range timestamps and don't want to
  * put off the next start time for the job indefinitely
  */
@@ -198,9 +199,14 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
 	TimestampTz res = 0;
 	volatile bool res_set = false;
 	TimestampTz last_finish = finish_time;
+	bool launch_failure = (job == NULL);
 	float8 multiplier = (consecutive_failures > MAX_FAILURES_MULTIPLIER ? MAX_FAILURES_MULTIPLIER :
 																		  consecutive_failures);
 	MemoryContext oldctx;
+	Assert(consecutive_failures > 0);
+	int64 max_slots =
+		(1 << consecutive_failures) - 1; /* 2^(consecutive_failures) - 1, at most 2^20 */
+	int64 rand_backoff = random() % (max_slots * USECS_PER_SEC);
 
 	if (!IS_VALID_TIMESTAMP(finish_time))
 	{
@@ -211,15 +217,31 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
 	BeginInternalSubTransaction("next start on failure");
 	PG_TRY();
 	{
-		/* ival = retry_period * (consecutive_failures)  */
-		Datum ival = DirectFunctionCall2(interval_mul,
-										 IntervalPGetDatum(&job->fd.retry_period),
-										 Float8GetDatum(multiplier));
-
+		Datum ival;
 		/* ival_max is the ceiling = MAX_INTERVALS_BACKOFF * schedule_interval */
-		Datum ival_max = DirectFunctionCall2(interval_mul,
-											 IntervalPGetDatum(&job->fd.schedule_interval),
-											 Float8GetDatum(MAX_INTERVALS_BACKOFF));
+		Datum ival_max;
+		// max wait time to launch job is 1 minute
+		Interval interval_max = { .time = 60000000 };
+		Interval retry_ival = { .time = 2000000 };
+		retry_ival.time += rand_backoff;
+
+		if (launch_failure)
+		{
+			// random backoff seconds in [2, 2 + 2^f]
+			ival = IntervalPGetDatum(&retry_ival);
+			ival_max = IntervalPGetDatum(&interval_max);
+		}
+		else
+		{
+			/* ival = retry_period * (consecutive_failures)  */
+			ival = DirectFunctionCall2(interval_mul,
+									   IntervalPGetDatum(&job->fd.retry_period),
+									   Float8GetDatum(multiplier));
+			/* ival_max is the ceiling = MAX_INTERVALS_BACKOFF * schedule_interval */
+			ival_max = DirectFunctionCall2(interval_mul,
+										   IntervalPGetDatum(&job->fd.schedule_interval),
+										   Float8GetDatum(MAX_INTERVALS_BACKOFF));
+		}
 
 		if (DatumGetInt32(DirectFunctionCall2(interval_cmp, ival, ival_max)) > 0)
 			ival = ival_max;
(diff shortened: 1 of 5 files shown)


mkindahl (commented):
In general, this looks good, but it would be good if you could add an outline in the commit message where you describe how:

- Jobs that fail while executing are re-tried
- Jobs that fail to start the background worker are re-tried (IIRC, this is different from the one above)
- Jobs that succeed are re-tried

gayyappan (commented):
IIRC we had exponential backoff initially and then switched to additive backoff. The problem is with jobs that have fairly longer schedules. Say the job is scheduled to run once a day. With exponential backoff, you will quickly schedule a failed job too far into the future. A better way might be to add an upper bound for the next start time after a failure.
i.e. if job failed:
```
      new_start_time = last_failure_time + schedule interval (with some back off)
      ub_start_time = last_failue_time + '1 h'
      if  ub_start_time < new_start_time
          new_start_time = worst_case.

```

Also make sure to test and catch potential overflow with exponential backoffs.

konskov:
Thank you so much for reviewing! Currently there is an upper bound but it is probably too high (10 times the schedule interval, and for jobs that fail to launch due to background worker shortage, the upper bound is 1 minute). I will change it to something fixed and smaller, '1h' as you suggest.

svenklemm (approved):

fabriziomello (approved):

src/bgw/job_stat.c
```diff
@@ -211,15 +212,37 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
...
+		{
+			// retry every 2 seconds
+			ival = IntervalPGetDatum(&retry_ival);
+		}
+
+		/* ival = retry_period ^ (consecutive_failures - 1)  */
+		/* arbitrarily choose 2 for exponential growth */
+		for (i = 0; i < exponent - 1; i++)
+		{
+			ival = DirectFunctionCall2(interval_mul, ival, Float8GetDatum(2));
```
mfundul, line 233: Exponential backoff works when the scheduled time is picked randomly between `now()` and `now() + ival` (regardless of jitter). If we always pick `now() + ival` then we risk livelocking the conflicting jobs ad infinitum.

This is orthogonal to jitter, which helps with scheduling at non-integer multiples of time units, so as to not generate high load spikes in the system (e.g. due to everything happening at integer seconds instead of spreading them out).

mfundul (changes requested):

src/bgw/job_stat.c
konskov: You are right, we are risking that. So for first case (out of background workers) I will change the backoff calculation to  randomly pick a backoff value between [0, 2^f - 1] where f is the number of consecutive failed launches of the job.

mfundul (changes requested):

src/bgw/job_stat.c
```diff
@@ -198,8 +199,19 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
...
+	 * the aim of backoff is to reduce conflicts in the case of jobs that fail to launch,
+	 * as there are jobs competing for available background workers in that case.
+	 * therefore we introduce some randomness in backoff calculation for that type of failure
+	 * by randomly selecting the exponent in the interval [1, consecutive_failures]
+	 * Otherwise as soon as a worker becomes available, all jobs will try to grab it, but only one
+	 * will anyway so why should they all try at once? so jobs that experience failures will delay
+	 * their next_start a random amount to reduce the probability of collisions
+	 */
+	Assert(consecutive_failures > 0);
+	int rand_exp = random() % consecutive_failures + 1;
```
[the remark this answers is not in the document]
konskov: Thank you for correcting me 😓 I changed that to select a random backoff time in the range at a microsec granularity.

src/bgw/job_stat.c
```diff
@@ -198,8 +199,25 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
...
+	 * their next_start a random amount to reduce the probability of collisions
+	 */
+	Assert(consecutive_failures > 0);
+	// 2^(consecutive_failures) - 1, at most 2^20
+	int64 max_slots = (1 << consecutive_failures) - 1;
+	// granularity: 1 usec (2^20)
+	int64 micros_per_minute = 1000000;
+	// will get a random int in [0, (2^f - 1) * 1000]
+	// this represents a random amount of microseconds to backoff
+	int64 rand_backoff = random() % (max_slots * micros_per_minute);
```
mfundul, line 220:
```suggestion
	int64 rand_backoff = random() % (max_slots * USECS_PER_SEC);
```
I don't see anything to do with minutes here.

src/bgw/job_stat.c
```diff
@@ -198,8 +199,25 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
...
+	 * Otherwise as soon as a worker becomes available, all jobs will try to grab it, but only one
+	 * will anyway so why should they all try at once? so jobs that experience failures will delay
+	 * their next_start a random amount to reduce the probability of collisions
+	 */
+	Assert(consecutive_failures > 0);
+	// 2^(consecutive_failures) - 1, at most 2^20
+	int64 max_slots = (1 << consecutive_failures) - 1;
+	// granularity: 1 usec (2^20)
+	int64 micros_per_minute = 1000000;
+	// will get a random int in [0, (2^f - 1) * 1000]
```
mfundul, line 218: Is this comment up-to-date?

src/bgw/job_stat.c
konskov: Indeed

src/bgw/job_stat.c
```diff
@@ -211,15 +228,40 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
...
-											 Float8GetDatum(MAX_INTERVALS_BACKOFF));
+		Datum ival_max;
+		if (job)
+			ival_max = DirectFunctionCall2(interval_mul,
+										   IntervalPGetDatum(&job->fd.schedule_interval),
+										   Float8GetDatum(MAX_INTERVALS_BACKOFF));
+		else
+		{
+			ival_max = IntervalPGetDatum(&interval_max);
+		}
```
mfundul, line 264: you do:
```
		if (job)
		{
		}
		else
		{
		}
```
3 to 4 times here and below. Maybe you should create a meaningful boolean variable for the `if (job)` case, and you should probably only do 1 `if-else` instead of 3 consecutive times to improve readability.

src/bgw/job_stat.c
```diff
@@ -211,15 +229,32 @@ calculate_next_start_on_failure(TimestampTz finish_time, int consecutive_failure
...
 		/* ival_max is the ceiling = MAX_INTERVALS_BACKOFF * schedule_interval */
-		Datum ival_max = DirectFunctionCall2(interval_mul,
-											 IntervalPGetDatum(&job->fd.schedule_interval),
-											 Float8GetDatum(MAX_INTERVALS_BACKOFF));
+		Datum ival_max;
+		// max wait time to launch job is 1 minute
+		Interval interval_max = { .time = 60000000 };
+		Interval retry_ival = { .time = 2000000 };
+		retry_ival.time += rand_backoff;
+		if (!launch_failure)
```
mfundul, line 240:
```suggestion
		if (launch_failure)
```
I think?

konskov: No, that should be the logic there. Launch failure is when the job couldn't be started because of lack of background workers, and in that case, we have the random backoff in seconds calculated above. 
In other cases (job threw runtime error) we have a retry period of 5 minutes, and so the backoff there grows like 5, 10, 20 minutes until it reaches the schedule interval to take the retry period into account. Unless you meant to put the launch_failure case first which would be more readable so I'll do that

mfundul (approved):

PR #4573, opened 2022-08-03
Load TSL library on compressed_data_out call

pmwkaa:
A call to `compressed_data_out` from a replication worker would
produce a misleading error saying that your license is "timescale"
and you should upgrade to "timescale" license, even if you have
already upgraded.

As a workaround, we try to load the TSL module it in this function.
It will still error out in the "apache" version as intended.

We already had the same fix for `compressed_data_in` function.

### DIFF
--- a/src/cross_module_fn.c
+++ b/src/cross_module_fn.c
@@ -207,22 +207,35 @@ error_no_default_fn_pg_community(PG_FUNCTION_ARGS)
 	pg_unreachable();
 }
 
+/*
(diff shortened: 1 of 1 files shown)


svenklemm:
Hmm that doesnt seem like the real fix and seems more like a bandaid. What happens when the license is set to Apache and someone tries to call this function?

fabriziomello (approved):

jnidzwetzki (approved):

PR #4577, opened 2022-08-05
Fix segfaults in compression code with corrupt data

svenklemm:
Sanity check the compression header for sane algorithm
before using it as index into an array. Previously
this would result in a segfault and could happen with
corrupted compressed data.

Disable-check: commit-count

### DIFF
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -164,6 +164,17 @@ static void row_compressor_finish(RowCompressor *row_compressor);
  ** compress_chunk **
  ********************/
 
+static CompressedDataHeader *
+get_compressed_data_header(Datum data)
+{
+	CompressedDataHeader *header = (CompressedDataHeader *) PG_DETOAST_DATUM(data);
+
+	if (header->compression_algorithm >= _END_COMPRESSION_ALGORITHMS)
+		elog(ERROR, "invalid compression algorithm %d", header->compression_algorithm);
+
+	return header;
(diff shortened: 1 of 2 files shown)


akuzm (approved):

tsl/src/compression/compression.c
```diff
@@ -1364,13 +1371,13 @@ tsl_compressed_data_decompress_reverse(PG_FUNCTION_ARGS)
...
 	if (SRF_IS_FIRSTCALL())
 	{
 		funcctx = SRF_FIRSTCALL_INIT();
 		oldcontext = MemoryContextSwitchTo(funcctx->multi_call_memory_ctx);
 
+		header = get_compressed_data_header(PG_GETARG_DATUM(0));
+
 		iter =
 			definitions[header->compression_algorithm]
 				.iterator_init_reverse(PG_GETARG_DATUM(0), get_fn_expr_argtype(fcinfo->flinfo, 1));
```
konskov, line 1383: I didn't understand why this is not `.iterator_init_reverse(PointerGetDatum(header), get_fn_expr_argtype(fcinfo->flinfo, 1));`

konskov (approved):

tsl/src/compression/compression.c
svenklemm: you are right, your suggestion is better cause it prevents  detoasting twice

gayyappan (approved):

PR #4583, opened 2022-08-07
Fix partitioning functions

### LINKED ISSUE #4575: [Bug]: unsupported expression argument node type 112 when calling `_timescaledb_internal.get_partition_hash`
### What type of bug is this?

Unexpected error

Query executor

### What happened?

When calling the `_timescaledb_internal.get_partition_hash` function from a procedure or immutable function, the query fails with the error message `Query 1 ERROR: ERROR:  unsupported expression argument node type 112`.
I could not find any additional info in the logs.

I initially ran into this bug on 2.7.0 and thought it might be related to: #4489. Therefore i tried upgrading to 2.7.2 where this issue should be fixed, but it seems to not fix this. 

2.7.0 - 2.7.2

### PostgreSQL version used

14.4

MAC OS 11.6.2

Docker

On prem/Self-hosted

```bash
Query 1 ERROR: ERROR:  unsupported expression argument node type 112
```

```bash
CREATE OR REPLACE FUNCTION my_function(IN partition_key integer) RETURNS integer
AS 'select _timescaledb_internal.get_partition_hash($1);'
LANGUAGE SQL;

CREATE OR REPLACE FUNCTION my_immutable_function(IN partition_key integer) RETURNS integer
AS 'select _timescaledb_internal.get_partition_hash($1);'
LANGUAGE SQL
IMMUTABLE;

CREATE OR REPLACE PROCEDURE my_procedure(IN partition_key integer)
AS $$
DECLARE
    partition_hash integer; 
BEGIN
	-- Need to do some things based on partition hash
	select _timescaledb_internal.get_partition_hash(partition_key) INTO partition_hash; -- this eventually fails with the unsupported node type error	
END;
$$ LANGUAGE PLPGSQL;

SELECT my_function(1); -- This will not fail
SELECT my_immutable_function(1); -- This will always fail with: Query 1 ERROR: ERROR:  unsupported expression argument node type 112
CALL my_procedure(1); -- This will eventually fail with same error as above
```

fabriziomello:
When executing `get_partition_{hash|for_key}` inside an IMMUTABLE
function we're getting the following error:

`ERROR: unsupported expression argument node type 112`

This error is because the underlying `resolve_function_argtype` was not
dealing with `T_Param` node type.

Fixed it by dealing properly with `T_Param` node type returning the
`paramtype` for the argument type.

Fixes #4575

### DIFF
--- a/src/partitioning.c
+++ b/src/partitioning.c
@@ -317,6 +317,9 @@ resolve_function_argtype(FunctionCallInfo fcinfo)
 			/* Argument is function, so our input is its result type */
 			argtype = ((FuncExpr *) node)->funcresulttype;
 			break;
+		case T_Param:
+			argtype = ((Param *) node)->paramtype;
+			break;
 		default:
 			elog(ERROR, "unsupported expression argument node type %u", nodeTag(node));
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,11 @@ accidentally triggering the load of a previous DB version.**
 
 **Bugfixes**
 * #4486 Adding boolean column with default value doesn't work on compressed table
+* #4575 Fix use of `get_partition_hash` and `get_partition_for_key` inside an IMMUTABLE function
 
 **Thanks**
 @janko for reporting
+@AlmiS for reporting error on `get_partition_hash` executed inside an IMMUTABLE function
 
 ## 2.7.2 (2022-07-26)
 
--- a/test/expected/hash.out
+++ b/test/expected/hash.out
@@ -302,3 +302,22 @@ SELECT _timescaledb_internal.get_partition_hash('08002b:010203'::macaddr);
           294987870
 (1 row)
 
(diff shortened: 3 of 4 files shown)


pmwkaa (approved):

src/partitioning.c
```diff
@@ -317,6 +317,9 @@ resolve_function_argtype(FunctionCallInfo fcinfo)
 			/* Argument is function, so our input is its result type */
 			argtype = ((FuncExpr *) node)->funcresulttype;
 			break;
+		case T_Param:
+			argtype = ((Param *) node)->paramtype;
+			break;
 		default:
 			elog(ERROR, "unsupported expression argument node type %u", nodeTag(node));
```
svenklemm, line 324: might be nice to use ts_get_node_name here but that would require adding the primnodes.h nodes to that function

svenklemm (approved):

src/partitioning.c
fabriziomello: will send a following PR adding it, thanks!

PR #4590, opened 2022-08-09
Do not perform chunk exclusion twice

akuzm:
Specifically, if we're doing a SELECT and already excluded the chunks
that do not match the WHERE clause while expanding the hypertable,
don't call the Postgres function that's supposed to do the same --
relation_excluded_by_constraints(). For DELETE/UPDATE, our inheritance
expansion code is not invoked, so this optimization does not apply.

#### Explicitly using the custom partitioning function in a condition
```
-- Exclude using the function on time
EXPLAIN (verbose, costs off)
SELECT * FROM part_time_func WHERE time_partfunc(time) < '2018-07-01';
```

Here `time_partfunc` is the custom partitioning function that converts the incoming unix epoch into timestamptz. This condition is not on `time`, so our chunk exclusion can't use it. But the postgres constraints have the form of:
 ```
 ((time_partfunc("time") >= 'Wed Jun 27 17:00:00 2018 PDT'::timestamp with time zone) AND (time_partfunc("time") < 'Wed Jul 04 17:00:00 2018 PDT'::timestamp with time zone))
 ```
 So it works.

#### Disjoint interval on time with OR clause
```
explain select * from metrics_timestamptz m1 WHERE m1.time < '2000-01-10' OR m1.time > '2001-01-01';
```

Our chunk exclusion doesn't support OR clauses. It can work only with a single interval on partitioning dimension, not multiple disjoint intervals.

#### Join clause for nullable side of join is not used for chunk exclusion
```
explain SELECT d.*,
    m.*
FROM device_tbl d
    LEFT OUTER JOIN metrics_ordered_idx m ON m.device_id = d.device_id
    AND m.time > '2019-01-01'
WHERE d.device_id = 8
ORDER BY m.v0;
```

This happens because we're using our own join qual processing logic based on the raw parse tree, see `collect_join_quals`. This logic cuts some corners. Not sure it is still needed, because in PG 14 the `baserestrictinfo` is properly filled at this point (`ts_plan_expand_hypertable_chunks`), and contains the needed clause.


erimatnor:
@akuzm Is this still something that you plan to take out of draft or should it be closed?

akuzm:
> @akuzm Is this still something that you plan to take out of draft or should it be closed?

I'll close it, it was mostly to investigate some things, and I wrote the result down in the PR description. Not something actionable at the moment.

PR #4591, opened 2022-08-09
Add inherited check constraints to OSM chunk

gayyappan:
When a table is added to an inheritance hierrachy, PG checks
if all check constraints are present on this table. When a OSM chunk
is added as a child of a hypertable with constraints,
make sure that all check constraints are replicated on the child OSM
chunk as well.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -1115,7 +1115,8 @@ chunk_create_table_constraints(const Chunk *chunk)
 								chunk->hypertable_relid,
 								chunk->fd.hypertable_id);
 
-	if (chunk->relkind == RELKIND_RELATION)
+	if (chunk->relkind == RELKIND_RELATION &&
+		!ts_flags_are_set_32(chunk->fd.status, CHUNK_STATUS_FOREIGN))
 	{
 		ts_trigger_create_all_on_chunk(chunk);
 		ts_chunk_index_create_all(chunk->fd.hypertable_id,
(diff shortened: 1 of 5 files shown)


src/chunk.c
```diff
@@ -4569,10 +4570,22 @@ add_foreign_table_as_chunk(Oid relid, Hypertable *parent_ht)
...
 	ts_dimension_slice_insert_multi(chunk->cube->slices, chunk->cube->num_slices);
-	ts_chunk_constraints_add_dimension_constraints(chunk->constraints, chunk->fd.id, chunk->cube);
 	/* check constraints are not automatically created for foreign tables.
 	 * See: ts_chunk_constraints_add_dimension_constraints.
+	 * Collect all the check constraints from the hypertable and add them to the
+	 * foreign table. Otherwise, cannot add as child of the hypertable (pg inheritance
+	 * code will error. Note that the name of the check constraint on the hypertable
+	 * and the foreign table chunk should match.
+	 * Also, the dimension constraint is not converted to a check constraint on the
+	 * foreign table chunk
```
svenklemm, line 4580: There should be no dimension constraints processed here as they are not present on the parent table only on individual chunks.

svenklemm (approved):

src/chunk.c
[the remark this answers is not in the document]
gayyappan: We add a dummy dimension constraint on the OSM chunk (line 4588). This is metadata in the chunk catalog table (cannot create a chunk without this info). But this is not translated into a check constraint. Is that what you are referring to?

gayyappan: We add a dummy dimension constraint on the OSM chunk (line 4588). This is metadata in the chunk catalog table (cannot create a chunk without this info). But this is not translated into a check constraint. Is that what you are referring to?

akuzm (approved):

src/chunk.c
gayyappan: Updated the comment in the source code.

PR #4595, opened 2022-08-10
Modify chunk exclusion to include OSM chunks  (draft)

gayyappan:
The chunk exclusion code examines the dimension_slice catalog
table to filter out chunks. OSM chunks do not register their
actual dimension ranges with the catalogs. This metadata is
managed by OSM. So an OSM chunk, if it exists, should not
be excluded.

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -1053,8 +1053,7 @@ chunk_create_table_constraints(const Chunk *chunk)
 								chunk->hypertable_relid,
 								chunk->fd.hypertable_id);
 
-	if (chunk->relkind == RELKIND_RELATION &&
-		!ts_flags_are_set_32(chunk->fd.status, CHUNK_STATUS_FOREIGN))
+	if (chunk->relkind == RELKIND_RELATION && !ts_chunk_is_osm_chunk(chunk))
 	{
 		ts_trigger_create_all_on_chunk(chunk);
 		ts_chunk_index_create_all(chunk->fd.hypertable_id,
@@ -1613,7 +1612,7 @@ chunk_tuple_found(TupleInfo *ti, void *arg)
 	chunk->hypertable_relid = ts_hypertable_id_to_relid(chunk->fd.hypertable_id);
 	chunk->relkind = get_rel_relkind(chunk->table_id);
 
-	if (chunk->relkind == RELKIND_FOREIGN_TABLE)
+	if (chunk->relkind == RELKIND_FOREIGN_TABLE && ts_chunk_is_osm_chunk(chunk) == false)
 		chunk->data_nodes = ts_chunk_data_node_scan_by_chunk_id(chunk->fd.id, ti->mctx);
 
 	return SCAN_DONE;
@@ -4250,6 +4249,12 @@ ts_chunk_is_uncompressed_or_unordered(const Chunk *chunk)
 			!ts_flags_are_set_32(chunk->fd.status, CHUNK_STATUS_COMPRESSED));
 }
 
+bool
+ts_chunk_is_osm_chunk(const Chunk *chunk)
+{
+	return ts_flags_are_set_32(chunk->fd.status, CHUNK_STATUS_FOREIGN);
+}
+
 static const char *
 get_chunk_operation_str(ChunkOperation cmd)
 {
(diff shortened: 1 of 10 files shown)


src/chunk_scan.c
```diff
@@ -240,6 +253,8 @@ ts_chunk_scan_by_constraints(const Hyperspace *hs, const List *dimension_vecs,
 				unlocked_chunks[unlocked_chunk_count] = chunk;
 				unlocked_chunk_count++;
 
+				if (ts_chunk_is_osm_chunk(chunk))
```
akuzm, line 256: If the OSM chunk doesn't have dimension slices, should we `Assert(!ts_chunk_is_osm_chunk(chunk))` here?

src/chunk.c
```diff
@@ -4613,3 +4618,70 @@ ts_chunk_attach_osm_table_chunk(PG_FUNCTION_ARGS)
...
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(osm_chunk_status));
+	ScanKeyInit(&scankey[1],
+				Anum_chunk_status_hypertable_id_idx_hypertable_id,
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(hypertable_id));
+
+	num_found = ts_scanner_scan(&scanctx);
```
akuzm, line 4677: Do you want to simplify this to an index-only scan? You'll need to make a covering index  with `include (id)`, set `.table = NULL` and `.want_itup` in the context, use `ts_scanner_next` to get the index tuple, and then `index_getattr()` to get the chunk id from the index tuple directly.

akuzm: This would be less I/O and less code.

akuzm (approved):

src/chunk_scan.c
gayyappan: the chunk has dimension slices, but are dummy ranges.

src/chunk.c
```diff
@@ -1675,7 +1674,7 @@ chunk_tuple_found(TupleInfo *ti, void *arg)
 	chunk->hypertable_relid = ts_hypertable_id_to_relid(chunk->fd.hypertable_id);
 	chunk->relkind = get_rel_relkind(chunk->table_id);
 
-	if (chunk->relkind == RELKIND_FOREIGN_TABLE)
+	if (chunk->relkind == RELKIND_FOREIGN_TABLE && ts_chunk_is_osm_chunk(chunk) == false)
```
svenklemm, line 1677:
```suggestion
	if (chunk->relkind == RELKIND_FOREIGN_TABLE && !ts_chunk_is_osm_chunk(chunk) )
```

src/chunk.c
```diff
@@ -4613,3 +4618,70 @@ ts_chunk_attach_osm_table_chunk(PG_FUNCTION_ARGS)
...
+	};
+
+	/*
+	 * Perform an index scan on status + hypertable ID
+	 */
+	ScanKeyInit(&scankey[0],
+				Anum_chunk_status_hypertable_id_idx_status,
+				BTEqualStrategyNumber,
+				F_INT4EQ,
+				Int32GetDatum(osm_chunk_status));
```
svenklemm, line 4670: Since chunk.status is a bit field this is not correct and might bite us in the future.

gayyappan: The bit operator, biteq , works on VARBIT type - this will not work with the int status type.
(see: https://www.postgresql.org/docs/current/datatype-bit.html).

Looks like I have to go back to the previous approach of adding a new column instead of reusing the status bits to identify an OSM chunk

akuzm: Is it ever combined with other statuses? Maybe we can assert that it's not and use it this way. It's kinda hackish, but adding another column will inevitably bloat the catalog and slow down all queries. Not sure by how much, but chunk catalog lookups do constitute a significant part of planning time already.

src/chunk_scan.c
akuzm: Is it ever going to appear in the results of a lookup by constraints? I guess not, so probably we could assert this.

akuzm: I mean, if it did appear there sometimes, it would be weird -- when, and why not always?

gayyappan: Yes, it should appear in this list with a query that has < operator:
e.g. select * from ht where ts < '2020-01-01'

PR #4597, opened 2022-08-11
Centralize CI settings

svenklemm:
This patch adds a new settings file for common github action settings.
Instead of repeating latest pg versions in every github workflow we
can read the settings from this central file.

### DIFF
--- a/.github/ci_settings.py
+++ b/.github/ci_settings.py
@@ -1,0 +1,16 @@
+#!/usr/bin/env python
+
+#  This file and its contents are licensed under the Apache License 2.0.
+#  Please see the included NOTICE for copyright information and
+#  LICENSE-APACHE for a copy of the license.
+
+# Common settings for our CI jobs
+
+PG12_EARLIEST = "12.0"
+PG12_LATEST = "12.12"
+PG13_EARLIEST = "13.2"
+PG13_LATEST = "13.8"
+PG14_EARLIEST = "14.0"
+PG14_LATEST = "14.5"
+
+PG_LATEST = [PG12_LATEST, PG13_LATEST, PG14_LATEST]
--- a/.github/gh_config_reader.py
+++ b/.github/gh_config_reader.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python
+
+#  This file and its contents are licensed under the Apache License 2.0.
+#  Please see the included NOTICE for copyright information and
+#  LICENSE-APACHE for a copy of the license.
+
+import json
+import ci_settings
+
+# generate commands to set github action variables
+for key in dir(ci_settings):
+    if not key.startswith("__"):
+        value = getattr(ci_settings, key)
+        print(str.format("::set-output name={0}::{1}", key, json.dumps(value)))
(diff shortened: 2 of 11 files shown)


.github/ci_settings.py
```diff
@@ -0,0 +1,17 @@
...
+#  LICENSE-APACHE for a copy of the license.
+
+# Common settings for our CI jobs
+
+PG12_EARLIEST = "12.0"
+PG12_LATEST = "12.11"
+PG13_EARLIEST = "13.2"
+PG13_LATEST = "13.7"
+PG14_EARLIEST = "14.0"
+PG14_LATEST = "14.4"
```
fabriziomello, line 14: What about get the LATEST versions from here: https://www.postgresql.org/versions.rss ?

svenklemm: We want this to be an explicit decission and not depend on external services. Also on release the new version is not available on all platforms so that would mean every new postgres release would break our CI till every platform has all the packages.

fabriziomello (approved):

.github/gh_config_reader.py
```diff
@@ -0,0 +1,15 @@
...
+#  LICENSE-APACHE for a copy of the license.
+
+import json
+import ci_settings
+
+# generate commands to set github action variables
+for key in dir(ci_settings):
+  if not key.startswith('__'):
+    value = getattr(ci_settings, key)
+    print(str.format("::set-output name={0}::{1}",key, json.dumps(value)))
```
jnidzwetzki, line 14: Nit

```suggestion
    print(str.format("::set-output name={0}::{1}", key, json.dumps(value)))
```

.github/workflows/sqlsmith.yaml
```diff
@@ -8,13 +8,24 @@ on:
...
+    - name: Checkout source code
+      uses: actions/checkout@v2
+    - name: Build matrix
+      id: setter
+      run: python .github/gh_config_reader.py
+
   regress:
     name: SQLsmith PG${{ matrix.pg }}
     runs-on: ${{ matrix.os }}
     strategy:
```
jnidzwetzki, line 25: Most other GitHub actions that are using `gh_config_reader.py` have a dependency on the new `config` job (`needs: config`). From my point of view, this would also make sense for the `regress` job.

.github/workflows/update-test.yaml
```diff
@@ -8,12 +8,23 @@ on:
...
+    steps:
+    - name: Checkout source code
+      uses: actions/checkout@v2
+    - name: Build matrix
+      id: setter
+      run: python .github/gh_config_reader.py
+
   update_test:
     name: Update test PG${{ matrix.pg }}
     runs-on: 'ubuntu-latest'
```
jnidzwetzki, line 24: Most other GitHub actions that are using `gh_config_reader.py` have a dependency on the new `config` job (`needs: config`). From my point of view, this would also make sense for the `update_test` job.

jnidzwetzki (approved):
A few minor comments, the rest looks good to me.

PR #4602, opened 2022-08-11
Regenerate SSL certs

svenklemm:
The SSL certs used for testing expired today. This patch adds new
certificates that last for 100 years.


mfundul (approved):
> This patch adds new certificates that last for 100 years.

Only 100? Show some faith!

fabriziomello (approved):

PR #4606, opened 2022-08-12
Fix coccinelle hash_create script

svenklemm:
Previously the hash_create coccinelle script would not detect
hash_create calls that declared the return variable in the same
statement or that used the return value of hash_create as return
value.

Disable-check: commit-count

### DIFF
--- a/coccinelle/hash_create.cocci
+++ b/coccinelle/hash_create.cocci
@@ -5,34 +5,35 @@
 // to be explicit about the memory context our hash tables live in so we enforce
 // usage of the flag.
 @ hash_create @
-expression res;
 position p;
 @@
 
-res@p = hash_create(...);
+hash_create@p(...)
 
 @safelist@
-expression res;
 expression arg1, arg2, arg3;
 expression w1, w2;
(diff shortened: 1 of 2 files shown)


coccinelle/hash_create.cocci
```diff
@@ -5,34 +5,35 @@
...
 )
-@ depends on !safelist @
+@safelist2@
 expression res;
+expression arg1, arg2, arg3;
+expression flags;
+position hash_create.p;
+@@
+Assert(flags & HASH_CONTEXT);
+res = hash_create@p(arg1,arg2,arg3, flags);
```
jnidzwetzki, line 32: Is the `res` expression still needed? The following line seems to capture the same code blocks in my tests and the syntax would be similar to the expressions in `safelist`:

```suggestion
hash_create@p(arg1,arg2,arg3, flags)
```

jnidzwetzki (approved):

akuzm (approved):
It does find https://github.com/timescale/timescaledb/pull/4603 , nice.

Issue #4576, opened 2022-08-04
[Bug]: OOM condition when loading large data set in to hypertable


ninjaltd:
### What type of bug is this?

Crash

### What subsystems and features are affected?

Other

### What happened?

Inserting a large sized data set into a hypertable triggering a memory leak and causes system to go OOM.
Please refer issue #4560.  Upgrading to 2.7.2 did not fix this but the issue got closed.  So I am opening a new one.

### TimescaleDB version affected

2.7.2

### PostgreSQL version used

14.4

### What operating system did you use?

Debian GNU/Linux 11

### What installation method did you use?

Deb/Apt

### What platform did you run on?

On prem/Self-hosted

### Relevant log output and stack trace

```bash
n/a
```

### How can we reproduce the bug?

```bash
CREATE DATABASE test;
\c test
CREATE TABLE inet_session_usages (
    id bigint NOT NULL,
    inet_session_id integer NOT NULL,
    bytes_in bigint,
    bytes_out bigint,
    created_at timestamp with time zone,
    year_created integer,
    month_created integer,
    day_created integer,
    hour_created integer,
    minute_created integer,
    created_at_svc_tz timestamp without time zone
);
SELECT create_hypertable('inet_session_usages', 'created_at_svc_tz', chunk_time_interval => INTERVAL '1 day');
create unique index inet_session_usages_session_id_idx on inet_session_usages (created_at_svc_tz DESC, inet_session_id);

wget -O - https://tyl.spirit.net.au/~z/inet_session_usages.sql.xz | xz -d | psql test
```

ninjaltd:
Sorry should have been:  

wget -O - https://tyl.spirit.net.au/~z/inet_session_usages.sql.xz | xz -d - | psql test

mrksngl:
I likely hit the same issue.

I inserted 170GiB data using `COPY ... FROM STDIN WITH BINARY`.

For reference, I tried this with normal postgres, including the indexes which are normally created by timescale. This peaked at around 1.38GiB memory.

Doing the same with timescale, memory consumption rises in huge steps and I get an OOM kill before data is fully copied. Depending on the memory limit (postgres runs inside a kubernetes pod), it can take more or less data, but even 8GiB of memory were not enough to copy the data -- it stopped when roughly 120GiB was transferred.

@ninjaltd also pointed out issue #4650, but different from there I get the same behavior whether space partitioning is used or not.

I also ran into the following two interesting facts:
[...]

PR #4611, opened 2022-08-15
Do not allocate temporary data in PortalContext

akuzm:
The hyperspace point corresponding to the inserted row should be
allocated in the per-tuple context.

Fixes: https://github.com/timescale/timescaledb/issues/4576 (Fabrizio repoted that this fix largely helps).


fabriziomello:
@akuzm I repeat the tests locally using this PR and we still leak somewhere, after some minutes running the COPY the backend memory went over 4GB

```bash
➜ ps faux | grep postgres | grep COPY
fabrizio   67960 92.7  6.0 4057900 3960964 ?     RNs  14:52   9:25      \_ postgres: fabrizio fabrizio [local] COPY
```

akuzm:
Somehow it only starts to leak for real after 30 minutes for me. I fixed some other leaks, fixed my tools that I used to find leaks, and also added a test for memory usage in PortalContext. Should be better now.

src/hypertable.c
```diff
@@ -1048,7 +1048,7 @@ hypertable_chunk_store_add(const Hypertable *h, const Chunk *input_chunk)
 	ts_subspace_store_add(h->chunk_cache,
 						  cached_chunk->cube,
 						  cached_chunk,
-						  /* object_free = */ NULL);
+						  (void (*)(void *)) ts_chunk_free);
```
mfundul, line 1051: That's a bit of a risky cast. Why not do it inside the function like in `/src/nodes/chunk_dispatch.c`:
```
static void
destroy_chunk_insert_state(void *cis)
{
	ts_chunk_insert_state_destroy((ChunkInsertState *) cis);
}
```

akuzm: You mean, in case the signature changes? Sure, I can do it.

mfundul: Technically it is considered undefined behaviour. I'm sure it works in practice e.g. for gcc.

fabriziomello:
> Somehow it only starts to leak for real after 30 minutes for me. I fixed some other leaks, fixed my tools that I used to find leaks, and also added a test for memory usage in PortalContext. Should be better now.

@akuzm In my laptop start leak immediately ... :-( 

Much better now with the current version.

```bash
➜ ps faux | grep postgres | grep COPY
fabrizio  696080 99.6  0.2 250356 165704 ?       RNs  18:24   6:28      \_ postgres: fabrizio fabrizio [local] COPY

➜ ps faux | grep postgres | grep COPY
fabrizio  696080 99.8  0.2 254560 168372 ?       RNs  18:24  10:58      \_ postgres: fabrizio fabrizio [local] COPY

➜ ps faux | grep postgres | grep COPY
fabrizio  696080 99.7  0.2 257068 169708 ?       RNs  18:24  15:24      \_ postgres: fabrizio fabrizio [local] COPY

➜ ps faux | grep postgres | grep COPY
fabrizio  696080 99.1  0.2 287640 196272 ?       RNs  18:24  20:35      \_ postgres: fabrizio fabrizio [local] COPY
```

Still increasing a bit over the time but orders of magnitude less than `main` branch.

Ps: and now my cpu fan is saying "thank you AlexK"

svenklemm (approved):

fabriziomello (approved):

PR #4612, opened 2022-08-15
Migrate windows tests to GitHub Actions

svenklemm:
This patch changes our windows test to run in GitHub actios instead
of appveyor. This allows us to drop appveyor and have all CI workflows
in one place. This also improves our windows testing as appveyor was
only running on PG12 and this patch changes the windows tests to run
on all supported PG versions. This patch also changes the windows
tests to use the same configuration as the other tests.
Since our test infrastructure heavily relies on shell scripting
the test runner cannot run natively on Windows. We use WSL instead
to get a unix environment to execute the test runner.

Disable-check: commit-count


fabriziomello (approved):

.github/workflows/windows-build-and-test.yaml
```diff
@@ -49,20 +57,89 @@ jobs:
...
+      if: matrix.pg != '12'
+      shell: wsl-bash {0}
+      run: |
+        export TEST_TABLESPACE1_PREFIX='${{ env.TABLESPACE1 }}'
+        export TEST_TABLESPACE2_PREFIX='${{ env.TABLESPACE2 }}'
+        cmake -B build_wsl -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DTEST_PGPORT_LOCAL=${{ env.PGPORT }}
+        make -C build_wsl regresschecklocal
+
+    - name: Setup postgres cluster for TSL tests
+      run: |
```
jnidzwetzki, line 110: Is it correct that this step should be performed with PG 12 (because the `Run TSL tests` step is excluded for PG 12)?

jnidzwetzki (approved):
I have one comment regarding PG 12, the rest looks good.

PR #4619, opened 2022-08-16
GROUP BY error when setting compress_segmentby with an enum column

sb230132:


svenklemm (changes requested):
You should not change the vartype of the Var but instead fall back to anyenum lookup for the operators when the type is an enum and the lookup with the explicit type fails.

fix
```diff
@@ -0,0 +1,132 @@
+diff --git a/tsl/src/nodes/decompress_chunk/decompress_chunk.c b/tsl/src/nodes/decompress_chunk/decompress_chunk.c
+index 7c88541f..237d7145 100644
+--- a/tsl/src/nodes/decompress_chunk/decompress_chunk.c
++++ b/tsl/src/nodes/decompress_chunk/decompress_chunk.c
```
mfundul, line 4: You should probably remove this file.

tsl/src/fdw/scan_plan.c
```diff
@@ -228,7 +228,9 @@ evaluate_stable_function(Oid funcid, Oid result_type, int32 result_typmod, Oid r
 		else
 			has_nonconst_input = true;
 	}
-
+	if (has_null_input)
+	{
+	}
```
fabriziomello, line 233:
```suggestion
```

fabriziomello:
Some comments:

1. You need provide a better commit message describing what is the problem you're solving and how you fixed it (for example #4582)
2. Remember squash commits before rebase it to the main branch, unless you really want to send several commits, and for that each one should have a good description and also use the `Disable-check: commit-count` in the PR description to silent the CI
3. Add to the thanks section in CHANGELOG the issue reporter

fix
fabriziomello: This file should be removed, no??

sb230132: I don't remember committing this file. Anyways removed it now.
Thanks.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -210,6 +212,20 @@ build_compressed_scan_pathkeys(SortInfo *sort_info, PlannerInfo *root, List *chu
...
+				{
+					sortop = get_opfamily_member(pk->pk_opfamily,
+												 ANYENUMOID,
+												 ANYENUMOID,
+												 pk->pk_strategy);
+				}
+				else
+				{
+					elog(ERROR, "Invalid segmentby column");
+				}
```
fabriziomello, line 227: Do we need this branch? If yes shouldn't have a test case for this?

fabriziomello:
Seems your code editor messed (automatically formatted) the CHANGELOG.md... we prefer to send coding format changes into a separated PR because it make useless the git blame. Check the .git-blame-ignore-revs in the root of source three.

svenklemm (commented):
I think we need to have similar check when we enable compression for the hypertable. No point having a compressed hypertable that errors when you try to query it.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -210,6 +212,20 @@ build_compressed_scan_pathkeys(SortInfo *sort_info, PlannerInfo *root, List *chu
...
+				if (type_is_enum(var->vartype))
+				{
+					sortop = get_opfamily_member(pk->pk_opfamily,
+												 ANYENUMOID,
+												 ANYENUMOID,
+												 pk->pk_strategy);
+				}
+				else
+				{
+					elog(ERROR, "Invalid segmentby column");
```
svenklemm, line 226: This does not conform to pg error message style guide: https://www.postgresql.org/docs/current/error-style-guide.html
Should probably be more along the lines of `sort operator lookup failed for column "%s"`

tsl/src/nodes/decompress_chunk/decompress_chunk.c
sb230132: Agree that its good to have a testcase. However iam not sure what test will hit this code.
I think it's better to have this code, so that it may hit in future. If we remove this, then code may further execute and report some unknown error or even worse case crash or return wrong results.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
sb230132: Fixed as per your suggestion.

sb230132:
> Seems your code editor messed (automatically formatted) the CHANGELOG.md... we prefer to send coding format changes into a separated PR because it make useless the git blame. Check the .git-blame-ignore-revs in the root of source three.


Right, its my editor which screwed up the formatting. Fixed now.

sb230132:
> I think we need to have similar check when we enable compression for the hypertable. No point having a compressed hypertable that errors when you try to query it.

May be yes. But i dont have all the context required to check for these conditions during ALTER TABLE ... ENABLE COMPRESSION.

tsl/src/fdw/scan_plan.c
sb230132: Fixed by removing the change.

fabriziomello: I guess you forgot to remove it, no?

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -210,6 +212,20 @@ build_compressed_scan_pathkeys(SortInfo *sort_info, PlannerInfo *root, List *chu
...
+				if (type_is_enum(var->vartype))
+				{
+					sortop = get_opfamily_member(pk->pk_opfamily,
+												 ANYENUMOID,
+												 ANYENUMOID,
+												 pk->pk_strategy);
+				}
+				else
+				{
+					elog(ERROR, "Sort operator lookup failed for column \"%s\"", column_name);
```
fabriziomello, line 226:
```suggestion
					elog(ERROR, "sort operator lookup failed for column \"%s\"", column_name);
```

tsl/src/fdw/scan_plan.c
sb230132: Right, i had this change in my main branch. Thus this change was re-appearing again and again. Reverted this change in main branch and fixed in current branch as well.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
sb230132: Fixed.

CHANGELOG.md
```diff
@@ -2495,3 +2496,6 @@ the next release.
 * [72f754a] use PostgreSQL's own `hash_any` function as default partfunc (thanks @robin900)
 * [39f4c0f] Remove sample data instructions and point to docs site
 * [9015314] Revised the `get_general_index_definition` function to handle cases where indexes have definitions other than just `CREATE INDEX` (thanks @bricklen)
+
+**Bugfixes**
+* #3481 GROUP BY error when setting compress_segmentby with an enum column
```
svenklemm, line 2501: we reference PRs here not issues
```suggestion
* #4619 GROUP BY error when setting compress_segmentby with an enum column
```

svenklemm (approved):

sb230132 (commented):
Fixed Sven comment.

fabriziomello (approved):

Issue #3376, opened 2021-06-30
Size of the chunks increase without being decompressed if decompress_chunk(i) fails in middle 


hardikm10:
**Relevant system information:**
OS: [e.g. Ubuntu 16.04, Windows 10 x64, etc]: Docker / timescale/timescaledb:2.3.0-pg12
 - PostgreSQL version (output of `postgres --version`): 12.7
 - TimescaleDB version (output of `\dx` in `psql`): 2.3.0
 - Installation method: [e.g., "using Docker", "apt install", "source"] :  "Using Docker"

**Describe the bug**
If `select decompress_chunk(i) FROM show_chunks('<hypertable>') i` fails in the middle because there was 'a chunk' which was already decompressed, the chunks on which it has acted upon (before the failure) still shows as compressed (which is expected) but their size increases somehow( which isn't expected).  Upon repeated execution of the above failing command, the size increase continuously with the status still remain as 'compressed'

**To Reproduce**
```
drop table sensor_data;

create table sensor_data( 
time timestamptz not null, 
sensor_id integer not null, 
cpu double precision null,
temperature double precision null );

select from create_hypertable('sensor_data','time');

INSERT INTO sensor_data
	SELECT
		time + (INTERVAL '1 minute' * random()) AS time,
		sensor_id,
		random() AS cpu,
		random()* 100 AS temperature
	FROM
		generate_series(now() - INTERVAL '1 months', now() - INTERVAL '1 week', INTERVAL '1 minute') AS g1(time),
		generate_series(1, 100, 1 ) AS g2(sensor_id)
	ORDER BY
		time;
```
```
SELECT *,pg_size_pretty(total_bytes) FROM chunks_detailed_size('sensor_data');
     SELECT chunk_name,pg_size_pretty(total_bytes) as total_size FROM chunks_detailed_size('sensor_data');
    chunk_name    | total_size
------------------+------------
 _hyper_1_1_chunk | 51 MB
 _hyper_1_2_chunk | 97 MB
 _hyper_1_3_chunk | 97 MB
 _hyper_1_4_chunk | 87 MB
 ```
```
alter table sensor_data set (timescaledb.compress, 
timescaledb.compress_orderby = 'time DESC');

defaultdb=> select compress_chunk(i) FROM show_chunks('sensor_data', older_than => '8 days'::INTERVAL) i;                   
------------------------------------------
 _timescaledb_internal._hyper_1_1_chunk
[...]

Issue #4578, opened 2022-08-05
[Bug]: INSERT into compressed chunks does not clean dead tuples


svenklemm:
### What type of bug is this?

Performance issue

### What subsystems and features are affected?

Compression

### What happened?

We have a "storage leak" in compression, more specifically the insert into compressed chunks feature.  The leak is due to the way we disable vacuum on these chunks leading to dead tuples never getting cleaned up. To detect these simply checking n_dead_tup is not enough because it will never be updated, you need to explicitly ANALYZE the chunk and check n_dead_tup after. This is also a performance issue as all those dead tuples have to be scanned when scanning the table.

### TimescaleDB version affected

2.3+

### PostgreSQL version used

14.4

### What operating system did you use?

Archlinux 2022.07.01 x64

### What installation method did you use?

Source

### What platform did you run on?

On prem/Self-hosted

### Relevant log output and stack trace

```bash
SELECT n_dead_tup from pg_stat_user_tables where relid=<chunk with compression>;
-[ RECORD 1 ]----------
n_dead_tup | 1363992372
```

svenklemm:
Seems like this is only a problem when the job errors. Couldn't reproduce it in normal operation.

antekresic:
Fixed by #5118

PR #4645, opened 2022-08-23
Size of the chunks increase without being decompressed if decompress_chunk(i) fails in middle

sb230132:
Issue:
Consider a compressed hypertable (sensor_data) consisting of 4 chunks named as below with total sizes.
    chunk_name    | total_size
------------------+------------
 _hyper_1_1_chunk | 11 MB
 _hyper_1_2_chunk | 20 MB
 _hyper_1_3_chunk | 20 MB
 _hyper_1_4_chunk | 87 MB

Except _hyper_1_4_chunk all other chunks are compressed.

Execute:
select decompress_chunk('_timescaledb_internal._hyper_1_2_chunk');

Execute:
select decompress_chunk(i) FROM show_chunks('sensor_data', older_than => '8 days'::INTERVAL) i;

Above query will uncompress _hyper_1_1_chunk and proceed to next chunk. Since next chunk is already decompressed, this query will error out. Thus we see an increase in size of _hyper_1_1_chunk.

    chunk_name    | total_size
------------------+------------
 _hyper_1_1_chunk | 54 MB <== this is increased.
 _hyper_1_2_chunk | 80 MB
 _hyper_1_3_chunk | 20 MB
 _hyper_1_4_chunk | 87 MB

While uncompressing _hyper_1_1_chunk  lot of stale data is left which will be cleaned up when VACUUM is run on this chunk.

Proposed patch:
=============
Consider a scenario where customer has 100 compressed chunks (each of 10GB or more) with few (1 or 2) uncompressed chunks, then select decompress_chunk(i) ...
query will be doing a lot of work trying to decompress all 100 chunks and when it encounters the last chunk which is not compressed, this query will error out, wasting all the time spent on decompressing.
We can overcome this issue by checking if the chunk being decompressed is compressed or not.
However customer needs to explicitly specify if_compressed argument.
This patch sets the default value of this 2nd argument to true, which saves a lot of time in these type of scenarios.


konskov:
This patch solves the problem mentioned in the description. Can we also add some tests to show that it does exactly that though? I mean something like the reproduction case where we check the chunk sizes before and after decompressing

sql/maintenance_utils.sql
```diff
@@ -39,7 +39,7 @@ CREATE OR REPLACE FUNCTION @extschema@.compress_chunk(
 
 CREATE OR REPLACE FUNCTION @extschema@.decompress_chunk(
     uncompressed_chunk REGCLASS,
-    if_compressed BOOLEAN = false
```
mfundul, line 42: pgspot complains about the redefinition of this function. You need to add this change to the upgrade/downgrade scripts as well.

svenklemm (changes requested):
I dont see how this change fixes the reported issue.

sb230132:
> I dont see how this change fixes the reported issue.

Stale data which is generated as part of decompression is not cleaned up. As a result querying compression metrics tend to report wrong values. Compression metrics will be updated to correct values only after VACCUM command is scheduled at some future point in time.

What the fix does is to skip all decompressed chunks, thus allowing the transaction to pass. This way query not only succeeds but also reflects catalog tables and compression metrics to be correct.

svenklemm:
The ticket is about dead tuples from failing decompress runs. How would changing the default trigger a vacuum run for these chunks? The scenario you describe is not relevant for the vacuum case because if the chunk is not in the right state it will not be decompressed and therefore not produce any dead tuples that need maintenance. This is about other classes of errors that lead to the decompression failing.

sb230132:
> The ticket is about dead tuples from failing decompress runs. How would changing the default trigger a vacuum run for these chunks? The scenario you describe is not relevant for the vacuum case because if the chunk is not in the right state it will not be decompressed and therefore not produce any dead tuples that need maintenance. This is about other classes of errors that lead to the decompression failing.

This is the ticket iam referring to. 
https://github.com/timescale/timescaledb/issues/3376

Ticket talks about chunk size increase continuously while the status still remain as 'compressed' when 
`select decompress_chunk(i) FROM show_chunks('<hypertable>') i` fails in the middle because there was a chunk which was already decompressed.

Current fix does not trigger VACUUM command. Current fix only skips chunks if it is already decompressed.

sb230132:
> 

Ok, i will add testcases too.

konskov:
@sb230132 @svenklemm apologies for my previous review comment, it was misleading. The size increase when the argument  `if_compressed` of `decompress_chunk` is false, will still occur. So the current PR is rather a workaround and not a fix for the linked issue.

sb230132:
> @sb230132 @svenklemm apologies for my previous review comment, it was misleading. The size increase when the argument `if_compressed` of `decompress_chunk` is false, will still occur. So the current PR is rather a workaround and not a fix for the linked issue.

@konskov,
You are right. Iam making the workaround a default behaviour as part of this patch so that user need not explicitly specify the 2nd argument. Only advantage with this patch is in below case:
A hypertable containing many compressed chunks and 1 or 2 decompressed, without my patch all the work of decompressing the chunks will be a waste, thus wasting lot of cpu and memory. My patch avoids this scenario.

konskov:
I also think that this default behavior is preferable. As this is not a fix though for the original issue, it would be good to remove the `Fixes #3376` from the commit message. And also squash all commits into a single one before merging.

mfundul (changes requested):
I'm with @svenklemm on this one. This PR does not fix the issue, and the API change is probably not what users expect, when compared to all other similar functions that behave this way. The workaround should probably be users being explicit with `if_compressed`. I think we need a proper fix for this one.

sb230132:
> I'm with @svenklemm on this one. This PR does not fix the issue, and the API change is probably not what users expect, when compared to all other similar functions that behave this way. The workaround should probably be users being explicit with `if_compressed`. I think we need a proper fix for this one.


Ok.
Side effects of not cleaning stale data (dead rows) causes compression metrics to go wrong. After decompressing each chunk we enable AUTO VACUUM to true for that chunk. So ideally these dead rows will get cleaned up at some future point in time.

A similar issue is raised in this PR: https://github.com/timescale/timescaledb/issues/4578
We close 1 of the PRs as a duplicate and track only one.

IMHO this is the right behaviour if we consider performance of a query. Way to fix it is to have a setting which when enabled should cleanup dead tuples immediately after decompression of each chunk.

sb230132:
Hi @svenklemm @konskov @mfundul 

I can reproduce a similar issue on vanilla postgres without having timescaledb extension.

Create a table with UNIQUE CONSTRAINT
```
test=# create table sensor_data(
test(#   time timestamptz not null,
test(#   sensor_id integer not null,
test(#   cpu double precision null,
test(#   temperature double precision null,
test(#   UNIQUE(time)
test(# );
CREATE TABLE
test=# 
```

Now check size of table

```
test=# select pg_relation_size('sensor_data');
 pg_relation_size 
------------------
                0
(1 row)
```

Insert into table with UNIQUE CONSTRAINT violation
```
test=# 
test=# INSERT INTO sensor_data SELECT time + (INTERVAL '1 minute' * random()) AS time, sensor_id, random() AS cpu, random()* 100 AS temperature
test-#     FROM generate_series(now() - INTERVAL '1 months', now() - INTERVAL '1 week', INTERVAL '1 minute') AS g1(time), generate_series(1, 100, 1 ) AS g2(sensor_id) ORDER BY time;

ERROR:  duplicate key value violates unique constraint "sensor_data_time_key"
DETAIL:  Key ("time")=(2022-08-18 06:16:30.900106+05:30) already exists.
test=# 
```

Table has 0 records.
```
test=# select count(*) from sensor_data;
 count 
-------
     0
(1 row)
```

Check table size:
```
test=# select pg_relation_size('sensor_data');
 pg_relation_size 
------------------
        100900864
(1 row)
```

Run VACUUM to clean up dead tuples
```
test=# vacuum "sensor_data";
VACUUM
```

Check size of table.
```
test=# select pg_relation_size('sensor_data');
 pg_relation_size 
------------------
                0
(1 row)
```

Coming to original PR issue: Decompression of chunk produces lot of dead tuples, which is not cleaned up. And thus when we query size of compressed chunks, we see increase in size.

We either document this behaviour when user executes:
```
select decompress_chunk(d) FROM show_chunks('sensor_data') d;
```

OR 

change interface of decompress_chunk() API to avoid such inconsistency.

sb230132:
> You might wanna read up on how MVCC and autovacuum works. The difference between the postgres example and the timescaledb example is that the stale data in the postgres table will get cleaned up the next time the autovacuum worker processes it which will never happen for the compressed hypertable because we explicitly disable autovacuum.

Hi @svenklemm 

Agreed.
What if user queries for size of postgresql table, before autovacuum worker process gets triggered. I think he will see wrong results. So can we assume this as a bug as well ?

sb230132:
Closing this PR since its a duplicate of #4578

PR #4659, opened 2022-08-31
Fix upgrade from 1.6.0 to later vesions.

### LINKED ISSUE #3114: Cascade Delete Fails on Compressed Hypertable After Upgrade From 1.6.1 to 2.1.0 (pg11)
## System Information

 - OS: Debian GNU/Linux 10 (buster)
 - PostgreSQL version: 11.11
 - TimescaleDB version: 2.1.0

## Description

We use a _compressed_ Timescale hypertable that references other tables, with foreign key constraints configured to `ON DELETE CASCADE`. The cascade delete used to work normally on `1.6.1`, but stopped working after migrating to `2.1.0`.

A workaround is to decompress and recompress all chunks, after which cascade delete starts working normally again.

It seems that `1.6.1` and `2.1.0` differ in how they deal with foreign key constraints on compressed chunks; `1.6.1` keeps the constraints on the normal chunks after creating the corresponding compressed ones, while `2.1.0` removes the constraints from original chunks after creating the compressed versions.

The migration to `2.1.0` does not remove the foreign constraints added by `1.6.1`, and seems to add additional checks that get triggered on cascade delete, resulting in errors.

## Reproduction

Run a new TimescaleDB container version 1.6.1 with PostgreSQL 11:
```
docker run -d --name timescaledb -e POSTGRES_PASSWORD=password \
        timescale/timescaledb:1.6.1-pg11
```

Wait a few seconds for the database to boot, then open a psql shell:
```
docker exec -it timescaledb psql -U postgres -X
```

Create a normal table `users`, and a hybertable `events`. The hypertable references the normal table, with a foreign key constraint configured to `ON DELETE CASCADE`:
```sql
CREATE TABLE users (id UUID PRIMARY KEY, name TEXT);

CREATE TABLE events (
        timestamp TIMESTAMPTZ not null,
        user_id UUID NOT NULL,
        value INTEGER NOT NULL,
        CONSTRAINT fk_user FOREIGN KEY(user_id)
        REFERENCES users(id) ON DELETE CASCADE
);

SELECT create_hypertable('events', 'timestamp');
```

Insert a few users and events:
```sql
INSERT INTO users VALUES
        ('332d6ce3-c4d5-4c88-9cd7-2b56c2ae8bef', 'Sam Fisher'),
        ('b0142cd9-e197-4182-8f6f-4de4e20d8ba3', 'Bob Ross');

INSERT INTO events VALUES
        ('2021-04-08T00:00:00Z', '332d6ce3-c4d5-4c88-9cd7-2b56c2ae8bef', 25),
        ('2021-04-08T00:00:00Z', 'b0142cd9-e197-4182-8f6f-4de4e20d8ba3', 30);

SELECT * FROM users;
SELECT * FROM events;
```

Enable compression on the hybertable:
```sql
ALTER TABLE "events" SET (
        timescaledb.compress,
        timescaledb.compress_segmentby = 'user_id'
);
```

Manually compress the first chunk:
```sql
SELECT compress_chunk('_timescaledb_internal._hyper_1_1_chunk');
```

As a control test, attempt to delete one of the users, notice how the related events got cascade deleted correctly:
```sql
DELETE FROM users WHERE id = 'b0142cd9-e197-4182-8f6f-4de4e20d8ba3';
SELECT * FROM users;
SELECT * FROM events;
```

Check the chunk's details; run:
```
\d+ _timescaledb_internal._hyper_1_1_chunk
```

Notice that despite compressing this chunk, it still has its foreign constraints. You will see:
```
Foreign-key constraints:
    "1_1_fk_user" FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE
```

Now close the shell:
```
\q
```

Migrate to the new `2.1.0` version of TimescaleDB. To do that, first run a container with the new docker image, using the same data volume:
```sh
export TIMESCALE_VOLUME=$(docker inspect timescaledb \
        --format='{{range .Mounts }}{{.Name}}{{end}}')

echo $TIMESCALE_VOLUME

docker stop timescaledb
docker rm timescaledb

docker run -d --name timescaledb -e POSTGRES_PASSWORD=password \
        -v $TIMESCALE_VOLUME:/var/lib/postgresql/data \
        timescale/timescaledb:2.1.0-pg11
```

Then run a psql shell:
```sh
docker exec -it timescaledb psql -U postgres -X
```

Upgrade the extension:
```sql
ALTER EXTENSION timescaledb UPDATE;
```

And check the extension version to make sure the update succeeded:
```
\dx timescaledb
```

Notice that the upgrade did not change the constraints on the chunk:
```
\d+ _timescaledb_internal._hyper_1_1_chunk
```

It still says:
```
Foreign-key constraints:
    "1_1_fk_user" FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE
```

Attempt to delete the second user:
```sql
DELETE FROM users WHERE id = '332d6ce3-c4d5-4c88-9cd7-2b56c2ae8bef';
```

This will fail with the error:
```
ERROR:  cannot update/delete rows from chunk "_hyper_1_1_chunk" as it is compressed
CONTEXT:  SQL statement "DELETE FROM ONLY "_timescaledb_internal"."_hyper_1_1_chunk" WHERE $1 OPERATOR(pg_catalog.=) "user_id""
```

No users or events were deleted:
```sql
SELECT * FROM users;
SELECT * FROM events;
```

To fix the constraints on this chunk, decompress it, and then recompress it again:
```sql
SELECT decompress_chunk('_timescaledb_internal._hyper_1_1_chunk');
SELECT compress_chunk('_timescaledb_internal._hyper_1_1_chunk');
```

Notice that the constraint on the chunk was removed:

```
\d+ _timescaledb_internal._hyper_1_1_chunk
```

Now reattempt user deletion:
```sql
DELETE FROM users WHERE id = '332d6ce3-c4d5-4c88-9cd7-2b56c2ae8bef';
```

It worked; the user was deleted, and all related events.
```sql
SELECT * FROM users;
SELECT * FROM events;
```

## Expected Behavior

Cascade deletion on compressed hybertables, working on `1.6.1`, should continue to work after upgrading to `2.1.0` without manual user intervention. Or, some warnings or instruction should be added to draw attention to this regression, and the available workaround.

## Actual Behavior

Cascade deletion on hybertables stops working after upgrade to `2.1.0`.

## Workaround

Decompress and recompress all chunks. Cascade delete should start working normally again.

## Additional Context

Kayisoft runs a production deployment of Timescale serving more than 10,000 concurrent users all around the world, with a number of rows approaching one billion. Recompressing all chunks is expected to take more than 5 hours on our current hardware. We still haven't upgraded our production deployment to `2.1.0`, and we'd rather have a way to upgrade without losing the ability to delete, and degrading service performance for our users for hours on end.

sb230132:
Compressed hypertables have FK constraints, which causes DELETE query on hypertable
to fail. This patch fixes upgrade script which will delete all FK constraints
to be dropped on compressed chunks if present.

Fixes #3114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ accidentally triggering the load of a previous DB version.**
 * #4575 Fix use of `get_partition_hash` and `get_partition_for_key` inside an IMMUTABLE function
 * #4611 Fix a potential OOM when loading large data sets into a hypertable
 * #4646 Fix time_bucket_ng origin handling
+* #3114 Fix upgrade from 1.6.0 to later by dropping FK constraints from compressed hypertable
 
 **Thanks**
 @AlmiS for reporting error on `get_partition_hash` executed inside an IMMUTABLE function
@@ -28,6 +29,7 @@ accidentally triggering the load of a previous DB version.**
 @michaelkitson for reporting permission errors using default privileges on Continuous Aggregates
 @ninjaltd and @mrksngl for reporting a potential OOM when loading large data sets into a hypertable
 @ssmoss for reporting an issue with time_bucket_ng origin handling
+@Mohammad-matini for reporting an upgrade issue from 1.6.1 to later versions
 
 ## 2.7.2 (2022-07-26)
 
--- a/sql/updates/post-update.sql
+++ b/sql/updates/post-update.sql
@@ -132,3 +132,34 @@ FROM _timescaledb_catalog.hypertable h
 INNER JOIN _timescaledb_catalog.dimension d ON (d.hypertable_id = h.id)
 WHERE d.interval_length IS NULL; 
 DROP FUNCTION _timescaledb_internal.update_dimension_partition;
+
+-- upgrade from 1.6.0 to later requires FK constraints to be dropped
+DO $$
+DECLARE
+ fk_constraint_row RECORD;
+BEGIN
+  FOR fk_constraint_row IN
+    SELECT chunk.schema_name AS schema_name, chunk.table_name AS table_name,
+       chunk_con.constraint_name AS constraint_name
+    FROM _timescaledb_catalog.chunk_constraint chunk_con
+       INNER JOIN _timescaledb_catalog.chunk chunk ON (
+          chunk_con.chunk_id = chunk.id
+          AND chunk.compressed_chunk_id <> 0)
+       INNER JOIN _timescaledb_catalog.hypertable hypertable ON
+          (chunk.hypertable_id = hypertable.id)
+       INNER JOIN pg_constraint pg_chunk_con ON (
+          pg_chunk_con.conrelid = format('%I.%I', chunk.schema_name, chunk.table_name)::regclass
+          AND pg_chunk_con.conname = chunk_con.constraint_name
+          AND pg_chunk_con.contype = 'f'
+          )
+    INNER JOIN pg_class pg_chunk_index_class ON (
+            pg_chunk_con.conindid = pg_chunk_index_class.oid
+    )
+    LOOP
+        EXECUTE format('ALTER TABLE %I.%I DROP CONSTRAINT %I',
+            fk_constraint_row.schema_name,
+            fk_constraint_row.table_name,
+            fk_constraint_row.constraint_name);
+    END LOOP;
+END
+$$;


mfundul (changes requested):
You can't do this kind of fix in `latest-dev.sql`, since this will just become `2.7.2--2.8.0.sql`. The correct place would be `post-update.sql`

sb230132:
In All DB engineering meeting, this bug has been discussed, and plan is to close the PR.

PR #4667, opened 2022-09-01
Add compat macros for PG15 signature changes

svenklemm:
PG15 refactors INSERT/UPDATE/DELETE code and changes the signatures
of some of the functions.

https://github.com/postgres/postgres/commit/25e777cf

### DIFF
--- a/src/compat/compat.h
+++ b/src/compat/compat.h
@@ -624,7 +624,92 @@ pg_strtoint64(const char *str)
 						  multixact_freeze_table_age,                                              \
 						  oldestXmin,                                                              \
 						  freezeLimit,                                                             \
-						  multiXactCutoff)
+						  multiXactCutoff,                                                         \
+						  NULL)
+#endif
+
+#if PG15_LT
+#define ExecARUpdateTriggersCompat(estate,                                                         \
+								   resultRelInfo,                                                  \
+								   src_partinfo,                                                   \
+								   dst_partinfo,                                                   \
+								   tupleid,                                                        \
+								   oldtuple,                                                       \
+								   inewslot,                                                       \
+								   recheckIndexes,                                                 \
+								   transtition_capture,                                            \
+								   is_crosspart_update)                                            \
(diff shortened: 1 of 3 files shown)


svenklemm:
With this applied timescaledb will compile against PG15

src/compat/compat.h
```diff
@@ -624,7 +624,92 @@ pg_strtoint64(const char *str)
...
+#define ExecARUpdateTriggersCompat(estate,                                                         \
+								   resultRelInfo,                                                  \
+								   src_partinfo,                                                   \
+								   dst_partinfo,                                                   \
+								   tupleid,                                                        \
+								   oldtuple,                                                       \
+								   inewslot,                                                       \
+								   recheckIndexes,                                                 \
+								   transtition_capture,                                            \
+								   is_crosspart_update)                                            \
```
fabriziomello, line 659: In our code we always set this parameter as `false` so we can simplify this signature removing the `is_crosspart_update` and set `false` for prior pg15 version.

src/compat/compat.h
```diff
@@ -624,7 +624,92 @@ pg_strtoint64(const char *str)
...
+#endif
+
+#if PG15_LT
+#define ExecBRUpdateTriggersCompat(estate,                                                         \
+								   epqstate,                                                       \
+								   resultRelInfo,                                                  \
+								   tupleid,                                                        \
+								   oldtuple,                                                       \
+								   slot,                                                           \
+								   tmfdp)                                                          \
```
fabriziomello, line 679: Same here for `tmfdp` that for prior pg15 version we pass `NULL`

fabriziomello (approved):

src/compat/compat.h
svenklemm: With these compat macros we usually adopt the signature of the latest version to create the least friction moving forward and also to make it easy to replace the macro once we no longer need to support the older versions that required the macro.

konskov (approved):

PR #4668, opened 2022-09-01
Hierarchical Continuous Aggregates

### LINKED ISSUE #1400: Continuous aggregate on top of another continuous aggregate
I'm trying to create a continuous aggregate on top of another continuous aggregate and gets `invalid SELECT query for continuous aggregate`. Is this not supported? I can't find any information on this. See below example where the first query works find and the second fails. This is a very important feature, since it allows me to make a weighted average.

 ```
 create view readings_5min
  with (timescaledb.continuous)
  as
  select 
  time_bucket('5 minutes',time) as time,
  "serial",
  avg(value) as mean,
  from readings_raw
  group by time_bucket('5 minutes',time),"serial"
    
  create view readings_hour
  with (timescaledb.continuous)
  as
  select 
  time_bucket('1 hour',time) as time,
  "serial",
  avg(mean) as mean,
  from readings_5min
  group by time_bucket('1 hour',time),"serial"
```

### PR DESCRIPTION
**Pull request by fabriziomello**
Enable users create Hierarchical Continuous Aggregates (aka Continuous
Aggregates on top of another Continuous Aggregates).

With this PR users can create levels of aggregation granularity in
Continuous Aggregates making the refresh process even faster.

A problem with this feature can be in upper levels we can end up with
the "average of averages". But to get the "real average" we can rely on
"stats_aggs" TimescaleDB Toolkit function that calculate and store the
partials that can be finalized with other toolkit functions like
"average" and "sum".

Closes #1400

### DIFF
--- a/src/ts_catalog/continuous_agg.c
+++ b/src/ts_catalog/continuous_agg.c
@@ -288,6 +288,14 @@ continuous_agg_formdata_make_tuple(const FormData_continuous_agg *fd, TupleDesc 
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_raw_hypertable_id)] =
 		Int32GetDatum(fd->raw_hypertable_id);
 
+	if (fd->parent_mat_hypertable_id == INVALID_HYPERTABLE_ID)
+		nulls[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] = true;
+	else
+	{
+		values[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] =
+			Int32GetDatum(fd->parent_mat_hypertable_id);
+	}
+
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_user_view_schema)] =
 		NameGetDatum(&fd->user_view_schema);
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_user_view_name)] =
@@ -328,6 +336,12 @@ continuous_agg_formdata_fill(FormData_continuous_agg *fd, const TupleInfo *ti)
 		DatumGetInt32(values[AttrNumberGetAttrOffset(Anum_continuous_agg_mat_hypertable_id)]);
 	fd->raw_hypertable_id =
 		DatumGetInt32(values[AttrNumberGetAttrOffset(Anum_continuous_agg_raw_hypertable_id)]);
+
+	if (nulls[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)])
+		fd->parent_mat_hypertable_id = INVALID_HYPERTABLE_ID;
+	else
+		fd->parent_mat_hypertable_id = DatumGetInt32(
+			values[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)]);
 
 	memcpy(&fd->user_view_schema,
 		   DatumGetName(values[AttrNumberGetAttrOffset(Anum_continuous_agg_user_view_schema)]),
--- a/tsl/src/continuous_aggs/create.c
+++ b/tsl/src/continuous_aggs/create.c
@@ -151,10 +151,11 @@ typedef struct FinalizeQueryInfo
 
 typedef struct CAggTimebucketInfo
 {
-	int32 htid;				/* hypertable id */
-	Oid htoid;				/* hypertable oid */
-	AttrNumber htpartcolno; /* primary partitioning column of raw hypertable */
-							/* This should also be the column used by time_bucket */
+	int32 htid;						/* hypertable id */
+	int32 parent_mat_hypertable_id; /* parent materialization hypertable id */
+	Oid htoid;						/* hypertable oid */
+	AttrNumber htpartcolno;			/* primary partitioning column of raw hypertable */
+									/* This should also be the column used by time_bucket */
 	Oid htpartcoltype;
 	int64 htpartcol_interval_len; /* interval length setting for primary partitioning column */
 	int64 bucket_width;			  /* bucket_width of time_bucket, stores BUCKET_WIDHT_VARIABLE for
@@ -206,7 +207,8 @@ static Query *mattablecolumninfo_get_partial_select_query(MatTableColumnInfo *ma
 static void caggtimebucketinfo_init(CAggTimebucketInfo *src, int32 hypertable_id,
 									Oid hypertable_oid, AttrNumber hypertable_partition_colno,
 									Oid hypertable_partition_coltype,
-									int64 hypertable_partition_col_interval);
+									int64 hypertable_partition_col_interval,
+									int32 parent_mat_hypertable_id);
 static void caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause,
 									List *targetList);
 static void finalizequery_init(FinalizeQueryInfo *inp, Query *orig_query,
@@ -227,7 +229,8 @@ static void
 create_cagg_catalog_entry(int32 matht_id, int32 rawht_id, const char *user_schema,
 						  const char *user_view, const char *partial_schema,
 						  const char *partial_view, int64 bucket_width, bool materialized_only,
-						  const char *direct_schema, const char *direct_view, const bool finalized)
+						  const char *direct_schema, const char *direct_view, const bool finalized,
+						  const int32 parent_mat_hypertable_id)
 {
 	Catalog *catalog = ts_catalog_get();
 	Relation rel;
@@ -249,6 +252,15 @@ create_cagg_catalog_entry(int32 matht_id, int32 rawht_id, const char *user_schem
 	memset(values, 0, sizeof(values));
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_mat_hypertable_id)] = matht_id;
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_raw_hypertable_id)] = rawht_id;
+
+	if (parent_mat_hypertable_id == INVALID_HYPERTABLE_ID)
+		nulls[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] = true;
+	else
+	{
+		values[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] =
+			parent_mat_hypertable_id;
+	}
+
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_user_view_schema)] =
 		NameGetDatum(&user_schnm);
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_user_view_name)] =
@@ -678,9 +690,10 @@ create_view_for_query(Query *selquery, RangeVar *viewrel)
 static void
 caggtimebucketinfo_init(CAggTimebucketInfo *src, int32 hypertable_id, Oid hypertable_oid,
 						AttrNumber hypertable_partition_colno, Oid hypertable_partition_coltype,
-						int64 hypertable_partition_col_interval)
+						int64 hypertable_partition_col_interval, int32 parent_mat_hypertable_id)
 {
 	src->htid = hypertable_id;
+	src->parent_mat_hypertable_id = parent_mat_hypertable_id;
 	src->htoid = hypertable_oid;
 	src->htpartcolno = hypertable_partition_colno;
 	src->htpartcoltype = hypertable_partition_coltype;
@@ -957,7 +970,7 @@ cagg_agg_validate(Node *node, void *context)
  *   added.
  */
 static bool
-cagg_query_supported(Query *query, StringInfo hint, StringInfo detail, const bool finalized)
+cagg_query_supported(const Query *query, StringInfo hint, StringInfo detail, const bool finalized)
 {
 /*
  * For now deprecate partial aggregates on release builds only.
@@ -1079,7 +1092,7 @@ cagg_query_supported(Query *query, StringInfo hint, StringInfo detail, const boo
 }
 
 static CAggTimebucketInfo
-cagg_validate_query(Query *query, bool finalized)
+cagg_validate_query(const Query *query, const bool finalized)
 {
 	CAggTimebucketInfo ret;
 	Cache *hcache;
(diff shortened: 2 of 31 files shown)

### CODE REVIEW

**Comment by Inviz** [comment]
Yeyyy

**Comment by Inviz** [comment]
It's amazing that the actual code has a pretty small surface area for this feature.

File: tsl/src/continuous_aggs/create.c [hunk 1]
```diff
@@ -1097,38 +1097,61 @@ cagg_validate_query(Query *query, bool finalized)
...
+		{
+			const ContinuousAgg *cagg;
+
+			cagg = ts_continuous_agg_find_by_relid(rte->relid);
+
+			if (!ContinuousAggIsFinalized(cagg))
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("not allowed old format of continuous aggregate")));
```
gayyappan, line 1126:
```suggestion
						 errmsg("old format of continuous aggregate is not supported for this feature")));
```

tsl/src/continuous_aggs/create.c
```diff
@@ -1097,38 +1097,61 @@ cagg_validate_query(Query *query, bool finalized)
...
-				ereport(ERROR,
-						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-						 errmsg("hypertable is a continuous aggregate materialization table")));
-			case HypertableIsRawTable:
-				break;
-			case HypertableIsNotContinuousAgg:
-				break;
-			default:
-				Assert(false);
+			/* there can only be one continuous aggregate per table */
```
gayyappan, line 1140: this comment is incorrect?  should be removed

tsl/src/continuous_aggs/create.c
```diff
@@ -1097,38 +1097,61 @@ cagg_validate_query(Query *query, bool finalized)
...
+			switch (ts_continuous_agg_hypertable_status(ht->fd.id))
```
gayyappan, line 1141: by the way, why do we need this check at all? don't we support all these cases with this new feature?

tsl/src/continuous_aggs/create.c
gayyappan: maybe a errhint descirbing how to migrate to new format is useful here?

tsl/src/continuous_aggs/create.c
fabriziomello: This comment is a bit wrong because the tests have another purpose

tsl/src/continuous_aggs/create.c
fabriziomello: we need this test to prevent users create CAGG over an existing materialization hypertable. See: https://github.com/timescale/timescaledb/blob/main/tsl/test/expected/cagg_ddl.out#L467-L476

src/ts_catalog/continuous_agg.c
```diff
@@ -288,6 +288,14 @@ continuous_agg_formdata_make_tuple(const FormData_continuous_agg *fd, TupleDesc
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_raw_hypertable_id)] =
 		Int32GetDatum(fd->raw_hypertable_id);
 
+	if (fd->parent_mat_hypertable_id == INVALID_CHUNK_ID)
```
gayyappan, line 291:
```suggestion
	if (fd->parent_mat_hypertable_id == INVALID_HYPERTABLE_ID)
```

src/ts_catalog/continuous_agg.c
```diff
@@ -329,6 +337,12 @@ continuous_agg_formdata_fill(FormData_continuous_agg *fd, const TupleInfo *ti)
 	fd->raw_hypertable_id =
 		DatumGetInt32(values[AttrNumberGetAttrOffset(Anum_continuous_agg_raw_hypertable_id)]);
 
+	if (nulls[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)])
+		fd->parent_mat_hypertable_id = INVALID_CHUNK_ID;
```
gayyappan, line 341: INVALID_HYPERTABLE_ID

tsl/src/continuous_aggs/create.c
```diff
@@ -1097,38 +1110,63 @@ cagg_validate_query(Query *query, bool finalized)
...
+								"feature"),
+						 errhint("Run \"CALL cagg_migrate('%s.%s');\" to migrate to the new "
+								 "format.",
+								 NameStr(cagg->data.user_view_schema),
+								 NameStr(cagg->data.user_view_name))));
+			}
+
+			parent_mat_hypertable_id = cagg->data.mat_hypertable_id;
+			hcache = ts_hypertable_cache_pin();
+			ht = ts_hypertable_cache_get_entry_by_id(hcache, cagg->data.mat_hypertable_id);
```
gayyappan, line 1150: Don't you need a check that the cagg's time_bucket interval is a multiple of the interval of the original cagg.
i.e. if the first cagg has time_bucket 15 minutes,
cagg on cagg cannot have time_bucket of 20 minutes

src/ts_catalog/continuous_agg.c
fabriziomello: Fixed

src/ts_catalog/continuous_agg.c
fabriziomello: Fixed

tsl/src/continuous_aggs/create.c
fabriziomello: You're correct... will take a note in this PR about it and another validations as we discussed in PVT.

fabriziomello:
As per discussed in pvt with @gayyappan this PR lack of some important validations:

1. bucket width of the nested cagg should be multiple of the parent cagg bucket width
2. prevent users create a real-time cagg over a materialized only cagg
3. when change a cagg from materialized only to real-time propagate this change to all nested caggs
4. prevent users create a nested cagg with bucket width less than the parent cagg bucket width

/cc @horzsolt @iroussos

davidkohn88:
While I appreciate the concerns from @fabriziomello and @gayyappan, and I agree, these are weird edge cases, I think that these checks are entirely unnecessary and potentially counterproductive and certainly will slow down our development. I'd suggest that we can document some of these as potential issues, but I don't think we should spend time trying to prevent odd cases that any user can run into no matter what. 

My basic philosophy on this is that these views should be treated essentially as separate entities and the behavior should be predictable based on that, we should not emphasize / interlink or try to plan for all the weird ways that someone can cause them to interact weirdly. Users can already do this with a normal view, and we don't stop them there, that's on them, and it should be here too. 

> As per discussed in pvt with @gayyappan this PR lack of some important validations:
> 
> 1. bucket width of the nested cagg should be multiple of the parent cagg bucket width
A) I'm not sure that this is possible or reasonable in a number of cases, for instance, can we actually enforce this with anything involving time zones? What's the test? Is a week a multiple of a day? What about a month being a multiple of 7 minutes? 
B) fundamentally, a time_bucket produces a new timestamp. If you want to re-bucket that imperfectly, I don't think we should stop you. There may be many perfectly valid reasons why a user would want to do that, including that for many, it's "close enough" ie the 7 minutes into a month thing...I'd probably be okay with bucketing it even if technically there was a little slosh at the edges, it'd even out over time. We don't stop users from doing this in queries now, why would we stop them here? 
C) This doesn't impact invalidations at all. The underlying table gets invalidations, then the first materialization gets invalidations, and they will all propagate properly up the chain expanding to the size of the bucket (assuming we've done our work correctly), so there's no special case where you're going to be "more wrong" here than if you just do this in a query. 
D) I *do* think that we should have a note in the docs saying something like: `We recommend that smaller buckets divide evenly into larger buckets` give a demonstration of why this could be a problem and then say, but we do not enforce this, it is up to you as the user to understand the implications of doing or not doing this. 
> 2. prevent users create a real-time cagg over a materialized only cagg
Why would we prevent this? There are a number of valid use cases where I would totally want the behavior that this provides. For instance, I have a big hypertable that I know isn't that performant if I do real time agg on it, so I create a frequently refreshed, say, 1 minute cagg on it, and then I do larger bucketed caggs over that. I *want* the real time view for those larger bucketed caggs because I want to see the new results ASAP once the 1 minute view gets refreshed, but I know that if I hit the hypertable my performance suffers too much, so I create a real time view on the larger bucketed ones and keep the 1 minute one materialized only. 
> 3. when change a cagg from materialized only to real-time propagate this change to all nested caggs
This would be a horrible, horrible user experience for anyone in the situation above, and I don't know why we would do this. 

It seems far, far easier for a user to reason about this on a per view basis rather than have side effects. The views are functionally independent, all we're doing is creating another cagg on top of the view. That is far, far easier to understand than if I start creating implicit linkages. 
> 4. prevent users create a nested cagg with bucket width less than the parent cagg bucket width
Why? because it's silly? I agree it's silly, but like, we can't keep everyone from doing all the silly things, that's too much work. If you do this, you'll find that it has the same number of rows as the original cagg. Okay. What's the big deal? Are they going to lose data or will something horrible happen?

My basic philosophy on most of this is that we should avoid doing too much to prevent users from shooting themselves in the foot, at least when the injury will not be that grave. None of the scenarios here will cause data loss or irrecoverable failures, so why spend the time both constraining user behavior unnecessarily (which means they can't make tradeoffs that we might not have considered) and also slowing down our development considerably. Trying to prevent all this stuff means so much more testing and development on our end and I really don't think we're going to cover all the cases. And the benefit to the user is so minimal.

mkindahl (commented):
Some minor comments, but in general I think it looks good.

tsl/src/continuous_aggs/create.c
```diff
@@ -152,6 +152,7 @@ typedef struct FinalizeQueryInfo
 typedef struct CAggTimebucketInfo
 {
 	int32 htid;				/* hypertable id */
+	int32 parent_mthtid;	/* parent materialization hypertable id */
```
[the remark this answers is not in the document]
fabriziomello: Fixed.

tsl/src/continuous_aggs/create.c
```diff
@@ -1097,38 +1110,63 @@ cagg_validate_query(Query *query, bool finalized)
...
 
-		ht = ts_hypertable_cache_get_cache_and_entry(rte->relid, CACHE_FLAG_NONE, &hcache);
+			cagg = ts_continuous_agg_find_by_relid(rte->relid);
+
+			if (!ContinuousAggIsFinalized(cagg))
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("old format of continuous aggregate is not supported for this "
+								"feature"),
```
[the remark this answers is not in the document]
fabriziomello: Fixed.

tsl/src/continuous_aggs/create.c
```diff
@@ -2248,7 +2288,8 @@ cagg_create(const CreateTableAsStmt *create_stmt, ViewStmt *stmt, Query *panquer
 							  materialized_only,
 							  dum_rel->schemaname,
 							  dum_rel->relname,
-							  finalized);
+							  finalized,
+							  origquery_ht->parent_mthtid);
```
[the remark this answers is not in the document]
fabriziomello: Fixed!

tsl/src/continuous_aggs/create.c
```diff
@@ -1097,38 +1110,63 @@ cagg_validate_query(Query *query, bool finalized)
...
-			case HypertableIsMaterializationAndRaw:
+			ContinuousAggHypertableStatus status = ts_continuous_agg_hypertable_status(ht->fd.id);
+
+			/* prevent create a CAGG over an existing materialization hypertable */
+			if (status == HypertableIsMaterialization ||
+				status == HypertableIsMaterializationAndRaw)
+			{
 				ereport(ERROR,
 						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
 						 errmsg("hypertable is a continuous aggregate materialization table")));
```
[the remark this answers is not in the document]
fabriziomello: Done!

tsl/src/continuous_aggs/create.c
```diff
@@ -1097,38 +1110,66 @@ cagg_validate_query(Query *query, bool finalized)
...
-				break;
-			case HypertableIsNotContinuousAgg:
-				break;
-			default:
-				Assert(false);
+						 errmsg("hypertable is a continuous aggregate materialization table"),
+						 errhint("An hypertable or a continuous aggregate should be used instead "
+								 "of the materialization hypertable \"%s.%s\".",
+								 NameStr(ht->fd.schema_name),
+								 NameStr(ht->fd.table_name))));
```
mkindahl, line 1171: You can find the continuous aggregate for the materialization table, so I would suggest something along the lines:
```suggestion
						 errhint("Do you want to use continuous aggregate \"%s.%s\" instead?",
								 NameStr(schema_name),
								 NameStr(table_name))));
```
And then an `errdetail` with the materialization table name.

mkindahl (approved):
Looks good. Some minor comments, but approving since they are not critical.

tsl/src/continuous_aggs/create.c
```diff
@@ -151,10 +151,11 @@ typedef struct FinalizeQueryInfo
...
 {
-	int32 htid;				/* hypertable id */
-	Oid htoid;				/* hypertable oid */
-	AttrNumber htpartcolno; /* primary partitioning column of raw hypertable */
-							/* This should also be the column used by time_bucket */
+	int32 htid;						/* hypertable id */
+	int32 parent_mat_hypertable_id; /* parent materialization hypertable id */
+	Oid htoid;						/* hypertable oid */
+	AttrNumber htpartcolno;			/* primary partitioning column of raw hypertable */
+									/* This should also be the column used by time_bucket */
```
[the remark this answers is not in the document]
fabriziomello: I have my editor configured to remove trailing spaces (I guess everyone should do it) but in this case it was changed by clang-format.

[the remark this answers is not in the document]
mkindahl: Interesting. Anyway, just something to keep in mind. Not a blocker.

src/ts_catalog/continuous_agg.c
```diff
@@ -288,6 +288,14 @@ continuous_agg_formdata_make_tuple(const FormData_continuous_agg *fd, TupleDesc
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_raw_hypertable_id)] =
 		Int32GetDatum(fd->raw_hypertable_id);
 
+	if (fd->parent_mat_hypertable_id == INVALID_HYPERTABLE_ID)
+		nulls[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] = true;
+	else
+	{
+		values[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] =
+			Int32GetDatum(fd->parent_mat_hypertable_id);
+	}
```
mfundul, line 297:
```suggestion
		values[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] =
			Int32GetDatum(fd->parent_mat_hypertable_id);
```

mfundul: nit

tsl/src/continuous_aggs/create.c
```diff
@@ -249,6 +252,15 @@ create_cagg_catalog_entry(int32 matht_id, int32 rawht_id, const char *user_schem
...
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_mat_hypertable_id)] = matht_id;
 	values[AttrNumberGetAttrOffset(Anum_continuous_agg_raw_hypertable_id)] = rawht_id;
+
+	if (parent_mat_hypertable_id == INVALID_HYPERTABLE_ID)
+		nulls[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] = true;
+	else
+	{
+		values[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] =
+			parent_mat_hypertable_id;
+	}
```
mfundul, line 262:
```suggestion
		values[AttrNumberGetAttrOffset(Anum_continuous_agg_parent_mat_hypertable_id)] =
			parent_mat_hypertable_id;
```

fabriziomello: It can't be done because the INVALID_HYPERTABLE_ID is equal to 0 (zero) and I want to store NULL in the `parent_mat_hypertable_id` column

gayyappan (approved):
Changes look good.
would be good to have an additional test for this case: 
If we have cagg3->cagg2->cagg1 and DROP MAT VEW cagg2
then cagg3 and cagg1 are unaffected by this drop.

PR #4670, opened 2022-09-05
Add timezone support to time_bucket_gapfill

svenklemm:
This patch adds a new function time_bucket_gapfill function that
allows bucketing in a specific timezone. 

You can gapfill with explicit timezone like so:
`SELECT time_bucket_gapfill('1 day', time, 'Europe/Berlin') ...`

Unfortunately this introduces an ambiguity with some previous
call variations when an untyped start/finish argument was passed
to the function. Some queries might need to be adjusted and either
explicitly name the positional argument or resolve the type ambiguity
by casting to the intended type.

Disable-check: commit-count


tsl/src/nodes/gapfill/exec.c
```diff
@@ -139,13 +141,22 @@ gapfill_internal_get_datum(int64 value, Oid type)
...
 
-static inline int64
-interval_to_usec(Interval *interval)
+static Expr *
+get_start_arg(GapFillState *state)
+{
+	if (!state->have_timezone)
+		return lthird(state->args);
+	else
+		return lfourth(state->args);
```
fabriziomello, line 150:
```suggestion

	return lfourth(state->args);
```

Nit.

tsl/src/nodes/gapfill/exec.c
```diff
@@ -139,13 +141,22 @@ gapfill_internal_get_datum(int64 value, Oid type)
...
+get_finish_arg(GapFillState *state)
 {
-	if (interval == NULL)
-		ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR), errmsg("interval cannot be NULL")));
-	return (interval->month * DAYS_PER_MONTH * USECS_PER_DAY) + (interval->day * USECS_PER_DAY) +
-		   interval->time;
+	if (!state->have_timezone)
+		return lfourth(state->args);
+	else
+		return lfifth(state->args);
```
fabriziomello, line 159:
```suggestion

	return lfifth(state->args);
```

Nit.

tsl/src/nodes/gapfill/gapfill.c
```diff
@@ -39,3 +39,18 @@ GAPFILL_TIMEBUCKET_WRAPPER(int64);
...
+
+Datum
+gapfill_timestamptz_timezone_time_bucket(PG_FUNCTION_ARGS)
+{
+	/*
+	 * since time_bucket is STRICT and time_bucket_gapfill
+	 * is not we need to add explicit checks for NULL here
+	 */
+	if (PG_ARGISNULL(0) || PG_ARGISNULL(1) || PG_ARGISNULL(2))
+		PG_RETURN_NULL();
```
fabriziomello, line 51: Add some tests here.

jnidzwetzki (approved):
Aside from the open comments, the PR looks good to me.

PR #4673, opened 2022-09-05
Fix planner chunk exclusion for VIEWs

### LINKED ISSUE #4524: [Bug]:  Planner chunk exclusion not working in VIEWs
### What type of bug is this?

Performance issue

Query planner

### What happened?

When querying a view on a hypertable constraints on dimensions are not used for plan time chunk exclusion. 

main

### PostgreSQL version used

14.4

Archlinux 2022.07.01 x64

Source

On prem/Self-hosted

```bash
explain SELECT * from v2 where time > now() - '168h'::interval;
                                  QUERY PLAN
-------------------------------------------------------------------------------
 HashAggregate  (cost=4.10..4.15 rows=4 width=18)
   Group Key: metrics."time", metrics.device
   ->  Custom Scan (ChunkAppend) on metrics  (cost=0.00..4.07 rows=4 width=18)
         Chunks excluded during startup: 3
         ->  Seq Scan on _hyper_1_4_chunk  (cost=0.00..1.02 rows=1 width=18)
               Filter: ("time" > (now() - '168:00:00'::interval))
(6 rows)
```

```bash
create table metrics(time timestamptz,device text, value float);
SELECT create_hypertable('metrics','time');
INSERT INTO metrics select '2000-01-01','a',0.5;
INSERT INTO metrics select '2001-01-01','b',0.5;
INSERT INTO metrics select '2002-01-01','b',0.5;
INSERT INTO metrics select now(),'b',0.5;
CREATE VIEW v2 AS SELECT time, device, avg(value) from metrics group by 1,2;
EXPLAIN SELECT * from v2 where time > now() - '168h'::interval;
```

svenklemm:
Allow planner chunk exclusion in subqueries. When we decicde on
whether a query may benefit from constifying now and encounter a
subquery peek into the subquery and check if the constraint
references a hypertable partitioning column.

Fixes #4524

### DIFF
--- a/src/planner/constify_now.c
+++ b/src/planner/constify_now.c
@@ -36,9 +36,9 @@
  * Interval needs to be Const in those expressions.
  */
 static const Dimension *
-get_hypertable_dimension(Oid relid)
+get_hypertable_dimension(Oid relid, int flags)
 {
-	Hypertable *ht = ts_planner_get_hypertable(relid, CACHE_FLAG_CHECK);
+	Hypertable *ht = ts_planner_get_hypertable(relid, flags);
 	if (!ht)
 		return NULL;
 	return hyperspace_get_open_dimension(ht->space, 0);
@@ -47,6 +47,7 @@ get_hypertable_dimension(Oid relid)
 static bool
 is_valid_now_expr(OpExpr *op, List *rtable)
 {
+	int flags = CACHE_FLAG_MISSING_OK | CACHE_FLAG_NOCREATE;
 	/* Var > or Var >= */
 	if ((op->opfuncid != F_TIMESTAMPTZ_GT && op->opfuncid != F_TIMESTAMPTZ_GE) ||
 		!IsA(linitial(op->args), Var))
@@ -63,7 +64,29 @@ is_valid_now_expr(OpExpr *op, List *rtable)
 	Assert(var->varno <= list_length(rtable));
 	RangeTblEntry *rte = list_nth(rtable, var->varno - 1);
 
-	const Dimension *dim = get_hypertable_dimension(rte->relid);
+	/*
+	 * If this query on a view we might have a subquery here
+	 * and need to peek into the subquery range table to check
+	 * if the constraints are on a hypertable.
+	 */
+	if (rte->rtekind == RTE_SUBQUERY)
+	{
+		/*
+		 * Unfortunately the mechanism used to warm up the
+		 * hypertable cache does not apply to hypertables
+		 * referenced indirectly eg through VIEWs. So we
+		 * have to do the lookup for this hypertable without
+		 * CACHE_FLAG_NOCREATE flag.
+		 */
+		flags = CACHE_FLAG_MISSING_OK;
+		TargetEntry *tle = list_nth(rte->subquery->targetList, var->varattno - 1);
+		if (!IsA(tle->expr, Var))
+			return false;
+		var = castNode(Var, tle->expr);
+		rte = list_nth(rte->subquery->rtable, var->varno - 1);
+	}
+
+	const Dimension *dim = get_hypertable_dimension(rte->relid, flags);
 	if (!dim || dim->fd.column_type != TIMESTAMPTZOID || dim->column_attno != var->varattno)
 		return false;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,11 +19,12 @@ argument or resolve the type ambiguity by casting to the intended type.
 
 **Bugfixes**
 * #4619 Improve handling enum columns in compressed hypertables
+* #4673 Fix now() constification for VIEWs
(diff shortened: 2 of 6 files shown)


akuzm:
Is it possible to implement plan-time now() constification at the moment after the subqueries are pulled up, and the quals are assigned to relations, but before the inheritance hierarchy is expanded? So that it would look at baserestrictinfos instead. This ways it feels like we're doing it too early, and have to mirror the general subquery pullup logic for a very specific case.

svenklemm:
> Is it possible to implement plan-time now() constification at the moment after the subqueries are pulled up, and the quals are assigned to relations, but before the inheritance hierarchy is expanded? So that it would look at baserestrictinfos instead. This ways it feels like we're doing it too early, and have to mirror the general subquery pullup logic for a very specific case.

The problem is for the view case in the subquery we don't see the hypertable. For general subqueries this should all work fine as the normal logic should apply when processing the subquery.

akuzm:
> > The problem is for the view case in the subquery we don't see the hypertable.
> 
> That's because we're doing the now() consitifcation before subquery pullup. But why so early? We could do it before a hypertable is expanded. It's easier because we have to only examine the baserestrictinfos, the subqueries are already flattened, the constants are evaluated and so on.

I mean, the scope of now() constification could be as small as the function that calculates which part of hyperspace matches the baserestrictinfo. It would be enough to evaluate now() in that function and we wouldn't even need to add and delete the additional quals with evaluated now(). Am I missing something?

svenklemm:
> > > The problem is for the view case in the subquery we don't see the hypertable.
> > 
> > 
> > That's because we're doing the now() consitifcation before subquery pullup. But why so early? We could do it before a hypertable is expanded. It's easier because we have to only examine the baserestrictinfos, the subqueries are already flattened, the constants are evaluated and so on.
> 
> I mean, the scope of now() constification could be as small as the function that calculates which part of hyperspace matches the baserestrictinfo. It would be enough to evaluate now() in that function and we wouldn't even need to add and delete the additional quals with evaluated now(). Am I missing something?

This optimization is very time sensitive and has to happen at the earliest point to gain the maximum benefit. While it is true that we could delay the constification that would also mean we would have to give up on a lot of the optimization potential. In addition this optimization works for any query (including UPDATE/DELETE) and works whether our own hypertable expansion or postgres expansion is used.

src/planner/constify_now.c
```diff
@@ -63,7 +64,22 @@ is_valid_now_expr(OpExpr *op, List *rtable)
...
 
-	const Dimension *dim = get_hypertable_dimension(rte->relid);
+	/*
+	 * If this query on a view we might have a subquery here
+	 * and need to peek into the subquery range table to check
+	 * if the constraints are on a hypertable.
+	 */
+	if (rte->rtekind == RTE_SUBQUERY)
+	{
+		flags = CACHE_FLAG_MISSING_OK;
```
jnidzwetzki, line 74: Could you add a comment why the `CACHE_FLAG_NOCREATE` flag should not be set in this case?

jnidzwetzki:
Please don't forget to include the bugfix in the changelog as well.

jnidzwetzki (approved):

akuzm:
>  This optimization is very time sensitive and has to happen at the earliest point to gain the maximum benefit. While it is true that we could delay the constification that would also mean we would have to give up on a lot of the optimization potential. 

The main benefit comes from having less chunks at the planning stage, right? For this it has to happen before the inheritance expansion. Are there other benefits for which it must happen earlier? I'm asking just so that I have a full understanding of what's going on.

svenklemm:
> > This optimization is very time sensitive and has to happen at the earliest point to gain the maximum benefit. While it is true that we could delay the constification that would also mean we would have to give up on a lot of the optimization potential.
> 
> The main benefit comes from having less chunks at the planning stage, right? For this it has to happen before the inheritance expansion. Are there other benefits for which it must happen earlier? I'm asking just so that I have a full understanding of what's going on.

Yes this is about having less chunks at planning stage. Which hook would you suggest that should run instead that is after subquery pullup and before hypertable expansion?

akuzm:
> Yes this is about having less chunks at planning stage. Which hook would you suggest that should run instead that is after subquery pullup and before hypertable expansion?

For SELECTs we could do this just in `ts_plan_expand_hypertable_chunks`, but not sure about UPDATEs...

svenklemm:
> > Yes this is about having less chunks at planning stage. Which hook would you suggest that should run instead that is after subquery pullup and before hypertable expansion?
> 
> For SELECTs we could do this just in `ts_plan_expand_hypertable_chunks`, but not sure about UPDATEs...

SELECT FOR UPDATE also does not use our own hypertable expansion so moving the code there would reduce the queries it could be applied to.

akuzm (approved):

PR #4678, opened 2022-09-08
Add interface for troubleshooting job failures

konskov:
This commit gives more visibility into job failures by making the
information regarding a job runtime error available in an extension
table (`job_errors`) that users can directly query.
This commit also adds an infromational view on top of the table for
convenience.
To prevent the `job_errors` table from growing too large,
a retention job is also set up with a default retention interval
of 1 month. The retention job is registered with a custom check
function that requires that a valid "drop_after" interval be provided
in the config field of the job.

### DIFF
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -172,6 +172,48 @@ BEGIN
     WHERE mat_hypertable_id OPERATOR(pg_catalog.=) _cagg_data.mat_hypertable_id;
 END;
 $BODY$;
+CREATE TABLE _timescaledb_internal.job_errors (
+  job_id integer not null, 
+  pid integer,
+  start_time timestamptz,
+  finish_time timestamptz,
+  error_data jsonb
+);
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_internal.job_errors', '');
+
+CREATE VIEW timescaledb_information.job_errors AS 
+SELECT
+    job_id,
+    error_data ->> 'proc_schema' as proc_schema,
+    error_data ->> 'proc_name' as proc_name,
+    pid,
+    start_time,
+    finish_time,
+    error_data ->> 'sqlerrcode' AS sqlerrcode,
+    CASE WHEN error_data ->>'message' IS NOT NULL THEN 
+      CASE WHEN error_data ->>'detail' IS NOT NULL THEN 
+        CASE WHEN error_data ->>'hint' IS NOT NULL THEN concat(error_data ->>'message', '. ', error_data ->>'detail', '. ', error_data->>'hint')
+        ELSE concat(error_data ->>'message', ' ', error_data ->>'detail')
+        END
+      ELSE
+        CASE WHEN error_data ->>'hint' IS NOT NULL THEN concat(error_data ->>'message', '. ', error_data->>'hint')
+        ELSE error_data ->>'message'
+        END
+      END
+    ELSE
+      'job crash detected, see server logs'
+    END
+    AS err_message
+FROM
+    _timescaledb_internal.job_errors;
+
+ALTER TABLE _timescaledb_internal.bgw_job_stat ADD COLUMN flags integer;
+UPDATE _timescaledb_internal.bgw_job_stat SET flags = 0;
+
+ALTER TABLE _timescaledb_internal.bgw_job_stat 
+ALTER COLUMN flags SET NOT NULL,
+ALTER COLUMN flags SET DEFAULT 0;
 
 -- Issue #4727
 ALTER TABLE _timescaledb_catalog.continuous_agg_migrate_plan_step
--- a/sql/views.sql
+++ b/sql/views.sql
@@ -299,4 +299,31 @@ ORDER BY table_name,
   segmentby_column_index,
   orderby_column_index;
 
+-- troubleshooting job errors view
+CREATE OR REPLACE VIEW timescaledb_information.job_errors AS 
+SELECT
+    job_id,
+    error_data ->> 'proc_schema' as proc_schema,
+    error_data ->> 'proc_name' as proc_name,
+    pid,
+    start_time,
+    finish_time,
+    error_data ->> 'sqlerrcode' AS sqlerrcode,
+    CASE WHEN error_data ->>'message' IS NOT NULL THEN 
+      CASE WHEN error_data ->>'detail' IS NOT NULL THEN 
+        CASE WHEN error_data ->>'hint' IS NOT NULL THEN concat(error_data ->>'message', '. ', error_data ->>'detail', '. ', error_data->>'hint')
+        ELSE concat(error_data ->>'message', ' ', error_data ->>'detail')
+        END
+      ELSE
+        CASE WHEN error_data ->>'hint' IS NOT NULL THEN concat(error_data ->>'message', '. ', error_data->>'hint')
+        ELSE error_data ->>'message'
+        END
+      END
+    ELSE
+      'job crash detected, see server logs'
+    END
+    AS err_message
+FROM
+    _timescaledb_internal.job_errors;
+
 GRANT SELECT ON ALL TABLES IN SCHEMA timescaledb_information TO PUBLIC;
--- a/src/bgw/job.c
+++ b/src/bgw/job.c
@@ -44,6 +44,7 @@
 #include "bgw/scheduler.h"
 
 #include <cross_module_fn.h>
+#include "jsonb_utils.h"
 
 #define TELEMETRY_INITIAL_NUM_RUNS 12
 
@@ -156,6 +157,57 @@ job_config_check(BgwJob *job, Jsonb *config)
 			 NameStr(job->fd.check_schema),
 			 NameStr(job->fd.check_name),
 			 job->fd.id);
+}
+
+/* this function fills in a jsonb with the non-null fields of
+ the error data and also includes the proc name and schema in the jsonb
+ we include these here to avoid adding these fields to the table */
+static Jsonb *
+ts_errdata_to_jsonb(ErrorData *edata, Name proc_schema, Name proc_name)
+{
+	JsonbParseState *parse_state = NULL;
+	pushJsonbValue(&parse_state, WJB_BEGIN_OBJECT, NULL);
+	if (edata->sqlerrcode)
+		ts_jsonb_add_str(parse_state, "sqlerrcode", unpack_sql_state(edata->sqlerrcode));
+	if (edata->message)
+		ts_jsonb_add_str(parse_state, "message", edata->message);
+	if (edata->detail)
+		ts_jsonb_add_str(parse_state, "detail", edata->detail);
+	if (edata->hint)
+		ts_jsonb_add_str(parse_state, "hint", edata->hint);
+	if (edata->filename)
+		ts_jsonb_add_str(parse_state, "filename", edata->filename);
+	if (edata->lineno)
+		ts_jsonb_add_int32(parse_state, "lineno", edata->lineno);
+	if (edata->funcname)
+		ts_jsonb_add_str(parse_state, "funcname", edata->funcname);
+	if (edata->domain)
+		ts_jsonb_add_str(parse_state, "domain", edata->domain);
+	if (edata->context_domain)
+		ts_jsonb_add_str(parse_state, "context_domain", edata->context_domain);
+	if (edata->context)
+		ts_jsonb_add_str(parse_state, "context", edata->context);
+	if (edata->schema_name)
+		ts_jsonb_add_str(parse_state, "schema_name", edata->schema_name);
+	if (edata->table_name)
+		ts_jsonb_add_str(parse_state, "table_name", edata->table_name);
+	if (edata->column_name)
+		ts_jsonb_add_str(parse_state, "column_name", edata->column_name);
+	if (edata->datatype_name)
+		ts_jsonb_add_str(parse_state, "datatype_name", edata->datatype_name);
+	if (edata->constraint_name)
+		ts_jsonb_add_str(parse_state, "constraint_name", edata->constraint_name);
+	if (edata->internalquery)
+		ts_jsonb_add_str(parse_state, "internalquery", edata->internalquery);
+	if (edata->detail_log)
+		ts_jsonb_add_str(parse_state, "detail_log", edata->detail_log);
+	if (strlen(NameStr(*proc_schema)) > 0)
+		ts_jsonb_add_str(parse_state, "proc_schema", NameStr(*proc_schema));
+	if (strlen(NameStr(*proc_name)) > 0)
+		ts_jsonb_add_str(parse_state, "proc_name", NameStr(*proc_name));
+	/* we add the schema qualified name here as well*/
+	JsonbValue *result = pushJsonbValue(&parse_state, WJB_END_OBJECT, NULL);
+	return JsonbValueToJsonb(result);
 }
 
 static BgwJob *
@@ -975,6 +1027,46 @@ zero_guc(const char *guc_name)
 				(errcode(ERRCODE_INTERNAL_ERROR), errmsg("could not set \"%s\" guc", guc_name)));
 }
 
+/*
+ * This function creates an entry in the job_errors table
+ * when a background job throws a runtime error or the job scheduler
+ * detects that the job crashed
+ */
+bool
+ts_job_errors_insert_tuple(const FormData_job_error *job_err)
+{
+	Catalog *catalog = ts_catalog_get();
+	Relation rel = table_open(catalog_get_table_id(catalog, JOB_ERRORS), RowExclusiveLock);
+	TupleDesc desc = RelationGetDescr(rel);
+	Datum values[Natts_job_error];
+	bool nulls[Natts_job_error] = { false };
+	CatalogSecurityContext sec_ctx;
+
+	values[AttrNumberGetAttrOffset(Anum_job_error_job_id)] = Int32GetDatum(job_err->job_id);
+	values[AttrNumberGetAttrOffset(Anum_job_error_start_time)] =
+		TimestampTzGetDatum(job_err->start_time);
+	values[AttrNumberGetAttrOffset(Anum_job_error_finish_time)] =
+		TimestampTzGetDatum(job_err->finish_time);
+
+	if (job_err->pid > 0)
+		values[AttrNumberGetAttrOffset(Anum_job_error_pid)] = Int64GetDatum(job_err->pid);
+	else
+		nulls[AttrNumberGetAttrOffset(Anum_job_error_pid)] = true;
+	if (job_err->error_data)
+		values[AttrNumberGetAttrOffset(Anum_job_error_error_data)] =
+			JsonbPGetDatum(job_err->error_data);
+	else
+		nulls[AttrNumberGetAttrOffset(Anum_job_error_error_data)] = true;
+
+	ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
+	ts_catalog_insert_values(rel, desc, values, nulls);
+	ts_catalog_restore_user(&sec_ctx);
+
+	table_close(rel, RowExclusiveLock);
+
+	return true;
+}
+
 extern Datum
 ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
 {
@@ -1016,6 +1108,9 @@ ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
 		elog(ERROR, "job %d not found when running the background worker", params.job_id);
 
 	pgstat_report_appname(NameStr(job->fd.application_name));
+	MemoryContext oldcontext = CurrentMemoryContext;
+	TimestampTz start_time = DT_NOBEGIN, finish_time = DT_NOBEGIN;
+	NameData proc_schema = { 0 }, proc_name = { 0 };
 
 	PG_TRY();
 	{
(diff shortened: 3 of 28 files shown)


konskov:
### need fixing:
- [ ] it's possible to insert multiple records pertaining to the same crash if there's another server crash and the job hasn't run again in the meantime, which is misleading
### need testing: 
- [ ] bypassing permission check to allow users to alter the error retention job

sql/updates/latest-dev.sql
```diff
@@ -46,3 +46,23 @@ BEGIN
...
+  job_id integer not null, 
+  pid bigint, -- is int enough? what's the max pid on a system
+  start_time timestamptz,
+  finish_time timestamptz,
+  error_data jsonb
+);
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_internal.job_errors', '');
+
+CREATE OR REPLACE VIEW timescaledb_information.job_errors AS
```
fabriziomello, line 59:
```suggestion
CREATE VIEW timescaledb_information.job_errors AS 
```

sql/updates/latest-dev.sql
```diff
@@ -46,3 +46,23 @@ BEGIN
...
+
+CREATE OR REPLACE VIEW timescaledb_information.job_errors AS 
+SELECT job_id, pid, start_time, finish_time, error_data->'sqlerrcode' as sqlerrcode, 
+(
+  CASE WHEN error_data IS NOT NULL THEN
+    (error_data->'message')::text
+  ELSE
+    'job crash detected, see server logs'
+  END) AS err_message, error_data->'detail' as err_detail, 
+error_data->'hint' as err_hint FROM _timescaledb_internal.job_errors;
```
fabriziomello, line 67:
```suggestion
SELECT
    job_id,
    pid,
    start_time,
    finish_time,
    error_data ->> 'sqlerrcode' AS sqlerrcode,
    COALESCE(error_data ->> 'message', 'job crash detected, see server logs') AS err_message,
    error_data ->> 'detail' AS err_detail,
    error_data ->> 'hint' AS err_hint
FROM
    _timescaledb_internal.job_errors;
```

Note the usage of `COALESCE` and the operator `->>` that result in a TEXT type instead of JSONB

sql/views.sql
```diff
@@ -299,4 +299,15 @@ ORDER BY table_name,
...
+-- troubleshooting job errors view
+CREATE OR REPLACE VIEW timescaledb_information.job_errors AS 
+SELECT job_id, pid, start_time, finish_time, error_data->'sqlerrcode' as sqlerrcode, 
+(
+  CASE WHEN error_data IS NOT NULL THEN
+    (error_data->'message')::text
+  ELSE
+    'job crash detected, see server logs'
+  END) AS err_message, error_data->'detail' as err_detail, 
+error_data->'hint' as err_hint FROM _timescaledb_internal.job_errors;
```
fabriziomello, line 311: Same as pointed out in latest-dev

src/bgw/job.c
```diff
@@ -1098,7 +1145,35 @@ ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
...
+		jerr.job_id = params.job_id;
+		jerr.start_time = start_time;
+		jerr.finish_time = finish_time;
+		jerr.pid = MyProcPid;
+
+		StartTransactionCommand();
+
+		ts_job_errors_insert_relation(&jerr);
+
+		CommitTransactionCommand();
```
fabriziomello, line 1174: Why two separated transactions here?

src/bgw/job_stat.c
```diff
@@ -565,7 +565,21 @@ ts_bgw_job_stat_next_start(BgwJobStat *jobstat, BgwJob *job, int32 consecutive_f
...
 
 	if (jobstat->fd.consecutive_crashes > 0)
+	{
+		// update the errors table regarding the crash
+		FormData_job_error jerr = { 0 };
+		jerr.error_data = NULL;
+		jerr.start_time = jobstat->fd.last_start;
+		// what should this be? maybe -infinity to indicate the job never finished, crashed instead?
+		// or now? (jobstat will set it to now in a bit)
+		jerr.finish_time = ts_timer_get_current_timestamp();
```
fabriziomello, line 575: NULL ??

src/jsonb_utils.c
```diff
@@ -239,3 +239,47 @@ ts_jsonb_get_interval_field(const Jsonb *json, const char *key)
 
 	return DatumGetIntervalP(interval_datum);
 }
+
+Jsonb *
+ts_errdata_to_jsonb(ErrorData *edata)
```
fabriziomello, line 244: Did u implemented it from scratch or copied from Postgres source code? If copied is good to place a comment about it and from where you copied

sql/updates/latest-dev.sql
konskov: Actually I think all these fields should be `TEXT` instead of `JSONB` and I'll change them accordingly.

src/jsonb_utils.c
konskov: I wrote it from scratch

src/bgw/job_stat.c
konskov: hmm actually if I set it to NULL, or -infinity, I'll run into some issues with the retention policy: Currently the retention policy drops the entries for which `finish_time < now() - retention_interval`.

src/bgw/job.c
konskov: Fixed

src/bgw/job_stat.c
```diff
@@ -333,6 +337,8 @@ bgw_job_stat_tuple_mark_end(TupleInfo *ti, void *const data)
 	fd->last_run_success = result_ctx->result == JOB_SUCCESS ? true : false;
 	fd->total_crashes--;
 	fd->consecutive_crashes = 0;
+	// last_crash_reported = false;
+	fd->flags = ts_clear_flags_32(fd->flags, LAST_CRASH_REPORTED);
```
mkindahl, line 341: Comment just duplicates code, so not needed. It is clear as it is.
```suggestion
	fd->flags = ts_clear_flags_32(fd->flags, LAST_CRASH_REPORTED);
```

src/bgw/job_stat.c
```diff
@@ -148,7 +151,8 @@ bgw_job_stat_tuple_mark_start(TupleInfo *ti, void *const data)
 	fd->last_run_success = false;
 	fd->total_crashes++;
 	fd->consecutive_crashes++;
-
+	// maybe unnecessary, but clear the flags (error repored -> false)
+	fd->flags = ts_clear_flags_32(fd->flags, LAST_CRASH_REPORTED);
```
mkindahl, line 155:
```suggestion
	fd->flags = ts_clear_flags_32(fd->flags, LAST_CRASH_REPORTED);
```

src/bgw/job.c
```diff
@@ -976,6 +978,41 @@ zero_guc(const char *guc_name)
 				(errcode(ERRCODE_INTERNAL_ERROR), errmsg("could not set \"%s\" guc", guc_name)));
 }
 
+bool
+ts_job_errors_insert_relation(FormData_job_error *job_err)
```
mkindahl, line 982: Strange name. It does not really insert a relation, does it? You should probably add a short comment before outlining what it does.

tsl/test/expected/bgw_db_scheduler.out
```diff
@@ -107,8 +107,8 @@ SELECT ts_bgw_db_scheduler_test_run_and_wait_for_scheduler_finish(50);
 
 -- empty
 SELECT * FROM _timescaledb_internal.bgw_job_stat;
- job_id | last_start | last_finish | next_start | last_successful_finish | last_run_success | total_runs | total_duration | total_successes | total_failures | total_crashes | consecutive_failures | consecutive_crashes 
---------+------------+-------------+------------+------------------------+------------------+------------+----------------+-----------------+----------------+---------------+----------------------+---------------------
+ job_id | last_start | last_finish | next_start | last_successful_finish | last_run_success | total_runs | total_duration | total_successes | total_failures | total_crashes | consecutive_failures | consecutive_crashes | flags 
+--------+------------+-------------+------------+------------------------+------------------+------------+----------------+-----------------+----------------+---------------+----------------------+---------------------+-------
```
mkindahl, line 111: You might want to take the opportunity to switch to vertical display (`\x on`) here so that it is easier to review in the future.

tsl/test/sql/bgw_custom.sql
```diff
@@ -51,7 +51,7 @@ SELECT add_job('custom_proc2','1h', config:= '{"type":"procedure"}'::jsonb);
 SELECT add_job('custom_func', '1h', config:='{"type":"function"}'::jsonb);
 SELECT add_job('custom_func_definer', '1h', config:='{"type":"function"}'::jsonb);
 
-SELECT * FROM timescaledb_information.jobs WHERE job_id != 1 ORDER BY 1;
+SELECT * FROM timescaledb_information.jobs WHERE job_id NOT IN (1,2) ORDER BY 1;
```
mkindahl, line 54: Why 1 and 2? Please add a comment what these jobs are and why you need to exclude them from the result.

tsl/test/t/006_job_crash_log.pl
```diff
@@ -0,0 +1,65 @@
+# This file and its contents are licensed under the Timescale License.
+# Please see the included NOTICE for copyright information and
+# LICENSE-TIMESCALE for a copy of the license.
+
+use strict;
+use warnings;
+use PostgresNode;
+use Test::More;
+use TestLib;
```
mkindahl, line 9: This might require some tweaking to support PG15. Please check #4734

tsl/test/t/006_job_crash_log.pl
```diff
@@ -0,0 +1,65 @@
+# This file and its contents are licensed under the Timescale License.
+# Please see the included NOTICE for copyright information and
+# LICENSE-TIMESCALE for a copy of the license.
+
+use strict;
```
mkindahl, line 5: Please add a comment on what this test is expected to test. It will make maintaining it in the future easier.

tsl/test/t/006_job_crash_log.pl
```diff
@@ -0,0 +1,65 @@
...
+use warnings;
+use PostgresNode;
+use Test::More;
+use TestLib;
+
+my $node = get_new_node('primary');
+$node->init(allows_streaming => 1);
+$node->append_conf('postgresql.conf',
+	"shared_preload_libraries = 'timescaledb'");
+$node->start;
```
mkindahl, line 15: These steps are already done if you create a [`TimescaleNode`](https://github.com/timescale/timescaledb/blob/main/test/perl/TimescaleNode.pm) instead of a `PostgresNode`. It will also make sure that you do all initialization that are necessary, now and in the future.

tsl/test/t/006_job_crash_log.pl
```diff
@@ -0,0 +1,65 @@
...
+
+# create proc to run as job
+my $query =
+  q[CREATE OR REPLACE PROCEDURE custom_proc_sleep60(jobid int, config jsonb) LANGUAGE PLPGSQL AS
+$$
+BEGIN
+  RAISE NOTICE 'im about to sleep for 60 seconds, plenty of time for you to kill me';
+  perform pg_sleep(60);
+END
+$$; ];
```
mkindahl, line 36: Tip: Here docs are easier to read.
 ```suggestion
my $query = <<'END_OF_SQL';
CREATE OR REPLACE PROCEDURE custom_proc_sleep60(jobid int, config jsonb) LANGUAGE PLPGSQL AS
$$
BEGIN
  RAISE NOTICE 'im about to sleep for 60 seconds, plenty of time for you to kill me';
  perform pg_sleep(60);
END
$$;
END_OF_SQL
```

File: tsl/src/bgw_policy/job_api.c [hunk 16]
```diff
@@ -288,8 +295,9 @@ job_alter(PG_FUNCTION_ARGS)
 	job = find_job(job_id, PG_ARGISNULL(0), if_exists);
 	if (job == NULL)
 		PG_RETURN_NULL();
-
-	ts_bgw_job_permission_check(job);
+	/* if this is the error retention job, skip the permission check and allow modifications */
+	if (!is_error_retention_job(job))
+		ts_bgw_job_permission_check(job);
```
mkindahl, line 300: Not sure why that is necessary. Circumventing permission checks is always a risk, so would be good to understand why this is necessary.

src/bgw/job.c
konskov: yes, shall I rename it to `ts_job_errors_insert_record` or `ts_job_errors_insert_tuple` instead? That is what it does, it inserts a new record into the `job_errors` table

mkindahl: Either one is fine, but please add a brief comment as well.

tsl/test/t/006_job_crash_log.pl
konskov: Thank you, fixed

src/bgw/job.c
```diff
@@ -1078,7 +1117,32 @@ ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
...
 		elog(LOG, "job %d threw an error", params.job_id);
-		PG_RE_THROW();
+
+		ErrorData *edata;
+		FormData_job_error jerr = { 0 };
+		// switch away from error context to not lose the data
+		MemoryContextSwitchTo(oldcontext);
+		edata = CopyErrorData();
+
+		StartTransactionCommand();
```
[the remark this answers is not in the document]
konskov: No, there is no need for a separate transaction. I removed it.

src/bgw/job.c
```diff
@@ -1078,7 +1117,32 @@ ts_bgw_job_entrypoint(PG_FUNCTION_ARGS)
...
+
+		jerr.error_data = ts_errdata_to_jsonb(edata);
+		jerr.job_id = params.job_id;
+		jerr.start_time = start_time;
+		jerr.finish_time = finish_time;
+		jerr.pid = MyProcPid;
+
+		ts_job_errors_insert_relation(&jerr);
+
+		CommitTransactionCommand();
```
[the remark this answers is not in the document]
konskov: Well, I am not sure when and why to use one over the other, and I thought using `CommitTransactionCommand` is in keeping with the rest of the operations here.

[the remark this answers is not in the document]
konskov: When I tried using `SPI_commit` instead that resulted in a crash

src/bgw/job.c
```diff
@@ -44,6 +44,8 @@
 #include "bgw/scheduler.h"
 
 #include <cross_module_fn.h>
+#include "jsonb_utils.h"
+#include "unistd.h"
```
mkindahl, line 48: What do you need this for?
```suggestion
#include <unistd.h>
```

src/bgw/job_stat.c
```diff
@@ -19,6 +19,9 @@
 #define MAX_FAILURES_MULTIPLIER 20
 #define MIN_WAIT_AFTER_CRASH_MS (5 * 60 * 1000)
 
+#define JOB_STAT_FLAGS_DEFAULT 0
+#define LAST_CRASH_REPORTED 1
+
```
mkindahl, line 24: Might be better to add in the header file, so other dependent modules can use it.

src/bgw/job_stat.c
```diff
@@ -565,7 +602,23 @@ ts_bgw_job_stat_next_start(BgwJobStat *jobstat, BgwJob *job, int32 consecutive_f
...
+	{
+		if (!ts_flags_are_set_32(jobstat->fd.flags, LAST_CRASH_REPORTED))
+		{
+			/* Update the errors table regarding the crash */
+			FormData_job_error jerr = { 0 };
+			jerr.error_data = NULL;
+			jerr.start_time = jobstat->fd.last_start;
+			jerr.finish_time = ts_timer_get_current_timestamp();
+			jerr.pid = -1;
+			jerr.job_id = jobstat->fd.id;
```
mkindahl, line 614: Nit: We support C99, and if you provide an initializer, the remaining fields will automatically be initialized to zero.

```suggestion
			const FormData_job_error jerr = {
			.error_data = NULL,
			.start_time = jobstat->fd.last_start,
			.finish_time = ts_timer_get_current_timestamp(),
			.pid = -1,
			.job_id = jobstat->fd.id,
		};
```

tsl/test/expected/bgw_db_scheduler.out
```diff
@@ -1434,9 +1448,9 @@ SELECT wait_for_job_1_to_run(2);
...
-   1024 | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.25 1999 PST | Fri Dec 31 16:00:00.15 1999 PST | t                |          2 | @ 0            |               2 |              0 |             0 |                    0 |                   0
+ job_id |           last_start            |           last_finish           |           next_start            |     last_successful_finish      | last_run_success | total_runs | total_duration | total_successes | total_failures | total_crashes | consecutive_failures | consecutive_crashes | flags 
+--------+---------------------------------+---------------------------------+---------------------------------+---------------------------------+------------------+------------+----------------+-----------------+----------------+---------------+----------------------+---------------------+-------
+   1024 | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.15 1999 PST | Fri Dec 31 16:00:00.25 1999 PST | Fri Dec 31 16:00:00.15 1999 PST | t                |          2 | @ 0            |               2 |              0 |             0 |                    0 |                   0 |     0
```
mkindahl, line 1453: Vertical display here as well?

tsl/test/expected/bgw_db_scheduler.out
```diff
@@ -1553,9 +1567,9 @@ SELECT * FROM sorted_bgw_log;
...
-   1025 | Fri Dec 31 16:00:00.48 1999 PST | Fri Dec 31 16:00:00.48 1999 PST | Fri Dec 31 16:00:00.49 1999 PST | Fri Dec 31 16:00:00.48 1999 PST | t                |          2 | @ 0            |               2 |              0 |             0 |                    0 |                   0
+ job_id |           last_start            |           last_finish           |           next_start            |     last_successful_finish      | last_run_success | total_runs | total_duration | total_successes | total_failures | total_crashes | consecutive_failures | consecutive_crashes | flags 
+--------+---------------------------------+---------------------------------+---------------------------------+---------------------------------+------------------+------------+----------------+-----------------+----------------+---------------+----------------------+---------------------+-------
+   1025 | Fri Dec 31 16:00:00.48 1999 PST | Fri Dec 31 16:00:00.48 1999 PST | Fri Dec 31 16:00:00.49 1999 PST | Fri Dec 31 16:00:00.48 1999 PST | t                |          2 | @ 0            |               2 |              0 |             0 |                    0 |                   0 |     0
```
mkindahl, line 1572: Vertical display?

tsl/test/sql/bgw_custom.sql
mkindahl: Looks good. Resolving comment.

tsl/test/t/006_job_crash_log.pl
mkindahl: Looks good. Resolving comment.

tsl/test/t/006_job_crash_log.pl
```diff
@@ -0,0 +1,64 @@
...
+# LICENSE-TIMESCALE for a copy of the license.
+
+use strict;
+use warnings;
+use TimescaleNode;
+use Test::More;
+use TestLib;
+
+# This test checks that a job crash is reported in the job errors table
+# (ie a record gets inserted into the table _timescaledb_internal.job_errors).
```
mkindahl, line 12: Nit: use either "i.e." or "that is".
```suggestion
# (that is, a record gets inserted into the table _timescaledb_internal.job_errors).
```

tsl/test/t/006_job_crash_log.pl
```diff
@@ -0,0 +1,64 @@
...
+my $query_add =
+  q[select add_job('custom_proc_sleep60', '5 minutes', initial_start => now())];
+my $jobid = $node->safe_psql('postgres', "$query_add");
+is($jobid, '1000', 'job was added');
+
+# sleep 10 to make sure job has started running
+$node->safe_psql('postgres', "select pg_sleep(10)");
+# select the pid of this job in order to kill it
+my $query_pid =
+  qq[select pid from pg_stat_activity where application_name like 'User-Defined Action%' and query like '%custom_proc_sleep60%'];
```
mkindahl, line 48: Heredocs work here as well. Perl is great in that way. Note the use of double-quotes for the heredoc tag.

```suggestion
my $query_pid = <<"END_OF_QUERY";
select pid from pg_stat_activity
where application_name like 'User-Defined Action%'
   and query like '%custom_proc_sleep60%'
END_OF_QUERY
```

mkindahl (approved):
Looks good. I added a few suggestions, but it's up to you if you want to incorporate them.

src/bgw/job.c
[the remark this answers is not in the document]
mkindahl: Ok. You can take a look at that in a separate PR.

tsl/test/t/006_job_crash_log.pl
mkindahl: I'm not sure you actually need the double-quotes here since you have no Perl variables in the query.

src/bgw/job.c
konskov: I don't need it, I probably added it for debugging and then forgot to remove. Removing it now.

konskov:
I added a second commit (that I will squash) to incorporate the feedback I received yesterday after the demo. 
Also, still remaining: 
- [ ] Handle job failures that don't lead to crashes and don't have a related ereport that gets caught by `PG_TRY - CATCH` (e.g. SIGTERM)

sql/updates/latest-dev.sql
```diff
@@ -172,6 +172,48 @@ BEGIN
...
+      END
+    ELSE
+      'job crash detected, see server logs'
+    END
+    AS err_message
+FROM
+    _timescaledb_internal.job_errors;
+
+ALTER TABLE _timescaledb_internal.bgw_job_stat ADD COLUMN flags integer;
+UPDATE _timescaledb_internal.bgw_job_stat SET flags = 0;
```
mfundul, line 212:
```suggestion
ALTER TABLE _timescaledb_internal.bgw_job_stat ADD COLUMN flags integer NOT NULL DEFAULT 0;
```

This should add the `NOT NULL` constraint during upgrades. It should also make sure the column is initialized without extra overhead at `ALTER TABLE` statement time.

src/bgw/job_stat.c
```diff
@@ -562,7 +598,34 @@ ts_bgw_job_stat_next_start(BgwJobStat *jobstat, BgwJob *job, int32 consecutive_f
...
+			pushJsonbValue(&parse_state, WJB_BEGIN_OBJECT, NULL);
+			ts_jsonb_add_str(parse_state, "proc_schema", NameStr(proc_schema));
+			ts_jsonb_add_str(parse_state, "proc_name", NameStr(proc_name));
+			JsonbValue *result = pushJsonbValue(&parse_state, WJB_END_OBJECT, NULL);
+
+			const FormData_job_error jerr = {
+				.error_data = JsonbValueToJsonb(result),
+				.start_time = jobstat->fd.last_start,
+				.finish_time = ts_timer_get_current_timestamp(),
+				.pid = -1,
```
mfundul, line 619: Is there a way to get the pid here?

sql/updates/latest-dev.sql
konskov: Hmm, apparently, this doesn't work as per the comment in `test/sql/updates/catalog_missing_columns.sql` so I think I'll have to do it as before.

src/bgw/job_stat.c
konskov: I think the most straightforward way is to update the `bgw_job_stat` table once the job starts, for example in a field `pid integer`. Maybe it should be `pid bigint` so I can use 32 bits for the pid and the rest for the flags. Then again, I think pids will be way less than 2^32 so it could be safe to use the `flags` field to store both the pid and the flag `last_crash_reported`

tsl/test/t/006_job_crash_log.pl
```diff
@@ -0,0 +1,67 @@
+# This file and its contents are licensed under the Timescale License.
+# Please see the included NOTICE for copyright information and
+# LICENSE-TIMESCALE for a copy of the license.
+
+use strict;
+use warnings;
+use TimescaleNode;
+use Test::More;
+use TestLib;
```
mfundul, line 9:
```suggestion
```

mfundul (approved):

tsl/test/t/006_job_crash_log.pl
konskov: Done

PR #4691, opened 2022-09-12
Remove test case causing bgw_custom crash

konskov:
Patch #4425 introduced regression test failures, namely, a crash in function `ts_bgw_job_update_by_id`. The failures are due to the COMMIT statement in the custom check procedure. This patch removes that particular test case from bgw_custom.


konskov:
some failed runs regarding this: https://github.com/timescale/timescaledb/runs/8236364647?check_suite_focus=true, https://github.com/timescale/timescaledb/runs/8171789810?check_suite_focus=true and there are more

fabriziomello (approved):
In any case we should fix this in a later PR because we introduced a failure here.

svenklemm (approved):

PR #4717, opened 2022-09-16
Fix insufficient permissions message in delete_job

konskov:
Previously, when the user executing delete_job did not have owner permissions, the error message stated that the owner has insufficient permissions to delete the job, which is not true and might be misleading. This patch alters the error message to print the name of the user attempting to perform the delete without the required privileges.


svenklemm (changes requested):
The user name shown is correct. Maybe the message can be misunderstood but it is not wrong.
It means the user running this command has insufficient permissions to delete jobs for the user named in the message which is the owner of the job.

PR #4720, opened 2022-09-16
Fix chunk exclusion for prepared statements and dst changes

svenklemm:
The constify code constifying TIMESTAMPTZ expressions when doing chunk exclusion did not account for daylight saving time switches leading to different calculation outcomes when timezone changes. This patch adds a 4 hour safety buffer to any such calculations.


src/planner/expand_hypertable.c
```diff
@@ -268,6 +269,47 @@ constify_timestamptz_op_interval(PlannerInfo *root, OpExpr *constraint)
...
+		if (!var_on_left)
+			add = !add;
+
+		/*
+		 * The safety buffer is chosen to be 4 hours because daylight saving time
+		 * changes seem to be in the range between -1 and 2 hours.
+		 */
+		if (add)
+			constified_tstz += 4 * USECS_PER_HOUR;
+		else
```
gayyappan, line 307: don't have  a test for this case.

gayyappan (approved):
Consider adding another test as well.
Changes look good.

akuzm (approved):

PR #4731, opened 2022-09-20
Add health check function

erimatnor:
A new health check function _timescaledb_internal.health() returns the
health and status of the database instance, including any configured
data nodes (in case the instance is an access node).

Since the function returns also the health of the data nodes, it tries
hard to avoid throwing errors. An error will fail the whole function
and therefore not return any node statuses, although some of the nodes
might be healthy.

The health check on the data nodes is a recursive (remote) call to the
same function on those nodes. Unfortunately, the check will fail with
an error if a connection cannot be established to a node (or an error
occurs on the connection), which means the whole function call will
fail. This will be addressed in a future change by returning the error
in the function result instead.

### DIFF
--- a/tsl/src/remote/healthcheck.c
+++ b/tsl/src/remote/healthcheck.c
@@ -1,0 +1,313 @@
+/*
+ * This file and its contents are licensed under the Timescale License.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-TIMESCALE for a copy of the license.
+ */
+#include <postgres.h>
+#include <funcapi.h>
+#include <utils/builtins.h>
+#include <utils/lsyscache.h>
+#include <utils/syscache.h>
+
+#include "healthcheck.h"
+#include "data_node.h"
+#include "dist_commands.h"
+#include "dist_util.h"
+
+/*
+ * Functions and data structures for printing a health check result.
+ */
+enum Anum_show_conn
+{
+	Anum_health_node_name = 1,
+	Anum_health_healthy,
+	Anum_health_in_recovery,
+	Anum_health_error,
+	_Anum_health_max,
+};
+
+#define Natts_health (_Anum_health_max - 1)
+
+/*
+ * Health check result.
+ *
+ * Produces a tuple with four columns:
+ *
+ * 1. Node name (NULL for the access node itself)
+ *
+ * 2. A "healthy" boolean. An access node returns "false" in case a data node
+ * is not responding.
+ *
+ * 3. A boolean telling if the node is in recovery. Note that the "healthy"
+ * status will be "true" while the node is in recovery.
+ *
+ * 4. An optional error message explaining the reason for "healthy" being
+ * false.
+ */
+static HeapTuple
+create_health_check_tuple(const char *data_node, bool in_recovery, TupleDesc tupdesc)
+{
+	Datum values[Natts_health];
+	bool nulls[Natts_health];
+
+	memset(nulls, true, sizeof(nulls));
+
+	if (NULL != data_node)
+	{
+		NameData *node_name = palloc(sizeof(NameData));
+		namestrcpy(node_name, data_node);
+		values[AttrNumberGetAttrOffset(Anum_health_node_name)] = NameGetDatum(node_name);
+		nulls[AttrNumberGetAttrOffset(Anum_health_node_name)] = false;
+	}
+
+	values[AttrNumberGetAttrOffset(Anum_health_healthy)] = BoolGetDatum(in_recovery ? false : true);
+	nulls[AttrNumberGetAttrOffset(Anum_health_healthy)] = false;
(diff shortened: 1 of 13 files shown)


tsl/src/remote/healthcheck.c
```diff
@@ -0,0 +1,270 @@
...
+			StringInfo cmd = makeStringInfo();
+			Oid fnamespaceid = get_func_namespace(fcinfo->flinfo->fn_oid);
+
+			appendStringInfo(cmd,
+							 "SELECT * FROM %s.%s()",
+							 get_namespace_name(fnamespaceid),
+							 get_func_name(fcinfo->flinfo->fn_oid));
+			data_node_list = data_node_get_node_name_list();
+			search_path = GetConfigOption("search_path", false, false);
+			result = ts_dist_cmd_invoke_on_data_nodes_using_search_path(cmd->data,
```
pmwkaa, line 111: But this will fail and throw an error, right?

erimatnor: Yes, exactly. This is pointed out in the commit message as future work.

erimatnor:
This PR is missing tests, which needs to be added before merge.

tsl/src/remote/healthcheck.c
```diff
@@ -0,0 +1,270 @@
...
+	_Anum_health_max,
+};
+
+#define Natts_health (_Anum_health_max - 1)
+
+static HeapTuple
+create_health_check_tuple(const char *data_node, bool in_recovery, TupleDesc tupdesc)
+{
+	Datum values[Natts_health];
+	bool nulls[Natts_health] = { true };
```
nikkhils, line 35: better to use nulls[Natts_health] = {false};?

erimatnor:
Tests added

nikkhils (approved):
LGTM!

tsl/src/remote/healthcheck.c
```diff
@@ -0,0 +1,304 @@
...
+
+	if (NULL != data_node)
+	{
+		NameData *node_name = palloc(sizeof(NameData));
+		namestrcpy(node_name, data_node);
+		values[AttrNumberGetAttrOffset(Anum_health_node_name)] = NameGetDatum(node_name);
+		nulls[AttrNumberGetAttrOffset(Anum_health_node_name)] = false;
+	}
+
+	values[AttrNumberGetAttrOffset(Anum_health_healthy)] = BoolGetDatum(true);
```
pmwkaa, line 63: So healthy column is reserved for the future use in such case? Like internal consistency check etc

tsl/src/remote/healthcheck.c
```diff
@@ -0,0 +1,304 @@
...
+			StringInfo cmd = makeStringInfo();
+			Oid fnamespaceid = get_func_namespace(fcinfo->flinfo->fn_oid);
+
+			appendStringInfo(cmd,
+							 "SELECT * FROM %s.%s()",
+							 get_namespace_name(fnamespaceid),
+							 get_func_name(fcinfo->flinfo->fn_oid));
+			data_node_list = data_node_get_node_name_list();
+			search_path = GetConfigOption("search_path", false, false);
+			result = ts_dist_cmd_invoke_on_data_nodes_using_search_path(cmd->data,
```
pmwkaa, line 136: not sure we need to set search_path here

tsl/src/remote/healthcheck.c
```diff
@@ -0,0 +1,304 @@
...
+#include "dist_util.h"
+
+/*
+ * Functions and data structures for printing a health check result.
+ */
+enum Anum_show_conn
+{
+	Anum_health_node_name = 1,
+	Anum_health_healthy,
+	Anum_health_in_recovery,
```
pmwkaa, line 24: Likely node being in_recovery should not be considered healthy as well...I guess it won't hurt to have extra details (ui maybe?), since we use it for AN as well

pmwkaa (approved):
looks good, just comments

tsl/src/remote/healthcheck.c
erimatnor: It would be `false` if, for example, the data node cannot establish a connection to the DN. But it is not possible to test now since it throws an error. It would be `true` if the node appears healthy, but, e.g., `in_recovery` is also `true`. We could, of course, simply use `false` in both cases, but I figured it might be useful information to also let the caller know that the node is recovering.

tsl/src/remote/healthcheck.c
erimatnor: Removed.

tsl/src/remote/healthcheck.c
erimatnor: Changed this so that healthy is `false` when `in_recovery` is true.

65278 (approved):
LGTM. The PR isn't hard to follow, that said I have not run it.

Issue #4728, opened 2022-09-20
[Bug]: Bulk insert fails


mrksngl:
### What type of bug is this?

Unexpected error

### What subsystems and features are affected?

Data ingestion, Distributed hypertable, Multi-node

### What happened?

When checking whether #4576 is solved with 2.8.0, I got a new error when having a distributed setup.

I set up an access node and a single data node using the `timescaledb:2.8.0-pg14` docker image.
I copy the same data that were used by the author of the issue referenced above. The only thing I changed is to use a distributed hypertable.

I experienced one of the following two situations:
 * Either the COPY does not fail but is painfully slow. It completely stalls now and then and in the end takes 4x the time needed compared to copying the same data in a non-distributed ht.
 * Or (this actually happens way more often) it transfers all data to the access node but then terminates the connection with the message below.

### TimescaleDB version affected

2.8.0

### PostgreSQL version used

14.5

### What operating system did you use?

timescaledb:2.8.0-pg14 docker image

### What installation method did you use?

Docker

### What platform did you run on?

Other

### Relevant log output and stack trace

```bash
ERROR:  XX000: epoll_ctl() failed: Operation not permitted
LOCATION:  WaitEventAdjustEpoll, latch.c:1083
WARNING:  01000: AbortTransaction while in ABORT state
LOCATION:  AbortTransaction, xact.c:2681
WARNING:  01000: transaction rollback on data node "dn0" failed
LOCATION:  dist_txn_xact_callback_abort, dist_txn.c:182
ERROR:  XX000: epoll_ctl() failed: Operation not permitted
LOCATION:  WaitEventAdjustEpoll, latch.c:1083
ERROR:  XX000: epoll_ctl() failed: Operation not permitted
LOCATION:  WaitEventAdjustEpoll, latch.c:1083
Time: 32905,319 ms (00:32,905)
ERROR:  42P01: relation "public.inet_session_usages_id_seq" does not exist
LINE 1: SELECT pg_catalog.setval('public.inet_session_usages_id_seq'...
                                 ^
LOCATION:  RangeVarGetRelidExtended, namespace.c:428
Time: 7,981 ms
```

### How can we reproduce the bug?

```bash
CREATE DATABASE test;
\c test
CREATE TABLE inet_session_usages (
    id bigint NOT NULL,
    inet_session_id integer NOT NULL,
    bytes_in bigint,
    bytes_out bigint,
    created_at timestamp with time zone,
    year_created integer,
    month_created integer,
[...]

PR #4738, opened 2022-09-21
Fix error when waiting for remote COPY to finish

akuzm:
Pass proper flags to WaitLatchOrSocket

Fixes https://github.com/timescale/timescaledb/issues/4728

### DIFF
--- a/tsl/src/remote/connection.c
+++ b/tsl/src/remote/connection.c
@@ -2278,8 +2278,8 @@ remote_connection_end_copy(TSConnection *conn, TSConnectionError *err)
 				 * writeable, we just retry.
 				 */
 				(void) WaitLatchOrSocket(MyLatch,
+										 WL_TIMEOUT | WL_SOCKET_WRITEABLE | WL_EXIT_ON_PM_DEATH,
 										 socket,
-										 WL_TIMEOUT | WL_SOCKET_WRITEABLE,
 										 /* timeout = */ 1000,
 										 /* wait_event_info = */ 0);
 			}
@@ -2348,12 +2348,8 @@ remote_connection_end_copy(TSConnection *conn, TSConnectionError *err)
 		ExecStatusType status = PQresultStatus(res);
 		if (status != PGRES_COMMAND_OK)
 		{
-			success =
-				fill_result_error(err,
-								  ERRCODE_CONNECTION_EXCEPTION,
-								  psprintf("invalid result status '%s' when ending remote COPY",
-										   PQresStatus(status)),
-								  res);
+			success = false;
+			remote_connection_get_result_error(res, err);
 		}
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ argument or resolve the type ambiguity by casting to the intended type.
 * #4685 Improve chunk exclusion for space dimensions
 * #4696 Report warning when enabling compression on hypertable
 * #4720 Fix chunk exclusion for prepared statements and dst changes
+* #4738 Fix the assorted epoll_ctl() errors that could occur with COPY into a distributed hypertable
 * #4739 Fix continuous aggregate migrate check constraint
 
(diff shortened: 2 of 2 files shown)


tsl/src/remote/connection.c
```diff
@@ -2278,8 +2278,8 @@ remote_connection_end_copy(TSConnection *conn, TSConnectionError *err)
 				 * writeable, we just retry.
 				 */
 				(void) WaitLatchOrSocket(MyLatch,
+										 WL_TIMEOUT | WL_SOCKET_WRITEABLE | WL_EXIT_ON_PM_DEATH,
```
jnidzwetzki, line 2281: Is the `WL_EXIT_ON_PM_DEATH` constant actually needed or is it a leftover from the first version of the PR?

tsl/src/remote/connection.c
```diff
@@ -2348,12 +2348,8 @@ remote_connection_end_copy(TSConnection *conn, TSConnectionError *err)
...
 		if (status != PGRES_COMMAND_OK)
 		{
-			success =
-				fill_result_error(err,
-								  ERRCODE_CONNECTION_EXCEPTION,
-								  psprintf("invalid result status '%s' when ending remote COPY",
-										   PQresStatus(status)),
-								  res);
+			success = false;
+			remote_connection_get_result_error(res, err);
```
jnidzwetzki, line 2352: In other functions (like in `remote_connection_begin_copy`) we use `fill_result_error` to report connection errors. Using this function, we can add additional information about the error location that might be useful for future error investigation.

```suggestion
                fill_result_error(err,
                                                  ERRCODE_CONNECTION_FAILURE,
                                                  "Invalid status when ending remote COPY",
                                                  res);
```

jnidzwetzki (approved):
Overall it looks good and in my test environment, I don’t see `epoll()` errors anymore using my 'bad' virtual network link.

```
# With 5% package loss and ~20 ms latency
test2=# \COPY rides FROM /tmp/nyc_data_rides.csv CSV;
COPY 10906860
Time: 3593856.448 ms (59:53.856)

# Without package loss and ~20 ms latency
test2=# \COPY rides FROM /tmp/nyc_data_rides.csv CSV;
COPY 10906860
Time: 298452.311 ms (04:58.452)
test2=# \COPY rides FROM /tmp/nyc_data_rides.csv CSV;
COPY 10906860
Time: 302073.666 ms (05:02.074) 
```

tsl/src/remote/connection.c
akuzm: I found that fill_result_error overwrites the original error, i.e. it doesn't report what the error on the connection was. Let's use this one instead, we also use it in other similar places such as row-by-row data fetcher.

tsl/src/remote/connection.c
akuzm: Yes, it is required according to the comments to `WaitLatchOrSocket`.

tsl/src/remote/connection.c
jnidzwetzki: I was mentioning this, because `remote_connection_get_result_error` simply calls `fill_result_error` ([link](https://github.com/timescale/timescaledb/blob/533a138ae22ddabf8a576d23dcb8f88a842fb6e5/tsl/src/remote/connection.c#L881)) with an empty `errmsg`.

tsl/src/remote/connection.c
pmwkaa: good catch :)

pmwkaa (approved):

tsl/src/remote/connection.c
akuzm: Yeah... It's weird that passing a non-empty message overwrites the remote message, can't tell how this can happen from the code. But I've definitely seen it show "Invalid status PGRES_FATAL_ERROR" instead of the remote error text, which is bad.

Issue #4617, opened 2022-08-15
[Enhancement]: Add index support to compress_chunk


svenklemm:
### What type of enhancement is this?

Performance

### What subsystems and features will be improved?

Compression

### What does the enhancement do?

Currently compress_chunk will never use an index even if one matching the order by configuration exists. Currently we always do a tuple sort over a table scan leading to temp file creation when not enough memory is configured. This explains some of the problems our users are seeing when compressing large chunks. When an index is present that matches the order by configuration we can skip the tuple sort completely and get sorted tuples from an index scan instead which should speed up compression and eliminate any temp file creation in compress_chunk.

PR #4743, opened 2022-09-22
[WIP] Added index support for compress chunk

shhnwz:
#4617


tsl/src/compression/compression.c
```diff
@@ -297,14 +321,44 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
 
 	TupleDesc in_desc = RelationGetDescr(in_rel);
 	TupleDesc out_desc = RelationGetDescr(out_rel);
+	in_rel_index_iods = RelationGetIndexList(in_rel);
```
nikkhils, line 324: `in_rel_index_oids` might be the name you want to use here.

tsl/src/compression/compression.c
```diff
@@ -297,14 +321,44 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+				bool compression_col_is_orderby = COMPRESSIONCOL_IS_ORDER_BY(keys[i]);
+				bool compression_col_is_segmentby = COMPRESSIONCOL_IS_SEGMENT_BY(keys[i]);
+				if (namestrcmp((Name) &keys[i]->attname,
+							   NameStr(index_rel->rd_att->attrs[i].attname)) == 0 &&
+					(compression_col_is_orderby || compression_col_is_segmentby))
+				{
+					if (compression_col_is_segmentby)
+						indexscan_direction = BackwardScanDirection;
+					else
+						indexscan_direction = ForwardScanDirection;
```
nikkhils, line 342: Why should `indexscan_direction` be specified by is_segmentby or is_orderby? ISTM, if we want to have same output as tuplesort (ascending AFAIK), then we should inspect the `matched_index_rel` and check if it's an asc/desc index and set the scan direction to FORWARD if it's asc or to BACKWARD if it's descending when we do index_getnext_slot on it.

tsl/src/compression/compression.c
```diff
@@ -314,12 +368,54 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
 						true /*need_bistate*/);
 
-	row_compressor_append_sorted_rows(&row_compressor, sorted_rel, in_desc);
+	if (matched_index_rel != NULL)
+	{
+		DEBUG_WAITPOINT("compress_chunk_indexscan_start");
+		index_scan = index_beginscan(in_rel, matched_index_rel, SnapshotAny, 0, 0);
+		slot = table_slot_create(in_rel, NULL);
+		index_rescan(index_scan, NULL, 0, NULL, 0);
+		while (index_getnext_slot(index_scan, indexscan_direction, slot))
```
nikkhils, line 377: this indexscan_direction should be based on if the matched index is itself a ASC/DESC index.

PR #4745, opened 2022-09-22
Hypertable FK reference to PostgreSQL partitioned table

### LINKED ISSUE #4684: [Bug]: Hypertable FK reference to vanilla PostgreSQL hash-partitioned table doesn't seem to work
### What type of bug is this?

Incorrect result

Command processing, Data ingestion, Partitioning, Query planner

### What happened?

It seems that foreign key references **from** hypertables **to** vanilla PostgreSQL hash-partitioned tables aren't supported? I get foreign key constraints triggered where I wasn't expecting them and don't get them when using plain ol' PGDB! I've attempted to craft a MRE on dbfiddle to demonstrate this. I've also written about it on one of the TSDB slack channels, which has lead me here.

https://timescaledb.slack.com/archives/C4GT3N90X/p1661445729287369
https://dbfiddle.uk/bO3D8EQ2

This is reproducible using the 'bleeding edge' TimescaleDB; The nightly docker build from 'main' based on top of PG14.

I've re-run this several times over with the same data now and each time its the same violation arising from the _wrong_ partition number (I think this is key!). Simply removing the foreign key constraint in the hypertable allows the ingest to succeed and me to inspect the database (which is how I can discover which partition the referenced key is in vs. which TSDB thinks its in). The constraint is not triggered if I replace the hypertable with a regular table. The dbfiddle above has more information too, in the code comments.

2.7+ ('main' in git)

### PostgreSQL version used

14

Ubuntu (whatever your docker base is)

Docker

On prem/Self-hosted

```bash
ERROR:  insert or update on table "_hyper_1_1_chunk" violates foreign key constraint "1_4_timescale_dev_id_fkey1"
DETAIL:  Key (id, dev)=(-7005382656368145192, -5972894815461654684) is not present in table "regular_1".
```

```bash
https://dbfiddle.uk/bO3D8EQ2
```

sb230132:
Consider a hypertable which has a foreign key constraint on a
referenced table, which is a parititioned table. In such case, foreign
key constraint is duplicated for each of the partitioned table. When
we insert into a hypertable, we end up checking the foreign key
constraint multiple times, which obviously leads to foreign key
constraint violation. Instead we only check foreign key constraint of
the parent table of the partitioned table.

Fixes #4684

### DIFF
--- a/src/chunk_constraint.c
+++ b/src/chunk_constraint.c
@@ -577,6 +577,15 @@ chunk_constraint_need_on_chunk(const char chunk_relkind, Form_pg_constraint conf
 		 */
 		return false;
 	}
+	/*
+	   Check if the foreign key constraint references a partition in a partitioned
+	   table. In that case, we shouldn't include this constraint as we will end up
+	   checking the foreign key constraint once for every partition, which obviously
+	   leads to foreign key constraint violation. Instead, we only include constraints
+	   referencing the parent table of the partitioned table.
+	*/
+	if (conform->contype == CONSTRAINT_FOREIGN && OidIsValid(conform->conparentid))
+		return false;
 
 	/* Foreign tables do not support non-check constraints, so skip them */
 	if (chunk_relkind == RELKIND_FOREIGN_TABLE)
--- a/test/sql/create_hypertable.sql
+++ b/test/sql/create_hypertable.sql
@@ -442,3 +442,161 @@ SELECT * FROM create_hypertable('sample_table', 'time');
 
 -- cleanup
 DROP TABLE sample_table CASCADE;
+
+-- github issue 4684
+-- test PARTITION BY HASH
+CREATE TABLE regular(
+   id INT NOT NULL,
+   dev INT NOT NULL,
+   value INT,
+   CONSTRAINT cstr_regular_pky PRIMARY KEY (id)
+) PARTITION BY HASH (id);
+
+DO $$
+BEGIN
+   FOR i IN 1..2
+   LOOP
+      EXECUTE format('
+         CREATE TABLE %I
+         PARTITION OF regular
+         FOR VALUES WITH (MODULUS 2, REMAINDER %s)',
+         'regular_' || i, i - 1
+      );
+   END LOOP;
+END;
+$$;
+
+INSERT INTO regular SELECT generate_series(1,1000), 44,55;
+
+CREATE TABLE timescale (
+   ts TIMESTAMP WITH TIME ZONE NOT NULL,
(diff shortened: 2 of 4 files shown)


test/sql/create_hypertable.sql
```diff
@@ -441,4 +441,63 @@ ALTER TABLE sample_table DROP COLUMN name;
...
\ No newline at end of file
+DROP TABLE sample_table CASCADE;
+
+-- github issue 4684
+CREATE TABLE regular(
+   id INT NOT NULL,
+   dev INT NOT NULL,
+   value INT,
+   CONSTRAINT cstr_regular_pky PRIMARY KEY (id)
+) PARTITION BY HASH (id);
```
erimatnor, line 452: The title of the PR seems to indicate that it problem is specific to hash-partitioned tables. Is it really the case that  this problem occurs only with hash-partitioned tables or does it also happen with LIST and RANGE partitioned ones as well? 

In either case, it might be good to test list and range partitioning as well. If this is a general problem with partitioned tables, then the title of the PR might need to be updated.

src/chunk_constraint.c
```diff
@@ -577,6 +577,16 @@ chunk_constraint_need_on_chunk(const char chunk_relkind, Form_pg_constraint conf
 		 */
 		return false;
 	}
+	/*
+	   Consider a hypertable which has a foreign key constraint on a referenced
+	   table, which is a parititioned table. In such case, foreign key constraint
+	   is duplicated for each of the partitioned table. When we insert into a
+	   hypertable, we end up checking the foreign key constraint multiple times,
+	   which obviously leads to foreign key constraint violation. Instead we only
+	   check for the parent table of the partitioned table.
```
erimatnor, line 586:
```suggestion
	   Check if the foreign key constraint references a partition in a partitioned table. In that case, we shouldn't include this constraint as we will end up checking the foreign key constraint once for every partition,
	   which obviously leads to foreign key constraint violation. Instead, we only
	   include constraints referencing the parent table of the partitioned table.
```

test/sql/create_hypertable.sql
sb230132: Thanks. Issue exists for all types of partitioned tables. Tested for LIST, RANGE and HASH based partitioned tables and added the same in create_hypertable.sql.
Also removed reference to hash partition in commit message.

erimatnor (approved):

mkindahl (approved):

maudrid:
Is this solved in 2.8.1, because I'm still having this same issue on this version.
Was it removed?
image

PR #4756, opened 2022-09-27
Improve compression job IO performance

antekresic:
This PR does a couple of changes to improve IO usage for compression jobs:

- changes sort memory limits from `work_mem` to `maintenance_work_mem`, reducing the chance to use temp buffers for larger chunks
- moves uncompressed chunk ANALYZE operation after heap has been loaded into shared buffers by the heap scan we need to do in order to sort the chunk contents

Disable-check: commit-count

### DIFF
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -20,6 +20,7 @@
 #include <funcapi.h>
 #include <libpq/pqformat.h>
 #include <miscadmin.h>
+#include <nodes/pg_list.h>
 #include <storage/lmgr.h>
 #include <storage/predicate.h>
 #include <utils/builtins.h>
@@ -390,6 +391,7 @@ static void compress_chunk_populate_sort_info_for_column(Oid table,
 														 const ColumnCompressionInfo *column,
 														 AttrNumber *att_nums, Oid *sort_operator,
 														 Oid *collation, bool *nulls_first);
+static void run_analyze_on_chunk(Oid chunk_relid);
 
 static Tuplesortstate *
 compress_chunk_sort_relation(Relation in_rel, int n_keys, const ColumnCompressionInfo **keys)
@@ -441,6 +443,12 @@ compress_chunk_sort_relation(Relation in_rel, int n_keys, const ColumnCompressio
 
 	heap_endscan(heapScan);
 
+	/* Perform an analyze on the chunk to get up-to-date stats before compressing.
+	 * We do it at this point because we've just read out the entire chunk into
+	 * tuplesort, so its pages are likely to be cached and we can save on I/O.
+	 */
+	run_analyze_on_chunk(in_rel->rd_id);
+
 	ExecDropSingleTupleTableSlot(heap_tuple_slot);
 
 	tuplesort_performsort(tuplesortstate);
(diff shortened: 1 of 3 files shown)


svenklemm (approved):

svenklemm:
Needs changelog entry

tsl/src/compression/compression.c
```diff
@@ -488,6 +496,25 @@ compress_chunk_populate_sort_info_for_column(Oid table, const ColumnCompressionI
 	ReleaseSysCache(tp);
 }
 
+static void
```
akuzm, line 499:
```suggestion
/* Run VACUUM ANALYZE on the given chunk. */
static void
```

akuzm: Maybe it should be called `run_vacuum_analyze` if that's what it does?

tsl/src/compression/compression.c
```diff
@@ -441,6 +443,12 @@ compress_chunk_sort_relation(Relation in_rel, int n_keys, const ColumnCompressio
 
 	heap_endscan(heapScan);
 
+	/* Perform an analyze on the chunk to get up-to-date stats before compressing.
+	 * We do it at this point to reduce the amount of random IO operations
+	 * necessary in case the statistic target is larger than the chunk heap size.
```
akuzm, line 448: We do it here because we just read out the entire chunk into tuplesort, its pages are cached and we can save on disk I/O, right? The wording with random I/O and statistics target is somewhat confusing...

akuzm (approved):

tsl/src/compression/compression.c
antekresic: That's true. I'm open to suggestions about better wording :smile:

akuzm: "We do it at this point because we've just read out the entire chunk into tuplesort, so its pages are likely to be cached and we can save on I/O".

Although I wonder to which extent this is true, because full heap scans use a dedicated cache strategy which doesn't cache much.

antekresic: I have definitely seen a rise in average read op size. In any case, heap will be present in os/disk cache if not shared buffers.

cevian: The Postgres cache uses the dedicated strategy but the os/disk cache doesn't and that's the caching we really care about.

mkindahl (approved):
Looks good to me. The lock that we are taking while compressing is very heavy, but this has nothing to do with this PR and I do not think that we can take a weaker lock.

tsl/src/compression/compression.c
antekresic: Updated to `run_analyze_on_chunk`.

tsl/src/compression/compression.c
antekresic: Updated comment to suggestion.

MarkCupitt:
I just ran into an issue on teh latest Helmn Chart, was getting a Patroni Startup Error 

`128 kB is outside the valid range for parameter "maintenance_work_mem`

adding `maintenance_work_mem: 64MB` into teh Helm Values 
```
      postgresql:
        parameters:
          maintenance_work_mem: 64MB
```
FIxed it ..

PR #4767, opened 2022-09-28
Allow manual index creation in CAggs

### LINKED ISSUE #4735: [Bug]: `permission denied for schema _timescaledb_internal` when manually creating index on a continuous aggregate
### What type of bug is this?

Unexpected error

Continuous aggregate

### What happened?

When creating a materialized view as a regular user, the following error is raised when adding an index to a materialized view:

```
ERROR:  42501: permission denied for schema _timescaledb_internal
```

2.8.0

### PostgreSQL version used

14.5

Ubuntu 22.04

Source

On prem/Self-hosted

```bash
ERROR:  42501: permission denied for schema _timescaledb_internal
LOCATION:  aclcheck_error, aclchk.c:3447
```

```bash
-- as a regular user, not as a superuser:
CREATE TABLE public.test(inserted timestamptz not null);
SELECT public.create_hypertable('test', 'inserted');
CREATE MATERIALIZED VIEW public.test_cagg WITH (
  timescaledb.continuous,
  timescaledb.materialized_only=false
) AS
SELECT
    public.time_bucket(interval '1 day', inserted) AS bucket,
    count(*) AS n_inserted
FROM
    public.test
GROUP BY
    bucket
WITH NO DATA;

CREATE INDEX ON public.test_cagg(n_inserted);
```

RafiaSabih:
The materialised hypertable resides in the _timescaledb.internal schema which resulted in permission error at the time of manual index creation by non super user. To solve this, it now switches to timescaledb user before index creation of CAgg.

Fixes #4735


svenklemm:
Needs test

fabriziomello:
You should create a failing test case using a non-superuser to create the index on CAgg

fabriziomello:
> You should create a failing test case using a non-superuser to create the index on CAgg

The test is correct. My mistake!!!

fabriziomello (approved):

svenklemm (changes requested):
This needs additional security checks. Currently this will allow anybody to create an index on a cagg. Additionally in the `IF NOT EXISTS` path when the object already exists the user is not restored.

src/process_utility.c
```diff
@@ -2703,12 +2708,25 @@ process_index_start(ProcessUtilityArgs *args)
 		PreventInTransactionBlock(true,
 								  "CREATE INDEX ... WITH (timescaledb.transaction_per_chunk)");
 
+	if (cagg)
+	{
+		/*
+		 * if this is an index creation fro cagg, then we need to switch user as the current
+		 * user might not be having permissions on internal schema where cagg index will be
+		 * created.
```
svenklemm, line 2716:
```suggestion
		 * If this is an index creation for cagg, then we need to switch user as the current
		 * user might not have permissions on then internal schema where the cagg index
		 * will be created.
```

svenklemm (approved):

PR #4786, opened 2022-10-01
Extend the now() optimization to also apply to CURRENT_TIMESTAMP

### LINKED ISSUE #4782: [Bug]: CURRENT_TIMESTAMP very slow when subtracting an interval
### What type of bug is this?

Performance issue

Query executor, Query planner

### What happened?

Reposting from the slack channel at the request of David Kohn

I'm using the timescale managed service to ingest about 1300 metrics every minute to a single table in a EAV style (organizationId, fieldId, timestamp, value), with an index on organizationId asc, fieldId asc, timestamp desc. I have continuous aggregates running to create 5m, 10m, 15m, 30m, 60m, 180m, 360m, 720m, and 1440m averages.
I'm running the following query:
EXPLAIN select "data".* from "automation"."data" 
where 
    "timestamp" >= CURRENT_TIMESTAMP - INTERVAL '10 minutes' 
    and "timestamp" <= CURRENT_TIMESTAMP 
    and "field_id" = 985
    and "organization_id" = '23c8c57f-e848-4162-8d09-6952558cdf4a'
order by "timestamp" desc
And so far it's been taking over 600s. It still hasn't finished yet. In general this query has been taking an absurd amount of time, and I have no idea why. It takes this long even without the EXPLAIN.

EDIT: Seems to run faster if I use a constant timestamp like '2022-09-30T14:55-0700'::timestamptz instead of CURRENT_TIMESTAMP
EDIT2: Found the problem, but don't understand why. Running this query seems to run as quickly as expected
select "data".* from "automation"."data" 
where 
    "timestamp" >= 'NOW'::timestamptz - INTERVAL '10 minutes' 
    and "timestamp" <= CURRENT_TIMESTAMP 
    and "field_id" = 985
    and "organization_id" = '23c8c57f-e848-4162-8d09-6952558cdf4a'
order by "timestamp" desc
So ultimately, the culprit appears to be doing a timestamp  >= CURRENT_TIMESTAMP - INTERVAL '10 minutes' but using 'NOW'::timestamptz seems to work just fine. Anyone have any idea as to why? 

2.7.0

### PostgreSQL version used

13.8

Cloud Managed

Not applicable

Managed Service for TimescaleDB (MST/Aiven), Microsoft Azure Cloud

_No response_

```bash
Run a query on a hypertable with a `CURRENT_TIMESTAMP` call that subtracts an `INTERVAL`, such as

select "data".* from "automation"."data" 
where 
    "timestamp" >= CURRENT_TIMESTAMP - INTERVAL '10 minutes' 
    and "timestamp" <= CURRENT_TIMESTAMP 
    and "field_id" = 985
    and "organization_id" = '23c8c57f-e848-4162-8d09-6952558cdf4a'
order by "timestamp" desc
```
```

### PR DESCRIPTION
**Pull request by svenklemm**
The optimization that constifies certain now() expressions before hypertable expansion did not apply to CURRENT_TIMESTAMP even though it is functionally similar to now(). This patch extends the optimization to CURRENT_TIMESTAMP.

Fixes #4782

### DIFF
--- a/src/planner/constify_now.c
+++ b/src/planner/constify_now.c
@@ -45,6 +45,19 @@ get_hypertable_dimension(Oid relid, int flags)
 }
 
 static bool
+is_valid_now_func(Node *node)
+{
+	if (IsA(node, FuncExpr) && castNode(FuncExpr, node)->funcid == F_NOW)
+		return true;
+
+	if (IsA(node, SQLValueFunction) &&
+		castNode(SQLValueFunction, node)->type == SVFOP_CURRENT_TIMESTAMP)
+		return true;
+
+	return false;
+}
+
+static bool
 is_valid_now_expr(OpExpr *op, List *rtable)
 {
 	int flags = CACHE_FLAG_MISSING_OK | CACHE_FLAG_NOCREATE;
@@ -93,7 +106,7 @@ is_valid_now_expr(OpExpr *op, List *rtable)
 		return false;
 
 	/* Var > now() or Var >= now() */
-	if (IsA(lsecond(op->args), FuncExpr) && lsecond_node(FuncExpr, op->args)->funcid == F_NOW)
+	if (is_valid_now_func(lsecond(op->args)))
 		return true;
 
 	if (!IsA(lsecond(op->args), OpExpr))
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/planner/constify_now.c [hunk 1]
```diff
@@ -44,6 +44,19 @@ get_hypertable_dimension(Oid relid, int flags)
...
 
+static bool
+is_valid_now_func(Node *node)
+{
+	if (IsA(node, FuncExpr) && castNode(FuncExpr, node)->funcid == F_NOW)
+		return true;
+
+	if (IsA(node, SQLValueFunction) &&
+		castNode(SQLValueFunction, node)->type == SVFOP_CURRENT_TIMESTAMP)
+		return true;
```
akuzm, line 55: Weird that codecov complains... probably this is a stale comment.

akuzm (approved):

fabriziomello (approved):

PR #4788, opened 2022-10-03
Add COSTS OFF to make test more stable

akuzm:
For some reason the cost estimates are different on my machine.


fabriziomello (approved):
LGTM. Don't forget to update expected test files for Postgres 13 and 14

PR #4790, opened 2022-10-04
Update .clang-format for version 14

akuzm:
The only thing we're missing is the configuration for braces after case labels, and the rest of the stuff was just missed by the version 8 we used to use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,18 +371,9 @@ configure_file(${EXT_CONTROL_FILE}.in ${EXT_CONTROL_FILE})
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${EXT_CONTROL_FILE}
         DESTINATION "${PG_SHAREDIR}/extension")
 
-# We look for specific versions installed by distributions before the default
-# name since distros that have installed clang-format-9 will not work and the
-# user need to install an earlier version, which will then be named
-# "clang-format-N".
-#
-# This breaks the CMake convention of using the "default" name first to handle
-# local installs. If this turns out to be something that we want to support, we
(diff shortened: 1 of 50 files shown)


fabriziomello:
Nice. IMHO you can organize this PR in three commits:
1) Code changed (scripts, yml, etc)
2) Code formatted (changes after run clang-format)
3) Adding formatting code commit hash to .git-blame-ignore-revs

akuzm:
The CI checks is not going to pass for 1 or 2 separately, and 1 is small, so I'd merge them. Hash of the 1 and 2 is going to change each time I rebase, and after it's merged, no? I'm not sure what trickery github does under hood... Let's do 3 as a separate PR later.

svenklemm (approved):

CMakeLists.txt
```diff
@@ -371,18 +371,9 @@ configure_file(${EXT_CONTROL_FILE}.in ${EXT_CONTROL_FILE})
...
-# "clang-format-N".
-#
-# This breaks the CMake convention of using the "default" name first to handle
-# local installs. If this turns out to be something that we want to support, we
-# need to look specifically for "clang-format" in the local installation paths
-# before looking for the versioned names in standard installation paths.
 find_program(
   CLANG_FORMAT
-  NAMES clang-format-8 clang-format-7 clang-format
+  NAMES clang-format
```
akuzm, line 376: The `-Wno-error=unknown` gives us a modicum of backwards compatibility, so I think we can just use _the_ clang-format that is present on the machine, without preferring a particular version. Or should we still prefer 14?

fabriziomello (approved):

PR #4803, opened 2022-10-07
Fix bgw_custom test

sb230132:
bgw_custom test fails randomly due to difference in number of chunks getting executed. This patch fixes this issue.


konskov:
Oh I guess I was too quick to approve.. The failing test case for the windows test is `bgw_custom` again

fabriziomello:
@sb230132 Konstantina's previous PR didn't solve it? https://github.com/timescale/timescaledb/pull/4802

konskov:
this test has been flaky for two reasons: one was the next_start field that occasionally was `-infinity` or had an actual value so I removed that from the alter_job return values where it wasn't needed. The other reason is that one of the chunks of the sensor_data table (last test in the file) seems to get created on windows...

svenklemm (changes requested):
now() should not be used in tests as it can result in flaky tests

PR #4810, opened 2022-10-10
Show information about OOM killer in CI 

svenklemm:
Include OOM kill event logs into error printout. Previously these
would only be visible by inspecting the postgres log and looking
for killed by signal 9.


fabriziomello (approved):

svenklemm:
Run with OOM kill: https://github.com/timescale/timescaledb/actions/runs/3226407743/jobs/5279892802

akuzm (approved):

PR #4812, opened 2022-10-11
Add indexes to job_errors table

konskov:
This commit adds two indexes to the job_errors table. One index on job_id to make filtering by job_id faster, and one on the finish_time of the failed job, as the retention job for the job_errors table selects which records to drop based on the value of finish_time.

### DIFF
--- a/sql/updates/latest-dev.sql
+++ b/sql/updates/latest-dev.sql
@@ -14,6 +14,10 @@ CREATE TABLE _timescaledb_internal.job_errors (
   error_data jsonb
 );
 
+-- don't need to drop indexes in reverse-dev because the table itself is dropped
+CREATE INDEX job_errors_job_id_idx ON _timescaledb_internal.job_errors(job_id);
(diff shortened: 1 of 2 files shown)


fabriziomello:
You didn't add DROP INDEX into reverse-dev.sql because it was not released yet and in your previous PR you added a DROP TABLE right??? 

Would be nice to add to comments that this PR (and in the commit message) is related to the previous one where you introduce the `job_errors` table.

fabriziomello (approved):

akuzm (approved):

sql/updates/latest-dev.sql
```diff
@@ -14,6 +14,10 @@ CREATE TABLE _timescaledb_internal.job_errors (
   error_data jsonb
 );
 
+-- don't need to drop indexes in reverse-dev because the table itself is dropped
```
kgyrtkirk, line 17: I don't think that's true anymore; this patch have lost that train :)

fabriziomello (changes requested):
Need to include reverse-dev.sql to remove indexes.

svenklemm:
closing since its outdated

PR #4821, opened 2022-10-13
Index support for compress chunk

### LINKED ISSUE #4617: [Enhancement]: Add index support to compress_chunk

shhnwz:
Index support for compress chunk

 It allows to override tuplesort with indexscan
 if compression setting keys matches with Index keys.
 Moreover this feature has Enable/Disable Toggle.
 To Disable from the client use the following command,
 SET timescaledb.enable_compression_indexscan = 'OFF'

Fixes #4617

### DIFF
--- a/src/guc.c
+++ b/src/guc.c
@@ -77,6 +77,7 @@ bool ts_guc_enable_osm_reads = true;
 TSDLLEXPORT bool ts_guc_enable_transparent_decompression = true;
 bool ts_guc_enable_per_data_node_queries = true;
 bool ts_guc_enable_async_append = true;
+TSDLLEXPORT bool ts_guc_enable_compression_indexscan = true;
 TSDLLEXPORT bool ts_guc_enable_skip_scan = true;
 int ts_guc_max_open_chunks_per_insert = 10;
 int ts_guc_max_cached_chunks_per_hypertable = 10;
@@ -380,6 +381,17 @@ _guc_init(void)
 							 "Enable getting and showing EXPLAIN output from remote nodes",
 							 &ts_guc_enable_remote_explain,
 							 false,
+							 PGC_USERSET,
+							 0,
+							 NULL,
+							 NULL,
+							 NULL);
+
+	DefineCustomBoolVariable("timescaledb.enable_compression_indexscan",
+							 "Enable compression to take indexscan path",
+							 "Enable indexscan during compression, if matching index is found",
+							 &ts_guc_enable_compression_indexscan,
+							 true,
 							 PGC_USERSET,
 							 0,
 							 NULL,
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -32,7 +32,7 @@
 #include <utils/syscache.h>
 #include <utils/tuplesort.h>
 #include <utils/typcache.h>
-
+#include <catalog/pg_am.h>
 #include <utils.h>
 
 #include "compat/compat.h"
@@ -49,7 +49,7 @@
 #include "segment_meta.h"
 #include "ts_catalog/hypertable_compression.h"
 #include "ts_catalog/catalog.h"
-
+#include "guc.h"
 #include <nodes/print.h>
 
 #define MAX_ROWS_PER_COMPRESSION 1000
@@ -165,6 +165,17 @@ static void row_compressor_init(RowCompressor *row_compressor, TupleDesc uncompr
 static void row_compressor_append_sorted_rows(RowCompressor *row_compressor,
 											  Tuplesortstate *sorted_rel, TupleDesc sorted_desc);
 static void row_compressor_finish(RowCompressor *row_compressor);
+static void row_compressor_update_group(RowCompressor *row_compressor, TupleTableSlot *row);
+static bool row_compressor_new_row_is_in_new_group(RowCompressor *row_compressor,
+												   TupleTableSlot *row);
+static void row_compressor_append_row(RowCompressor *row_compressor, TupleTableSlot *row);
+static void row_compressor_flush(RowCompressor *row_compressor, CommandId mycid,
+								 bool changed_groups);
+
+static SegmentInfo *segment_info_new(Form_pg_attribute column_attr);
+static void segment_info_update(SegmentInfo *segment_info, Datum val, bool is_null);
+static bool segment_info_datum_is_in_group(SegmentInfo *segment_info, Datum datum, bool is_null);
+static void run_analyze_on_chunk(Oid chunk_relid);
 
 /********************
  ** compress_chunk **
@@ -300,6 +311,18 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
 			   int num_compression_infos)
 {
 	int n_keys;
+	ListCell *lc;
+	int indexscan_direction = NoMovementScanDirection;
+	List *in_rel_index_oids;
+	Relation matched_index_rel = NULL;
+	TupleTableSlot *slot;
+	IndexScanDesc index_scan;
+	bool first_iteration = true;
+	bool changed_groups, compressed_row_is_full;
+	MemoryContext old_ctx;
+	CommandId mycid = GetCurrentCommandId(true);
+	HeapTuple in_table_tp = NULL, index_tp = NULL;
+	Form_pg_attribute in_table_attr_tp, index_attr_tp;
 	const ColumnCompressionInfo **keys;
 	CompressionStats cstat;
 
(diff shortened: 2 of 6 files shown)


tsl/src/compression/compression.c
```diff
@@ -298,14 +320,48 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
 
 	TupleDesc in_desc = RelationGetDescr(in_rel);
 	TupleDesc out_desc = RelationGetDescr(out_rel);
+	in_rel_index_iods = RelationGetIndexList(in_rel);
```
akuzm, line 323: `_oids`?

tsl/src/compression/compression.c
```diff
@@ -298,14 +320,48 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+	if (n_keys == 1)
```
akuzm, line 324: Let's make it work for arbitrary number of columns. We have to just loop over them and if the next column doesn't match or the direction is different, don't use the index.

akuzm:
Needs more tests, probably good to put them into a separate test file.
1) some way to tell from the tests which code is used -- maybe a debug message? I don't really like them because then suddenly other debug messages pop up from unrelated parts of code, and your test becomes flaky. Maybe we can add a debug option which enables an INFO message about which code path is taken.
2) all combinations of desc/asc and nulls first/last that match the index or not.
3) different index type -- we can only use btree here, not gist and so on.

tsl/src/compression/compression.c
svenklemm: The initial PR needs to at least have support for n_keys <= 2 as that will cover most configurations. n_keys = 1 would only apply for compressed hypertables with no segmentby columns.

shhnwz: Implemented for multi-column support

tsl/src/compression/compression.c
shhnwz: Taken care Thanks!

tsl/src/compression/compression.c
```diff
@@ -298,14 +320,232 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+					{
+						if (option)
+						{
+							/* Case NullFirst(DESC) e.g. [NULL->2->1]*/
+							if (((option & INDOPTION_DESC) != 0) &&
+								((option & INDOPTION_NULLS_FIRST) != 0) &&
+								(is_orderby_asc && is_null_first))
+							{
+								current_direction = BackwardScanDirection;
+							}
```
akuzm, line 392: We can simplify these branches as discussed on standup:

```
bool index_orderby_asc = (option & INDOPTION_DESC) == 0;
bool index_nulls_first = (option & INDOPTION_NULLS_FIRST) == 0;

if (is_orderby_asc = index_orderby_asc && is_nulls_fist = index_nulls_first)
{
    current_direction = ForwardScanDirection;
}
else if (is_orderby_asc != index_orderby_asc && is_nulls_first != index_nulls_first)
{
    current_direction = BackwardScanDirection;
}
else
{
    // doesn't match.
}
```

shhnwz: Looks good so far, it only misses couple of corner case like this, but I think its good to go with your simplification. let me check after making the changes.
`if (((option & INDOPTION_DESC) != 0) &&
								((option & INDOPTION_NULLS_FIRST) != 0) &&
								(is_orderby_asc && is_null_first))
							{
								current_direction = BackwardScanDirection;
							}`

svenklemm (commented):
I think we are missing check for collation. Can you add tests with different collation. In the index check the collation must match for an index to be elegible for datatypes that are collation sensitive.

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,165 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+				{
+					int16 att_num = get_attnum(in_table, NameStr(keys[i]->attname));
+
+					int16 option = index_rel->rd_indoption[i];
+					bool index_orderby_asc = ((option & INDOPTION_DESC) == 0);
+					bool index_null_first = ((option & INDOPTION_NULLS_FIRST) != 0);
+					bool is_orderby_asc = keys[i]->orderby_asc;
+					bool is_null_first = keys[i]->orderby_nullsfirst;
+
+					if (index_info->ii_IndexAttrNumbers[i] == att_num)
```
svenklemm, line 394:
```suggestion
					if (OidIsValid(att_num) && index_info->ii_IndexAttrNumbers[i] == att_num)
```

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,165 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+								 "index \"%s\" does not have column \"%s\"",
+								 get_rel_name(index_oid),
+								 NameStr(keys[i]->attname));
+
+						in_table_attr_tp = (Form_pg_attribute) GETSTRUCT(in_table_tp);
+						index_attr_tp = (Form_pg_attribute) GETSTRUCT(index_tp);
+
+						if (COMPRESSIONCOL_IS_SEGMENT_BY(keys[i]))
+						{
+							if (option)
```
akuzm, line 415: What does this check mean? Let's add a comment.

akuzm:
Looks like we still need more tests:
1) When collation doesn't match
2) Two segmentby columns, both the case where they require the same order and the different one. Codecov seems to complain that this is not covered.
3) Would be good to construct some kind of test that shows the compression went correctly, and the compressed  groups are the same as w/o the index. Maybe select segmentby + metadata columns from the compressed chunk?

src/guc.c
```diff
@@ -373,6 +374,18 @@ _guc_init(void)
 							 NULL,
 							 NULL);
 
+	DefineCustomBoolVariable("timescaledb.enable_compression_index_lookup",
+							 "Enable index lookup during compression",
+							 "Enable index lookup during compression, use the matching index"
+							 "to get the sorted tuples in order to save the sorting cpu cycles",
```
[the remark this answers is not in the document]
nikkhils: +1. I don't think we need a guc to enable/disable this.

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,165 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
 
 	TupleDesc in_desc = RelationGetDescr(in_rel);
 	TupleDesc out_desc = RelationGetDescr(out_rel);
+	in_rel_index_oids = RelationGetIndexList(in_rel);
+	int i = 0;
+	/* We create index on the compressed chunk based on ColumnCompressionInfo,
+	 * So, before calling row compressor we need to sort the rows as per ColumnCompressionInfo.
+	 * Cost of sorting can be mitigated if we find and existing BTREE index defined for
```
nikkhils, line 353: typo `and existing` should be `an existing`

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,165 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+							{
+								current_direction = BackwardScanDirection;
+							}
+							else
+							{
+								current_direction = NoMovementScanDirection;
+								break;
+							}
+						}
+						if (HeapTupleIsValid(in_table_tp))
```
nikkhils, line 451: redundant IsValid checks here and below. They will be valid otherwise we error out above

tsl/src/compression/compression.c
```diff
@@ -340,12 +514,65 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+			}
 
-	row_compressor_finish(&row_compressor);
+			row_compressor_append_row(&row_compressor, slot);
+			MemoryContextSwitchTo(old_ctx);
+			ExecClearTuple(slot);
+		}
 
-	tuplesort_end(sorted_rel);
+		run_analyze_on_chunk(in_rel->rd_id);
```
nikkhils, line 554: why do we want to run analyze on chunk?

tsl/src/compression/compression.c
shhnwz: done..

tsl/src/compression/compression.c
shhnwz: Thanks Nikhil, 
you are right, we can safely remove the redundant check. Done.

tsl/src/compression/compression.c
shhnwz: We need to run analyze to sync the stats. Couple of stats test cases were failing without analyze call.

tsl/src/compression/compression.c
shhnwz: Comment added. Done

src/guc.c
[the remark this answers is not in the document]
shhnwz: I discuss the rationale of having toggle for this feature. I will help in providing workaround in case of regression in production. We can take a call in future once this gets stabilized.

tsl/src/compression/compression.c
shhnwz: Yup, I simply could think it of. Added the check

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,179 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
 
 	TupleDesc in_desc = RelationGetDescr(in_rel);
 	TupleDesc out_desc = RelationGetDescr(out_rel);
+	in_rel_index_oids = RelationGetIndexList(in_rel);
+	int i = 0;
+	/* We create index on the compressed chunk based on ColumnCompressionInfo,
+	 * So, before calling row compressor we need to sort the rows as per ColumnCompressionInfo.
```
svenklemm, line 352: These 2 are completely independant operations. We need sorted input because we want batches to be ordered by compress_orderby. This is independant of presence of indexes.

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,179 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+	 * Cost of sorting can be mitigated if we find an existing BTREE index defined for
+	 * uncompressed chunk otherwise expensive tuplesort will come into play.
+	 *
+	 * This interesting piece of code is trying to find any existing index that
+	 * matches the ColumnCompressionInfo just to override tuplesort with indexscan.
```
svenklemm, line 357:
```suggestion
	 * The following code is trying to find an existing index that
	 * matches the ColumnCompressionInfo so we can skip the sequential scan and tuplesort.
```

tsl/src/compression/compression.c
```diff
@@ -340,12 +528,65 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
 						true /*need_bistate*/);
 
-	row_compressor_append_sorted_rows(&row_compressor, sorted_rel, in_desc);
+	if (matched_index_rel != NULL)
+	{
+#ifdef TS_DEBUG
+		const char *compression_path =
+			GetConfigOption("timescaledb.show_compression_path_info", true, false);
+		if (compression_path != NULL && strcmp(compression_path, "on") == 0)
+			elog(INFO, "compress_chunk_indexscan_start");
```
svenklemm, line 537: might be useful to print the index it picked here too

svenklemm (approved):

tsl/src/compression/compression.c
shhnwz: Taken care

tsl/src/compression/compression.c
shhnwz: Taken care.

tsl/src/compression/compression.c
akuzm: OK, I understand, segmentby column is always ASC. But why do you check for `option != 0` here? It might contain some flags unrelated to the ordering. Can you rewrite it using the usual variables `index_orderby_asc` and `index_nulls_first`?

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,180 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+							 * the segment_by always ASC.
+							 */
+							if (option)
+							{
+								if (!index_orderby_asc &&
+									in_table_attr_tp->attcollation == index_attr_tp->attcollation)
+								{
+									current_direction = BackwardScanDirection;
+								}
+								else if (in_table_attr_tp->attcollation ==
```
akuzm, line 431: Codecov complains that there's no test coverage for this and the next condition, let's add it.

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,180 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+						if (!HeapTupleIsValid(index_tp))
+							elog(ERROR,
+								 "index \"%s\" does not have column \"%s\"",
+								 get_rel_name(index_oid),
+								 NameStr(keys[i]->attname));
+
+						in_table_attr_tp = (Form_pg_attribute) GETSTRUCT(in_table_tp);
+						index_attr_tp = (Form_pg_attribute) GETSTRUCT(index_tp);
+
+						if (COMPRESSIONCOL_IS_SEGMENT_BY(keys[i]))
```
akuzm, line 414: By the way, why do we need a completely separate branch for segmentby columns? Can we just set the `is_orderby_asc` and `is_null_first` for them properly, and then use the same checks as for the orderby  columns?

shhnwz: Makes sense, Let me try out with it..

tsl/src/compression/compression.c
shhnwz: Now coverage looks good

tsl/src/compression/compression.c
shhnwz: Done...

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,142 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+						}
+						else if (previous_direction != current_direction)
+						{
+							break;
+						}
+					}
+					else
+					{
+						break;
+					}
```
akuzm, line 450: You're not resetting `current_direction` here, this will lead to an index still being chosen if the last column didn't match, right?

We have to fix some things:
1) add a test for this bug. Try to reproduce it on the unfixed version and verify that the last column is chosen incorrectly.
2) Reset `current_direction` at the start of the loop to avoid bugs like this.
3) Rewrite this as early exit from the loop:
```
					if (att_num == 0 || index_info->ii_IndexAttrNumbers[i] != att_num)
					{
					    break;
					}
```
Long if branches are hard to read, so early exit with break/return is preferrable.

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,142 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+
+					int16 option = index_rel->rd_indoption[i];
+					bool index_orderby_asc = ((option & INDOPTION_DESC) == 0);
+					bool index_null_first = ((option & INDOPTION_NULLS_FIRST) != 0);
+					bool is_orderby_asc =
+						COMPRESSIONCOL_IS_SEGMENT_BY(keys[i]) ? true : keys[i]->orderby_asc;
+					bool is_null_first =
+						COMPRESSIONCOL_IS_SEGMENT_BY(keys[i]) ? false : keys[i]->orderby_nullsfirst;
+
+					if (att_num > 0 && index_info->ii_IndexAttrNumbers[i] == att_num)
```
akuzm, line 397: Attnum can't be zero here, right? The orderby/segmentby keys must be present in the table. We should check this with an assertion.

tsl/src/compression/compression.c
```diff
@@ -323,14 +346,139 @@ compress_chunk(Oid in_table, Oid out_table, const ColumnCompressionInfo **column
...
+
+					int16 option = index_rel->rd_indoption[i];
+					bool index_orderby_asc = ((option & INDOPTION_DESC) == 0);
+					bool index_null_first = ((option & INDOPTION_NULLS_FIRST) != 0);
+					bool is_orderby_asc =
+						COMPRESSIONCOL_IS_SEGMENT_BY(keys[i]) ? true : keys[i]->orderby_asc;
+					bool is_null_first =
+						COMPRESSIONCOL_IS_SEGMENT_BY(keys[i]) ? false : keys[i]->orderby_nullsfirst;
+
+					if (att_num == 0 || index_info->ii_IndexAttrNumbers[i] != att_num)
```
akuzm, line 397: Can `att_num == 0` happen in practice? This means that the orderby/segmentby column is not found in the uncompressed chunk. That would be either a program error or a data corruption. Let's use `Ensure` macro to check this, not `if`.

akuzm (approved):

PR #4822, opened 2022-10-13
Miss segmentby compression option in CAGGs

### LINKED ISSUE #4816: [Bug]: Enabling Compression on a Cagg Results with no "segment_by"
### What type of bug is this?

Performance issue

Compression

### What happened?

When I enable compression on my continous aggregate table and check with `_timescaledb_catalog.hypertable_compression;`, I see that "segment_by" is not compatible with continous aggregate definition.

I was expected **sensor_id** would be a "segment_by" option but it didnt. Because of that, my "sersor_id" based queries are very slow when compressin is enabled...

when I check from compre. def. table; I see **stream_id** is not selected as a segment. When I read source (src/continuous_aggs/options.c) it says so:
> 
> /* get the compression parameters for cagg. The parameters are
> --
>   |   |   |   | * derived from the cagg view definition.
>   |   |   |   | * Computes:
>   |   |   |   | * compress_segmentby = GROUP BY columns from cagg query
>   |   |   |   | * compress_orderby = time_bucket column from cagg query
>   |   |   |   | */
>   |   |   |   | static List *
>   |   |   |   | cagg_get_compression_params(ContinuousAgg *agg, Hypertable *mat_ht)
> 

```
SELECT * FROM _timescaledb_catalog.hypertable_compression ;
 hypertable_id |    attname     | compression_algorithm_id | segmentby_column_index | orderby_column_index | orderby_asc | orderby_nullsfirst 
---------------+----------------+--------------------------+------------------------+----------------------+-------------+--------------------
             1 | time           |                        4 |                        |                    1 | f           | t
             1 | stream_id      |                        0 |                      1 |                      |             | 
             1 | value          |                        3 |                        |                      |             | 

            11 | stream_id      |                        4 |                        |                      |             | 
            11 | bucket         |                        4 |                        |                    1 | t           | f
            11 | value          |                        3 |                        |                      |             | 
            11 | num_values     |                        4 |                        |                      |             | 
```

1: hypertable, 11: cagg

2.7.2

### PostgreSQL version used

14.5

Ubuntu 14.5-1.pgdg22.04+1

Deb/Apt

Google Cloud Platform (GCP)

_No response_

```bash
-- 1. create table
CREATE TABLE public.data (
  time timestamptz NOT NULL,
  stream_id int4 NOT NULL,
  value float8 NOT NULL
);

-- 2. create hypertable
SELECT create_hypertable('data', 'time', create_default_indexes => false, chunk_time_interval => INTERVAL '1 week');

-- 3. import data

-- 4. create indexes
CREATE INDEX ON data(time desc);
CREATE INDEX ON data(stream_id, time desc);

-- 5. create cagg
CREATE MATERIALIZED VIEW data_aggregate_3600
SELECT stream_id,
      time_bucket(INTERVAL '1 hour', time) AS bucket,
      SUM(value) as value,
      count(value) as num_values
FROM data_aggregate
GROUP BY stream_id, bucket;

-- 6. enable compr on cagg
ALTER MATERIALIZED VIEW data_3600 set (timescaledb.compress = true);

--7 sample slow query
SELECT time_bucket('3600 seconds',  bucket) AS time, avg(value) AS value
  FROM data_3600   
WHERE stream_id = 11111
    AND bucket >= '2015-01-11 04:00:00+00'
    AND bucket < '2022-10-01 20:21:46+00'
GROUP BY time
```

fabriziomello:
that store the final aggregation state instead of the byte array of
the partial aggregate state, offering multiple opportunities of
optimizations as well a more compact form.

This new version also removes the unecessary `chunk_id` column from the
materialization hypertable and consequently the re-aggregation in the
user view. It means the user view that query the materialization
hypertable don't have a GROUP BY clause anymore that was problematic
for query performance.

Before 2.7 when users turn compression ON we infer compression options
`segmentby` and `orderby` based on the GROUP BY clause and time bucket
respectively. With the new version without a GROUP BY clause in the user
view the inferetion for the 'segmentby' compression option stopped to
work.

Fixed it by changing the code to the compression on the new version of
Continuous Aggregate (aka finals form) behave the same as the old
version.

Fix #4816

### DIFF
--- a/tsl/src/continuous_aggs/options.c
+++ b/tsl/src/continuous_aggs/options.c
@@ -67,22 +67,55 @@ update_materialized_only(ContinuousAgg *agg, bool materialized_only)
 }
 
 /*
- * Retrieve the cagg view query and find the groupby clause and
- * time_bucket clause. Map them to the column names(of mat.hypertable)
- * Note that cagg_view_query has 2 forms : with union and without UNION
- * We have to extract the part of the query that has finalize_agg on
- * the materialized  hypertable to find the group by clauses.
- * (see continuous_aggs/create.c for more info on the query structure)
- * Returns: list of column names used in group by clause of the cagg query.
+ * This function is responsible to return a list of column names used in
+ * GROUP BY clause of the cagg query. It behaves a bit different depending
+ * of the type of the Continuous Aggregate.
+ *
+ * 1) Partials form (finalized=false)
+ *
+ *    Retrieve the "user view query" and find the GROUP BY clause and
+ *    "time_bucket" clause. Map them to the column names (of mat.hypertable)
+ *
+ *    Note that the "user view query" has 2 forms:
+ *    - with UNION
+ *    - without UNION
+ *
+ *    We have to extract the part of the query that has "finalize_agg" on
+ *    the materialized hypertable to find the GROUP BY clauses.
+ *    (see continuous_aggs/create.c for more info on the query structure)
+ *
+ * 2) Finals form (finalized=true) (>= 2.7)
+ *
+ *    Retrieve the "direct view query" and find the GROUP BY clause and
+ *    "time_bucket" clause. We use the "direct view query" because in the
+ *    "user view query" we removed the re-aggregation in the part that query
+ *    the materialization hypertable so we don't have a GROUP BY clause
+ *    anymore.
+ *
+ *    Get the column name from the GROUP BY clause because all the column
+ *    names are the same in all underlying objects (user view, direct view,
+ *    partial view and materialization hypertable).
  */
 static List *
 cagg_find_groupingcols(ContinuousAgg *agg, Hypertable *mat_ht)
 {
 	List *retlist = NIL;
 	ListCell *lc;
-	Oid cagg_view_oid =
-		get_relname_relid(NameStr(agg->data.user_view_name),
-						  get_namespace_oid(NameStr(agg->data.user_view_schema), false));
+	Oid cagg_view_oid;
+
+	/*
+	 * Get the direct_view definition for the finalized version because
+	 * the user view doesn't have the "GROUP BY" clause anymore.
+	 */
+	if (ContinuousAggIsFinalized(agg))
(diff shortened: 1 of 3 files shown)


tsl/src/continuous_aggs/options.c
```diff
@@ -80,9 +80,21 @@ cagg_find_groupingcols(ContinuousAgg *agg, Hypertable *mat_ht)
 {
 	List *retlist = NIL;
 	ListCell *lc;
-	Oid cagg_view_oid =
-		get_relname_relid(NameStr(agg->data.user_view_name),
-						  get_namespace_oid(NameStr(agg->data.user_view_schema), false));
+	Oid cagg_view_oid;
+
+	/*
+	 * Get the direct_view definition for the finalized version because
```
gayyappan, line 86: Could you also update the comment above the function to clarify that we now have 2 forms for caggs (since 2.7) and how they differ ? (the comment now refers to the old style)

gayyappan (approved):
Code changes look good.

tsl/src/continuous_aggs/options.c
fabriziomello: Updated.

tsl/src/continuous_aggs/options.c
```diff
@@ -117,11 +150,20 @@ cagg_find_groupingcols(ContinuousAgg *agg, Hypertable *mat_ht)
...
+		if (ContinuousAggIsFinalized(agg))
+		{
+			/* "resname" is the same as "mat column names" in the finalized version */
+			if (!cagg_tle->resjunk && cagg_tle->resname)
+				retlist = lappend(retlist, cagg_tle->resname);
+		}
+		else
+		{
+			/* groupby clauses are columns from the mat hypertable */
+			Assert(IsA(cagg_tle->expr, Var));
```
svenklemm, line 163: this is redundant as castNode will Assert the correct type

fabriziomello: removed

PR #4820, opened 2022-10-13
Support parameterized data node scans in joins

akuzm:
This allows us to perform a nested loop join of a small outer local table to an inner distributed hypertable, without downloading the entire hypertable to the access node.

Related to https://github.com/timescale/timescaledb/issues/4305

On the example query from the above issue, this plan gives about 2x speedup, but in practice this can be arbitrarily high for a join with high selectivity and a large distributed hypertable.

Disable-check: commit-count

PR #4827, opened 2022-10-14
Use prepared statements for parameterized data node queries

akuzm:
This + `plan_cache_mode = force_generic_plan` avoids planning the remote query anew, which is important for cases when there are a lot of loops. The code is mostly the resurrected old row-by-row fetcher.

A quick benchmark on a query from here: https://www.timescale.com/forum/t/distributed-hypertable-poor-performance-for-simple-bucketed-aggregation/498

Speedup | Time, s | Plan 
---:|---:|:---
-1.4x | 2.8  | Prepared statements (this PR)
1.0x  | 3.9  | Parameterized data node scan with CURSOR (current baseline)   
+1.2x   | 4.7   | No parameterized data node scan with COPY    

Merge after:
* https://github.com/timescale/timescaledb/pull/4826
* https://github.com/timescale/timescaledb/pull/4820
* https://github.com/timescale/timescaledb/pull/5035


svenklemm (approved):

tsl/src/remote/connection.c
```diff
@@ -708,6 +712,7 @@ static const char *default_connection_options[] = {
 	"SET datestyle = ISO",
 	"SET intervalstyle = postgres",
 	"SET extra_float_digits = 3",
+	"SET plan_cache_mode = force_generic_plan",
```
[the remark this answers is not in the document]
akuzm: I'll try to set it only for the parameterized scans.

tsl/src/remote/connection.c
```diff
@@ -326,6 +328,8 @@ fill_result_error(TSConnectionError *err, int errcode, const char *errmsg, const
 	err->remote.stmtpos = get_error_field_copy(res, PG_DIAG_STATEMENT_POSITION);
 	if (err->remote.msg == NULL)
 		err->remote.msg = pstrdup(PQresultErrorMessage(res));
+	if (err->remote.msg == NULL)
+		err->remote.msg = pstrdup(PQerrorMessage(entry->conn->pg_conn));
```
akuzm, line 332: This won't work, should either change to length check or remove. I think this was needed for error reporting when the data node segfaults.

tsl/src/remote/prepared_statement_fetcher.c
```diff
@@ -0,0 +1,402 @@
...
+	/*
+	 * Force using the generic plan for each execution of the data node query,
+	 * because it would be very expensive and pointless to replan it for each
+	 * subsequent parameter value. The cleanup will be done by the remote
+	 * connection layer when it initializes a remote connection. This way we
+	 * avoid an extra network request on an erroneous exit or
+	 * cancellation, which would be both prone to cascading errors and
+	 * time-consuming.
+	 */
+	PGresult *res = remote_connection_exec(conn, "SET plan_cache_mode = 'force_generic_plan'");
```
pmwkaa, line 331: When we execute it, is it on first connection to the DN? seems like expensive to do

tsl/src/remote/data_fetcher.h
```diff
@@ -20,17 +20,33 @@ typedef struct DataFetcher DataFetcher;
 
 typedef struct DataFetcherFuncs
```
pmwkaa, line 21: That looks nice, do we need to refactor other fetchers to follow this interface?

pmwkaa (approved):
Looks nice, you can ignore my nit comments

tsl/src/remote/prepared_statement_fetcher.c
akuzm: This code runs only when the fetcher is created, basically once per top-level query that the user runs on the AN.

tsl/src/remote/data_fetcher.h
akuzm: They can't do anything smart for the new method we're adding, "rescan on changed parameters". COPY can't be parameterized at all, and cursors can't be prepared. So they continue to work as before, only needed some superficial changes.

tsl/src/remote/prepared_statement_fetcher.c
pmwkaa: So we send an additional command for each query each time? Does it makes sense to keep a state per connection and send it only if necessary to reset. We are doing something similar already for connection options

PR #4829, opened 2022-10-15
USE C function for time_bucket() offset

SachinSetiya:
Instead of using SQL UDF for handling offset parameter added ts_timestamp/tz/date_offset_bucket() which will handle offset.


svenklemm (changes requested):
Datum and Timestamp/Timestamptz are not interchangeable and you need to use the proper macro to convert between them. While code will work on amd64 it will probably segfault on 32bit architectures.

SachinSetiya:
Okay, will chnage

src/time_bucket.c
```diff
@@ -281,6 +309,30 @@ ts_timestamptz_bucket(PG_FUNCTION_ARGS)
...
+    {
+        timestamp = DirectFunctionCall2(ts_timestamptz_bucket, period, timestamp);
+        PG_RETURN_TIMESTAMPTZ(timestamp);
+    }
+    /* Apply offset. */
+    timestamp = DirectFunctionCall2(timestamptz_mi_interval, timestamp, PG_GETARG_DATUM(2));
+    timestamp = DirectFunctionCall2(ts_timestamptz_bucket, period, timestamp);
+    /* Remove offset. */
+    timestamp = DirectFunctionCall2(timestamptz_pl_interval, timestamp, PG_GETARG_DATUM(2));
+    PG_RETURN_TIMESTAMPTZ(timestamp);
```
svenklemm, line 333:
```suggestion
    Datum timestamp = PG_GETARG_DATUM(1);

    /*
     * Offset can be null
     */
    if (PG_NARGS() <= 2 || PG_ARGISNULL(2))
    {
        timestamp = DirectFunctionCall2(ts_timestamptz_bucket, period, timestamp);
        PG_RETURN_TIMESTAMPTZ(timestamp);
    }
    /* Apply offset. */
    timestamp = DirectFunctionCall2(timestamptz_mi_interval, timestamp, PG_GETARG_DATUM(2));
    timestamp = DirectFunctionCall2(ts_timestamptz_bucket, period, timestamp);
    /* Remove offset. */
    timestamp = DirectFunctionCall2(timestamptz_pl_interval, timestamp, PG_GETARG_DATUM(2));
    PG_RETURN_TIMESTAMPTZ(timestamp);
```

svenklemm (commented):
You need to add these functions to src/func_cache.c

test/expected/query.out
```diff
@@ -286,14 +286,16 @@ BEGIN;
...
+-----------------------------------------------------------------------------------------------------------------
  Limit
    ->  GroupAggregate
-         Group Key: ((time_bucket('@ 1 min'::interval, (_hyper_1_1_chunk."time" - '@ 30 secs'::interval)) + '@ 30 secs'::interval))
-         ->  Result
-               ->  Index Scan using _hyper_1_1_chunk_time_plain on _hyper_1_1_chunk
-(5 rows)
+         Group Key: (time_bucket('@ 1 min'::interval, _hyper_1_1_chunk."time", '@ 30 secs'::interval))
+         ->  Sort
+               Sort Key: (time_bucket('@ 1 min'::interval, _hyper_1_1_chunk."time", '@ 30 secs'::interval)) DESC
```
svenklemm, line 295: why is it no longer using index here?

svenklemm: nvm i think the func_cache changes actually fix this

.github/workflows/linux-32bit-build-and-test.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       options: --privileged --ulimit core=-1
       env:
         DEBIAN_FRONTEND: noninteractive
-        IGNORES: append-* debug_notice transparent_decompression-* transparent_decompress_chunk-* plan_skip_scan-12 pg_dump
+        IGNORES: append-* debug_notice transparent_decompression-* transparent_decompress_chunk-* plan_skip_scan-12 pg_dump query
```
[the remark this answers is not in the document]
SachinSetiya: By looking at 32 bit runs , query test is passing, So I will remove it.

src/time_bucket.c
```diff
@@ -234,6 +234,25 @@ ts_timestamp_bucket(PG_FUNCTION_ARGS)
...
 
+TS_FUNCTION_INFO_V1(ts_timestamp_offset_bucket);
+
+TSDLLEXPORT Datum
+ts_timestamp_offset_bucket(PG_FUNCTION_ARGS)
+{
+	Datum period = PG_GETARG_DATUM(0);
+	Datum timestamp = PG_GETARG_DATUM(1);
+
+	if (TIMESTAMP_NOT_FINITE(DatumGetTimestamp(timestamp)))
```
jnidzwetzki, line 245: Could you add a comment why this test is needed? And is it possible to construct a test case for this?

src/time_bucket.c
```diff
@@ -402,6 +437,25 @@ ts_date_bucket(PG_FUNCTION_ARGS)
...
 }
 
+TS_FUNCTION_INFO_V1(ts_date_offset_bucket);
+
+TSDLLEXPORT Datum
+ts_date_offset_bucket(PG_FUNCTION_ARGS)
+{
+	Datum period = PG_GETARG_DATUM(0);
+	Datum date = PG_GETARG_DATUM(1);
+	/* Apply offset. */
```
jnidzwetzki, line 447: Nit: The other functions are formatted in a slightly different way (e.g., you have an empty line before `/* Apply offset. */` and no empty line before `/* Remove offset. */`). Could you use consistent formatting? Then it is easier to spot the differences between the functions.

jnidzwetzki (commented):
I have two comments, the rest looks good to me.

src/time_bucket.c
SachinSetiya: We have the similar code in ts_timestamp_bucket and relevant functions , and there are tests for these cases in timestamp.sql

src/time_bucket.c
SachinSetiya: Changed

src/time_bucket.c
jnidzwetzki: I was asking for a test case because when I did the review, the return in Line 246 was not covered by a test case. Currently, there are failing tests for this PR, and the recent test coverage is not reported. Could you look at the test coverage when the tests have been run successfully? If the current tests do not cover the return, please add an appropriate test case.

SachinSetiya: Okay

SachinSetiya: I have added the test cases

jnidzwetzki (approved):

PR #4830, opened 2022-10-15
Add function to alter data nodes

### LINKED ISSUE #2104: Divert reads away from data nodes
When a node is unavailable, we should not only block new chunks from being created on it; we should also divert away reads from it. In other words, we should make sure the we set another "primary" data node for the chunks no the data node that is unavailable. The primary data node for a chunk is the foreign server referenced in the chunks foreign table.

erimatnor:
Add a new function, `alter_data_node()`, which can be used to change the data node's configuration originally set up via `add_data_node()` on the access node.

The new functions introduces a new option "available" that allows configuring the availability of the data node. Setting
`available=>false` means that the node should no longer be used for reads and writes. Only read "failover" is implemented as part of this change, however.

To fail over reads, the alter data node function finds all the chunks for which the unavailable data node is the "primary" query target and "fails over" to a chunk replica on another data node instead. If some chunks do not have a replica to fail over to, a warning will be raised.

When a data node is available again, the function can be used to switch back to using the data node for queries.

Closes #2104

### DIFF
--- a/tsl/src/chunk.c
+++ b/tsl/src/chunk.c
@@ -13,6 +13,7 @@
 #include <access/htup_details.h>
 #include <access/xact.h>
 #include <nodes/makefuncs.h>
+#include <nodes/parsenodes.h>
 #include <utils/acl.h>
 #include <utils/builtins.h>
 #include <utils/syscache.h>
@@ -36,6 +37,7 @@
 #include <errors.h>
 #include <error_utils.h>
 #include <hypertable_cache.h>
+#include "hypercube.h"
 
 #include "chunk.h"
 #include "chunk_api.h"
@@ -44,6 +46,7 @@
 #include "dist_util.h"
 #include "remote/dist_commands.h"
 #include "ts_catalog/chunk_data_node.h"
+#include "utils.h"
 
 static bool
 chunk_match_data_node_by_server(const Chunk *chunk, const ForeignServer *server)
@@ -66,7 +69,7 @@ chunk_match_data_node_by_server(const Chunk *chunk, const ForeignServer *server)
 }
 
 static bool
-chunk_set_foreign_server(Chunk *chunk, ForeignServer *new_server)
+chunk_set_foreign_server(const Chunk *chunk, const ForeignServer *new_server)
 {
 	Relation ftrel;
 	HeapTuple tuple;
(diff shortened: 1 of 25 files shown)


tsl/test/expected/data_node.out
```diff
@@ -1579,3 +1579,216 @@ DROP DATABASE :DN_DBNAME_3;
...
+(1 row)
+
+INSERT INTO hyper1
+SELECT t, (abs(timestamp_hash(t::timestamp)) % 3) + 1, random() * 30
+FROM generate_series('2022-01-01 00:00:00'::timestamptz, '2022-01-05 00:00:00', '1 h') t;
+INSERT INTO hyper2 SELECT * FROM hyper1;
+INSERT INTO hyper3 SELECT * FROM hyper1;
+-- create view to see the data nodes and default data node of all
+-- chunks
+CREATE TEMPORARY VIEW chunk_default_data_node AS
```
nikkhils, line 1636: Does it make sense to create this view in system catalog?

nikkhils (approved):
Overall looks good to me barring one comment!

tsl/test/expected/data_node.out
erimatnor: Probably in timescaledb_information or timescaledb_experimental, but let's make a separate PR in that case.

tsl/src/data_node.c
```diff
@@ -1416,6 +1429,279 @@ data_node_detach(PG_FUNCTION_ARGS)
...
+	server = data_node_get_foreign_server(node_name, ACL_NO_CHECK, false, false /* missing_ok */);
+
+	if (host == NULL && database == NULL && port == -1 && arropts == NULL)
+		PG_RETURN_DATUM(
+			HeapTupleGetDatum(create_alter_data_node_tuple(tupdesc, node_name, server->options)));
+
+	current_options = list_copy(server->options);
+
+	if (host != NULL)
+		options = append_data_node_option(options,
```
pmwkaa, line 1620: Thats seems a bit odd that we need to keep both arguments and array of options, where we put those arguments in. Why not to just have an array of options then? Or remove it completely

tsl/src/data_node.c
```diff
@@ -1416,6 +1429,279 @@ data_node_detach(PG_FUNCTION_ARGS)
...
+ * Switch data node to use for queries on chunks.
+ *
+ * When reset=false it will switch from the given data node to another one,
+ * but only if the data node is currently used for queries on the chunk.
+ *
+ * When reset=true it will switch to the given data node, if it is "primary"
+ * for the chunk (according to the current partitioning configuration).
+ */
+static void
+switch_data_node_on_chunks(const ForeignServer *datanode, bool reset)
```
pmwkaa, line 1503: I am finding it a bit difficult to follow the `reset` semantics here, why not to call it `available`everywhere starting from the intitial call in the `data_node_alter()`?

pmwkaa (commented):
Overall looks good, but I found it a bit clunky that we have two ways of defining options now. I would rather prefer either we do define everything using array or just use arguments and remove array. Also to mention, that `available` argument is not presented as an argument and moved to options

tsl/src/data_node.c
erimatnor: I guess I figured the parameter is semantically about what the function should do and not the status of the data node.

src/bgw/job.c
```diff
@@ -1397,10 +1397,10 @@ ts_bgw_job_validate_schedule_interval(Interval *schedule_interval)
...
 	if (has_month && (has_day || has_time))
 		ereport(ERROR,
-				errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-				errmsg("month intervals cannot have day or time component"),
-				errdetail("Fixed schedule jobs do not support such schedule intervals."),
-				errhint("Express the interval in terms of days or time instead."));
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("month intervals cannot have day or time component"),
+				 errdetail("Fixed schedule jobs do not support such schedule intervals."),
+				 errhint("Express the interval in terms of days or time instead.")));
```
nikkhils, line 1403: totally unrelated change to the PR.

tsl/src/data_node.c
nikkhils: Agreed on the common options being individual arguments like "port", "host" etc.

nikkhils (approved):
LGTM overall.

tsl/src/data_node.c
erimatnor: After discussion, decided to make "available" a parameter and punt on advanced options array.

tsl/src/data_node.c
erimatnor: Fixed.

src/bgw/job.c
erimatnor: I know, but it gave me compilation warnings/errors.

tsl/src/data_node.c
```diff
@@ -1415,6 +1428,240 @@ data_node_detach(PG_FUNCTION_ARGS)
...
+				 errmsg("function returning record called in context "
+						"that cannot accept type record")));
+
+	tupdesc = BlessTupleDesc(tupdesc);
+
+	/* Check if a data node with the given name actually exists, or raise an error. */
+	server = data_node_get_foreign_server(node_name, ACL_NO_CHECK, false, false /* missing_ok */);
+
+	if (host == NULL && database == NULL && port == -1 && available_is_null)
+		PG_RETURN_DATUM(
```
pmwkaa, line 1617: If no arguments supplied, it will return current state?

tsl/test/sql/data_node.sql
```diff
@@ -777,3 +779,161 @@ DROP DATABASE :DN_DBNAME_3;
...
+\set ON_ERROR_STOP 1
+
+-- query some data from all hypertables to show its working before
+-- simulating the node being down
+SELECT time, location FROM hyper1 ORDER BY time LIMIT 1;
+SELECT time, location FROM hyper2 ORDER BY time LIMIT 1;
+SELECT time, location FROM hyper3 ORDER BY time LIMIT 1;
+SELECT time, location FROM hyper_1dim ORDER BY time LIMIT 1;
+
+-- simulate a node being down by renaming the database for
```
pmwkaa, line 838: Maybe I am missing it, but do we have a test where try to mark all the nodes as unavailable?

pmwkaa (approved):
Looks good.

The only thing that I find a bit odd that we can't mark all nodes unavailable, which is a bit counter intuitive. If the alter function produces and error, this is something that should be handled by the orchestration service on top of it.

I guess the main reason for that is that we can lose data if we drop a last chunk during write. I am just thinking if this should be simpler, like allow to mark all nodes unavailable, but do not allow do drop chunks in case there is only 1 copy left or something.

tsl/test/sql/data_node.sql
```diff
@@ -777,3 +779,161 @@ DROP DATABASE :DN_DBNAME_3;
...
+WHERE hypertable_name IN ('hyper3', 'hyper_1dim')
+AND range_start::timestamptz <= '2022-01-10 00:00:00'
+AND range_end::timestamptz > '2022-01-10 00:00:00'
+ORDER BY 1, 2;
+
+-- re-enable the data node and the chunks should "switch back" to
+-- using the data node. However, only the chunks for which the node is
+-- "primary" should switch to using the data node for queries
+ALTER DATABASE data_node_1_unavailable RENAME TO :DN_DBNAME_1;
+SELECT * FROM alter_data_node('data_node_1', available=>true);
```
nikkhils, line 889: The following output doesn't make sense IMO when we make a DN available

```
db_data_node=# SELECT * FROM alter_data_node('data_node_2', available=>true);
WARNING:  insufficient number of data nodes
DETAIL:  There are not enough data nodes to replicate chunks according to the configured replication factor.
HINT:  Attach 1 or more data nodes to hypertable "hyper_1dim".
WARNING:  insufficient number of data nodes
DETAIL:  There are not enough data nodes to replicate chunks according to the configured replication factor.
HINT:  Attach 1 or more data nodes to hypertable "hyper_1dim".
WARNING:  insufficient number of data nodes
DETAIL:  There are not enough data nodes to replicate chunks according to the configured replication factor.
HINT:  Attach 1 or more data nodes to hypertable "hyper_1dim".
```

tsl/src/chunk.c
```diff
@@ -134,32 +137,134 @@ chunk_set_foreign_server(Chunk *chunk, ForeignServer *new_server)
...
+			/* For hypertables without a space partition, use the data node
+			 * assignment logic to figure out whether to use the data node as
+			 * query data node. The "default" query data node is the first in
+			 * the list. The chunk assign logic only returns available data
+			 * nodes. */
+			List *datanodes = ts_hypertable_assign_chunk_data_nodes(ht, chunk->cube);
+			const char *node_name = linitial(datanodes);
+			server = GetForeignServerByName(node_name, false);
+
+			if (strcmp(server->servername, node_name) == 0)
```
nikkhils, line 234: clearly this a bug @erimatnor? 

The node_name and server->servername will always match :-D

I believe we need

should_switch_data_node = (server->serverid == data_node_id);

erimatnor: Yes, it should compare serverid with the function parameter data_node_id

tsl/src/data_node.c
erimatnor: yes, that's the intention.

tsl/test/sql/data_node.sql
erimatnor: I can add one, but what is it that you'd like to see from that test?

pmwkaa: Nothing concrete, seems like an edge case

PR #4840, opened 2022-10-17
Fix performance regressions in the copy code

jnidzwetzki:
In 8375b9aa536a619a5ac2644e0dae3c25880a4ead, a patch was added to handle chunks closes during an ongoing copy operation. However, this patch introduces a performance regression. All `MultiInsertBuffers` are deleted after they are flushed. In this PR, the performance regression is fixed. The most commonly used `MultiInsertBuffers` survive flushing. 

The 51259b31c4c62b87228b059af0bbf28caa143eb3 commit changes the way the per-tuple context is used. Since this commit, more objects are stored in this context. The size of the context was used to set the tuple size to PG < 14. The extra objects in the context lead to wrong (very large) results and flushes almost after every tuple read.

The cache synchronization introduced in 296601b1d7aba7f23aea3d47c617e2d6df81de3e is reverted. With the current implementation, `MAX_PARTITION_BUFFERS` survive the flash. If `timescaledb.max_open_chunks_per_insert` is lower than `MAX_PARTITION_BUFFERS` , a buffer flush would be performed after each tuple read.

# Evaluation
The evaluation is performed using PG 14.

## 2.6.1 (Version Before Regression)
```
Report for benchmark suite 'copy'
+-------------------------------------------------------------------------------------------------------------------+----------+
| Query                                                                                                             | 9ae47c68 |
+-------------------------------------------------------------------------------------------------------------------+----------+
| TRUNCATE sensor_data;                                                                                             |    21.72 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_sensor_data.csv' WITH CSV HEADER /* initial copy */;          | 46357.29 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_sensor_data.csv' WITH CSV HEADER /* second copy */;           | 50986.47 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_time_data.csv' WITH CSV HEADER /* initial copy */;            | 44632.01 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_time_data.csv' WITH CSV HEADER /* second copy */;             | 51910.59 |
+-------------------------------------------------------------------------------------------------------------------+----------+
```

## 2.8.1 (Version With Regression)
```
Report for benchmark suite 'copy'
+-------------------------------------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                                             | 65078486025a5cfdd00663f524831619066c5946 |
+-------------------------------------------------------------------------------------------------------------------+------------------------------------------+
| TRUNCATE sensor_data;                                                                                             |                                    23.90 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_sensor_data.csv' WITH CSV HEADER /* initial copy */;          |                                 47655.08 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_sensor_data.csv' WITH CSV HEADER /* second copy */;           |                                 52555.64 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_time_data.csv' WITH CSV HEADER /* initial copy */;            |                                 49493.31 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_time_data.csv' WITH CSV HEADER /* second copy */;             |                                 53489.15 |
+-------------------------------------------------------------------------------------------------------------------+------------------------------------------+
```

## PR Version
```
Report for benchmark suite 'copy'
+-------------------------------------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                                             | d55df01a5c8b2875835cf56d8c63dd2fd1be3f8d |
+-------------------------------------------------------------------------------------------------------------------+------------------------------------------+
| TRUNCATE sensor_data;                                                                                             |                                    22.82 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_sensor_data.csv' WITH CSV HEADER /* initial copy */;          |                                 31970.58 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_sensor_data.csv' WITH CSV HEADER /* second copy */;           |                                 36304.62 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_time_data.csv' WITH CSV HEADER /* initial copy */;            |                                 30163.68 |
| COPY sensor_data FROM '%%WORK_DIR%%/copy_ordered_by_time_data.csv' WITH CSV HEADER /* second copy */;             |                                 34985.83 |
+-------------------------------------------------------------------------------------------------------------------+------------------------------------------+
```

#### Input data sorted by sensor_data, time
copy1

#### Input data sorted by time, sensor_data
copy2

**Note:** Only if the `timescaledb.max_cached_chunks_per_hypertable` and `timescaledb.max_open_chunks_per_insert` parameters are decreased together, the execution time increases significantly. If only one of the parameters is decreased, the chunks do not need to be closed or can be found in the chunk cache.

### DIFF
--- a/src/copy.c
+++ b/src/copy.c
@@ -85,6 +85,9 @@
  */
 #define MAX_BUFFERED_BYTES 65535
 
+/* Trim the list of buffers back down to this number after flushing */
+#define MAX_PARTITION_BUFFERS 32
+
 /* Stores multi-insert data related to a single relation in CopyFrom. */
 typedef struct TSCopyMultiInsertBuffer
 {
@@ -151,10 +154,8 @@ on_chunk_insert_state_changed(ChunkInsertState *state, void *data)
 {
 	BulkInsertState bistate = data;
 
-	/* Different chunk so must release BulkInsertState */
-	if (bistate->current_buf != InvalidBuffer)
-		ReleaseBuffer(bistate->current_buf);
-	bistate->current_buf = InvalidBuffer;
+	/* Chunk changed, so release the buffer held in BulkInsertState */
+	ReleaseBulkInsertStatePin(bistate);
 }
 
 static CopyChunkState *
@@ -266,16 +267,6 @@ TSCopyMultiInsertInfoIsFull(TSCopyMultiInsertInfo *miinfo)
 		miinfo->bufferedBytes >= MAX_BUFFERED_BYTES)
 		return true;
 
-	if (hash_get_num_entries(miinfo->multiInsertBuffers) >= ts_guc_max_open_chunks_per_insert)
-	{
-		/*
-		 * Flushing each multi-insert buffer will require looking up the
-		 * corresponding chunk insert state in the cache, so don't accumulate
-		 * more inserts than the cache can fit, to avoid thrashing.
-		 */
-		return true;
-	}
-
 	return false;
 }
 
@@ -291,7 +282,7 @@ TSCopyMultiInsertInfoIsEmpty(TSCopyMultiInsertInfo *miinfo)
 /*
  * Write the tuples stored in 'buffer' out to the table.
  */
-static inline void
+static inline int
 TSCopyMultiInsertBufferFlush(TSCopyMultiInsertInfo *miinfo, TSCopyMultiInsertBuffer *buffer)
 {
 	MemoryContext oldcontext;
@@ -318,11 +309,15 @@ TSCopyMultiInsertBufferFlush(TSCopyMultiInsertInfo *miinfo, TSCopyMultiInsertBuf
 	 * table is closed and pointers (e.g., result_relation_info point) to invalid
 	 * addresses. Re-reading the chunk insert state ensures that the table is
 	 * open and the pointers are valid.
+	 *
+	 * No callback on changed chunk is needed, the bulk insert state buffer is
+	 * freed in TSCopyMultiInsertBufferCleanup().
 	 */
-	ChunkInsertState *cis = ts_chunk_dispatch_get_chunk_insert_state(miinfo->ccstate->dispatch,
-																	 buffer->point,
-																	 on_chunk_insert_state_changed,
-																	 buffer->bistate);
+	ChunkInsertState *cis =
+		ts_chunk_dispatch_get_chunk_insert_state(miinfo->ccstate->dispatch,
+												 buffer->point,
+												 NULL /* on chunk changed function */,
+												 NULL /* payload for on chunk changed function */);
 
 	ResultRelInfo *resultRelInfo = cis->result_relation_info;
 
@@ -412,6 +407,14 @@ TSCopyMultiInsertBufferFlush(TSCopyMultiInsertInfo *miinfo, TSCopyMultiInsertBuf
 	/* Mark that all slots are free */
 	buffer->nused = 0;
 
+	/* Chunk could be closed on a subsequent call of ts_chunk_dispatch_get_chunk_insert_state
+	 * (e.g., due to timescaledb.max_open_chunks_per_insert). So, ensure the bulk insert is
+	 * finished after the flush is complete.
+	 */
+	ResultRelInfo *result_relation_info = cis->result_relation_info;
+	Assert(result_relation_info != NULL);
+	table_finish_bulk_insert(result_relation_info->ri_RelationDesc, miinfo->ti_options);
+
 	/* Reset cur_lineno and line_buf_valid to what they were */
 #if PG14_GE
 	if (cstate != NULL)
@@ -420,6 +423,8 @@ TSCopyMultiInsertBufferFlush(TSCopyMultiInsertInfo *miinfo, TSCopyMultiInsertBuf
 		cstate->cur_lineno = save_cur_lineno;
 	}
 #endif
+
+	return cis->chunk_id;
 }
 
 /*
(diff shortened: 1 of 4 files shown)


akuzm:
A high-level question: why is the CIM_MULTI path slow? Maybe we can make it always fast and always use it? As far as I understand from briefly reading this PR, we stop using this in some scenarios where it has inferior performance. When we have several code path with different performance characteristics, and use complex conditions to switch between them, it becomes hard to reason about the performance, and benchmarking becomes more dependent on a particular data set.

akuzm:
I mean, even before this patch it was complex enough already, I was completely puzzled by that the COPY throughput is significantly higher with random input order than with sequential. I remember seeing a lot of creating/destroying tuple slots or something in the flame graph, but couldn't figure out what was wrong. Maybe there's some bug there that we can fix and get better performance...

akuzm:
> I remember seeing a lot of creating/destroying tuple slots or something in the flame graph, but couldn't figure out what was wrong. Maybe there's some bug there that we can fix and get better performance...

https://sendeyo.com/get/d/7053bff117?x=371.5&y=1685

Found the flamegraph. Why is it spending so much time in ResourceOwner? Looks suspicious... This is for insert ordered by time, and the insert in random order is faster, main difference being that it doesn't spend so much time in ResourceOwner.

src/copy.c
```diff
@@ -840,6 +929,19 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, MemoryConte
 				(errmsg("Using normal unbuffered copy operation (CIM_SINGLE) "
 						"because triggers are defined on the destination table.")));
 	}
+	else if (Max(ts_guc_max_open_chunks_per_insert, ts_guc_max_cached_chunks_per_hypertable) <
+			 MIN_CHUNKS_FOR_MULTI_COPY)
```
akuzm, line 933: I'm not sure we should allow such low values of these caches at all, besides for testing. Maybe we should ereport a warning here.

src/copy.c
```diff
@@ -85,6 +85,15 @@
  */
 #define MAX_BUFFERED_BYTES 65535
 
+/* Trim the list of buffers back down to this number after flushing */
+#define MAX_PARTITION_BUFFERS 32
+
```
akuzm, line 90: Why not keep all of them? This doesn't influence the max memory usage of the query, and saves the work of (de)allocating these buffers back and forth. I think max memory usage is the most important because this is what determines if the query fails with OOM. Transiently reducing the memory usage helps somewhat, but I'm not sure it's worth the work.

src/copy.c
akuzm: Another argument against this is that it makes testing stuff such as reusing insert buffers much more difficult.

jnidzwetzki: The copy code has two code paths: 

* `CIM_SINGLE` for chunks that use features (e.g., triggers) that require a sequential order for insertion. 
* `CIM_MULTI_CONDITIONAL` is used when multi-inserts are possible. 

So, we have had two code paths so far, which are used depending on the configuration of the table.

This new condition ensures that `CIM_SINGLE` is used when only a few chunks can be opened (and cached) in parallel. The overhead for maintaining the multi-insert buffer is higher in this case than using direct bulk inserts (see attached graphs). 

The condition is deterministic and can be tested by setting the parameters in the test properly. In the test, you can easily check whether multi-buffer optimization is used or not by setting the log level to DEBUG1. We also already have tests that verify the used code path in that way.

src/copy.c
jnidzwetzki: We should definitely delete the buffers after some time. Imagine you have a copy operation from STDIN running which ingests data from a stream over days. The buffers for the older chunks are at some point no longer needed and should not lead to an OOM.

src/copy.c
akuzm: Let me expand on the test coverage stuff. Say, I want to test re-using insert buffer for a different chunk. Now I can set `ts_guc_max_open_chunks_per_insert = 1` and insert two lines. But if we disable CIM_MULTI for small number of chunks, I'll have to make a bigger data set for this. Why not allow it with a warning? Such low values are not a valid configuration anyway, but useful for tests.

src/copy.c
akuzm: Oh, they are actually fixed after a specific chunk... I had a wrong understanding of this. Then we have to remove them eventually indeed.

akuzm (approved):
The only thing I'd change is remove the auto-switch to CIM_MULTI, because it has no value for production use (the very low GUC values are not a sane configuration), and complicates understanding and testing.

jnidzwetzki:
@akuzm The auto-switch to `CIM_SINGLE` based on configuration parameters is removed in the current version of the PR.

src/copy.c
```diff
@@ -431,37 +436,119 @@ static inline void
...
+{
+	int b1 = ((const TSCopyMultiInsertBuffer *) lfirst(a))->nused;
+	int b2 = ((const TSCopyMultiInsertBuffer *) lfirst(b))->nused;
+
+	Assert(b1 >= 0);
+	Assert(b2 >= 0);
+
+	return (b1 > b2) ? 1 : (b1 == b2) ? 0 : -1;
+}
+#endif
```
fabriziomello, line 479:
```suggestion
/* list_sort comparator to sort TSCopyMultiInsertBuffer by usage */
#if PG13_LT
static int
TSCmpBuffersByUsage(const void *a, const void *b)
{
	int b1 = ((const TSCopyMultiInsertBuffer *) lfirst(*(ListCell **) a))->nused;
	int b2 = ((const TSCopyMultiInsertBuffer *) lfirst(*(ListCell **) b))->nused;
#else
static int
TSCmpBuffersByUsage(const ListCell *a, const ListCell *b)
{
	int b1 = ((const TSCopyMultiInsertBuffer *) lfirst(a))->nused;
	int b2 = ((const TSCopyMultiInsertBuffer *) lfirst(b))->nused;
#endif

	Assert(b1 >= 0);
	Assert(b2 >= 0);

	return (b1 > b2) ? 1 : (b1 == b2) ? 0 : -1;
}
```
Nit. Just a simplification. It's up to you.

fabriziomello (approved):

src/copy.c
jnidzwetzki: I thought about writing it that way too. But I decided to write the more verbose version to improve readability and stay consistent with other existing compare functions like [this one](https://github.com/timescale/timescaledb/blob/40a6c4cf87edb3ca261cebcf4b2c6af11fcc6f0e/src/compat/compat.h#L560).

PR #4842, opened 2022-10-17
Fix the flaky pg_dump test

akuzm:
It was frequently failing on Windows. Sort by what is actually printed.


pmwkaa (approved):

erimatnor (approved):
So, if I understand the issue correctly, it was sorting on the numerical OID (in text format)?

akuzm:
> So, if I understand the issue correctly, it was sorting on the numerical OID (in text format)?

yep

PR #4843, opened 2022-10-17
Bump github action versions

fabriziomello:
All github actions that run on Node12 are deprecated so bumped github action versions to run on Node16.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/abi.yaml
+++ b/.github/workflows/abi.yaml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
     - name: Checkout source code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
     - name: Read configuration
       id: config
       run: python .github/gh_config_reader.py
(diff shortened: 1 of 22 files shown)


jnidzwetzki (approved):
I'm not sure if we use more actions that need to be updated as well. But this is at least a good start. 👍

fabriziomello:
> I'm not sure if we use more actions that need to be updated as well. But this is at least a good start. +1

I've also bumped the `actions/upload-artifact` and `actions/setup-python`.

svenklemm (commented):
windows-build-and-test.yaml also has a cache action that the PR missed

svenklemm (approved):

fabriziomello:
> windows-build-and-test.yaml also has a cache action that the PR missed

Fixed, thanks!

PR #4851, opened 2022-10-18
Add clang-tidy warning readability-suspicious-call-argument

akuzm:
Helps find accidentally swapped arguments, like in the recent epoll_ctl() error.

Disable-check: commit-count

### DIFF
--- a/tsl/src/continuous_aggs/invalidation.c
+++ b/tsl/src/continuous_aggs/invalidation.c
@@ -552,11 +552,11 @@ invalidation_entry_reset(Invalidation *entry)
  * invalidations.
  */
 static void
-invalidation_expand_to_bucket_boundaries(Invalidation *inv, Oid timetype, int64 bucket_width,
+invalidation_expand_to_bucket_boundaries(Invalidation *inv, Oid time_type_oid, int64 bucket_width,
 										 const ContinuousAggsBucketFunction *bucket_function)
 {
-	const int64 min_for_type = ts_time_get_min(timetype);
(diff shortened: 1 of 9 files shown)


tsl/src/continuous_aggs/invalidation.c
```diff
@@ -552,11 +552,11 @@ invalidation_entry_reset(Invalidation *entry)
...
  */
 static void
-invalidation_expand_to_bucket_boundaries(Invalidation *inv, Oid timetype, int64 bucket_width,
+invalidation_expand_to_bucket_boundaries(Invalidation *inv, Oid time_type_oid, int64 bucket_width,
 										 const ContinuousAggsBucketFunction *bucket_function)
 {
-	const int64 min_for_type = ts_time_get_min(timetype);
-	const int64 max_for_type = ts_time_get_max(timetype);
+	const int64 timestamp_min = ts_time_get_min(time_type_oid);
+	const int64 timestamp_max = ts_time_get_max(time_type_oid);
```
fabriziomello, line 559:
```suggestion
	const int64 time_dimension_min = ts_time_get_min(time_type_oid);
	const int64 time_dimension_max = ts_time_get_max(time_type_oid);
```
IMHO I prefer to not use `timestamp` in those variable names because the time dimension can be different datatypes (DATE, TIMESTAMP, INTEGER, etc).

akuzm: Thanks, this name works. clang-tidy is surprisingly picky about these names so I had a hard time choosing the new one.

PR #4855, opened 2022-10-19
Add telemetry for fixed schedule jobs

konskov:
Previous commit #4664 introduced the ability to execute background jobs on a fixed schedule.
This commit updates our telemetry data to include the number of jobs scheduled to execute on a fixed schedule vs the number registered to execute on a drifting schedule.

### DIFF
--- a/src/telemetry/telemetry.c
+++ b/src/telemetry/telemetry.c
@@ -59,6 +59,11 @@
 #define REQ_BUILD_ARCHITECTURE "build_architecture"
 #define REQ_DATA_VOLUME "data_volume"
 
+#define REQ_NUM_POLICY_CAGG_FIXED "num_continuous_aggs_policies_fixed"
+#define REQ_NUM_POLICY_COMPRESSION_FIXED "num_compression_policies_fixed"
+#define REQ_NUM_POLICY_REORDER_FIXED "num_reorder_policies_fixed"
+#define REQ_NUM_POLICY_RETENTION_FIXED "num_retention_policies_fixed"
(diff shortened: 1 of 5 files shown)


src/telemetry/telemetry.c
```diff
@@ -226,10 +254,15 @@ add_job_counts(JsonbParseState *state)
 	BgwJobTypeCount counts = bgw_job_type_counts();
 
 	ts_jsonb_add_int32(state, REQ_NUM_POLICY_CAGG, counts.policy_cagg);
+	ts_jsonb_add_int32(state, REQ_NUM_POLICY_CAGG_FIXED, counts.policy_cagg_fixed);
```
akuzm, line 257: Does it make sense to define macros for these names? They're used in one place, another layer of indirection just makes it harder to read.

src/telemetry/telemetry.c
```diff
@@ -101,19 +106,42 @@ bgw_job_type_counts()
...
+				else
+					counts.policy_cagg++;
+			}
 			else if (namestrcmp(&job->fd.proc_name, "policy_compression") == 0)
-				counts.policy_compression++;
+			{
+				if (job->fd.fixed_schedule)
+					counts.policy_compression_fixed++;
+				else
+					counts.policy_compression++;
```
akuzm, line 120: A lot of stuff is not covered, but I see that you have these jobs in the test... maybe this is a stale comment by codecov bot.

akuzm (approved):

src/telemetry/telemetry.c
konskov: I think defining the macro follows the same pattern as for the rest of the parameters, so makes sense to do it for that reason.

PR #4860, opened 2022-10-20
Block multi-statement DDL command in one query

### LINKED ISSUE #4818: [Bug] dist analyze: cannot insert multiple commands into a prepared statement
```
$ psql -c "create table metric_dist(ts timestamptz, id int, value float);
select create_distributed_hypertable('metric_dist', 'ts', 'id');"

$ psql -c "analyze metric_dist; analyze metric_dist;"
ERROR:  42601: [data2]: cannot insert multiple commands into a prepared statement
LOCATION:  async_response_report_error, async.c:418
```
Looks like the distributed ddl handling is broken.

pmwkaa:
Ensure that queries involving several distributed DDL commands in one query string are blocked.

Fix #4818

### DIFF
--- a/tsl/src/remote/dist_ddl.c
+++ b/tsl/src/remote/dist_ddl.c
@@ -117,6 +117,21 @@ dist_ddl_error_if_not_allowed_data_node_session(void)
 }
 
 static void
+dist_ddl_error_if_multi_command(const ProcessUtilityArgs *args)
+{
+	List *parsetree_list;
+
+	/* Parse the SQL string into a list of raw parse trees */
+	parsetree_list = pg_parse_query(args->query_string);
+
+	/* Prevent 'command;command' execution */
+	if (list_length(parsetree_list) != 1)
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("nested commands are not supported on distributed hypertable")));
+}
+
+static void
 dist_ddl_state_add_remote_command(const char *cmd)
 {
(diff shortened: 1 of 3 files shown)


tsl/src/remote/dist_ddl.c
```diff
@@ -116,6 +116,19 @@ dist_ddl_error_if_not_allowed_data_node_session(void)
...
+dist_ddl_error_if_multi_command(const ProcessUtilityArgs *args)
+{
+	List *parsetree_list;
+
+	/* Parse the SQL string into a list of raw parse trees */
+	parsetree_list = pg_parse_query(args->query_string);
+
+	/* Prevent 'command;command' execution */
+	if (list_length(parsetree_list) != 1)
+		dist_ddl_error_raise_unsupported();
```
akuzm, line 129: `ERROR:  operation not supported on distributed hypertable` sounds misleading. ANALYZE is supported, multi-commands aren't.

akuzm: Why not say "multiple commands are not supported on distributed tables"? That would be the correct description of the limitation we have. If I run "analyze this; analyze that" and it says "the operation is not supported", the operation here is analyze. That's just wrong.

pmwkaa: I would assume the operation here is the supplied query text to the server. Also saying `multiple commands are not supported on distributed hypertable` is also a bit misleading, because they are supported. Even multi-statement transactions are supported. There are some exceptions of course, like you cannot mix local and distributed hypertables, etc. The problem mostly, because we do not rewrite DDL queries (with some exceptions).

Yet I understand your point, messaging could be better. Let me change it

PR #4867, opened 2022-10-21
Build on Ubuntu 22.04

akuzm:
It has newer GCC which should detect more warnings.


svenklemm:
There seem to be a lot of unrelated changes in this PR. Please also test release-builds when using USED_FOR_ASSERTS_ONLY.

src/chunk.c
```diff
@@ -1911,7 +1911,7 @@ chunk_resurrect(const Hypertable *ht, int chunk_id)
 {
 	ScanIterator iterator;
 	Chunk *chunk = NULL;
-	int count = 0;
+	PG_USED_FOR_ASSERTS_ONLY int count = 0;
```
akuzm, line 1914: This is the new warning reported by clang for builds w/o assertions: the variable is initialized and incremented, but the resulting value is only checked in assertions.

akuzm:
> There seem to be a lot of unrelated changes in this PR

I removed the `Ensure` stuff. There seems to be too much clang-tidy warnings in builds w/o assertions, probably we shouldn't try to enable it at the moment.

akuzm:
Builds fine on CentOS 7 with `gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC)` and `-DUSE_OPENSSL=0 -DWARNINGS_AS_ERRORS=OFF -DLINTER=OFF  -DTAP_CHECKS=OFF -DCMAKE_C_STANDARD=11 -DREGRESS_CHECKS=OFF  -DCMAKE_BUILD_TYPE=RelWithDebInfo`

src/hypertable.c
```diff
@@ -466,8 +466,8 @@ ts_hypertable_scan_with_memory_context(const char *schema, const char *table,
 									   bool tuplock, MemoryContext mctx)
 {
 	ScanKeyData scankey[2];
-	NameData schema_name = { { 0 } };
-	NameData table_name = { { 0 } };
+	NameData schema_name = { .data = { 0 } };
```
akuzm, line 469: clang-15 doesn't like double nested braces, and gcc 4.8.5 doesn't like single braces, so this is how we have to compromise to avoid warnings

akuzm: (for clang-15 this happens with -Wextra that we're probably going to enable soon)

svenklemm (approved):

Issue #4708, opened 2022-09-15
[Bug]: Assertion hit in add_baserel_cache_entry_for_chunk


jnidzwetzki:
### What type of bug is this?

Crash

### What subsystems and features are affected?

Query planner

### What happened?

SQLSmith found the following assertion hit when running with seed `11872767`.

```
[...]
#2  0x000055edd1a3c53c in ExceptionalCondition (conditionName=0x7f16f8ddd02d "entry->type == chunk_reltype", 
    errorType=0x7f16f8ddcc47 "FailedAssertion", fileName=0x7f16f8ddcfa8 "/home/jan/timescaledb/src/planner/planner.c", lineNumber=176)
    at assert.c:69
[...]

(gdb) frame 3
#3  0x00007f16f8da8baf in add_baserel_cache_entry_for_chunk (chunk_reloid=20516, chunk_status=1, hypertable=0x55edd29171a8, 
    chunk_reltype=TS_REL_CHUNK_CHILD) at /home/jan/timescaledb/src/planner/planner.c:176
176			Assert(entry->type == chunk_reltype);
(gdb) print entry->type
$1 = TS_REL_CHUNK
(gdb) print chunk_reltype
$2 = TS_REL_CHUNK_CHILD
````

### TimescaleDB version affected

2.8.0  and 2.9.0-dev (424f6f7648e88907b9373ef44ec851787db21080)

### PostgreSQL version used

14.2

### What operating system did you use?

Debian 11

### What installation method did you use?

Source

### What platform did you run on?

On prem/Self-hosted

### Relevant log output and stack trace

Tested with TimescaleDB 2.8.0

```bash
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f170444b537 in __GI_abort () at abort.c:79
#2  0x000055edd1a3c53c in ExceptionalCondition (conditionName=0x7f16f8ddd02d "entry->type == chunk_reltype", 
    errorType=0x7f16f8ddcc47 "FailedAssertion", fileName=0x7f16f8ddcfa8 "/home/jan/timescaledb/src/planner/planner.c", lineNumber=176)
    at assert.c:69
#3  0x00007f16f8da8baf in add_baserel_cache_entry_for_chunk (chunk_reloid=20516, chunk_status=1, hypertable=0x55edd29171a8, 
[...]

PR #4870, opened 2022-10-21
Do not cache the classify_relation result

akuzm:
It depends on the context, not only on the relation id. The same chunk can be expanded both as a child of hypertable and as an independent table.

Fixes https://github.com/timescale/timescaledb/issues/4708

### DIFF
--- a/src/planner/planner.c
+++ b/src/planner/planner.c
@@ -75,8 +75,6 @@
 typedef struct BaserelInfoEntry
 {
 	Oid reloid;
-	/* Either a chunk or plain baserel (TS_REL_OTHER). */
-	TsRelType type;
 	Hypertable *ht;
 	uint32 chunk_status; /* status of chunk, if this is a chunk */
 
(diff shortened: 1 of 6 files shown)


src/planner/planner.c
```diff
@@ -657,15 +651,16 @@ get_parent_rte(const PlannerInfo *root, Index rti)
  * related metadata: like chunk_status and pointer to hypertable entry.
  * It is okay to cache a pointer to the hypertable, since this cache is
  * confined to the lifetime of the query and not used across queries.
+ * If parent reolid is know, the caller can specify it to avoid the costly
+ * lookup. Otherwise pass InvalidOid.
```
akuzm, line 655: @jnidzwetzki This is the part that should improve planning performance

svenklemm, line 655:
```suggestion
 * If parent reloid is known, the caller can specify it to avoid the costly
 * lookup. Otherwise pass InvalidOid.
```

svenklemm (approved):

PR #4875, opened 2022-10-24
Enable -Wextra

akuzm:
Our code mostly has warnings about assignment/comparison of int types of different signedness.


tsl/src/compression/compression.c
```diff
@@ -1466,7 +1466,7 @@ Datum
 tsl_compressed_data_recv(PG_FUNCTION_ARGS)
 {
 	StringInfo buf = (StringInfo) PG_GETARG_POINTER(0);
-	CompressedDataHeader header = { { 0 } };
+	CompressedDataHeader header = { .vl_len_ = { 0 } };
```
akuzm, line 1469: clang-15 doesn't like double nested braces, and gcc 4.8.5 doesn't like single braces, so this is how we have to compromise to avoid warnings

Also see this PR https://github.com/timescale/timescaledb/pull/4867/files#r1004751330

akuzm:
Doesn't introduce new warnings on gcc 4.8.5 as of [ab0e3ca](https://github.com/timescale/timescaledb/pull/4875/commits/ab0e3ca5132cfea337412be52830aeddde54e0ad)

svenklemm (approved):

jnidzwetzki (approved):

PR #4879, opened 2022-10-25
Add hook for chunk creation

gayyappan:
After data is tiered using OSM, we cannot insert data into the same range. Need a callback that can be invoked by timescaledb to check for range overlaps before creating a new chunk

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -1153,11 +1153,62 @@ ts_chunk_create_only_table(Hypertable *ht, Hypercube *cube, const char *schema_n
 	return chunk;
 }
 
+#if PG14_GE
+#define OSM_CHUNK_INSERT_CHECK_HOOK "osm_chunk_insert_check_hook"
+typedef int (*ts_osm_chunk_insert_hook_type)(Oid ht_oid, int64 range_start, int64 range_end);
+static ts_osm_chunk_insert_hook_type
+get_osm_chunk_insert_hook()
+{
+	ts_osm_chunk_insert_hook_type *func_ptr =
+		(ts_osm_chunk_insert_hook_type *) find_rendezvous_variable(OSM_CHUNK_INSERT_CHECK_HOOK);
+	return *func_ptr;
+}
+#endif
+
 static Chunk *
 chunk_create_from_hypercube_after_lock(const Hypertable *ht, Hypercube *cube,
 									   const char *schema_name, const char *table_name,
 									   const char *prefix)
 {
+#if PG14_GE
+	ts_osm_chunk_insert_hook_type insert_func_ptr = get_osm_chunk_insert_hook();
+
+	if (insert_func_ptr)
+	{
+		/* OSM only uses first dimension . doesn't work with multinode tables yet*/
+		Dimension *dim = &ht->space->dimensions[0];
+		/* convert to PG timestamp from timescaledb internal format */
+		int64 range_start =
+			ts_internal_to_time_int64(cube->slices[0]->fd.range_start, dim->fd.column_type);
+		int64 range_end =
+			ts_internal_to_time_int64(cube->slices[0]->fd.range_end, dim->fd.column_type);
+		int chunk_exists = insert_func_ptr(ht->main_table_relid, range_start, range_end);
+		if (chunk_exists)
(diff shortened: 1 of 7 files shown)


src/chunk.c
```diff
@@ -1153,11 +1153,62 @@ ts_chunk_create_only_table(Hypertable *ht, Hypercube *cube, const char *schema_n
...
+#define OSM_CHUNK_INSERT_CHECK_HOOK "osm_chunk_insert_check_hook"
+typedef int (*TS_OSM_CHUNK_INSERT_HOOK_TYPE)(Oid ht_oid, int64 range_start, int64 range_end);
+static TS_OSM_CHUNK_INSERT_HOOK_TYPE
+get_osm_chunk_insert_hook()
+{
+	TS_OSM_CHUNK_INSERT_HOOK_TYPE *func_ptr =
+		(TS_OSM_CHUNK_INSERT_HOOK_TYPE *) find_rendezvous_variable(OSM_CHUNK_INSERT_CHECK_HOOK);
+	return *func_ptr;
+}
+#endif
```
fabriziomello, line 1166:
```suggestion
#if PG14_GE
#define OSM_CHUNK_INSERT_CHECK_HOOK "osm_chunk_insert_check_hook"

/* chunk insert hook type defintion*/
typedef int (*ts_osm_chunk_insert_hook_type) (Oid ht_oid, int64 range_start, int64 range_end);

static ts_osm_chunk_insert_hook_type
get_osm_chunk_insert_hook()
{
	ts_osm_chunk_insert_hook_type *func_ptr =
		(ts_osm_chunk_insert_hook_type *) find_rendezvous_variable(OSM_CHUNK_INSERT_CHECK_HOOK);
	return *func_ptr;
}
#endif
```
IMHO is better to use a lower case for hook type names.

src/chunk.c
```diff
@@ -1153,11 +1153,62 @@ ts_chunk_create_only_table(Hypertable *ht, Hypercube *cube, const char *schema_n
...
+	if (insert_func_ptr)
+	{
+		/* OSM only uses first dimension . doesn't work with multinode tables yet*/
+		Dimension *dim = &ht->space->dimensions[0];
+		/* convert to PG timestamp from timescaledb internal format */
+		int64 range_start =
+			ts_internal_to_time_int64(cube->slices[0]->fd.range_start, dim->fd.column_type);
+		int64 range_end =
+			ts_internal_to_time_int64(cube->slices[0]->fd.range_end, dim->fd.column_type);
+		int chunk_exists = insert_func_ptr(ht->main_table_relid, range_start, range_end);
```
fabriziomello, line 1185:
```suggestion
		int chunk_exists = insert_func_ptr(ht->main_table_relid, range_start, range_end);

```
Nit.

src/chunk.c
```diff
@@ -1153,11 +1153,62 @@ ts_chunk_create_only_table(Hypertable *ht, Hypercube *cube, const char *schema_n
...
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("distributed hypertable member cannot create chunk on its own"),
+					 errmsg("Cannot insert into tiered chunk range of %s.%s - attempt to create "
+							"new chunk "
+							"with range  [%s %s] failed",
+							NameStr(ht->fd.schema_name),
+							NameStr(ht->fd.table_name),
+							DatumGetCString(OidFunctionCall1(outfuncid, start_ts)),
+							DatumGetCString(OidFunctionCall1(outfuncid, end_ts))),
```
fabriziomello, line 1206: Shouldn't be an `errdetail` ???

fabriziomello (approved):
Left some minor comments, but in any case LGTM!

src/chunk.c
gayyappan: fixed

src/chunk.c
gayyappan: want to see info about the overlap as part of the error.

src/chunk.c
gayyappan: fixed formatting.

svenklemm (approved):

PR #4880, opened 2022-10-26
Fix const null clauses in runtime chunk exclusion

akuzm:
The code we inherited from postgres expects that if we have a const null or false clause, it's going to be the single one, but that's not true for runtime chunk exclusion because we don't try to fold such restrictinfos after evaluating the mutable functions. Fix it to also work for multiple restrictinfos.

Related to https://github.com/timescale/Support-Dev-Collab/issues/620

Disable-check: commit-count

### DIFF
--- a/.github/workflows/linux-build-and-test.yaml
+++ b/.github/workflows/linux-build-and-test.yaml
@@ -196,9 +196,9 @@ jobs:
         sudo coredumpctl gdb <<<"
           set verbose on
           show debug-file-directory
-          printf "query = '%s'\n\n", debug_query_string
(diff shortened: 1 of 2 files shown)


fabriziomello:
@akuzm just an advice, you can ask for support folks to convert the SDC issue into a public issue.

svenklemm (commented):
Why do the counts in the EXPLAIN output change due to this?

akuzm:
> Why do the counts in the EXPLAIN output change due to this?

There was one null value in the `device_id` column on which the subquery is filtered, so now for this rows all chunks are excluded.

svenklemm (approved):

PR #4890, opened 2022-10-27
Do not clobber the baserel cache on UDF error

### LINKED ISSUE #4823: [Bug]: 2.7 crashes when executing multiple nested IMMUTABLE functions (2.6 was fine)
### What type of bug is this?

Crash

Command processing

### What happened?

I previously logged #4489 which caused a segfault when two IMMUTABLE functions were called at top level.

Now I have a similar bug when two IMMUTABLE functions are called in a transaction, but "nested" in another function.

Selecting from a table in the top-level function, having two arguments in the top-level function, etc... all seems relevant. These replication steps are as "boiled down" as I could.

No crash when executing against regular postgresql 13.8, or timescaledb 2.6.2

2.7.0 to 2.8.1 (2.6.2 was fine)

### PostgreSQL version used

13.8

Ubuntu 18.04

Docker

On prem/Self-hosted

```bash
LOG:  terminating any other active server processes
WARNING:  terminating connection because of crash of another server process
DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
HINT:  In a moment you should be able to reconnect to the database and repeat your command.
LOG:  all server processes terminated; reinitializing
LOG:  database system was interrupted; last known up at 2022-10-14 00:31:02 UTC
LOG:  database system was not properly shut down; automatic recovery in progress
LOG:  redo starts at 0/1E0D430
LOG:  invalid record length at 0/1E31780: wanted 24, got 0
LOG:  redo done at 0/1E31748
LOG:  database system is ready to accept connections
LOG:  TimescaleDB background worker launcher connected to shared catalogs
```

```bash
DROP TABLE IF EXISTS valid_ids;
CREATE TABLE valid_ids
(
  id UUID PRIMARY KEY
);

CREATE OR REPLACE FUNCTION DEFAULT_UUID(TEXT DEFAULT '') RETURNS UUID AS $$
  BEGIN
    RETURN COALESCE($1, '')::UUID;
  EXCEPTION WHEN invalid_text_representation THEN
    RETURN '00000000-0000-0000-0000-000000000000';
  END;
$$ LANGUAGE PLPGSQL IMMUTABLE;

CREATE OR REPLACE FUNCTION KNOWN_ID(UUID, TEXT) RETURNS UUID AS $$
  SELECT COALESCE(
    (SELECT id FROM valid_ids WHERE id = $1), 
    DEFAULT_UUID()
  );
$$ LANGUAGE SQL;

SELECT KNOWN_ID(NULL, ''), KNOWN_ID(NULL, '');
```

akuzm:
The baserel cache should only be allocated and freed by the top-level query.

Fixes #4823


src/planner/planner.c
```diff
@@ -444,8 +440,12 @@ timescaledb_planner(Query *parse, int cursor_opts, ParamListInfo bound_params)
...
 	PlannedStmt *stmt;
 	ListCell *lc;
-	bool reset_fetcher_type = false;
-	bool reset_baserel_info = false;
+	/*
+	 * Volatile is needed because these are the local variables that are
+	 * modified between setjmp/longjmp calls.
+	 */
+	volatile bool reset_fetcher_type = false;
+	volatile bool reset_baserel_info = false;
```
akuzm, line 448: So that's what the `-Wclobbered` warning was about... I'll fix the other places separately.

fabriziomello (approved):

svenklemm (approved):

PR #4895, opened 2022-10-28
Allow foreign tables in hypertable modify path

gayyappan:
OSM chunks are foreign tables. Modify assert to allow updates/deletes on hypertables with foreign table chunks.

### DIFF
--- a/src/nodes/hypertable_modify.c
+++ b/src/nodes/hypertable_modify.c
@@ -820,8 +820,11 @@ ExecModifyTable(PlanState *pstate)
 			bool isNull;
 
 			relkind = resultRelInfo->ri_RelationDesc->rd_rel->relkind;
-			/* Since this is a hypertable relkind should always be RELKIND_RELATION. */
-			Assert(relkind == RELKIND_RELATION);
+			/* Since this is a hypertable relkind should be RELKIND_RELATION for a local
+			 * chunk or  RELKIND_FOREIGN_TABLE for a chunk that is a foreign table
+			 * (OSM chunks)
+			 */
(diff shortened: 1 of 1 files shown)


src/nodes/hypertable_modify.c
```diff
@@ -820,8 +820,9 @@ ExecModifyTable(PlanState *pstate)
 			bool isNull;
 
 			relkind = resultRelInfo->ri_RelationDesc->rd_rel->relkind;
-			/* Since this is a hypertable relkind should always be RELKIND_RELATION. */
-			Assert(relkind == RELKIND_RELATION);
+			/* Since this is a hypertable relkind should be RELKIND_RELATION. or
+			 * a chunk that is a foreign table (OSM chunks) */
+			Assert(relkind == RELKIND_RELATION || relkind == RELKIND_FOREIGN_TABLE);
```
svenklemm, line 825: OSM chunks are still part of a hypertable so why do we need to allow foreign tables here?

svenklemm, line 825:
```suggestion
			/* Since this is a hypertable relkind should be RELKIND_RELATION for a local chunk  or
			 * RELKIND_FOREIGN_TABLE for a chunk that is a foreign table (OSM chunks) */
			Assert(relkind == RELKIND_RELATION || relkind == RELKIND_FOREIGN_TABLE);
```

svenklemm (approved):

src/nodes/hypertable_modify.c
gayyappan: fixed.

fabriziomello (approved):

PR #4897, opened 2022-10-28
Upload test results to the database

akuzm:
This will help determine the most flaky tests.


akuzm:
> @fabriziomello, @pmwkaa: please review this pull request.
> 
> > Powered by [pull-review](https://github.com/imsky/pull-review)

Oh wow, what is this stuff :open_mouth: Please disregard it.

.github/workflows/linux-build-and-test.yaml
```diff
@@ -59,10 +60,20 @@ jobs:
...
 
+    - name: Checkout TimescaleDB
+      uses: actions/checkout@v3
+
+    # We are going to rebuild Postgres at least once a month, so that it doesn't
+    # suddenly break ages after the original problem.
+    - name: Get date for build caching
+      id: get-date
+      run: |
+        echo "date=$(date +"%m")" >> $GITHUB_OUTPUT
```
svenklemm, line 75: hmm i think weekly or even daily would be better

scripts/upload_ci_stats.sh
```diff
@@ -0,0 +1,178 @@
...
+select create_hypertable('job', 'job_date');
+
+create table if not exists test(
+    job_date timestamptz,
+    test_name text,
+    test_status text,
+    test_duration float
+);
+
+create unique index on test(job_date, test_name);
```
akuzm, line 43: I think it's convenient to have unique names for our tests. We had some duplicates (e.g. an isolation and a regression test with the same name), so I renamed them.

scripts/upload_ci_stats.sh
```diff
@@ -0,0 +1,178 @@
...
+                to_print = lines_buffered > max_context_lines ? min_context_lines : lines_buffered;
+
+                if (lines_buffered > to_print)
+                    print "<" lines_buffered - to_print " lines skipped>";
+
+                for (i = 0; i < to_print; i++) {
+                    print buf[(NR + i - to_print) % max_context_lines]
+                }
+
+                printf("c %04d: %s\n", NR, $0);
```
akuzm, line 132: The `c %04d: ` part is for debugging. I'd like to leave it in for a while, until I'm absolutely sure that this diff snipping code works. It doesn't harm readability too much, and actually helps notice errors, because it has line numbers.

.github/workflows/linux-build-and-test.yaml
akuzm: Changed to daily.

svenklemm:
Hmm not a fan of this being in the matrix builder, can we split this into separate file to keep the matrix builder cleaner

akuzm:
> Hmm not a fan of this being in the matrix builder, can we split this into separate file to keep the matrix builder cleaner

You mean, the flaky check? I don't know, logically it has to be somewhere around the matrix builder because it adds a special configuration and computes the installcheck options. I can split out the calculation of the changed tests into a separate procedure. Or even into a separate file, although to my taste that's an overkill.

Sorry for this PR being confusing, I already do three things here -- fixing flaky check, fixing the LLVM options like you asked me before, and finally adding the upload itself. Maybe it's better to split it up.

akuzm:
> I already do three things here 

Four things, also changing how the Postgres builds are cached.

akuzm:
Factored some stuff out:

* [ ] https://github.com/timescale/timescaledb/pull/4947
* [ ] https://github.com/timescale/timescaledb/pull/4948

svenklemm (approved):

.github/workflows/windows-build-and-test.yaml
```diff
@@ -177,43 +184,63 @@ jobs:
...
-            if re.search('failed', line, re.IGNORECASE):
-              print(line, end='')
-
-        for path in Path('build_wsl').rglob('regression.diffs'):
-          for line in path.open():
-            print(line, end='')
-        
-        with open(os.environ['GITHUB_OUTPUT'], 'a') as output:
-          print('regression_diff=true', file=output)
+        find . -name regression.diffs -exec cat {} + > regression.log
```
konskov, line 199: Why remove the python code here? for convenience or is there another reason?

akuzm: I wanted to make it as similar as possible to the other checks like Linux. This way it's easier to maintain, I can just diff those files and copy the fragments of the code, no need to write same thing anew in another language.

konskov (approved):

PR #4898, opened 2022-10-28
Fix failure resuming a CAgg migration

fabriziomello:
Trying to resume a failed Continuous Aggregate raise an exception that the migration plan already exists, but this is wrong and the expected behaviour should be resume the migration and continue from the last failed step.


svenklemm (commented):
No test for resume?

fabriziomello:
> No test for resume?

There are, search for "resuming" in the test output ... Should I make it more clear?

PR #4899, opened 2022-10-28
Create index fails if hypertable has foreign table chunk

gayyappan:
We cannot create indexes on foreign tables. This PR modifies process_index_chunk to skip OSM chunks

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -2486,7 +2486,11 @@ process_index_chunk(Hypertable *ht, Oid chunk_relid, void *arg)
 	Assert(!hypertable_is_distributed(ht));
 
 	chunk = ts_chunk_get_by_relid(chunk_relid, true);
-
+	if (IS_OSM_CHUNK(chunk)) /*cannot create index on foreign OSM chunk */
+	{
+		ereport(NOTICE, (errmsg("skipping index creation for tiered data")));
+		return;
+	}
 	chunk_rel = table_open(chunk_relid, ShareLock);
 	hypertable_index_rel = index_open(info->obj.objectId, AccessShareLock);
 	indexinfo = BuildIndexInfo(hypertable_index_rel);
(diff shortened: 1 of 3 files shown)


tsl/test/sql/chunk_utils_internal.sql
```diff
@@ -394,6 +394,14 @@ FROM chunk_view
 WHERE hypertable_name = 'hyper_constr'
 ORDER BY chunk_name;
 
+--Utility functions -check index creation on hypertable with OSM chunk
```
fabriziomello, line 397: IMHO will be nice if you can improve a bit the comments here for people understand that we're skipping CREATE INDEX on OSM chunks...

fabriziomello (approved):

src/process_utility.c
```diff
@@ -2430,7 +2430,8 @@ process_index_chunk(Hypertable *ht, Oid chunk_relid, void *arg)
 	Assert(!hypertable_is_distributed(ht));
 
 	chunk = ts_chunk_get_by_relid(chunk_relid, true);
-
+	if (IS_OSM_CHUNK(chunk)) /*cannot create index on foreign OSM chunk */
```
fabriziomello, line 2433: What about add a NOTICE message where about skipping?

src/process_utility.c
```diff
@@ -2494,6 +2495,14 @@ process_index_chunk_multitransaction(int32 hypertable_id, Oid chunk_relid, void
 	}
 #endif
 
+	chunk = ts_chunk_get_by_relid(chunk_relid, true);
+	if (IS_OSM_CHUNK(chunk)) /*cannot create index on foreign OSM chunk */
```
fabriziomello, line 2499: What about add a NOTICE message where about skipping?

tsl/test/sql/chunk_utils_internal.sql
gayyappan: added more comments here.

src/process_utility.c
mfundul: By moving `chunk = ts_chunk_get_by_relid(chunk_relid, true);` above you're breaking the lock semantics described below:
```
...
	 * Hold a lock on the hypertable index, and the chunk to prevent
	 * from being altered. Since we use the same relids across transactions,
...
```

mfundul: I don't really know.

fabriziomello: Seems this is to prevent a concurrent session alter the same object (alter/drop columns, add/drop indexes, etc). Will be better @gayyappan to move your check and skip to after `table_open` and `index_open`

zilder: @gayyappan is there any problems with moving this block back into the lock-protected section?

Patch: https://gist.github.com/zilder/a9b2b61af07361ae6a938b75ae516fb2

gayyappan: yes, it  is ok to move the check after the table_open. This code block will not work as-is. Note that we should clear state correctly before exiting from the function. Probably easiest with an if-else stmt.

src/process_utility.c
gayyappan: done

src/process_utility.c
gayyappan: moved the code around to address the problem described by mfundul.

fabriziomello (approved):
LGTM

zilder (approved):

PR #4902, opened 2022-10-31
Fix error: variable not found in subplan target list on PG15

### LINKED ISSUE #4834: ERROR:  variable not found in subplan target list on PG15
### What type of bug is this?

Incorrect result

Compression

### What happened?

TimescaleDB 2.9.0 when compiled against PG15 causes few regresscheck tests to fail with below error:
ERROR:  variable not found in subplan target list

2.9.0

### PostgreSQL version used

15.0

Mac OS X

Source

On prem/Self-hosted

_No response_

```bash
Run below tests:

rowsecurity-15
compression
compression_bgw
modify_exclusion
```

sb230132:
On PG15 new flag CUSTOMPATH_SUPPORT_PROJECTION is introduced. This flag
tells if a planner node is projection capable or not. CustomScan created
in TimescaleDB by default is not projection capable, this causes CustomScan
node to be wrapped around Result node. Update query on a hypertable has
a logic which is based on assumption that "ModifyTable" plan nodes lefttree 
should be CustomScan node. With PG15 this assumption is broken which causes
"ERROR:  variable not found in subplan target list".

Fixes #4834


mfundul:
Can you point to the PG commit that caused this change?

sb230132:
Please refer below details:

commit 955b3e0f9269639fb916cee3dea37aee50b82df0
Author: Tom Lane <tgl@sss.pgh.pa.us>
Date:   Tue Jul 6 18:10:11 2021 -0400

    Allow CustomScan providers to say whether they support projections.
    
    Previously, all CustomScan providers had to support projections,
    but there may be cases where this is inconvenient.  Add a flag
    bit to say if it's supported.
    
    Important item for the release notes: this is non-backwards-compatible
    since the default is now to assume that CustomScan providers can't
    project, instead of assuming that they can.  It's fail-soft, but could
    result in visible performance penalties due to adding unnecessary
    Result nodes.
    
    Sven Klemm, reviewed by Aleksander Alekseev; some cosmetic fiddling
    by me.
    
    Discussion: https://postgr.es/m/CAMCrgp1kyakOz6c8aKhNDJXjhQ1dEjEnp+6KNT3KxPrjNtsrDg@mail.gmail.com

mfundul:
I meant put this reference in the commit message, you can check out other PG15 commits and do it in a similar manner.

svenklemm (changes requested):
That seems like not the correct fix for this issue and incidental that it fixes the error.

sb230132:
> That seems like not the correct fix for this issue and incidental that it fixes the error.

Here is my analysis of the issue:
While creating "ModifyTable" plan node, the subplan returned is, CustomScan node(which is actually ChunkAppend).
CustomScan node instead of being set as lefttree of "ModifyTable", this node is wrapped around a "Result" node. This is
because CustomScan node is not marked as projection capable. Later when creating root node which is "HypertableModify"
there is code to replace ROWID_VAR in targetlist with NULL constants. Refer below code in hypertable_modify.c

```
	if (mt->operation == CMD_UPDATE || mt->operation == CMD_DELETE)
	{
		cscan->scan.plan.targetlist =
			ts_replace_rowid_vars(root, cscan->scan.plan.targetlist, mt->nominalRelation);

		if (mt->operation == CMD_UPDATE && ts_is_chunk_append_plan(mt->plan.lefttree))
		{
			mt->plan.lefttree->targetlist =
				ts_replace_rowid_vars(root, mt->plan.lefttree->targetlist, mt->nominalRelation);
		}
	}
``` 
Since "ModifyTable" nodes lefttree is not ChunkAppend node, ChunkAppend nodes targetlist ROWID_VAR is not replaced with NULLs, thus causing the error.

Fix is to mark ChunkAppend node as projection capable. If this is not the right fix, then please let me know the right solution.

svenklemm:
We need to be able to support result nodes there cause if we enable projection like you suggested we will just paper over the actual issue and a different plan will expose the same error

sb230132:
> We need to be able to support result nodes there cause if we enable projection like you suggested we will just paper over the actual issue and a different plan will expose the same error

Agreed. Iam not enabling projection on CustomScan node. Iam fixing ts_is_chunk_append_plan() to check for chunkAppend node in case this node is wrapped around Result node in PG15.

svenklemm (approved):

mfundul (approved):

PR #4904, opened 2022-11-01
Remove BitmapScan support in DecompressChunk

svenklemm:
We don't want to support BitmapScans below DecompressChunk as this adds additional complexity to support and there is little benefit in doing so.
This fixes a bug that can happen when we have a parameterized BitmapScan that is parameterized on a compressed column and will lead to an execution failure with an error regarding incorrect attribute types in the expression.


akuzm:
> No tests since this is quite complex to set up.

```
explain (costs off) select * from metrics_ordered_idx where device_id = 1;
 Append
   ->  Custom Scan (DecompressChunk) on _hyper_126_12255_chunk
         ->  Bitmap Heap Scan on compress_hyper_127_12271_chunk
               Recheck Cond: (device_id = 1)
               ->  Bitmap Index Scan on compress_hyper_127_12271_chunk__compressed_hypertable_127_devic
                     Index Cond: (device_id = 1)
```
Is this the kind of the plan we're talking about? I get it by setting `enable_seqscan` and `enable_indexscan` to off. The table is `metrics_ordered_idx` from the `transparent_decompression_ordered_index` test.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -425,6 +425,17 @@ ts_decompress_chunk_generate_paths(PlannerInfo *root, RelOptInfo *chunk_rel, Hyp
 		Path *child_path = lfirst(lc);
 		DecompressChunkPath *path;
 
+		/*
+		 * We skip any BitmapScan paths here as supporting those
+		 * would require fixing up the internal scan. Since we
+		 * currently do not do this BitmapScans would be generated
+		 * when we have a parameterized path on a compressed column
+		 * that would have invalid references due to our
```
akuzm, line 433: Does bitmap scan work for the non-parameterized path? If so, maybe we should move this condition under the `if (child_path->param_info != NULL)`.

svenklemm:
> > > No tests since this is quite complex to set up.
> > 
> > 
> > ```
> > explain (costs off) select * from metrics_ordered_idx where device_id = 1;
> >  Append
> >    ->  Custom Scan (DecompressChunk) on _hyper_126_12255_chunk
> >          ->  Bitmap Heap Scan on compress_hyper_127_12271_chunk
> >                Recheck Cond: (device_id = 1)
> >                ->  Bitmap Index Scan on compress_hyper_127_12271_chunk__compressed_hypertable_127_devic
> >                      Index Cond: (device_id = 1)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is this the kind of the plan we're talking about? I get it by setting `enable_seqscan` and `enable_indexscan` to off. The table is `metrics_ordered_idx` from the `transparent_decompression_ordered_index` test.
> 
> That's not parameterized path.

For a parameterized path you need a more complex plan with eg JOIN. This is just a plan with a qual but not parameterized.

akuzm:
>  For a parameterized path you need a more complex plan with eg JOIN. This is just a plan with a qual but not parameterized.

Yeah, I can get it to parameterize on segmentby, but on compressed column it's complicated indeed (and not sure it makes much sense).

akuzm (approved):

PR #4907, opened 2022-11-01
Remove stale chunk before copying chunk

### LINKED ISSUE #4847: Multinode: Enhance copy_chunk/move_chunk to truncate/remove an existing chunk

API improvement

Access node, Data node, Multi-node

https://github.com/timescale/timescaledb/issues/4846 plans to remove metadata on Access -node (AN) for down datanode (DN) visavis a chunk that gets written to while its down.

When DN comes back, rebalancer service in Timescale Cloud (TSC) will call "copy_chunk" to refresh a stale chunk. We need to modify `copy_chunk` to take a parameter and drop/truncate an existing target chunk on the DN while doing the copy.

### Implementation challenges

_No response_

pmwkaa:
This change adds additional stage to the copy/move chunk operation logic which is used to remove previosly created chunks on data node, known as stale chunks.

Stale chunk is a chunk which exists only on a data node without access node awareness. This situation is possible when access node decided to abandon the chunk to support write HA.

Note: this PR does not add additional argument to the copy/move chunk function. We need to decide, whether we actually need it. Potentially it might be in conflict with the cleanup function, if it was not executed after previously failed copy attempt.

Fix #4847

### DIFF
--- a/tsl/test/expected/dist_move_chunk.out
+++ b/tsl/test/expected/dist_move_chunk.out
@@ -611,6 +611,167 @@ WHERE hypertable_name = 'dist_test';
  _timescaledb_internal._dist_hyper_3_12_chunk | {db_dist_move_chunk_3}
 (4 rows)
 
+DROP TABLE dist_test;
+-- Ensure stale chunks are handled during chunk copy
+CREATE TABLE dist_test(time timestamp NOT NULL, device int, temp float);
+SELECT create_distributed_hypertable('dist_test', 'time', 'device', 3);
+WARNING:  column type "timestamp without time zone" used for "time" does not follow best practices
+ create_distributed_hypertable 
+-------------------------------
+ (4,public,dist_test,t)
+(1 row)
+
+INSERT INTO dist_test SELECT t, (abs(timestamp_hash(t::timestamp)) % 10) + 1, 0.10 FROM generate_series('2018-03-02 1:00'::TIMESTAMPTZ, '2018-03-08 1:00', '1 hour') t;
+SELECT * from show_chunks('dist_test');
+                 show_chunks                  
+----------------------------------------------
+ _timescaledb_internal._dist_hyper_4_13_chunk
+ _timescaledb_internal._dist_hyper_4_14_chunk
+ _timescaledb_internal._dist_hyper_4_15_chunk
+ _timescaledb_internal._dist_hyper_4_16_chunk
+(4 rows)
+
+SELECT * FROM test.remote_exec(NULL, $$ SELECT * from show_chunks('dist_test'); $$);
+NOTICE:  [db_dist_move_chunk_1]:  SELECT * from show_chunks('dist_test')
+NOTICE:  [db_dist_move_chunk_1]:
+show_chunks                                 
+--------------------------------------------
+_timescaledb_internal._dist_hyper_4_13_chunk
+_timescaledb_internal._dist_hyper_4_16_chunk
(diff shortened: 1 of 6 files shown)


tsl/test/sql/dist_move_chunk.sql
```diff
@@ -255,6 +255,33 @@ SELECT chunk_schema || '.' ||  chunk_name, data_nodes
...
+CREATE TABLE dist_test(time timestamp NOT NULL, device int, temp float);
+SELECT create_distributed_hypertable('dist_test', 'time', 'device', 3);
+INSERT INTO dist_test SELECT t, (abs(timestamp_hash(t::timestamp)) % 10) + 1, 0.10 FROM generate_series('2018-03-02 1:00'::TIMESTAMPTZ, '2018-03-08 1:00', '1 hour') t;
+
+SELECT * from show_chunks('dist_test');
+SELECT * FROM test.remote_exec(NULL, $$ SELECT * from show_chunks('dist_test'); $$);
+
+-- create dirty chunk
+\c :DATA_NODE_2 :ROLE_CLUSTER_SUPERUSER;
+CREATE TABLE _timescaledb_internal._dist_hyper_4_13_chunk(time timestamp NOT NULL, device int, temp float);
```
pmwkaa, line 270: Had to create it manually, could not found a way to do this right now, before we have a proper detach/drop chunk logic on the AN for HA.

nikkhils: @pmwkaa I would recommend to adopt the "stale" chunk nomenclature instead of "dirty" chunk overall for the patch. We can also use "orphan" chunks if you prefer. But in my current work I am using "stale" right now.

nikkhils (approved):
LGTM overall. I think we should discuss the name. "Dirty" doesn't make sense. Maybe "Orphan" makes more sense? Before we merge this in we should discuss and firm up the naming and then merge.

tsl/test/expected/dist_move_chunk.out
```diff
@@ -611,6 +611,162 @@ WHERE hypertable_name = 'dist_test';
...
+-------------
+ 
+(1 row)
+
+-- create dirty chunk
+\c :DATA_NODE_2 :ROLE_CLUSTER_SUPERUSER;
+CREATE TABLE _timescaledb_internal._dist_hyper_4_13_chunk(time timestamp NOT NULL, device int, temp float);
+\c :TEST_DBNAME :ROLE_CLUSTER_SUPERUSER;
+-- try to copy chunk copy a chunk with dirty chunk table exists
+CALL timescaledb_experimental.copy_chunk(chunk=>'_timescaledb_internal._dist_hyper_4_13_chunk', source_node=> :'DATA_NODE_1', destination_node => :'DATA_NODE_2');
```
erimatnor, line 670: Should also test the opposite copy direction:

```
CALL timescaledb_experimental.copy_chunk(chunk=>'_timescaledb_internal._dist_hyper_4_13_chunk', source_node=> :'DATA_NODE_2', destination_node => :'DATA_NODE_1');
```

I guess in this case an error should be raised?

erimatnor: Btw, should we introduce an "fail_if_exists" option to the function? I.e., we fail with an error if the chunk exists already unless we set `fail_if_exists=>false`.

pmwkaa:
Closing this PR in favor of https://github.com/timescale/timescaledb/pull/4971

PR #4908, opened 2022-11-01
Honor usage of OidIsValid() macro

fabriziomello:
Postgres source code define the macro `OidIsValid()` to check if the Oid is valid or not (comparing against the `InvalidOid` type). See `src/include/c.h` in Postgres source three.

Changed all direct comparisons against `InvalidOid` for the `OidIsValid` call and add a coccinelle check to make sure the future changes will use it correctly.

### DIFF
--- a/coccinelle/oidisvalid.cocci
+++ b/coccinelle/oidisvalid.cocci
@@ -1,0 +1,30 @@
+//
+// find comparisons against `InvalidOid`
+//
+// Postgres has `OidIsValid()` macro to check if a given Oid is valid or not
+// (see https://github.com/postgres/postgres/blob/master/src/include/c.h).
+//
+// This script look for comparisons to `InvalidOid` and recommend the usage of
+// `OidIsValid()` instead.
+//
+// For example:
+//    `oid != InvalidOid` should be `OidIsValid(oid)`
+//    `oid == InvalidOid` should be `!OidIsValid(oid)`
+//
+@@
+symbol InvalidOid;
+expression oid;
+@@
+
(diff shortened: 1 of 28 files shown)


coccinelle/oidisvalid.cocci
```diff
@@ -0,0 +1,13 @@
+@@
+expression O;
+@@
+
+- (O != InvalidOid)
++ OidIsValid(O)
```
jnidzwetzki, line 6: In the other Coccinelle checks, we also add a comment if the patch can be applied (see [this](https://github.com/timescale/timescaledb/blob/c06b647680dabac74cd027735ad7176f855c24a0/coccinelle/hash_create.cocci#L37) or [this](https://github.com/timescale/timescaledb/blob/c06b647680dabac74cd027735ad7176f855c24a0/coccinelle/hash_create.cocci#L37)). 

When the CI runs and one of the Coccinelle patches can be applied, the differences between the unpatched and the patched file are shown. The added comment makes it clearer for the user what the problem is.

jnidzwetzki (approved):
I have added a minor comment, the rest looks good to me.

svenklemm (approved):

coccinelle/oidisvalid.cocci
svenklemm: I agree please add a header to the file describing the check and a message when code does not conform so it's clear what the problem is

fabriziomello: Added! Thanks!

PR #4911, opened 2022-11-02
Fix sanitizer builds with -Wclobbered

akuzm:
They use GCC 10 which has some other set of false positives.

### DIFF
--- a/src/loader/loader.c
+++ b/src/loader/loader.c
@@ -496,7 +496,8 @@ static void
 check_uuid(const char *label)
 {
 	const MemoryContext oldcontext = CurrentMemoryContext;
-	const char *uuid = strchr(label, SECLABEL_DIST_TAG_SEPARATOR);
+	/* Volatile is to work around the incorrect GCC -Wclobbered diagnostics. */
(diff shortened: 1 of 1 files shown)


mfundul:
Can you post the exact error message please? I am curious.

mfundul:
> ```
> /home/runner/work/timescaledb/timescaledb/src/loader/loader.c: In function ‘check_uuid’:
> /home/runner/work/timescaledb/timescaledb/src/loader/loader.c:500:23: error: variable ‘uuid’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]
>   500 |  const char *uuid = strchr(label, SECLABEL_DIST_TAG_SEPARATOR);
>       |              ^~~~
> ```
> 
> uuid is not used after PG_CATCH, so it's not a problem, but gcc still complains.

So does this false positive get fixed with gcc version > 10?

akuzm:
> So does this false positive get fixed with gcc version > 10?

Yes, for the normal CI builds we use gcc 11 and it doesn't complain.

mfundul (approved):

pmwkaa (approved):

fabriziomello (approved):

PR #4912, opened 2022-11-02
Check if sanitizers still build

akuzm:


akuzm:
Checked what I wanted.

PR #4916, opened 2022-11-03
Disable llvm on macos

akuzm:
I accidentally re-enabled it when adding the flaky check.

### DIFF
--- a/.github/workflows/linux-build-and-test.yaml
+++ b/.github/workflows/linux-build-and-test.yaml
@@ -92,7 +92,12 @@ jobs:
         mkdir -p ~/$PG_SRC_DIR
         tar --extract --file postgresql.tar.bz2 --directory ~/$PG_SRC_DIR --strip-components 1
         cd ~/$PG_SRC_DIR
-        ./configure --prefix=$HOME/$PG_INSTALL_DIR --with-llvm LLVM_CONFIG=${{ matrix.llvm_config }} --with-openssl --without-readline --without-zlib --without-libxml ${{ matrix.pg_extra_args }}
+        if [[ "${{ runner.os }}" == "Linux" ]]; then
+          ./configure --prefix=$HOME/$PG_INSTALL_DIR --with-llvm LLVM_CONFIG=${{ matrix.llvm_config }} --with-openssl --without-readline --without-zlib --without-libxml ${{ matrix.pg_extra_args }}
+        else
+          # the current github macos image has a buggy llvm installation so we build without llvm on mac
+          ./configure --prefix=$HOME/$PG_INSTALL_DIR --with-openssl --without-readline --without-zlib --without-libxml ${{ matrix.pg_extra_args }}
+        fi
         make -j $MAKE_JOBS
         make -j $MAKE_JOBS -C src/test/isolation
         make -j $MAKE_JOBS -C contrib/postgres_fdw


.github/workflows/linux-build-and-test.yaml
```diff
@@ -92,7 +92,12 @@ jobs:
         mkdir -p ~/$PG_SRC_DIR
         tar --extract --file postgresql.tar.bz2 --directory ~/$PG_SRC_DIR --strip-components 1
         cd ~/$PG_SRC_DIR
-        ./configure --prefix=$HOME/$PG_INSTALL_DIR --with-llvm LLVM_CONFIG=${{ matrix.llvm_config }} --with-openssl --without-readline --without-zlib --without-libxml ${{ matrix.pg_extra_args }}
+        if [[ "${{ runner.os }}" == "Linux" ]]; then
+          ./configure --prefix=$HOME/$PG_INSTALL_DIR --with-llvm LLVM_CONFIG=${{ matrix.llvm_config }} --with-openssl --without-readline --without-zlib --without-libxml ${{ matrix.pg_extra_args }}
+        else
```
pmwkaa, line 97: Maybe check for MacOS explicitly?

pmwkaa (approved):

.github/workflows/linux-build-and-test.yaml
akuzm: Yep, I'll fix it here: https://github.com/timescale/timescaledb/pull/4897
For now just returted the old version to get the CI back fast.

.github/workflows/linux-build-and-test.yaml
```diff
@@ -92,7 +92,12 @@ jobs:
...
+          # the current github macos image has a buggy llvm installation so we build without llvm on mac
+          ./configure --prefix=$HOME/$PG_INSTALL_DIR --with-openssl --without-readline --without-zlib --without-libxml ${{ matrix.pg_extra_args }}
+        fi
```
svenklemm, line 100: Cant you put `--without-llvm` in pg_extra_args in matrix builder instead

akuzm: Yep, I'll fix it here: https://github.com/timescale/timescaledb/pull/4897
For now just returted the old version to get the CI back fast.

akuzm: I didn't notice it because of the build caching, going to fix this problem as well.

svenklemm: then we dont need matrix.llvm_config anymore if we move this all into matrix builder into pg_extra_args, keeps the workflow cleaner

akuzm: Yep, that's what I've done there.

svenklemm: this PR has no matrix builder changes

akuzm: There in #4897 I mean.

https://github.com/timescale/timescaledb/pull/4897/files#diff-93f32462b75c120eacd7ebe4d9a44caefe0627c996ca7a39091da0707bb8dd40R62

Let's merge this PR first, which is a revert of the earlier incorrect changes. Our CI is actually broken now, Bharaty reported this failure.

fabriziomello (approved):

PR #4925, opened 2022-11-04
Add new chunk state CHUNK_STATUS_COMPRESSED_PARTIAL

svenklemm:
A chunk is in this state when it is compressed but also has uncompressed data in the uncompressed chunk. Individual tuples can only ever exist in either area. This is preparation patch to add support for uncompressed staging area for DML operations.

### DIFF
--- a/sql/maintenance_utils.sql
+++ b/sql/maintenance_utils.sql
@@ -68,18 +68,24 @@ BEGIN
 
     -- Chunk names are in the internal catalog, but we only care about
     -- the chunk name here.
+    -- status bits:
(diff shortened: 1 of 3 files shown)


fabriziomello (approved):
I'm assuming you will add tests in following PR, right?

jnidzwetzki (approved):
I have the same comment about the tests as @fabriziomello, the rest looks good to me.

svenklemm:
This will be heavily used by followup PRs (#4926,...) that improve DML support for compressed chunks.

PR #4928, opened 2022-11-07
Extra Result node on top of CustomScan on PG15

### LINKED ISSUE #4833: Extra Result node on top of CustomScan on PG15
### What type of bug is this?

Incorrect result

Continuous aggregate

### What happened?

TimescaleDB when compiled against PG15 causes many cagg tests to fail. 
Initial failure is related to permissions which reported error like "permission denied for schema public"
Once permission issues is fixed we get incorrect results as shown below.

```
+++ /Volumes/Work/pg15-regresscheck-testing/build/tsl/test/results/exp_cagg_monthly.out 2022-10-17 09:35:26.000000000 +0530
@@ -1044,6 +1044,7 @@
 (3 rows)

 GRANT USAGE ON FOREIGN SERVER :DATA_NODE_1, :DATA_NODE_2, :DATA_NODE_3 TO PUBLIC;
+GRANT CREATE ON SCHEMA public TO :ROLE_DEFAULT_PERM_USER;
 CREATE TABLE conditions_dist(
   day DATE NOT NULL,
@@ -1065,7 +1066,7 @@
    MAX(temperature)
 FROM conditions_dist
 GROUP BY bucket;
-NOTICE:  refreshing continuous aggregate "conditions_dist_1m"
+NOTICE:  continuous aggregate "conditions_dist_1m" is already up-to-date
 SELECT mat_hypertable_id AS cagg_id
 WHERE user_view_name = 'conditions_dist_1m'
@@ -1091,11 +1092,9 @@
 (1 row)

 SELECT * FROM conditions_dist_1m ORDER BY bucket;
-   bucket   | min | max
-------------+-----+-----
- 01-01-2010 | 101 | 131
- 02-01-2010 | 201 | 228
-(2 rows)
+ bucket | min | max
+--------+-----+-----
+(0 rows)
```

2.9.0

### PostgreSQL version used

15.0

Mac OS X 

Source

On prem/Self-hosted

_No response_

```bash
Please run below regresscheck tests after fixing permission issues.

cagg_ddl_dist_ht
telemetry_stats
cagg_migrate_integer_dist_ht
cagg_migrate_timestamp_dist_ht
cagg_bgw_dist_ht
cagg_ddl_dist_ht
chunk_api
compression_bgw
data_node
dist_api_calls
dist_commands
dist_compression
dist_copy_format_long
dist_copy_long
dist_ddl
dist_move_chunk
dist_partial_agg
dist_policy
dist_query
exp_cagg_monthly
exp_cagg_origin
exp_cagg_timezone
cagg_concurrent_refresh_dist_ht
cagg_drop_chunks
cagg_multi_dist_ht
continuous_aggs
continuous_aggs_deprecated
```

sb230132:
On PG15 CustomScan by default is not projection capable, thus wraps this node in Result node. This change in PG15 causes tests result files which have EXPLAIN output to fail. This patch fixes the plan outputs.

Fixes #4833


mfundul (approved):

akuzm (approved):
Sven explained the difference with earlier PG versions:

> On PG <= 14 postgres will modify targetlists of any custom nodes which requires us to fix up targetlists after planning to not end up with broken plans. On PG 15 when the custom node is not marked as projection capable postgres will not mess with the targetlist of the custom node and instead inject a result node if any projection needs to happen. This means we no longer have to do the post-planning fixup. Those targetlist modifications of postgres were source of lots of subtle bugs.

So we want to actually disable these TL modifications to avoid bug-prone code, at least in the future.

PR #4938, opened 2022-11-08
Add nothrow codepath to health check

65278:
Add handling for offline nodes in health check
    
_timescaledb_internal.health() must not throw errors when a remote
node is offline. Since the remote connection cache throws, a new
function ts_dist_cmd_invoke_on_data_nodes_no_throw is introduced
to the dist command api.
    
ts_dist_cmd_invoke_on_data_nodes_no_throw establishes a connection
for each dist node and spawns an async request. If a connection can
not be established, it creates a PGresult with PGRES_FATAL_ERROR for
this connection.
    
This function has a couple of trade offs. Since it reuses remote
connection open functions, it needs to make copies of all PGresults.
This is because PGresults are owned by the connection and we register
an event handler which frees all PGresults on connection close. It
also can't create an AsyncResponseResult, since that API is not
accessible in this scope. Therefore struct DistCmdResponse was
changed along with the functions that handle DistCmdResponses.

### DIFF
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -251,6 +251,17 @@ create_foreign_server(const char *const node_name, const char *const host, int32
 }
 
 TSConnection *
+data_node_get_connection_nothrow(const char *const data_node, char **errmsg)
+{
+	const ForeignServer *server;
+
+	Assert(data_node != NULL);
+	server = data_node_get_foreign_server(data_node, ACL_NO_CHECK, false, false);
+
+	return remote_connection_open_nothrow(server->serverid, GetUserId(), errmsg);
+}
+
+TSConnection *
 data_node_get_connection(const char *const data_node, RemoteTxnPrepStmtOption const ps_opt,
 						 bool transactional)
 {
--- a/tsl/src/remote/async.c
+++ b/tsl/src/remote/async.c
@@ -79,6 +79,7 @@ typedef struct AsyncResponseResult
 typedef struct AsyncResponseCommunicationError
 {
 	AsyncResponse base;
+	TSConnectionError *errmsg;
 	AsyncRequest *request;
 } AsyncResponseCommunicationError;
 
@@ -86,6 +87,7 @@ typedef struct AsyncResponseError
 {
 	AsyncResponse base;
 	const char *errmsg;
+	AsyncRequest *request;
 } AsyncResponseError;
 
 typedef struct AsyncRequestSet
@@ -300,9 +302,12 @@ static AsyncResponseCommunicationError *
 async_response_communication_error_create(AsyncRequest *req)
 {
 	AsyncResponseCommunicationError *ares = palloc0(sizeof(AsyncResponseCommunicationError));
+	TSConnectionError *errmsg = palloc(sizeof(TSConnectionError));
+	remote_connection_get_error(req->conn, errmsg);
 
 	*ares = (AsyncResponseCommunicationError){
 		.base = { .type = RESPONSE_COMMUNICATION_ERROR },
+		.errmsg = errmsg,
 		.request = req,
 	};
 
@@ -322,14 +327,13 @@ async_response_timeout_create()
 }
 
 static AsyncResponse *
-async_response_error_create(const char *errmsg)
+async_response_error_create(const char *errmsg, AsyncRequest *req)
 {
 	AsyncResponseError *ares = palloc0(sizeof(AsyncResponseError));
 
-	*ares = (AsyncResponseError){
-		.base = { .type = RESPONSE_ERROR },
-		.errmsg = pstrdup(errmsg),
-	};
+	*ares = (AsyncResponseError){ .base = { .type = RESPONSE_ERROR },
+								  .errmsg = pstrdup(errmsg),
+								  .request = req };
 
 	return &ares->base;
 }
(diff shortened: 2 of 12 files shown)


tsl/src/data_node.c
```diff
@@ -243,6 +243,17 @@ create_foreign_server(const char *const node_name, const char *const host, int32
...
+TSConnection *
+data_node_get_connection_nothrow(const char *const data_node, char **errmsg)
+{
+	const ForeignServer *server;
+
+	Assert(data_node != NULL);
+	server = data_node_get_foreign_server(data_node, ACL_NO_CHECK, false, false);
+
+	return remote_connection_open_nothrow(server->serverid, GetUserId(), errmsg);
+}
```
nikkhils, line 255: @65278 why aren't we using the existing `remote_connection_ping` API here?

nikkhils: Hmm, self answer, because we want to send the healthcheck query to run on the DN itself.

tsl/src/remote/dist_commands.c
```diff
@@ -76,6 +78,117 @@ ts_dist_cmd_collect_responses(List *requests)
...
+	while ((ar = async_request_set_wait_ok_result(rs)))
+	{
+		DistCmdResponse *response = &results->responses[i];
+
+		/* Closing remote connections will trigger event processing
+		 * which in turn clears PGresults from Async Responses.
+		 * Make a deep copy before closing.
+		 * Todo: a deep copy will always be PGRES_TUPLES_OK */
+		response->pg_result =
+			PQcopyResult(async_response_result_get_pg_result(ar), PG_COPYRES_TUPLES);
```
erimatnor, line 164: Hmm, I am hesitant about this approach. Ideally, we should life-cycle manage also Result copies, although we aren't at this time because we haven't used any Result copies so far.

In other words, we should probably store the Result copy on the connection list when `PGEVT_RESULTCOPY` occurs, although we currently don't handle that event because of the above reason. However, if we do in the future, then this will break.

The reason we life-cycle manage the Results is because they are malloc'd  instead of allocated on PG memory contexts. So, there's a huge risk of memory leaks otherwise.

erimatnor: Maybe one way to fix this is to change the API for this function to require the caller to pass in the connection. Then the connection can be managed outside the scope of the function and one can ensure it is not closed until responses are processed.

65278: I thought about that too, will do. There's no harm to keeping the connection in the DistCmdResponse.

65278:
> Overall, I would have preferred a simpler `SELECT 1` kinda approach via existing `nothrow` functions. But I see that we want to invoke the `healthcheck` function on the DNs as well so we need this generic infra that this patch adds to send any command to the DNs in a `nothrow` fashion.
> 
> So from that perspective, this patch seems fine to me. Please add some more tests and make it work with search path

I've done the tests a while ago, I had some false positives, but I'll be able to make it pass.

Regarding `search_path`. I've added this, but it adds a really high amount of code complexity and doesn't seem worth it to me. The problem is that there needs to be logic to handle the `set search_path` result now, which is different from the logic used to handle the actual query result. On top of that, it needs extra logic to handle clearing the connection. This code is in need of refactoring already.

65278:
I've added an extra codepath to `async.c`: `async_request_set_wait_any_result_nothrow`. This fixes a possible exception I hadn't noticed before. I'm fairly positive all the edge cases left absolutely require throwing.

65278:
I've done a run of code clean-ups, streamlined the logic, made it more readable. I'm satisfied with the search path branch now.

tsl/src/remote/async.c
```diff
@@ -357,6 +361,61 @@ async_response_close(AsyncResponse *res)
...
+			break;
+		case RESPONSE_COMMUNICATION_ERROR:
+		{
+			AsyncResponseCommunicationError *arc = (AsyncResponseCommunicationError *) res;
+			TSConnectionError *err = arc->errmsg;
+			if (node_name)
+				*node_name = err->nodename;
+
+			int errcode = err->remote.errcode != 0 ? err->remote.errcode :
+						  err->remote.errcode != 0 ? ((err)->remote.errcode) :
```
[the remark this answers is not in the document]
65278: You're absolutely right. I had a way longer expression which I shortened ... with this result.

tsl/src/remote/dist_commands.c
```diff
@@ -76,6 +86,287 @@ ts_dist_cmd_collect_responses(List *requests)
...
+				 errmsg("no data nodes to execute command on"),
+				 errhint("Add data nodes before executing a distributed command.")));
+
+	data_nodes = ts_dist_cmd_sanitize_data_node_list(data_nodes);
+
+	/* Since libpq uses malloc, register a callback that frees results
+	 * on Memory Context Reset/Delete */
+	callback_data = palloc0(sizeof(List *));
+	mcb = palloc0(sizeof(MemoryContextCallback));
+	mcb->func = &ts_dist_cmd_mctx_reset_callback;
```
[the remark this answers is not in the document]
65278: That's not a bug, but it's not idiosyncratic C. I actually had to look this up in the standard as I was as bewildered as you are. The compiler has no choice but to turn the rvalue into a pointer, therefore the ampersand can be omitted. Writing it down is explicit and doesn't turn the lvalue into a pointer pointer.
I've written a lot of perl lately, and in perl you'd call the symbol, so you reference it instead.
Here's the assembly btw:
```
│    0x7f44ffa3f3db <ts_dist_cmd_invoke_on_data_nodes_no_throw+95>   call   0x7f44ff9ce130 <palloc0@plt>                                                                                     │
│    0x7f44ffa3f3e0 <ts_dist_cmd_invoke_on_data_nodes_no_throw+100>  mov    rsi,rax                                                                                                          │
│    0x7f44ffa3f3e3 <ts_dist_cmd_invoke_on_data_nodes_no_throw+103>  lea    rax,[rip+0xfffffffffffffc45]        # 0x7f44ffa3f02f <ts_dist_cmd_mctx_reset_callback>                           │
│    0x7f44ffa3f3ea <ts_dist_cmd_invoke_on_data_nodes_no_throw+110>  mov    QWORD PTR [rsi],rax                                                                                              │
│    0x7f44ffa3f3ed <ts_dist_cmd_invoke_on_data_nodes_no_throw+113>  mov    QWORD PTR [rsi+0x8],rbx                                                                                          │
│    0x7f44ffa3f3f1 <ts_dist_cmd_invoke_on_data_nodes_no_throw+117>  mov    QWORD PTR [rsi+0x10],0x0                                                                                         │
│  > 0x7f44ffa3f3f9 <ts_dist_cmd_invoke_on_data_nodes_no_throw+125>  mov    rax,QWORD PTR [rip+0x47990]
```

tsl/src/remote/dist_commands.c
```diff
@@ -76,6 +86,287 @@ ts_dist_cmd_collect_responses(List *requests)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("no data nodes to execute command on"),
+				 errhint("Add data nodes before executing a distributed command.")));
+
+	data_nodes = ts_dist_cmd_sanitize_data_node_list(data_nodes);
+
+	/* Since libpq uses malloc, register a callback that frees results
+	 * on Memory Context Reset/Delete */
+	callback_data = palloc0(sizeof(List *));
```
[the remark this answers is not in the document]
65278: That's actually not the problem here. It should be `sizeof(List **)`. A list is allowed to change its initial pointer when using `lappend()`. Since this is a list which needs to end up in a callback, I need to store the memory address to the list somewhere.
Still, good catch. The intent is more readable now.

PR #4947, opened 2022-11-09
Change the flaky check to use output files

akuzm:
This is simpler, accounts for both sql and isolation tests, changes in included files, and versioned tests.


akuzm:
Tested manually on a different PR with lots of test changes.

konskov (approved):

fabriziomello (approved):

PR #4954, opened 2022-11-10
Use async API to drop database from delete_data_node

### LINKED ISSUE #4838: Regresscheck tests hanging on PG15
### What type of bug is this?

Unexpected error

Data node

### What happened?

TimescaleDB when compiled against PG15 causes many few tests to hang forever.

2.9.0

### PostgreSQL version used

15.0

Mac OS X

Source

On prem/Self-hosted

_No response_

```bash
make installcheck TESTS="data_node_bootstrap dist_hypertable-15 bgw_custom"
```

lkshminarayanan:
PG15 introduced a ProcSignalBarrier mechanism in drop database implementation to force all backends to close the file handles for dropped tables. The backend that is executing the drop database command will emit a new process signal barrier and wait for other backends to accept it. But the backend which is executing the delete_data_node function will not be able to process the above mentioned signal as it will be stuck waiting for the drop database query to return. Thus the two backends end up waiting for each other causing a deadlock.

Fixed it by using the async API to execute the drop database command from delete_data_node instead of the blocking remote_connection_cmdf_ok call.

Fixes #4838


lkshminarayanan:
Looks like some other tests are still stuck. I'm having a look at it.

lkshminarayanan:
The hanging tests are now fixed. There are still a couple of failures in the PG15 debug run in CI but they are unrelated to this PR and I'm already looking at them.

lkshminarayanan:
Hi @erimatnor and @sb230132, the PR is ready for review.

erimatnor (approved):
This is OK to get this test to work.

We need a bigger overhaul of our use of remote libpq API calls to ensure they are interruptable. This is handled separately though.

sb230132 (approved):

Issue #4944, opened 2022-11-09
High memory usage for a distributed SELECT


akuzm:
```
create table public.dhtab (
    tms timestamp not null,
    dim bigint not null
);

select create_distributed_hypertable('dhtab', 'tms');

insert into dhtab(tms, dim)
select '2020-01-01'::timestamp + i * '1 week'::interval, 1                                                          
from generate_series(1, 10000) t(i); 

explain (analyze) select * from dhtab where dim = any(array(select * from generate_series(1, 10000)));
```

This table has 10k chunks which is bad, but why would this query need 10GB of memory to complete? Looks suspicious.

PR #4960, opened 2022-11-10
Reduce planning memory usage with lots of chunks

akuzm:
Selectivity functions, like scalararrayinsel, allocate _a lot_ in the top planner context. Use a temporary memory context to call them.

Fixes https://github.com/timescale/timescaledb/issues/4944

### DIFF
--- a/src/planner/planner.c
+++ b/src/planner/planner.c
@@ -11,6 +11,7 @@
 #include <executor/nodeAgg.h>
 #include <miscadmin.h>
 #include <nodes/makefuncs.h>
+#include <nodes/pg_list.h>
 #include <nodes/plannodes.h>
 #include <optimizer/appendinfo.h>
 #include <optimizer/clauses.h>
--- a/src/planner/planner.c
+++ b/src/planner/planner.c
@@ -150,6 +150,8 @@ DataFetcherType ts_data_node_fetcher_scan_type = AutoFetcherType;
  */
 static struct BaserelInfo_hash *ts_baserel_info = NULL;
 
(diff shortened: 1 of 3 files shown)


src/planner/planner.c
```diff
@@ -553,11 +570,21 @@ timescaledb_planner(Query *parse, int cursor_opts, ParamListInfo bound_params)
...
 													 /* private_data = */ NULL);
 			}
+
+			/*
+			 * Create a temporary memory context that can optionally be used by
+			 * the planner at this level of timescaledb_planner(). The inner
+			 * recursive invocations of timescaledb_planner() will also create
+			 * their own temporary contexts.
+			 */
+			MemoryContext mcxt = AllocSetContextCreate(CurrentMemoryContext,
```
akuzm, line 584: I'm kinda reluctant to add another weird cache based on static variables, and an entire stack this time... On the other hand, the general Postgres planning approach of "leaking memory like there's no tomorrow" ((c) the mailing lists) becomes problematic when you have a lot of partitions.

svenklemm: I 'm not a fan of adding nested memory contexts like this.

akuzm: Yeah, I don't like it either, feels fragile. We can just drop this as wontfix. Or maybe you have better ideas?

akuzm: The planning memory usage proportional to the number of chunks times the number of array elements also doesn't feel right, but not sure how to fix it in extension.

akuzm:
Closing for now because this is an upstream problem, and the fix is potentially more problematic.

PR #4964, opened 2022-11-11
Fix dist_query dist_partial_agg on PG 15

sb230132:
Fix dist_query dist_partial_agg on PG 15


lkshminarayanan (approved):
Couple of questions but LGTM. 
1. Can we remove the generic `dist_partial_agg.out` and `dist_query.out` now that we have version specific output files?
2. A new `result` fields seems to have been added in 15.0 query plan output. Should we refer that change in the commit message?

sb230132:
> Couple of questions but LGTM.
> 
> 1. Can we remove the generic `dist_partial_agg.out` and `dist_query.out` now that we have version specific output files?
> 2. A new `result` fields seems to have been added in 15.0 query plan output. Should we refer that change in the commit message?

Yes, stale .out files have been removed.
New "Result" node which is specific to PG15 is tracked as part of #4928

akuzm (approved):

PR #4968, opened 2022-11-11
Check for interrupts in gapfill main loop

### LINKED ISSUE #4963: [Bug]: Unable to cancel or kill running query
### What type of bug is this?

Locking issue, Performance issue, Other

Continuous aggregate, Query executor

### What happened?

We have this query which makes use of caggs:
```
SELECT intvl AS time
, sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>"", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>", sum("<MASKED>") AS "<MASKED>"
	FROM (
	  SELECT time_bucket_gapfill('18000s', time) AS intvl
	      , locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 0.000001", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 0.00001", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 0.0001", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 0.001", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 0.01", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 0.1", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 1.0", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 10.0", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 100.0", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 1000.0", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 10000.0", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency 100000.0", locf(max("<MASKED>"), treat_null_as_missing=>true) as "Latency +Inf", <MASKED>, "<MASKED>"
	  FROM "<MASKED>-60min"

	  WHERE time >= '2022-10-26T18:00:00.000Z' AND time < '2022-11-10T19:00:00.000Z'
	  AND "<MASKED>" LIKE '<MASKED>'

	  GROUP BY <MASKED>, "<MASKED>"
	         , intvl
	  WINDOW w AS(
	      PARTITION BY <MASKED>, "<MASKED>"
	      ORDER BY time_bucket_gapfill('18000s', time)
	      ROWS BETWEEN 1 PRECEDING AND CURRENT ROW
	  )
	  ORDER BY intvl
	) AS subquery

	GROUP BY intvl
	ORDER BY intvl
	OFFSET 1;
```

Disregard the complexity and performance characteristics of this for now, it's not good and we know that :) The problem that we're seeing is that we are unable to cancel or kill this query when we need to via `pg_cancel_backend` or `pg_terminate_backend`. These calls return immediately and successfully but the query remains running for a long time. The process handling the query can be seen via `top` as well, and while I know it's not the right thing to do we tried a `kill <PID>` on the postgres process handling the query which also had no effect.

I am not well-versed in the internals or postgres or timescale, but I had found some information online a while back that discussed similar problems that were due to the query using too many locks. Looking at the lock count for this query via it's PID shows that it uses a few hundred locks, but even when we tuned the database to increase `max_locks_per_transaction` nothing changed. 

The only other relevant information I could find was around the `CHECK_FOR_INTERRUPTS()` [macro](https://docs.huihoo.com/doxygen/postgresql/miscadmin_8h_source.html). As far as I know there should be no reason for the database to not cancel/kill a query unless there's a bug in the database itself (going off of postgres dev discussions here). I am wondering if timescale is not making use of this macro in all the spots that it should.

Maybe my analysis above is wrong, but the behavior we see with this query is consistent.

2.8.1

### PostgreSQL version used

14

Base image of official timescale container

Docker

Amazon Web Services (AWS)

_No response_

With a matching schema and cagg setup that the query would work with run the query for a 15 day time window once that time window is filled with data. The query is of course extremely slow and takes a huge amount of memory to run but the real problem is not being able to cancel/kill it.

svenklemm:
Add CHECK_FOR_INTERRUPTS() macro to gapfill main loop.

Fixes #4963

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,11 +29,14 @@ argument or resolve the type ambiguity by casting to the intended type.
 * #4840 Fix performance regressions in the copy code
 * #4823 Fix a crash that could occur when using nested user-defined functions with hypertables
 * #4898 Fix cagg migration failure when trying to resume
+* #4910 Fix a typo in process_compressed_data_out
 * #4955 Fix cagg migration for hypertables using timestamp without timezone
+* #4968 Check for interrupts in gapfill main loop
 
 **Thanks**
 * @jflambert for reporting a crash with nested user-defined functions.
 * @jvanns for reporting hypertable FK reference to vanilla PostgreSQL partitioned table doesn't seem to work
+* @kou for fixing a typo in process_compressed_data_out
 * @xvaara for helping reproduce a bug with bitmap scans in transparent decompression
 
 ## 2.8.1 (2022-10-06)
--- a/tsl/src/nodes/gapfill/gapfill_exec.c
+++ b/tsl/src/nodes/gapfill/gapfill_exec.c
@@ -11,6 +11,7 @@
 #include <catalog/pg_cast.h>
 #include <catalog/pg_collation.h>
 #include <catalog/pg_type.h>
+#include <miscadmin.h>
 #include <nodes/extensible.h>
 #include <nodes/makefuncs.h>
 #include <nodes/nodeFuncs.h>
@@ -810,6 +811,8 @@ gapfill_exec(CustomScanState *node)
 
 	while (true)
 	{
+		CHECK_FOR_INTERRUPTS();
+
 		/* fetch next tuple from subplan */
 		if (FETCHED_NONE == state->state)
 		{


fabriziomello (approved):

jnidzwetzki (approved):
Could you add the fix to the changelog? Rest LGTM.

aaronjwood:
Is there a rough ETA for the next release that would contain this patch?

PR #4971, opened 2022-11-14
Introduce drop_stale_chunks() function

### LINKED ISSUE #4848: Multinode HA: status of stale chunks on a datanode
### What problem does the new feature solve?

https://github.com/timescale/timescaledb/issues/4846 will remove metadata an Access node for a datanode (DN) that is down visavis a chunk on it that gets written to.

When such a DN comes back, it will have a list of chunks on it which have stale data, the DN needs to report back this list to the AN or additionally might also need to carry out any cleanup/maintenance locally.

### What does the feature do?

The DN could report back the dimensions of chunks and then based on the discrepancies in the dimensions we can determine which chunk is stale on it. 

### Implementation challenges

_No response_

pmwkaa:
This function drops chunks on a specified data node if those chunks are not known by the access node.

Call `drop_stale_chunks()` automatically when data node becomes available again.

Fix #4848

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -3261,6 +3261,28 @@ ts_chunk_get_chunk_ids_by_hypertable_id(int32 hypertable_id)
 	return chunkids;
 }
 
+List *
+ts_chunk_get_all_chunk_ids(LOCKMODE lockmode)
+{
+	List *chunkids = NIL;
+	ScanIterator iterator = ts_scan_iterator_create(CHUNK, lockmode, CurrentMemoryContext);
+	ts_scan_iterator_set_index(&iterator, CHUNK, CHUNK_ID_INDEX);
+	ts_scan_iterator_scan_key_init(&iterator,
+								   Anum_chunk_idx_id,
+								   BTEqualStrategyNumber,
+								   F_INT4GE,
+								   Int32GetDatum(0));
+	ts_scanner_foreach(&iterator)
+	{
+		bool isnull;
+		Datum id = slot_getattr(ts_scan_iterator_slot(&iterator), Anum_chunk_id, &isnull);
+		if (!isnull)
+			chunkids = lappend_int(chunkids, DatumGetInt32(id));
+	}
+
+	return chunkids;
+}
+
 static ChunkResult
 chunk_recreate_constraint(ChunkScanCtx *ctx, ChunkStub *stub)
 {
--- a/tsl/src/chunk.c
+++ b/tsl/src/chunk.c
@@ -25,6 +25,7 @@
 #include <utils/snapmgr.h>
 #include <executor/executor.h>
 #include <parser/parse_func.h>
+#include <storage/lmgr.h>
 #include <funcapi.h>
 #include <miscadmin.h>
 #include <fmgr.h>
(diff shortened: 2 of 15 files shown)


tsl/src/chunk.c
```diff
@@ -507,3 +508,187 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+														NameStr(ht->fd.table_name))));
+
+		/* serialize chunk creation around a lock on the "main table" to avoid
+		 * multiple processes trying to create new chunks */
+		LockRelationOid(ht->main_table_relid, ShareUpdateExclusiveLock);
+		ts_hypertable_has_chunks(table_relid, AccessShareLock);
+
+		/* collect a list of chunks which should exists on the specified data node
+		 * according to the access node */
+		chunk_id_list = ts_chunk_get_chunk_ids_by_hypertable_id(ht->fd.id);
```
erimatnor, line 579: Instead of doing the all of these scans, can't we just scan _only_ `chunk_data_node` with a node_name index scan? 

If you want the chunks for just a specific hypertable (although I don't know why we do that), you can use ` `ts_chunk_get_relid` to map the chunk_id to a relid.

tsl/src/chunk.c
```diff
@@ -507,3 +508,187 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+		appendStringInfo(cmd, "array[");
+		foreach (lc, chunk_id_list)
+		{
+			int32_t chunk_id = lfirst_int(lc);
+			int32_t node_chunk_id = -1;
+			Chunk *chunk = ts_chunk_get_by_id(chunk_id, true);
+
+			if (ts_chunk_has_data_node(chunk, node_name))
+			{
+				/* we need to use node_chunk_id instead of chunk_id */
```
erimatnor, line 590: This information is already in `chunk->data_nodes` which is a list of `ChunkDataNode` structs. So this inner scan seems unnecessary.

tsl/src/chunk.c
```diff
@@ -507,3 +508,187 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+			while (array_iterate(it, &chunk_id_datum, &isnull))
+			{
+				if (isnull)
+					ereport(ERROR,
+							(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+							 errmsg("chunks array arguments cannot be NULL")));
+
+				if (chunk_id == DatumGetInt32(chunk_id_datum))
+				{
+					match = true;
```
erimatnor, line 672: I suggesty putting converting the array into a hashtable before the loop so that you can do O(1) lookups to filter chunks.

erimatnor (commented):
It is not clear to me why we look for stale chunks only on a specific hypertable. Aren't we interested in all stale chunks irrespective of hypertables?

By doing this on a per-hypertable basis, it seems like we are complicating the code by having to filter chunks based on hypertable. Ideally, we should only need to scan `chunk_data_node` since it has all the mappings.

I also wish the function gave a bit more info/feedback, like a list of chunks dropped. Or, maybe it makes sense to split this into two functions, one that gets the list of stale chunks and then a separate function to drop them. I realize this might be less efficient in terms of back-and-forth, but I am not sure that matters too much given that this is not performance sensitive and it might be preferable to have more insights, making the functionality more composable and easier to debug.

pmwkaa:
@erimatnor I've made some changes, please take a look again

tsl/src/chunk.c
```diff
@@ -507,3 +508,188 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("node_name argument cannot be NULL")));
+		if (chunks_array != NULL)
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("chunks argument cannot be used on the access node")));
+
+		/* get an exclusive lock on the chunks catalog table to prevent new chunk
+		 * creation during this operation */
+		LockRelationOid(ts_catalog_get()->tables[CHUNK].id, AccessExclusiveLock);
```
pmwkaa, line 625: Probably we need to include CHUNK_DATA_NODE here as well

pmwkaa: Also, do we need isolation tests?

tsl/src/chunk.c
```diff
@@ -507,3 +508,188 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+		id_list = lappend_int(id_list, DatumGetInt32(id_datum));
+	}
+	array_free_iterator(it);
+
+	(void) list_sort_compat(id_list, list_int_cmp_compat);
+	return id_list;
+}
+
+static List *
+chunk_id_list_exclusive_right_merge_join(const List *an_list, const List *dn_list)
```
nikkhils, line 537: Hmm, not sure the name with `merge_join` makes sense.

tsl/src/chunk.c
```diff
@@ -507,3 +508,188 @@ chunk_drop_replica(PG_FUNCTION_ARGS)
...
+		/* scan for chunks that reference the given data node */
+		ScanIterator it = ts_chunk_data_nodes_scan_iterator_create(CurrentMemoryContext);
+		ts_chunk_data_nodes_scan_iterator_set_node_name(&it, node_name);
+		ts_scanner_foreach(&it)
+		{
+			TupleTableSlot *slot = ts_scan_iterator_slot(&it);
+			bool PG_USED_FOR_ASSERTS_ONLY isnull = false;
+			int32_t node_chunk_id;
+
+			node_chunk_id =
```
nikkhils, line 639: @pmwkaa I don't think that the chunk_id is the SAME on the AN and the DN. Looks like we might have to send the chunk_names instead of ids..

pmwkaa: We are iterating over `chunk_data_node` table in order to get chunk id on the data node. This is a mapping table for chunks we use on the access node: https://github.com/timescale/timescaledb/blob/main/src/ts_catalog/catalog.h#L591. S

tsl/src/chunk.c
pmwkaa: We can rename it if people like, but it is essentially what the function do

tsl/src/chunk.c
nikkhils: Ah, ok. Cool

pmwkaa:
Added support for `alter_data_node()`. Now it calls `drop_stale_chunks()` automatically when data node becomes available again.

tsl/src/data_node.c
```diff
@@ -1617,6 +1617,11 @@ data_node_alter(PG_FUNCTION_ARGS)
 		PG_RETURN_DATUM(
 			HeapTupleGetDatum(create_alter_data_node_tuple(tupdesc, node_name, server->options)));
 
+	/* drop stale chunks on the unavailable data node, if we are going to
```
nikkhils, line 1620: The location of this code could be after we run `AlterForeignServer`

src/chunk.c
```diff
@@ -3261,6 +3261,28 @@ ts_chunk_get_chunk_ids_by_hypertable_id(int32 hypertable_id)
...
+{
+	List *chunkids = NIL;
+	ScanIterator iterator = ts_scan_iterator_create(CHUNK, lockmode, CurrentMemoryContext);
+	ts_scan_iterator_set_index(&iterator, CHUNK, CHUNK_ID_INDEX);
+	ts_scan_iterator_scan_key_init(&iterator,
+								   Anum_chunk_idx_id,
+								   BTEqualStrategyNumber,
+								   F_INT4GE,
+								   Int32GetDatum(0));
+	ts_scanner_foreach(&iterator)
```
erimatnor, line 3275: Would be nice if the join could be done in this loop using a per-tuple memory context to avoid memory O(N) memory usage.

tsl/src/data_node.c
pmwkaa: So that means that we run drop_stale_chunks when data node is marked as available, I am not sure this is what we intended. It should conflict with locking, etc, right?

pmwkaa: I guess your point is to move the check before the call to CommandCounterIncrement()  but after the AlterForeignServer() call

tsl/src/chunk.c
```diff
@@ -561,3 +562,206 @@ chunk_unfreeze_chunk(PG_FUNCTION_ARGS)
...
+			if (compare == 0)
+			{
+				/* l = r */
+				l = lnext_compat(an_list, l);
+				r = lnext_compat(dn_list, r);
+			}
+			else if (compare < 0)
+			{
+				/* l < r */
+				/* chunk exists only on the access node */
```
erimatnor, line 614: Should this case every happen? Seems odd that it would. Should we report error here (corrupted metadata)?

tsl/test/expected/dist_ddl.out
```diff
@@ -2805,6 +2806,584 @@ SELECT is_distributed, replication_factor FROM timescaledb_information.hypertabl
...
+
+ remote_exec 
+-------------
+ 
+(1 row)
+
+---- drop stale chunks 36, 37 on data nodes
+SELECT _timescaledb_internal.drop_stale_chunks(:'DATA_NODE_1');
+ drop_stale_chunks 
+-------------------
```
erimatnor, line 3101: Again, would be good to see output of dropped chunks here.

erimatnor (approved):
Looks like the function is doing the right thing, although I have these general comments. Perhaps some of them can be addressed in a follow-up PR.

* Would be good to see dropped stale chunks as output of the function.
* The implementation seems it could be more efficient. Right now the join seems overly complicated and requires materializing multiple lists of chunks IDs. That requires O(N) memory instead of doing the join in a scan loop using a per-tuple memory context. (Included some suggested pseudo code as inline comment).

pmwkaa:
@erimatnor 

I understand your point, but this will improve only one 1/2 of used memory. We need to hold array of ids and we need to have it sorted, which will create a List (I couldn't see a way to a way to sort directly on ArrayType). We can't rely on the fact that the supplied array is sorted, we need to do it.

Given the tight timeline I suggest to do this as a future improvement or follow up PRs as we decide.

nikkhils (approved):
It looks to me overall as a first commit. We can make additional changes later since this is not a regularly used code.

Issue #4972, opened 2022-11-14
Test dist_move_chunk hangs in PG15


lkshminarayanan:
### What type of bug is this?

Other

### What subsystems and features are affected?

Data node, Distributed hypertable

### What happened?

When compiled and run in PG15, the `dist_move_chunk` testcase hangs due to a deadlock caused between the copy_chunk function and a DROP DATABASE command run in parallel. This is not locally reproducible but hangs in Github CI.

### TimescaleDB version affected

2.9.0

### PostgreSQL version used

15.0

### What operating system did you use?

Ubuntu 22.04

### What installation method did you use?

Source

### What platform did you run on?

Not applicable

### Relevant log output and stack trace

```bash
The logs downloaded from CI has following repeated lines :

2022-11-14 07:57:38.482 UTC cluster_super_user [24872] db_exp_cagg_timezone LOG:  00000: still waiting for backend with PID 24834 to accept ProcSignalBarrier
2022-11-14 07:57:38.482 UTC cluster_super_user [24872] db_exp_cagg_timezone LOCATION:  WaitForProcSignalBarrier, procsignal.c:418
2022-11-14 07:57:38.482 UTC cluster_super_user [24872] db_exp_cagg_timezone STATEMENT:  DROP DATABASE db_exp_cagg_timezone_1;
2022-11-14 07:57:38.482 UTC cluster_super_user [24881] db_exp_cagg_origin LOG:  00000: still waiting for backend with PID 24834 to accept ProcSignalBarrier
2022-11-14 07:57:38.482 UTC cluster_super_user [24881] db_exp_cagg_origin LOCATION:  WaitForProcSignalBarrier, procsignal.c:418
2022-11-14 07:57:38.482 UTC cluster_super_user [24881] db_exp_cagg_origin STATEMENT:  DROP DATABASE db_exp_cagg_origin_1;
2022-11-14 07:57:38.482 UTC cluster_super_user [24841] db_dist_triggers LOG:  00000: still waiting for backend with PID 24834 to accept ProcSignalBarrier
2022-11-14 07:57:38.482 UTC cluster_super_user [24841] db_dist_triggers LOCATION:  WaitForProcSignalBarrier, procsignal.c:418
2022-11-14 07:57:38.482 UTC cluster_super_user [24841] db_dist_triggers STATEMENT:  DROP DATABASE db_dist_triggers_1;
[...]

PR #4973, opened 2022-11-14
Disable dist_move_chunk test in PG15

lkshminarayanan:
The dist_move_chunk causes the CI to hang when compiled and run with PG15 as explained in #4972.


akuzm (approved):

lkshminarayanan:
Thank you for the review @akuzm !

PR #4974, opened 2022-11-14
Move freeze/unfreeze chunk to tsl

gayyappan:
Move code for freeze and unfreeze chunk to tsl directory.


src/chunk.c
```diff
@@ -3549,6 +3547,17 @@ ts_chunk_unset_frozen(Chunk *chunk)
 #endif
 }
 
+bool
+ts_chunk_is_frozen(Chunk *chunk)
```
jnidzwetzki, line 3551: Is this exported function intended for later use? I could not find any calls at the moment.

jnidzwetzki (approved):
One question, rest LGTM.

akuzm (approved):

src/chunk.c
gayyappan: called from the tsl code.

PR #4977, opened 2022-11-15
Read until EOF in COPY fetcher

erimatnor:
Ensure the COPY fetcher implementation reads data until EOF with `PQgetCopyData()`. Also ensure the malloc'ed copy data is freed with `PQfreemem()` if an error is thrown in the processing loop.

Previously, the COPY fetcher didn't read until EOF, and instead assumed EOF when the COPY file trailer is received. Since EOF wasn't reached, it required terminating the COPY with an extra call to the (deprecated) `PQendcopy()` function.

Still, there are cases when a COPY needs to be prematurely terminated, for example, when querying with a LIMIT clause. Therefore, distinguish between "normal" end (when receiving EOF) and forceful end (cancel the ongoing query).

### DIFF
--- a/tsl/src/remote/copy_fetcher.c
+++ b/tsl/src/remote/copy_fetcher.c
@@ -6,6 +6,7 @@
 
 #include <postgres.h>
 #include <port/pg_bswap.h>
+#include <libpq-fe.h>
 
 #include "copy_fetcher.h"
 #include "tuplefactory.h"
@@ -18,6 +19,7 @@ typedef struct CopyFetcher
 	/* Data for virtual tuples of the current retrieved batch. */
 	Datum *batch_values;
 	bool *batch_nulls;
+	bool file_trailer_received;
 } CopyFetcher;
 
 static void copy_fetcher_send_fetch_request(DataFetcher *df);
@@ -58,6 +60,7 @@ static void
 copy_fetcher_reset(CopyFetcher *fetcher)
 {
 	fetcher->state.open = false;
+	fetcher->file_trailer_received = false;
 	data_fetcher_reset(&fetcher->state);
 }
 
@@ -71,7 +74,6 @@ copy_fetcher_send_fetch_request(DataFetcher *df)
 	if (fetcher->state.open)
 	{
 		/* data request has already been sent */
-		Assert(fetcher->state.data_req != NULL);
 		return;
 	}
 
@@ -107,13 +109,9 @@ copy_fetcher_send_fetch_request(DataFetcher *df)
 					 errmsg("could not set single-row mode on connection to \"%s\"",
 							remote_connection_node_name(fetcher->state.conn)),
 					 errdetail("The aborted statement is: %s.", fetcher->state.stmt),
-					 errhint(
-						 "Row-by-row fetching of data is not supported together with sub-queries."
-						 " Use cursor fetcher instead.")));
+					 errhint("Copy fetcher is not supported together with sub-queries."
+							 " Use cursor fetcher instead.")));
 		}
-
-		fetcher->state.data_req = req;
-		fetcher->state.open = true;
 
 		PGresult *res = PQgetResult(remote_connection_get_pg_conn(fetcher->state.conn));
 		if (res == NULL)
@@ -130,7 +128,9 @@ copy_fetcher_send_fetch_request(DataFetcher *df)
 			remote_connection_error_elog(&err, ERROR);
 		}
 
+		fetcher->state.open = true;
 		PQclear(res);
+		pfree(req);
 	}
 	PG_CATCH();
 	{
@@ -140,6 +140,7 @@ copy_fetcher_send_fetch_request(DataFetcher *df)
 		PG_RE_THROW();
 	}
 	PG_END_TRY();
+
 	MemoryContextSwitchTo(oldcontext);
 }
 
@@ -241,6 +242,75 @@ copy_data_check_header(StringInfo copy_data)
 }
 
 /*
+ * End the COPY after receiving EOF or canceling a query (e.g., due to a LIMIT
+ * being reached).
+ *
+ * This should be called after canceling a query, or, after reading all data,
+ * the file trailer, and getting an EOF return value.
+ */
+static void
+end_copy(CopyFetcher *fetcher, bool canceled)
+{
+	PGconn *conn = remote_connection_get_pg_conn(fetcher->state.conn);
+	PGresult *final_pgres = NULL;
+	PGresult *pgres = NULL;
+	ExecStatusType received_status;
+
+	Assert(fetcher->state.open);
+
+	/* Read results until NULL */
+	while ((pgres = PQgetResult(conn)))
+	{
+		if (final_pgres == NULL)
+			final_pgres = pgres;
+		else
+			remote_result_close(pgres);
+	}
+
+	received_status = PQresultStatus(final_pgres);
+	remote_result_close(final_pgres);
+
+	if (canceled)
+	{
+		/* If the query was canceled during query execution by the access node
+		 * (e.g., due to reaching a LIMIT), expect either PGRES_COMMAND_OK
+		 * (query completed before cancel happened) or PGRES_FATAL_ERROR
+		 * (query abandoned before completion) */
+		if (received_status != PGRES_COMMAND_OK && received_status != PGRES_FATAL_ERROR)
+			remote_connection_elog(fetcher->state.conn, ERROR);
+	}
+	else if (received_status != PGRES_COMMAND_OK)
+	{
+		Assert(received_status == PGRES_FATAL_ERROR || received_status == PGRES_NONFATAL_ERROR);
+		remote_connection_elog(fetcher->state.conn, ERROR);
+	}
+
+	fetcher->state.open = false;
+	remote_connection_set_status(fetcher->state.conn, CONN_IDLE);
+}
+
+/*
+ * Prematurely end the COPY before EOF is received.
+ *
+ * This happens for queries that are abandoned before reaching EOF, e.g., when
+ * using LIMITs.
+ */
+static void
+end_copy_before_eof(CopyFetcher *fetcher)
+{
+	/*
+	 * The fetcher state might not be open if the fetcher got initialized but
+	 * never executed due to executor constraints.
+	 */
+	if (!fetcher->state.open)
+		return;
+
+	Assert(!fetcher->state.eof);
+	remote_connection_cancel_query(fetcher->state.conn);
+	end_copy(fetcher, true);
+}
+
+/*
  * Process response for ongoing async request
  */
 static int
@@ -248,15 +318,12 @@ copy_fetcher_complete(CopyFetcher *fetcher)
 {
 	/* Marked as volatile since it's modified in PG_TRY used in PG_CATCH */
 	AsyncResponseResult *volatile response = NULL;
+	char *volatile dataptr = NULL;
 	MemoryContext oldcontext;
-	AsyncRequestSet *fetch_req_wrapper = async_request_set_create();
+	PGconn *conn = remote_connection_get_pg_conn(fetcher->state.conn);
 
 	Assert(fetcher->state.open);
-	Assert(fetcher->state.data_req != NULL);
-
 	data_fetcher_validate(&fetcher->state);
-
-	async_request_set_add(fetch_req_wrapper, fetcher->state.data_req);
 
 	/*
 	 * We'll store the tuples in the batch_mctx.  First, flush the previous
@@ -286,43 +353,30 @@ copy_fetcher_complete(CopyFetcher *fetcher)
 
 			StringInfoData copy_data = { 0 };
 
-			copy_data.len = PQgetCopyData(remote_connection_get_pg_conn(fetcher->state.conn),
+			copy_data.len = PQgetCopyData(conn,
 										  &copy_data.data,
 										  /* async = */ false);
 
+			/* Set dataptr to ensure data is freed with PQfreemem() in
+			 * PG_CATCH() clause in case error is thrown. */
+			dataptr = copy_data.data;
+
 			if (copy_data.len == -1)
 			{
-				/*
-				 * According to the docs, this means EOF, but in practice we get
-				 * it when there is an error. Check for error and report it.
-				 */
+				/* Note: it is possible to get EOF without having received the
+				 * file trailer in case there's e.g., a remote error. */
 				fetcher->state.eof = true;
-
-				PGresult *res = PQgetResult(remote_connection_get_pg_conn(fetcher->state.conn));
-				if (res == NULL)
-				{
-					/* Shouldn't really happen but technically possible. */
-					TSConnectionError err;
-					remote_connection_get_error(fetcher->state.conn, &err);
-					remote_connection_error_elog(&err, ERROR);
-				}
-				if (PQresultStatus(res) != PGRES_COPY_OUT)
-				{
-					TSConnectionError err;
-					remote_connection_get_result_error(res, &err);
-					remote_connection_error_elog(&err, ERROR);
-				}
+				/* Should read final result with PQgetResult() until it
+				 * returns NULL. This happens below. */
 				break;
 			}
 			else if (copy_data.len == -2)
 			{
 				/*
 				 * Error. The docs say: consult PQerrorMessage() for the reason.
-				 * remote_connection_get_error() will do this for us.
+				 * remote_connection_elog() will do this for us.
 				 */
-				TSConnectionError err;
-				remote_connection_get_error(fetcher->state.conn, &err);
-				remote_connection_error_elog(&err, ERROR);
+				remote_connection_elog(fetcher->state.conn, ERROR);
 			}
 
 			copy_data.maxlen = copy_data.len;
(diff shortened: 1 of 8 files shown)


erimatnor:
Note: The diff looks bigger than it is because some code is at different indentation.

tsl/src/remote/copy_fetcher.c
```diff
@@ -104,12 +107,11 @@ copy_fetcher_send_fetch_request(DataFetcher *df)
 		{
 			ereport(ERROR,
 					(errcode(ERRCODE_CONNECTION_FAILURE),
-					 errmsg("could not set single-row mode on connection to \"%s\"",
+					 errmsg("could not set COPY mode on connection to \"%s\"",
```
akuzm, line 110: The previous text was correct, it fails to set single-row mode here. COPY mode is set by sending the COPY query.

tsl/src/remote/copy_fetcher.c
```diff
@@ -556,35 +532,49 @@ copy_fetcher_close(DataFetcher *df)
...
+		while ((pgres = PQgetResult(conn)))
+		{
+			if (final_pgres == NULL)
+				final_pgres = pgres;
+			else
+				remote_result_close(pgres);
+			num_results++;
+		}
+
+		Assert(num_results == 1);
```
akuzm, line 559: I'd make this an `elog` because this is closer to an environment error, we don't control the remote server.

akuzm: Probably would also be good to see what other PQResults we got instead of the single one we expected. Although I'm not sure if this ever happens.

akuzm (approved):

akuzm:
By the way, is this covered by the old tests that fail on PG 15, or we need something else?

tsl/src/remote/copy_fetcher.c
```diff
@@ -556,35 +532,49 @@ copy_fetcher_close(DataFetcher *df)
...
-		 */
-		PGresult *pgres = PQgetResult(remote_connection_get_pg_conn(fetcher->state.conn));
-		if (pgres)
+		/* Read results until NULL */
+		while ((pgres = PQgetResult(conn)))
+		{
+			if (final_pgres == NULL)
+				final_pgres = pgres;
+			else
+				remote_result_close(pgres);
```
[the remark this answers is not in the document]
erimatnor: We don't strictly need it, but I think a goal was initially to wrap all libpq stuff in our own APIs. But, TBH, we haven't been consistent about it. The internals don't really matter, but provides to ability to do additional debugging, for example.

tsl/src/remote/copy_fetcher.c
erimatnor: There's a check above for eof being set, so we shouldn't get here if the connection isn't in the right state. This is what the assert checks. For tests, it is good to crash here for something that shouldn't happen.

akuzm:
> PQendCopy doesn't cancel the query, right? If we want to remove it because it's deprecated, maybe for the premature end case we can cancel the query + do a loop with PQgetCopyData discarding the results?

Maybe I lost track of the original rationale for the PR. I thought we're just removing it because it's deprecated. If we're now adding it back, there should be some other rationale, right? Some broken test on PG 15? Ideally I think it would be good to remove it using the approach I mentioned above.

erimatnor:
> > PQendCopy doesn't cancel the query, right? If we want to remove it because it's deprecated, maybe for the premature end case we can cancel the query + do a loop with PQgetCopyData discarding the results?
> 
> Maybe I lost track of the original rationale for the PR. I thought we're just removing `PQendCopy` because it's deprecated. If we're now adding it back, there should be some other rationale for the changes, right? Some broken test on PG 15? Ideally I think it would be good to remove it using the approach I mentioned above.

Sorry for the confusion.

Removing `PQendCopy()` was not the rationale, but I thought the function call wasn't necessary initially. The main part of the change was to properly read to EOF instead of invoking `PQendCopy()` when reading the fail trailer. It turns out, however, that we need to prematurely end the COPY in some case, e.g., when abandoning a scan due to a LIMIT or similar. The only way to do that is `PQendCopy()`, AFAICT. 

So, the difference from before is that we now read to EOF in the normal case and only abort the COPY with `PQendCopy()` for other cases. Previously, we aborted all scans with `PQendCopy()`, which wasn't the "correct" use of the API.

akuzm:
Looking at the source, `PQendCopy()` doesn't end copy, it just does `PQgetResult()` and verifies that it is PGRES_COMMAND_OK. If the only reason is to avoid the deprecated function, let's replace this call with `PQgetResult`. I'd avoid adding another code path and another variable. The existing code is already complicated and poorly tested.

Speaking about tests, I'm not sure how early cancellation can work with `PQendCopy()`. From reading the source, it looks like if the remote server is still sending the COPY data, `PQendCopy` will fail because it expects `PGRES_COMMAND_OK`, not `PGRES_COPY_OUT` as would be the case in COPY mode. The second problem is that we should cancel the remote query in this case, and it seemingly doesn't exist. Do we have a test case that covers this?

erimatnor:
> OK, I found where this happens, `pqParseInput3` silently drops the copy data `'d'` messages. So if we decide to stop COPY early, it is technically enough to call `PQgetResult` which will skip the copy data for us. But from a performance standpoint we should also cancel the query.

Thanks, I will try that.

erimatnor:
> Looking at the source, `PQendCopy()` doesn't end copy, it just does `PQgetResult()` and verifies that it is PGRES_COMMAND_OK. If the only reason is to avoid the deprecated function, let's replace this call with `PQgetResult`. I'd avoid adding another code path and another variable. The existing code is already complicated and poorly tested.
> 
> Speaking about tests, I'm not sure how early cancellation can work with `PQendCopy()`. From reading the source, it looks like if the remote server is still sending the COPY data, `PQendCopy` will fail because it expects `PGRES_COMMAND_OK`, not `PGRES_COPY_OUT` as would be the case in COPY mode. The second problem is that we should cancel the remote query in this case, and there's seemingly no code to do that. Do we have a test case that covers this?

We have a function to cancel a query. So, I will use that one.

FWIW, it is odd that the postgresql protocol does not have a way to do control command signaling across the same connection, to e.g., stop an ongoing query. What the libpq cancel function does is open up a new connection to cancel the query.

tsl/src/remote/copy_fetcher.c
```diff
@@ -338,85 +385,101 @@ copy_fetcher_complete(CopyFetcher *fetcher)
...
-
 			PQfreemem(copy_data.data);
+			dataptr = NULL;
 		}
 
-		fetcher->state.num_tuples = row;
+		/* Don't count the file trailer as a row if this was the last batch */
+		fetcher->state.num_tuples = fetcher->file_trailer_received ? row - 1 : row;
 		fetcher->state.next_tuple_idx = 0;
+
```
akuzm, line 482: Should we check that the file trailer is received here as well?

akuzm: Github is being weird, I mean in the if below

tsl/src/remote/copy_fetcher.c
```diff
@@ -338,85 +385,101 @@ copy_fetcher_complete(CopyFetcher *fetcher)
...
-			}
+				/*
+				 * From the PostgreSQL (libpq) docs: The file trailer consists
+				 * of a 16-bit integer word containing -1. This is easily
+				 * distinguished from a tuple's field-count word. A reader
+				 * should report an error if a field-count word is neither -1
+				 * nor the expected number of columns. This provides an extra
+				 * check against somehow getting out of sync with the data.
+				 */
+				fetcher->file_trailer_received = true;
```
akuzm, line 396: Why do we need separate logic for the file trailer as compared to other kinds of eof? We do the same thing for them anyway, right?

tsl/src/remote/copy_fetcher.c
```diff
@@ -550,41 +584,13 @@ copy_fetcher_close(DataFetcher *df)
...
-				err.msg = "internal program error: remaining activity on the data node connection "
-						  "after finishing COPY";
-			}
-			remote_connection_error_elog(&err, ERROR);
-		}
+	/* Check for premature ending of COPY (i.e., before reaching EOF). Note,
+	 * if EOF was reached, we closed already in the main processing loop so no
+	 * need to end_copy() here in that case. */
+	if (!fetcher->state.eof)
+		end_copy_before_eof(fetcher);
```
akuzm, line 591: If we already received the file trailer, we shouldn't cancel the query, right? It's finished already.

tsl/src/remote/copy_fetcher.c
```diff
@@ -240,6 +241,66 @@ copy_data_check_header(StringInfo copy_data)
...
+			final_pgres = pgres;
+		else
+			remote_result_close(pgres);
+	}
+
+	received_status = PQresultStatus(final_pgres);
+	remote_result_close(final_pgres);
+
+	if (received_status != expected_status)
+		remote_connection_elog(fetcher->state.conn, ERROR);
```
akuzm, line 277: This might give us messages like "Fatal error: success". Let's also add a line about what kind of status we were expecting and that it didn't match.

tsl/src/remote/copy_fetcher.c
```diff
@@ -240,6 +241,66 @@ copy_data_check_header(StringInfo copy_data)
...
+end_copy_before_eof(CopyFetcher *fetcher)
+{
+	/*
+	 * The fetcher state might not be open if the fetcher got initialized but
+	 * never executed due to executor constraints.
+	 */
+	if (!fetcher->state.open)
+		return;
+
+	Assert(!fetcher->state.eof);
```
akuzm, line 299: Probably also doesn't make sense to cancel if we have received the file trailer. I'm just going through all usages of `trailer`/`eof` to check if they're used in the same way.

tsl/src/remote/copy_fetcher.c
```diff
@@ -338,85 +385,101 @@ copy_fetcher_complete(CopyFetcher *fetcher)
...
 			}
-
-			Datum *values = &fetcher->batch_values[tupdesc_natts * row];
-			bool *nulls = &fetcher->batch_nulls[tupdesc_natts * row];
-			for (int i = 0; i < tupdesc_natts; i++)
+			else
 			{
-				nulls[i] = true;
-			}
+				Assert(!fetcher->file_trailer_received);
```
akuzm, line 405: `Assert(!eof)` also must hold, so the usage of `trailer`/`eof` is the same here again.

tsl/src/remote/copy_fetcher.c
```diff
@@ -286,32 +344,21 @@ copy_fetcher_complete(CopyFetcher *fetcher)
...
-					remote_connection_error_elog(&err, ERROR);
-				}
-				if (PQresultStatus(res) != PGRES_COPY_OUT)
-				{
-					TSConnectionError err;
-					remote_connection_get_result_error(res, &err);
-					remote_connection_error_elog(&err, ERROR);
-				}
+				/* Should read final result with PQgetResult() until it
+				 * returns NULL. This happens when the fetcher is closed. */
```
akuzm, line 361: Looks like it will happen in line 497 instead (the call to `end_copy()` at the end of the top-level loop).

tsl/src/remote/copy_fetcher.c
erimatnor: You mean to assert it?

tsl/src/remote/copy_fetcher.c
erimatnor: I guess, but I don't really see how it would happen without us also reading EOF. And, if we would cancel at file trailer without receiving EOF, then that's fine too because if we explicitly cancel it is not an error case.

tsl/src/remote/copy_fetcher.c
erimatnor: The file trailer and EOF are technically different things.

The file trailer _is_ data in the stream. EOF is the return value from `PGgetCopyData()`. I imagine that one could in theory send two or more files in the same COPY stream, delineated by a file trailer. EOF indicates end of the stream. These are technically two different things.

I prefer to treat the file trailer as data, and not do special casing for what to do in that case. Instead, just continue reading in the loop until `PGgetCopyData()` return -1 as specified by the protocol. Whether we read EOF  here or in the next iteration of the loop doesn't matter technically, but it seems more elegant to me to have `PGgetCopyData()` in  only one place.

tsl/src/remote/copy_fetcher.c
akuzm: I think we shouldn't cancel at file trailer, it's just another kind of eof, so cancelling would be pointless extra work, and it's heavy.

> I guess, but I don't really see how it would happen without us also reading EOF.

Let's add an assertion then. I just want to make sure that `trailer` and `eof` almost always have the same value, and the only exception is `trailer = true and eof = false` while we're making the final `PGgetCopyData` call which is just a sanity check.

tsl/src/remote/copy_fetcher.c
akuzm: yep

erimatnor: Actually, EOF can be read without the file trailer being received. For example, when an error or cancel happens at the other end.

tsl/src/remote/copy_fetcher.c
akuzm: > I imagine that one could in theory send two or more files in the same COPY stream, delineated by a file trailer.

This doesn't happen and it's not something that we should try to support. I agree that we should add a check that the file trailer is followed by EOF. This means that in practice, 'eof' and 'trailer' are the same thing.

> it seems more elegant to me to have PQgetCopyData() in only one place.

What's wrong with adding one call and checking that it returns EOF? Instead of that, we have two flags with barely distinguishable meaning, both of which we have to check in several other places. Basically we're using the difference between these flags to make additional iteration of the loop, which should only call `PQgetCopyData`. To me this looks confusing and unnecessarily complicated.

tsl/src/remote/copy_fetcher.c
akuzm: Looks like we get to this branch only when there's no error. EOF with error does `ereport` earlier.

tsl/src/remote/copy_fetcher.c
erimatnor: > What's wrong with adding one call and checking that it returns EOF? Instead of that, we have two flags with barely distinguishable meaning, both of which we have to check in several other places. Basically we're using the difference between these flags to make additional iteration of the loop, which should only call `PQgetCopyData`. To me this looks confusing and unnecessarily complicated.

I guess there are pros and cons with both approaches. But the cons are bigger in my book.

The flags effectively express the  state machine that the fetcher implements. Perhaps it makes sense to turn them into an enum: RECEIVING, FILE_TRAILER, EOF, or similar, instead of separate booleans in case you are concerned about multiple flags. I see nothing wrong with explicitly expressing the state machine. I imagine that for many that would actually bring clarity in what's going on as opposed to it being implicit.

The main reason I don't think your suggest approach is great, however,  is because it requires checking for not only the expected condition in both places where `PQgetCopyData()` is called, but also handle the conditions that are not expected. `PQgetCopyData()` returns three states represented by the values  >=0, -1, -2 for receiving data, EOF, and error, respectively. If I check for those conditions in one place, all of them will likely be executed at some point. But, if I split the handling of those states across different places in the code, I need to check for the "expected" condition but also add error handling for the unexpected ones. The unexpected condition is likely to never going to be executed, with more dead code as a result and worse CodeCov.

Further, if I read EOF as a special case after receiving the file trailer, I also need to break out of the loop. But that means I need to duplicate some code to free memory read for the file trailer which is normally done at the end of the loop (thus having `PQmemfree()` in two places as well). I also need to modify `fetcher->state.num_tuples` in two places to account for having received a legitimate data tuple vs a file trailer.

To summarize, I guess there are pros and cons with both approaches, although I personally don't like special-casing like you suggest. It is, IMO, cleaner with a general loop where you handle _all_ return values of a function in one place. Right now I am not convinced by your arguments for this change that has no functional improvement. It will only increase the amount of similar code and also dead code.

I suggest keeping this as is unless the next reviewer also feels strongly that we should change it.

tsl/src/remote/copy_fetcher.c
```diff
@@ -240,6 +241,66 @@ copy_data_check_header(StringInfo copy_data)
...
+	 * due to a LIMIT), expect FATAL_ERROR (query failed) */
+	ExecStatusType expected_status = fetcher->state.eof ? PGRES_COMMAND_OK : PGRES_FATAL_ERROR;
+	ExecStatusType received_status;
+
+	Assert(fetcher->state.open);
+
+	/* Read results until NULL */
+	while ((pgres = PQgetResult(conn)))
+	{
+		if (final_pgres == NULL)
```
65278, line 267: Help me across the street here. final_pgres starts out as NULL, so the branch is true on the first time. So the first result is the final result? 

Apart of that, if we move  `received_status = PQresultStatus(final_pgres);`  into this branch and initialised `ExecStatusType received_status = PGRES_FATAL_ERROR;` , we can do away with the else branch and just call `remote_result_status(pgres)` unconditionally here.

65278 (commented):
Apart of one question, this looks good to me. The discussion on copy file trailers vs EOF exhausts my knowledge, so I can't possibly comment.

tsl/src/remote/copy_fetcher.c
akuzm: > Help me across the street here. final_pgres starts out as NULL, so the branch is true on the first time. So the first result is the final result?

We expect there to be only one PGResult. Maybe we should change this code to complain if there are more PGResults than one, or even simpler, just request the first result and not check what happens next. This is what `PQendCopy` does.

65278: Thanks for the explanation. I'll say that the intent of the code isn't too obvious from the code. Also if there are extra results, that definitely warrants a warning.

erimatnor: Yes, we expect one result, but the libpq doc explicitly says: `PQgetResult must be called repeatedly until it returns a null pointer, indicating that the command is done.` So, following suggested practice here.

tsl/src/remote/copy_fetcher.c
```diff
@@ -240,6 +241,66 @@ copy_data_check_header(StringInfo copy_data)
...
+ */
+static void
+end_copy(CopyFetcher *fetcher)
+{
+	PGconn *conn = remote_connection_get_pg_conn(fetcher->state.conn);
+	PGresult *final_pgres = NULL;
+	PGresult *pgres = NULL;
+	/* Expect COMMAND OK if we received EOF. If the query was canceled (e.g.,
+	 * due to a LIMIT), expect FATAL_ERROR (query failed) */
+	ExecStatusType expected_status = fetcher->state.eof ? PGRES_COMMAND_OK : PGRES_FATAL_ERROR;
```
nikkhils, line 259: In case of LIMIT, why are we setting FATAL_ERROR? We want to close early and with SUCCESS, no?

tsl/src/remote/copy_fetcher.c
```diff
@@ -338,85 +383,102 @@ copy_fetcher_complete(CopyFetcher *fetcher)
...
-					 */
-					if (!attconv->conv_funcs[att].fn_strict)
-					{
-						values[att] = ReceiveFunctionCall(&attconv->conv_funcs[att],
-														  NULL,
-														  attconv->ioparams[att],
-														  attconv->typmods[att]);
-					}
-					else
+					nulls[i] = true;
```
nikkhils, line 425: `memset(nulls, 1, sizeof(nulls));` might be better than a loop

nikkhils (approved):
LGTM in general. Complicated, hard to follow code this

tsl/src/remote/copy_fetcher.c
erimatnor: Ok, I'll consider this change for a refactoring PR as a follow up.

tsl/src/remote/copy_fetcher.c
erimatnor: This is not us. This is the error returned from the data node where the query is canceled. From `PQcancel()` docs:

" If the cancellation is effective, the current command will terminate early and return an error result."

erimatnor: Actually, when we cancel, the docs says we can receive either COMMAND_OK or ERROR. For example, the query could have just completed when the canceled happened. 

I will tweak the code to account for this case and make things more clear.

tsl/src/remote/copy_fetcher.c
erimatnor: This was old code, just indented differently. Also, looks like the loop is necessary anyway and I guess nulls was initialized to "false" so this is only called in the "true" case.  I will leave further changes out of this PR.

tsl/src/remote/copy_fetcher.c
erimatnor: It is not strictly wrong to cancel at file trailer before reading EOF. This is very unlikely to happen, though, so why optimize for it?

erimatnor: Added an extra Assert(), although not sure this is what you were looking for?

PR #4980, opened 2022-11-15
Fix GITHUB_OUTPUT on Windows

akuzm:
We have to add it to WSLENV variable or else it's not passed to the WSL processes.

Disable-check: commit-count

### DIFF
--- a/.github/workflows/linux-32bit-build-and-test.yaml
+++ b/.github/workflows/linux-32bit-build-and-test.yaml
@@ -44,7 +44,7 @@ jobs:
         apt-get update
         apt-get install -y gnupg postgresql-common
         yes | /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh
-        apt-get install -y gcc make cmake libssl-dev libkrb5-dev libipc-run-perl libtest-most-perl sudo gdb git wget
+        apt-get install -y gcc make cmake libssl-dev libkrb5-dev libipc-run-perl libtest-most-perl sudo gdb git wget gawk
         apt-get install -y postgresql-${PG_MAJOR} postgresql-server-dev-${PG_MAJOR}
 
     - name: Build pg_isolation_regress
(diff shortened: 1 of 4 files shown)


.github/workflows/windows-build-and-test.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
       run: python .github/gh_config_reader.py
     - name: Set build_type
       id: build_type
+      env:
+        WSLENV: GITHUB_OUTPUT/p
```
jnidzwetzki, line 29: For the other actions that use the environment variable `GITHUB_OUTPUT`, we also set `shell: wsl-bash {0}`. Is this intentionally not set here?

jnidzwetzki (approved):
One minor question, rest LGTM

.github/workflows/windows-build-and-test.yaml
akuzm: This is the config action that runs on ubuntu-latest, so bash is the default shell.

akuzm: I guess I have to remove the WSLENV...

fabriziomello (approved):

.github/workflows/linux-32bit-build-and-test.yaml
```diff
@@ -82,24 +82,33 @@ jobs:
...
+        # wait in case there are in-progress coredumps
+        sleep 10
         if compgen -G "/tmp/core*" > /dev/null; then
           PG_MAJOR=$(echo "${{ matrix.pg }}" | sed -e 's![.].*!!')
           apt-get install postgresql-${PG_MAJOR}-dbgsym >/dev/null
           for file in /tmp/core*
           do
-            gdb /usr/lib/postgresql/${PG_MAJOR}/bin/postgres -c $file <<EOT | tee -a stacktraces.log
-              printf "%s\n\n", debug_query_string
+            gdb /usr/lib/postgresql/${PG_MAJOR}/bin/postgres -c $file <<<"
```
akuzm, line 101: For some reason, heredoc doesn't always work correctly in these scripts.

PR #4986, opened 2022-11-16
Run all the tests on every commit in main

akuzm:
If we test every commit in main, we can allow GitHub to merge the PRs automatically without requiring a manual rebase on the current master. These rebases are a real time sink.

Also it will help to get more statistics for our test result database, and get more coverage for probabilistic tests like sqlsmith.

### DIFF
--- a/.github/workflows/coccinelle.yaml
+++ b/.github/workflows/coccinelle.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches:
+      - main
       - prerelease_test
 jobs:
   coccinelle:
--- a/.github/workflows/code_style.yaml
+++ b/.github/workflows/code_style.yaml
@@ -1,10 +1,8 @@
 name: Code style
 on:
-  schedule:
-    # run daily 20:00 on main branch
-    - cron: '0 20 * * *'
   push:
     branches:
+      - main
       - prerelease_test
   pull_request:
 jobs:
--- a/.github/workflows/cron-tests.yaml
+++ b/.github/workflows/cron-tests.yaml
@@ -1,10 +1,8 @@
 name: Additional tests
 on:
-  schedule:
-    # run daily 20:00 on main branch
-    - cron: '0 20 * * *'
   push:
     branches:
+      - main
       - prerelease_test
 jobs:
   config:
--- a/.github/workflows/linux-32bit-build-and-test.yaml
+++ b/.github/workflows/linux-32bit-build-and-test.yaml
@@ -1,10 +1,8 @@
 name: Regression Linux i386
 on:
-  schedule:
-    # run daily 20:00 on main branch
-    - cron: '0 20 * * *'
   push:
     branches:
+      - main
       - prerelease_test
   pull_request:
 jobs:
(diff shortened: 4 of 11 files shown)


.github/workflows/apt-arm-packages.yaml
```diff
@@ -1,13 +1,11 @@
...
 on:
-  schedule:
-    # run daily 0:00 on main branch
-    - cron: '0 0 * * *'
   push:
     tags:
     - '*'
     branches:
+    - main
     - release_test
```
akuzm, line 9: What's that, by the way? We don't have the 'release_test' branch.

svenklemm (changes requested):
I dont think the package tests should be changed by this

akuzm:
```
$ ~/clickhouse local -S "time DateTime('CET')" -q "select toHour(time) t, bar(count(*), 0, 100) from table group by t order by t" < <(git log --pretty=format:%at | head -1000)
0	████████████████▋
1	███████▏
2	█████▌
3	████
4	██████▍
5	████████▋
6	████████
7	█████████████████▌
8	███████████████████████████▏
9	██████████████████████████████████████████████████▍
10	███████████████████████████████████████████████████████████████████▏
11	████████████████████████████████████████████▋
12	█████████████████████████████████████████████████████▌
13	████████████████████████████████████████████████████████████████████
14	███████████████████████████████████████████████████████▏
15	█████████████████████████████████████████████████████████▌
16	████████████████████████████████████████████████████▋
17	███████████████████████████████████████████████████████▏
18	███████████████████████████████████████▏
19	████████████████████████████████▋
20	█████████████████████████████████████▌
21	████████████████████████▋
22	███████████████████████▏
23	████████████████████████████████████
```

Number of commits by hour in CET. Just wondering what would be the best time to run checks that are not per-commit, but at the same time require the build cache, which is per-day...

.github/workflows/abi.yaml
```diff
@@ -8,11 +8,9 @@
 # in a new minor postgresql version.
 name: ABI Test
 on:
-  schedule:
-    # run daily 20:00 on main branch
-    - cron: '0 20 * * *'
```
svenklemm, line 13: Since the purpose of this is to detect incompatible upstream changes we might wanna keep it on schedule and not triggered by commit in our repo

.github/workflows/homebrew.yaml
```diff
@@ -1,13 +1,11 @@
 # Test installation of our homebrew tap for latest version
 name: Homebrew
 on:
-  schedule:
-    # run daily 20:00 on main branch
-    - cron: '0 20 * * *'
```
svenklemm, line 6: since this is a package test whose result is not changed by commits to main this should not be changed

.github/workflows/apt-arm-packages.yaml
svenklemm: i guess this was supposed to be used to trigger all the package tests but we never started using it

akuzm:
> @svenklemm, @RafiaSabih: please review this pull request.

No idea why it decided to act now and not before :) @konskov should we finish configuring it?

konskov:
@akuzm you and @jnidzwetzki were right and the reason reviewer assignment works here is because this PR was not opened from a fork but within the repo. One solution Jan pointed out would be to use `pull_request_target` in the pull-review workflow. It's just a one-line change. But I am uncertain if there are security implications and what they are.

svenklemm (approved):

sb230132 (approved):

PR #4990, opened 2022-11-17
INSERT .. SELECT on distributed hypertable fails on PG15

### LINKED ISSUE #4983: [Bug]: `INSERT INTO distributed_hypertable SELECT * FROM another_distributed_hypertable;` fails in PG15
### What type of bug is this?

Unexpected error

Data node, Distributed hypertable

### What happened?

Given two distributed_hypertables hyper1 and hyper2, inserting tuples fetched from hyper1 into hyper2 using the `INSERT INTO hyper2 SELECT * FROM hyper1` query fails if it was preceded by a `INSERT INTO hyper1 ...` query.

The query returns the following error :
```
postgres=# INSERT INTO hyper2 SELECT * FROM hyper1;
ERROR:  request already in progress on port 5432
```

It should have instead succeeded after inserting the tuples from hyper1 into hyper2.

The `data_node` testcase in PG15 CI fails due to the same issue.

2.9.0

### PostgreSQL version used

15.0

Ubuntu 22.04

Source

On prem/Self-hosted

_No response_

```bash
-- Run the following SQL queries in Timescale 2.9.0 on PG15.

SELECT node_name, database, node_created, database_created, extension_created   
FROM add_data_node('data_node_1', host => 'localhost', database => 'TESTDB_1');
SELECT node_name, database, node_created, database_created, extension_created
FROM add_data_node('data_node_2', host => 'localhost', database => 'TESTDB_2');

CREATE TABLE hyper1 (time timestamptz, location int, temp float);               
CREATE TABLE hyper2 (LIKE hyper1); 

SELECT create_distributed_hypertable('hyper1', 'time', 'location');
SELECT create_distributed_hypertable('hyper2', 'time', 'location');

INSERT INTO hyper1 VALUES ('2022-01-01 00:00:00', 10, 23.45);
INSERT INTO hyper2 SELECT * FROM hyper1;
```

sb230132:
INSERT .. SELECT query containing distributed hypertables generates plan with DataNodeCopy node which is not supported. Issue is in function tsl_create_distributed_insert_path() where we decide if we should generate DataNodeCopy or DataNodeDispatch node based on the kind of query. In PG15 for INSERT .. SELECT query timescaledb planner generates DataNodeCopy as rte->subquery is set to NULL. This is because of a commit in PG15 where rte->subquery is set to NULL as part of a fix.

This patch checks if SELECT subquery has distributed hypertables or not by looking into root->parse->jointree which represents subquery.

Fixes #4983


tsl/src/planner.c
```diff
@@ -315,12 +315,47 @@ tsl_create_distributed_insert_path(PlannerInfo *root, ModifyTablePath *mtpath, I
 					if (rte->rtekind == RTE_SUBQUERY)
 					{
 						distributed = false;
+#if PG15_GE
+						Node *jtnode = (Node *) root->parse->jointree;
+						if (IsA(jtnode, FromExpr))
```
akuzm, line 320: Can we put this code for traversing the jointree hierarchy into the `distributed_rtes_wakler`?

tsl/src/planner.c
```diff
@@ -315,12 +315,47 @@ tsl_create_distributed_insert_path(PlannerInfo *root, ModifyTablePath *mtpath, I
 					if (rte->rtekind == RTE_SUBQUERY)
 					{
 						distributed = false;
+#if PG15_GE
```
akuzm, line 318: Does this new way works on the older PG versions? If it does, let's make it the only way, to support less variants of the code.

akuzm (approved):

tsl/src/planner.c
sb230132: Not possible, since i dont have all context to parse jointree in distributed_rtes_wakler().

tsl/src/planner.c
sb230132: Right, made code common for all versions.

nikkhils (commented):
LGTM! Initially I was confused as to where the jointree was getting the rte from. Now I can see it comes from the `rtable`.

nikkhils (approved):
LGTM! Initially I was confused as to where the jointree was getting the rte from. Now I can see it comes from the rtable.

PR #4991, opened 2022-11-17
Add a GUC variable to enable/disable OSM reads

zilder:
The GUC variable controls whether OSM chunk is used in a select query plan.

### DIFF
--- a/src/guc.c
+++ b/src/guc.c
@@ -73,6 +73,7 @@ bool ts_guc_enable_constraint_exclusion = true;
 bool ts_guc_enable_qual_propagation = true;
 bool ts_guc_enable_cagg_reorder_groupby = true;
 bool ts_guc_enable_now_constify = true;
+bool ts_guc_enable_osm_reads = true;
 TSDLLEXPORT bool ts_guc_enable_transparent_decompression = true;
 bool ts_guc_enable_per_data_node_queries = true;
 bool ts_guc_enable_async_append = true;
@@ -285,6 +286,18 @@ _guc_init(void)
 							 "Enable the optimization that combines different chunks belonging to "
 							 "the same hypertable into a single query per data_node",
 							 &ts_guc_enable_per_data_node_queries,
+							 true,
+							 PGC_USERSET,
+							 0,
+							 NULL,
+							 NULL,
+							 NULL);
+
+	DefineCustomBoolVariable("timescaledb.enable_tiered_reads",
+							 "Enable tiered data reads",
+							 "Enable reading of tiered data by including a foreign table "
+							 "representing the data in the object storage into the query plan",
+							 &ts_guc_enable_osm_reads,
 							 true,
 							 PGC_USERSET,
 							 0,
--- a/src/hypertable_restrict_info.c
+++ b/src/hypertable_restrict_info.c
@@ -20,6 +20,7 @@
 #include "dimension.h"
 #include "dimension_slice.h"
 #include "dimension_vector.h"
+#include "guc.h"
 #include "hypercube.h"
 #include "partitioning.h"
 #include "scan_iterator.h"
@@ -664,6 +665,18 @@ ts_hypertable_restrict_info_get_chunks(HypertableRestrictInfo *hri, Hypertable *
 		 * No restrictions on hyperspace. Just enumerate all the chunks.
 		 */
 		chunk_ids = ts_chunk_get_chunk_ids_by_hypertable_id(ht->fd.id);
+
+		/*
+		 * If the hypertable has an OSM chunk it would end up in the list
+		 * as well. We need to remove it when OSM reads are disabled via GUC
+		 * variable.
+		 */
+		if (!ts_guc_enable_osm_reads)
+		{
+			int32 osm_chunk_id = ts_chunk_get_osm_chunk_id(ht->fd.id);
+
+			chunk_ids = list_delete_int(chunk_ids, osm_chunk_id);
+		}
 	}
 	else
 	{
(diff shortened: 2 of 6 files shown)


src/hypertable_restrict_info.c
```diff
@@ -686,16 +699,21 @@ ts_hypertable_restrict_info_get_chunks(HypertableRestrictInfo *hri, Hypertable *
...
 		int32 osm_chunk_id = ts_chunk_get_osm_chunk_id(ht->fd.id);
-		if (osm_chunk_id != 0 && !list_member_int(chunk_ids, osm_chunk_id))
+
+		if (osm_chunk_id != 0)
 		{
-			chunk_ids = lappend_int(chunk_ids, osm_chunk_id);
+			chunk_ids = !ts_guc_enable_osm_reads ? list_delete_int(chunk_ids, osm_chunk_id) :
+						!list_member_int(chunk_ids, osm_chunk_id) ?
+												   lappend_int(chunk_ids, osm_chunk_id) :
+												   chunk_ids;
```
gayyappan, line 716: I am surprised we have a code cov issue here. Could you take a look?

src/hypertable_restrict_info.c
```diff
@@ -686,16 +699,21 @@ ts_hypertable_restrict_info_get_chunks(HypertableRestrictInfo *hri, Hypertable *
...
+		 * so we have to check if it's already there not to add a duplicate.
+		 * Similarly if OSM reads are disabled then we exclude the OSM chunk.
 		 */
 		int32 osm_chunk_id = ts_chunk_get_osm_chunk_id(ht->fd.id);
-		if (osm_chunk_id != 0 && !list_member_int(chunk_ids, osm_chunk_id))
+
+		if (osm_chunk_id != 0)
 		{
-			chunk_ids = lappend_int(chunk_ids, osm_chunk_id);
+			chunk_ids = !ts_guc_enable_osm_reads ? list_delete_int(chunk_ids, osm_chunk_id) :
```
jnidzwetzki, line 713: The nested expression is difficult to understand. Is it possible to write it down in a more verbose way?

src/hypertable_restrict_info.c
```diff
@@ -686,16 +699,21 @@ ts_hypertable_restrict_info_get_chunks(HypertableRestrictInfo *hri, Hypertable *
...
+		 * enabled. It has some virtual dimension slices (at the moment,
+		 * (+inf, +inf) slice for time, but it used to be different and might
+		 * change again.) So sometimes it will match and sometimes it won't,
+		 * so we have to check if it's already there not to add a duplicate.
+		 * Similarly if OSM reads are disabled then we exclude the OSM chunk.
 		 */
 		int32 osm_chunk_id = ts_chunk_get_osm_chunk_id(ht->fd.id);
-		if (osm_chunk_id != 0 && !list_member_int(chunk_ids, osm_chunk_id))
+
+		if (osm_chunk_id != 0)
```
jnidzwetzki, line 711:
```suggestion
		if (osm_chunk_id != INVALID_CHUNK_ID)
```

src/guc.c
```diff
@@ -292,6 +293,17 @@ _guc_init(void)
 							 NULL,
 							 NULL);
 
+	DefineCustomBoolVariable("timescaledb.enable_tiered_reads",
```
jnidzwetzki, line 296: The setting is called `enable_tiered_reads` the variable is called `ts_guc_enable_osm_reads`. Is this different terminology (tiered vs. osm) intentional?

src/guc.c
```diff
@@ -292,6 +293,17 @@ _guc_init(void)
...
+							 "Enable tiered data reads",
+							 NULL,
```
jnidzwetzki, line 298: Most other GUCs have a more detailed explanation as third parameter. It would be good if we could add this explanation to this setting as well.

src/guc.c
zilder: It's intentianal. We decided that we don't want to expose the name "OSM" in the user's api, but internally we refer to it as OSM.

src/guc.c
zilder: Added a long description

src/hypertable_restrict_info.c
zilder: Fixed, thanks

src/hypertable_restrict_info.c
zilder: I converted it into a regular if/else statement

src/hypertable_restrict_info.c
zilder: I rewrote it into an if/else statement after Jan's suggestion. Codecov now complains about the closing braces, but I don't think I can do anything about it.

src/hypertable_restrict_info.c
```diff
@@ -686,16 +699,25 @@ ts_hypertable_restrict_info_get_chunks(HypertableRestrictInfo *hri, Hypertable *
...
-		if (osm_chunk_id != 0 && !list_member_int(chunk_ids, osm_chunk_id))
+
+		if (osm_chunk_id != INVALID_CHUNK_ID)
 		{
-			chunk_ids = lappend_int(chunk_ids, osm_chunk_id);
+			if (!ts_guc_enable_osm_reads)
+			{
+				chunk_ids = list_delete_int(chunk_ids, osm_chunk_id);
+			}
+			else if (!list_member_int(chunk_ids, osm_chunk_id))
```
jnidzwetzki, line 717: Nit: `list_append_unique_int` could be an alternative

jnidzwetzki (approved):

src/hypertable_restrict_info.c
zilder: Oh, I missed that function. Thanks for the review!

PR #5002, opened 2022-11-22
Fix dist_fetcher_type test on PG15

sb230132:
Fix dist_fetcher_type test on PG15

### DIFF
--- a/tsl/test/shared/expected/dist_fetcher_type-15.out
+++ b/tsl/test/shared/expected/dist_fetcher_type-15.out
@@ -1,47 +1,88 @@
 -- This file and its contents are licensed under the Timescale License.
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-TIMESCALE for a copy of the license.
-
 \set ON_ERROR_STOP off
-
 -- Test that we use the correct type of remote data fetcher.
 set timescaledb.remote_data_fetcher = 'auto';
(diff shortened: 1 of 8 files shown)


lkshminarayanan (approved):

tsl/test/shared/expected/dist_fetcher_type-15.out
```diff
@@ -0,0 +1,271 @@
...
+select 1 x from distinct_on_distributed t1, distinct_on_distributed t2
+where t1.id = t2.id + 1
+limit 1;
+QUERY PLAN
+ Limit (actual rows=1 loops=1)
+   Output: 1
+   ->  Nested Loop (actual rows=1 loops=1)
+         Output: 1
+         Join Filter: (t1.id = (t2.id + 1))
+         Rows Removed by Join Filter: 40000
```
akuzm, line 54: So do we know why the plan has changed? Looks like the LIMIT stopped working for the distributed tables, this is probably a bug.

akuzm (approved):
Just for context to other readers: we concluded that the reason for reading more rows is the changes of ordering of rows in PG 15. Because of this, the matching rows arrive later than in PG 14, and the join has to filter out more rows before there is a match. To fix, we changed the condition to just equality on id. This way, the first row should match.

PR #5004, opened 2022-11-22
Use custom node to block frozen chunk modifications

svenklemm:
This patch changes the code that blocks frozen chunk modifications to no longer use triggers but to use custom node instead. Frozen chunks is a timescaledb internal object and should therefore not be protected by TRIGGER which is external and creates several hazards. TRIGGERs created to protect internal state contend with user-created triggers. The trigger created to protect frozen chunks does not work well with our restoring GUC which we use when restoring logical dumps. Thirdly triggers are not functional for any internal operations but are only working in code pathes that explicitly added trigger support.

### DIFF
--- a/src/nodes/chunk_dispatch.c
+++ b/src/nodes/chunk_dispatch.c
@@ -148,6 +148,17 @@ ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
 		 */
 		bool found;
 		Chunk *new_chunk = ts_hypertable_find_chunk_for_point(dispatch->hypertable, point);
+
+#if PG14_GE
+		/*
+		 * Frozen chunks require at least PG14.
+		 */
+		if (new_chunk && ts_chunk_is_frozen(new_chunk))
+			elog(ERROR,
+				 "cannot INSERT into frozen chunk \"%s\"",
+				 get_rel_name(new_chunk->table_id));
+#endif
+
 		if (new_chunk == NULL)
 		{
 			new_chunk = ts_hypertable_create_chunk_for_point(dispatch->hypertable, point, &found);
--- a/tsl/src/planner.c
+++ b/tsl/src/planner.c
@@ -5,6 +5,7 @@
  */
 #include <postgres.h>
 #include <catalog/pg_trigger.h>
+#include <commands/extension.h>
 #include <optimizer/paths.h>
 #include <parser/parsetree.h>
 #include <foreign/fdwapi.h>
@@ -24,6 +25,7 @@
 #include "ts_catalog/hypertable_compression.h"
 #include "hypertable.h"
 #include "nodes/compress_dml/compress_dml.h"
+#include "nodes/frozen_chunk_dml/frozen_chunk_dml.h"
 #include "nodes/decompress_chunk/decompress_chunk.h"
 #include "nodes/data_node_dispatch.h"
 #include "nodes/data_node_copy.h"
@@ -31,6 +33,8 @@
 #include "planner.h"
 
 #include <math.h>
+
+#define OSM_EXTENSION_NAME "timescaledb_osm"
 
 static bool
 is_dist_hypertable_involved(PlannerInfo *root)
@@ -48,6 +52,21 @@ is_dist_hypertable_involved(PlannerInfo *root)
 
 	return false;
 }
+
+#if PG14_GE
+static int osm_present = -1;
+
+static bool
+is_osm_present()
+{
+	if (osm_present == -1)
+	{
+		Oid osm_oid = get_extension_oid(OSM_EXTENSION_NAME, true);
+		osm_present = OidIsValid(osm_oid);
+	}
+	return osm_present;
+}
+#endif
 
 void
 tsl_create_upper_paths_hook(PlannerInfo *root, UpperRelationKind stage, RelOptInfo *input_rel,
@@ -114,10 +133,28 @@ tsl_set_rel_pathlist_query(PlannerInfo *root, RelOptInfo *rel, Index rti, RangeT
 			ts_decompress_chunk_generate_paths(root, rel, ht, chunk);
 	}
 }
+
 void
 tsl_set_rel_pathlist_dml(PlannerInfo *root, RelOptInfo *rel, Index rti, RangeTblEntry *rte,
 						 Hypertable *ht)
 {
+#if PG14_GE
+	if (is_osm_present())
+	{
+		Chunk *chunk = ts_chunk_get_by_relid(rte->relid, false);
+		if (chunk && ts_chunk_is_frozen(chunk))
+		{
+			ListCell *lc;
+			foreach (lc, rel->pathlist)
+			{
+				Path **pathptr = (Path **) &lfirst(lc);
+				*pathptr = frozen_chunk_dml_generate_path(*pathptr, chunk);
+			}
+			return;
+		}
+	}
+#endif
+
 	if (ht != NULL && TS_HYPERTABLE_HAS_COMPRESSION_TABLE(ht))
 	{
 		ListCell *lc;
--- a/tsl/test/sql/chunk_utils_internal.sql
+++ b/tsl/test/sql/chunk_utils_internal.sql
@@ -27,11 +27,14 @@ CREATE OR REPLACE VIEW chunk_view AS
 GRANT SELECT on chunk_view TO PUBLIC;
 
 \c :TEST_DBNAME :ROLE_SUPERUSER
+-- fake presence of timescaledb_osm
+INSERT INTO pg_extension(oid,extname,extowner,extnamespace,extrelocatable,extversion) SELECT 1,'timescaledb_osm',10,11,false,'1.0';
+
 CREATE SCHEMA test1;
 GRANT CREATE ON SCHEMA test1 TO :ROLE_DEFAULT_PERM_USER;
 GRANT USAGE ON SCHEMA test1 TO :ROLE_DEFAULT_PERM_USER;
 
---mock hooks for OSM intercation with timescaledb
+-- mock hooks for OSM interaction with timescaledb
 CREATE OR REPLACE FUNCTION ts_setup_osm_hook( ) RETURNS VOID
 AS :TSL_MODULE_PATHNAME LANGUAGE C VOLATILE;
 
@@ -61,10 +64,8 @@ WHERE hypertable_name = 'hyper1' and hypertable_schema = 'test1'
 ORDER BY chunk_name LIMIT 1
 \gset
 
--- Freeze and check trigger
-SELECT tgname, tgtype FROM pg_trigger WHERE tgrelid = :'CHNAME'::regclass ORDER BY tgname, tgtype;
+-- Freeze
 SELECT  _timescaledb_internal.freeze_chunk( :'CHNAME');
-SELECT tgname, tgtype FROM pg_trigger WHERE tgrelid = :'CHNAME'::regclass ORDER BY tgname, tgtype;
 
 SELECT * from test1.hyper1 ORDER BY 1;
 
@@ -121,7 +122,7 @@ ROLLBACK;
 INSERT INTO test1.hyper1 VALUES ( 11, 11);
 
 -- Test truncating table should fail
-TRUNCATE test1.hyper1;
+TRUNCATE :CHNAME;
 
 SELECT * from test1.hyper1 ORDER BY 1;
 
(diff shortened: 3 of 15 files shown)


konskov:
> only working in code pathes
Nit: typo in the commit message

src/nodes/chunk_dispatch.c
```diff
@@ -145,6 +145,15 @@ ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
 		 * where the chunk already exists.
 		 */
 		Chunk *new_chunk = ts_hypertable_find_chunk_for_point(dispatch->hypertable, point);
+
+#if PG14_GE
+		/*
+		 * Frozen chunks require at least PG14.
+		 */
+		if (new_chunk && ts_chunk_is_frozen(new_chunk))
+			elog(ERROR, "cannot INSERT into frozen chunk");
```
konskov, line 154: I think it would be good to print the chunk relname here also

tsl/test/sql/chunk_utils_internal.sql
```diff
@@ -517,10 +515,6 @@ SELECT _timescaledb_internal.stop_background_workers();
```
jnidzwetzki, line 515: The tests in Lines 497 - 516 (`-- Test dump & restore`) can also be removed. They were introduced to test backup and restore when the block trigger is active. In addition, Lines 133 - 140 in `tsl/test/sql/CMakeLists.txt`can be removed.

src/nodes/chunk_dispatch.c
svenklemm: done

jnidzwetzki:
I did a few benchmarks. It seems that the current version of the PR re-introduces the performance regression that was solved by 380464df9bb35784853f186b9e268cd9d50e442b. From my understanding, we now need to [read the status of the chunk](https://github.com/timescale/timescaledb/pull/5004/files#diff-138c70c5e0855b42b17a9b0f2f095dbda5f2f7489dab6f0fd4de4ef41328b003R124) to decide if we should construct the custom node or not. This seems to be done at query planning time and it is performed for all chunks that might be affected by the query. 

Reading the chunk state is a costly meta-data lookup. Path pruning is not always possible during query planning (e.g., due to a cast in the query predicate). Therefore, this lookup has to be performed sometimes for all chunks and not only for the chunks that are actually affected by the query (this is my assumption of the regression, which should be validated with a Flamegraph). 

Might it be possible to delay the check until the query actually accesses the chunk (e.g., construct the custom node always, perform the meta-data lookup on the first call, and cache the result for all further calls)? This might help to move this check entirely into the executor code path and only perform the meta-data lookup when the query actually accesses a chunk.

# Benchmark
The benchmarks are performed on PostgreSQL 14.2 (release build).

## Last commit before the changes are introduced (40297f1897ab73d5b9b7d960b5ae83338f14e1bf)

```
Report for benchmark suite 'delete_exclusion_space'
+---------------------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                             | 40297f1897ab73d5b9b7d960b5ae83338f14e1bf |
+---------------------------------------------------------------------------------------------------+------------------------------------------+
| DELETE FROM ht_chunk_space_1k WHERE time > '2000-01-01';                                          |                                   112.75 |
| DELETE FROM ht_chunk_space_1k WHERE time > now();                                                 |                                   117.14 |
| DELETE FROM ht_chunk_space_1k WHERE time > '2000-01-01'::text::timestamptz;                       |                                   221.07 |
| DELETE FROM ht_chunk_space_1k WHERE device = 1 AND time > '2000-01-01';                           |                                   126.43 |
| DELETE FROM ht_chunk_space_1k WHERE device = 1 AND time > now();                                  |                                   129.23 |
| DELETE FROM ht_chunk_space_1k WHERE device = 1 AND time > '2000-01-01'::text::timestamptz;        |                                   162.00 |
| DELETE FROM ht_chunk_space_1k WHERE device IN (2,3,4) AND time > '2000-01-01';                    |                                   142.34 |
| DELETE FROM ht_chunk_space_1k WHERE device IN (2,3,4) AND time > now();                           |                                   145.48 |
| DELETE FROM ht_chunk_space_1k WHERE device IN (2,3,4) AND time > '2000-01-01'::text::timestamptz; |                                   226.06 |
| DELETE FROM ht_chunk_space_4k WHERE time > '2000-01-01';                                          |                                   509.54 |
| DELETE FROM ht_chunk_space_4k WHERE time > now();                                                 |                                   525.03 |
| DELETE FROM ht_chunk_space_4k WHERE time > '2000-01-01'::text::timestamptz;                       |                                   994.08 |
| DELETE FROM ht_chunk_space_4k WHERE device = 1 AND time > '2000-01-01';                           |                                   553.00 |
| DELETE FROM ht_chunk_space_4k WHERE device = 1 AND time > now();                                  |                                   571.97 |
| DELETE FROM ht_chunk_space_4k WHERE device = 1 AND time > '2000-01-01'::text::timestamptz;        |                                   705.55 |
| DELETE FROM ht_chunk_space_4k WHERE device IN (2,3,4) AND time > '2000-01-01';                    |                                   630.39 |
| DELETE FROM ht_chunk_space_4k WHERE device IN (2,3,4) AND time > now();                           |                                   649.52 |
| DELETE FROM ht_chunk_space_4k WHERE device IN (2,3,4) AND time > '2000-01-01'::text::timestamptz; |                                   975.96 |
| DELETE FROM ht_metrics_space WHERE device = 1 AND time > '2020-01-21';                            |                                  2656.20 |
| DELETE FROM ht_metrics_space WHERE device IN (2,2) AND time > '2020-01-21';                       |                                  2822.64 |
| DELETE FROM ht_metrics_space WHERE device = ANY(ARRAY[3,3]) AND time > '2020-01-21';              |                                  2330.39 |
+---------------------------------------------------------------------------------------------------+------------------------------------------+
Report for benchmark suite 'update_exclusion'
+------------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                    | 40297f1897ab73d5b9b7d960b5ae83338f14e1bf |
+------------------------------------------------------------------------------------------+------------------------------------------+
| UPDATE ht_chunk_1k SET value = value + 0.5 WHERE time > '2000-01-01';                    |                                    76.28 |
| UPDATE ht_chunk_1k SET value = value + 0.5 WHERE time > now();                           |                                    81.27 |
| UPDATE ht_chunk_1k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz; |                                   162.40 |
| UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01';                    |                                   707.87 |
| UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > now();                           |                                   748.07 |
| UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz; |                                  1457.20 |
+------------------------------------------------------------------------------------------+------------------------------------------+
```

## Commit that contains the changes (76d1d3a02962c90c289dca7f75f9ed0478be0acb)
```
Report for benchmark suite 'delete_exclusion_space'
+---------------------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                             | 76d1d3a02962c90c289dca7f75f9ed0478be0acb |
+---------------------------------------------------------------------------------------------------+------------------------------------------+
| DELETE FROM ht_chunk_space_1k WHERE time > '2000-01-01';                                          |                                   111.69 |
| DELETE FROM ht_chunk_space_1k WHERE time > now();                                                 |                                   120.25 |
| DELETE FROM ht_chunk_space_1k WHERE time > '2000-01-01'::text::timestamptz;                       |                                   297.24 |
| DELETE FROM ht_chunk_space_1k WHERE device = 1 AND time > '2000-01-01';                           |                                   124.99 |
| DELETE FROM ht_chunk_space_1k WHERE device = 1 AND time > now();                                  |                                   128.34 |
| DELETE FROM ht_chunk_space_1k WHERE device = 1 AND time > '2000-01-01'::text::timestamptz;        |                                   181.38 |
| DELETE FROM ht_chunk_space_1k WHERE device IN (2,3,4) AND time > '2000-01-01';                    |                                   140.89 |
| DELETE FROM ht_chunk_space_1k WHERE device IN (2,3,4) AND time > now();                           |                                   143.96 |
| DELETE FROM ht_chunk_space_1k WHERE device IN (2,3,4) AND time > '2000-01-01'::text::timestamptz; |                                   265.58 |
| DELETE FROM ht_chunk_space_4k WHERE time > '2000-01-01';                                          |                                   496.82 |
| DELETE FROM ht_chunk_space_4k WHERE time > now();                                                 |                                   518.90 |
| DELETE FROM ht_chunk_space_4k WHERE time > '2000-01-01'::text::timestamptz;                       |                                  1298.56 |
| DELETE FROM ht_chunk_space_4k WHERE device = 1 AND time > '2000-01-01';                           |                                   548.51 |
| DELETE FROM ht_chunk_space_4k WHERE device = 1 AND time > now();                                  |                                   566.24 |
| DELETE FROM ht_chunk_space_4k WHERE device = 1 AND time > '2000-01-01'::text::timestamptz;        |                                   781.48 |
| DELETE FROM ht_chunk_space_4k WHERE device IN (2,3,4) AND time > '2000-01-01';                    |                                   625.78 |
| DELETE FROM ht_chunk_space_4k WHERE device IN (2,3,4) AND time > now();                           |                                   637.41 |
| DELETE FROM ht_chunk_space_4k WHERE device IN (2,3,4) AND time > '2000-01-01'::text::timestamptz; |                                  1110.49 |
| DELETE FROM ht_metrics_space WHERE device = 1 AND time > '2020-01-21';                            |                                  2601.20 |
| DELETE FROM ht_metrics_space WHERE device IN (2,2) AND time > '2020-01-21';                       |                                  2801.96 |
| DELETE FROM ht_metrics_space WHERE device = ANY(ARRAY[3,3]) AND time > '2020-01-21';              |                                  2330.24 |
+---------------------------------------------------------------------------------------------------+------------------------------------------+
Report for benchmark suite 'update_exclusion'
+------------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                    | 76d1d3a02962c90c289dca7f75f9ed0478be0acb |
+------------------------------------------------------------------------------------------+------------------------------------------+
| UPDATE ht_chunk_1k SET value = value + 0.5 WHERE time > '2000-01-01';                    |                                    75.94 |
| UPDATE ht_chunk_1k SET value = value + 0.5 WHERE time > now();                           |                                    81.21 |
| UPDATE ht_chunk_1k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz; |                                   235.34 |
| UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01';                    |                                   701.54 |
| UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > now();                           |                                   737.11 |
| UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz; |                                  2038.73 |
+------------------------------------------------------------------------------------------+------------------------------------------+
```

## Summary

Affected are queries like `UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz;` (old 1457.20 ms, new 2038.73 ms) or `DELETE FROM ht_chunk_space_4k WHERE device IN (2,3,4) AND time > '2000-01-01'::text::timestamptz;` (old 975.96 ms, new 1110.49 ms).

akuzm:
> This might help to move this check entirely into the executor code path and only perform the meta-data lookup when the query actually accesses a chunk.

But this will have the same performance problem, right? It is also a per-chunk lookup of this metadata. We could speed it up in two ways:
1) batch it (do for all chunks in the query at the same time)
2) make it faster by adding some kind of index
I think we need to do both. The ideal place would be during planning at some point where the chunks have been expanded and we have access to the parent hypertable.

jnidzwetzki:
@akuzm In the table `ht_chunk_8k`, we have 8001 chunks. 8000 chunks that are not affected by the query `UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz;` and only one chunk that is affected by the query. My assumption is that when we perform the meta-data check only on demand, we will perform the meta-data lookup only for the one affected chunk.

However, I absolutely agree with you. Anything that makes meta-data lookup operations faster will help us.

akuzm:
> @akuzm In the table `ht_chunk_8k`, we have 8001 chunks. 8000 chunks that are not affected by the query `UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz;` and only one chunk that is affected by the query. My assumption is that when we perform the meta-data check only on demand, we will perform the meta-data lookup only for the one affected chunk.

I forgot, do we have runtime chunk exclusion for update? If so, maybe this lookup should be just after we do the runtime exclusion.

svenklemm:
> > @akuzm In the table `ht_chunk_8k`, we have 8001 chunks. 8000 chunks that are not affected by the query `UPDATE ht_chunk_8k SET value = value + 0.5 WHERE time > '2000-01-01'::text::timestamptz;` and only one chunk that is affected by the query. My assumption is that when we perform the meta-data check only on demand, we will perform the meta-data lookup only for the one affected chunk.
> 
> I forgot, do we have runtime chunk exclusion for update? If so, maybe this lookup should be just after we do the runtime exclusion.

Yes we do but benefitting from this here is a bit more involved and runtime exclusion uses postgres constrait exclusion instead of internal metadata.

svenklemm:
Latest version of the PR does not have the regression:
https://grafana.ops.savannah-dev.timescale.com/d/NdmLnOk4z/compare-benchmark-runs?orgId=1&var-run1=1639&var-run2=1641&var-postgres=14&var-branch=All&var-threshold=0.1

tsl/test/sql/chunk_utils_internal.sql
svenklemm: done

jnidzwetzki:
If I understand the change correctly, it now checks whether the [OSM extension is loaded or not](https://github.com/timescale/timescaledb/pull/5004/files#diff-138c70c5e0855b42b17a9b0f2f095dbda5f2f7489dab6f0fd4de4ef41328b003R140). If it is not loaded, the freeze check is omitted, which removes the regression. 

Did you change anything else to fix the regression, or does the regression occur when the OSM extension is loaded?

svenklemm:
The real fix is https://github.com/timescale/eng-database/issues/339 which will also fix the planning time for compression/any chunk metadata lookup but it's not a quick fix.

konskov (approved):

fabriziomello (approved):

tsl/src/planner.c
```diff
@@ -49,6 +51,21 @@ is_dist_hypertable_involved(PlannerInfo *root)
 	return false;
 }
 
+#if PG14_GE
+static int osm_present = -1;
```
fabriziomello, line 55:
```suggestion
static bool osm_present = false;
```
Nit. Shouldn't this be `bool` ??

jnidzwetzki (approved):

tsl/src/planner.c
svenklemm: this is tri-state, since we want to cache the result of  the check for this backend, -1 = not yet checked

PR #5007, opened 2022-11-22
Change time of scheduled CI run

svenklemm:
Since we now use the date as a part of the cache key to ensure no stale cache entries hiding build failures we need to make sure we have a cache entry present before workflows that depend on cache are run.

### DIFF
--- a/.github/workflows/linux-build-and-test.yaml
+++ b/.github/workflows/linux-build-and-test.yaml
@@ -1,8 +1,12 @@
 name: Regression
 on:
   schedule:
-    # run daily 20:00 on main branch
-    - cron: '0 20 * * *'
+    # run daily 0:00 on main branch
+    # Since we use the date as a part of the cache key to ensure no
+    # stale cache entries hiding build failures we need to make sure
+    # we have a cache entry present before workflows that depend on cache
+    # are run.
+    - cron: '0 0 * * *'
   push:
     branches:
       - prerelease_test


akuzm (approved):

akuzm:
Does it have some way to specify dependencies between workflows? If we move to per-commit runs, we'll probably need it.

svenklemm:
> Does it have some way to specify dependencies between workflows? If we move to per-commit runs, we'll probably need it.

AFAIK you can only have dependencies with workflows. But this only affects coverity and sqlsmith so triggering a run at midnight is acceptable imo.

fabriziomello (approved):

konskov (approved):

akuzm:
> AFAIK you can only have dependencies with workflows. But this only affects coverity and sqlsmith so triggering a run at midnight is acceptable imo.

By the way, they are still broken:

1) We didn't update the cache key for these workflows. It now includes the day and the workflow file hash
2) They use ubuntu 20.04 and our builds have been using ubuntu 22.04 for a long time already. 

I guess the reason they still work is that they use the old cached builds.

akuzm:
I'll try to fix it here: https://github.com/timescale/timescaledb/pull/4986

svenklemm:
yea i noticed that we didnt adjust the cache key but i changed them now to no longer depend on the cache

PR #5008, opened 2022-11-22
Fix TRUNCATE on hierarchical caggs

konskov:
When truncating a cagg that had another cagg defined on top of it, the nested cagg would not get invalidated accordingly. That was because we were not adding a corresponding entry in the hypertable invalidation log for the materialization hypertable of the base cagg.
This commit adds an invalidation entry in the table so that subsequent refreshes see and properly process this invalidation.

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -1082,6 +1082,17 @@ process_truncate(ProcessUtilityArgs *args)
 																		  TS_TIME_NOBEGIN,
 																		  TS_TIME_NOEND);
 
+						/* Additionally, this cagg's materialization hypertable could be the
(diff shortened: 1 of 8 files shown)


svenklemm (approved):

mkindahl (commented):
Patch looks fine, but:

- It looks like you should squash the commits before merging them. In that case, adding "disable-check: commit-count" will risk merging commits before they are squashed (been there done that).
- You also have a typo in the message, so please check the final commit message.

konskov:
thank you, I `wouild` not have noticed the typo otherwise 😄 . Done!

fabriziomello (approved):

mkindahl (approved):

PR #5009, opened 2022-11-22
Enable few CI checks for PG15.

sb230132:
This patch will enable Regression Linux i386 / PG15.1 Debug linux-i386 workflow.


.github/workflows/update-test.yaml
```diff
@@ -35,6 +35,8 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Update tests ${{ matrix.pg }}
+      # skip update test for PG15 and above
+      if: matrix.pg < '15'
```
jnidzwetzki, line 39: Is the test disabled because there are no container images for PG 15 yet or are there other reasons?

jnidzwetzki (approved):
Overall, it looks good to me. I would recommend making the title of the commit more precise. With this commit, only some (and not all) PG 15 CI checks are enabled.

.github/ci_settings.py
```diff
@@ -22,4 +22,6 @@
 PG14_LATEST = "14.6"
 PG14_ABI_MIN = "14.0"
 
-PG_LATEST = [PG12_LATEST, PG13_LATEST, PG14_LATEST]
+PG15_LATEST = "15.1"
```
jnidzwetzki, line 25: Maybe it would be good to introduce the `PG15_EARLIEST` setting with this commit as well. Or are there reasons that prevent this?

.github/workflows/update-test.yaml
sb230132: Yes, we skip because there is no container images for PG 15.

fabriziomello (approved):

sb230132:
Closing this since its fixed as part of https://github.com/timescale/timescaledb/pull/5042

PR #5020, opened 2022-11-24
Test dist caggs with an unavailable data node

### LINKED ISSUE #4978: Multinode - Distributed Continous aggs test when a DN is down
Distributed Continous Agg saves data on the AN. Caggs should work for a refresh when RF>1

pmwkaa:
Add additional test cases to ensure caggs functionality on distributed hypertable during data node being unavailable.

Fix #4978

### DIFF
--- a/tsl/test/expected/dist_cagg.out
+++ b/tsl/test/expected/dist_cagg.out
@@ -1,0 +1,146 @@
+-- This file and its contents are licensed under the Timescale License.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-TIMESCALE for a copy of the license.
+-- Need to be super user to create extension and add data nodes
+\c :TEST_DBNAME :ROLE_CLUSTER_SUPERUSER;
+-- Support for execute_sql_and_filter_server_name_on_error()
+\unset ECHO
+psql:include/remote_exec.sql:5: NOTICE:  schema "test" already exists, skipping
+psql:include/filter_exec.sql:5: NOTICE:  schema "test" already exists, skipping
+\set DATA_NODE_1 :TEST_DBNAME _1
+\set DATA_NODE_2 :TEST_DBNAME _2
+\set DATA_NODE_3 :TEST_DBNAME _3
+CREATE SCHEMA some_schema AUTHORIZATION :ROLE_1;
+SELECT node_name, database, node_created, database_created, extension_created
+FROM (
+  SELECT (add_data_node(name, host => 'localhost', DATABASE => name)).*
+  FROM (VALUES (:'DATA_NODE_1'), (:'DATA_NODE_2'), (:'DATA_NODE_3')) v(name)
+) a;
+   node_name    |    database    | node_created | database_created | extension_created 
+----------------+----------------+--------------+------------------+-------------------
+ db_dist_cagg_1 | db_dist_cagg_1 | t            | t                | t
+ db_dist_cagg_2 | db_dist_cagg_2 | t            | t                | t
+ db_dist_cagg_3 | db_dist_cagg_3 | t            | t                | t
+(3 rows)
+
+GRANT USAGE ON FOREIGN SERVER :DATA_NODE_1, :DATA_NODE_2, :DATA_NODE_3 TO PUBLIC;
+GRANT CREATE ON SCHEMA public TO :ROLE_1;
+-- test alter_data_node(unvailable) with caggs
+--
+-- create cagg on distributed hypertable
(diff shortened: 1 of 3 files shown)


tsl/test/expected/dist_ddl.out
```diff
@@ -3384,6 +3384,100 @@ SELECT * from show_chunks('dist_test');
...
+-- case 2: ensure cagg can be updated while data node is not available
+INSERT INTO conditions_dist(day, temperature)
+SELECT ts, date_part('month', ts)*100 + date_part('day', ts)
+FROM generate_series('2010-03-01 00:00:00 MSK' :: timestamptz, '2010-04-01 00:00:00 MSK' :: timestamptz - interval '1 day', '1 day') as ts;
+-- case 3: ensure refresh cagg does not work on a data node
+\set ON_ERROR_STOP 0
+CALL refresh_continuous_aggregate('conditions_dist_1m', '2010-01-01', '2010-04-01');
+ERROR:  could not connect to "db_dist_ddl_1"
+\set ON_ERROR_STOP 1
+-- case 4: ensure that cagg is still updated and data is visible using real-time aggregation
```
erimatnor, line 3445:
```suggestion
-- case 4: ensure that new data is visible using real-time aggregation
```

tsl/test/expected/dist_ddl.out
```diff
@@ -3384,6 +3384,100 @@ SELECT * from show_chunks('dist_test');
...
+SELECT * FROM conditions_dist_1m ORDER BY bucket;
+            bucket            | min | max  
+------------------------------+-----+------
+ Thu Dec 31 13:00:00 2009 PST | 101 | 1231
+ Sun Jan 31 13:00:00 2010 PST | 131 |  227
+ Sun Feb 28 13:00:00 2010 PST | 228 |  330
+(3 rows)
+
+-- case 5: TRUNCATE will not work while data node is not available
```
erimatnor, line 3454:
```suggestion
-- case 5: TRUNCATE will not work while data node is not available because invalidations are sent to the data nodes
```

tsl/test/expected/dist_ddl.out
```diff
@@ -3384,6 +3384,100 @@ SELECT * from show_chunks('dist_test');
...
+(1 row)
+
+SELECT alter_data_node(:'DATA_NODE_1', available => true);
+                 alter_data_node                 
+-------------------------------------------------
+ (db_dist_ddl_1,localhost,55432,db_dist_ddl_1,t)
+(1 row)
+
+CALL refresh_continuous_aggregate('conditions_dist_1m', '2010-01-01', '2010-04-01');
+SELECT * FROM conditions_dist_1m ORDER BY bucket;
```
erimatnor, line 3478: Would be nice to turn off real-time aggregation before doing this query so that we can see that the new value is actually materialized.

tsl/test/expected/dist_ddl.out
```diff
@@ -3384,6 +3384,105 @@ SELECT * from show_chunks('dist_test');
...
+------------------------------+-----+------
+ Thu Dec 31 13:00:00 2009 PST | 101 | 1231
+ Sun Jan 31 13:00:00 2010 PST | 131 |  227
+ Sun Feb 28 13:00:00 2010 PST | 228 |  330
+(3 rows)
+
+-- case 5: TRUNCATE will not work while data node is not available
+\set ON_ERROR_STOP 0
+TRUNCATE TABLE conditions_dist;
+ERROR:  could not connect to "db_dist_ddl_1"
```
erimatnor, line 3457: Fails because DDL doesn't work when DN unavailable

tsl/test/expected/dist_ddl.out
```diff
@@ -3384,6 +3384,105 @@ SELECT * from show_chunks('dist_test');
...
+\set ON_ERROR_STOP 1
+-- case 6: TRUNCATE cagg directly
+\set ON_ERROR_STOP 0
+TRUNCATE conditions_dist_1m;
```
erimatnor, line 3461: Fails because it is trying to send invalidation to DN. Note: data in cagg is stored on AN, so the TRUNCATION itself doesn't fail, but it fails as a result of the failed invalidation.

erimatnor (approved):
Covers the test cases we talked about but some of the comments are a bit misleading so you might want to tweak those according to suggestions.

Also, I think the test probably belongs in a different test file.

nikkhils (approved):
LGTM as well. Please fix Erik's review comments before merging.

PR #5021, opened 2022-11-25
Use packaged postgres for sqlsmith and coverity CI

svenklemm:
The sqlsmith and coverity workflows used the cache postgres build but could not produce a build by themselves and therefore relied on other workflows to produce the cached binaries. This patch changes those workflows to use normal postgres packages instead of custom built postgres to remove that dependency.

### DIFF
--- a/.github/workflows/coverity.yaml
+++ b/.github/workflows/coverity.yaml
@@ -7,53 +7,23 @@ on:
     branches:
       - coverity_scan
 jobs:
-  config:
-    runs-on: ubuntu-latest
-    outputs:
-      pg_latest: ${{ steps.setter.outputs.PG_LATEST }}
-    steps:
-    - name: Checkout source code
-      uses: actions/checkout@v3
-    - name: Read configuration
-      id: setter
-      run: python .github/gh_config_reader.py
 
   coverity:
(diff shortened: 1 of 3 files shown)


.github/workflows/coverity.yaml
```diff
@@ -25,35 +15,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pg: ${{ fromJson(needs.config.outputs.pg_latest) }}
+        pg: [12,13,14]
         os: [ubuntu-20.04]
```
akuzm, line 19: Should we switch to 22.04 while we're at it?

akuzm (approved):

.github/workflows/coverity.yaml
```diff
@@ -25,35 +15,20 @@ jobs:
...
     env:
       PG_SRC_DIR: pgbuild
```
SachinSetiya, line 21: I dont think these 3 variables are used anymore ?

SachinSetiya (approved):

.github/workflows/coverity.yaml
svenklemm: removed

PR #5028, opened 2022-11-26
Use RelationGetSmgr instead of rd_smgr

svenklemm:
rd_smgr should not be accessed directly but RelationGetSmgr should be used instead. Accessing it directly can lead to segfaults when parallel relcache flushes are happening.

https://github.com/postgres/postgres/commit/f10f0ae420ee62400876ab34dca2c09c20dcd030

### DIFF
--- a/src/copy.c
+++ b/src/copy.c
@@ -703,6 +703,31 @@ has_other_before_insert_row_trigger_than_ts(ResultRelInfo *resultRelInfo)
 	return false;
 }
 
+#if PG13_GE
+/*
+ * RelationGetSmgr
+ *		Returns smgr file handle for a relation, opening it if needed.
+ *
+ * Very little code is authorized to touch rel->rd_smgr directly.  Instead
+ * use this function to fetch its value.
+ *
+ * Note: since a relcache flush can cause the file handle to be closed again,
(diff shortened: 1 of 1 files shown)


src/copy.c
```diff
@@ -703,6 +703,31 @@ has_other_before_insert_row_trigger_than_ts(ResultRelInfo *resultRelInfo)
...
+ * copied verbatim from postgres because it is a static function
+ */
+static inline SMgrRelation
+RelationGetSmgr(Relation rel)
+{
+	if (unlikely(rel->rd_smgr == NULL))
+		smgrsetowner(&(rel->rd_smgr), smgropen(rel->rd_node, rel->rd_backend));
+	return rel->rd_smgr;
+}
+#endif
```
fabriziomello, line 729: Shouldn't this code reside in the compat module?

fabriziomello (approved):
Left a minor comment but approved anyway.

src/copy.c
svenklemm: We use compat for wrapper functions that abstract API differences between the PG versions. For verbatim imports from PG source we use src/import but since copy.c was the only call site i put it there directly

konskov (approved):

PR #5031, opened 2022-11-26
Fix segfault in cagg creation

svenklemm:
When trying to create cagg on top of any relation that is a neither a hypertable nor a continuous aggregate the command would segfault. This patch changes the code to handle this case gracefully and error out when trying to create a cagg on top of a relation that is not supported. Found by coverity.

### DIFF
--- a/tsl/src/continuous_aggs/create.c
+++ b/tsl/src/continuous_aggs/create.c
@@ -1105,7 +1105,7 @@ cagg_validate_query(const Query *query, const bool finalized, const char *cagg_s
 	StringInfo detail = makeStringInfo();
 	bool is_nested = false;
 	Query *prev_query = NULL;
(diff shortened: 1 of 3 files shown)


fabriziomello (approved):

tsl/src/continuous_aggs/create.c
```diff
@@ -1155,6 +1155,15 @@ cagg_validate_query(const Query *query, const bool finalized, const char *cagg_s
 		{
 			cagg_source = ts_continuous_agg_find_by_relid(rte->relid);
 
+			if (!cagg_source)
```
fabriziomello, line 1158: Can u rename this var to "cagg_parent"? Is a better name and I forgot to rename it in the original PR. I'm using "parent" in several places already!

konskov (approved):

PR #5035, opened 2022-11-28
Improve cost model for data node scans

akuzm:
Simplify the path generation for the parameterized data node scans. Adjust the data node scan cost if it's an index scan, instead of always treating it as a sequential scan. Add the GUC to disable parameterized data node scan. Add more tests.

Merge after:
- [x] https://github.com/timescale/timescaledb/pull/5044


erimatnor:
@akuzm can you describe what "saner values" means in this context, i.e., how did you conclude that the new values are saner?

akuzm:
> @akuzm can you describe what "saner values" means in this context, i.e., how did you conclude that the new values are saner?

The network transfer should be more costly than the transfer to local parallel executors, right? Not sure what's the factor, I set it to ten. The postgres_fdw defaults were lower.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -297,33 +297,15 @@ ts_ec_member_matches_foreign(PlannerInfo *root, RelOptInfo *rel, EquivalenceClas
...
-static void
-add_data_node_scan_paths(PlannerInfo *root, RelOptInfo *baserel)
+/*
+ * Build parameterizations that are useful for performing joins with the given
+ * hypertable relation. We will use them to generate the parameterized data node
+ * scan paths. The code is mostly copied from postgres_fdw,
+ * postgresGetForeignPaths().
+ */
+static List *
+build_parameterizations(PlannerInfo *root, RelOptInfo *hyper_rel)
```
akuzm, line 307: This is now a separate function, because it makes no sense to call it for each data node, it does the same work.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -462,32 +444,107 @@ add_data_node_scan_paths(PlannerInfo *root, RelOptInfo *baserel)
 		}
 	}
 
+	return ppi_list;
+}
+
+static void
+add_data_node_scan_paths(PlannerInfo *root, RelOptInfo *data_node_rel, RelOptInfo *hyper_rel,
```
akuzm, line 451: This the second part of the now-splitted function. It is called per data node and generates scan path.

tsl/src/fdw/estimate.c
```diff
@@ -121,14 +121,26 @@ get_upper_rel_estimate(PlannerInfo *root, RelOptInfo *rel, CostEstimate *ce)
 		get_agg_clause_costs_compat(root, (Node *) fpinfo->grouped_tlist, aggsplit, &aggcosts);
 	}
 
-	/* Get number of grouping columns and possible number of groups */
+	/*
+	 * Get number of grouping columns and possible number of groups. We don't
+	 * have per-column ndistinct statistics on access node for distributed
+	 * tables, so in this case hardcode it as 1/10 of all rows to incentivize
+	 * grouping push down.
+	 */
```
akuzm, line 129: The previous estimate was just totally incorrect because of the missing statistics. This improves over it, because GROUP BY should generally produce less rows than the ungrouped relation.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -462,32 +444,107 @@ add_data_node_scan_paths(PlannerInfo *root, RelOptInfo *baserel)
...
+			}
+		}
+
+		/*
+		 * As a baseline, cost the data node scan as a seq scan.
+		 */
 		Cost startup_cost = 0;
 		Cost run_cost = 0;
-		double rows = baserel->tuples > 1 ? baserel->tuples : 123456;
+		double rows = data_node_rel->tuples > 1 ? data_node_rel->tuples : 123456;
```
svenklemm, line 526: hmm that is weird i think postgres uses a different convention for unknown tuple estimates any reason why we are not following that

akuzm: You mean what `table_block_relaion_estimate_size` does when there are no stats? It knows the number of pages, but we don't, since this is a distributed table. The particular number is arbitrary, I just chose something that would be easy to spot when looking at EXPLAIN.

svenklemm (approved):

tsl/test/expected/dist_hypertable-14.out
```diff
@@ -3796,15 +3795,15 @@ ANALYZE hyper_estimate;
...
- Append  (cost=100.00..607.58 rows=15 width=20)
-   ->  Foreign Scan on _dist_hyper_16_38_chunk  (cost=100.00..101.40 rows=4 width=20)
-   ->  Foreign Scan on _dist_hyper_16_39_chunk  (cost=100.00..101.20 rows=2 width=20)
-   ->  Foreign Scan on _dist_hyper_16_40_chunk  (cost=100.00..101.30 rows=3 width=20)
-   ->  Foreign Scan on _dist_hyper_16_41_chunk  (cost=100.00..101.20 rows=2 width=20)
-   ->  Foreign Scan on _dist_hyper_16_42_chunk  (cost=100.00..101.30 rows=3 width=20)
-   ->  Foreign Scan on _dist_hyper_16_43_chunk  (cost=100.00..101.10 rows=1 width=20)
+                                        QUERY PLAN                                        
+------------------------------------------------------------------------------------------
+ Append  (cost=10000.00..60021.38 rows=15 width=20)
```
nikkhils, line 3800: Why have the cost ranges shifted from 100 to 10000 ranges?

akuzm: That's because I significantly increased the default fdw tuple cost. It was unreasonably low, lower than even the parallel tuple cost. I think now the plans are better, we started to prefer aggregation pushdown in more cases.

erimatnor:
If I am not mistaken, this PR changes costing more generally than only for parameterized scans. Not saying there is something wrong with that if it is an improvement, but I think the PR description/title should reflect that.

akuzm:
> If I am not mistaken, this PR changes costing more generally than only for parameterized scans. Not saying there is something wrong with that if it is an improvement, but I think the PR description/title should reflect that.

I extended the description. The reason this PR has many different changes is that they're all dependent and don't give good plans in isolation, so I had to do them together.

tsl/src/fdw/estimate.c
erimatnor: We have n_distinct stats for chunks (the root table holds no data so no stats). 

After I run ANALYZE on a distributed hypertable I see this:

```
postgres=# select tablename, attname, n_distinct from pg_stats where tablename LIKE '_dist_hyper_1_16_chunk';
       tablename        | attname | n_distinct 
------------------------+---------+------------
 _dist_hyper_1_16_chunk | time    |         -1
 _dist_hyper_1_16_chunk | device  |          3
 _dist_hyper_1_16_chunk | temp    |         -1
(3 rows)

```
While it makes sense to estimate the n_distinct stats for hypertables when no stats exist, I think we should try to do better with the stats we have.
 
For row count stats we use the sum of all chunks in a data node scan to get the stats for that DN. For n_distinct we could do something similar, but for n_distinct we probably need to use something like the average of all chunks' n_distinct.

If we are calculating the n_distinct for the whole hypertable, we need to look at two cases: (1) the column is the space partitioning column (2) it is not the space partition column.

For case (1) (column is space partition), then we can just sum the n_distinct of all data nodes.

For case (2), we probably need to take the average of all chunks.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -462,32 +444,107 @@ add_data_node_scan_paths(PlannerInfo *root, RelOptInfo *baserel)
...
+		startup_cost += data_node_rel->reltarget->cost.startup;
 		startup_cost += fpinfo->remote_conds_cost.startup;
 		run_cost += fpinfo->remote_conds_cost.per_tuple * rows;
 		run_cost += cpu_tuple_cost * rows;
-		run_cost += seq_page_cost * baserel->pages;
+		run_cost += seq_page_cost * data_node_rel->pages;
 		rows *= remote_sel_sane;
 
+		/* Discount it if it's likely to be an index scan on the remote. */
+		run_cost *= scan_cost_discount_multiplier;
```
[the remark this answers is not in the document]
akuzm: Makes sense, I switched it to use a bool variable.

tsl/src/fdw/estimate.c
akuzm: Sounds reasonable. We would use `get_relation_stats_hook` to return the averaged statistics for a given data node scan into `examine_simple_variable`, and then the Postgres cost estimation functions would make use of it. This looks like a relatively big change though, do you think we can do it later? This PR is already oversized somewhat, I had to make many changes to get good plans. It's better than the status quo, plus it brings us closer to using the parameterized data node scans, which will up the baseline performance for many kinds of multinode joins.

akuzm: This particular place might be simpler, I was thinking about join selectivity estimate in `add_data_node_scan_paths`. I would prefer to implement your suggested approach in both of these places and wherever else it is needed, so that we have a more consistent cost model. I'm going to correct the comments and mention this possibility.

erimatnor (approved):
As discussed, let's look at cost models and tuple costs separately (I think the PR description needs an update to reflect). The query plan changes are a little hard to grasp by just looking at test output, so I guess I partly trust that this is an improvement in that regard. We will need to continue to evolve this in any case.

Approving, since there seems to be valuable changes beyond the above.

tsl/test/expected/dist_param.out
```diff
@@ -45,125 +45,250 @@ select count(*) from show_chunks('metric_dist');
...
-               Remote SQL: SELECT id, value FROM public.metric_dist WHERE _timescaledb_internal.chunks_in(public.metric_dist.*, ARRAY[52]) AND ((ts >= '2022-02-01 15:02:02-08'::timestamp with time zone)) AND ((ts <= '2022-02-02 15:02:02-08'::timestamp with time zone)) AND (($1::integer = id))
-(13 rows)
+               ->  Index Scan using metric_name_name on public.metric_name
+                     Output: metric_name.id, metric_name.name
+                     Index Cond: ((metric_name.name >= 'cpu'::text) AND (metric_name.name < 'cpv'::text))
+                     Filter: (metric_name.name ~~ 'cpu%'::text)
+               ->  Custom Scan (DataNodeScan) on public.metric_dist
+                     Output: metric_dist.id, metric_dist.value
+                     Data node: data_node_1
+                     Chunks: _dist_hyper_1_3_chunk, _dist_hyper_1_16_chunk, _dist_hyper_1_20_chunk, _dist_hyper_1_37_chunk, _dist_hyper_1_52_chunk
```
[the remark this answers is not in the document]
akuzm: Yes, I widened the condition to transfer more data, so that the cost changes are more apparent with plain join vs parameterized subqueries.

PR #5041, opened 2022-12-01
Fix remote_txn on PG15

### LINKED ISSUE #4837: Reporting WARNING after ERROR on PG15
### What type of bug is this?

Incorrect result

Other

### What happened?

TimescaleDB when compiled against PG15 causes many a test to report WARNING after an ERROR.
ERROR should be the last one to be reported.

2.9.0

### PostgreSQL version used

15.0

Mac OS X

Source

On prem/Self-hosted

_No response_

```bash
make installcheck TESTS="remote_txn"

Below is the diff

 COMMIT;
 WARNING:  kill event: pre-commit
-WARNING:  transaction rollback on data node "loopback" failed
 ERROR:  [loopback]: terminating connection due to administrator command
+WARNING:  transaction rollback on data node "loopback" failed
 SELECT debug_waitpoint_release('remote_conn_xact_end');
  debug_waitpoint_release
 -------------------------
@@ -27
```

sb230132:
In remote_txn, testcases which kill remote processes on data nodes,
tend to rollback transactions and as part of the process, WARNINGS/ERROR
are reported to client. Client however reports WARNINGS/ERROR in different
order intermittently. This is an issue specific to psql utility. Placing 
psql in gdb and trying to diagnose the problem does not reproduce the issue.

This patch fixes the tests by not reporting WARNINGS.

Fixes #4837

### DIFF
--- a/.github/gh_matrix_builder.py
+++ b/.github/gh_matrix_builder.py
@@ -155,9 +155,7 @@ m["include"].append(
             # below tests are tracked as part of #4972
             "installcheck_args": "SKIPS='dist_move_chunk' "
             # below tests are tracked as part of #4835
-            "IGNORES='telemetry_stats partialize_finalize "
-            # below tests are tracked as part of #4837
-            "remote_txn'",
+            "IGNORES='telemetry_stats partialize_finalize'",
         }
     )
 )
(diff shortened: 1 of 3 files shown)


fabriziomello (approved):

mkindahl (approved):
LGTM

PR #5043, opened 2022-12-01
Use fixed alpine version for upgrade/downgrade tests

SachinSetiya:
Updating alpine version causes libssl version to change, which in turn cause error in downgrade tests.

PR #5046, opened 2022-12-01
Fix ABI openssl issue

lkshminarayanan:
The latest Alpine Linux versions on which the recent PostgreSQL docker
images are based, updated their OpenSSL default to version 3. Due to
this there is a difference in the version of the OpenSSL libs available
between the builder and test images of the ABI tests causing them to
fail. Fixed that by updating the workflow to use the same version of the
OpenSSL lib available in both builder and test images.


lkshminarayanan:
The scheduled jobs have started failing  - https://github.com/timescale/timescaledb/actions/runs/3596121308 and this PR will fix them.

lkshminarayanan:
This is now fixed by #5043

PR #5049, opened 2022-12-02
Refactor CAggs on CAggs regression tests

fabriziomello:
When CAggs on CAggs was introduced in commit 3749953 the regression tests was splited into 6 (six) different test suites.

Simplyfied it grouping tests and reduced to just 2 (two) different test suites. It save resources and time because each suite test spawn it own Postgres instance.


akuzm (approved):
Does it give a big speedup? Personally I prefer small tests because the big output is hard to navigate when something breaks.

sb230132 (approved):

PR #5054, opened 2022-12-06
Fix segfault after second ANALYZE

### LINKED ISSUE #4857: [Bug]: Segfault after second ANALYZE with PG 14.5/TimescaleDB 2.8.1
### What type of bug is this?

Crash

Other

### What happened?

When I execute `ANALYZE` twice from my application using SQLx the second execution consistently leads to the worker process segfaulting.

Strangely it does not happen when repeatedly executing `ANALYZE` in `psql`.

2.8.1

### PostgreSQL version used

14.5

Arch Linux x86_64

Other

Other

```bash
Oct 19 12:40:45 beryllium postgres[57717]: 2022-10-19 12:40:45.770 CEST [57717] LOG:  server process (PID 58663) was terminated by signal 11: Segmentation fault
Oct 19 12:40:45 beryllium postgres[57717]: 2022-10-19 12:40:45.770 CEST [57717] DETAIL:  Failed process was running: ANALYZE
Oct 19 12:40:45 beryllium postgres[57717]: 2022-10-19 12:40:45.770 CEST [57717] LOG:  terminating any other active server processes
Oct 19 12:40:45 beryllium postgres[57717]: 2022-10-19 12:40:45.771 CEST [57717] LOG:  all server processes terminated; reinitializing
Oct 19 12:40:45 beryllium postgres[58711]: 2022-10-19 12:40:45.785 CEST [58711] LOG:  database system was interrupted; last known up at 2022-10-19 12:40:40 CEST
Oct 19 12:40:45 beryllium postgres[58711]: 2022-10-19 12:40:45.785 CEST [58711] LOG:  database system was not properly shut down; automatic recovery in progress
Oct 19 12:40:45 beryllium postgres[58711]: 2022-10-19 12:40:45.785 CEST [58711] LOG:  redo starts at 0/C80EC4F8
Oct 19 12:40:45 beryllium postgres[58711]: 2022-10-19 12:40:45.797 CEST [58711] LOG:  redo done at 0/C853DF80 system usage: CPU: user: 0.01 s, system: 0.00 s, elapsed: 0.01 s
Oct 19 12:40:45 beryllium postgres[57717]: 2022-10-19 12:40:45.803 CEST [57717] LOG:  database system is ready to accept connections
Oct 19 12:40:45 beryllium postgres[58717]: 2022-10-19 12:40:45.803 CEST [58717] LOG:  TimescaleDB background worker launcher connected to shared catalogs


#0  RangeVarGetRelidExtended (relation=0x0, lockmode=1, flags=0, callback=0x0, callback_arg=0x0) at catalog/namespace.c:246
#1  0x000055f41b74e6af in expand_vacuum_rel (options=66, vrel=0x55f41cd93710) at commands/vacuum.c:778
#2  vacuum (relations=0x55f41cda8b78, params=0x7ffed1ac1ed0, bstrategy=<optimized out>, isTopLevel=<optimized out>) at commands/vacuum.c:392
#3  0x000055f41b74f1c8 in ExecVacuum (pstate=0x55f41cd86120, vacstmt=vacstmt@entry=0x55f41cd8e6f0, isTopLevel=isTopLevel@entry=true) at commands/vacuum.c:268
#4  0x00007f896bfc900c in process_vacuum (args=0x7ffed1ac1ff0) at /usr/src/debug/timescaledb/src/process_utility.c:950
#5  0x00007f896bfc7f64 in process_ddl_command_start (args=0x7ffed1ac1ff0) at /usr/src/debug/timescaledb/src/process_utility.c:4156
#6  timescaledb_ddl_command_start (pstmt=0x55f41cd8e660, query_string=<optimized out>, readonly_tree=<optimized out>, context=PROCESS_UTILITY_TOPLEVEL, 
    params=0x0, queryEnv=<optimized out>, dest=<optimized out>, completion_tag=<optimized out>) at /usr/src/debug/timescaledb/src/process_utility.c:4399
#7  0x000055f41b900684 in ProcessUtility (qc=0x7ffed1ac2460, dest=0x55f41bd32720 <donothingDR.lto_priv.0>, queryEnv=<optimized out>, params=<optimized out>, 
    context=PROCESS_UTILITY_TOPLEVEL, readOnlyTree=<optimized out>, queryString=<optimized out>, pstmt=0x55f41cd8e660) at tcop/utility.c:523
#8  PortalRunUtility (portal=portal@entry=0x55f41cd09518, pstmt=pstmt@entry=0x55f41cd8e660, isTopLevel=isTopLevel@entry=true, 
    setHoldSnapshot=setHoldSnapshot@entry=false, dest=0x55f41bd32720 <donothingDR.lto_priv.0>, qc=0x7ffed1ac2460) at tcop/pquery.c:1155
#9  0x000055f41b900807 in PortalRunMulti (portal=portal@entry=0x55f41cd09518, isTopLevel=isTopLevel@entry=true, setHoldSnapshot=setHoldSnapshot@entry=false, 
    dest=0x55f41bd32720 <donothingDR.lto_priv.0>, dest@entry=0x55f41cc9be90, altdest=0x55f41bd32720 <donothingDR.lto_priv.0>, altdest@entry=0x55f41cc9be90, 
    qc=qc@entry=0x7ffed1ac2460) at tcop/pquery.c:1319
#10 0x000055f41b900d53 in PortalRun (portal=0x55f41cd09518, count=9223372036854775807, isTopLevel=<optimized out>, run_once=<optimized out>, 
    dest=0x55f41cc9be90, altdest=0x55f41cc9be90, qc=0x7ffed1ac2460) at tcop/pquery.c:788
#11 0x000055f41b90279e in exec_execute_message (max_rows=<optimized out>, portal_name=0x55f41cc9ba48 "") at tcop/postgres.c:2187
#12 PostgresMain (argc=<optimized out>, argv=<optimized out>, dbname=<optimized out>, username=<optimized out>) at tcop/postgres.c:4557
#13 0x000055f41b8751b0 in BackendRun (port=0x55f41ccc7320) at postmaster/postmaster.c:4530
#14 BackendStartup (port=0x55f41ccc7320) at postmaster/postmaster.c:4252
#15 ServerLoop () at postmaster/postmaster.c:1745
#16 0x000055f41b875fb2 in PostmasterMain (argc=<optimized out>, argv=<optimized out>) at postmaster/postmaster.c:1417
#17 0x000055f41b58e14b in main (argc=3, argv=0x55f41cc95240) at main/main.c:209
```

```bash
I've only been able to reproduce it so far through SQLx in our proprietary application, not through `psql`.

If required I could try to write a separate reproducing application using SQLx that I could publish, but literally all it's doing is execute `ANALYZE` twice using the same PostgreSQL connection, even against a database with no tables in it.
```

sb230132:
Issue occurs in extended query protocol mode only where every query goes through PREPARE and EXECUTE phase. First time ANALYZE is executed, a list of relations to be vaccumed is extracted and saved in a list. This list is referenced in parsetree node. Once execution of ANALYZE is complete, this list is cleaned up, however reference to the same is not cleaned up in parsetree node. When second time ANALYZE is executed, segfault happens as we access an invalid memory location.

Fixed the issue by restoring the actual value in parsetree node once ANALYZE completes its execution.

Fixes #4857

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 `psql` with the `-X` flag to prevent any `.psqlrc` commands from
 accidentally triggering the load of a previous DB version.**
 
+**Bugfixes**
+* #5054 Fix segfault after second ANALYZE
+
+**Thanks**
+* @kyrias for reporting a crash when ANALYZE is executed on extended query protocol mode with extension loaded.
+
 ## 2.9.0
 This release adds major new features since the 2.8.1 release.
 We deem it moderate priority for upgrading.
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -906,6 +906,8 @@ process_vacuum(ProcessUtilityArgs *args)
 	Hypertable *ht;
 	List *vacuum_rels = NIL;
 	bool is_vacuumcmd;
+	/* save original VacuumRelation list */
+	List *saved_stmt_rels = stmt->rels;
 
 	is_vacuumcmd = stmt->is_vacuumcmd;
 
@@ -975,6 +977,12 @@ process_vacuum(ProcessUtilityArgs *args)
 															  cp->compressed_relid);
 		}
 	}
+	/*
+	Restore original list. stmt->rels which has references to
+	VacuumRelation list is freed up, however VacuumStmt is not
+	cleaned up because of which there is a crash.
+	*/
+	stmt->rels = saved_stmt_rels;
 	return DDL_DONE;
 }
 
--- a/test/sql/size_utils.sql
+++ b/test/sql/size_utils.sql
@@ -273,3 +273,18 @@ SELECT * FROM chunks_detailed_size('hypersize') ORDER BY node_name;
 SELECT * FROM hypertable_compression_stats('hypersize') ORDER BY node_name;
 SELECT * FROM chunk_compression_stats('hypersize') ORDER BY node_name;
 SELECT * FROM hypertable_index_size('hypersize_time_idx');
+
+-- github issue #4857
+-- below procedure should not crash
+SET client_min_messages = ERROR;
+do
+$$
+DECLARE
+  o INT;
+BEGIN
+  FOR c IN 1..20 LOOP
+    ANALYZE;
+  END LOOP;
+END;
+$$;
+RESET client_min_messages;
--- a/test/expected/size_utils.out
+++ b/test/expected/size_utils.out
@@ -744,3 +744,17 @@ SELECT * FROM hypertable_index_size('hypersize_time_idx');
                  24576
 (1 row)
 
+-- github issue #4857
+-- below procedure should not crash
+SET client_min_messages = ERROR;
+do
+$$
+DECLARE
+  o INT;
+BEGIN
+  FOR c IN 1..20 LOOP
+    ANALYZE;
+  END LOOP;
+END;
+$$;
+RESET client_min_messages;


src/process_utility.c
```diff
@@ -975,6 +977,12 @@ process_vacuum(ProcessUtilityArgs *args)
 															  cp->compressed_relid);
 		}
 	}
+	/*
+	Restore original list. stmt->rels which has references to
+	VacuumRelation list is freed up, however VacuumStmt is not
+	cleaned up because of which there is a crash.
+	*/
+	stmt->rels = list_copy(stmt_rels);
```
jnidzwetzki, line 985: I am curious, do you know in which places `stmt->rels` is changed? Is there any change besides the one in line 963 (`stmt->rels = list_concat(ctx.chunk_rels, vacuum_rels);`). 

Maybe we could also fix the problem by creating a new list there and using it for the iteration a few lines later. Then we would save the copy and restore operation and `stmt->rels` would always contain valid entries.

sb230132: You are right. We can fix it either way.
I tried the same approach initially. I created a new list instead of modifying stmt->rels, but i could not pass this new list to ExecVacuum(), as this API expects VacuumStmt as 2nd parameter. Which in turn requires me to create a new VacuumStmt object and set the new list inside it.

Thus i took alternative approach of saving the original list, and restoring the saved list.

test/sql/size_utils.sql
```diff
@@ -273,3 +273,18 @@ SELECT * FROM chunks_detailed_size('hypersize') ORDER BY node_name;
...
+do
+$$
+DECLARE
+  o INT;
+BEGIN
+  FOR c IN 1..2 LOOP
+    ANALYZE;
+  END LOOP;
+END;
+$$;
```
fabriziomello, line 289:
```suggestion
do
$$
BEGIN
  FOR c IN 1..5 LOOP
    RAISE INFO 'ANALYZE %', c;
    ANALYZE;
  END LOOP;
END;
$$;
```
Minor code adjustment. What about execute it more than two times just to make sure it will work beyond the second execution?

src/process_utility.c
fabriziomello: Isn't this similar to this? https://github.com/timescale/timescaledb/commit/74ca546565076676b4b18af3d369465dea4147e3

fabriziomello:
@sb230132 isn't this issue similar to https://github.com/timescale/timescaledb/commit/74ca546565076676b4b18af3d369465dea4147e3 ?

sb230132:
> 

@fabriziomello Issue you pointed out is same as this one. I did try to allocate a new memory context for stmt rel list and free it up, however during completion of this function process_vacuum() i need to restore the stmt rel list. As changes to this data structure should not be visible to the caller of this function.
Something similar is done in src/backend/commands/vacuum.c/vacuum(). Stmt rel list which is passed to vacuum() is pass by value, thus any changes done inside vacuum() is lost once this function call execution is done.

test/sql/size_utils.sql
sb230132: Sure.

fabriziomello:
> > 
> 
> @fabriziomello Issue you pointed out is same as this one. I did try to allocate a new memory context for stmt rel list and free it up, however during completion of this function process_vacuum() i need to restore the stmt rel list. As changes to this data structure should not be visible to the caller of this function. Something similar is done in src/backend/commands/vacuum.c/vacuum(). Stmt rel list which is passed to vacuum() is pass by value, thus any changes done inside vacuum() is lost once this function call execution is done.
> 

Didn't get your point. Where exactly Postgres do it in `src/backend/commands/vacuum.c:vacuum()`? The problem here is subsequent executions of an utility command. The first execution is prepared/cached (extended protocol / plpgsql do it) but the changed relation list is allocated in other memory context, so the subsequent execution will segfault due to memory violation. Am I missing something here?

sb230132:
> > > 
> > 
> > 
> > @fabriziomello Issue you pointed out is same as this one. I did try to allocate a new memory context for stmt rel list and free it up, however during completion of this function process_vacuum() i need to restore the stmt rel list. As changes to this data structure should not be visible to the caller of this function. Something similar is done in src/backend/commands/vacuum.c/vacuum(). Stmt rel list which is passed to vacuum() is pass by value, thus any changes done inside vacuum() is lost once this function call execution is done.
> 
> Didn't get your point. Where exactly Postgres do it in `src/backend/commands/vacuum.c:vacuum()`? The problem here is subsequent executions of an utility command. The first execution is prepared/cached (extended protocol / plpgsql do it) but the changed relation list is allocated in other memory context, so the subsequent execution will segfault due to memory violation. Am I missing something here?


You are right.
Let me give more description:
Check below function:
```
void
ExecVacuum(ParseState *pstate, VacuumStmt *vacstmt, bool isTopLevel)
```
In this function there is a call to 
```
vacuum(vacstmt->rels, &params, NULL, isTopLevel);
```
vacstmt->rels hold list of relations which needs to be vaccumed.
ex:
ANALYZE t1,t2,t2;
vacstmt->rels will have 3 entries representing t1,t2,t3

ANALYZE;
In this case vacstmt->rels has 0 entries.

When timescaledb extension is installed.
First time ANALYZE is executed:
1. In src/process_utility.c::process_vacuum() call,  stmt->rels is filled with list of relations which needs to be vacuumed.
2. This list is then passed on to ExecVacuum().
3. Once process_vacuum() is finished execution the list which holds all relations to be vacuumed is freed up.

Second time ANALYZE is executed:
1. In src/process_utility.c::process_vacuum(),  stmt->rels has the same list which has been deallocated.
2. Same list is accessed in ExecVacuum() and thus causes a crash.
3. Thus we need to reset  stmt->rels everytime at end of process_vacuum() execution.

src/process_utility.c
```diff
@@ -906,6 +906,8 @@ process_vacuum(ProcessUtilityArgs *args)
 	Hypertable *ht;
 	List *vacuum_rels = NIL;
 	bool is_vacuumcmd;
+	/* save original VacuumRelation list */
+	List *stmt_rels = list_copy(stmt->rels);
```
fabriziomello, line 910:
```suggestion
	List *stmt_rels = stmt->rels;
```
Do we really need to copy the list?? Below we're assigning `stmt->rels` to another memory area, no? Specifically in this line:
```
stmt->rels = list_concat(ctx.chunk_rels, vacuum_rels);
```

sb230132: Ok, fixed.

src/process_utility.c
```diff
@@ -906,6 +906,8 @@ process_vacuum(ProcessUtilityArgs *args)
 	Hypertable *ht;
 	List *vacuum_rels = NIL;
 	bool is_vacuumcmd;
+	/* save original VacuumRelation list */
+	List *stmt_rels = stmt->rels;
```
fabriziomello, line 910:
```suggestion
	List *saved_stmt_rels = stmt->rels;
```

src/process_utility.c
fabriziomello, line 985:
```suggestion
	stmt->rels = saved_stmt_rels;
```

fabriziomello (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 `psql` with the `-X` flag to prevent any `.psqlrc` commands from
 accidentally triggering the load of a previous DB version.**
 
+**Thanks**
```
jnidzwetzki, line 7: Nit: The `Thanks` section is usually located below `Features` and `Bugfixes` (i.e., as in the 2.9.0 section https://github.com/timescale/timescaledb/blob/main/CHANGELOG.md).

jnidzwetzki (approved):
One comment, rest looks good to me.

PR #5060, opened 2022-12-07
Correct Sanitizer log directory

jnidzwetzki:
So far, we have treated the `log_path` setting of the sanitizer like a file. In fact, this value is used as a prefix for the created log file. Since we expected the exact file name when uploading the sanitizer output, this file was not found and we lost the messages of the sanitizer. This PR changes the behavior. We now treat the setting as a prefix and upload all files created in a dedicated sanitizer output folder.

Example output when `log_path` is set to `/tmp/sanitizer/sanitizer` :

```
$ ls -l /tmp/sanitizer/
total 16
-rw------- 1 jan jan 13757 Dec  7 17:48 sanitizer.postgres.1894562
```

Link to CI run with failed upload: https://github.com/timescale/timescaledb/actions/runs/3633428885


fabriziomello (approved):

jnidzwetzki:
@akuzm Good point, I will do this in a follow up PR.

PR #5063, opened 2022-12-08
Fix memory leak for compression with merge chunks

cevian:
The RelationInitIndexAccessInfo call leaks cache memory and seems to be unnecessary.


akuzm (approved):
Yeah, not sure why we're calling it directly, it's already called by `relation_open -> RelationIdGetRelation -> RelationBuildDesc`.

antekresic (approved):
Yeah, this was my bad, was trying to fill it manually because of some other issue I was dealing with and it got left in when I fixed that issue.

PR #5080, opened 2022-12-12
Allow analyze to work on dist hypertable list

### LINKED ISSUE #4046: [Enhancement]: ANALYZE should analyze distributed hypertables

User experience

Command processing, Distributed hypertable

Running a "global" `ANALYZE` will not import stats for distributed hypertables, although such tables can be individually analyzed with `ANALYZE disttable` at which point stats will be imported from data nodes.

For convenience, the global `ANALYZE` should also include distributed hypertables.

### Implementation challenges

Importing stats from data nodes could be costly and might slow down `ANALYZE`. In some cases, data nodes might not respond due to temporary failures and that might break `ANALYZE`. Ideally,  in those failure scenarios, the global `ANALYZE` should probably succeed but raise a notice/warning that a distributed hypertable couldn't be analyzed.

65278:
Analyze currently does not work on a list of distributed hypertables because this would require deparsing of vacuum statements and sending a separate vacuum statement per data node.

To execute an unqualified vacuum (all tables), we need to decide if the sql of a command includes specific tables. We can not do this from a VacuumStmt node but would need to add an extra field to ProcessUtilityArgs.

I decided to go the extra mile and implement a vacuum statement deparser, solving the entire problem space.

A compromise was made on unqualified VERBOSE statements, as these would require parsing the DistResponseSet. If VERBOSE is specified in such a way, we simply emit a notice about skipping these tables. This remains an error when a list of tables is specified.

Fixes #4046

### DIFF
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -826,10 +826,9 @@ add_chunk_to_vacuum(Hypertable *ht, Oid chunk_relid, void *arg)
  * from vacuum.c, only it filters out distributed hypertables and chunks
  * that have been compressed.
  */
-static List *
-ts_get_all_vacuum_rels(bool is_vacuumcmd)
+static void
+ts_get_all_vacuum_rels(bool is_vacuumcmd, List **vacrels, List **disthtrels)
 {
-	List *vacrels = NIL;
 	Relation pgclass;
 	TableScanDesc scan;
 	HeapTuple tuple;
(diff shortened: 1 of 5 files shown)


65278:
Work in progress, a locking test fails and I've not implemented test cases yet.

erimatnor:
Closing for now since no progress.

PR #5089, opened 2022-12-14
Fix git permission issue in CI build

jnidzwetzki:
The new permissions checks to fix CVE-2022-29187 in Git caused some issues in our CI pipeline. This patch adds the checkout directory to Git's "safe.directory" setting.

Failed CI run: https://github.com/timescale/timescaledb/actions/runs/3690974658/jobs/6248545042

### DIFF
--- a/.github/workflows/linux-32bit-build-and-test.yaml
+++ b/.github/workflows/linux-32bit-build-and-test.yaml
@@ -63,6 +63,10 @@ jobs:
 
     - name: Build TimescaleDB
       run: |
+        # The owner of the checkout directory and the files do not match. Add the directory to
+        # Git's "safe.directory" setting. Otherwise git would complain about
+        # 'detected dubious ownership in repository'
+        git config --global --add safe.directory $(pwd)
         ./bootstrap -DCMAKE_BUILD_TYPE="${{ matrix.build_type }}" -DPG_SOURCE_DIR=~/postgresql -DREQUIRE_ALL_TESTS=ON
         make -C build install
         chown -R postgres:postgres .


svenklemm (changes requested):
I think the better approach here is to make git skip that check by adjusting the configuration as we do in other places. E.g. https://github.com/timescale/timescaledb/commit/17844a4087

jnidzwetzki:
I am in favor of fixing the permission problem than disabling the new Git security checks. In this specific case, we have a mismatch between the owner of the checkout directory and the checkout. Is there any technical problem with this solution?

You mentioned in Slack, that the user is controlled by the GitHub runner. This is true, but if GitHub decides to let the workflow run as another user as root, a lot of things (e.g., `apt-get`) would also break. I could not find a specification for it. But I don't suspect that this will change unannounced.

svenklemm:
> I am in favor of fixing the permission problem than disabling the new Git security checks. In this specific case, we have a mismatch between the owner of the checkout directory and the checkout. Is there any technical problem with this solution?
> 
> You mentioned in Slack, that the user is controlled by the GitHub runner. This is true, but if GitHub decides to let the workflow run as another user as root, a lot of things (e.g., `apt-get`) would also break. I could not find a specification for it. But I don't suspect that this will change unannounced.

Running a recursive chown is several orders of magnitude slower than adjusting the git config and this does not account for the additional runtime of the check itself. The suggested solution with chown only works for that specific user and prevents sharing the checkout between multiple steps when the checkout directory is bind mounted into docker containers as we do in some of the other workflows.

fabriziomello (approved):

jnidzwetzki:
@svenklemm  Changed, let's do in that way.

svenklemm (approved):

Issue #5090, opened 2022-12-14
[Bug]: Roll up chunks during compression causes planner issues, no query plan generated 


antekresic:
### What type of bug is this?

Unexpected error

### What subsystems and features are affected?

Compression

### What happened?

When compressing chunks with compressed chunk interval parameter set, chunks are being rolled up into a single chunk.

The issue occurs when trying to query the hypertable using a single segment by column qualifier.

### TimescaleDB version affected

main branch (soon to be 2.9.0)

### PostgreSQL version used

14.5

### What operating system did you use?

Ubuntu 22.04

### What installation method did you use?

Source

### What platform did you run on?

Other

### Relevant log output and stack trace

```bash
postgres=# select * from metric where series_id = 1;
ERROR:  could not devise a query plan for the given query
```

### How can we reproduce the bug?

```bash
create extension timescaledb;

CREATE TABLE metric(time TIMESTAMPTZ NOT NULL, value DOUBLE PRECISION NOT NULL, series_id BIGINT NOT NULL);

select create_hypertable('metric', 'time', chunk_time_interval => interval '1 h');

ALTER TABLE metric set (timescaledb.compress,
    timescaledb.compress_segmentby = 'series_id',
    timescaledb.compress_orderby = 'time',
    timescaledb.compress_chunk_time_interval = '1 day'
    );


insert into metric (time, series_id, value) select t, s, 1 from generate_series(now(), now()+interval'4h', interval '30s') t cross  join generate_series(0, 100, 1) s;

select compress_chunk(c, true) from show_chunks('metric') c;
select * from metric where series_id = 1;
```

PR #5093, opened 2022-12-14
Allow BitmapHeapScans on non-parametrized paths

antekresic:
Planner can decide to use BitmapHeapScans exclusively when
scanning compressed chunks. Since we filter out such scans
due to previously reported issues, this can lead to no
query plan devised when querying compressed chunks. This
change allows Bitmap scans on compressed chunks only when
it is not parametrized.

Fixes https://github.com/timescale/timescaledb/issues/5090

Disable-check: commit-count

### DIFF
--- a/tsl/src/nodes/decompress_chunk/decompress_chunk.c
+++ b/tsl/src/nodes/decompress_chunk/decompress_chunk.c
@@ -426,14 +426,14 @@ ts_decompress_chunk_generate_paths(PlannerInfo *root, RelOptInfo *chunk_rel, Hyp
 		DecompressChunkPath *path;
 
 		/*
-		 * We skip any BitmapScan paths here as supporting those
-		 * would require fixing up the internal scan. Since we
+		 * We skip any BitmapScan parameterized paths here as supporting
+		 * those would require fixing up the internal scan. Since we
 		 * currently do not do this BitmapScans would be generated
 		 * when we have a parameterized path on a compressed column
 		 * that would have invalid references due to our
 		 * EquivalenceClasses.
 		 */
-		if (IsA(child_path, BitmapHeapPath))
+		if (IsA(child_path, BitmapHeapPath) && child_path->param_info)
 			continue;
 
 		/*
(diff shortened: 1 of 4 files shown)


tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -426,14 +426,14 @@ ts_decompress_chunk_generate_paths(PlannerInfo *root, RelOptInfo *chunk_rel, Hyp
 		DecompressChunkPath *path;
 
 		/*
-		 * We skip any BitmapScan paths here as supporting those
-		 * would require fixing up the internal scan. Since we
+		 * We skip any BitmapScan parametrized paths here as supporting
```
svenklemm, line 429:
```suggestion
		 * We skip any BitmapScan parameterized paths here as supporting
```

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -1204,6 +1207,9 @@ create_compressed_scan_paths(PlannerInfo *root, RelOptInfo *compressed_rel, int
 		add_partial_path(compressed_rel, compressed_path);
 	}
 
+	bool old_bitmapscan = enable_bitmapscan;
+	enable_bitmapscan = false;
+
```
svenklemm, line 1212:
```suggestion
	  /*
   * We set enable_bitmapscan to false here to ensure any pathes with bitmapscan do not
   * displace other pathes. Note that setting the postgres GUC will not actually disable
   * the bitmapscan path creation but will instead create them with very high cost.
   * If bitmapscan were the dominant path after postgres planning we could end up
   * in a situation where we have no valid plan for this relation because we remove
   * bitmapscan pathes from the pathlist.
  */ 
	bool old_bitmapscan = enable_bitmapscan;
	enable_bitmapscan = false;
```

svenklemm (approved):

tsl/src/nodes/decompress_chunk/decompress_chunk.c
antekresic: Updated.

tsl/src/nodes/decompress_chunk/decompress_chunk.c
```diff
@@ -1204,6 +1207,18 @@ create_compressed_scan_paths(PlannerInfo *root, RelOptInfo *compressed_rel, int
 		add_partial_path(compressed_rel, compressed_path);
 	}
 
+	/*
+	 * We set enable_bitmapscan to false here to ensure any pathes with bitmapscan do not
```
akuzm, line 1211: "paths", but let's ignore this typo not to restart the checks.

akuzm (approved):

PR #5106, opened 2022-12-19
Run regression tests on pg15 windows

svenklemm:
Disable-check: commit-count

### DIFF
--- a/src/bgw/job_stat.c
+++ b/src/bgw/job_stat.c
@@ -272,7 +272,7 @@ static float8
 calculate_jitter_percent()
 {
 	/* returns a number in the range [-0.125, 0.125] */
-	uint8 percent = random();
+	uint8 percent = rand();
 	return ldexp((double) (16 - (int) (percent % 32)), -7);
 }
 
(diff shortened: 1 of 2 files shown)


src/bgw/job_stat.c
```diff
@@ -272,7 +272,7 @@ static float8
 calculate_jitter_percent()
 {
 	/* returns a number in the range [-0.125, 0.125] */
-	uint8 percent = random();
+	uint8 percent = rand();
```
konskov, line 275: it's not obvious to me why this change is required?

svenklemm: The explanation is in the commit message of the 2nd commit:
```
Use rand() instead of random() cause the latter is not available
on Windows and postgres stopped backporting it with PG15.
Ideally we switch to the crypto functions added in PG15 but that
requires a bit more work and this is the minimal change required
to get it to build against PG15 on Windows.
```

konskov: I see - sorry I made you repeat yourself here!

konskov (approved):

sb230132 (approved):

PR #5108, opened 2022-12-20
Fix column ordering in compressed table index.

### LINKED ISSUE #5104: [Bug]: Segment by index ordering determined by table definition
### What type of bug is this?

Configuration

Compression

### What happened?

When defining compression segment by parameter using multiple columns, the parameter ordering is not respected for index creation. Ordering used is based on the table definition which is counter intuitive and can lead to performance degradation.

2.9.0

### PostgreSQL version used

14.5

Ubuntu 22.04

Source

On prem/Self-hosted

```bash
postgres=# \d+ _timescaledb_internal.compress_hyper_2_10_chunk
                                               Table "_timescaledb_internal.compress_hyper_2_10_chunk"
        Column         |                 Type                  | Collation | Nullable | Default | Storage  | Compression | Stats target | Description 
-----------------------+---------------------------------------+-----------+----------+---------+----------+-------------+--------------+-------------
 time                  | _timescaledb_internal.compressed_data |           |          |         | external |             | 0            | 
 value                 | double precision                      |           |          |         | plain    |             | 1000         | 
 series_id             | bigint                                |           |          |         | plain    |             | 1000         | 
 _ts_meta_count        | integer                               |           |          |         | plain    |             | 1000         | 
 _ts_meta_sequence_num | integer                               |           |          |         | plain    |             | 1000         | 
 _ts_meta_min_1        | timestamp with time zone              |           |          |         | plain    |             | 1000         | 
 _ts_meta_max_1        | timestamp with time zone              |           |          |         | plain    |             | 1000         | 
Indexes:
    "compress_hyper_2_10_chunk__compressed_hypertable_2_value_series" btree (value, series_id, _ts_meta_sequence_num)
Inherits: _timescaledb_internal._compressed_hypertable_2
Access method: heap
Options: toast_tuple_target=128
```

```bash
postgres=# CREATE TABLE metric(time TIMESTAMPTZ NOT NULL, value DOUBLE PRECISION NOT NULL, series_id BIGINT NOT NULL);                                                                                                                                      
postgres=# select create_hypertable('metric', 'time', chunk_time_interval => interval '1 h', create_default_indexes => false);
postgres=# ALTER TABLE metric set (timescaledb.compress,
    timescaledb.compress_segmentby = 'series_id, value',
    timescaledb.compress_orderby = 'time'
    );
```

sb230132:
When defining compression segment by parameter using multiple columns, the parameter ordering is not respected for index creation.

This patch fixes the issue by maintaining the same order in which user has defined columns in segment by clause.

Fixes #5104

### DIFF
--- a/tsl/test/sql/compression_ddl.sql
+++ b/tsl/test/sql/compression_ddl.sql
@@ -517,3 +517,32 @@ SELECT timescaledb_pre_restore();
 DROP TABLE :ctable;
 SELECT timescaledb_post_restore();
 DROP TABLE issue4140;
+
+-- github issue 5104
+CREATE TABLE metric(
+	time TIMESTAMPTZ NOT NULL,
+	value DOUBLE PRECISION NOT NULL,
+	series_id BIGINT NOT NULL);
+
+SELECT create_hypertable('metric', 'time',
+	chunk_time_interval => interval '1 h',
+	create_default_indexes => false);
+
+-- enable compression
+ALTER TABLE metric set(timescaledb.compress,
+    timescaledb.compress_segmentby = 'series_id, value',
+    timescaledb.compress_orderby = 'time'
+);
+
+SELECT
+      comp_hypertable.schema_name AS "COMP_SCHEMA_NAME",
+      comp_hypertable.table_name AS "COMP_TABLE_NAME"
+FROM _timescaledb_catalog.hypertable uc_hypertable
+INNER JOIN _timescaledb_catalog.hypertable comp_hypertable ON (comp_hypertable.id = uc_hypertable.compressed_hypertable_id)
+WHERE uc_hypertable.table_name like 'metric' \gset
+
+-- get definition of compressed hypertable and notice the index
+\d :COMP_SCHEMA_NAME.:COMP_TABLE_NAME
+
+DROP TABLE metric CASCADE;
+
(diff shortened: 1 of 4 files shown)


akuzm (approved):
Would be good to have a specific test for this.

fabriziomello (changes requested):
Need to add regression tests according to the issue!

sb230132:
> Need to add regression tests according to the issue!

Added testcase.

tsl/test/sql/compression_ddl.sql
```diff
@@ -517,3 +517,31 @@ SELECT timescaledb_pre_restore();
...
+      comp_hypertable.schema_name AS "COMP_SCHEMA_NAME",
+      comp_hypertable.table_name AS "COMP_TABLE_NAME"
+FROM _timescaledb_catalog.hypertable uc_hypertable
+INNER JOIN _timescaledb_catalog.hypertable comp_hypertable ON (comp_hypertable.id = uc_hypertable.compressed_hypertable_id)
+WHERE uc_hypertable.table_name like 'metric' \gset
+
+-- get definition of compressed hypertable and notice the index
+\d :COMP_SCHEMA_NAME.:COMP_TABLE_NAME
+
+DROP TABLE metric CASCADE;
```
fabriziomello, line 547: Missing newline at the end of file

fabriziomello (approved):

PR #5109, opened 2022-12-20
Skip package downgrade test for PG15

svenklemm:
Since we currently only have 1 released version that supports PG15 we cannot test downgrade with PG15.


.github/workflows/apt-packages.yaml
```diff
@@ -4,6 +4,7 @@ on:
   schedule:
     # run daily 0:00 on main branch
     - cron: '0 0 * * *'
+  pull_request:
```
akuzm, line 7: Do we want to run this test for PRs?

akuzm (approved):

.github/workflows/apt-packages.yaml
svenklemm: no this was just for testing that the change works, will remove before merging

fabriziomello (approved):

PR #5114, opened 2022-12-22
Fix issue with deleting data node and dropping database

erimatnor:
When deleting a data node with the option `drop_database=>true`, the database is deleted even if the command fails for other reasons.

Fix this behavior by dropping the remote database at the end of the delete data node operation so that other checks fail first.


nikkhils (approved):
LGTM!

PR #5130, opened 2022-12-26
Fix CAgg on CAgg variable bucket size validation

### LINKED ISSUE #5126: [Bug]: CAGG on CAGG Error
### What type of bug is this?

Unexpected error

Continuous aggregate

### What happened?

When attempting to create a CAG using a `1 HOUR` bucket based on a CAG with a `5 MINUTE` bucket the following error is thrown:
```
[2022-12-24 23:40:15] [22P01] ERROR: floating-point exception
[2022-12-24 23:40:15] Detail: An invalid floating-point operation was signaled. This probably means an out-of-range result or an invalid operation, such as division by zero.
```

2.9.1

### PostgreSQL version used

14.6

psql (15.1 (Ubuntu 15.1-1.pgdg20.04+1), server 14.6 (Ubuntu 14.6-1.pgdg20.04+1))

Deb/Apt

On prem/Self-hosted

```bash
[2022-12-24 23:40:15] [22P01] ERROR: floating-point exception
[2022-12-24 23:40:15] Detail: An invalid floating-point operation was signaled. This probably means an out-of-range result or an invalid operation, such as division by zero.
```

```bash
DROP MATERIALIZED VIEW IF EXISTS bug_timescaledb_table_1h;
DROP MATERIALIZED VIEW IF EXISTS bug_timescaledb_table_5m;
DROP TABLE             IF EXISTS bug_timescaledb_table;

CREATE TABLE bug_timescaledb_table (
    recorded_at    TIMESTAMPTZ NOT NULL,

    a VARCHAR(50),
    b VARCHAR(50),

    c REAL,
    d REAL,
    e REAL,

    f REAL,
    g REAL,
    h REAL
);

SELECT CREATE_HYPERTABLE(
    'bug_timescaledb_table',
    'recorded_at',
    chunk_time_interval => INTERVAL '30 DAY'
);

CREATE MATERIALIZED VIEW bug_timescaledb_table_5m
        SELECT
            TIME_BUCKET('5 MINUTE', recorded_at, 'US/Pacific') AS recorded_at,
            a AS a,
            MIN(c) AS c_min
        FROM bug_timescaledb_table
        GROUP BY 1, 2
        WITH NO DATA;

CREATE MATERIALIZED VIEW bug_timescaledb_table_1h
        SELECT
            TIME_BUCKET('1 HOUR', recorded_at, 'US/Pacific') AS recorded_at,
            a AS a,
            MIN(c_min) AS c_min
        FROM bug_timescaledb_table_5m
        GROUP BY 1, 2
        WITH NO DATA;
```
The error will be thrown when attempting to create `bug_timescaledb_table_1h`.

fabriziomello:
During the creation of a CAgg on top of another CAgg we check if the bucket width is multiple of the parent and for this arithmetic we made an assumption that picking just the `month` part of the `interval` for variable bucket size was enought.

This assumption was wrong for bucket sizes that doesn't have the month part (i.e: using timezones) leading to division by zero error because in that case the `month` part of an `interval` is equal to 0 (zero).

Fixed it by properly calculating the bucket width for variable sized buckets.

Fixes #5126

### DIFF
--- a/tsl/src/continuous_aggs/create.c
+++ b/tsl/src/continuous_aggs/create.c
@@ -160,6 +160,7 @@ typedef struct CAggTimebucketInfo
 	int64 htpartcol_interval_len; /* interval length setting for primary partitioning column */
 	int64 bucket_width;			  /* bucket_width of time_bucket, stores BUCKET_WIDTH_VARIABLE for
 									 variable-sized buckets */
+	Oid bucket_width_type;		  /* type of bucket_width */
 	Interval *interval;			  /* stores the interval, NULL if not specified */
 	const char *timezone;		  /* the name of the timezone, NULL if not specified */
 
@@ -698,10 +699,11 @@ caggtimebucketinfo_init(CAggTimebucketInfo *src, int32 hypertable_id, Oid hypert
 	src->htpartcolno = hypertable_partition_colno;
 	src->htpartcoltype = hypertable_partition_coltype;
 	src->htpartcol_interval_len = hypertable_partition_col_interval;
-	src->bucket_width = 0;			/* invalid value */
-	src->interval = NULL;			/* not specified by default */
-	src->timezone = NULL;			/* not specified by default */
-	TIMESTAMP_NOBEGIN(src->origin); /* origin is not specified by default */
+	src->bucket_width = 0;				 /* invalid value */
+	src->bucket_width_type = InvalidOid; /* invalid oid */
+	src->interval = NULL;				 /* not specified by default */
+	src->timezone = NULL;				 /* not specified by default */
+	TIMESTAMP_NOBEGIN(src->origin);		 /* origin is not specified by default */
 }
 
 static Const *
@@ -743,6 +745,7 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
 	{
 		SortGroupClause *sgc = (SortGroupClause *) lfirst(l);
 		TargetEntry *tle = get_sortgroupclause_tle(sgc, targetList);
+
 		if (IsA(tle->expr, FuncExpr))
 		{
 			FuncExpr *fe = ((FuncExpr *) tle->expr);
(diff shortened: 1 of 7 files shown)


tsl/src/continuous_aggs/create.c
```diff
@@ -743,6 +745,7 @@ caggtimebucket_validate(CAggTimebucketInfo *tbinfo, List *groupClause, List *tar
 	{
 		SortGroupClause *sgc = (SortGroupClause *) lfirst(l);
 		TargetEntry *tle = get_sortgroupclause_tle(sgc, targetList);
+
```
RafiaSabih, line 748: extra line

fabriziomello: I know we should try to avoid reformatting code in our normal PRs but this is a minor change and just to make the code a bit more readable. Also added an extra line in `cagg_validate_query` .

RafiaSabih (approved):

tsl/src/continuous_aggs/create.c
```diff
@@ -1091,6 +1096,33 @@ cagg_query_supported(const Query *query, StringInfo hint, StringInfo detail, con
...
 }
 
+static inline int64
+get_bucket_width(CAggTimebucketInfo bucket_info)
+{
+	if (bucket_info.bucket_width == BUCKET_WIDTH_VARIABLE)
+	{
+		/* if bucket size is VARIABLE and have month part then pick it for the arithmetic */
+		if (bucket_info.interval->month != 0)
+			return bucket_info.interval->month;
```
svenklemm, line 1106: hmm this seems completely different from the return value in the other codepath i think it would be cleaner if caller does the switch and decides whether he wants microseconds interval or month interval

fabriziomello: thinking better about it I guess we always can get the microseconds interval for variable buckets, right?? I mean don't need this switch

akuzm (approved):
LGTM. Codecov is at it again, seems that it doesn't work for the cagg on cagg test?

fabriziomello:
> LGTM. Codecov is at it again, seems that it doesn't work for the cagg on cagg test?

I'll have a look on this. Perhaps the "dynamic" way that I wrote those tests to exercise all variations of time dimensions in caggs on caggs can make the codedov confused.

PR #5133, opened 2022-12-28
Fix CAgg on CAgg different column order

### LINKED ISSUE #5131: [Bug]: Time bucket error while creating an CAGG on another one
### What type of bug is this?

Unexpected error

Continuous aggregate

### What happened?

When trying to create a CAGG on another one, we encountered the following error :   
 `ERROR: time bucket function must reference a hypertable dimension column`

While investigate on the error, we found that the bug was thrown if the order of the time partitioning column mismatch between the first materialized view and the hypertable. For exemple, if the time partitioning column is the second in the table and the `time bucket` column is the first in the CAGG, creating another CAGG will cause this error.

As workaround, we use the same position for the `time bucket`  as the time column on the table ; but we manage to have in some case the time column in the sixth position and only four column in the CAGG :sweat_smile: (recreating the CAGG with empty column did the trick)

2.9.1

### PostgreSQL version used

13

timescale/timescaledb:2.9.1-pg13

Docker

On prem/Self-hosted

```bash
ERROR:  time bucket function must reference a hypertable dimension column
```

```bash
CREATE TABLE data(
    temperature INT8,
    date TIMESTAMPTZ NOT NULL
);

SELECT create_hypertable('data', 'date');

CREATE MATERIALIZED VIEW cagg
WITH(timescaledb.continuous) AS
SELECT time_bucket('1hour', date) AS hour,
  min(temperature),  
  avg(temperature),
  max(temperature)
FROM data
GROUP BY hour;

CREATE MATERIALIZED VIEW cagg_on_cagg
WITH(timescaledb.continuous) AS
SELECT time_bucket('1day', hour) AS day,
  min(min),
  avg(avg),
  max(max)
FROM cagg
GROUP BY day;
```

fabriziomello:
Creating a CAgg on a CAgg where the time column is in a different order of the original hypertable was raising the following exception:

`ERROR:  time bucket function must reference a hypertable dimension column`

The problem was during the validation we're initializing internal data structure with the wrong hypertable metadata.

Fixes #5131

### DIFF
--- a/tsl/test/expected/cagg_on_cagg.out
+++ b/tsl/test/expected/cagg_on_cagg.out
@@ -18,10 +18,13 @@
 \set BUCKET_WIDTH_1ST 'INTEGER \'1\''
 \set BUCKET_WIDTH_2TH 'INTEGER \'5\''
 \set BUCKET_WIDTH_3TH 'INTEGER \'10\''
-\ir include/cagg_on_cagg_common.sql
+-- Different order of time dimension in raw ht
+\set IS_DEFAULT_COLUMN_ORDER FALSE
+\ir include/cagg_on_cagg_setup.sql
 -- This file and its contents are licensed under the Timescale License.
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-TIMESCALE for a copy of the license.
+-- CAGGs on CAGGs tests
 \if :IS_DISTRIBUTED
 \echo 'Running distributed hypertable tests'
 \else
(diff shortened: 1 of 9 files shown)


tsl/test/expected/cagg_on_cagg.out
```diff
@@ -18,22 +18,33 @@
...
 Running local hypertable tests
 \endif
 SET ROLE :ROLE_DEFAULT_PERM_USER;
--- CAGGs on CAGGs tests
-CREATE TABLE conditions (
-  time :TIME_DIMENSION_DATATYPE NOT NULL,
-  temperature NUMERIC
-);
+DROP TABLE IF EXISTS conditions CASCADE;
+psql:include/cagg_on_cagg_setup.sql:14: NOTICE:  table "conditions" does not exist, skipping
```
RafiaSabih, line 36: Is this necessary?

fabriziomello: At this point no but the new file `include/cagg_on_cagg_setup.sql` will be included several times so we'll need to cleanup the hypertable to re-setup the test case

RafiaSabih (approved):

akuzm (approved):
Weird that codecov complains about the lines that should be covered...

fabriziomello:
> Weird that codecov complains about the lines that should be covered...

It is really weird. In another simple fix I've added a failing test and codecov complains exactly the code I added to fix the failing test  :-(

PR #5136, opened 2023-01-02
SELECT from partial compressed chunks crashes

### LINKED ISSUE #5134: [Bug]: Crash when running `explain (analyze, verbose)` after insert in compressed chunk
### What type of bug is this?

Crash

Compression

### What happened?

Server crashed when running `explain (verbose, analyze) select * from _timescaledb_internal._hyper_1_2_chunk ;` on a chunk that has been inserted into after being compressed

2.10.0-dev

### PostgreSQL version used

13.6

macOS ventura 13.1

Source

On prem/Self-hosted

_No response_

```bash
create table mytab (time timestamptz not null, a int, b int, c int);

SELECT create_hypertable('mytab', 'time', chunk_time_interval => interval '1 day');

insert into mytab
select time, 
case when (now() - time < interval '1 days') THEN 1 
when (now() - time < interval '2 days') THEN 2 
when (now() - time < interval '3 days') THEN 3 ELSE 4 END as a
from generate_series(now() - interval '3 days', now(), interval '5 sec') as g1(time);

UPDATE mytab SET c = a;

alter table mytab set (timescaledb.compress, timescaledb.compress_segmentby = 'a, c');
```
Compress one of the chunks and insert a value into the table that falls in the range of the compressed chunk (so it will be inserted into the compressed chunk). Then do 
`explain (verbose, analyze) select * from _timescaledb_internal._hyper_1_2_chunk ;` for the compressed chunk
```

### PR DESCRIPTION
**Pull request by sb230132**
SELECT from partially compressed chunk crashes due to reference to NULL pointer. When generating paths for DecompressChunk, uncompressed_partial_path is null which is not checked, thus causing a crash. This patch checks for NULL before calling create_append_path().

Fixes #5134

### DIFF
--- a/tsl/test/shared/sql/compression_dml.sql
+++ b/tsl/test/shared/sql/compression_dml.sql
@@ -54,3 +54,35 @@ INSERT INTO metric_5m (time, series_id, value)
 -- clean up
 RESET work_mem;
 DROP TABLE metric_5m;
+
+-- github issue 5134
+CREATE TABLE mytab (time TIMESTAMPTZ NOT NULL, a INT, b INT, c INT);
+SELECT table_name FROM create_hypertable('mytab', 'time', chunk_time_interval => interval '1 day');
+
+INSERT INTO mytab
+    SELECT time,
+        CASE WHEN (:'start_date'::timestamptz - time < interval '1 days') THEN 1
+             WHEN (:'start_date'::timestamptz - time < interval '2 days') THEN 2
+             WHEN (:'start_date'::timestamptz - time < interval '3 days') THEN 3 ELSE 4 END as a
+    from generate_series(:'start_date'::timestamptz - interval '3 days', :'start_date'::timestamptz, interval '5 sec') as g1(time);
+
+-- enable compression
+ALTER TABLE mytab SET (
+    timescaledb.compress,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by svenklemm — APPROVED** [review]

File: tsl/test/shared/sql/compression_dml.sql [hunk 1]
```diff
@@ -53,4 +53,35 @@ INSERT INTO metric_5m (time, series_id, value)
...
+       WHERE hypertable_name = 'mytab' ORDER BY range_start limit 1 \gset
+
+-- compress only the first chunk
+SELECT compress_chunk(:'chunk_table');
+
+-- insert a row into first compressed chunk
+INSERT INTO mytab SELECT '2022-10-07 05:30:10+05:30'::timestamp with time zone, 3, 3;
+-- should not crash
+EXPLAIN (costs off) SELECT * FROM :chunk_table;
+DROP TABLE mytab CASCADE;
```
fabriziomello, line 87: Missing new line at the end of file!

fabriziomello (approved):

tsl/test/shared/sql/compression_dml.sql
sb230132: Fixed.

PR #5144, opened 2023-01-03
Implement incremental materialized view

RafiaSabih:
Expose functions to enable incremental materialized views using the existing continuous aggregates infrastructure.
This will add flexibility of creating views with joins of hypertables and/or normal tables and refresh them as per the
requirement.

### DIFF
--- a/sql/cagg_utils.sql
+++ b/sql/cagg_utils.sql
@@ -120,3 +120,71 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
     OUT ret_window_start BIGINT,
     OUT ret_window_end BIGINT
 ) RETURNS RECORD AS '@MODULE_PATHNAME@', 'ts_invalidation_process_cagg_log' LANGUAGE C STRICT VOLATILE;
+
+--Get the maximum value materialized for the given hypertable
+CREATE OR REPLACE FUNCTION timescaledb_experimental.get_materialization_threshold(
+   mat_ht varchar,
+   time_col varchar
+)  RETURNS varchar
+LANGUAGE 'plpgsql'
+ AS    $func$
+ DECLARE mat_thresh TIMESTAMP;
+         query_string text;
+BEGIN
+    query_string = format('SELECT max(%s) FROM %s', time_col, mat_ht);
+    EXECUTE query_string into mat_thresh;
+    RETURN mat_thresh;
+END;
+$func$;
+
+--Function to materialize the hypertable --mat_ht using join definition in join_view
+--using the target hypertable target_ht over the provided target_thresh
+CREATE OR REPLACE FUNCTION timescaledb_experimental.execute_materialization(
+    join_view varchar,
+    mat_ht varchar,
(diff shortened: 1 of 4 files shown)


sql/cagg_utils.sql
```diff
@@ -120,3 +120,38 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
...
+    mat_ht varchar,
+    target_ht varchar,
+    time_col varchar,
+    prev_mat_thresh varchar,
+    target_mat_thresh varchar
+)  RETURNS VOID
+ LANGUAGE 'plpgsql'
+ AS    $func$
+BEGIN
+   EXECUTE format('LOCK TABLE ONLY %s IN SHARE ROW EXCLUSIVE MODE', target_ht);
```
davidkohn88, line 153: Why is this lock being taken?

sql/cagg_utils.sql
```diff
@@ -120,3 +120,38 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
...
+   EXECUTE format('REFRESH MATERIALIZED VIEW %s',join_view);
```
davidkohn88, line 154: What materialized view is this?

sql/cagg_utils.sql
```diff
@@ -120,3 +120,38 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
...
+BEGIN
+    query_string = format('SELECT max(%s) FROM %s', time_col, mat_ht);
+    EXECUTE query_string into mat_thresh;
+    RETURN mat_thresh;
+END;
+$func$;
+
+--Function to materialize the hypertable --mat_ht using join definition in join_view
+--using the target hypertable target_ht over the provided target_thresh
+CREATE OR REPLACE FUNCTION timescaledb_experimental.execute_materialization(
```
davidkohn88, line 142: This isn't one of the APIs we've talked about...and I'm not sure that it's what we want. I don't think we're doing the right level of abstraction here.

davidkohn88 (commented):
I've left a few comments I'm a bit confused as to what we're doing here...I don't think this was the type of abstraction we've talked about before and feels like it's too tied to a specific implementation...but I might be missing something?

sql/cagg_utils.sql
```diff
@@ -120,3 +120,38 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.invalidation_process_cagg_log(
...
+LANGUAGE 'plpgsql'
+ AS    $func$
+ DECLARE mat_thresh TIMESTAMP;
+         query_string text;
+BEGIN
+    query_string = format('SELECT max(%s) FROM %s', time_col, mat_ht);
+    EXECUTE query_string into mat_thresh;
+    RETURN mat_thresh;
+END;
+$func$;
```
davidkohn88, line 138: Make a set version even if it's a no-op now.

horzsolt:
We don't have the bandwidth currently for this one, so closing.

PR #5145, opened 2023-01-03
Add join functionality to the deparser code

jnidzwetzki:
This patch adds the missing functionality to handle joins to the deparser. Most of the code is a backport from PG Upstream.

**Note:** This is the first PR for the join functionality in the multi-node code. Test cases will follow with the PR containing the planner changes.


akuzm (approved):
I'd prefer to commit this as part of the main PR, so that it's not dead code and we could see the test coverage.

nikkhils (approved):
LGTM!

PR #5150, opened 2023-01-05
MERGE support on hypertables

### LINKED ISSUE #5139: Support MERGE command

User experience

Command processing

MERGE sql command on hypertables should function as expected.

### Implementation challenges

_No response_

sb230132:
This patch does following:

1. Planner changes to create ChunkDispatch node when MERGE command
   has INSERT action.
2. Changes to map partition attributes from a tuple returned from
   child node of ChunkDispatch against physical targetlist, so that
   ChunkDispatch node can read the correct value from partition column.
3. Fixed issues with MERGE on compressed hypertable.
4. Added more testcases.
5. MERGE in distributed hypertables is not supported.
6. Since there is no Custom Scan (HypertableModify) node for MERGE
   with UPDATE/DELETE on compressed hypertables, we don't support this.

Fixes #5139

### DIFF
--- a/src/nodes/chunk_dispatch/chunk_dispatch.c
+++ b/src/nodes/chunk_dispatch/chunk_dispatch.c
@@ -11,6 +11,7 @@
 #include <nodes/nodeFuncs.h>
 #include <parser/parsetree.h>
 #include <utils/rel.h>
+#include <utils/syscache.h>
 #include <catalog/pg_type.h>
 
 #include "compat/compat.h"
@@ -20,6 +21,7 @@
 #include "subspace_store.h"
 #include "dimension.h"
 #include "guc.h"
+#include "nodes/hypertable_modify.h"
 #include "ts_catalog/chunk_data_node.h"
 
 static Node *chunk_dispatch_state_create(CustomScan *cscan);
@@ -244,6 +246,15 @@ chunk_dispatch_plan_create(PlannerInfo *root, RelOptInfo *relopt, CustomPath *be
 	cscan->custom_scan_tlist = tlist;
 	cscan->scan.plan.targetlist = tlist;
 
+#if PG15_GE
+	if (root->parse->mergeUseOuterJoin)
+	{
+		/* replace expressions of ROWID_VAR */
+		tlist = ts_replace_rowid_vars(root, tlist, relopt->relid);
+		cscan->scan.plan.targetlist = tlist;
+		cscan->custom_scan_tlist = tlist;
+	}
+#endif
 	return &cscan->scan.plan;
 }
 
@@ -339,8 +350,61 @@ chunk_dispatch_exec(CustomScanState *node)
 	/* Switch to the executor's per-tuple memory context */
 	old = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
 
+#if PG15_GE
+	TupleTableSlot *newslot = NULL;
+	if (dispatch->dispatch_state->mtstate->operation == CMD_MERGE)
+	{
+		HeapTuple tp;
+		AttrNumber natts;
+		AttrNumber attno;
+
+		tp = SearchSysCache1(RELOID, ObjectIdGetDatum(ht->main_table_relid));
+		if (!HeapTupleIsValid(tp))
+			elog(ERROR, "cache lookup failed for relation %u", ht->main_table_relid);
+		natts = ((Form_pg_class) GETSTRUCT(tp))->relnatts;
+		ReleaseSysCache(tp);
+		for (attno = 1; attno <= natts; attno++)
+		{
+			tp = SearchSysCache2(ATTNUM,
+								 ObjectIdGetDatum(ht->main_table_relid),
+								 Int16GetDatum(attno));
+			if (!HeapTupleIsValid(tp))
+				continue;
+			Form_pg_attribute att_tup = (Form_pg_attribute) GETSTRUCT(tp);
+			ReleaseSysCache(tp);
+			if (att_tup->attisdropped || att_tup->atthasmissing)
+			{
+				state->is_dropped_attr_exists = true;
+				continue;
+			}
+		}
+		for (int i = 0; i < ht->space->num_dimensions; i++)
+		{
+			List *actionStates =
+				dispatch->dispatch_state->mtstate->resultRelInfo->ri_notMatchedMergeAction;
+			ListCell *l;
+			foreach (l, actionStates)
+			{
+				MergeActionState *action = (MergeActionState *) lfirst(l);
+				CmdType commandType = action->mas_action->commandType;
+				if (commandType == CMD_INSERT)
+				{
+					/* fetch full projection list */
+					action->mas_proj->pi_exprContext->ecxt_innertuple = slot;
+					newslot = ExecProject(action->mas_proj);
+					break;
+				}
+			}
+			if (newslot)
+				break;
+		}
+	}
 	/* Calculate the tuple's point in the N-dimensional hyperspace */
+	point = ts_hyperspace_calculate_point(ht->space, (newslot ? newslot : slot));
+
+#else
 	point = ts_hyperspace_calculate_point(ht->space, slot);
+#endif
 
 	/* Save the main table's (hypertable's) ResultRelInfo */
 	if (!dispatch->hypertable_result_rel_info)
@@ -375,7 +439,7 @@ chunk_dispatch_exec(CustomScanState *node)
 	MemoryContextSwitchTo(old);
 
 	/* Convert the tuple to the chunk's rowtype, if necessary */
-	if (cis->hyper_to_chunk_map != NULL)
+	if (cis->hyper_to_chunk_map != NULL && state->is_dropped_attr_exists == false)
 		slot = execute_attr_map_slot(cis->hyper_to_chunk_map->attrMap, slot, cis->slot);
 
 	return slot;
--- a/src/nodes/chunk_dispatch/chunk_dispatch.h
+++ b/src/nodes/chunk_dispatch/chunk_dispatch.h
@@ -72,6 +72,8 @@ typedef struct ChunkDispatchState
 	 */
 	ChunkDispatch *dispatch;
 	ResultRelInfo *rri;
+	/* flag to represent dropped attributes */
+	bool is_dropped_attr_exists;
 } ChunkDispatchState;
 
 extern TSDLLEXPORT bool ts_is_chunk_dispatch_state(PlanState *state);
--- a/src/partitioning.c
+++ b/src/partitioning.c
@@ -260,6 +260,13 @@ ts_partitioning_func_apply(PartitioningInfo *pinfo, Oid collation, Datum value)
 	return result;
 }
 
+/*
+ * Helper function to find the right partition value from a tuple,
+ * for space partitioned hypertables. Since attributes in tuple can
+ * be of different order when compared to physical table columns order,
+ * we pass partition_col_idx which points to correct space parititioned
+ * column in the given tuple.
+ */
 TSDLLEXPORT Datum
 ts_partitioning_func_apply_slot(PartitioningInfo *pinfo, TupleTableSlot *slot, bool *isnull)
 {
--- a/test/expected/merge.out
+++ b/test/expected/merge.out
@@ -1,22 +1,24 @@
 -- This file and its contents are licensed under the Apache License 2.0.
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-APACHE for a copy of the license.
--- Create conditions table with location and temperature
-CREATE TABLE conditions (
+\c :TEST_DBNAME :ROLE_SUPERUSER
+-- Create target table with location and temperature
+CREATE TABLE target (
    time        TIMESTAMPTZ       NOT NULL,
    location    SMALLINT          NOT NULL,
-   temperature DOUBLE PRECISION  NULL
+   temperature DOUBLE PRECISION  NULL,
+   val text default 'string -'
 );
 SELECT create_hypertable(
-  'conditions',
+  'target',
   'time',
   chunk_time_interval => INTERVAL '5 seconds');
-    create_hypertable    
--------------------------
- (1,public,conditions,t)
+  create_hypertable  
+---------------------
+ (1,public,target,t)
 (1 row)
 
-INSERT INTO conditions
+INSERT INTO target
 SELECT time, location, 14 as temperature
 FROM generate_series(
 	'2021-01-01 00:00:00',
@@ -24,23 +26,23 @@ FROM generate_series(
     INTERVAL '5 seconds'
   ) as time,
 generate_series(1,4) as location;
--- Create conditions_updated table with location and temperature
-CREATE TABLE conditions_updated (
+-- Create source table with location and temperature
+CREATE TABLE source (
    time        TIMESTAMPTZ       NOT NULL,
    location    SMALLINT          NOT NULL,
    temperature DOUBLE PRECISION  NULL
 );
 SELECT create_hypertable(
-  'conditions_updated',
+  'source',
   'time',
   chunk_time_interval => INTERVAL '5 seconds');
-        create_hypertable        
----------------------------------
- (2,public,conditions_updated,t)
+  create_hypertable  
+---------------------
+ (2,public,source,t)
 (1 row)
 
--- Generate data that overlaps with conditions table
-INSERT INTO conditions_updated
+-- Generate data that overlaps with target table
+INSERT INTO source
 SELECT time, location, 80 as temperature
 FROM generate_series(
 	'2021-01-01 00:00:05',
@@ -49,20 +51,20 @@ FROM generate_series(
   ) as time,
 generate_series(1,4) as location;
 -- Print table/rows/num of chunks
-select * from conditions order by time, location asc;
-             time             | location | temperature 
-------------------------------+----------+-------------
- Fri Jan 01 00:00:00 2021 PST |        1 |          14
- Fri Jan 01 00:00:00 2021 PST |        2 |          14
- Fri Jan 01 00:00:00 2021 PST |        3 |          14
- Fri Jan 01 00:00:00 2021 PST |        4 |          14
- Fri Jan 01 00:00:05 2021 PST |        1 |          14
- Fri Jan 01 00:00:05 2021 PST |        2 |          14
- Fri Jan 01 00:00:05 2021 PST |        3 |          14
- Fri Jan 01 00:00:05 2021 PST |        4 |          14
+select * from target order by time, location asc;
+             time             | location | temperature |   val    
+------------------------------+----------+-------------+----------
+ Fri Jan 01 00:00:00 2021 PST |        1 |          14 | string -
+ Fri Jan 01 00:00:00 2021 PST |        2 |          14 | string -
+ Fri Jan 01 00:00:00 2021 PST |        3 |          14 | string -
+ Fri Jan 01 00:00:00 2021 PST |        4 |          14 | string -
+ Fri Jan 01 00:00:05 2021 PST |        1 |          14 | string -
+ Fri Jan 01 00:00:05 2021 PST |        2 |          14 | string -
+ Fri Jan 01 00:00:05 2021 PST |        3 |          14 | string -
+ Fri Jan 01 00:00:05 2021 PST |        4 |          14 | string -
 (8 rows)
 
-select * from conditions_updated order by time, location asc;
+select * from source order by time, location asc;
              time             | location | temperature 
 ------------------------------+----------+-------------
  Fri Jan 01 00:00:05 2021 PST |        1 |          80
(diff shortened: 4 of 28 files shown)


svenklemm (commented):
This needs test for hypertables with chunks with different physical layout you can look in the regresscheck-shared test for how to set this up.

akuzm:
>    Refactor ExecInsert/Delete/Update
>    Backported commit 25e777cf8e547d7423d2e1e9da71f98b9414d59e
>    Backport all MERGE related interfaces and its implementations.
>    Backported commit 7103ebb7aae8ab8076b7e85f335ceb8fe799097c
    
Can't find these commits in the main repo or in your fork, what are they?

sb230132:
> 

These commits you will find in postgresql master branch. Since these interfaces are not exported, i need to backport to our source.

```
commit 25e777cf8e547d7423d2e1e9da71f98b9414d59e
Author: Alvaro Herrera <alvherre@alvh.no-ip.org>
Date:   Thu Mar 17 11:47:04 2022 +0100

    Split ExecUpdate and ExecDelete into reusable pieces

    Create subroutines ExecUpdatePrologue / ExecUpdateAct /
    ExecUpdateEpilogue, and similar for ExecDelete.

    Introduce a new struct to be used internally in nodeModifyTable.c,
    dubbed ModifyTableContext, which contains all context information needed
    to perform these operations, as well as ExecInsert and others.

    This allows using a different schedule and a different way of evaluating
    the results of these operations, which can be exploited by a later
    commit introducing support for MERGE.  It also makes ExecUpdate and
    ExecDelete proper shorter and (hopefully) simpler.

    Author: Álvaro Herrera <alvherre@alvh.no-ip.org>
    Reviewed-by: Amit Langote <amitlangote09@gmail.com>
    Reviewed-by: Japin Li <japinli@hotmail.com>
    Reviewed-by: Zhihong Yu <zyu@yugabyte.com>
    Discussion: https://postgr.es/m/202202271724.4z7xv3cf46kv@alvherre.pgsql


commit 7103ebb7aae8ab8076b7e85f335ceb8fe799097c
Author: Alvaro Herrera <alvherre@alvh.no-ip.org>
Date:   Mon Mar 28 16:45:58 2022 +0200

    Add support for MERGE SQL command

    MERGE performs actions that modify rows in the target table using a
    source table or query. MERGE provides a single SQL statement that can
    conditionally INSERT/UPDATE/DELETE rows -- a task that would otherwise
    require multiple PL statements.  For example,

    MERGE INTO target AS t
    USING source AS s
    ON t.tid = s.sid
    WHEN MATCHED AND t.balance > s.delta THEN
      UPDATE SET balance = t.balance - s.delta
    WHEN MATCHED THEN
      DELETE
    WHEN NOT MATCHED AND s.delta > 0 THEN
      INSERT VALUES (s.sid, s.delta)
    WHEN NOT MATCHED THEN
      DO NOTHING;
```

akuzm:
It's becoming really hard to follow which part we are copying from Postgres and what we are modifying. What's worse, we now even have mixed code from different versions of Postgres. So I would suggest this:

1) create a new file in src/import, that would contain all the functions of nodeModifyTable copied from postgres.
2) move there the copied code from modify_hypertable.c
3) disable clang-format for this file with /* clang-format off */ in the beginning
4) run pgindent on it like described here https://www.interdb.jp/blog/pgsql/pgindent/
5) rename functions by e.g. adding a prefix like `ht_`, so that we don't confuse them with the original ones.
6) mention which versions of postgres they are copied from.

This way, at least we will be able to compare to postgres code with a diff tool, and tell which changes we made ourselves and which were made by the upstream.

I think it might be easier to start with the code that we currently have there, w/o the code required for merge. We could merge this separately, and then add the MERGE code on top of it.  

Do you think it's feasible?

akuzm:
Also the test coverage still seems to be lacking, even after I fixed the bug in coverage collection. If you open the "files" tab, you can see the comments from the codecov bot about the missing coverage for some lines. Can you please investigate it?

akuzm:
> I think it might be easier to start with the code that we currently have there, w/o the code required for merge. We could merge this separately, and then add the MERGE code on top of it.

Or maybe we can do this just for the new code you are adding. It would be still somewhat confusing, but better than now, and less work than moving everything.

sb230132:
> It's becoming really hard to follow which part we are copying from Postgres and what we are modifying. What's worse, we now even have mixed code from different versions of Postgres. So I would suggest this:
> 
> 1. create a new file in src/import, that would contain all the functions of nodeModifyTable copied from postgres.
> 2. move there the copied code from modify_hypertable.c
> 3. disable clang-format for this file with /* clang-format off */ in the beginning
> 4. run pgindent on it like described here https://www.interdb.jp/blog/pgsql/pgindent/
> 5. rename functions by e.g. adding a prefix like `ht_`, so that we don't confuse them with the original ones.
> 6. mention which versions of postgres they are copied from.
> 
> This way, at least we will be able to compare to postgres code with a diff tool, and tell which changes we made ourselves and which were made by the upstream.
> 
> I think it might be easier to start with the code that we currently have there, w/o the code required for merge. We could merge this separately, and then add the MERGE code on top of it.
> 
> Do you think it's feasible?

Ok i will create a new PR where i will only include all the code ported from postgresql as suggested.

sb230132:
> 4\. https://www.interdb.jp/blog/pgsql/pgindent/

here is the [PR](https://github.com/timescale/timescaledb/pull/5335) with MERGE specific postgresql code.

sb230132:
> Also the test coverage still seems to be lacking, even after I fixed the bug in coverage collection. If you open the "files" tab, you can see the comments from the codecov bot about the missing coverage for some lines. Can you please investigate it?

Right. Fixed it by adding more tests. Now i get 91.88% coverage.

merge_code_coverage

src/partitioning.c
```diff
@@ -261,13 +261,17 @@ ts_partitioning_func_apply(PartitioningInfo *pinfo, Oid collation, Datum value)
...
+ts_partitioning_func_apply_slot(PartitioningInfo *pinfo, TupleTableSlot *slot, bool *isnull,
+								AttrNumber partition_col_idx)
 {
 	Datum value;
 	bool null;
 	Oid collation;
 
-	value = slot_getattr(slot, pinfo->column_attnum, &null);
+	if (partition_col_idx != pinfo->column_attnum)
+		value = slot_getattr(slot, partition_col_idx, &null);
```
akuzm, line 272: This line also works for the `else` case, because they are equal, right? So we can remove the `if` and keep this line only.

src/partitioning.c
```diff
@@ -261,13 +261,17 @@ ts_partitioning_func_apply(PartitioningInfo *pinfo, Oid collation, Datum value)
 }
 
 TSDLLEXPORT Datum
-ts_partitioning_func_apply_slot(PartitioningInfo *pinfo, TupleTableSlot *slot, bool *isnull)
+ts_partitioning_func_apply_slot(PartitioningInfo *pinfo, TupleTableSlot *slot, bool *isnull,
+								AttrNumber partition_col_idx)
```
akuzm, line 265: Might be good to add a small comment that mentions that this function is applied to chunk tuples also, which might have different column order, and that's why we pass the `partition_col_idx` separately instead of the one in `pinfo`.

src/partitioning.c
sb230132: Right, fixed.

src/partitioning.c
sb230132: Added comment.

src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -290,8 +362,40 @@ chunk_dispatch_exec(CustomScanState *node)
...
+		if (!OidIsValid(partition_col_idx))
+		{
+			List *actionStates =
+				dispatch->dispatch_state->mtstate->resultRelInfo->ri_notMatchedMergeAction;
+			ListCell *l;
+			foreach (l, actionStates)
+			{
+				MergeActionState *action = (MergeActionState *) lfirst(l);
+				CmdType commandType = action->mas_action->commandType;
+				if (commandType == CMD_INSERT)
```
akuzm, line 382: We should only have CMD_INSERT here, because we don't insert the chunk dispatch node otherwise, right?

src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -290,8 +362,40 @@ chunk_dispatch_exec(CustomScanState *node)
...
+				}
+			}
+		}
+	}
 	/* Calculate the tuple's point in the N-dimensional hyperspace */
-	point = ts_hyperspace_calculate_point(ht->space, slot);
+	if (!OidIsValid(partition_col_idx) && newslot)
+		point = ts_hyperspace_calculate_point(ht->space, newslot, NULL);
+	else
+		point = ts_hyperspace_calculate_point(ht->space, slot, state->colnames_list);
```
akuzm, line 395: What is the slot here, is it from some chunk? So this would be the case of `WHEN MATCHED THEN INSERT`. It's a little cumbersome that we now have special code for tuple conversion in the partitioning functions. I wonder if we could determine the chunk reloid here, fetch the matching chunk insert state, and then use the tuple conversion map like `hyper_to_chunk_map` to fetch the partitioning columns.

src/nodes/chunk_dispatch/chunk_dispatch.h
```diff
@@ -72,6 +72,17 @@ typedef struct ChunkDispatchState
...
+	 * Save attributes of targetList to calculate correct attribute no: in
+	 * case tuple has a different order incase of MERGE with INSERT
+	 */
+	List *colnames_list;
+	/*
+	 * Save dropped attributes to be used during execution to map
+	 * attributes numbers between different tuples.
+	 */
+	List *dropped_attrList;
+	List *chunk_resultrelInfoList;
```
akuzm, line 85: Can't find other usages of this one, is it needed?

src/nodes/chunk_dispatch/chunk_dispatch.c
sb230132: ChunkDispatch node is inserted when INSERT action is specified in MERGE, however it does not mean that MERGE cannot have other actions like UPDATE/DELETE. Thus we check for commandType.

src/nodes/chunk_dispatch/chunk_dispatch.h
sb230132: dropped_attrList is needed. I removed chunk_resultrelInfoList.

src/nodes/chunk_dispatch/chunk_dispatch.c
sb230132: Here slot is the tuple which ChunkDispatch node receives from its child node. This is the tuple which needs to be inserted into the target table. However if you notice from below plan output, you will see that we don't have fully projected tuple. Received slot has only 2 attributes (Output: s.temperature, t.tableoid, t.ctid). From this slot we cannot get the partitioned column value for us to decide in which chunk this tuple should be inserted into. Thus i call ExecProject() if calcuated partition slot number is invalid. Iam not doing any mapping here.

```
postgres=# explain (costs off, verbose) MERGE INTO target t
MERGE INTO target t
USING source s
ON t.time = s.time AND t.location = s.location
WHEN MATCHED AND t.location = 560076 THEN
 UPDATE SET temperature = (t.temperature + s.temperature) * 2, val = val || ' UPDATED BY MERGE'
WHEN MATCHED AND t.location = 560083 THEN
 DELETE
WHEN NOT MATCHED THEN
 INSERT VALUES ('2021-11-01 00:00:05'::timestamp with time zone, 5, 210, 'string - INSERTED BY MERGE');
                                      QUERY PLAN
------------------------------------------------------------------------------------------------
 Custom Scan (HypertableModify)
   ->  Merge on public.target t
         Merge on public.target t_1
         Merge on _timescaledb_internal._hyper_20_39_chunk t_2
         Merge on _timescaledb_internal._hyper_20_40_chunk t_3
         ->  Custom Scan (ChunkDispatch)
               ->  Merge Left Join
                     Output: s.temperature, t.tableoid, t.ctid
                     Merge Cond: ((s."time" = t."time") AND (s.location = t.location))
                     ->  Sort
                           Output: s.temperature, s."time", s.location
                           Sort Key: s."time", s.location
                           ->  Seq Scan on public.source s
                                 Output: s.temperature, s."time", s.location
                     ->  Sort
                           Output: t."time", t.location, t.tableoid, t.ctid
                           Sort Key: t."time", t.location
                           ->  Append
                                 ->  Seq Scan on public.target t_1
                                       Output: t_1."time", t_1.location, t_1.tableoid, t_1.ctid
                                 ->  Seq Scan on _timescaledb_internal._hyper_20_39_chunk t_2
                                       Output: t_2."time", t_2.location, t_2.tableoid, t_2.ctid
                                 ->  Seq Scan on _timescaledb_internal._hyper_20_40_chunk t_3
                                       Output: t_3."time", t_3.location, t_3.tableoid, t_3.ctid
(24 rows)
```

src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -185,16 +187,86 @@ chunk_dispatch_plan_create(PlannerInfo *root, RelOptInfo *relopt, CustomPath *be
 		cscan->scan.plan.plan_rows += subplan->plan_rows;
 		cscan->scan.plan.plan_width += subplan->plan_width;
 	}
-
-	cscan->custom_private = list_make1_oid(cdpath->hypertable_relid);
+	cscan->custom_private = lappend(cscan->custom_private, makeInteger(cdpath->hypertable_relid));
```
svenklemm, line 190: Why are we using Value nodes here instead of plain values?

sb230132: I cant use lappend_int() as it will create List of type T_IntList. And later when i append other values, i get assert.
Want to want is to be able to append relid + few other values. To achieve this i first convert relid as Integer node and save in the custom_private list.

How else can i store unsigned int and other values in a List ?

src/dimension.c
```diff
@@ -957,8 +957,84 @@ ts_point_create(int16 num_dimensions)
...
+
+/*
+ * Helper function to find the right partition value from a tuple,
+ * in case attributes in tuple are not ordered correctly.
+ *
+ * Assume columns in table are in below order:
+ *  [time, temperature, location, value]
+ * "time" is the partition column based on this columns value
+ * ChunkDispatch node decides if to create or reuse an existing
+ * chunk.
```
svenklemm, line 996: I think we need to have a test with the specific configuration you mention here.

src/dimension.c
```diff
@@ -957,8 +957,84 @@ ts_point_create(int16 num_dimensions)
...
+int
+get_partition_attrno_from_tuple(const Hyperspace *hs, TupleTableSlot *slot, List *colnames_list,
+								bool is_space_partition)
+{
+	AttrNumber partition_col_idx = 0;
+	Dimension *d;
+	if (is_space_partition)
+		d = (Dimension *) &hs->dimensions[1];
+	else
+		d = (Dimension *) &hs->dimensions[0];
```
svenklemm, line 1013: This will break with multiple space partition or when the the 2nd dimension is not the space dimension.

src/dimension.c
```diff
@@ -970,10 +1046,18 @@ ts_hyperspace_calculate_point(const Hyperspace *hs, TupleTableSlot *slot)
 		bool isnull;
 		Oid dimtype;
 
+		AttrNumber partition_col_idx = d->column_attno;
+		if (colnames_list)
+			partition_col_idx = get_partition_attrno_from_tuple(hs,
+																slot,
+																colnames_list,
+																(d->partitioning ? true : false));
```
svenklemm, line 1054: This is not the correct way to determine the dimension type. There are two macros IS_OPEN_DIMENSION and IS_CLOSED_DIMENSION to determine dimension type.

src/nodes/chunk_dispatch/chunk_dispatch.h
```diff
@@ -72,6 +72,16 @@ typedef struct ChunkDispatchState
...
+	/*
+	 * Save attributes of targetList to calculate correct attribute no: in
+	 * case tuple has a different order incase of MERGE with INSERT
+	 */
+	List *colnames_list;
+	/*
+	 * Save dropped attributes to be used during execution to map
+	 * attributes numbers between different tuples.
+	 */
+	List *dropped_attrList;
```
svenklemm, line 84: It shouldnt be necessary to pass down the column list from planner to executor. Is there no other way to deal with this?

src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -290,8 +362,40 @@ chunk_dispatch_exec(CustomScanState *node)
...
+				dispatch->dispatch_state->mtstate->resultRelInfo->ri_notMatchedMergeAction;
+			ListCell *l;
+			foreach (l, actionStates)
+			{
+				MergeActionState *action = (MergeActionState *) lfirst(l);
+				CmdType commandType = action->mas_action->commandType;
+				if (commandType == CMD_INSERT)
+				{
+					/* in case received tuple has only junk attributes, then
+					 * fetch the actual values. */
```
svenklemm, line 385: why would the received tuple only have junk attributes?

sb230132: Junk attributes, means system column attributes.
If you notice in below plan the tuple recieved by ChunkDispatch has only 2 attributes which are  t.tableoid, t.ctid
From this tuple i cannot know the dimension value. I need to call ExecProject() to get full row and the get the dimension value.

```
postgres=# explain (costs off, verbose) MERGE INTO target t
USING source s
ON t.location = 1234
WHEN NOT MATCHED THEN
INSERT VALUES ('2021-11-01 00:00:05'::timestamp with time zone, 5, 210, 'string - INSERTED BY MERGE');
                                          QUERY PLAN
----------------------------------------------------------------------------------------------
 Custom Scan (HypertableModify)
   ->  Merge on public.target t
         Merge on public.target t_1
         Merge on _timescaledb_internal._hyper_38_13_chunk t_2
         Merge on _timescaledb_internal._hyper_38_14_chunk t_3
         ->  Custom Scan (ChunkDispatch)
               ->  Nested Loop Left Join
                     Output: t.tableoid, t.ctid
                     ->  Append
                           ->  Seq Scan on public.source s_1
                           ->  Seq Scan on _timescaledb_internal._hyper_39_15_chunk s_2
                           ->  Seq Scan on _timescaledb_internal._hyper_39_16_chunk s_3
                     ->  Materialize
                           Output: t.tableoid, t.ctid
                           ->  Append
                                 ->  Seq Scan on public.target t_1
                                       Output: t_1.tableoid, t_1.ctid
                                       Filter: (t_1.location = 1234)
                                 ->  Seq Scan on _timescaledb_internal._hyper_38_13_chunk t_2
                                       Output: t_2.tableoid, t_2.ctid
                                       Filter: (t_2.location = 1234)
                                 ->  Seq Scan on _timescaledb_internal._hyper_38_14_chunk t_3
                                       Output: t_3.tableoid, t_3.ctid
                                       Filter: (t_3.location = 1234)
(24 rows)
```

src/dimension.c
sb230132: I do have a testcase for this in merge.sql
Please notice the column order in table vs the column order in the tuple received from child node of ChunkDispatch.

```
postgres=# \d target
                              Table "public.target"
   Column    |           Type           | Collation | Nullable |     Default
-------------+--------------------------+-----------+----------+------------------
 time        | timestamp with time zone |           | not null |
 location    | smallint                 |           | not null |
 temperature | double precision         |           |          |
 val         | text                     |           |          | 'string -'::text
Indexes:
    "target_time_idx" btree ("time" DESC)
Triggers:
    ts_insert_blocker BEFORE INSERT ON target FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker()
Number of child tables: 2 (Use \d+ to list them.)

postgres=#
postgres=# explain (costs off, verbose) MERGE INTO target t                                                                                                                                                                                USING source s                                                                                                                                                                                                                             ON t.time = s.time AND t.location = s.location                                                                                                                                                                                             WHEN MATCHED THEN                                                                                                                                                                                                                          UPDATE SET temperature = (t.temperature + s.temperature)/2, val = val || ' UPDATED BY MERGE'                                                                                                                                               WHEN NOT MATCHED THEN                                                                                                                                                                                                                      INSERT (time, location, temperature, val) VALUES (s.time, s.location, s.temperature, 'string - INSERTED BY MERGE');
                                           QUERY PLAN
------------------------------------------------------------------------------------------------
 Custom Scan (HypertableModify)
   ->  Merge on public.target t
         Merge on public.target t_1
         Merge on _timescaledb_internal._hyper_38_13_chunk t_2
         Merge on _timescaledb_internal._hyper_38_14_chunk t_3
         ->  Custom Scan (ChunkDispatch)
               ->  Merge Left Join
                     Output: s.temperature, s."time", s.location, t.tableoid, t.ctid
                     Merge Cond: ((s."time" = t."time") AND (s.location = t.location))
                     ->  Sort
                           Output: s.temperature, s."time", s.location
                           Sort Key: s."time", s.location
                           ->  Append
                                 ->  Seq Scan on public.source s_1
                                       Output: s_1.temperature, s_1."time", s_1.location
                                 ->  Seq Scan on _timescaledb_internal._hyper_39_15_chunk s_2
                                       Output: s_2.temperature, s_2."time", s_2.location
                                 ->  Seq Scan on _timescaledb_internal._hyper_39_16_chunk s_3
                                       Output: s_3.temperature, s_3."time", s_3.location
                     ->  Sort
                           Output: t."time", t.location, t.tableoid, t.ctid
                           Sort Key: t."time", t.location
                           ->  Append
                                 ->  Seq Scan on public.target t_1
                                       Output: t_1."time", t_1.location, t_1.tableoid, t_1.ctid
                                 ->  Seq Scan on _timescaledb_internal._hyper_38_13_chunk t_2
                                       Output: t_2."time", t_2.location, t_2.tableoid, t_2.ctid
                                 ->  Seq Scan on _timescaledb_internal._hyper_38_14_chunk t_3
                                       Output: t_3."time", t_3.location, t_3.tableoid, t_3.ctid
(29 rows)
```

src/dimension.c
sb230132: Here iam not determining dimension type. 
For a given tuple received from subplan of ChunkDispatch, I need to know the position where time or space dimension column is located. get_partition_attrno_from_tuple() will return the slot number where the time or space dimension column is present in the tuple.

src/dimension.c
sb230132: I do have a testcase for this in merge.sql
Please notice the column order in table vs the column order in the tuple received from child node of ChunkDispatch.

```
postgres=# \d target
                              Table "public.target"
   Column    |           Type           | Collation | Nullable |     Default
-------------+--------------------------+-----------+----------+------------------
 time        | timestamp with time zone |           | not null |
 location    | smallint                 |           | not null |
 temperature | double precision         |           |          |
 val         | text                     |           |          | 'string -'::text
Indexes:
    "target_time_idx" btree ("time" DESC)
Triggers:
    ts_insert_blocker BEFORE INSERT ON target FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker()
Number of child tables: 2 (Use \d+ to list them.)

postgres=#
postgres=# explain (costs off, verbose) MERGE INTO target t                                                                                                                                                                                USING source s                                                                                                                                                                                                                             ON t.time = s.time AND t.location = s.location                                                                                                                                                                                             WHEN MATCHED THEN                                                                                                                                                                                                                          UPDATE SET temperature = (t.temperature + s.temperature)/2, val = val || ' UPDATED BY MERGE'                                                                                                                                               WHEN NOT MATCHED THEN                                                                                                                                                                                                                      INSERT (time, location, temperature, val) VALUES (s.time, s.location, s.temperature, 'string - INSERTED BY MERGE');
                                           QUERY PLAN
------------------------------------------------------------------------------------------------
 Custom Scan (HypertableModify)
   ->  Merge on public.target t
         Merge on public.target t_1
         Merge on _timescaledb_internal._hyper_38_13_chunk t_2
         Merge on _timescaledb_internal._hyper_38_14_chunk t_3
         ->  Custom Scan (ChunkDispatch)
               ->  Merge Left Join
                     Output: s.temperature, s."time", s.location, t.tableoid, t.ctid
                     Merge Cond: ((s."time" = t."time") AND (s.location = t.location))
                     ->  Sort
                           Output: s.temperature, s."time", s.location
                           Sort Key: s."time", s.location
                           ->  Append
                                 ->  Seq Scan on public.source s_1
                                       Output: s_1.temperature, s_1."time", s_1.location
                                 ->  Seq Scan on _timescaledb_internal._hyper_39_15_chunk s_2
                                       Output: s_2.temperature, s_2."time", s_2.location
                                 ->  Seq Scan on _timescaledb_internal._hyper_39_16_chunk s_3
                                       Output: s_3.temperature, s_3."time", s_3.location
                     ->  Sort
                           Output: t."time", t.location, t.tableoid, t.ctid
                           Sort Key: t."time", t.location
                           ->  Append
                                 ->  Seq Scan on public.target t_1
                                       Output: t_1."time", t_1.location, t_1.tableoid, t_1.ctid
                                 ->  Seq Scan on _timescaledb_internal._hyper_38_13_chunk t_2
                                       Output: t_2."time", t_2.location, t_2.tableoid, t_2.ctid
                                 ->  Seq Scan on _timescaledb_internal._hyper_38_14_chunk t_3
                                       Output: t_3."time", t_3.location, t_3.tableoid, t_3.ctid
(29 rows)
```

src/dimension.c
sb230132: Right. I fixed it now.

src/nodes/chunk_dispatch/chunk_dispatch.h
sb230132: I need actual physical layout of table to determine the correct values of dimension columns from the received tuple especially when columns order in tuple received from child node of ChunkDispatch node is different from actual physical ordering.

src/nodes/chunk_dispatch/chunk_dispatch.c
sb230132: I think iam already doing it.
If you check in chunk_dispatch_plan_create() i have below code.

```
#if PG15_GE
	if (root->parse->mergeUseOuterJoin)
	{
		/* replace expressions of ROWID_VAR */
		tlist = ts_replace_rowid_vars(root, tlist, relopt->relid);
		cscan->scan.plan.targetlist = tlist;
		cscan->custom_scan_tlist = tlist;
	}
#endif
```

pmwkaa (commented):
I've looked through the changes and it looked good to me (you can ignore my comments), unfortunately I don't have a deep knowledge about our planner code.

Testing against normal PG table and hypertable was a good idea. In order to move forward with this PR what I think we should do is to integrate as much tests as possible from the original PostgreSQL merge command tests and try to them with hypertable: https://github.com/postgres/postgres/blob/master/src/test/regress/sql/merge.sql. 
This is might be done to some extent, but this will allow us to know if we missed any important cases and tell the level of comparability with the regular PG table (which ideally should be 1 to 1).

Ideally we should run original merge tests for PG table and hypertable test to compare diff between two files output. Can we run it as a template by changing destination table in each test? I believe we did something like this before with multi-node when tested different planner outputs, etc.

src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -322,8 +332,76 @@ chunk_dispatch_exec(CustomScanState *node)
...
+			if (!HeapTupleIsValid(tp))
+				continue;
+			Form_pg_attribute att_tup = (Form_pg_attribute) GETSTRUCT(tp);
+			ReleaseSysCache(tp);
+			if (att_tup->attisdropped || att_tup->atthasmissing)
+			{
+				state->is_dropped_attr_exists = true;
+				continue;
+			}
+			colnames_list = lappend(colnames_list, pstrdup(NameStr(att_tup->attname)));
```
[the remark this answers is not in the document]
sb230132: I don't see this piece of code being reused anywhere. Keeping it as is for now.

src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -322,8 +332,76 @@ chunk_dispatch_exec(CustomScanState *node)
...
+		AttrNumber attno;
+
+		tp = SearchSysCache1(RELOID, ObjectIdGetDatum(ht->main_table_relid));
+		if (!HeapTupleIsValid(tp))
+			elog(ERROR, "cache lookup failed for relation %u", ht->main_table_relid);
+		natts = ((Form_pg_class) GETSTRUCT(tp))->relnatts;
+		ReleaseSysCache(tp);
+		for (attno = 1; attno <= natts; attno++)
+		{
+			tp = SearchSysCache2(ATTNUM,
```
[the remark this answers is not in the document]
sb230132: I need to know if any attribute is dropped or not. Thus cant use SearchSysCacheAttNum

sb230132:
> I've looked through the changes and it looked good to me (you can ignore my comments). Unfortunately I don't have a deep knowledge about our planner code to tell if we missed something.
> 
> Testing against a normal PG table and hypertable was a good idea.
> 
> In order to move forward with this PR what I think we should do is to integrate as much tests as possible from the original PostgreSQL merge command tests and try to them with hypertable: https://github.com/postgres/postgres/blob/master/src/test/regress/sql/merge.sql. This is might be done to some extent, but this will allow us to know if we missed any important cases and tell the level of comparability with the regular PG table (which ideally should be 1 to 1).
> 
> Ideally we should run original merge tests for PG table and hypertable test to compare diff between two files output. Can we run it as a template by changing destination table in each test? I believe we did something like this before with multi-node when tested different planner outputs, etc.

Added ts_merge.sql which will run the entire test with/without hypertables and compare diffs.
I took merge.sql from postgres source code, renamed it to ts_merge.sql and did few modifications to make the test run on timescale.
Had to skip a few tests as it had EXPLAIN and also removed few -ve cases as it was giving different error message for pg table and hypertables. 
Iam skipping this test on windows as there seem to be a different behaviour of MERGE on windows vs linux.

src/partitioning.c
```diff
@@ -260,6 +260,13 @@ ts_partitioning_func_apply(PartitioningInfo *pinfo, Oid collation, Datum value)
 	return result;
 }
 
+/*
+ * Helper function to find the right partition value from a tuple,
+ * for space partitioned hypertables. Since attributes in tuple can
+ * be of different ordere when compared to physical table columns order,
```
fabriziomello, line 266:
```suggestion
 * be of different order when compared to physical table columns order,
```

test/sql/include/ts_merge_query.sql
```diff
@@ -0,0 +1,1248 @@
...
+DROP FUNCTION measurement_insert_trigger();
+
+RESET SESSION AUTHORIZATION;
+DROP TABLE target CASCADE;
+DROP TABLE target2 CASCADE;
+DROP TABLE source, source2;
+DROP FUNCTION merge_trigfunc();
+REVOKE CREATE ON SCHEMA public FROM regress_merge_privs;
+DROP USER regress_merge_privs;
+DROP USER regress_merge_no_privs;
```
fabriziomello, line 1248: Missing EOF new empty line

test/expected/merge.out
```diff
@@ -1,46 +1,48 @@
 -- This file and its contents are licensed under the Apache License 2.0.
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-APACHE for a copy of the license.
--- Create conditions table with location and temperature
-CREATE TABLE conditions (
+\c :TEST_DBNAME :ROLE_SUPERUSER
```
fabriziomello, line 4: What about testing the MERGE commands without a SUPERUSER?? Just because on cloud we do not give a SUPERUSER for our customers.

fabriziomello (approved):
Please just add the changelog to the new .unreleased folder

test/sql/include/ts_merge_load.sql
```diff
@@ -0,0 +1,64 @@
+-- This file and its contents are licensed under the Apache License 2.0.
+-- Please see the included NOTICE for copyright information and
+-- LICENSE-APACHE for a copy of the license.
+
+CREATE USER regress_merge_privs;
```
pmwkaa, line 5: Hm, I am not sure it is safe to create users inside tests this way since users are shared between databases. For instance, will you be able to run same test one after another if same user already exists?

pmwkaa (approved):

test/sql/include/ts_merge_load.sql
sb230132: We do drop the users in ts_merge_query.sql

Issue #5140, opened 2023-01-03
add optional "reference" functionality in "create_hypertable"


nikkhils:
We need to be able to create special type of distributed hypertables which are basically regular Postgres tables that are replicated uniformly (meaning entire content is replicated) to all the Datanodes. These are called as "Reference Tables" (RT).

RT is a special type of distributed hypertable. It does not need a time or space dimension columns. In RT, data is replicated to all DNs and is not hash partitioned by either time and/or space dimensions. We will enhance the create_hypertable API to include an additional optional argument reference.

```
CREATE OR REPLACE FUNCTION @extschema@.create_hypertable(
    relation                REGCLASS,
    ...
    distributed             BOOLEAN = NULL,
    reference               BOOLEAN = NULL 
) 
```

* If the reference argument is set to true then it has to additionally specify the distributed option as well. 
* Additionally, if replication_factor is specified then we will error out since the data needs to be present on all DNs uniformly. 
* All other time/space dimension columns references, we error out if they are specified along with reference argument. 
* Compression or freezing will NOT be allowed on RT objects.

nikkhils:
New proposal is to create dimensionless distributed hypertables without using any additional options like reference. The semantics of dimensionless distributed hypertables will be that they will have replication factor set to the current number of datanodes.

Erik suggests that for such dimensionless distributed hypertables, we do not create even one chunk the way the current implementation does. Instead, we route all changes to the same root table itself. We can do this by entering the root table entry in the chunk metadata. The alternative requires further investigation.

svenklemm:
multinode has been removed

PR #5155, opened 2023-01-09
Add "reference' column to hypertable

### LINKED ISSUE #5141: add "is_reference" boolean in _timescaledb_catalog.hypertable
As per #5140 we need to know that a specific distributed hypertable is a "REFERENCE TABLE" implementation. Add a "is_reference" boolean field for that. Also make changes to the hypertable views in all relevant locations to show that it's a reference table.

While we are doing this. Also add an additional "is_distributed" boolean field because right now it's derived indirectly by checking "replication_factor > 0" in many views, etc. So this will be helping remove tech-debt as well.

We could have considered using of bit flags in on single column and adding bits for REF/DISTRIBUTED etc. but we might not save a lot of space due to alignment padding. A boolean column uses 1 byte. Effectively, we can fit four or more boolean columns for one integer column. This also keeps the code simple.

nikkhils:
A distributed hypertable can be created as a "reference" table. Persist this information in the hypertable catalog table.

Fixes #5141

### DIFF
--- a/test/expected/drop_extension.out
+++ b/test/expected/drop_extension.out
@@ -11,9 +11,9 @@ NOTICE:  adding not-null constraint to column "time"
 (1 row)
 
 SELECT * FROM _timescaledb_catalog.hypertable;
- id | schema_name | table_name | associated_schema_name | associated_table_prefix | num_dimensions | chunk_sizing_func_schema |  chunk_sizing_func_name  | chunk_target_size | compression_state | compressed_hypertable_id | replication_factor 
-----+-------------+------------+------------------------+-------------------------+----------------+--------------------------+--------------------------+-------------------+-------------------+--------------------------+--------------------
-  1 | public      | drop_test  | _timescaledb_internal  | _hyper_1                |              2 | _timescaledb_internal    | calculate_chunk_interval |                 0 |                 0 |                          |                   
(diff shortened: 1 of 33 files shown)


test/expected/drop_extension.out
```diff
@@ -11,9 +11,9 @@ NOTICE:  adding not-null constraint to column "time"
...
-  1 | public      | drop_test  | _timescaledb_internal  | _hyper_1                |              2 | _timescaledb_internal    | calculate_chunk_interval |                 0 |                 0 |                          |                   
+ id | schema_name | table_name | associated_schema_name | associated_table_prefix | num_dimensions | chunk_sizing_func_schema |  chunk_sizing_func_name  | chunk_target_size | compression_state | compressed_hypertable_id | replication_factor | reference 
+----+-------------+------------+------------------------+-------------------------+----------------+--------------------------+--------------------------+-------------------+-------------------+--------------------------+--------------------+-----------
+  1 | public      | drop_test  | _timescaledb_internal  | _hyper_1                |              2 | _timescaledb_internal    | calculate_chunk_interval |                 0 |                 0 |                          |                    | f
```
jnidzwetzki, line 16: The test cases to modify this value will be added in #5140?

nikkhils: @jnidzwetzki no. Adding in this PR

jnidzwetzki:
The upgrade/downgrade tests are failing with `ERROR:  42P01: view "hypertables" does not exist`. Could you take a look at that?

svenklemm:
No longer relevant since multinode has been removed

PR #5172, opened 2023-01-11
Remove parallel safe from partialize_agg

### LINKED ISSUE #4922: [Bug]: Inconsistent and Incorrect Aggregations
### What type of bug is this?

Incorrect result

Multi-node

### What happened?

Queries utilising partial aggregation return incorrect and inconsistent results.

Example:
Query 1
`SELECT sum(value)
FROM table;`

Query 2
`SELECT SUM(value) FROM (
SELECT value
FROM table
ORDER BY 1) a;`

Query 1 will produce incorrect results, where Query 2 will bypass partial aggregation on the data nodes and return correctly.

2.8.1

### PostgreSQL version used

14.5

Ubuntu 22.04.1 x64

Docker

Amazon Web Services (AWS)

_No response_

Running `timescaledb-ha pg14.5-ts2.8.1-p0` on 1 access node with 4 data nodes using EKS with all nodes in the same availability zone.

```SQL
create table example_table_name
(
    time  timestamp not null,
    key_1 text,
    key_2 text,
    key_3 text,
    value double precision
);

SELECT create_distributed_hypertable(example_table_name, "time");

SELECT set_chunk_time_interval(example_table_name, INTERVAL '1h');

SELECT add_retention_policy(example_table_name, INTERVAL '30d');
```

Data of the form `(1664496000, 'a', '1efa7394-5540-410d-a468-b5f204d66ff9', 'b', 1)`

I had to insert 2M rows across 400 chunks before I saw this bug appear. Note that I have seen this issue occur on multiple tables now, all of which have the same partitioning and retention policies, but different schemas. `AVG` is also affected, the issue is not limited to `SUM`.

Editing to add: this also happens when the value is an int, and the string columns are unnecessary

fabriziomello:
Previous PR #4307 mark `partialize_agg` and `finalize_agg` as parallel safe but this change is leading to incorrect results in some cases.

Those functions are supposed work in parallel but seems is not the case and it is not evident yet the root cause and how to properly use it in parallel queries so we decided to revert this change and provide correct results to users.

Fixes #4922

### DIFF
--- a/sql/partialize_finalize.sql
+++ b/sql/partialize_finalize.sql
@@ -2,30 +2,24 @@
 -- Please see the included NOTICE for copyright information and
 -- LICENSE-APACHE for a copy of the license.
 
--- These wrapper functions are used to push down aggregation to data nodes.
--- They can be marked parallel-safe, and the parallel plan will be chosen
--- depending on whether the underlying aggregate function itself is
--- parallel-safe.
-
 CREATE OR REPLACE FUNCTION _timescaledb_internal.partialize_agg(arg ANYELEMENT)
-RETURNS BYTEA AS '@MODULE_PATHNAME@', 'ts_partialize_agg' LANGUAGE C STABLE PARALLEL SAFE;
+RETURNS BYTEA AS '@MODULE_PATHNAME@', 'ts_partialize_agg' LANGUAGE C VOLATILE;
 
 CREATE OR REPLACE FUNCTION _timescaledb_internal.finalize_agg_sfunc(
 tstate internal, aggfn TEXT, inner_agg_collation_schema NAME, inner_agg_collation_name NAME, inner_agg_input_types NAME[][], inner_agg_serialized_state BYTEA, return_type_dummy_val ANYELEMENT)
 RETURNS internal
 AS '@MODULE_PATHNAME@', 'ts_finalize_agg_sfunc'
-LANGUAGE C IMMUTABLE PARALLEL SAFE;
+LANGUAGE C IMMUTABLE;
 
 CREATE OR REPLACE FUNCTION _timescaledb_internal.finalize_agg_ffunc(
 tstate internal, aggfn TEXT, inner_agg_collation_schema NAME, inner_agg_collation_name NAME, inner_agg_input_types NAME[][], inner_agg_serialized_state BYTEA, return_type_dummy_val ANYELEMENT)
 RETURNS anyelement
 AS '@MODULE_PATHNAME@', 'ts_finalize_agg_ffunc'
-LANGUAGE C IMMUTABLE PARALLEL SAFE;
+LANGUAGE C IMMUTABLE;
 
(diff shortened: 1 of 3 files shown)


tsl/test/sql/partialize_finalize.sql
```diff
@@ -296,3 +296,44 @@ select _timescaledb_internal.finalize_agg( 'aggregate_to_test_ffunc_extra(int, a
...
+
+INSERT INTO issue4922 (time, value)
+SELECT time, (random()*30) as value
+FROM generate_series(now() - interval '100 hours', now(), '1 sec') AS time;
+
+SET force_parallel_mode = 'on';
+SET parallel_setup_cost = 0;
+
+-- Compare the results (should return NO ROWS)
+(SELECT
```
akuzm, line 317: Slightly confused by this test -- which subselect is the broken one? Maybe we could generate stable pseudo-random data and just show the results of both subselects. For pseudo-random we can use something like this: https://github.com/timescale/timescaledb/blob/main/tsl/test/sql/dist_param.sql#L20-L33

akuzm (approved):
OK as a quick fix, but I wonder what's the root cause.

tsl/test/sql/partialize_finalize.sql
akuzm: Also might be good to have an explain query so that we are absolutely sure it runs in parallel, because parallel execution is notoriously hard to achieve sometimes.

sql/partialize_finalize.sql
```diff
@@ -8,24 +8,23 @@
 -- parallel-safe.
 
 CREATE OR REPLACE FUNCTION _timescaledb_internal.partialize_agg(arg ANYELEMENT)
-RETURNS BYTEA AS '@MODULE_PATHNAME@', 'ts_partialize_agg' LANGUAGE C STABLE PARALLEL SAFE;
+RETURNS BYTEA AS '@MODULE_PATHNAME@', 'ts_partialize_agg' LANGUAGE C VOLATILE;
```
akuzm, line 11: Even volatile? I'd think it should be immutable, it's a pure function.

sql/partialize_finalize.sql
```diff
@@ -8,24 +8,23 @@
 -- parallel-safe.
```
akuzm, line 8: The comment above needs updating.

sql/partialize_finalize.sql
erimatnor: Yeah, why not keep STABLE and just remove parallel safe?

erimatnor: Is it because some aggs can be volatile themselves?

akuzm: > Is it because some aggs can be volatile themselves?

Postgres check the volatility of the entire expression tree, so this shouldn't matter.

fabriziomello: I didn't thought too much about it because the idea here is to REVERT to the previous state.

tsl/test/sql/partialize_finalize.sql
fabriziomello: You're correct will update the tests.

sql/partialize_finalize.sql
erimatnor: Yeah, I also think that this function is essentially just a planner hint and not actually executed.

tsl/test/sql/partialize_finalize.sql
fabriziomello: Tests updated

PR #5184, opened 2023-01-16
Fix default data node availability status

erimatnor:
Ensure that the availability of a data node is "true" if no metadata is set about the node's availability.


erimatnor:
Already being fixed in https://github.com/timescale/timescaledb/issues/5154

PR #5212, opened 2023-01-24
Allow pushdown of reference table joins

jnidzwetzki:
This patch adds the functionality that is needed to perform distributed, parallel joins on reference tables on access nodes. This code allows the pushdown of a join if:

 * (1) The setting "ts_guc_enable_per_data_node_queries" is enabled
 * (2) The outer relation is a distributed hypertable
 * (3) The inner relation is marked as a reference table
 * (4) The join is a left join or an inner join

### DIFF
--- a/src/cross_module_fn.c
+++ b/src/cross_module_fn.c
@@ -402,6 +402,13 @@ ts_tsl_loaded(PG_FUNCTION_ARGS)
 	PG_RETURN_BOOL(ts_cm_functions != &ts_cm_functions_default);
 }
 
+static void
+mn_get_foreign_join_path_default_fn_pg_community(PlannerInfo *root, RelOptInfo *joinrel,
+												 RelOptInfo *outerrel, RelOptInfo *innerrel,
+												 JoinType jointype, JoinPathExtraData *extra)
+{
+}
+
 /*
  * Define cross-module functions' default values:
  * If the submodule isn't activated, using one of the cm functions will throw an
@@ -555,6 +562,7 @@ TSDLLEXPORT CrossModuleFunctions ts_cm_functions_default = {
 	.hypertable_distributed_set_replication_factor = error_no_default_fn_pg_community,
 	.update_compressed_chunk_relstats = update_compressed_chunk_relstats_default,
 	.health_check = error_no_default_fn_pg_community,
+	.mn_get_foreign_join_paths = mn_get_foreign_join_path_default_fn_pg_community,
 };
 
 TSDLLEXPORT CrossModuleFunctions *ts_cm_functions = &ts_cm_functions_default;
--- a/src/planner/planner.c
+++ b/src/planner/planner.c
@@ -110,6 +110,7 @@ void _planner_fini(void);
 
 static planner_hook_type prev_planner_hook;
 static set_rel_pathlist_hook_type prev_set_rel_pathlist_hook;
+static set_join_pathlist_hook_type prev_set_join_pathlist_hook;
 static get_relation_info_hook_type prev_get_relation_info_hook;
 static create_upper_paths_hook_type prev_create_upper_paths_hook;
 static void cagg_reorder_groupby_clause(RangeTblEntry *subq_rte, Index rtno, List *outer_sortcl,
@@ -1662,13 +1663,36 @@ cagg_reorder_groupby_clause(RangeTblEntry *subq_rte, Index rtno, List *outer_sor
 	}
 }
 
+/* Register our join pushdown hook. Becuase for PostgreSQL the tables we are operating
+ * on are local tables. So, the FDW hooks are not called. Register our join path
+ * generation as a generic planer hook.
+ */
+
+static void
+timescaledb_set_join_pathlist_hook(PlannerInfo *root, RelOptInfo *joinrel, RelOptInfo *outerrel,
+								   RelOptInfo *innerrel, JoinType jointype,
+								   JoinPathExtraData *extra)
+{
+	/* Left table has to be a distributed hypertable */
+	TimescaleDBPrivate *outerrel_private = outerrel->fdw_private;
+	if (outerrel_private != NULL && outerrel_private->fdw_relation_info != NULL)
+		ts_cm_functions
+			->mn_get_foreign_join_paths(root, joinrel, outerrel, innerrel, jointype, extra);
+
+	/* Call next hook in chain */
+	if (prev_set_join_pathlist_hook != NULL)
+		(*prev_set_join_pathlist_hook)(root, joinrel, outerrel, innerrel, jointype, extra);
+}
+
 void
 _planner_init(void)
 {
 	prev_planner_hook = planner_hook;
 	planner_hook = timescaledb_planner;
 	prev_set_rel_pathlist_hook = set_rel_pathlist_hook;
+	prev_set_join_pathlist_hook = set_join_pathlist_hook;
 	set_rel_pathlist_hook = timescaledb_set_rel_pathlist;
+	set_join_pathlist_hook = timescaledb_set_join_pathlist_hook;
 
 	prev_get_relation_info_hook = get_relation_info_hook;
 	get_relation_info_hook = timescaledb_get_relation_info_hook;
@@ -1682,6 +1706,7 @@ _planner_fini(void)
 {
 	planner_hook = prev_planner_hook;
 	set_rel_pathlist_hook = prev_set_rel_pathlist_hook;
+	set_join_pathlist_hook = prev_set_join_pathlist_hook;
 	get_relation_info_hook = prev_get_relation_info_hook;
 	create_upper_paths_hook = prev_create_upper_paths_hook;
 }
--- a/tsl/src/fdw/data_node_scan_plan.c
+++ b/tsl/src/fdw/data_node_scan_plan.c
@@ -43,6 +43,10 @@
 #include "fdw_utils.h"
 #include "relinfo.h"
 #include "scan_plan.h"
+#include "estimate.h"
+#include "planner/planner.h"
+#include "chunk.h"
+#include "debug_assert.h"
 
 /*
  * DataNodeScan is a custom scan implementation for scanning hypertables on
@@ -70,10 +74,20 @@ static Path *data_node_scan_path_create(PlannerInfo *root, RelOptInfo *rel, Path
 										double rows, Cost startup_cost, Cost total_cost,
 										List *pathkeys, Relids required_outer, Path *fdw_outerpath,
 										List *private);
+
+static Path *data_node_join_path_create(PlannerInfo *root, RelOptInfo *rel, PathTarget *target,
+										double rows, Cost startup_cost, Cost total_cost,
+										List *pathkeys, Relids required_outer, Path *fdw_outerpath,
+										List *private);
+
 static Path *data_node_scan_upper_path_create(PlannerInfo *root, RelOptInfo *rel,
 											  PathTarget *target, double rows, Cost startup_cost,
 											  Cost total_cost, List *pathkeys, Path *fdw_outerpath,
 											  List *private);
+
+static bool fdw_pushdown_foreign_join(PlannerInfo *root, RelOptInfo *joinrel, JoinType jointype,
+									  RelOptInfo *outerrel, RelOptInfo *innerrel,
+									  JoinPathExtraData *extra);
 
 static AppendRelInfo *
 create_append_rel_info(PlannerInfo *root, Index childrelid, Index parentrelid)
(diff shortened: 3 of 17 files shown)


tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+}
+
+/*
+ * Check if the query performs a join between a hypertble (outer) and a reference
+ * table (inner) and the join type is a LEFT JOIN, an INNER JOIN, or an implicit
+ * join.
+ */
+static bool
+is_safe_to_pushdown_reftable_join(PlannerInfo *root, List *join_reference_tables,
+								  RangeTblEntry *outertableref, RangeTblEntry *innertableref,
```
erimatnor, line 736: `outertableref` not used in this function except for asserts. I guess it is maybe because we don't support all kinds of joins (yet), but consider eliding this parameter from the parameter list.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	 */
+	fpinfo = fdw_relinfo_get(joinrel);
+	fpinfo_o = fdw_relinfo_get(outerrel);
+	fpinfo_i = fdw_relinfo_get(innerrel);
+
+	/* shouldn't happen in the current implementation */
+	if (!fpinfo_o || !fpinfo_o->pushdown_safe || !fpinfo_i || !fpinfo_i->pushdown_safe)
+		ereport(ERROR,
+				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+				 errmsg("join pushdown failed, a partition is not marked as pushdown safe")));
```
erimatnor, line 805:
```suggestion
				 errmsg("join pushdown failed because a partition is not marked as pushdown safe")));
```

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+
+	/*
+	 * If either of the joining relations is marked as unsafe to pushdown, the
+	 * join can not be pushed down.
+	 */
+	fpinfo = fdw_relinfo_get(joinrel);
+	fpinfo_o = fdw_relinfo_get(outerrel);
+	fpinfo_i = fdw_relinfo_get(innerrel);
+
+	/* shouldn't happen in the current implementation */
```
erimatnor, line 801: Consider asserting for now if it cannot happen. We can always change asserts to checks later, when needed.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	if (!fpinfo_o || !fpinfo_o->pushdown_safe || !fpinfo_i || !fpinfo_i->pushdown_safe)
+		ereport(ERROR,
+				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+				 errmsg("join pushdown failed, a partition is not marked as pushdown safe")));
+
+	/*
+	 * If joining relations have local conditions, those conditions are
+	 * required to be applied before joining the relations. Hence the join can
+	 * not be pushed down (shouldn't happen in the current implementation).
+	 */
```
erimatnor, line 811: Same here: if the condition is not possible, consider asserting instead of having errors that will never display.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	if (fpinfo_o->local_conds || fpinfo_i->local_conds)
+		ereport(ERROR,
+				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+				 errmsg("join pushdown failed, local conditions are present")));
```
erimatnor, line 815:
```suggestion
				 errmsg("join pushdown failed because local conditions are present")));
```

tsl/src/fdw/data_node_scan_plan.c
erimatnor: Should this really be an error? I guess I would simply expect no pushdown, but the query executing anyway in non-optimal way.

tsl/src/fdw/data_node_scan_plan.c
erimatnor: Again, should we really fail as opposed to just producing a non-pushdown plan?

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	{
+		RestrictInfo *rinfo = lfirst_node(RestrictInfo, lc);
+		bool is_remote_clause = ts_is_foreign_expr(root, joinrel, rinfo->clause);
+
+		if (IS_OUTER_JOIN(jointype) && !RINFO_IS_PUSHED_DOWN(rinfo, joinrel->relids))
+		{
+			if (!is_remote_clause)
+				ereport(ERROR,
+						(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
+						 errmsg("join pushdown failed, a clause can not be pushed down")));
```
erimatnor, line 845:
```suggestion
						 errmsg("join pushdown failed because a clause can not be pushed down")));
```

Again, should we really fail here?

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+#else
+			fpinfo->joinclauses =
+				list_concat(fpinfo->joinclauses, list_copy(fpinfo_i->remote_conds));
+			fpinfo->remote_conds =
+				list_concat(fpinfo->remote_conds, list_copy(fpinfo_o->remote_conds));
+#endif
+			break;
+
+		case JOIN_RIGHT:
+#if PG14_GE
```
erimatnor, line 949: Looks like we need a test case for JOIN_RIGHT

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+			fpinfo->remote_conds = list_concat(fpinfo->remote_conds, fpinfo_i->remote_conds);
+#else
+			fpinfo->joinclauses =
+				list_concat(fpinfo->joinclauses, list_copy(fpinfo_o->remote_conds));
+			fpinfo->remote_conds =
+				list_concat(fpinfo->remote_conds, list_copy(fpinfo_i->remote_conds));
+#endif
+			break;
+
+		case JOIN_FULL:
```
erimatnor, line 960: Also need a test case for JOIN FULL

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	 * Skip check if the join result has been considered already.
+	 */
+	if (joinrel->fdw_private)
+		return;
+
+	/*
+	 * This code does not work for joins with lateral references, since those
+	 * must have parameterized paths, which we don't generate yet.
+	 */
+	if (!bms_is_empty(joinrel->lateral_relids))
```
erimatnor, line 1156: Can we add a test case with join lateral?

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	 * because otherwise we should not start the planning for distributed tables. However
+	 * we can check this condition here and fail more nicely by doing nothing than hitting
+	 * a failed assert. */
+	TimescaleDBPrivate *outerrel_private = outerrel->fdw_private;
+	if (outerrel_private == NULL || outerrel_private->fdw_relation_info == NULL)
+		return;
+
+	/* Ensure pushdown planning is only performed on the AN. On the DNs the outererl is not
+	 * partitioned. */
+	if (outerrel->nparts <= 0)
```
erimatnor, line 1169: I guess this code isn't even invoked on data nodes, so we can probably assert this.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -674,6 +688,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+		return;
+
+	/* Get the hypertable from the outer relation. */
+	RangeTblEntry *rte_outer = planner_rt_fetch(outerrel->relid, root);
+
+	/* Test that the fetched outer relation is a distributed hypertable. If not,
+	 * the pushdown is not possible. This should be already tested by the previous
+	 * checks on 'outerrel'. But fail nicely if the condition is not meet. */
+	if (rte_outer == NULL || !is_distributed_hypertable(rte_outer->relid))
```
erimatnor, line 1178: This should probably be an assertion. This function should never be invoked on a non-distributed hypertable.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -955,6 +1652,54 @@ data_node_scan_path_create(PlannerInfo *root, RelOptInfo *rel, PathTarget *targe
...
+ */
+static Path *
+data_node_join_path_create(PlannerInfo *root, RelOptInfo *rel, PathTarget *target, double rows,
+						   Cost startup_cost, Cost total_cost, List *pathkeys,
+						   Relids required_outer, Path *fdw_outerpath, List *private)
+{
+	DataNodeScanPath *scanpath = palloc0(sizeof(DataNodeScanPath));
+
+	if (rel->lateral_relids && !bms_is_subset(rel->lateral_relids, required_outer))
+		required_outer = bms_union(required_outer, rel->lateral_relids);
```
erimatnor, line 1673: Add test case?

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -955,6 +1652,54 @@ data_node_scan_path_create(PlannerInfo *root, RelOptInfo *rel, PathTarget *targe
...
+
+	/*
+	 * We should use get_joinrel_parampathinfo to handle parameterized paths,
+	 * but the API of this function doesn't support it, and existing
+	 * extensions aren't yet trying to build such paths anyway.  For the
+	 * moment just throw an error if someone tries it; eventually we should
+	 * revisit this.
+	 */
+	if (!bms_is_empty(required_outer) || !bms_is_empty(rel->lateral_relids))
+		elog(ERROR, "parameterized foreign joins are not supported yet");
```
erimatnor, line 1683: Add test case?

tsl/src/fdw/data_node_scan_plan.c
akuzm: Yes, looks like this shouldn't be an error. We generally have to test four cases: ec/non-ec join, with shippable/non-shippable functions. I have some examples in dist_param test, scroll down from here: https://github.com/timescale/timescaledb/blob/main/tsl/test/sql/dist_param.sql#L64

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Good point, the parameter is removed.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Changed, the query is now executed in that case without a pushed-down join.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: @erimatnor Changed, the join is not pushed down in that case.

@akuzm Thank you for pointing out that shippable/non-shippable / EC,non-EC cases are not tested. The test cases are now added.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Changed

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: We are now returning (and not pushing down the join) instead of showing an error message. I could change this to an assert statement. But I am more in favor of returning this case. What is your opinion on this?

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Same as in Line 801

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Regarding the test cases / test coverage: The code is from PG upstream and is not triggered in the current implementation. We could remove these conditions and the support for other join types. However, if we decide to support these joins, it makes a lot of work to determine which statements have been removed (especially if these are single statements and not obvious things like entire functions or parameters).

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: I have seen join replanning on the DNs. So, I added this code to ensure that only the AN performs the planning.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: These parameters are replaced before the `data_node_scan_plan` is generated in our implementation; this check is from PG upstream. However, I added a test case for a prepared statement to test that the join pushdown also works in this case. 

We have a similar check also for [regular scans](https://github.com/timescale/timescaledb/blob/206056ca12c611b2038b35bfa0eac43372a563c2/tsl/src/fdw/data_node_scan_plan.c#L937) in place. I wanted to check if we were covering this line in the existing tests. But unfortunately, codecov currently has [an outage](https://status.codecov.com/) at the moment and I can't access the reports.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Changed to an assertion.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: @erimatnor Codecov works now and I checked the coverage of the closely related line in the existing code. [This line](https://github.com/timescale/timescaledb/blob/206056ca12c611b2038b35bfa0eac43372a563c2/tsl/src/fdw/data_node_scan_plan.c#L937) in the existing `data_node_scan_path_create` function is also one of these lines that are copied from PG upstream and cannot directly be tested in our implementation. So, it is hard to create a test case for the new/similar line. 

https://app.codecov.io/gh/timescale/timescaledb/commit/0562c6d89aa757e7aa7f3acfaea3f3d386fe9f6c/blob/tsl/src/fdw/data_node_scan_plan.c#L1635

tsl/src/fdw/data_node_scan_plan.c
akuzm: > I have seen join replanning on the DNs.

This would be a serious bug, the data nodes don't use data node scans. Let's make this an Assert or Ensure.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	 * This code does not work for joins with lateral references, since those
+	 * must have parameterized paths, which we don't generate yet.
+	 */
+	if (!bms_is_empty(joinrel->lateral_relids))
+		return;
+
+	/* The outerrel has to be distributed. This condition should be always hold
+	 * because otherwise we should not start the planning for distributed tables. However
+	 * we can check this condition here and fail more nicely by doing nothing than hitting
+	 * a failed assert. */
```
akuzm, line 1219: This sounds like an internal program error, and we don't normally try to fail gracefully for them. Is this case different?

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	if (jointype != JOIN_INNER && jointype != JOIN_LEFT)
+		return false;
+
+	/* Check that at least one reference table is defined. */
+	if (join_reference_tables == NIL)
+		return false;
+
+	/* Only queries with two tables and one join are supported. */
+	fromList = query->jointree->fromlist;
+	if (!is_inner_or_left_join_two_tables(fromList) && !is_implicit_join_two_tables(fromList))
```
akuzm, line 826: I have a bad feeling about using the raw parse tree at this stage of the planning. Generally we look at RelOptInfos. If you want to limit it to two participating tables, can you check the relids of the innerrel and outerrel?

tsl/src/init.c
```diff
@@ -235,6 +236,7 @@ CrossModuleFunctions tsl_cm_functions = {
 	.cache_syscache_invalidate = cache_syscache_invalidate,
 	.update_compressed_chunk_relstats = update_compressed_chunk_relstats,
 	.health_check = ts_dist_health_check,
+	.mn_set_foreign_join_paths = ts_get_foreign_join_paths,
```
akuzm, line 239: Can you use the same name for the field? Easier to grep.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	return true;
+}
+
+/*
+ * Check if the join fromList belongs to an implicit join (e.g.,
+ * SELECT * FROM metric m1, metric_name m2 WHERE m1.id=m2.id). This type
+ * of join has two RANGETBLREF entries in the fromList.
+ */
+static bool
+is_implicit_join_two_tables(List *fromList)
```
akuzm, line 778: Is there anything special about the grammatically implicit join? It's just a particular kind of grammar for the usual inner join, right?

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,689 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+ * (1) The setting "ts_guc_enable_per_data_node_queries" is enabled
+ * (2) The outer relation is a distributed hypertable
+ * (3) The inner relation is marked as a reference table
+ * (4) The join is a left join or an inner join
+ *
+ * The join will be performed between the multiple DataNodeRels (see function
+ * build_data_node_part_rels) and the original innerrel of the join (the reftable).
+ */
+void
+data_node_generate_pushdown_join_paths(PlannerInfo *root, RelOptInfo *joinrel, RelOptInfo *outerrel,
```
akuzm, line 1184: This is basically the top-level GetForeignJoinPaths hook, right? Can you please add the name of the corresponding fdw function and the postgres version (maybe even sha) that you used? Will be easier to port future changes this way.

tsl/src/fdw/data_node_scan_plan.c
akuzm: The corresponding check is performed by the top-level join pathlist hook `get_foreign_join_paths_hook`

src/planner/planner.c
```diff
@@ -1662,13 +1663,35 @@ cagg_reorder_groupby_clause(RangeTblEntry *subq_rte, Index rtno, List *outer_sor
...
 }
 
+/* Register our join pushdown hook. Becuase for PostgreSQL
+ * the tables we are operating on are local tables. So,
+ * the FDW hooks are not called. Register our join path
+ * generation as a generic planer hook.
+ */
+
+static void
+get_foreign_join_paths_hook(PlannerInfo *root, RelOptInfo *joinrel, RelOptInfo *outerrel,
```
akuzm, line 1673: The name might be a little misleading, this is the join pathlist hook, not the `GetForeignJoinPaths` fdw method.

tsl/src/fdw/fdw.c
```diff
@@ -395,6 +406,8 @@ static FdwRoutine timescaledb_fdw_routine = {
 	/* explain/analyze */
 	.ExplainForeignScan = explain_foreign_scan,
 	.ExplainForeignModify = explain_foreign_modify,
+	/* Support functions for join push-down */
+	.GetForeignJoinPaths = ts_get_foreign_join_paths,
```
akuzm, line 410: For the first version, are we going with reference tables as normal manually replicated tables? This would mean that this hook is not needed, because it is only called for joins of two foreign tables from the same foreign server. Might make sense to clarify this in comments or even remove it.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: So far, the function was called by the generic planner hook every time a join was part of the query and `fdw_private` was not empty. I made the generic planner hook (`timescaledb_set_join_pathlist_hook`) more selective when this FDW function is called. With this change, I could convert this statement to an Assert.

tsl/src/fdw/fdw.c
jnidzwetzki: Good point, it is removed.

tsl/src/init.c
jnidzwetzki: The name of the function is changed.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: The statement is changed to an assert.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Could you elaborate on this? I would like to understand why this approach is bad. We use the fromlist also in [other places](https://github.com/timescale/timescaledb/blob/484a4ea3fc5acb58468f19d445b38c7aaf38fb7b/tsl/src/continuous_aggs/create.c#L1223) to detect the join type and the involved tables.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: They are indeed special. For left / right joins, you see a `JoinExpr` in the `fromList`. For these implicit joins, you have only `RangeTblRef` entries in the list (see the comment of the function `is_implicit_join_two_tables`).

tsl/src/fdw/data_node_scan_plan.c
akuzm: The parse tree (`Query`) is not normally used at this point by Postgres. If you look at other join planning code like `make_one_rel`, it only uses the RelOptInfos. I'd say most Path-generation code uses RelOptInfos, which we could roughly call a "relational algebra" representation. If you are using the parse tree at the later stages of planning, you'll have to redo a lot of work done earlier. For example, here you have to recursively traverse the fromlist to find the joins written as subqueries. This was already done for you by `pull_up_subqueries`. Or if you want to examine the join clauses, you'll have to redo the work of `distribute_quals_to_rels`. 

Sometimes using the parse tree is unavoidable because we don't have the suitable hooks. But here it's not the case.

Not familiar with the caggs code, maybe it's the hook problem, or maybe they're just doing it wrong :)

tsl/src/fdw/data_node_scan_plan.c
akuzm: Hahah, I scroll down and here is an example of what I was talking about above! Just use reloptinfos :)

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Thank you for the detailed explanation. In the actual join path generation (e.g., in the function `data_node_generate_pushdown_join_paths`) the `RelOptInfo` data structures are used as proposed. Also here in the `is_safe_to_pushdown_reftable_join` function, I use the data structures of the planner (i.e., `JoinType`) wherever I can. As you mentioned, a lot of work is applied to generate these data structures and they distinguish from the raw query (e.g., subqueries are pulled up). 

However, the current join pushdown implementation should only be applied if a query contains one join and two tables. My intention here is to access the raw query to check how the query looked like before it was pre-processed to get this information (the `root->simple_rel_array` contains a lot of additional `RelOptInfo`, which makes it harder to get this information from there).

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+						"'enable_per_data_node_queries' GUC is disabled")));
+
+		return;
+	}
+
+	/* The inner table can be a distributed hypertable or a plain table. Plain tables don't have
+	 * a TsFdwRelInfo at this point. So, it needs to be created.
+	 */
+	if (innerrel->fdw_private == NULL)
+		fdw_relinfo_create(root, innerrel, InvalidOid, InvalidOid, TS_FDW_RELINFO_REFJOIN_TABLE);
```
akuzm, line 1309: Does it really need an fpinfo?

Also please note that it might be a table managed by another extension (e.g. `postgres_fdw`). This is not likely, but we have to be careful about accessing `fdw_private` in such cases. The other extensions are the reason why we don't touch the `fdw_private` field unless our `classify_relation` says it's our table.

Also the name is somewhat confusing, maybe `TS_FDW_RELINFO_REFERENCE_TABLE`?

akuzm: > Does it really need an fpinfo?

Oh, it probably does, at least we're storing the local/remote conditions there.

tsl/src/fdw/data_node_scan_plan.c
akuzm: > However, the current join pushdown implementation should only be applied if a query contains one join and two tables. 

Does anything break if the query contains other joins? I'd say it shouldn't, but not sure. For the particular join you're pushing down, you can check that it's two tables with  `bms_num_members(jonrel->relids) == 2`.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+		return false;
+
+	/* Join can be pushed down */
+	return true;
+}
+
+/*
+ * Assess whether the join between inner and outer relations can be pushed down
+ * to the foreign server. As a side effect, save information we obtain in this
+ * function to TsFdwRelInfo passed in.
```
akuzm, line 840:
```suggestion
 * function to TsFdwRelInfo passed in.
 *
 * Based on foreign_join_ok() from postgres_fdw v. 15.1.
```

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	fpinfo_o = fdw_relinfo_get(outerrel);
+	fpinfo_i = fdw_relinfo_get(innerrel);
+
+	/* shouldn't happen in the current implementation */
+	if (!fpinfo_o || !fpinfo_o->pushdown_safe || !fpinfo_i || !fpinfo_i->pushdown_safe)
+		return false;
+
+	/*
+	 * If joining relations have local conditions, those conditions are
+	 * required to be applied before joining the relations. Hence the join can
```
akuzm, line 866: Should be applied before -- so that the outer joins work properly? Not sure if we want this optimization to apply when distributed table is outer -- only makes us fetch more data.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	 * deparseExplicitTargetList() isn't smart enough to handle anything other
+	 * than a Var.  In particular, if there's some PlaceHolderVar that would
+	 * need to be evaluated within this join tree (because there's an upper
+	 * reference to a quantity that may go to NULL as a result of an outer
+	 * join), then we can't try to push the join down because we'll fail when
+	 * we get to deparseExplicitTargetList().  However, a PlaceHolderVar that
+	 * needs to be evaluated *at the top* of this join tree is OK, because we
+	 * can do that locally after fetching the results from the remote side.
+	 */
+	foreach (lc, root->placeholder_list)
```
akuzm, line 926: PlaceHolderVars always make my head hurt... I see that this is also postgres_fdw code, but let's add a test for it. Codecov complains. We should think twice whether we need outer joins, might be easier to delete all this.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+}
+#endif
+
+/*
+ * Create a new join partition RelOptInfo data structure for a partition. The data
+ * structure is based on the parameter joinrel. The paramter is taken as template
+ * and adjusted for the partition provided by the parameter hyper_table_rel.
+ */
+static RelOptInfo *
+create_join_partition(PlannerInfo *root, RelOptInfo *innerrel, RelOptInfo *joinrel,
```
akuzm, line 1106: `create_join_partition_for_data_node`? Or `create_data_node_partition_for_join` :thinking:  Just to make it sound less generic, we have other kinds of partition-wise joins.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	RelOptInfo *join_partition = palloc(sizeof(RelOptInfo));
+	memcpy(join_partition, joinrel, sizeof(RelOptInfo));
+
+	/* Create a new relinfo for the join partition */
+	join_partition->fdw_private = NULL;
+	TsFdwRelInfo *join_part_fpinfo = fdw_relinfo_create(root,
+														join_partition,
+														hyper_table_rel->serverid,
+														InvalidOid,
+														TS_FDW_RELINFO_REFJOIN_PARTITION);
```
akuzm, line 1118: Is `hyper_table_rel` actually a `data_node_rel` here? These are different things, I'd rename to avoid confusion.

tsl/src/fdw/data_node_scan_plan.c
akuzm: `create_data_node_partition_for_joinrel` is even more specific.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+		return;
+
+	/*
+	 * This code does not work for joins with lateral references, since those
+	 * must have parameterized paths, which we don't generate yet.
+	 */
+	if (!bms_is_empty(joinrel->lateral_relids))
+		return;
+
+	/* Ensure pushdown planning is only performed on the AN. On the DNs the outererl is not
```
akuzm, line 1219:
```suggestion
	/* Ensure pushdown planning is only performed on the AN. On the DNs the outerrel is not
```

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	Query *query = root->parse;
+	List *fromList;
+
+	Assert(root != NULL);
+	Assert(innertableref != NULL);
+
+	/*
+	 * We support pushing down of INNER and LEFT joins only.
+	 *
+	 * Constructing queries representing partitioned OUTER, SEMI and ANTI
```
akuzm, line 814:
```suggestion
	 * Constructing queries representing partitioned FULL, SEMI and ANTI
```

Did you mean FULL? The LEFT and RIGHT are both a variants of outer.

tsl/src/fdw/data_node_scan_plan.c
akuzm: I only just realized that this is exactly the same kind of partitioning we have for data node scans :) Somehow the word "partition" confused me into thinking about generic partition-wise joins, that would be chunk-wise in our terms. Maybe `data_node_joinrel` would be even better...

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: When you perform a query like:

```sql
 SELECT * FROM metric m1, metric m2, metric_name mn WHERE m1.id = mn.id AND m2.id = mn.id;
```

(`metric` is a DHT, `metric_name` a reference_table)

PostgreSQL could perform one join on all three tables or join two tables and join this result with the remaining table. In the latter case, you end up with `joinrels` where the condition `bms_num_members(joinrel->relids) == 2` holds but the join should not be pushed down.

My intention is here to detect queries where more than two tables are involved and don't push down the join in this case. This condition could be lifted in a further PR. However, we decided to use a strict filter here and remove these restrictions in the future when we know that the first join pushdown implementation works.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: In the current implementation, we don't support outer joins. We could lift all of these restrictions later as soon as we have the first version of the join pushdown working.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: I could not trigger this code in the tests because we don't support outer joins or joins on views with constants at the moment. We could remove this code for the current implementation. As I mentioned [here](https://github.com/timescale/timescaledb/pull/5212#discussion_r1095802858), if we decide to remove this kind of code, supporting these features later takes more effort because you need to compare our implementation and the PG upstream implementation line-by-line.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	join_partition->reltarget = create_empty_pathtarget();
+	join_partition->reltarget->sortgrouprefs = joinrel->reltarget->sortgrouprefs;
+	join_partition->reltarget->cost = joinrel->reltarget->cost;
+	join_partition->reltarget->width = joinrel->reltarget->width;
+#if PG14_GE
+	join_partition->reltarget->has_volatile_expr = joinrel->reltarget->has_volatile_expr;
+#endif
+	join_partition->reltarget->exprs =
+		castNode(List,
+				 adjust_appendrel_attrs(root, (Node *) joinrel->reltarget->exprs, 1, &appinfo));
```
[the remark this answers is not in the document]
jnidzwetzki: Yes, this is needed. The join is performed partition-wise. Each of the partitions belongs to one data node and we are adjusting the expressions here using the `appinfo` of the data node.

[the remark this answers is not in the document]
akuzm: Let's call it "data-node-wise" or something, because "partition" is ambiguous between chunks and data nodes. So which kind of adjustment exactly do we do? I see that `adjust_apprendrel_attrs` can change varno or convert row type. For data node joinrels, the varno and rowtype are the same, because we just copy them from the base joinrel. Does this function do anything else?

tsl/src/fdw/data_node_scan_plan.c
akuzm: > In the latter case, you end up with joinrels where the condition bms_num_members(joinrel->relids) == 2 holds but the join should not be pushed down.

Yes, we should also check that outer table is distributed, and inner table is a reference table. These three conditions should be sufficient, no? And we can check all of them without using the parse tree.

tsl/src/fdw/data_node_scan_plan.c
akuzm: OK, let's keep it for the ease of comparison. Maybe makes sense to add a comment and/or assert about this somewhere.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: I have converted the statements to `Asserts.`

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: As suggested, converted to `Asserts`.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: At this point, the `innerrel` has to be marked as a reference table. So it has to be 'our' table. Depending on the used replication method, it could be a DHT or a plain PostgreSQL table. I am not sure that `classify_relation` would help much here to protect this table.

As suggested `TS_FDW_RELINFO_REFJOIN_TABLE` was renamed to `TS_FDW_RELINFO_REFERENCE_TABLE`.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: As suggested, I added a comment that this code is currently not used.

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+
+	return true;
+}
+
+#ifdef USE_ASSERT_CHECKING
+/*
+ * Check if the given hypertable is a distributed hypertable.
+ */
+static bool
+is_distributed_hypertable(Oid hypertable_id)
```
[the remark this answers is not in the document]
jnidzwetzki: As suggested, the variable has been renamed to `hypertable_reloid`.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: The function has been renamed to `create_data_node_joinrel`.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: It is actually a `data_node_rel` here. I renamed the variable to reflect this better.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: Thanks, you are correct, the comment has been updated.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: I just tested it. This is not sufficient to check. If you execute a join as described [here](https://github.com/timescale/timescaledb/pull/5212#discussion_r1105469959), PostgreSQL generates multiple ways of performing the join. 

The first step could be the join between `metric m1` and `metric_name mn` and then a further join between this intermediate result and `metric m2` is performed. For the first join `bms_num_members(joinrel->relids) == 2` is true and we have a distributed outer table and an inner reference table. However, we don't want to push down the join in this case.

tsl/src/fdw/data_node_scan_plan.c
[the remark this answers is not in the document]
jnidzwetzki: The function is renamed as suggested.

With `adjust_apprendrel_attrs`, the relid in the expressions is translated from the hypertable relid to the relid of the data node that is responsible for this partition. Otherwise, the deparser could not pushdown these expressions properly.

tsl/src/fdw/data_node_scan_plan.c
akuzm: > However, we don't want to push down the join in this case.

But why? `m1` and `m2` are used symmetrically here, this distinction looks arbitrary.

akuzm: > But why? m1 and m2 are used symmetrically here, this distinction looks arbitrary.

This would be very weird behavior to have, we would favor some of the grammatically implicit joins in the join order, based on their relative position in the query? That's not normally done in SQL.

tsl/src/fdw/data_node_scan_plan.c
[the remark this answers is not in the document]
akuzm: > With adjust_apprendrel_attrs, the relid in the expressions is translated from the hypertable relid to the relid of the data node that is responsible for this partition.

Yes, but aren't they exactly equal? As set here: https://github.com/timescale/timescaledb/pull/5212/files#diff-47f57773f9558fb5ef0345e680eb19c0d1c9378c1701011b7a930a96e8a0813cR1135

This is relid in the sense of `varno` and `RelOptInfo`, not the oid from RTE.

tsl/src/fdw/data_node_scan_plan.c
akuzm: Even the explicit inner joins don't enforce the particular order, you have to set a special guc for it: https://www.postgresql.org/docs/current/explicit-joins.html

tsl/src/fdw/data_node_scan_plan.c
```diff
@@ -741,6 +755,691 @@ push_down_group_bys(PlannerInfo *root, RelOptInfo *hyper_rel, Hyperspace *hs,
...
+	/*
+	 * Create unfinished PgFdwRelationInfo entry which is used to indicate
+	 * that the join relation is already considered, so that we won't waste
+	 * time in judging safety of join pushdown and adding the same paths again
+	 * if found safe. Once we know that this join can be pushed down, we fill
+	 * the entry.
+	 */
+	fpinfo =
+		fdw_relinfo_create(root, joinrel, InvalidOid, InvalidOid, TS_FDW_RELINFO_UNINITIALIZED);
+	Assert(fpinfo->type == TS_FDW_RELINFO_UNINITIALIZED);
```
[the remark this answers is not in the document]
jnidzwetzki: The PostgreSQL logic is to add a dummy value to joinrel->fpinfo to indicate that we have already processed this joinrel  (see [this](https://github.com/postgres/postgres/blob/ef7002dbe0b06e4e5b42c89becd4eb9be2e9aa89/contrib/postgres_fdw/postgres_fdw.c#L6249) and [this](https://github.com/postgres/postgres/blob/ef7002dbe0b06e4e5b42c89becd4eb9be2e9aa89/contrib/postgres_fdw/postgres_fdw.c#L6260)). 

I introduced the `TS_FDW_RELINFO_JOIN` type to distinguish this fpinfo type from an actual `TS_FDW_RELINFO_UNINITIALIZED` fpinfo.

tsl/src/fdw/data_node_scan_plan.c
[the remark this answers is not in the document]
jnidzwetzki: Yes, these values are equal. However, in this function, we have to adjust the expressions for the join partition and make sure that the expressions (they are generated for the unpartitioned join) belong to this partition. We do something similar in [adjust_data_node_rel_attrs](https://github.com/timescale/timescaledb/blob/46632f887365a3c64377571741114fc4cd11a7d6/tsl/src/fdw/data_node_scan_plan.c#L217) for the DataNodes.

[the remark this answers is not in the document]
akuzm: OK, I still don't understand what exactly it adjusts, given that the varno and row type are exactly the same, but I'll have to trust you on this :)

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: I restructured my code. Now it is enough to check `bms_num_members(root->all_baserels)` at this place; accessing the parse tree is no longer needed.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: In the most recent version of the PR, this function was removed and we don't access the `query->jointree` anymore.

akuzm (approved):
Thanks for all the fixes, I think it's almost ideal now :)

tsl/src/fdw/data_node_scan_plan.c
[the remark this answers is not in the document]
jnidzwetzki: Sorry for the imprecise answer. The `varno` is adjusted at this point 

For example, here you see the `exprs` before and after the adjustment. The `varno` of the expressions is translated from the input table position to the proper position of this specific DataNode entry in the simple_rel_array (the data node is at position 8 in this example). This is needed for the deparser to generate the SQL for the partition properly.

### Original List
```
(
   {VAR 
   :varno 1 
   :varattno 2 
   :vartype 23 
   :vartypmod -1 
   :varcollid 0 
   :varlevelsup 0 
   :varnosyn 1 
   :varattnosyn 2 
   :location -1
   [...]
)
```

### Adjusted List
```
(
   {VAR 
   :varno 8 
   :varattno 2 
   :vartype 23 
   :vartypmod -1 
   :varcollid 0 
   :varlevelsup 0 
   :varnosyn 8 
   :varattnosyn 2 
   :location -1
   }
   [...]
)
```

src/cross_module_fn.c
```diff
@@ -555,6 +555,7 @@ TSDLLEXPORT CrossModuleFunctions ts_cm_functions_default = {
 	.hypertable_distributed_set_replication_factor = error_no_default_fn_pg_community,
 	.update_compressed_chunk_relstats = update_compressed_chunk_relstats_default,
 	.health_check = error_no_default_fn_pg_community,
+	.mn_get_foreign_join_paths = NULL,
```
erimatnor, line 558: Following the pattern here, you might want to add a dummy function instead (I guess one that does nothing by default). Then you can avoid a NULL check when using the function.

tsl/src/fdw/data_node_scan_plan.c
erimatnor: Can we make dead code clear,  by wrapping it in `#if ENABLE_DEAD_CODE` or marking it with `pg_unreachable()`, or similar so that the code is there but you can avoid having code coverage complain and it is obvious to the reader that the code isn't executed.

tsl/src/fdw/data_node_scan_plan.c
erimatnor: Hmm, I don't see the assertion. I mean, the fact that his is a data node scan means it must be a distributed hypertable.

erimatnor (approved):
Approving, although I have some inline comments.

src/cross_module_fn.c
jnidzwetzki: @erimatnor Good point. I introduced a dummy function.

jnidzwetzki: @erimatnor Good point. I introduced a dummy function.

jnidzwetzki:
@erimatnor I like the idea with the `#ifdef ENABLE_DEAD_CODE` statements. I added them to the larger blocks of currently unused code.

tsl/src/fdw/data_node_scan_plan.c
jnidzwetzki: We are called by the generic PostgreSQL `set_join_pathlist_hook` planner hook and have only checked that `fdw_private` and `fdw_relation_info` [are not null](https://github.com/timescale/timescaledb/pull/5212/files#diff-5d89399a9796f5e49176485595f03e7ce25cea608cdd8417d8d91d00c57396faR1678). However, we have not checked if the relation is a (distributed) hypertable. I changed this statement to an assert, which was triggered during my tests. Unfortunately, this only occurred once, and I was unable to reproduce this behavior again.

Therefore, I reverted this to a `return` statement; I would like to make sure that the `is_distributed_hypertable` check is also performed in production code and I would like to be safe here (e.g., we are not processing a table that has a populated `fdw_private` field from another FDW extension).

PR #5214, opened 2023-01-24
Fix use of prepared statement in async module

erimatnor:
Broken code caused the async connection module to never send queries
using prepared statements. Instead, queries were sent using the
parameterized query statement instead.

The regression was introduced a long time ago in commit https://github.com/timescale/timescaledb/commit/86858e36e9fb6243d52191bcedc6be05ee2ab734

Fix this so that prepared statements are used when created.

### DIFF
--- a/tsl/src/remote/async.c
+++ b/tsl/src/remote/async.c
@@ -152,6 +152,8 @@ async_request_set_state(AsyncRequest *req, AsyncRequestState new_state)
 static AsyncRequest *
 async_request_send_internal(AsyncRequest *req, int elevel)
 {
+	int ret;
+
 	if (req->state != DEFERRED)
 		elog(elevel, "can't send async request in state \"%d\"", req->state);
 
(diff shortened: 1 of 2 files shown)


tsl/src/remote/async.c
```diff
@@ -162,6 +164,16 @@ async_request_send_internal(AsyncRequest *req, int elevel)
 	remote_connection_configure_if_changed(req->conn);
 
 	if (req->stmt_name)
+	{
+		ret = PQsendQueryPrepared(remote_connection_get_pg_conn(req->conn),
```
erimatnor, line 168: `PQsendQueryPrepared()` didn't exist in our code, so prepared statements were never executed.

pmwkaa (approved):
Is there a way to test it?

tsl/src/remote/async.c
```diff
@@ -234,7 +234,33 @@ async_request_send_prepare(TSConnection *conn, const char *sql, int n_params)
 		elog(ERROR, "cannot create prepared statement name");
 
 	req = async_request_create(conn, sql, stmt_name, n_params, NULL, FORMAT_TEXT);
-	req = async_request_send_internal(req, ERROR);
+
+	if (req->state != DEFERRED)
+		elog(ERROR, "can't send async request in state \"%d\"", req->state);
+
+	if (remote_connection_is_processing(req->conn))
```
jnidzwetzki, line 241: @erimatnor I am not sure if this case is handled correctly. As far as I understand, the statement is not prepared in this code path and the `AsyncRequest` is returned in `DEFERRED` state.

I could not find any code that deals with a returned `AsyncRequest` in this state (i.e., does `async_request_wait_prepared_statement` can deal with it?). Do I miss something?

erimatnor: You are right. Was too quick to adopt code from another place.

tsl/src/remote/async.c
```diff
@@ -162,6 +164,16 @@ async_request_send_internal(AsyncRequest *req, int elevel)
 	remote_connection_configure_if_changed(req->conn);
```
jnidzwetzki, line 164: @erimatnor Nit: This line is not touched by your PR. However, in Line 233, we check the return value of `remote_connection_configure_if_changed`. In this line, we are ignoring it. I would suggest checking the return value of the function also here.

jnidzwetzki (approved):
I added one comment, the rest looks good to me.

tsl/src/remote/async.c
erimatnor: Fixed

Issue #4185, opened 2022-03-20
[Bug]: assertion failure in get_appendrel_parampathinfo 


svenklemm:
### What type of bug is this?

Other

### What subsystems and features are affected?

Query planner

### What happened?

sqlsmith failed with an assertion failure

Link to failed run: https://github.com/timescale/timescaledb/runs/5608836995?check_suite_focus=true

### TimescaleDB version affected

master

### PostgreSQL version used

14.2

### What operating system did you use?

Ubuntu 20.04

### What installation method did you use?

Source

### What platform did you run on?

On prem/Self-hosted

### Relevant log output and stack trace

```bash
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
(gdb) #0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
        set = {__val = {4194304, 0, 0, 0, 0, 94835184952096, 
            4611686018427387904, 0, 0, 140028848755930, 0, 94835184952142, 
            128, 0, 140724912339648, 140028842829120}}
        pid = <optimized out>
        tid = <optimized out>
        ret = <optimized out>
#1  0x00007f5b015b9859 in __GI_abort () at abort.c:79
        save_stage = 1
        act = {__sigaction_handler = {sa_handler = 0x7f5b0174cbf7, 
            sa_sigaction = 0x7f5b0174cbf7}, sa_mask = {__val = {1, 
              140028843188254, 3, 140724912339828, 12, 140028843188258, 2, 
              3473460813870661632, 7292510206346934584, 140724912339936, 
              7291948136861554232, 94835309690808, 8248173052380786944, 
              140724912339920, 1596, 20}}, sa_flags = 308886496, 
          sa_restorer = 0x564089e05638}
        sigs = {__val = {32, 0 <repeats 15 times>}}
#2  0x0000564089ca8fb8 in ExceptionalCondition (
    conditionName=conditionName@entry=0x564089e05638 "bms_is_subset(appendrel->lateral_relids, required_outer)", 
    errorType=errorType@entry=0x564089cff00b "FailedAssertion", 
    fileName=fileName@entry=0x564089e056a8 "relnode.c", 
    lineNumber=lineNumber@entry=1596) at assert.c:69
No locals.
#3  0x0000564089ad58aa in get_appendrel_parampathinfo (
    appendrel=appendrel@entry=0x564090f23fb8, 
[...]

PR #5220, opened 2023-01-25
Run sqlsmith on all commits in main

akuzm:
This would help to gather more statistics. We have some rare assertion failures that happen with SQLSmith like #4185

### DIFF
--- a/.github/workflows/sqlsmith.yaml
+++ b/.github/workflows/sqlsmith.yaml
@@ -1,5 +1,5 @@
 name: SQLsmith
-on:
+"on":
   schedule:
     # run daily 2:00 on main branch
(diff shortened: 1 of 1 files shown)


.github/workflows/sqlsmith.yaml
```diff
@@ -1,12 +1,10 @@
 name: SQLsmith
 on:
-  schedule:
-    # run daily 2:00 on main branch
-    - cron: '0 2 * * *'
```
svenklemm, line 5: no harm in leaving the scheduled run in

jnidzwetzki (approved):

fabriziomello (approved):

PR #5227, opened 2023-01-26
Add a workflow for automated backporting

akuzm:
1. Find the latest release branch
1. For each commit in main and not in release branch (compared by title), find the corresponding PR.
1. If the PR fixes an issue labeled "bug", and neither the PR nor the issue are labeled "no-backport", cherry-pick the commits from the PR onto the release branch, and create a PR with these changes.

The resulting backport PRs: https://github.com/timescale/timescaledb/labels/pr-backport

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,38 @@
+name: Backport Bug Fixes
+on:
+  schedule:
+    # Run weekdays 12:00 on main branch, so that it doesn't wreak havoc on
+    # weekends.
+    - cron: '0 12 * * 1-5'
+  workflow_dispatch:
+  push:
+    branches:
+      # You can run and debug new versions of the backport script by pushing it
+      # to this branch. workflow_dispatch can only be run through github cli for
+      # branches that are not main, so it's inconvenient.
+      - backport/trigger
+
+jobs:
+  backport:
+    name: Backport Bug Fixes
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install Linux Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install pip
+
+      - name: Install Python Dependencies
+        run: |
+          pip install PyGithub requests
+
+      - name: Checkout TimescaleDB
+        uses: actions/checkout@v3
+
+      - name: Run the Backport Script
+        env:
+          GITHUB_TOKEN: ${{ secrets.ORG_AUTOMATION_TOKEN }}
+        run: |
+          git remote --verbose
+          scripts/backport.py
--- a/.github/workflows/changelog-check.yaml
+++ b/.github/workflows/changelog-check.yaml
@@ -2,7 +2,10 @@ name: Check CHANGELOG.md updated
 "on":
   pull_request:
     types: [opened, synchronize, reopened, edited]
-    branches: [main]
+    # It's important to check that the changelog is updated with bug fixes that
+    # we backport to the release branches, so these branches are included as
+    # well.
+    branches: [main, "[0-9]+.[0-9]+.x"]
 jobs:
   # Check that the CHANGELOG is updated by the pull request. This can be
   # disabled by adding a trailer line of the following form to the
--- a/scripts/backport.py
+++ b/scripts/backport.py
@@ -1,0 +1,378 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import string
+import subprocess
+import sys
+
+from github import Github  # This is PyGithub.
+import requests
+
+
+def run_query(query):
+    """A simple function to use requests.post to make the GraphQL API call."""
+
+    request = requests.post(
+        "https://api.github.com/graphql",
+        json={"query": query},
+        headers={"Authorization": f'Bearer {os.environ.get("GITHUB_TOKEN")}'},
+        timeout=20,
+    )
+    response = request.json()
+
+    # Have to work around the unique GraphQL convention of returning 200 for errors.
+    if request.status_code != 200 or "errors" in response:
+        raise Exception(
+            f"Query failed to run by returning code of {request.status_code}."
+            f"\nQuery: '{query}'"
+            f"\nResponse: '{request.json()}'"
+        )
+
+    return response
+
+
+def get_referenced_issue(pr_number):
+    """Get the number of issue fixed by the given pull request.
+    Returns None if no issue is fixed, or more than one issue"""
+
+    ref_result = run_query(
+        string.Template(
+            """
+        query {
+            repository(owner: "timescale", name: "timescaledb") {
+              pullRequest(number: $pr_number) {
+                closingIssuesReferences(first: 1) {
+                  edges {
+                    node {
+                      number
+                    }
+                  }
+                }
+              }
+            }
+          }
+          """
+        ).substitute({"pr_number": pr_number})
+    )
+
+    # The above returns {'data': {'repository': {'pullRequest': {'closingIssuesReferences': {'edges': [{'node': {'number': 4944}}]}}}}}
+
+    ref_edges = ref_result["data"]["repository"]["pullRequest"][
+        "closingIssuesReferences"
+    ]["edges"]
+
+    if ref_edges and len(ref_edges) == 1:
+        return ref_edges[0]["node"]["number"]
+
+    return None
+
+
+def set_auto_merge(pr_number):
+    """Enable auto-merge for the given PR"""
+
+    owner, name = target_repo_name.split("/")
+    # We first have to find out the PR id, which is some base64 string, different
+    # from its number.
+    query = string.Template(
+        """query {
+          repository(owner: "$owner", name: "$name") {
+            pullRequest(number: $pr_number) {
+              id
+            }
+          }
+        }"""
+    ).substitute(pr_number=pr_number, owner=owner, name=name)
+    result = run_query(query)
+    pr_id = result["data"]["repository"]["pullRequest"]["id"]
+
+    query = string.Template(
+        """mutation {
+            enablePullRequestAutoMerge(
+                input: {
+                    pullRequestId: "$pr_id",
+                    mergeMethod: REBASE
+                }
+            ) {
+                clientMutationId
+            }
+        }"""
+    ).substitute(pr_id=pr_id)
+    run_query(query)
+
+
+def git_output(command):
+    """Get output from the git command, checking for the successful exit code"""
+    return subprocess.check_output(f"git {command}", shell=True, text=True)
+
+
+def git_check(command):
+    """Run a git command, checking for the successful exit code"""
+    subprocess.run(f"git {command}", shell=True, check=True)
+
+
+def git_returncode(command):
+    """Run a git command, returning the exit code"""
+    return subprocess.run(f"git {command}", shell=True, check=False).returncode
+
+
+# The token has to have the "access public repositories" permission, or else creating a PR returns 404.
+github = Github(os.environ.get("GITHUB_TOKEN"))
+
+# If we are running inside Github Action, will modify the main repo.
+source_remote = "origin"
+source_repo_name = os.environ.get("GITHUB_REPOSITORY")
+target_remote = source_remote
+target_repo_name = source_repo_name
+
+if not source_repo_name:
+    # We are running manually for debugging, probably want to modify a fork.
+    source_repo_name = "timescale/timescaledb"
+    target_repo_name = os.environ.get("BACKPORT_TARGET_REPO")
+    target_remote = os.environ.get("BACKPORT_TARGET_REMOTE")
+    if not target_repo_name or not target_remote:
+        print(
+            "Please specify the target repositories for debugging, using the "
+            "environment variables BACKPORT_TARGET_REPO (e.g. `timescale/timescaledb`) "
+            "and BACKPORT_TARGET_REMOTE (e.g. `origin`).",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+print(
+    f"Will look at '{source_repo_name}' (git remote '{source_remote}') for bug fixes, "
+    f"and create the backport PRs in '{target_repo_name}' (git remote '{target_remote}')."
+)
+
+source_repo = github.get_repo(source_repo_name)
+target_repo = github.get_repo(target_repo_name)
+
+# Set git name and email corresponding to the token user.
+token_user = github.get_user()
+os.environ["GIT_COMMITTER_NAME"] = token_user.name
+
+# This is an email that is used by Github when you opt to hide your real email
+# address. It is required so that the commits are recognized by Github as made
+# by the user. That is, if you use a wrong e-mail, there won't be a clickable
+# profile picture next to the commit in the Github interface.
+os.environ[
+    "GIT_COMMITTER_EMAIL"
+] = f"{token_user.id}+{token_user.login}@users.noreply.github.com"
+print(
+    f"Will commit as {os.environ['GIT_COMMITTER_NAME']} <{os.environ['GIT_COMMITTER_EMAIL']}>"
+)
+
+# Fetch the sources
+git_check(f"fetch {source_remote}")
+git_check(f"fetch {target_remote}")
+
+# Find out what is the branch corresponding to the previous version compared to
+# main. We will backport to that branch.
+version_config = dict(
+    [
+        re.match(r"^(.+)\s+=\s+(.+)$", line).group(1, 2)
+        for line in git_output(f"show {source_remote}/main:version.config").splitlines()
+        if line
+    ]
+)
+
+previous_version = version_config["update_from_version"]
+previous_version_parts = previous_version.split(".")
+previous_version_parts[-1] = "x"
+backport_target = ".".join(previous_version_parts)
+
+print(f"Will backport to {backport_target}")
+
+# Find out which commits are unique to main and target branch. Also build sets of
+# the titles of these commits. We will compare the titles to check whether a
+# commit was backported.
+main_commits = [
+    line.split("\t")
+    for line in git_output(
+        f'log -1000 --pretty="format:%h\t%s" {source_remote}/{backport_target}..{source_remote}/main'
+    ).splitlines()
+    if line
+]
+main_titles = {x[1] for x in main_commits}
+
+branch_commits = [
+    line.split("\t")
+    for line in git_output(
+        f'log -1000 --pretty="format:%h\t%s" {source_remote}/main..{source_remote}/{backport_target}'
+    ).splitlines()
+    if line
+]
+branch_commit_titles = {x[1] for x in branch_commits}
+
+# We will do backports per-PR, because one PR, though not often, might contain
+# many commits. So as the first step, go through the commits unique to main, find
+# out which of them have to be backported, and remember the corresponding PRs.
+# We also have to remember which commits to backport. The list from PR itself is
+# not what we need, these are the original commits from the PR branch, and we
+# need the resulting commits in master. Store them as dict(pr number -> PRInfo).
+prs_to_backport = {}
+
+
+class PRInfo:
+    """Information about the PR to be backported."""
+
+    def __init__(self, pygithub_pr_, issue_number_):
+        self.pygithub_pr = pygithub_pr_
+        self.pygithub_commits = []
+        self.issue_number = issue_number_
+
+
+for commit_sha, commit_title in main_commits:
+    if commit_title in branch_commit_titles:
+        print(f"{commit_sha[:9]} '{commit_title}' is already in the branch.")
+        continue
+
+    pygithub_commit = source_repo.get_commit(sha=commit_sha)
+
+    pulls = pygithub_commit.get_pulls()
+    if not pulls:
+        print(f"{commit_sha[:9]} '{commit_title}' does not belong to a PR.")
+        continue
+
+    pull = pulls[0]
+    issue_number = get_referenced_issue(pull.number)
+    if not issue_number:
+        print(
+            f"{commit_sha[:9]} belongs to the PR #{pull.number} '{pull.title}' that does not close an issue."
+        )
+        continue
+
+    pr_labels = {label.name for label in pull.labels}
+    stopper_labels = pr_labels.intersection(["no-backport", "pr-auto-backport-failed"])
+    if stopper_labels:
+        print(
+            f"{commit_sha[:9]} belongs to the PR #{pull.number} '{pull.title}' labeled as '{list(stopper_labels)[0]}' which prevents automated backporting."
+        )
+        continue
+
+    changed_files = {file.filename for file in pull.get_files()}
+    stopper_files = changed_files.intersection(
+        ["sql/updates/latest-dev.sql", "sql/updates/reverse-dev.sql"]
+    )
+    if stopper_files:
+        print(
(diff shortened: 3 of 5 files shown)


.github/workflows/backport.yaml
```diff
@@ -0,0 +1,38 @@
+name: Backport Bug Fixes
+on:
+  schedule:
+    # run daily 2:00 on main branch
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+  push:
+    branches:
+      - aku/backport-trigger
```
akuzm, line 9: For debugging, don't forget to remove.

scripts/backport.py
```diff
@@ -0,0 +1,258 @@
...
+    if subprocess.run(
+        # f"git cherry-pick -m 1 -x {commit_list}",
+        f"git cherry-pick -m 1 -x {commit_list} > /dev/null 2>&1",
+        shell=True,
+    ).returncode:
+        subprocess.run(
+            f"git cherry-pick --abort > /dev/null", shell=True
+        ).check_returncode()
+        print(
+            f'Conflict while backporting pull request #{original_pr.number}: "{original_pr.title}"'
```
akuzm, line 233: Should also report it to the original PR/issue. Here we need to be careful not to spam it with comment, either check for comment existence or apply a label like 'auto-backport-failed'. I think the second way might be preferrable because it's easy to filter on it, to review what we might need to backport manually.

akuzm: Added a tag for now.

scripts/backport.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Push the branch and create the backport PR
+    subprocess.run(
+        f"git push {target_remote} @:refs/heads/{backport_branch} > /dev/null",
+        shell=True,
+    ).check_returncode()
+
+    backport_pr = target_repo.create_pull(
+        title=f"""
+            Backport to {backport_target}: #{original_pr.number}: {original_pr.title}
```
akuzm, line 248: Should also mention the issue that is being fixed.

svenklemm:
Can we have a sanity check that aborts when latest-dev/reverse-dev is touched since those would require manual intervention.

scripts/backport.py
```diff
@@ -0,0 +1,261 @@
...
+
+print(f"Will backport to {backport_target}")
+
+# Find out which commits are unique to main and target branch. Also build sets of
+# the titles of these commits. We will compare the titles to check whether a
+# commit was backported.
+main_commits = [
+    l.split("\t")
+    for l in subprocess.check_output(
+        f'git log -1000 --pretty="format:%h\t%s" {source_remote}/2.9.x..{source_remote}/main',
```
jnidzwetzki, line 112: Could we make the branch name dynamic (here and in Line 123)?

scripts/backport.py
```diff
@@ -0,0 +1,261 @@
...
+            Backport to {backport_target}: #{original_pr.number}: {original_pr.title}
+            Merge it after all the relevant CI checks pass.
+            If this fix should not be backported, or will be backported manually, just close this PR.
+        """,
+        body=f"This is an automated backport of #{original_pr.number}: {original_pr.title}",
+        head=backport_branch,
+        base=backport_target,
+    )
+    backport_pr.add_to_labels("pr-backport")
+    backport_pr.add_to_assignees(original_pr.user.login)
```
jnidzwetzki, line 257: Was this line added after you opened the test backport PRs? The open PRs are not assigned to anybody?

jnidzwetzki: A nice to have feature: is it be possible to enable auto merge for the PR? I mean, if it can be built and all tests are successful, it could be merged.

akuzm: > Was this line added after you opened the test backport PRs? The open PRs are not assigned to anybody?

Yes, I fixed this later.

> A nice to have feature: is it be possible to enable auto merge for the PR? I mean, if it can be built and all tests are successful, it could be merged.

Unfortunately, I didn't find a way to do this through Github API. I think to later add a "phase 3" to this script, which would just go through the open PRs marked as 'pr-backport' with green checks, and merges them. Probably in the next PR, because the API for checks seems to be tricky.

jnidzwetzki: Maybe this helps: https://docs.github.com/en/graphql/reference/mutations#enablepullrequestautomerge

akuzm: Thanks, that's why I didn't find it, it's the scary GraphQL API :) I try to stick to the RESTful v3 where possible. I'll take a look at it.

scripts/backport.py
akuzm: Thanks, fixed.

akuzm:
> Can we have a sanity check that aborts when latest-dev/reverse-dev is touched since those would require manual intervention.

Sure. They'll probably fail because of the conflicts anyway, but I added a check to ignore such PRs.

scripts/backport.py
akuzm: Although it got me thinking, auto-merge would require us setting the required checks for the release branches, do we want this? Currently we have none. Maybe it'll make things less convenient.

jnidzwetzki: Good point... On the other hand, do we want to be able to merge PRs if tests fail? Maybe just mark the tests as required and not require reviews in that branch?

akuzm: OK, let's try it, I added some branch-protection rules (regression and up/downgrade tests like in main), and the code to enable auto-merge.

akuzm: Github graphql returns status 200 on failed requests, and apparently this is a "common convention" :facepalm:

.github/workflows/backport.yaml
akuzm: I guess I'll make the name more generic and leave it in.

scripts/backport.py
```diff
@@ -0,0 +1,334 @@
...
+              id
+            }
+          }
+        }"""
+    ).substitute(pr_number=pr_number, owner=owner, name=name)
+    result = run_query(query)
+    pr_id = result["data"]["repository"]["pullRequest"]["id"]
+
+    query = string.Template(
+        'mutation { enablePullRequestAutoMerge( input: { pullRequestId: "$pr_id", mergeMethod: REBASE } ) { clientMutationId } }'
```
jnidzwetzki, line 90: Nit: Could you format the GraphQL query like the ones above?

jnidzwetzki (approved):
Before this PR is merged, I would recommend to:

* Mention this action in our eng-database call. So, everybody is aware of this new behavior.
* Update the release documentation.

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/checkout@v3
+
+      - name: Run the Backport Script
+        env:
+          GITHUB_TOKEN: ${{ secrets.ORG_AUTOMATION_TOKEN }}
+          BACKPORT_TARGET_REPO: timescale/timescaledb
+          BACKPORT_TARGET_REMOTE: origin
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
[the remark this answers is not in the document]
akuzm: These credentials are needed so that the git commits show up as made by the github actions bot. I added a comment.

scripts/backport.py
```diff
@@ -0,0 +1,334 @@
...
+    ).substitute(pr_id=pr_id)
+    run_query(query)
+
+
+# The token has to have the "access public repositories" permission, or else creating a PR returns 404.
+github = Github(os.environ.get("GITHUB_TOKEN"))
+source_repo = github.get_repo("timescale/timescaledb")
```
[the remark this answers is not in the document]
akuzm: Yes, let's use the default env variable `GITHUB_ACTION_REPOSITORY`.

.github/workflows/backport.yaml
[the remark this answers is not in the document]
jnidzwetzki: I was curious about the user id. But `41898282` [seems to be](https://api.github.com/user/41898282) the official Bot account.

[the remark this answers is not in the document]
mkindahl: There is a [timescale automation user](https://github.com/timescale-automation) now if that helps, but the secret is not updated yet.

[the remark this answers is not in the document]
akuzm: Can you update the secret, please? I'll remove these credentials from here and will get the credentials corresponding to the token using the github api.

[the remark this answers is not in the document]
akuzm: All updated and works.

svenklemm (approved):

akuzm:
Added a couple of lines to the wiki: https://github.com/timescale/eng-database/wiki/Releasing-TimescaleDB#automated-cherry-picking-of-bug-fixes

scripts/backport.py
```diff
@@ -0,0 +1,371 @@
...
+    for x in [
+        line.split("=")
+        for line in subprocess.check_output(
+            f"git show {source_remote}/main:version.config",
+            shell=True,
+            text=True,
+        ).splitlines()
+        if line
+    ]
+}
```
mkindahl, line 165: A little easier to read. You can also split it into multiple lines.
```suggestion
# Find out what is the branch corresponding to the previous version compared to
# main. We will backport to that branch.
version_config = dict(
    re.split(r"\s+=\s+", line) 
     for line in subprocess.check_output(
         f"git show {source_remote}/main:version.config",
         shell=True,
         text=True,
     ).splitlines()
     if line
)
```

scripts/backport.py
```diff
@@ -0,0 +1,371 @@
...
+previous_version = version_config["update_from_version"]
+previous_version_parts = previous_version.split(".")
+previous_version_parts[-1] = "x"
+backport_target = ".".join(previous_version_parts)
+
+print(f"Will backport to {backport_target}")
+
+# Find out which commits are unique to main and target branch. Also build sets of
+# the titles of these commits. We will compare the titles to check whether a
+# commit was backported.
```
mkindahl, line 176: Doesn't this risk missing some commits if they have similar titles?

Since we never modify the master, can't we use the SHA on the master to decide if we should backport the commit or not. We could add the original SHA either to the commit message using `interpret-trailer` or to a note using `git notes`. (I would prefer the former since that does not require special treatment to push and pull notes.)

scripts/backport.py
```diff
@@ -0,0 +1,371 @@
...
+    if line
+]
+branch_titles = {x[1] for x in branch_commits}
+
+# We will do backports per-PR, because one PR, though not often, might contain
+# many commits. So as the first step, go through the commits unique to main, find
+# out which of them have to be backported, and remember the corresponding PRs.
+# We also have to remember which commits to backport. The list from PR itself is
+# not what we need, these are the original commits from the PR branch, and we
+# need the resulting commits in master. Store them as dict(pr number -> PRINfo).
```
mkindahl, line 204: Hmm.... quite often PRs contain multiple commit messages because they are fixing separate issues but are added to a single PR for convenience. In that case, it might make sense to cherry-pick them separately and create a PR for each.

scripts/backport.py
```diff
@@ -0,0 +1,371 @@
...
+            f"git rev-parse {target_remote}/{backport_branch} > /dev/null 2>&1",
+            shell=True,
+            check=False,
+        ).returncode
+        == 0
+    ):
+        print(
+            f'Backport branch {backport_branch} for PR #{original_pr.number}: "{original_pr.title}" already exists. Skipping.'
+        )
+        continue
```
mkindahl, line 303: It is a lot easier to read if you split out the git commands into separate functions and use them in the code.

scripts/backport.py
```diff
@@ -0,0 +1,371 @@
...
+
+    # Push the branch and create the backport PR
+    subprocess.run(
+        f"git push --quiet {target_remote} @:refs/heads/{backport_branch} > /dev/null 2>&1",
+        shell=True,
+        check=True,
+    )
+
+    original_description = original_pr.body
+    # Comment out the keywords like 'Fixes #1234' to avoid confusion.
```
mkindahl, line 334: Not sure what confusion you have in mind.

It is good if the backport PRs reference the issue as well so that you can track that it was actually backported. This will allow you to eliminate that our backporting system is making a mistake.

mkindahl (commented):
I have some questions about the workflow below that might be good to discuss.

It would be better to split it up into more functions to make it easier to maintain. In the current state, it is quite hard to read and follow the code.

scripts/backport.py
```diff
@@ -0,0 +1,371 @@
...
+
+source_repo = github.get_repo(source_repo_name)
+target_repo = github.get_repo(target_repo_name)
+
+# Set git name and email corresponding to the token user.
+token_user = github.get_user()
+os.environ["GIT_COMMITTER_NAME"] = token_user.name
+os.environ[
+    "GIT_COMMITTER_EMAIL"
+] = f"{token_user.id}+{token_user.login}@users.noreply.github.com"
```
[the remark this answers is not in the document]
akuzm: This is an email that is used by Github when you opt to hide your real email address. It is required so that the commits are recognized by Github as made by the user. That is, if you use a wrong e-mail, there won't be a clickable profile picture next to the commit in the Github interface.

scripts/backport.py
akuzm: The linter tells me to use dict comprehension instead of `dict()`, but re.split might be nicer indeed, I'll try to incorporate it.

scripts/backport.py
akuzm: Similar -- no. Exactly the same titles -- yes, they will be missed. But how likely this is in practice? I think the bug fixes usually have very distinct titles. I like the titles because this doesn't require any parsing.

```
$ git log --oneline --pretty=format:%s origin/main | sort | uniq -c | sort -nr | head -100
     18 Prepare the repo for next development cycle
      5 Prepare the repo for the next development cycle
      5 Add missing gitignore entry
      4 Fix flaky pg_dump test
      4 Bump postgres versions used in CI
      3 Remove unused functions
      3 Fix typos in comments
      3 Fix memory overflow
      2 Update CHANGELOG
      2 Upadting docs
      2 Run pgindent on code
      2 Remove unused functions from utils.c
      2 Remove dead code
      2 Release 2.0.0
      2 Move enterprise features to community
      2 Make older_than / newer_than error message user friendly
      2 Fix various misspellings
      2 Fix issues discovered by coverity
      2 Fix flaky bgw_custom test
      2 Fix compression_ddl isolation test
      2 Fix CAgg on CAgg variable bucket size validation
      2 Fix assertion failure in cursor_fetcher_rewind
      2 cleanup
```

scripts/backport.py
akuzm: I think we can't really handle that, because we can't understand which commit fixes which issue. One might need to be backported, and the other not. I can disallow PRs with multiple commits.

scripts/backport.py
mkindahl: Adding the SHA of the original commit as a trailer would not require any parsing either. It would also allow finding the original commit easy to make sure that no mistakes were done.

scripts/backport.py
mkindahl: No strong opinion, just a suggestion.

scripts/backport.py
akuzm: I don't think so... the first thing I need to know when I look at this code is which git command it is running. Putting this command somewhere else just makes it more complicated. I agree these lines look kinda ugly, especially when formatted by "black", but it's all just boilerplate around the only meaningful thing, the git command.

I can instead move boilerplate to functions, e.g. `git_check_output` and `git_returncode`, to remove the amount of visual noise. Thankfully I only need two these modes for subprocess.

scripts/backport.py
akuzm: It's going to reference the issue, but I don't want Github to treat it as a "fixes" reference. Granted, most of the time it's going to be closed already, so it's not important. But there might be some weird case when the original issue was reopened, but the backport of the old fix is in flight, and then it closes the issue. Or maybe we would want to reason about backporting another way, starting from issues, and then it would help to have only one PR that fixes an issue. I think it is more straightforward to keep the "issue"<->"PR that fixes it" a one-to-one correspondence.

akuzm: > It's going to reference the issue

e.g. it says "Original issue #xxxx" in the description of the backport PR

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Checkout TimescaleDB
+        uses: actions/checkout@v3
+
+      - name: Run the Backport Script
+        env:
+          GITHUB_TOKEN: ${{ secrets.ORG_AUTOMATION_TOKEN }}
+        run: |
+          git remote --verbose
+          scripts/backport.py
```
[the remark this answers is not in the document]
akuzm: Python 2 is long deprecated, no? And environment is under our control, this workflow requires `ubuntu-latest`. I will change the shebang to `#/usr/bin/env python3` because the `.py` script is not compatible with python 2. Can't really tell what's the recommended approach these days: https://peps.python.org/pep-0394/

.github/workflows/changelog-check.yaml
```diff
@@ -2,7 +2,10 @@ name: Check CHANGELOG.md updated
 on:
   pull_request:
     types: [opened, synchronize, reopened, edited]
-    branches: [main]
+    # It's important to check that the changelog is updated with bug fixes that
+    # we backport to the release branches, so these branches are included as
+    # well.
+    branches: [main, "*.?.x"]
```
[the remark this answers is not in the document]
akuzm: Oh, we don't match 2.10.x with both of these. Should be `[0-9]+.[0-9]+.x` I guess: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

scripts/backport.py
akuzm: Now I'm using `cherry-pick -x`, which adds `(cherry picked from commit c5e496a554e9f4d04578f39669108554c22c918d)` to the end of the message. I think we have to support this kind of trailer, because this is a git-standard thing that you can also use when cherry-picking manually.

So matching by SHA is going to require:
1) some minimal parsing (e.g. re.match);
2) human operators not forgetting to add `-x` when they cherry-pick manually -- this one I'm worried about the most;
3) some special handling when you actually do a different manual fix in the release branch. You'll have to pretend it's a cherry-pick not to confuse the automation.

Using titles solves all three problems, because it's easy and natural to use the same title for manual cherry-pick/separate fix, and instead of parsing I just do `git log --pretty=format:%s`.

Let's observe for a while how problematic this title matching is going to be?

.github/workflows/changelog-check.yaml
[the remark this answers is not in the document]
mkindahl: Right.

.github/workflows/backport.yaml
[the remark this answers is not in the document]
mkindahl: I usually prefer to play it safe, but if you feel confident it will always be python3, go with it.

scripts/backport.py
akuzm: > Thankfully I only need two these modes for subprocess.

Actually three, but anyway, done that, looks better now.

.github/workflows/backport.yaml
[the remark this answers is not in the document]
akuzm: Changed the shebang to require python3.

scripts/backport.py
akuzm: Wrote it even simpler with capture groups.

scripts/backport.py
akuzm: > I can skip PRs with multiple commits.

Did that.

mkindahl: > I think we can't really handle that, because we can't understand which commit fixes which issue.

Why not? Each commit would contain a "fixed" label in this case, so it should be straightforward to cherry-pick these and create separate PRs?

> One might need to be backported, and the other not. I can skip PRs with multiple commits.

Not sure why one should be backported and the other one not. Both are committed in a PR, so both should be part of the release.

scripts/backport.py
mkindahl: > e.g. it says "Original issue #xxxx" in the description of the backport PR

That works for me.

scripts/backport.py
akuzm: > Not sure why one should be backported and the other one not. Both are committed in a PR, so both should be part of the release.

If both should be backported together, separate PRs seem to contradict it? 

In the case you described above, we have two commits fixing two different issues in one PR. The current rule for backporting is "if the fixed issue is labeled 'bug' and not labeled 'no-backport'". These labels can differ for these issues, so the decision to backport will also differ.

I think it is more convenient to make backport decision on a per-PR basis. Commits often don't have "fixes" references (and I think it's good, because they tend to spam the issues endlessly at each rebase). We will also need more code to support both per-PR and per-commit basis, and pretty tricky logic to combine these approaches, because they might easily give different results.

akuzm:
Oh my, I forgot to disable the auto merge :sweat_smile:

scripts/backport.py
mkindahl: > Now I'm using `cherry-pick -x`, which adds `(cherry picked from commit c5e496a554e9f4d04578f39669108554c22c918d)` to the end of the message. I think we have to support this kind of trailer, because this is a git-standard thing that you can also use when cherry-picking manually.
> 
> So matching by SHA is going to require:
> 
>     1. some minimal parsing (e.g. re.match);

Yeah, this one is simpler.
 
>     2. human operators not forgetting to add `-x` when they cherry-pick manually -- this one I'm worried about the most;

Since human operators can be quite creative in what they are doing we're probably going to need some separate CI checks for release branches any way.
 
>     3. some special handling when you actually do a different manual fix in the release branch. You'll have to pretend it's a cherry-pick not to confuse the automation.

Not sure that is a big problem. If you do a manual fix, you would probably do something like this anyway:

```
$ git cherry-pick -x <whatever>
<hack hack hack>
$ git commit -a -m'My tweaks'
<test test test>
$ git rebase -i upstream/main
<merge it all, and keep the original message>
```

> Using titles solves all three problems, because it's easy and natural to use the same title for manual cherry-pick/separate fix, and instead of parsing I just do `git log --pretty=format:%s`.
> 
> Let's observe for a while how problematic this title matching is going to be?

Seems it's already decided. :)

scripts/backport.py
mkindahl: > In the case you described above, we have two commits fixing two different issues in one PR. The current rule for backporting is "if the fixed issue is labeled 'bug' and not labeled 'no-backport'". These labels can differ for these issues, so the decision to backport will also differ.

But then they should not be merged as a single PR. They should actually be separate PRs.

To be able to backport on a per-PR basis, it is important that if a PR consists of multiple commits, it is because they are **impossible** to submit as separate PRs. In that case, and only in that case, you can safely assume that the entire PR should be backported.

mkindahl: > Commits often don't have "fixes" references (and I think it's good, because they tend to spam the issues endlessly at each rebase).

All mine have. :)

mkindahl: > I think it is more convenient to make backport decision on a per-PR basis. [...] We will also need more code to support both per-PR and per-commit basis, and pretty tricky logic to combine these approaches, because they might easily give different results.

I am leaning towards the per-PR basis for another reason than "more code to support both". It is in some situations necessary to have two commits for a single change. For example, if you rename a lot of files and then modify them, it is better to have one commit for the renaming and one for the following changes. In this case, the PR would be the glue for these two PRs.

PR #5237, opened 2023-01-27
Run yamllint in CI 

akuzm:
Helps against some weird YAML errors when working with Github actions.


fabriziomello:
Mostly changes are just reformatting right? If yes would be good to add it to `.git-blame-ignore-revs` as well!

fabriziomello (approved):

akuzm:
> Mostly changes are just reformatting right? If yes would be good to add it to .git-blame-ignore-revs as well!

Yep, I'll add in a follow-up PR because the hashes are going to change after merge.

svenklemm (approved):

.github/workflows/windows-packages.yaml
```diff
@@ -109,10 +119,13 @@ jobs:
...
-        ~/PostgreSQL/${{ matrix.pg }}/bin/psql -U postgres -d postgres -X -c "CREATE EXTENSION timescaledb;SELECT extname,extversion,version() FROM pg_extension WHERE extname='timescaledb';"
-        $installed_version = ~/PostgreSQL/${{ matrix.pg }}/bin/psql -U postgres -d postgres -X -t -c "SELECT extversion FROM pg_extension WHERE extname='timescaledb';"
+        ~/PostgreSQL/${{ matrix.pg }}/bin/psql -U postgres -d postgres -X `
+          -c "CREATE EXTENSION timescaledb" `
+          -c "SELECT extname,extversion,version() FROM pg_extension WHERE extname='timescaledb'"
+        $installed_version = ~/PostgreSQL/${{ matrix.pg }}/bin/psql -U postgres `
+          -d postgres -X -t -c "SELECT extversion FROM pg_extension WHERE extname='timescaledb';"
         $installed_version = $installed_version.Trim()
-        if (${installed_version} -notmatch "${{ steps.version.outputs.version }}")
+        if ("${installed_version}" -ne "${{ steps.version.outputs.version }}")
```
akuzm, line 128: Also fixing this check while I'm at it. Somehow the version was empty, and this check always returned true because it was using `-nomatch` which means "doesn't match a regular expression", which in turn is always false for an empty regex.

akuzm:
All the pr, release, prerelease and coverity checks seem to be in order.

PR #5239, opened 2023-01-27
Fix next_start calculation for fixed schedules 

### LINKED ISSUE #5216: [Bug]: add_job() is not respecting scheduled interval when it's big enough
### What type of bug is this?

Incorrect result

Background worker

### What happened?

When creating a user defined action and adding a background job to execute this action in background, the next_start field is getting wrong after the first execution, if the provided scheduled interval is big enough and fixed_scheduled is true.

For example, I've created the user defined action "events.compress_hypertable", and then added it to the background scheduler by doing:

```
select add_job('events.compress_hypertable', '10y', config => '{"hypertable": "xxxx_evt", "schema":"events"}',
			  check_config => 'events.check_jobs_config', fixed_schedule => true)
```

This job runs once as expected, but the next start is set to 2036-01-25.

Simirlarly, if I change the scheduled interval to 10 months, the resulting next start is 2024-05-26.

For 5 years, it goes to 2031-01-24.

But, when the interval is "small", the next start get updated correctly, with for example, 1h, 1 month or 1 hour. It only happens when the interval is considerably spaced.

Bellow is a screenshot demonstrating this issue. I've not changed anything in the job after adding it.

I'm currently using timescaledb 2.9.0.

To give you additional info, the hypettable I'm testing has only 3 events, the the action defined compress its chunks with:
`select compress_chunk(i, true) from show_chunks('events.xxxx_evt') as i`

This action runs really fast, as the hypertable is small. It's just for testing purposes.

If you need any additional info, please feel free to ask.

I couldn't note this inconsistent behavior when using fixed_scheduled with FALSE. Only happens when it's TRUE.

Screenshot from 2023-01-24 16-19-09

2.9.0

### PostgreSQL version used

14

Ubuntu 20

Docker

Not applicable

_No response_

```bash
- Create an hypertable and enable compression

- Create a user defined action that compresses this hypertable by doing
select compress_chunk(i, true) from show_chunks('hypertable_name') as i

- add this action to be run in the background using the add_job() API, providing a scheduled interval considerably spaced (for example, 10 months). Use the default fixed_schedule, that is, TRUE.

- query the information view timescaledb_information.jobs, and the next_start column will be incosistent with the scheduled interval provided.
```

konskov:
This patch fixes several issues with next_start calculation.

- Previously, the offset was added twice in some cases.
This is fixed by this patch.

- Additionally, schedule intervals with month components
were not handled correctly.
Internally, time_bucket with origin is used to calculate
the next start. However, in the case of month intervals, the
timestamp calculated for a bucket is always aligned on the first
day of the month, regardless of origin.
Therefore, previously the result was aligned with origin by adding
the difference between origin and its respective time bucket.
This difference was computed as a fixed length interval in terms
of days and time. That computation led to incorrect computation of
next start occasionally, for example when a job should be executed on
the last day of a month.
That is fixed by adding an appropriate interval of months to
initial_start and letting Postgres handle this computation properly.

Fixes #5216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ accidentally triggering the load of a previous DB version.**
 * #5218 Add role-level security to job error log
 * #5214 Fix use of prepared statement in async module
 * #5290 Compression can't be enabled on continuous aggregates when segmentby/orderby columns need quotation
+* #5239 Fix next_start calculation for fixed schedules
 
 ## 2.9.3 (2023-02-03)
 
@@ -34,6 +35,7 @@ upgrade as soon as possible.
 **Thanks**
 * @ssmoss for reporting issues on continuous aggregates
 * @jaskij for reporting the compliation issue that occurred with clang
+
 
 ## 2.9.2 (2023-01-26)
 
--- a/src/bgw/job_stat.c
+++ b/src/bgw/job_stat.c
@@ -170,61 +170,105 @@ typedef struct
 } JobResultCtx;
 
 /*
- * time_bucket(schedule_interval, finish_time, origin => initial_start)
+ * logic is the following
+ * Ideally we would return
+ * time_bucket(schedule_interval, finish_time, origin => initial_start, timezone).
+ * That is what we return when the schedule interval does not have month components.
+ * However, when there is a month component in the schedule interval,
+ * then supplying the origin in time_bucket
+ * does not work and the returned bucket is aligned on the start of the month.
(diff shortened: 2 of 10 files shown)


mkindahl (commented):
You need a better commit message, in particular, the first line.

konskov:
Fixed the commit message I hope @mkindahl ?

mahipv (approved):
The patch looks good. Thank you for the extensive test case.

svenklemm (approved):

PR #5244, opened 2023-01-30
Small decompress_chunk refactor

svenklemm:
Refactor the decompression code to move the decompressor initialization into a separate function.

Disable-check: commit-count

### DIFF
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -10,11 +10,12 @@
 #include <access/htup_details.h>
 #include <access/multixact.h>
 #include <access/xact.h>
+#include <catalog/heap.h>
+#include <catalog/index.h>
 #include <catalog/namespace.h>
+#include <catalog/pg_am.h>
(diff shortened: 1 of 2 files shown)


tsl/src/compression/compression.c
```diff
@@ -1541,74 +1578,48 @@ decompress_chunk(Oid in_table, Oid out_table)
 	 */
 	Relation in_rel = relation_open(in_table, ExclusiveLock);
 
-	TupleDesc in_desc = RelationGetDescr(in_rel);
-	TupleDesc out_desc = RelationGetDescr(out_rel);
-
 	Oid compressed_data_type_oid = ts_custom_type_cache_get(CUSTOM_TYPE_COMPRESSED_DATA)->type_oid;
```
akuzm, line 1581: It's the same regardless of the table, right? Makes sense to move it inside the build function as well.

akuzm (approved):

PR #5245, opened 2023-01-30
Life-cycle manage connections via memory contexts

erimatnor:
Tie the life cycle of connections to the memory context they are created on. Previously, connections were automatically closed at the end of a transaction, although often a connection needs to live beyond a single transaction. For example, a connection cache is maintained for data node connections, and, for such cases, a flag was set on a connection to avoid closing it automatically.

Instead of tying connections to transactions, they are now life-cycle managed via memory contexts. This simplifies the handling of connections and avoids having to create exceptions to closing connections at transaction end.


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ accidentally triggering the load of a previous DB version.**
 
 **Features**
 * #5241 Allow RETURNING clause when inserting into compressed chunks
+* #5245 Life-cycle manage connections via memory contexts
```
nikkhils, line 11: `life-cycle manage` -> Manage life-cycle of connections via memory contexts?

nikkhils (approved):
LGTM overall.

pmwkaa (approved):

PR #5246, opened 2023-01-30
Make connection establishment interruptible

### LINKED ISSUE #2757: Make data node connection establishment non-blocking
Data node connections are established with a blocking API, which means that, e.g., the `add_data_node` command cannot be canceled (ctrl-c) if the data node is not responding.

An easy way to test this manually is to run a listening netcat process that pretends to be a data node endpoint:

```bash
netcat -l 6543
```

Then call `add_data_node` connecting to this IP and port and the result is a hanging that cannot be canceled.

Non-blocking connection establishment was implemented as part https://github.com/timescale/timescaledb/pull/2751, but it was reverted due to causing flaky tests in Appveyor. This needs further investigation.

erimatnor:
Refactor the data node connection establishment so that it is interruptible, e.g., by ctrl-c or `statement_timeout`.

Previously, the connection establishment used blocking libpq calls. By instead using asynchronous connection APIs and integrating with PostgreSQL interrupt handling, the connection establishment can be canceled by an interrupt caused by a statement timeout or a user.

Fixes #2757

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ accidentally triggering the load of a previous DB version.**
 **Features**
 * #5241 Allow RETURNING clause when inserting into compressed chunks
 * #5245 Mange life-cycle of connections via memory contexts
+* #5246 Make connection establishment interruptible
 
 **Bugfixes**
 * #4804 Skip bucketing when start or end of refresh job is null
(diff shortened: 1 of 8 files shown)


nikkhils (approved):
LGTM!

pmwkaa (approved):

PR #5252, opened 2023-01-30
Support ON CONFLICT DO NOTHING on compressed hypertables

svenklemm:
This patch allows unique constraints on compressed chunks. When trying to INSERT into compressed chunks with unique constraints any potentially conflicting compressed batches will be decompressed to let postgres do constraint checking on the INSERT. With this patch only INSERT ON CONFLICT DO NOTHING will be supported. For decompression only segment by information is considered to determine conflicting batches. This will be enhanced in a follow-up patch to also include orderby metadata to require decompressing less batches.

### DIFF
--- a/src/copy.c
+++ b/src/copy.c
@@ -324,6 +324,7 @@ TSCopyMultiInsertBufferFlush(TSCopyMultiInsertInfo *miinfo, TSCopyMultiInsertBuf
 	ChunkInsertState *cis =
 		ts_chunk_dispatch_get_chunk_insert_state(miinfo->ccstate->dispatch,
 												 buffer->point,
+												 buffer->slots[0],
 												 NULL /* on chunk changed function */,
 												 NULL /* payload for on chunk changed function */);
 
@@ -948,6 +949,7 @@ copyfrom(CopyChunkState *ccstate, List *range_table, Hypertable *ht, MemoryConte
 		/* Find or create the insert state matching the point */
 		cis = ts_chunk_dispatch_get_chunk_insert_state(dispatch,
 													   point,
+													   myslot,
 													   on_chunk_insert_state_changed,
 													   bistate);
 
--- a/src/nodes/chunk_dispatch/chunk_dispatch.c
+++ b/src/nodes/chunk_dispatch/chunk_dispatch.c
@@ -58,6 +58,7 @@ destroy_chunk_insert_state(void *cis)
  */
 extern ChunkInsertState *
 ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
+										 TupleTableSlot *slot,
 										 const on_chunk_changed_func on_chunk_changed, void *data)
 {
 	ChunkInsertState *cis;
@@ -91,6 +92,7 @@ ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
 		 * where the chunk already exists.
 		 */
 		bool found;
+		Assert(slot);
 		Chunk *chunk = ts_hypertable_find_chunk_for_point(dispatch->hypertable, point);
 
 #if PG14_GE
@@ -106,7 +108,12 @@ ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
 			chunk = ts_hypertable_create_chunk_for_point(dispatch->hypertable, point, &found);
 		}
 		else
+		{
 			found = true;
+		}
+
+		if (!chunk)
+			elog(ERROR, "no chunk found or created");
 
 		/* get the filtered list of "available" DNs for this chunk but only if it's replicated */
 		if (found && dispatch->hypertable->fd.replication_factor > 1)
@@ -126,11 +133,25 @@ ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
 			list_free(chunk_data_nodes);
 		}
 
-		if (!chunk)
-			elog(ERROR, "no chunk found or created");
-
 		cis = ts_chunk_insert_state_create(chunk, dispatch);
 		ts_subspace_store_add(dispatch->cache, chunk->cube, cis, destroy_chunk_insert_state);
+
+		if (found && ts_chunk_is_compressed(chunk) && !ts_chunk_is_distributed(chunk))
+		{
+			/*
+			 * If this is an INSERT into a compressed chunk with UNIQUE or
+			 * PRIMARY KEY constraints we need to make sure any batches that could
+			 * potentially lead to a conflict are in the decompressed chunk so
+			 * postgres can do proper constraint checking.
+			 */
+			if (ts_cm_functions->decompress_batches_for_insert)
+				ts_cm_functions->decompress_batches_for_insert(cis, chunk, slot);
+			else
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("functionality not supported under the current \"%s\" license",
+								ts_guc_license)));
+		}
 
 		MemoryContextSwitchTo(old_context);
 	}
@@ -308,6 +329,7 @@ chunk_dispatch_exec(CustomScanState *node)
 	/* Find or create the insert state matching the point */
 	cis = ts_chunk_dispatch_get_chunk_insert_state(dispatch,
 												   point,
+												   slot,
 												   on_chunk_insert_state_changed,
 												   state);
 
--- a/src/process_utility.c
+++ b/src/process_utility.c
@@ -2679,7 +2679,7 @@ process_index_start(ProcessUtilityArgs *args)
 	hcache = ts_hypertable_cache_pin();
 	ht = ts_hypertable_cache_get_entry_rv(hcache, stmt->relation);
 
-	if (NULL == ht)
+	if (!ht)
 	{
 		/* Check if the relation is a Continuous Aggregate */
 		cagg = ts_continuous_agg_find_by_rv(stmt->relation);
@@ -2702,7 +2702,7 @@ process_index_start(ProcessUtilityArgs *args)
 			}
 		}
 
-		if (NULL == ht)
+		if (!ht)
 		{
 			ts_cache_release(hcache);
 			return DDL_CONTINUE;
@@ -2715,18 +2715,6 @@ process_index_start(ProcessUtilityArgs *args)
 
 		/* Make the RangeVar for the underlying materialization hypertable */
 		stmt->relation = makeRangeVar(NameStr(ht->fd.schema_name), NameStr(ht->fd.table_name), -1);
-	}
-	else if (TS_HYPERTABLE_HAS_COMPRESSION_ENABLED(ht))
-	{
-		/* unique indexes are not allowed on compressed hypertables*/
-		if (stmt->unique || stmt->primary || stmt->isconstraint)
-		{
-			ts_cache_release(hcache);
-			ereport(ERROR,
-					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
-					 errmsg("operation not supported on hypertables that have compression "
-							"enabled")));
-		}
 	}
 
 	ts_hypertable_permissions_check_by_id(ht->fd.id);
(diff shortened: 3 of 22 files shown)


src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -106,8 +108,27 @@ ts_chunk_dispatch_get_chunk_insert_state(ChunkDispatch *dispatch, Point *point,
 			chunk = ts_hypertable_create_chunk_for_point(dispatch->hypertable, point, &found);
 		}
 		else
+		{
 			found = true;
 
+			if (ts_chunk_is_compressed(chunk) && chunk->data_nodes == NIL && slot)
```
akuzm, line 114: slot can't be null, right? Might be more clear to convert the above Assert into Ensure if it's just to prevent segfaults

tsl/src/compression/compression.c
```diff
@@ -814,6 +817,7 @@ index_scan_sequence_number(Relation table_rel, Oid index_oid, ScanKeyData *scank
 	index_scan->xs_want_itup = true;
 	index_rescan(index_scan, scankey, num_scankeys, NULL, 0);
 
+	/* TODO: add sanity check */
```
akuzm, line 820: Can you expand on this?

tsl/src/compression/compression.c
```diff
@@ -2042,3 +2048,199 @@ update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
...
+
+	RowDecompressor decompressor = build_decompressor(in_rel, out_rel);
+	Bitmapset *unique_columns = relation_unique_columns(out_rel);
+
+	int num_scankeys;
+	ScanKeyData *scankeys =
+		build_scankeys(chunk->fd.hypertable_id, decompressor, unique_columns, slot, &num_scankeys);
+
+	TableScanDesc heapScan =
+		table_beginscan(in_rel, GetTransactionSnapshot(), num_scankeys, scankeys);
```
akuzm, line 2204: This is a seq scan, probably we want to have an index scan?

tsl/src/compression/compression.c
```diff
@@ -2042,3 +2048,199 @@ update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
...
+
+	for (HeapTuple compressed_tuple = heap_getnext(heapScan, ForwardScanDirection);
```
akuzm, line 2206: Suspiciously low coverage for this loop, is it codecov mistake or missing tests?

tsl/src/compression/compression.c
```diff
@@ -2042,3 +2048,199 @@ update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
...
+				TypeCacheEntry *tce =
+					lookup_type_cache(decompressor.in_desc->attrs[attno - 1].atttypid,
+									  TYPECACHE_EQ_OPR_FINFO);
+
+				if (!tce || !OidIsValid(tce->eq_opr_finfo.fn_oid))
+					elog(ERROR,
+						 "no equality function for type \"%s\"",
+						 format_type_be(decompressor.in_desc->attrs[attno - 1].atttypid));
+
+				if (isnull)
```
akuzm, line 2147: Would be nice to have coverage for null value in segmentby column as well.

tsl/src/compression/compression.c
svenklemm: We need to support both, you cant have scankeys on non-index column in indexscan, right?
But indexscan support will not be part of initial PR

akuzm: Right. But Postgres requires there to be an index for the ON CONFLICT columns, so we should have one. I think it's OK with seq scan for starters, but performance won't be good and we probably need to add index scan soon.

akuzm: Oh, it's more difficult with orderby columns, we don't have min/max in the default index for them.

src/nodes/chunk_dispatch/chunk_dispatch.c
svenklemm: there is an assert earlier that is never hit and i think that is leftover from an earlier version

tsl/src/compression/compression.c
svenklemm: that isnt even code added by this PR, removed

tsl/src/compression/compression.c
svenklemm: the index will be on the uncompressed part though, not the compressed part

tsl/src/compression/compression.c
svenklemm: tests are WIP

tsl/src/compression/compression.c
svenklemm: done

tsl/src/compression/compression.c
```diff
@@ -2042,3 +2047,226 @@ update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
...
+
+		heap_deform_tuple(compressed_tuple,
+						  decompressor.in_desc,
+						  decompressor.compressed_datums,
+						  decompressor.compressed_is_nulls);
+
+		row_decompressor_decompress_row(&decompressor);
+
+		TM_FailureData tmfd;
+		table_tuple_delete(in_rel,
```
erimatnor, line 2246: Suggest handling return value here, or at least asserting the result if only one possibility.

tsl/src/compression/compression.c
```diff
@@ -2042,3 +2047,226 @@ update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
...
+										   decompressor,
+										   unique_columns,
+										   &null_columns,
+										   slot,
+										   &num_scankeys);
+
+	TableScanDesc heapScan =
+		table_beginscan(in_rel, GetTransactionSnapshot(), num_scankeys, scankeys);
+
+	for (HeapTuple compressed_tuple = heap_getnext(heapScan, ForwardScanDirection);
```
erimatnor, line 2211: Heap scans like this without a per-tuple memory context make me suspicious. I worry that there's a risk this allocates memory proportional to the size of the table. You should double check the memory usage, especially for big tables.

src/copy.c
```diff
@@ -324,6 +324,7 @@ TSCopyMultiInsertBufferFlush(TSCopyMultiInsertInfo *miinfo, TSCopyMultiInsertBuf
 	ChunkInsertState *cis =
 		ts_chunk_dispatch_get_chunk_insert_state(miinfo->ccstate->dispatch,
 												 buffer->point,
+												 buffer->slots[0],
```
akuzm, line 327: This looks incorrect, we should have decompressed everything we needed. Let's move decompression out of this function? It looks error-prone in cases like this one, when you just want to re-create the insert state, but don't actually want to decompress anything. Or when you use some batching, get the chunk insert state only once, but then have to call decompression for every row.

tsl/src/compression/compression.c
svenklemm: the row decompressor switches to a per tuple memory context

src/copy.c
svenklemm: Hmm in the presence of unique constraints you have to do this check for every tuple. Cause different tuples routed to the same chunk might still require different batches to be decompressed. Where would you move this to?

tsl/src/compression/compression.c
akuzm: Let's discuss the indexes a little more. I think the implementation with seq scan is OK to merge for starters, but it's not something we should publish. Doing a seq scan for each inserted row is going to be prohibitively slow.

I think it is reasonable for us to require an index on the decompressed chunk that matches the ON CONFLICT clause. It's the same thing that Postgres requires.

1) We already create the index on segmentby columns, so it always can be used.
2) For min/max, can probably use a seq scan?
3) Ideally, if Postgres managed to find an arbiter index on the uncompressed chunk, we should have a matching index on the decompressed chunk. Not sure if we do anything to guarantee that currently.

Do we have a plan for this somewhere?

src/copy.c
akuzm: To `chunk_dispatch_exec` and `copyfrom`, just after they look up the insert state. And here, in `TSCopyMultiInsertBufferFlush`, you flush the multi-insert buffers with tuples that you have already seen in `copyfrom`, so you don't have to do decompression for them again.

tsl/src/compression/compression.c
svenklemm: There will never be a matching index if you have unique constraints. Since unique constraints need to include the partitioning columns that means time will be part of the index but you will never want time as segmentby .

tsl/src/compression/compression.c
```diff
@@ -2042,3 +2047,226 @@ update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
...
+		ReleaseSysCache(index_tuple);
+	}
+
+	list_free(indexoidlist);
+
+	return unique_columns;
+}
+
+static ScanKeyData *
+build_scankeys(int32 hypertable_id, RowDecompressor decompressor, Bitmapset *unique_columns,
```
akuzm, line 2099: This part looks like "planning-time", similar to the deduction of arbiter indexes. But probably there's no convenient way to do it there. I think it would be good not do it per-row and cache this in the chunk insert state.

svenklemm: This needs to be per row as the scankeys are built from row values. I dont think you can easily cache this.

akuzm: Postgres index scans do that by just replacing the `sk_argument` (`ExecIndexEvalRuntimeKeys`).

akuzm (approved):

src/process_utility.c
```diff
@@ -2667,7 +2667,7 @@ process_index_start(ProcessUtilityArgs *args)
 	hcache = ts_hypertable_cache_pin();
 	ht = ts_hypertable_cache_get_entry_rv(hcache, stmt->relation);
 
-	if (NULL == ht)
+	if (!ht)
```
konskov, line 2670: nit: I think these changes as well as https://github.com/timescale/timescaledb/pull/5252/files#diff-038649a1ce33b21f89e080d566755f896778f29c2ae7c5d9a1bf6e7975b5f48cR1409 could be made separately in a different PR, if they are necessary?

tsl/src/compression/compression.c
```diff
@@ -2042,3 +2047,226 @@ update_compressed_chunk_relstats(Oid uncompressed_relid, Oid compressed_relid)
...
+			 * we have no metadata.
+			 */
+
+			if (COMPRESSIONCOL_IS_SEGMENT_BY(fd))
+			{
+				bool isnull;
+				Datum value = slot_getattr(slot, attno, &isnull);
+
+				TypeCacheEntry *tce =
+					lookup_type_cache(decompressor.in_desc->attrs[attno - 1].atttypid,
```
konskov, line 2138: `attno` is taken from the uncompressed chunk. is it guaranteed to be the same for the compressed? or should we be using the attname again to get the attno for the compressed chunk?

tsl/src/compression/compression.c
```diff
@@ -1503,6 +1507,7 @@ typedef struct RowDecompressor
 
 	TupleDesc out_desc;
 	Relation out_rel;
+	CatalogIndexState indexstate;
```
konskov, line 1510: this is not used anywhere yet, right?

tsl/src/compression/compression.c
svenklemm: good find, we do indeed need to do proper mapping here

antekresic: Any reason not to convert `attno - 1` to `AttrNumberGetAttrOffset(attno)`?

tsl/src/compression/compression.c
svenklemm: correct, removing

tsl/src/compression/compression.c
svenklemm: i'll leave that for a followup PR

tsl/src/compression/compression.c
svenklemm: done

tsl/src/compression/compression.c
svenklemm: done

konskov:
approving, I think this should be 1 commit though, right?

konskov (approved):

Issue #4958, opened 2022-11-10
Graceful command exit using CTRL+C on MN


vineethapai:

nikkhils:
Also check that setting `statement_timeout` on AN also cancels queries gracefully on the DNs

pmwkaa:
Might be related https://github.com/timescale/Support-Dev-Collab/issues/682

horzsolt:
@erimatnor , @nikkhils do you think this will fix that test [failure](https://github.com/timescale/timescaledb/issues/4972)?

svenklemm:
multinode has been removed

PR #5253, opened 2023-01-30
Make data node command execution interruptible

erimatnor:
The function to execute remote commands on data nodes used a blocking libpq API that doesn't integrate with PostgreSQL interrupt handling, making it impossible for a user or statement timeout to cancel a remote command.

Refactor the remote command execution function to use a non-blocking API and integrate with PostgreSQL signal handling via WaitEventSets.

Partial fix for #4958.

### DIFF
--- a/tsl/src/data_node.c
+++ b/tsl/src/data_node.c
@@ -257,7 +257,6 @@ data_node_get_connection(const char *const data_node, RemoteTxnPrepStmtOption co
 	const ForeignServer *server;
 	TSConnectionId id;
 
-	Assert(data_node != NULL);
 	server = data_node_get_foreign_server(data_node, ACL_NO_CHECK, false, false);
 	id = remote_connection_id(server->serverid, GetUserId());
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ accidentally triggering the load of a previous DB version.**
 * #5241 Allow RETURNING clause when inserting into compressed chunks
 * #5245 Mange life-cycle of connections via memory contexts
 * #5246 Make connection establishment interruptible
+* #5253 Make data node command execution interruptible
 
 **Bugfixes**
 * #4804 Skip bucketing when start or end of refresh job is null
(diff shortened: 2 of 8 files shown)


tsl/src/data_node.c
```diff
@@ -2113,3 +2113,33 @@ data_node_name_list_check_acl(List *data_node_names, AclMode mode)
...
+	TSConnection *conn = data_node_get_connection(nodename, REMOTE_TXN_NO_PREP_STMT, true);
+	const text *cmdstr;
+	PGresult *res;
+
+	if (PG_ARGISNULL(1))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("command string cannot be NULL")));
+
+	cmdstr = PG_GETARG_TEXT_P(1);
```
nikkhils, line 2135: This is not safe I feel. What if the cmdstr contains a `DML` command?

sql/data_node.sql
```diff
@@ -10,3 +10,6 @@ CREATE OR REPLACE FUNCTION _timescaledb_internal.remote_txn_heal_data_node(forei
 RETURNS INT
 AS '@MODULE_PATHNAME@', 'ts_remote_txn_heal_data_node'
 LANGUAGE C STRICT;
+
+CREATE OR REPLACE PROCEDURE _timescaledb_internal.data_node_exec(node_name NAME, command TEXT)
```
pmwkaa, line 14: Why are we exposing it, for testing?

pmwkaa (approved):

nikkhils (approved):
Eric explained that the new API has similar restrictions to distributed_exec. So fine. LGTM.

tsl/src/data_node.c
erimatnor: Not sure why that is an issue. Any command execution is still subject to regular permission control, so a user can only do things it is allowed to.

Already, you can use `distributed_exec` to execute any statement on data nodes.

sql/data_node.sql
erimatnor: yes, for testing. I can limit it to debug builds I guess.

PR #5256, opened 2023-01-31
Make remote COPY calls interruptible

erimatnor:
Refactor COPY code for both "in" and "out" mode so that it is non-blocking and interruptible. This will allow canceling queries and ingest paths that use COPY between the access node and data nodes. Canceling can be done using statement timeouts or user
interrupts (ctrl-c).

Closes https://github.com/timescale/timescaledb/issues/4958


erimatnor:
The code changes were split into a smaller set of PRs.

PR #5261, opened 2023-01-31
Improve Realtime Continuous Aggregate performance

### LINKED ISSUE #4699: Improve planning performance of real-time continuous aggs
### What type of bug is this?

Performance issue

Continuous aggregate

### What happened?

Hello everybody,

We are currently in the process of building a solution for candle stick (ohlc) data from different crypto exchanges (binance/coinbase etc.) and we noticed that we have very slow (magnitudes slower than expected) planning times on the caggs (with real time aggregation on), than we have on the time bucket query on the original hypertable + the cagg (without real time agg on). In our use case we have very bursty workloads, we are expecting around 500 - 1000 requests in approximately the same 1-10 second period. The drop off the performance is especially noticeable at around the 100+ queries in parallel on these views.

The raw data in the hypertables is around:
* 300-400 Gigabyte uncompressed
* 120 Gigabyte compressed

We did a lot of initial investigation for choosing the correct chunk size for each exchange (since they have different amount of instruments). We base our assumption currently on a server with 4 cpu cores/16gb ram (+1 replica). We looked up best practices for the chunks and found the recommendation from the timescale team that you should be able to fit one chunk of all your hypertables in around 25% (shared_buffers) of the available memory. The following chunks in each table reflect that (below 400 Megabyte for each chunk).

Our current setup consists of the following example schemas (with binance as an example):

* Hypertable, we are using an integer based column for the partitioning (unix timestamp in milliseconds)

```
CREATE TABLE candles_binance_60
(
    start       BIGINT          NOT NULL,
    instrument  TEXT            NOT NULL,
    open        NUMERIC(30, 15) NOT NULL,
    high        NUMERIC(30, 15) NOT NULL,
    low         NUMERIC(30, 15) NOT NULL,
    close       NUMERIC(30, 15) NOT NULL,
    volume      NUMERIC(30, 15) NOT NULL,
    -- other non relevant fields
);

SELECT create_hypertable('candles_binance_60', 'start', chunk_time_interval => 43200000);
SELECT set_integer_now_func('candles_binance_60', 'unix_milliseconds');
```

* Indices on the hypertable
```
create unique index on candles_binance_60(instrument, start DESC);
```

* Example view (we have different intervals 300, 900, ..., 86400)

```
CREATE MATERIALIZED VIEW candles_binance_300
    SELECT
        time_bucket(300000, start) AS bucket_start,
        instrument,
        FIRST(open, start) AS "open",
        MAX(high) AS high,
        MIN(low) AS low,
        LAST(close, start) AS "close",
        SUM(volume) as volume,
        -- other non relevant fields
    FROM candles_binance_60
    GROUP BY instrument, bucket_start
WITH NO DATA;
```

* Compression on hypertable + views (after 2 months)

```
ALTER TABLE candles_binance_60
SET (
    timescaledb.compress,
    timescaledb.compress_segmentby = 'instrument'
);
SELECT add_compression_policy('candles_binance_60', 5184000000);

-- Views
ALTER MATERIALIZED VIEW candles_binance_300 SET (
    timescaledb.compress = true
);
SELECT add_compression_policy('candles_binance_300', 5184000000);
```

After we inserted all the raw data into the hypertables, we created and refreshed all the views, to make sure we have data in the materialised view. After this we enabled compression and let the background jobs finish.

When we did our first test benchmarks on the caggs with different intervals, we noticed a very big increase in planning time, which became especially noticeable the more queries (100+) were running in parallel on a single cagg. Below is an example query and the explain which causes a very high planning time:

query:
 ```
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS) 
SELECT *
FROM candles_binance_900
WHERE instrument = 'FILUSDT' and bucket_start between 1659783328000 and 1661580254000
ORDER BY bucket_start
LIMIT 30000
```

explain:
```
Limit  (cost=1163.26..1242.98 rows=2007 width=92) (actual time=10.353..10.789 rows=1997 loops=1)
  Output: _materialized_hypertable_29.bucket_start, _materialized_hypertable_29.instrument, _materialized_hypertable_29.open, _materialized_hypertable_29.high, _materialized_hypertable_29.low, _materialized_hypertable_29.close, _materialized_hypertable_29.volume, _materialized_hypertable_29.trades, _materialized_hypertable_29.buy_volume, _materialized_hypertable_29.sell_volume, _materialized_hypertable_29.buys, _materialized_hypertable_29.sells, _materialized_hypertable_29.vwap, _materialized_hypertable_29.missed
  Buffers: shared hit=52 read=14
  I/O Timings: read=8.976
  ->  Merge Append  (cost=1163.26..1242.98 rows=2007 width=92) (actual time=10.352..10.662 rows=1997 loops=1)
        Sort Key: _materialized_hypertable_29.bucket_start
        Buffers: shared hit=52 read=14
        I/O Timings: read=8.976
        ->  Sort  (cost=1153.25..1158.21 rows=1985 width=92) (actual time=10.350..10.509 rows=1997 loops=1)
              Output: _materialized_hypertable_29.bucket_start, _materialized_hypertable_29.instrument, _materialized_hypertable_29.open, _materialized_hypertable_29.high, _materialized_hypertable_29.low, _materialized_hypertable_29.close, _materialized_hypertable_29.volume, _materialized_hypertable_29.trades, _materialized_hypertable_29.buy_volume, _materialized_hypertable_29.sell_volume, _materialized_hypertable_29.buys, _materialized_hypertable_29.sells, _materialized_hypertable_29.vwap, _materialized_hypertable_29.missed
              Sort Key: _materialized_hypertable_29.bucket_start
              Sort Method: quicksort  Memory: 500kB
              Buffers: shared hit=52 read=14
              I/O Timings: read=8.976
              ->  Custom Scan (ChunkAppend) on _timescaledb_internal._materialized_hypertable_29  (cost=0.43..1024.67 rows=1985 width=92) (actual time=0.875..9.905 rows=1997 loops=1)
                    Output: _materialized_hypertable_29.bucket_start, _materialized_hypertable_29.instrument, _materialized_hypertable_29.open, _materialized_hypertable_29.high, _materialized_hypertable_29.low, _materialized_hypertable_29.close, _materialized_hypertable_29.volume, _materialized_hypertable_29.trades, _materialized_hypertable_29.buy_volume, _materialized_hypertable_29.sell_volume, _materialized_hypertable_29.buys, _materialized_hypertable_29.sells, _materialized_hypertable_29.vwap, _materialized_hypertable_29.missed
                    Startup Exclusion: true
                    Runtime Exclusion: false
                    Buffers: shared hit=52 read=14
                    I/O Timings: read=8.976
                    ->  Index Scan using _hyper_29_19961_chunk__materialized_hypertable_29_instrument_bu on _timescaledb_internal._hyper_29_19961_chunk  (cost=0.43..471.21 rows=913 width=92) (actual time=0.874..2.664 rows=916 loops=1)
                          Output: _hyper_29_19961_chunk.bucket_start, _hyper_29_19961_chunk.instrument, _hyper_29_19961_chunk.open, _hyper_29_19961_chunk.high, _hyper_29_19961_chunk.low, _hyper_29_19961_chunk.close, _hyper_29_19961_chunk.volume, _hyper_29_19961_chunk.trades, _hyper_29_19961_chunk.buy_volume, _hyper_29_19961_chunk.sell_volume, _hyper_29_19961_chunk.buys, _hyper_29_19961_chunk.sells, _hyper_29_19961_chunk.vwap, _hyper_29_19961_chunk.missed
                          Index Cond: ((_hyper_29_19961_chunk.instrument = 'FILUSDT'::text) AND (_hyper_29_19961_chunk.bucket_start < COALESCE(_timescaledb_internal.cagg_watermark(29), '-9223372036854775808'::bigint)) AND (_hyper_29_19961_chunk.bucket_start >= '1659783328000'::bigint) AND (_hyper_29_19961_chunk.bucket_start <= '1661580254000'::bigint))
                          Buffers: shared hit=23 read=7
                          I/O Timings: read=2.398
                    ->  Index Scan using _hyper_29_19962_chunk__materialized_hypertable_29_instrument_bu on _timescaledb_internal._hyper_29_19962_chunk  (cost=0.43..489.49 rows=951 width=92) (actual time=0.052..0.762 rows=960 loops=1)
                          Output: _hyper_29_19962_chunk.bucket_start, _hyper_29_19962_chunk.instrument, _hyper_29_19962_chunk.open, _hyper_29_19962_chunk.high, _hyper_29_19962_chunk.low, _hyper_29_19962_chunk.close, _hyper_29_19962_chunk.volume, _hyper_29_19962_chunk.trades, _hyper_29_19962_chunk.buy_volume, _hyper_29_19962_chunk.sell_volume, _hyper_29_19962_chunk.buys, _hyper_29_19962_chunk.sells, _hyper_29_19962_chunk.vwap, _hyper_29_19962_chunk.missed
                          Index Cond: ((_hyper_29_19962_chunk.instrument = 'FILUSDT'::text) AND (_hyper_29_19962_chunk.bucket_start < COALESCE(_timescaledb_internal.cagg_watermark(29), '-9223372036854775808'::bigint)) AND (_hyper_29_19962_chunk.bucket_start >= '1659783328000'::bigint) AND (_hyper_29_19962_chunk.bucket_start <= '1661580254000'::bigint))
                          Buffers: shared hit=28 read=1
                          I/O Timings: read=0.304
                    ->  Index Scan using _hyper_29_19963_chunk__materialized_hypertable_29_instrument_bu on _timescaledb_internal._hyper_29_19963_chunk  (cost=0.43..63.97 rows=121 width=90) (actual time=2.347..6.344 rows=121 loops=1)
                          Output: _hyper_29_19963_chunk.bucket_start, _hyper_29_19963_chunk.instrument, _hyper_29_19963_chunk.open, _hyper_29_19963_chunk.high, _hyper_29_19963_chunk.low, _hyper_29_19963_chunk.close, _hyper_29_19963_chunk.volume, _hyper_29_19963_chunk.trades, _hyper_29_19963_chunk.buy_volume, _hyper_29_19963_chunk.sell_volume, _hyper_29_19963_chunk.buys, _hyper_29_19963_chunk.sells, _hyper_29_19963_chunk.vwap, _hyper_29_19963_chunk.missed
                          Index Cond: ((_hyper_29_19963_chunk.instrument = 'FILUSDT'::text) AND (_hyper_29_19963_chunk.bucket_start < COALESCE(_timescaledb_internal.cagg_watermark(29), '-9223372036854775808'::bigint)) AND (_hyper_29_19963_chunk.bucket_start >= '1659783328000'::bigint) AND (_hyper_29_19963_chunk.bucket_start <= '1661580254000'::bigint))
                          Buffers: shared hit=1 read=6
                          I/O Timings: read=6.275
        ->  GroupAggregate  (cost=10.01..64.47 rows=22 width=298) (actual time=0.001..0.001 rows=0 loops=1)
              Output: (time_bucket('900000'::bigint, candles_binance_60.start)), candles_binance_60.instrument, first(candles_binance_60.open, candles_binance_60.start), max(candles_binance_60.high), min(candles_binance_60.low), last(candles_binance_60.close, candles_binance_60.start), sum(candles_binance_60.volume), sum(candles_binance_60.trades), sum(candles_binance_60.buy_volume), sum(candles_binance_60.sell_volume), sum(candles_binance_60.buys), sum(candles_binance_60.sells), CASE WHEN (sum(candles_binance_60.volume) > '0'::numeric) THEN (sum((candles_binance_60.vwap * candles_binance_60.volume)) / sum(candles_binance_60.volume)) ELSE '0'::numeric END, max(candles_binance_60.missed)
              Group Key: candles_binance_60.instrument, (time_bucket('900000'::bigint, candles_binance_60.start))
              ->  Custom Scan (ConstraintAwareAppend)  (cost=10.01..62.93 rows=22 width=76) (actual time=0.000..0.001 rows=0 loops=1)
                    Output: (time_bucket('900000'::bigint, candles_binance_60.start)), candles_binance_60.instrument, candles_binance_60.open, candles_binance_60.start, candles_binance_60.high, candles_binance_60.low, candles_binance_60.close, candles_binance_60.volume, candles_binance_60.trades, candles_binance_60.buy_volume, candles_binance_60.sell_volume, candles_binance_60.buys, candles_binance_60.sells, candles_binance_60.vwap, candles_binance_60.missed
                    Hypertable: candles_binance_60
                    Chunks excluded during startup: 22
Query Identifier: 2417736485277390874
Planning:
  Buffers: shared hit=84615
Planning Time: 13711.166 ms <-- Extrem planning times
Execution Time: 11.281 ms
```

This affected all views. After turning off the real time aggregation: `timescaledb.materialized_only = true`, we saw the high planning times go away. So we started to dig deeper and looked into the explain from the view again (with real time aggregation on) and found the following line, which we think caused the slow down:

```
_hyper_29_19961_chunk.bucket_start < COALESCE(_timescaledb_internal.cagg_watermark(29), '-9223372036854775808'::bigint))
```

So we tried to replicate the same benchmark, just running with `_timescaledb_internal.cagg_watermark(29)` to see how this function behaves under load. (Note: id 29 -> materialised hypertable of one of the binance views)

Example query, we ran with 100+ queries in parallel:

```
SELECT _timescaledb_internal.cagg_watermark(29);
```

After running it in parallel, we immediately noticed now the exact same behaviour we experience on the real time aggregation on the views. But instead of the planning time, we now see the increased `execution_time` since i assume this query is used in the planning phase of timescaledb for the real time aggregations.

After looking more into the source code, we noticed that it tries to load the `cagg_watermark` function (basically max(bucket) on the underlying hypertable) for each query on the view with real time aggregation, which causes an extreme load on the database and results in very slow response times. It also seems it is holding a lock on certain tables, which could block the other queries from resuming.

We also saw there have already been attempts to cache the `watermark`, but it seems they will get invalidated on each new transaction/query: https://github.com/timescale/timescaledb/pull/2828 

In order to confirm our suspicion, we also run the exact same queries that the real time aggression would do on a view (but with real time aggregation off, to avoid the `cagg_watermark` function call):
* select rows on cagg until last 3 elements in time bucket
* select with group by + all the other aggregate functions (min/first/last/max) on the underlying hypertable for the last 3 elements (which have not been materialised)

We now saw an improved query times in the form of up to 5-10x faster query times, even when running them in parallel on the database.

With real time on:
```
100 samples of 100 events
Cumulative:	20m39.62135204s
HMean:		12.383585422s
Avg.:		12.39621352s
p50: 		12.435145364s
p75:		12.73965338s
p95:		12.938279098s
p99:		12.979909272s
p999:		13.096055059s
Long 5%:	12.989388337s
Short 5%:	11.597400698s
Max:		13.096055059s
Min:		11.366040485s
Range:		1.730014574s
StdDev:		393.364365ms
Rate/sec.:	0.08
```

Without real time + query on the underlying hypertable (both in parallel):
```
100 samples of 100 events
Cumulative:	47.728964203s
HMean:		24.56293ms
Avg.:		238.644821ms
p50: 		138.6285ms
p75:		472.18237ms
p95:		606.845871ms
p99:		673.48164ms
p999:		703.028222ms
Long 5%:	652.111827ms
Short 5%:	3.135312ms
Max:		703.028222ms
Min:		2.445328ms
Range:		700.582894ms
StdDev:		222.952233ms
Rate/sec.:	4.19
```

Real time functionality of the views is therefore basically unusable and we have to revert back to querying the cagg with real time off + the underlying hypertable.

Is this behaviour intended? Or is this a configuration issue? If it is intended is there any planned improvement on the real time aggregation + watermark function? Could the watermark of `select max(time) from <underlying table>` not be cached, until the next insert into that table happens?

Happy to hear any insights/solutions to our problem! 

2.8.0

### PostgreSQL version used

14.5

Timescale Cloud

Other

Timescale Cloud

_No response_

```bash
See schema above
```

fabriziomello:
When calling the `cagg_watermark` function to get the watermark of a
Continuous Aggregate we execute a `SELECT MAX(time_dimension)` query
in the underlying materialization hypertable.

The problem is that a `SELECT MAX(time_dimention)` query can be
expensive because it will scan all hypertable chunks increasing the
planning time for a Realtime Continuous Aggregates.

Improved it by creating a new catalog table to serve as a cache table
to store the current Continous Aggregate watermark in the following
situations:
- Create CAgg: store the minimum value of hypertable time dimension
  data type;
- Refresh CAgg: store the last value of the time dimension materialized
  in the underlying materialization hypertable (or the minimum value of
  materialization hypertable time dimension data type if there's no
  data materialized);
- Drop CAgg Chunks: the same as refresh cagg.

Closes #4699, #5307

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ accidentally triggering the load of a previous DB version.**
 
 **Features**
 * #5212 Allow pushdown of reference table joins
+* #5221 Improve Realtime Continuous Aggregate performance
 * #5312 Add timeout support to the ping_data_node()
 * #5361 Add parallel support for partialize_agg()
 * #5252 Improve unique constraint support on compressed hypertables
@@ -26,6 +27,7 @@ accidentally triggering the load of a previous DB version.**
 **Thanks**
 * @nikolaps for reporting an issue with the COPY fetcher
 * @S-imo-n for reporting the issue on Background Worker Scheduler crash
+* @kovetskiy and @DZDomi for reporting peformance regression in Realtime Continuous Aggregates
 
 ## 2.10.1 (2023-03-07)
 
--- a/sql/pre_install/tables.sql
+++ b/sql/pre_install/tables.sql
@@ -404,6 +404,18 @@ CREATE TABLE _timescaledb_catalog.continuous_aggs_invalidation_threshold (
 
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_invalidation_threshold', '');
 
+CREATE TABLE _timescaledb_catalog.continuous_aggs_watermark (
+  mat_hypertable_id integer NOT NULL,
+  watermark bigint NOT NULL,
+  -- table constraints
+  CONSTRAINT continuous_aggs_watermark_pkey PRIMARY KEY (mat_hypertable_id),
+  CONSTRAINT continuous_aggs_watermark_mat_hypertable_id_fkey FOREIGN KEY (mat_hypertable_id) REFERENCES _timescaledb_catalog.continuous_agg (mat_hypertable_id) ON DELETE CASCADE
+);
+
+SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_watermark', '');
+
+
+
 -- this does not have an FK on the materialization table since INSERTs to this
 -- table are performance critical
 CREATE TABLE _timescaledb_catalog.continuous_aggs_hypertable_invalidation_log (
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -65,6 +65,7 @@
 #include "ts_catalog/chunk_data_node.h"
 #include "ts_catalog/compression_chunk_size.h"
 #include "ts_catalog/continuous_agg.h"
+#include "ts_catalog/continuous_aggs_watermark.h"
 #include "ts_catalog/hypertable_data_node.h"
 #include "utils.h"
 
@@ -3893,7 +3894,7 @@ ts_chunk_do_drop_chunks(Hypertable *ht, int64 older_than, int64 newer_than, int3
 	Chunk *chunks;
 	const char *schema_name, *table_name;
 	const int32 hypertable_id = ht->fd.id;
-	bool has_continuous_aggs;
+	bool has_continuous_aggs, is_materialization_hypertable;
 	const MemoryContext oldcontext = CurrentMemoryContext;
 	ScanTupLock tuplock = {
 		.waitpolicy = LockWaitBlock,
@@ -3913,13 +3914,17 @@ ts_chunk_do_drop_chunks(Hypertable *ht, int64 older_than, int64 newer_than, int3
 	 * well. Do not unlock - let the transaction semantics take care of it. */
 	lock_referenced_tables(ht->main_table_relid);
 
+	is_materialization_hypertable = false;
+
 	switch (ts_continuous_agg_hypertable_status(hypertable_id))
 	{
 		case HypertableIsMaterialization:
 			has_continuous_aggs = false;
+			is_materialization_hypertable = true;
 			break;
 		case HypertableIsMaterializationAndRaw:
 			has_continuous_aggs = true;
+			is_materialization_hypertable = true;
 			break;
 		case HypertableIsRawTable:
 			has_continuous_aggs = true;
@@ -4027,6 +4032,14 @@ ts_chunk_do_drop_chunks(Hypertable *ht, int64 older_than, int64 newer_than, int3
 			ChunkDataNode *cdn = lfirst(lc);
 			data_nodes = list_append_unique_oid(data_nodes, cdn->foreign_server_oid);
 		}
+	}
+
+	/* When dropping chunks for a given CAgg then force set the watermark */
+	if (is_materialization_hypertable)
+	{
+		bool isnull;
+		int64 watermark = ts_hypertable_get_open_dim_max_value(ht, 0, &isnull);
+		ts_cagg_watermark_update(ht, watermark, isnull, true);
 	}
 
 	if (affected_data_nodes)
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -2902,7 +2902,7 @@ ts_hypertable_func_call_on_data_nodes(const Hypertable *ht, FunctionCallInfo fci
 /*
  * Get the max value of an open dimension.
  */
-Datum
+int64
 ts_hypertable_get_open_dim_max_value(const Hypertable *ht, int dimension_index, bool *isnull)
 {
 	StringInfo command;
@@ -2910,11 +2910,14 @@ ts_hypertable_get_open_dim_max_value(const Hypertable *ht, int dimension_index, 
 	int res;
 	bool max_isnull;
 	Datum maxdat;
+	Oid timetype;
 
 	dim = hyperspace_get_open_dimension(ht->space, dimension_index);
 
 	if (NULL == dim)
 		elog(ERROR, "invalid open dimension index %d", dimension_index);
+
+	timetype = ts_dimension_get_partition_type(dim);
 
 	/*
 	 * Query for the last bucket in the materialized hypertable.
@@ -2941,7 +2944,7 @@ ts_hypertable_get_open_dim_max_value(const Hypertable *ht, int dimension_index, 
 				 (errmsg("could not find the maximum time value for hypertable \"%s\"",
 						 get_rel_name(ht->main_table_relid)))));
 
-	Ensure(SPI_gettypeid(SPI_tuptable->tupdesc, 1) == ts_dimension_get_partition_type(dim),
+	Ensure(SPI_gettypeid(SPI_tuptable->tupdesc, 1) == timetype,
 		   "partition types for result (%d) and dimension (%d) do not match",
 		   SPI_gettypeid(SPI_tuptable->tupdesc, 1),
 		   ts_dimension_get_partition_type(dim));
@@ -2953,7 +2956,7 @@ ts_hypertable_get_open_dim_max_value(const Hypertable *ht, int dimension_index, 
 	if ((res = SPI_finish()) != SPI_OK_FINISH)
 		elog(ERROR, "SPI_finish failed: %s", SPI_result_code_string(res));
 
-	return maxdat;
+	return max_isnull ? ts_time_get_min(timetype) : ts_time_value_to_internal(maxdat, timetype);
 }
 
 bool
--- a/src/ts_catalog/continuous_aggs_watermark.c
+++ b/src/ts_catalog/continuous_aggs_watermark.c
@@ -1,0 +1,402 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+
+/*
+ * This file handles continuous aggs watermark functions.
+ */
+
+#include <postgres.h>
+#include <access/xact.h>
+#include <fmgr.h>
+#include <miscadmin.h>
+#include <utils/acl.h>
+
+#include "ts_catalog/continuous_agg.h"
+#include "ts_catalog/continuous_aggs_watermark.h"
+#include "hypertable.h"
+
+typedef struct ContinuousAggregateWatermark
+{
+	int32 mat_hypertable_id;
+	MemoryContext mctx;
+	MemoryContextCallback cb;
+	CommandId cid;
+	int64 value;
+} ContinuousAggregateWatermark;
+
+/*
+ * Cache the watermark in the current transaction for better performance
+ * (by avoiding repeated max bucket calculations). The watermark will be
+ * reset at the end of the transaction, when the watermark function's input
+ * argument (materialized hypertable ID) changes, or when a new command is
+ * executed.
+ *
+ * One could potentially create a hashtable of watermarks keyed on materialized
+ * hypertable ID, but this is left as a future optimization since it doesn't
+ * seem to be common case that multiple continuous aggregates exist in the same
+ * query. Besides, the planner can constify calls to the watermark function
+ * during planning since the function is STABLE. Therefore, this is only a
+ * fallback if the planner needs to constify it many times (e.g., if used as
+ * an index condition on many chunks).
+ */
+static ContinuousAggregateWatermark *cagg_watermark_cache = NULL;
+
+/*
+ * Callback handler to reset the watermark after the transaction ends. This is
+ * triggered by the deletion of the associated memory context.
+ */
+static void
+cagg_watermark_reset(void *arg)
+{
+	cagg_watermark_cache = NULL;
+}
+
+/*
+ * ContinuousAggregateWatermark is valid for the duration of one command execution on the same
+ * materialized hypertable.
+ */
+static bool
+cagg_watermark_is_valid(const ContinuousAggregateWatermark *cagg_watermark, int32 mat_hypertable_id)
+{
+	return cagg_watermark != NULL && cagg_watermark->mat_hypertable_id == mat_hypertable_id &&
+		   cagg_watermark->cid == GetCurrentCommandId(false);
+}
+
+static void
+cagg_watermark_init_scan_by_mat_hypertable_id(ScanIterator *iterator, const int32 mat_hypertable_id)
+{
+	iterator->ctx.index = catalog_get_index(ts_catalog_get(),
+											CONTINUOUS_AGGS_WATERMARK,
+											CONTINUOUS_AGGS_WATERMARK_PKEY);
+
+	ts_scan_iterator_scan_key_init(iterator,
+								   Anum_continuous_aggs_watermark_pkey_mat_hypertable_id,
+								   BTEqualStrategyNumber,
+								   F_INT4EQ,
+								   Int32GetDatum(mat_hypertable_id));
+}
+
+static int64
+cagg_watermark_get(Hypertable *mat_ht)
+{
+	PG_USED_FOR_ASSERTS_ONLY short count = 0;
+	Datum watermark = (Datum) 0;
+	bool value_isnull = true;
+	ScanIterator iterator =
+		ts_scan_iterator_create(CONTINUOUS_AGGS_WATERMARK, AccessShareLock, CurrentMemoryContext);
+
+	cagg_watermark_init_scan_by_mat_hypertable_id(&iterator, mat_ht->fd.id);
+
+	ts_scanner_foreach(&iterator)
+	{
+		watermark = slot_getattr(ts_scan_iterator_slot(&iterator),
+								 Anum_continuous_aggs_watermark_watermark,
+								 &value_isnull);
+		count++;
+	}
+	Assert(count <= 1);
+	ts_scan_iterator_close(&iterator);
+
+	if (value_isnull)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("watermark not defined for continuous aggregate: %d", mat_ht->fd.id)));
+
+	return DatumGetInt64(watermark);
+}
+
+static ContinuousAggregateWatermark *
+cagg_watermark_create(const ContinuousAgg *cagg, MemoryContext top_mctx)
+{
+	Hypertable *ht;
+	ContinuousAggregateWatermark *watermark;
+	MemoryContext mctx = AllocSetContextCreate(top_mctx,
+											   "ContinuousAggregateWatermark function",
+											   ALLOCSET_DEFAULT_SIZES);
+
+	watermark = MemoryContextAllocZero(mctx, sizeof(ContinuousAggregateWatermark));
+	watermark->mctx = mctx;
+	watermark->mat_hypertable_id = cagg->data.mat_hypertable_id;
+	watermark->cid = GetCurrentCommandId(false);
+	watermark->cb.func = cagg_watermark_reset;
+	MemoryContextRegisterResetCallback(mctx, &watermark->cb);
+
+	/* Hypertable associated to the Continuous Aggregate */
+	ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
+
+	if (NULL == ht)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid materialization hypertable ID: %d",
+						cagg->data.mat_hypertable_id)));
+
+	/* Get the stored watermark */
+	watermark->value = cagg_watermark_get(ht);
+
+	return watermark;
+}
+
+TS_FUNCTION_INFO_V1(ts_continuous_agg_watermark);
+
+/*
+ * Get the watermark for a real-time aggregation query on a continuous
+ * aggregate.
+ *
+ * The watermark determines where the materialization ends for a continuous
+ * aggregate. It is used by real-time aggregation as the threshold between the
+ * materialized data and real-time data in the UNION query.
+ *
+ * The watermark is stored into `_timescaledb_catalog.continuous_aggs_watermark`
+ * catalog table by the `refresh_continuous_agregate` procedure. It is defined
+ * as the end of the last (highest) bucket in the materialized hypertable of a
+ * continuous aggregate.
+ *
+ * The materialized hypertable ID is given as input argument.
+ */
+Datum
+ts_continuous_agg_watermark(PG_FUNCTION_ARGS)
+{
+	const int32 mat_hypertable_id = PG_GETARG_INT32(0);
+	ContinuousAgg *cagg;
+	AclResult aclresult;
+
+	if (NULL != cagg_watermark_cache)
+	{
+		if (cagg_watermark_is_valid(cagg_watermark_cache, mat_hypertable_id))
+			PG_RETURN_INT64(cagg_watermark_cache->value);
+
+		MemoryContextDelete(cagg_watermark_cache->mctx);
+	}
+
+	cagg = ts_continuous_agg_find_by_mat_hypertable_id(mat_hypertable_id);
+
+	if (NULL == cagg)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid materialized hypertable ID: %d", mat_hypertable_id)));
+
+	/*
+	 * Preemptive permission check to ensure the function complains about lack
+	 * of permissions on the cagg rather than the materialized hypertable
+	 */
+	aclresult = pg_class_aclcheck(cagg->relid, GetUserId(), ACL_SELECT);
+	aclcheck_error(aclresult, OBJECT_MATVIEW, get_rel_name(cagg->relid));
+	cagg_watermark_cache = cagg_watermark_create(cagg, TopTransactionContext);
+
+	PG_RETURN_INT64(cagg_watermark_cache->value);
+}
+
+static int64
+cagg_compute_watermark(ContinuousAgg *cagg, int64 watermark, bool isnull)
+{
+	if (isnull)
+	{
+		watermark = ts_time_get_min(cagg->partition_type);
+	}
+	else
+	{
+		/*
+		 * The materialized hypertable is already bucketed, which means the
+		 * max is the start of the last bucket. Add one bucket to move to the
+		 * point where the materialized data ends.
+		 */
+		if (ts_continuous_agg_bucket_width_variable(cagg))
+		{
+			/*
+			 * Since `value` is already bucketed, `bucketed = true` flag can
+			 * be added to ts_compute_beginning_of_the_next_bucket_variable() as
+			 * an optimization, if necessary.
+			 */
+			watermark =
+				ts_compute_beginning_of_the_next_bucket_variable(watermark, cagg->bucket_function);
+		}
+		else
+		{
+			watermark = ts_time_saturating_add(watermark,
+											   ts_continuous_agg_bucket_width(cagg),
+											   cagg->partition_type);
+		}
+	}
+
+	return watermark;
+}
+
+TS_FUNCTION_INFO_V1(ts_continuous_agg_watermark_materialized);
+
+/*
+ * Get the materialized watermark for a real-time aggregation query on a
+ * continuous aggregate.
+ *
+ * The difference between this function and `ts_continuous_agg_watermark` is
+ * that this one get the max open dimension of the materialization hypertable
+ * insted of get the stored value in the catalog table.
+ */
+Datum
+ts_continuous_agg_watermark_materialized(PG_FUNCTION_ARGS)
+{
+	const int32 mat_hypertable_id = PG_GETARG_INT32(0);
+	ContinuousAgg *cagg;
+	AclResult aclresult;
+	bool isnull;
+	Hypertable *ht;
+	int64 watermark;
+
+	cagg = ts_continuous_agg_find_by_mat_hypertable_id(mat_hypertable_id);
+
+	if (NULL == cagg)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid materialized hypertable ID: %d", mat_hypertable_id)));
+
+	/*
+	 * Preemptive permission check to ensure the function complains about lack
+	 * of permissions on the cagg rather than the materialized hypertable
+	 */
+	aclresult = pg_class_aclcheck(cagg->relid, GetUserId(), ACL_SELECT);
+	aclcheck_error(aclresult, OBJECT_MATVIEW, get_rel_name(cagg->relid));
+
+	ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
+	watermark = ts_hypertable_get_open_dim_max_value(ht, 0, &isnull);
+
+	watermark = cagg_compute_watermark(cagg, watermark, isnull);
+
+	PG_RETURN_INT64(watermark);
+}
+
+TSDLLEXPORT void
+ts_cagg_watermark_insert(Hypertable *mat_ht, int64 watermark, bool watermark_isnull)
+{
+	Catalog *catalog = ts_catalog_get();
+	Relation rel =
+		table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_WATERMARK), RowExclusiveLock);
+	TupleDesc desc = RelationGetDescr(rel);
+	Datum values[Natts_continuous_aggs_watermark];
+	bool nulls[Natts_continuous_aggs_watermark] = { false };
+	CatalogSecurityContext sec_ctx;
+
+	/* if trying to insert a NULL watermark then get the MIN value for the time dimension */
+	if (watermark_isnull)
+	{
+		const Dimension *dim = hyperspace_get_open_dimension(mat_ht->space, 0);
+
+		if (NULL == dim)
+			elog(ERROR, "invalid open dimension index %d", 0);
+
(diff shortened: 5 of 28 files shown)


fabriziomello:
Some preliminary results:

## **1. Last TimescaleDB release: 2.10.0**
```bash
tsbench on  main [?] via tsbench took 12s 
➜ python -m src.tsbench --with-connection pgsql://fabrizio@localhost:5432/2.10.0 --benchmarks cagg_watermark              
*** Processing benchmarks on connection: pgsql://fabrizio@localhost:5432/2.10.0
*** Executing benchmark 'cagg_watermark'
*** All benchmarks are executed - done
============================================
Report for benchmark suite 'cagg_watermark'
+--------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                | 8b549b08e28d121946eccbc7452bc476c7754dc8 |
+--------------------------------------------------------------------------------------+------------------------------------------+
| SELECT bucket, a, value FROM agg_1m  WHERE a = 1 AND bucket > '2023-01-01 00:00:00'; |                                     2.03 |
| SELECT bucket, a, value FROM agg_5m  WHERE a = 1 AND bucket > '2023-01-01 00:00:00'; |                                     8.89 |
| SELECT bucket, a, value FROM agg_15m WHERE a = 1 AND bucket > '2023-01-01 00:00:00'; |                                   158.18 |
+--------------------------------------------------------------------------------------+------------------------------------------+
```
**Flamegraph:** 
result_file_name3


## **2. This PR**
```bash
tsbench on  main [?] via tsbench took 11s 
➜ python -m src.tsbench --with-connection pgsql://fabrizio@localhost:5432/fabrizio --benchmarks cagg_watermark --no-teardown
*** Processing benchmarks on connection: pgsql://fabrizio@localhost:5432/fabrizio
*** Executing benchmark 'cagg_watermark'
*** All benchmarks are executed - done
============================================
Report for benchmark suite 'cagg_watermark'
+--------------------------------------------------------------------------------------+------------------------------------------+
| Query                                                                                | b7ad720d8243fe17a14a257434beff7430bc97c7 |
+--------------------------------------------------------------------------------------+------------------------------------------+
| SELECT bucket, a, value FROM agg_1m  WHERE a = 1 AND bucket > '2023-01-01 00:00:00'; |                                     1.76 |
| SELECT bucket, a, value FROM agg_5m  WHERE a = 1 AND bucket > '2023-01-01 00:00:00'; |                                     3.17 |
| SELECT bucket, a, value FROM agg_15m WHERE a = 1 AND bucket > '2023-01-01 00:00:00'; |                                    28.63 |
+--------------------------------------------------------------------------------------+------------------------------------------+
```
**Flamegraph:** 
result_file_name4

mkindahl (commented):
To link both issues to this pull request, [you need an additional "Closes" in the list](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,319 @@
...
+											cagg_watermark_scan_update /*=tuple_found*/,
+											RowExclusiveLock /*=lockmode*/,
+											CONTINUOUS_AGGS_WATERMARK_TABLE_NAME /*=table_name*/,
+											&new_watermark /*=data*/);
+
+	if (!watermark_updated)
+	{
+		Catalog *catalog = ts_catalog_get();
+		Relation rel =
+			table_open(catalog_get_table_id(catalog, CONTINUOUS_AGGS_WATERMARK), RowExclusiveLock);
```
mkindahl, line 238: Since you release the lock on the table inside `ts_catalog_scan_one` it might well be the case that a watermark is added by another process between this part of the code and the call of `ts_catalog_scan_one` This would result in a duplicate entry.

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,319 @@
...
+		values[AttrNumberGetAttrOffset(Anum_continuous_aggs_watermark_mat_hypertable_id)] =
+			Int32GetDatum(mat_hypertable_id);
+		values[AttrNumberGetAttrOffset(Anum_continuous_aggs_watermark_watermark)] =
+			Int64GetDatum(new_watermark);
+
+		ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
+		ts_catalog_insert_values(rel, desc, values, nulls);
+		ts_catalog_restore_user(&sec_ctx);
+
+		table_close(rel, RowExclusiveLock);
```
mkindahl, line 253: Since you're releasing the `RowExclusiveLock` here another transaction that started earlier might come in and change the value to a watermark that is not the correct one. I think you need to keep the lock until the end of the transaction.

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,319 @@
...
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid materialized hypertable ID: %d", mat_ht->fd.id)));
+
+	if (watermark_isnull)
+	{
+		watermark = ts_time_get_min(ts_dimension_get_partition_type(dim));
+	}
+	else
+	{
+		Oid timetype = ts_dimension_get_partition_type(dim);
```
mkindahl, line 274: Nit: This is only used in one of the branches below. Since unnecessary to compute it for both cases.

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,319 @@
...
+	const Dimension *dim = hyperspace_get_open_dimension(mat_ht->space, 0);
+
+	if (NULL == cagg)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid materialized hypertable ID: %d", mat_ht->fd.id)));
+
+	if (watermark_isnull)
+	{
+		watermark = ts_time_get_min(ts_dimension_get_partition_type(dim));
```
mkindahl, line 270: Nit: The inner function call is computed in both branches. Seems unnecessary to duplicate it.

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,319 @@
...
+
+	cagg_watermark_init_scan_by_mat_hypertable_id(&iterator, mat_hypertable_id);
+
+	ts_scanner_foreach(&iterator)
+	{
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		ts_catalog_delete_tid(ti->scanrel, ts_scanner_get_tuple_tid(ti));
+		count++;
+	}
+	return count > 0;
```
mkindahl, line 318: Nit: This return value is never checked. Seems redundant.

tsl/src/continuous_aggs/create.c
```diff
@@ -2863,6 +2865,16 @@ tsl_process_continuous_agg_viewstmt(Node *node, const char *query_string, void *
 
 		continuous_agg_refresh_internal(cagg, &refresh_window, CAGG_REFRESH_CREATION, true, true);
 	}
+	else
+	{
+		/* Set NULL watermark */
+		Hypertable *ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
+		Assert(NULL != ht);
```
mkindahl, line 2872: Since this is reading from the catalog, you might want to use `Ensure` here to not crash in production builds if the catalog data is corrupt.

mkindahl (commented):
A few comments below.

src/chunk.c
```diff
@@ -3911,14 +3912,22 @@ ts_chunk_do_drop_chunks(Hypertable *ht, int64 older_than, int64 newer_than, int3
...
 			has_continuous_aggs = false;
+			is_materialization_hypertable = true;
 			break;
+		}
 		case HypertableIsMaterializationAndRaw:
+		{
 			has_continuous_aggs = true;
+			is_materialization_hypertable = true;
 			break;
+		}
```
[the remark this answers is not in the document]
fabriziomello: Fixed!

src/hypertable.c
```diff
@@ -2930,10 +2932,10 @@ ts_hypertable_get_open_dim_max_value(const Hypertable *ht, int dimension_index,
 					 quote_identifier(NameStr(ht->fd.schema_name)),
 					 quote_identifier(NameStr(ht->fd.table_name)));
 
-	if (SPI_connect() != SPI_OK_CONNECT)
+	if (connect_to_spi && SPI_connect() != SPI_OK_CONNECT)
 		elog(ERROR, "could not connect to SPI");
 
-	res = SPI_execute(command->data, true /* read_only */, 0 /*count*/);
+	res = SPI_execute(command->data, connect_to_spi /* read_only */, 0 /*count*/);
```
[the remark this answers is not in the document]
fabriziomello: Actually I'm skipping connect to SPI when the caller already do it... and this is when we execute this function inside the materialization. I'm reusing the `connect_to_spi` flag to teach SPI that we're not executing read-only queries that is the case when we need to recovery then data updated in the current transaction.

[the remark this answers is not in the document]
mkindahl: Ok.

Could you please add a comment before the call so that it is clear that the caller is expected to do `SPI_connect` and `SPI_finish`?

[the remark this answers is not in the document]
fabriziomello: After some refactoring removed this change completely.

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,319 @@
...
+ * hypertable ID) changes, or when a new command is executed. One could
+ * potentially create a hashtable of watermarks keyed on materialized
+ * hypertable ID, but this is left as a future optimization since it doesn't
+ * seem to be common case that multiple continuous aggregates exist in the
+ * same query. Besides, the planner can constify calls to the watermark
+ * function during planning since the function is STABLE. Therefore, this is
+ * only a fallback if the planner needs to constify it many times (e.g., if
+ * used as an index condition on many chunks).
+ */
+static ContinuousAggregateWatermark *watermark_cache = NULL;
```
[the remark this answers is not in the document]
fabriziomello: You're correct, it is a cache backend... I didn't wrote this comment just copy/paste from the original implementation. Rephrased it a bit to be more clear.

src/ts_catalog/continuous_aggs_watermark.c
fabriziomello: Fixed

src/ts_catalog/continuous_aggs_watermark.c
fabriziomello: Fixed

src/ts_catalog/continuous_aggs_watermark.c
fabriziomello: Removed.

tsl/src/continuous_aggs/create.c
fabriziomello: Fixed!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ accidentally triggering the load of a previous DB version.**
 
 **Features**
 * #5212 Allow pushdown of reference table joins
+* #5221 Improve Realtime Continuous Aggregate performance
 * #5312 Add timeout support to the ping_data_node()
 * #5361 Add parallel support for partialize_agg()
```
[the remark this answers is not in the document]
fabriziomello: Yes it is a bordeline... added a Thanks entry

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,396 @@
...
+	PG_RETURN_INT64(cagg_watermark_cache->value);
+}
+
+static int64
+cagg_compute_watermark(ContinuousAgg *cagg, int64 watermark, bool isnull)
+{
+	if (isnull)
+	{
+		watermark = ts_time_get_min(cagg->partition_type);
+	}
```
mkindahl, line 193: Nit: you can just return the value in each branch. Feel free to ignore.

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,396 @@
...
+
+TSDLLEXPORT void
+ts_cagg_watermark_insert(Hypertable *mat_ht, int64 watermark, bool watermark_isnull)
+{
+	ContinuousAgg *cagg = ts_continuous_agg_find_by_mat_hypertable_id(mat_ht->fd.id);
+
+	if (NULL == cagg)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("invalid materialized hypertable ID: %d", mat_ht->fd.id)));
```
mkindahl, line 272: Since you have already fetched the materialized hypertable, this error would mean that something changed between fetching it and using it. The check here will prevent a crash in that case, but if the above scenario can happen, isn't this an indication of something wrong how the locks are handled?

tsl/src/continuous_aggs/refresh.c
```diff
@@ -18,6 +18,7 @@
 
 #include "ts_catalog/catalog.h"
 #include "ts_catalog/continuous_agg.h"
+// #include "ts_catalog/continuous_aggs_watermark.h"
 #include <dimension.h>
```
mkindahl, line 22: Remove it if you don't need it.
```suggestion
#include "ts_catalog/continuous_agg.h"
#include <dimension.h>
```

mkindahl (commented):
In general it looks good. A few minor things that I think need to be fixed. There is a bigger question on why some of the error would be raised since these would rather be indications of us doing something wrong with the locking. Raising errors here are safe, since they would normally not fire, but not sure if we want to have some additional tests for these cases.

sql/pre_install/tables.sql
```diff
@@ -404,6 +404,18 @@ CREATE TABLE _timescaledb_catalog.continuous_aggs_invalidation_threshold (
 
 SELECT pg_catalog.pg_extension_config_dump('_timescaledb_catalog.continuous_aggs_invalidation_threshold', '');
 
+CREATE TABLE _timescaledb_catalog.continuous_aggs_watermark (
```
erimatnor, line 407: Idea (feel free to ignore): Can't we simply create a bigserial for a cagg to track its watermark? We would have one bigserial for each cagg in the internal schema (next to the materialized hypertable itself).

That would avoid all the extra plumbing with catalog tables, indexes, etc., and we would benefit from internal (rel)caches. Plus we could just use `setval()` and `pg_sequence_last_value()` to update and read the watermark.

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,396 @@
...
+}
+
+static int64
+cagg_watermark_get(Hypertable *mat_ht)
+{
+	PG_USED_FOR_ASSERTS_ONLY short count = 0;
+	Datum watermark = (Datum) 0;
+	bool value_isnull = true;
+	ScanIterator iterator =
+		ts_scan_iterator_create(CONTINUOUS_AGGS_WATERMARK, RowExclusiveLock, CurrentMemoryContext);
```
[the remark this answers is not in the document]
erimatnor: Why RowExclusiveLock if we are just reading the watermark. Shouldn't it be AccessShareLock?

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,396 @@
...
+								 &value_isnull);
+		count++;
+	}
+	Assert(count <= 1);
+	ts_scan_iterator_close(&iterator);
+
+	if (value_isnull)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("Watermark not defined for continuous aggregate: %d", mat_ht->fd.id)));
```
erimatnor, line 106:
```suggestion
				 errmsg("watermark not defined for continuous aggregate: %d", mat_ht->fd.id)));
```

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,396 @@
...
+	{
+		watermark = slot_getattr(ts_scan_iterator_slot(&iterator),
+								 Anum_continuous_aggs_watermark_watermark,
+								 &value_isnull);
+		count++;
+	}
+	Assert(count <= 1);
+	ts_scan_iterator_close(&iterator);
+
+	if (value_isnull)
```
erimatnor, line 103: If there is no watermark, maybe we should return NULL or -INF instead of throwing an error?

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,396 @@
...
+											cagg_watermark_update_scan_internal /*=tuple_found*/,
+											RowExclusiveLock /*=lockmode*/,
+											CONTINUOUS_AGGS_WATERMARK_TABLE_NAME /*=table_name*/,
+											&data /*=data*/);
+
+	if (!watermark_updated)
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("Watermark not defined for continuous aggregate: %d", mat_hypertable_id)));
```
erimatnor, line 362:
```suggestion
				 errmsg("watermark not defined for continuous aggregate: %d", mat_hypertable_id)));
```

src/chunk.c
```diff
@@ -3893,7 +3894,7 @@ ts_chunk_do_drop_chunks(Hypertable *ht, int64 older_than, int64 newer_than, int3
 	Chunk *chunks;
 	const char *schema_name, *table_name;
 	const int32 hypertable_id = ht->fd.id;
-	bool has_continuous_aggs;
+	bool has_continuous_aggs, is_materialization_hypertable;
```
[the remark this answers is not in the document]
fabriziomello: Fixed

src/ts_catalog/continuous_aggs_watermark.c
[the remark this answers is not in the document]
fabriziomello: > Hmm... what context is being used here? You might want to check that you are not using a more long-lived context. It should probably be the tuple context since this is only relevant for the duration of the expression.

ExprContext is used

src/ts_catalog/continuous_aggs_watermark.c
fabriziomello: This will happen just if user shoot their own foot and remove the watermark tuple. I also added some failing tests for this.

src/ts_catalog/continuous_aggs_watermark.c
fabriziomello: Actually I don't need to populate the ContinuousAgg data structure here, I just need to get the time datatype for the open dimension of the Hypertable passed by parameter.

Fixed!

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,396 @@
...
+	watermark->cid = GetCurrentCommandId(false);
+	watermark->cb.func = cagg_watermark_reset;
+	MemoryContextRegisterResetCallback(mctx, &watermark->cb);
+
+	/* Hypertable associated to the Continuous Aggregate */
+	ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
+	Assert(NULL != ht);
+
+	/* Get the stored watermark */
+	watermark->value = cagg_watermark_get(ht);
```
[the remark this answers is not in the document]
fabriziomello: Fixed.

tsl/src/continuous_aggs/create.c
```diff
@@ -2816,33 +2824,48 @@ tsl_process_continuous_agg_viewstmt(Node *node, const char *query_string, void *
...
+		   stmt->into->rel->relname);
+
+	cagg = ts_continuous_agg_find_by_relid(relid);
+	Ensure(NULL != cagg,
+		   "continuous aggregate \"%s\".\"%s\" not found",
+		   schema_name,
+		   stmt->into->rel->relname);
+
+	mat_ht = ts_hypertable_get_by_id(cagg->data.mat_hypertable_id);
+	Ensure(NULL != cagg, "materialization hypertable %d not found", cagg->data.mat_hypertable_id);
```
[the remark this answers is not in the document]
fabriziomello: Fixed.

tsl/src/continuous_aggs/refresh.c
fabriziomello: Fixed!

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,319 @@
...
+	MemoryContext mctx = AllocSetContextCreate(top_mctx,
+											   "ContinuousAggregateWatermark function",
+											   ALLOCSET_DEFAULT_SIZES);
+
+	watermark = MemoryContextAllocZero(mctx, sizeof(ContinuousAggregateWatermark));
+	watermark->mctx = mctx;
+	watermark->mat_hypertable_id = cagg->data.mat_hypertable_id;
+	watermark->cid = GetCurrentCommandId(false);
+	watermark->cb.func = cagg_watermark_reset;
+	MemoryContextRegisterResetCallback(mctx, &watermark->cb);
```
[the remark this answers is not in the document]
fabriziomello: I prefer to don't mess with this in this PR. I know there are potential improvements here but this code is exact the same that is current working for long time now. We're already doing a deep refactoring here and I plan to do some follow-up PRs with other ideas to completely cache all the caggs watermark in a hashtable and also add a constify in the planner for even more query improvements, but first need to figure out a way to don't move back the watermark to the past (when dropping chunks).

sql/pre_install/tables.sql
fabriziomello: The problem of sequences is they're independent of transaction. Imagine that for some reason the materialization transaction of refresh cagg fail, in this case the watermark will move forward but the data will not be materialized... not sure if a sequence is a good idea here!

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,401 @@
...
+	ScanIterator iterator =
+		ts_scan_iterator_create(CONTINUOUS_AGGS_WATERMARK, RowExclusiveLock, CurrentMemoryContext);
+
+	cagg_watermark_init_scan_by_mat_hypertable_id(&iterator, mat_hypertable_id);
+
+	ts_scanner_foreach(&iterator)
+	{
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		ts_catalog_delete_tid(ti->scanrel, ts_scanner_get_tuple_tid(ti));
+	}
```
mkindahl, line 400: Nit. The iterator will automatically be closed if there are no more tuples, so this is just for good measure.

```suggestion
	}
    	ts_scan_iterator_close(&iterator);
```

src/ts_catalog/continuous_aggs_watermark.c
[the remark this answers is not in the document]
mkindahl: It's a nit, so it is fine to ignore.

mkindahl (commented):
We've had problems with `VACUUM` previously, so I think we should be careful about releasing locks too early, but otherwise it looks good.

erimatnor (approved):

src/ts_catalog/continuous_aggs_watermark.c
```diff
@@ -0,0 +1,401 @@
...
+	values[AttrNumberGetAttrOffset(Anum_continuous_aggs_watermark_mat_hypertable_id)] =
+		Int32GetDatum(mat_ht->fd.id);
+	values[AttrNumberGetAttrOffset(Anum_continuous_aggs_watermark_watermark)] =
+		Int64GetDatum(watermark);
+
+	ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
+	ts_catalog_insert_values(rel, desc, values, nulls);
+	ts_catalog_restore_user(&sec_ctx);
+
+	table_close(rel, RowExclusiveLock);
```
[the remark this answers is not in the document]
fabriziomello: Make sense. Fixed!

src/ts_catalog/continuous_aggs_watermark.c
fabriziomello: Fixed!

mkindahl (approved):

Issue #5161, opened 2023-01-09
[Enhancement]: Extend enabling compression on a continuous aggregrate with 'segmentby' and 'orderby' parameters


emizero:
### What type of enhancement is this?

API improvement, Performance

### What subsystems and features will be improved?

Compression, Continuous aggregate

### What does the enhancement do?

Extend "ALTER MATERIALIZED VIEW" with the 'segmentby' and 'orderby' parameters when enabling compression.
segmentby parameter is crucial for performance when caggs are especially large.  At the moment trying the query results in the error below.

```
ALTER MATERIALIZED VIEW test_table_cagg SET (
  timescaledb.compress = true,
  timescaledb.compress_segmentby = 'device_id'
);
ERROR:  unrecognized parameter "timescaledb.compress_segmentby"
```

Currently, there is a workaround that allows us to enable compression directly on the internal hypertable for the continuous aggregate.
```
ALTER TABLE _timescaledb_internal._materialized_hypertable_71 SET (                              
  timescaledb.compress = true,
  timescaledb.compress_segmentby = 'device_id'
);
ALTER TABLE
```

### Implementation challenges

Unknown at the moment

kgyrtkirk:
@fabriziomello: it would probably make sense to cover all options in the compression with clause [here](https://github.com/timescale/timescaledb/blob/5d12a3883d80f026252eaac8416143de6cd5653d/src/compression_with_clause.c#L24-L42); however that raises the question - does all of them make sense for an MV as well?

If they do make sense - then I would probably need to add some end-markers to [CompressHypertableOption](https://github.com/timescale/timescaledb/blob/5d12a3883d80f026252eaac8416143de6cd5653d/src/compression_with_clause.h#L16) and [ContinuousAggViewOption](https://github.com/timescale/timescaledb/blob/5d12a3883d80f026252eaac8416143de6cd5653d/src/ts_catalog/continuous_agg.h#L44) enums.

PR #5262, opened 2023-01-31
#5161 segmentby param

kgyrtkirk:


kgyrtkirk:
relates to: #5161

tsl/src/continuous_aggs/options.c
```diff
@@ -218,28 +218,37 @@ cagg_get_compression_params(ContinuousAgg *agg, Hypertable *mat_ht)
...
+		// merge defaults if there's any
+		for (int i = 0; i < _CompressOptionMax; i++)
+		{
+			if (with_clause_options[i].is_default && !default_with_clause_options[i].is_default)
+			{
+				with_clause_options[i] = default_with_clause_options[i];
+				// ereport(NOTICE,
+				// 		errmsg("defaulting %s to %s",
+				// 			   ts_compress_with_clause_def_get_name(i),
+				// 			   DatumGetCString(default_with_clause_options[i].parsed)));
```
kgyrtkirk, line 243: these `WithClauseDefinition` structs are are not really giving much advantage...
instead of enum keyed arrays - which could possibly wreck havoc if someone uses a different enum to index them - I would either add the key/type to the result as well; or try some other approach - like creating a full standard struct with the names of the fileds documenting what those informations are

without the name of the field its even hard to give proper messages to the user...

kgyrtkirk:
@mkindahl could you please take a look?

mkindahl (commented):
You need to update the commit message to match the guidelines in https://github.com/timescale/timescaledb/blob/main/CONTRIBUTING.md

tsl/src/continuous_aggs/options.c
kgyrtkirk: I've decided to add a pointer for the definition to make unparsing/name access simple and straightforward

CHANGELOG.md
```diff
@@ -10,12 +10,14 @@ accidentally triggering the load of a previous DB version.**
...
 * #5245 Mange life-cycle of connections via memory contexts
 * #5246 Make connection establishment interruptible
+* #5262 Extend enabling compression on a continuous aggregrate with 'segmentby' and 'orderby' parameters
 * #5253 Make data node command execution interruptible
 
 **Bugfixes**
 * #4926 Fix corruption when inserting into compressed chunks
 * #5218 Add role-level security to job error log
 * #5214 Fix use of prepared statement in async module
+* #5290 Compression can't be enabled on MVs when segmentby/orderby columns need quotation
```
akuzm, line 20:
```suggestion
```

The next line is the same.

CHANGELOG.md
```diff
@@ -10,12 +10,14 @@ accidentally triggering the load of a previous DB version.**
 * #5241 Allow RETURNING clause when inserting into compressed chunks
 * #5245 Mange life-cycle of connections via memory contexts
 * #5246 Make connection establishment interruptible
+* #5262 Extend enabling compression on a continuous aggregrate with 'segmentby' and 'orderby' parameters
```
akuzm, line 13: Maybe something like "allow to specify `compress_segmentby` and `compress_orderby` parameters when enabling a compression on a continuous aggregate table"?

src/compression_with_clause.h
```diff
@@ -37,5 +38,4 @@ extern TSDLLEXPORT List *ts_compress_hypertable_parse_order_by(WithClauseResult
 extern TSDLLEXPORT Interval *
 ts_compress_hypertable_parse_chunk_time_interval(WithClauseResult *parsed_options,
 												 Hypertable *hypertable);
-
```
akuzm, line 40: Accidental change?

src/with_clause_parser.c
```diff
@@ -113,6 +114,24 @@ ts_with_clauses_parse(const List *def_elems, const WithClauseDefinition *args, S
 	return results;
 }
 
+extern TSDLLEXPORT char *
+ts_with_clause_result_unparse_value(const WithClauseResult *result)
```
akuzm, line 118: Our code mostly uses "deparse", let's use it here as well to avoid confusion.

src/ts_catalog/continuous_agg.c
```diff
@@ -77,6 +90,52 @@ ts_continuous_agg_with_clause_parse(const List *defelems)
...
+				option_index = ContinuousViewOptionCompressSegmentBy;
+				break;
+			case CompressOrderBy:
+				option_index = ContinuousViewOptionCompressOrderBy;
+				break;
+			case CompressChunkTimeInterval:
+				option_index = ContinuousViewOptionCompressChunkTimeInterval;
+				break;
+			default:
+				ereport(ERROR,
```
akuzm, line 117: For internal program errors, there's a shorter form `elog(ERROR, ...` which also sets the right error code for you. Or `Ensure` if you want it to be an assertion in debug.

tsl/src/continuous_aggs/options.c
```diff
@@ -194,28 +194,37 @@ cagg_get_compression_params(ContinuousAgg *agg, Hypertable *mat_ht)
...
-															(Node *) makeString("false"),
-										  DEFELEM_UNSPEC,
-										  -1);
-	defelems = lappend(defelems, enable);
+	if (with_clause_options[CompressEnabled].parsed)
+	{
+		List *default_compress_defelems = cagg_get_compression_params(agg, mat_ht);
+		WithClauseResult *default_with_clause_options =
+			ts_compress_hypertable_set_clause_parse(default_compress_defelems);
+		// merge defaults if there's any
```
akuzm, line 210:
```suggestion
		/* Merge defaults if there's any. */
```

Let's follow the Postgres comment style.

akuzm (approved):

src/ts_catalog/continuous_agg.c
```diff
@@ -77,6 +90,51 @@ ts_continuous_agg_with_clause_parse(const List *defelems)
 								 continuous_aggregate_with_clause_def,
 								 TS_ARRAY_LEN(continuous_aggregate_with_clause_def));
 }
+
+List *
+ts_continuous_agg_unparse_compression_defelems(const WithClauseResult *with_clauses)
```
mkindahl, line 95: We call it "deparse", but I am not sure if this is really deparsing. It looks more like extracting elements from a structure. More like "get names...."
```suggestion
ts_continuous_agg_deparse_compression_defelems(const WithClauseResult *with_clauses)
```

src/ts_catalog/continuous_agg.h
```diff
@@ -48,6 +48,10 @@ typedef enum ContinuousAggViewOption
 	ContinuousViewOptionMaterializedOnly,
 	ContinuousViewOptionCompress,
 	ContinuousViewOptionFinalized,
+	ContinuousViewOptionCompressSegmentBy,
+	ContinuousViewOptionCompressOrderBy,
+	ContinuousViewOptionCompressChunkTimeInterval,
+	_ContinuousViewOptionMax
```
mkindahl, line 54:
```suggestion
	ContinuousViewOptionMax
```

src/with_clause_parser.c
mkindahl: Agree on that.
```suggestion
ts_with_clause_result_deparse_value(const WithClauseResult *result)
```

src/with_clause_parser.c
```diff
@@ -113,6 +114,24 @@ ts_with_clauses_parse(const List *def_elems, const WithClauseDefinition *args, S
...
+{
+	Oid oid = result->definition->type_id;
+	if (!OidIsValid(oid))
+		elog(ERROR, "Argument \"%d\" has invalid Oid!", oid);
```
mkindahl, line 122: [Error Message Style Guide](https://www.postgresql.org/docs/current/error-style-guide.html)

```suggestion
		elog(ERROR, "argument \"%d\" has invalid OID", oid);
```

tsl/src/continuous_aggs/options.c
```diff
@@ -194,28 +194,37 @@ cagg_get_compression_params(ContinuousAgg *agg, Hypertable *mat_ht)
...
+		for (int i = 0; i < _CompressOptionMax; i++)
+		{
+			if (with_clause_options[i].is_default && !default_with_clause_options[i].is_default)
+			{
+				with_clause_options[i] = default_with_clause_options[i];
+				ereport(NOTICE,
+						errmsg("defaulting %s to %s",
+							   with_clause_options[i].definition->arg_name,
+							   ts_with_clause_result_unparse_value(&with_clause_options[i])));
```
mkindahl, line 219: Using `elog` works fine here, unless you want to provide hints and/or details.

mkindahl (commented):
Missing a commit message. Reviewing a PR involves reviewing the commit and PR message as well.

src/compression_with_clause.h
```diff
@@ -19,6 +19,7 @@ typedef enum CompressHypertableOption
 	CompressSegmentBy,
 	CompressOrderBy,
 	CompressChunkTimeInterval,
+	_CompressOptionMax
```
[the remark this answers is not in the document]
kgyrtkirk: sure; I can remove the `_` ; but in this repository there are a lot of enums which has an underscore prefix in their Max; I just followed the existing convention...see [here](https://github.com/timescale/timescaledb/blob/348796f9d9dcfc01716e78886ac1cb3de8623f8f/src/ts_catalog/catalog.h#L61) for example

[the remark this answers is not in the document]
mkindahl: For nits like this, it is your decision if you want to do it or not.

src/ts_catalog/continuous_agg.c
kgyrtkirk: thank you; I was using ereport because it seemed `elog` was deprecated...but its more convinient for sure! :)

src/compression_with_clause.h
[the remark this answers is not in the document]
kgyrtkirk: I've renamed it...but I find it a bit confusing - without any distinctive things in the name its not clear at first glance that this enum entry supposed to mark the max value and its not there because there is a 'max' in the options....but since its the last and I guess everyone working on this will kinda know what could that be - I don't think thats a big hazard..

I looked at all the existing enum keys which have 'max' in their name; and all of them have a `_`  prefix:
```
git grep  -EC1  '^\s*[A-Za-z_]+M[Aa][xX]'  $(find . -name *.h)
```

however these 2 enums doesn't follow convention by using snake-case in their names anyway; so it's mixed up anyway..maybe all the enum keys should be renamed to follow the existing conventions?

CHANGELOG.md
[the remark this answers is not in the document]
kgyrtkirk: the other PR was merged into thispoint at some point; now this change is not there

src/ts_catalog/continuous_agg.c
kgyrtkirk: I've renamed the method to `ts_continuous_agg_get_compression_defelems`

kgyrtkirk:
@mkindahl I've squashed the commits - and improved the commit message; let me know what you think!

mkindahl (commented):
Looks good, but do you need to add a test case for the error that you added? I can't see one.

src/with_clause_parser.c
```diff
@@ -113,6 +114,24 @@ ts_with_clauses_parse(const List *def_elems, const WithClauseDefinition *args, S
 	return results;
 }
 
+extern TSDLLEXPORT char *
+ts_with_clause_result_deparse_value(const WithClauseResult *result)
+{
+	Oid oid = result->definition->type_id;
+	if (!OidIsValid(oid))
+		elog(ERROR, "argument \"%d\" has invalid OID", oid);
```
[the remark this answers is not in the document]
akuzm: `elog(ERROR ...)` is for internal program errors, like an assertion failure, so we probably don't want to be able to trigger it. If it's a user error, it should use `ereport` instead.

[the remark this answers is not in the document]
kgyrtkirk: no; there is no way for the user to affect this - I don't expect it to go off - its here for just in case something really odd is happening....

I could probably:
* change it to an assertion instead
* force-it to fail from an api test
* or remove it

I think an assertion would be the most appropriate in this case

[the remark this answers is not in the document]
mkindahl: > no; there is no way for the user to affect this - I don't expect it to go off - its here for just in case something really odd is happening....

Ok. Makes sense to not have a test for this then.

You might want to consider using `Ensure` here, which is intended for this usage. It will abort execution in debug build if the condition does not hold, which makes it easier to debug since you will have a core dump. In release builds, it will just be an error message, just like here.

> or remove it

I do not think that we should remove such checks. It will just lead to crashes in production code, which causes a lot of other problems since it brings the entire server down. Better to print an error and abort if the impossible, or less plausible, happens.

src/with_clause_parser.c
```diff
@@ -113,6 +114,22 @@ ts_with_clauses_parse(const List *def_elems, const WithClauseDefinition *args, S
...
+ts_with_clause_result_deparse_value(const WithClauseResult *result)
+{
+	Oid oid = result->definition->type_id;
+	Assert(OidIsValid(oid));
+
+	Oid in_fn;
+	bool typIsVarlena;
+
+	getTypeOutputInfo(oid, &in_fn, &typIsVarlena);
+	Assert(OidIsValid(in_fn));
```
mkindahl, line 127: This reads from PostgreSQL catalog tables, so a corrupt catalog table would then potentially cause a crash, which does not seem like a good idea. You would want to error out if the catalog tables are corrupt, however unlikely this is.
```suggestion
	getTypeOutputInfo(oid, &in_fn, &typIsVarlena);
	Ensure(OidIsValid(in_fn), "no output function for type with OID %d", oid);
```

src/with_clause_parser.c
```diff
@@ -113,6 +114,22 @@ ts_with_clauses_parse(const List *def_elems, const WithClauseDefinition *args, S
...
 }
 
+extern TSDLLEXPORT char *
+ts_with_clause_result_deparse_value(const WithClauseResult *result)
+{
+	Oid oid = result->definition->type_id;
+	Assert(OidIsValid(oid));
+
+	Oid in_fn;
+	bool typIsVarlena;
```
mkindahl, line 124: This generate errors with linters if the variable is not used.
```suggestion
	bool typIsVarlena pg_attribute_unused();
```

src/with_clause_parser.c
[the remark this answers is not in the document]
kgyrtkirk: replaced it with `Ensure` - that's even better than the original check. I usually try to retain these kind of checks as much as possible.

src/with_clause_parser.c
kgyrtkirk: replaced it with `Assert`

I was looking around a bit to see what's the usual practice regarding this; and I've found some which doesn't really take care of this:
* [ts_internal_to_time_string](https://github.com/timescale/timescaledb/blob/f13214891cf47fed0db6183ce6b56d9273865de3/src/utils.c#L390-L392)
* [cagg_validate_query](https://github.com/timescale/timescaledb/blob/main/tsl/src/continuous_aggs/create.c#L1559-L1565)

and possibly there are a few more places

interesting that in Postgres's [ruleutils](https://github.com/postgres/postgres/blob/7da51590ed6cd46ff886e8e4d08e8703db9c2b5b/src/backend/utils/adt/ruleutils.c#L10405-L10408) this not checked during the deparse of const expressions; maybe they've omitted the check because the expr is most likely was just parsed and the fn for deparsing should also exists - but I was really expecting some check there...

mkindahl (approved):

PR #5265, opened 2023-02-01
Remove SECURITY DEFINER from get_telemetry_report

svenklemm:
We should not broadly make functions security definer as that increases the attack surface for our extension. Especially for our telemetry we should strive to only run it with the minimum required privileges.


akuzm (approved):
It was just added here: https://github.com/timescale/timescaledb/pull/5218
Should we have some guidelines on this so that everybody has the same understanding of why it is problematic?

konskov:
I believe SECURITY DEFINER was added to the telmetry report function so that it could run with owner privileges. The owner is postgres and so, telemetry would get all the available information from job_errors. Let me do a test without SECURITY DEFINER.

konskov:
I'm approving, just a note: The telemetry process will be run as its owner, postgres, so it will have access to all the error data, regardless of SECURITY DEFINER. Also changing the query for aggregate job errors to query the view instead of the table will allow a regular user to call `get_telemetry_report()` and get information about their own jobs only.

konskov (approved):

PR #5267, opened 2023-02-01
Fix the python code style check 

akuzm:
For an unknown reason, pip started to install an older version of
prospector which is incompatible with the current pylint. Require the
new prospector version explicitly.


akuzm:
Seems to be some incompatibility between some prospector and pylint versions: https://github.com/PyCQA/prospector/issues/539

jnidzwetzki (approved):

svenklemm (approved):

PR #5270, opened 2023-02-01
Fix pylint again

akuzm:
Apparently a new version is out and it complains about Exception being too general.


svenklemm (approved):

akuzm:
Yes, I think I'll pin it, and the other tools as well. We can update them from time to time.

erimatnor (approved):

PR #5285, opened 2023-02-05
Don't reindex relation during decompress_chunk

svenklemm:
Reindexing a relation requires AccessExclusiveLock which prevents queries on that chunk. This patch changes decompress_chunk to update the index during decompression instead of reindexing.


tsl/src/compression/compression.c
```diff
@@ -1617,24 +1619,7 @@ decompress_chunk(Oid in_table, Oid out_table)
...
-	ReindexParams params = { 0 };
-	ReindexParams *options = &params;
-#endif
-
-	/* The reindex_relation() function creates an AccessExclusiveLock on the
-	 * chunk index (if present). After calling this function, concurrent
-	 * SELECTs have to wait until the index lock is released. When no
-	 * index is present concurrent SELECTs can be still performed in
-	 * parallel. */
-	DEBUG_WAITPOINT("decompress_chunk_impl_before_reindex");
```
jnidzwetzki, line 1635: These two `DEBUG_WAITPOINTs` are used in isolation tests (i.e., `debug_waitpoint_enable`). These tests should also be adjusted.

jnidzwetzki (approved):
Two regression tests are failing at the moment. I added a note about the debug waitpoints / isolation tests, the rest looks good to me.

tsl/src/compression/compression.c
svenklemm: Yes I saw that and stumbled across a cache invalidation issue. It looks we are missing cache invalidation when decompress_chunk deletes a chunk leading to invalid object references in parallel session. This might have been shadowed by the stronger locks we previously required which prevented parallel queries after decompression. I'll investigate.

lkshminarayanan: Hi @svenklemm, 

I looked into this issue. It is not due to a cache invalidation rather due to the standard `Read Committed` behaviour. 

So what happens is, 

**Session 1 :**

```
BEGIN;
SELECT count(*) FROM (SELECT decompress_chunk(i, if_compressed => true) 
  FROM show_chunks('sensor_data') i) i;
```

decompress_chunk takes ExclusiveLock on the sensor_data hypertable's compressed and decompressed chunks, performs the decompression, updates the related metadata in the internal chunk table and waits for the session to commit.

**Session 2 :**
```
SELECT FROM sensor_data;
```

In this parallel session `SELECT` now looks up the internal chunk table - does an index scan via `chunk_schema_name_table_name_key` index on the chunk table. Finds the chunk it is looking for and finds out that it is still compressed (as Session 1 has not committed yet). Reads the chunk details from the chunk table and finds out the OID of the related compressed chunk. When it tries to open the compressed chunk for reading, it is unable to as Session 1 still has the ExclusiveLock on it. (Note that Session 1 has a lock on the compressed chunk as it intends to delete it). So this parallel Session 2 waits for the lock on the compressed chunk. 

Once Session 1 commits, it deletes the compressed chunk and the Session 2 gets a `ERROR: could not open relation with OID <oid>` error.

antekresic (approved):
lgtm

PR #5290, opened 2023-02-06
Compression can't be enabled on continuous aggregates when segmentby/orderby columns need quotation

kgyrtkirk:
consider the sql commands:
```
drop table if EXISTS "tEst2" ;
-- Creating hypertable
CREATE TABLE IF NOT EXISTS "tEst2" (
    "Id" uuid NOT NULL,
    "Time" timestamp with time zone NOT NULL,
    CONSTRAINT "test2_pkey" PRIMARY KEY ("Id", "Time")
);

SELECT create_hypertable(
  '"tEst2"',
  'Time',
  chunk_time_interval => INTERVAL '1 day'
);

alter table "tEst2" set (timescaledb.compress=true, timescaledb.compress_segmentby='"Id"');

DROP MATERIALIZED VIEW if EXISTS "tEst2_mv";

CREATE MATERIALIZED VIEW "tEst2_mv"
SELECT "Id" as "Idd",
   time_bucket(INTERVAL '1 day', "Time") AS "bUcket"
FROM public."tEst2" 
GROUP BY "Idd", "bUcket";

ALTER MATERIALIZED VIEW "tEst2_mv" SET (timescaledb.compress = true);
```

before this patch the last command failed because:
* the `Idd` column lost quotation; leading to `ERROR:  column "idd" does not exist` 
* or the `bUcket` column lost quotation; leading to `ERROR:  column "bucket" does not exist`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ accidentally triggering the load of a previous DB version.**
 * #4926 Fix corruption when inserting into compressed chunks
 * #5218 Add role-level security to job error log
 * #5214 Fix use of prepared statement in async module
+* #5290 Compression can't be enabled on continuous aggregates when segmentby/orderby columns need quotation
 
 ## 2.9.3 (2023-02-03)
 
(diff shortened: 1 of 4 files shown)


svenklemm (changes requested):
doing `DROP IF EXISTS` is not necessary in tests since they each run in dedicated databases

kgyrtkirk:
@svenklemm done

akuzm (approved):

mkindahl (approved):

svenklemm (approved):

PR #5297, opened 2023-02-09
Warn about mismatched chunk cache sizes

akuzm:
Just noticed abysmal INSERT performance when experimenting with one of our customers' data set, and turns out my cache sizes were misconfigured, leading to constant hypertable chunk cache thrashing. Show a warning to detect this misconfiguration. Also use more generous defaults, we're not supposed to run on a microwave (unlike Postgres).

### DIFF
--- a/src/guc.c
+++ b/src/guc.c
@@ -107,6 +107,12 @@ bool ts_shutdown_bgw = false;
 char *ts_current_timestamp_mock = "";
 #endif
 
+/*
+ * We have to understand if we have finished initializing the GUCs, so that we
+ * know when it's OK to check their values for mutual consistency.
+ */
+static bool gucs_are_initialized = false;
+
 /* Hook for plugins to allow additional SSL options */
 set_ssl_options_hook_type ts_set_ssl_options_hook = NULL;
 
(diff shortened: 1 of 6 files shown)


src/guc.c
```diff
@@ -453,14 +453,7 @@ _guc_init(void)
...
 							&ts_guc_max_open_chunks_per_insert,
-							Min(work_mem * INT64CONST(1024) / INT64CONST(25000),
-								PG_INT16_MAX), /* Measurements via
-												* `MemoryContextStats(TopMemoryContext)`
-												* show chunk insert
-												* state memory context
-												* takes up ~25K bytes
-												* (work_mem is in
-												* kbytes) */
+							1024,
```
akuzm, line 456: These old measurements are not really relevant, because chunk lookup was redone several times, and we also have multi-insert now which completely changes the memory usage.

kgyrtkirk: I think the old default was not neccessarily the best - as I guess `work_mem` should be available to be used for other things - but the formula planned to use all of it to be used for this cache only
note: I wonder if the 25k estimate still stands...this formula was last changed 3 years ago

with the new constant 1024 and using the 25k estimate from the old comment; the max cache size is: `25K*1024 ~ 25M` ; which could be greater than `work_mem` - I think the default `work_mem=4M` 

I don't know if there is any counter infra which could be added to the underlying cache to track hit/miss counts and size so that trashing/etc can be tracked/identified

src/guc.c
```diff
@@ -473,7 +466,7 @@ _guc_init(void)
 							"Maximum cached chunks",
 							"Maximum number of chunks stored in the cache",
 							&ts_guc_max_cached_chunks_per_hypertable,
-							100,
+							1024,
```
kgyrtkirk, line 469: we seem to be caching chunks for  [up-to-16 hypertables](https://github.com/timescale/timescaledb/blob/caf79e0f5e3b789f70361f4fd3093afcea3a1da0/src/hypertable_cache.c#L68) at a time..

so if the above 25K per entry applies here as well - this cache will probably take up at tops `25K*1024*16 ~ 406MB` with the current setting - I'm not sure if that should be allowed in the default configuration

jnidzwetzki:
@akuzm Could you trigger a tsbench run for this branch? I don't think that the better default values should introduce a performance regression. But just to be sure.

src/guc.c
akuzm: If a session touches 16 hypertables with 1024 chunks with a modest size of 1 GB each, it is processing 16 TB of data. I think it is reasonable to use some 400 MB for metadata in this case.

src/guc.c
akuzm: Not sure the size of our various metadata caches should be governed by `work_mem`. It's basically how much memory postgres uses for sort before it starts spilling to disk. The Postgres defaults like 4MB are outdated and don't really apply if we're working with a 1000-chunk (this means a terabyte at least) hypertable. `timescaledb-tune` should configure more sensible values.

akuzm:
> @akuzm Could you trigger a tsbench run for this branch? I don't think that the better default values should introduce a performance regression. But just to be sure.

Doesn't look like we have any meaningful changes. Some cagg_simple queries are up to 10% slower, but it's within their normal variability, and doesn't reproduce locally for me.

https://grafana.ops.savannah-dev.timescale.com/d/NdmLnOk4z/compare-benchmark-runs?orgId=1&var-branch=All&var-postgres=14&var-run1=1894&var-run2=1896&var-threshold=0.02

src/guc.c
akuzm: For context, on my machine `timescaledb-tune` sets `work_mem` to 5 MB :confused: 
The old default value for `timescaledb.max_open_chunks_per_insert` with this setting is 208.

akuzm: > The old default value for timescaledb.max_open_chunks_per_insert with this setting is 208.

Not great, not terrible, still leads to cache thrashing combined with the other old default.

svenklemm (changes requested):
I dont think we want to do this in chunk_dispatch, this will trigger the warning with no way to disable it on every insert. I think we might wanna do this once on extension load anything else will be very spammy.

Loquacity:
Question: is it worth adding something to the docs about this?

akuzm:
> Question: is it worth adding something to the docs about this?

I think no, these are somewhat obscure parameters that we don't want our users to touch, it should "just work".

akuzm:
> I dont think we want to do this in chunk_dispatch, this will trigger the warning with no way to disable it on every insert. I think we might wanna do this once on extension load anything else will be very spammy.

Done

jnidzwetzki (approved):

svenklemm (approved):

PR #5313, opened 2023-02-13
Add timeout argument to the ping_data_node()

### LINKED ISSUE #5312: [Feature]: Add timeout support to the ping_data_node function

API improvement, Tech debt reduction, User experience

Data node

Add a way to specify ping out for the ping_data_node function by reaching which function should return false instead of being blocked

### Implementation challenges

_No response_

pmwkaa:
This PR introduces a timeout argument and a new logic to the timescale_internal.ping_data_node() function which allows to handle io timeouts for nodes being unresponsive.

Fix #5312


pmwkaa:
tests are coming

sql/data_node.sql
```diff
@@ -3,7 +3,7 @@
 -- LICENSE-APACHE for a copy of the license.
 
 -- Check if a data node is up
-CREATE OR REPLACE FUNCTION _timescaledb_internal.ping_data_node(node_name NAME) RETURNS BOOLEAN
+CREATE OR REPLACE FUNCTION _timescaledb_internal.ping_data_node(node_name NAME, timeout INTEGER = NULL) RETURNS BOOLEAN
```
erimatnor, line 6: What's the unit here? Is it seconds or milliseconds?
I think it would be better to use an `Interval` type so that one can type, e.g., `4 seconds` or `4s`.

Another option is to use a timestamp to set a "deadline": `ping_data_node('dn1', now() + interval '4s')`

erimatnor (commented):
I suggest that the interface should use an interval or timestamp type that have units (see inline comment).

As a side note: we should also do this to the health check function, which is the function I think we should use instead of ping.

sql/updates/reverse-dev.sql
```diff
@@ -3,3 +3,6 @@ GRANT ALL ON _timescaledb_internal.job_errors TO PUBLIC;
 ALTER EXTENSION timescaledb DROP VIEW timescaledb_information.job_errors;
 
 DROP VIEW timescaledb_information.job_errors;
+
+CREATE OR REPLACE FUNCTION _timescaledb_internal.ping_data_node(node_name NAME) RETURNS BOOLEAN
```
nikkhils, line 7: I believe we will need to DROP the new API first and then create with older signature?

tsl/src/data_node.c
```diff
@@ -2056,14 +2056,16 @@ Datum
 data_node_ping(PG_FUNCTION_ARGS)
 {
 	const char *node_name = PG_ARGISNULL(0) ? NULL : PG_GETARG_CSTRING(0);
+	TimestampTz timeout = PG_ARGISNULL(1) ? TS_NO_TIMEOUT : PG_GETARG_TIMESTAMPTZ(1);
```
nikkhils, line 2059: This has to be `PG_GETARG_INTERVAL_P` right?

nikkhils: and type has to be `Interval`

tsl/src/remote/connection.c
```diff
@@ -1623,7 +1689,7 @@ TSConnection *
 remote_connection_open_session(const char *node_name, List *connection_options, bool set_dist_id)
 {
 	char *err = NULL;
-	TSConnection *conn = remote_connection_open(node_name, connection_options, &err);
+	TSConnection *conn = remote_connection_open(node_name, connection_options, TS_NO_TIMEOUT, &err);
```
akuzm, line 1692: Is it ever a good idea to wait w/o timeout?

As an aside, the `TS_NO_TIMEOUT` define is pretty confusing, it's `DT_NOBEGIN`, which is in turn `PG_INT64_MIN`, which happens to match the `(long) -1` used by `WaitEventSetWait` on 64-bit systems, but not on 32-bit ones I guess? Does this lead to any problems?

akuzm (approved):

tsl/src/remote/connection.c
```diff
@@ -933,11 +951,23 @@ remote_connection_get_result(const TSConnection *conn)
...
+			uint32 events;
 			WaitEvent event;
+			long timeout_ms;
+			int ret;
 
-			ModifyWaitEvent(conn->wes, conn->sockeventpos, WL_SOCKET_READABLE, NULL);
-			WaitEventSetWait(conn->wes, -1, &event, 1, PG_WAIT_EXTENSION);
+			events = WL_SOCKET_READABLE;
+			if (endtime != TS_NO_TIMEOUT)
+				events |= WL_TIMEOUT;
```
akuzm, line 961: For simplicity, I'd specify `WL_TIMEOUT` always, no harm in it.

pmwkaa: it requires timeout >= 0

tsl/src/data_node.c
pmwkaa: Changed type to timestamptz in the function definition

tsl/src/remote/connection.c
pmwkaa: I think TS_NO_TIMEOUT is for TimestampTz type, not for the timeout itself, which is calculated based on it and has long type

erimatnor: @akuzm Yes, it makes sense to wait without timeouts. Queries typically time out based on `statement_timeout` or interrupt by the user (ctrl-c). Both of these generate a signal that will trigger the latch, which is part of the wait set.

tsl/src/data_node.c
```diff
@@ -2056,14 +2056,28 @@ Datum
...
 
 	Assert(NULL != server);
 
-	success = remote_connection_ping(server->servername);
+	if (timeout)
+	{
+		if (timeout->day || timeout->month)
+			ereport(ERROR,
+					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+					 errmsg("ping interval timeout is too large")));
```
erimatnor, line 2074: Why should we check this and make the decision a timeout is too big? What's the use case?

erimatnor (approved):

tsl/src/data_node.c
erimatnor: I think a more reasonable sanity check is to ensure that ping timeout is always smaller than statement_timeout. Otherwise, the ping function will always time out on statement_timeout before reaching its own timeout.

pmwkaa: Shouldn't this be expected? Ok, computing month and day now.

PR #5211, opened 2023-01-23
Recompress chunk rewrite

konskov:

PR #5321, opened 2023-02-14
Rewrite recompress_chunk for segmentwise processing

konskov:
This patch introduces a C-function to perform the recompression at a finer granularity instead of decompressing and subsequently compressing the entire chunk.

This improves performance for the following reasons:
- it needs to sort less data at a time and
- it avoids recreating the decompressed chunk and the heap inserts associated with that by decompressing each segment into a tuplesort instead. 

If no segmentby is specified when enabling compression or if an index does not exist on the compressed chunk then the operation is performed as before, decompressing and subsequently compressing the entire chunk.


konskov:
Failing tests that need fixing: 

- [ ] compression_bgw
- [ ] compression_ddl_iso
- [ ] compression_ddl
- [ ] dist_compression

konskov:
I might close #5211 in favor of this one

konskov:
TODOs:
- [ ] test prepared statements and status flags

sql/maintenance_utils.sql
```diff
@@ -86,16 +97,23 @@ BEGIN
...
-        PERFORM @extschema@.decompress_chunk(chunk);
-        COMMIT;
-        -- SET LOCAL is only active until end of transaction.
-        -- While we could use SET at the start of the function we do not
-        -- want to bleed out search_path to caller, so we do SET LOCAL
-        -- again after COMMIT
-        SET LOCAL search_path TO pg_catalog, pg_temp;
+        -- first check if there's an index. Might have to use a heuristic to determine if index usage would be efficient,
+        -- or if we'd better fall back to decompressing & recompressing entire chunk
+        SELECT _timescaledb_internal.get_compressed_chunk_index_for_recompression(chunk) INTO STRICT compressed_chunk_index;
```
jnidzwetzki, line 102: Do you need the index name or would it be enough to have a more generic function like `is_recompress_chunk_segmentwise_supported` here? Such a function would encapsulate the functionality and not publish unnecessary details about the internals.

tsl/src/compression/api.c
```diff
@@ -44,6 +45,9 @@
 #include "scanner.h"
 #include "scan_iterator.h"
 #include "utils.h"
+#include "chunk.h"
+#include "custom_type_cache.h" // for the compressed data datatype
+#include <utils/snapmgr.h>
```
jnidzwetzki, line 50: Please consider moving this to the group of other `utils` headers (i.e., line 27).

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
 	tsl_compress_chunk_wrapper(chunk, false);
 	return true;
 }
+
+/* this should be a wrapper around row_compressor_append_sorted_rows */
```
jnidzwetzki, line 900:
```suggestion
/* This is a wrapper around row_compressor_append_sorted_rows. */
```

sql/maintenance_utils.sql
konskov: You are right, I don't need the index oid, I just want to verify that the index exists. I will make the suggested change

sql/updates/reverse-dev.sql
```diff
@@ -8,3 +8,6 @@ DROP FUNCTION _timescaledb_internal.ping_data_node(NAME, INTERVAL);
 
 CREATE OR REPLACE FUNCTION _timescaledb_internal.ping_data_node(node_name NAME) RETURNS BOOLEAN
 AS '@MODULE_PATHNAME@', 'ts_data_node_ping' LANGUAGE C VOLATILE;
+
+DROP FUNCTION _timescaledb_internal.recompress_chunk_segmentwise(REGCLASS, BOOLEAN);
+DROP FUNCTION _timescaledb_internal.get_compressed_chunk_index_for_recompression(REGCLASS);
```
svenklemm, line 13: you need to have mirror-statements in latest-dev that create the function

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+ * instead, when an index is found.
+ */
+extern Datum
+tsl_get_compressed_chunk_index_for_recompression(PG_FUNCTION_ARGS)
+{
+	Oid uncompressed_chunk_id = PG_ARGISNULL(0) ? InvalidOid : PG_GETARG_OID(0);
+	Chunk *uncompressed_chunk = ts_chunk_get_by_relid(uncompressed_chunk_id, true);
+
+	if (NULL == uncompressed_chunk)
+		elog(ERROR, "unknown chunk id %d", uncompressed_chunk_id);
```
svenklemm, line 995: You are calling `ts_chunk_get_by_relid` so that it errors when the chunk does not exist so this code is redundant.

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	{
+		current_segment[i] = palloc(sizeof(CompressedSegmentInfo));
+	}
+	bool first_iteration = true;
+
+	/************** snapshot ****************************/
+	Snapshot snapshot = RegisterSnapshot(GetTransactionSnapshot());
+
+	/* Index scan */
+	elog(DEBUG1, "Index oid for recomrpess is %u", row_compressor.index_oid);
```
svenklemm, line 1320:
```suggestion
	elog(DEBUG1, "Index oid for recompress is %u", row_compressor.index_oid);
```

tsl/src/compression/compression.h
```diff
@@ -153,4 +287,44 @@ extern DecompressionIterator *(*tsl_get_decompression_iterator_init(
...
+extern void populate_per_compressed_columns_from_data(PerCompressedColumn *per_compressed_cols,
+													  int16 num_cols, Datum *compressed_datums,
+													  bool *compressed_is_nulls);
+extern void row_compressor_append_sorted_rows(RowCompressor *row_compressor,
+											  Tuplesortstate *sorted_rel, TupleDesc sorted_desc);
+extern void segment_info_update(SegmentInfo *segment_info, Datum val, bool is_null);
+
+extern Oid get_compressed_chunk_index(Relation compressed_chunk,
+									  int16 *uncompressed_col_to_compressed_col,
+									  PerColumn *per_column, int n_input_columns);
```
svenklemm, line 327: Your PR does not add a caller for this so this seems unnecessary.  The function is currently also static but it is only called locally in your PR so there is no reason to change that.

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+			PG_RETURN_OID(uncompressed_chunk_id);
+	}
+	int32 srcht_id =
+		uncompressed_chunk->fd.hypertable_id; // need it to find the segby cols from the catalog
+	Chunk *compressed_chunk = ts_chunk_get_by_id(uncompressed_chunk->fd.compressed_chunk_id, true);
+
+	List *htcols_list = ts_hypertable_compression_get(srcht_id);
+	ListCell *lc;
+	int htcols_listlen = list_length(htcols_list);
+	// find segmentby columns (need to know their index)
```
jnidzwetzki, line 1013: I'm not sure what the outcome of the discussion about the comments was, but I would suggest using consistent comment characters within a function.

```suggestion
	/* find segmentby columns (need to know their index) */
```

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+{
+	TableScanDesc heapScan;
+	HeapTuple uncompressed_tuple;
+	TupleDesc uncompressed_rel_tupdesc = RelationGetDescr(uncompressed_chunk_rel);
+	ScanKeyData *scankey = palloc0(sizeof(*scankey) * nsegmentby_cols);
+
+	for (int seg_col = 0; seg_col < nsegmentby_cols; seg_col++)
+	{
+		Datum val = current_segment[seg_col]->segment_info->val;
+		// what is the attno of this segby col in the uncompressed chunk?
```
jnidzwetzki, line 1116:
```suggestion
		/* Get the attno of this segby col in the uncompressed chunk */ 
```

svenklemm:
I think we should add a test that ensures prepared statements are replanned when chunk status changes from/to partial compressed.

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+		elog(ERROR, "unknown chunk id %d", uncompressed_chunk_id);
+
+	/* need to push down to data nodes if this is an access node */
+	if (uncompressed_chunk->relkind == RELKIND_FOREIGN_TABLE)
+	{
+		if (!recompress_remote_chunk(fcinfo, uncompressed_chunk))
+			PG_RETURN_NULL();
+
+		/* TODO: probably redundant? */
+		ts_chunk_set_compressed_chunk(uncompressed_chunk, INVALID_CHUNK_ID);
```
jnidzwetzki, line 1180: Could you check this in the final version of the PR?

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	}
+
+	/*
+	 * only proceed if status in (3, 9, 11)
+	 * 1: compressed
+	 * 2: compressed_unordered
+	 * 4: frozen
+	 * 8: compressed_partial
+	 */
+	if (status != 3 && status != 9 && status != 11)
```
jnidzwetzki, line 1205: I think the readability of the condition could be improved if you use `ts_chunk_is_unordered` / `ts_chunk_is_compressed` / `ts_chunk_is_partial` here.

jnidzwetzki:
@konskov Might it be possible to explain in more detail in the commit message why this change improves recompression? It would be great to document the idea in the commit history.

tsl/src/compression/compression.c
```diff
@@ -1761,11 +1623,34 @@ row_decompressor_decompress_row(RowDecompressor *decompressor)
...
+							0 /*=options*/,
+							decompressor->bistate);
+			}
+			else
+			{
+				TupleTableSlot *slot =
+					MakeSingleTupleTableSlot(decompressor->out_desc, &TTSOpsVirtual);
+
+				// create the virtual tuple slot
+				ExecClearTuple(slot); // might not need this
```
jnidzwetzki, line 1640: Please check this in the final version of the PR.

tsl/src/compression/compression.h
```diff
@@ -12,6 +12,11 @@
 #include <fmgr.h>
 #include <lib/stringinfo.h>
 #include <utils/relcache.h>
+#include <access/heapam.h> // for BulkInsertState
+#include <utils/tuplesort.h>
+#include "segment_meta.h"
+#include <utils/sortsupport.h>
```
jnidzwetzki, line 18: I would suggest to swap Lines 17 and 18 to group the `utils` includes

tsl/test/sql/recompress_chunk_segmentwise.sql
```diff
@@ -0,0 +1,134 @@
...
+
+ALTER TABLE test_defaults ADD COLUMN c1 int;
+ALTER TABLE test_defaults ADD COLUMN c2 int NOT NULL DEFAULT 42;
+SELECT * FROM test_defaults ORDER BY 1,2;
+
+INSERT INTO test_defaults SELECT '2000-01-01', 2;
+SELECT * FROM test_defaults ORDER BY 1,2;
+
+call recompress_chunk(:'compressed_chunk');
+SELECT * FROM test_defaults ORDER BY 1,2;
```
jnidzwetzki, line 134: Do you have a test to check which strategy is chosen by `recompress_chunk`?  

One idea is adding a `DEBUG1` log message to both strategies and checking the chosen strategy in the regression tests.

tsl/src/compression/api.c
```diff
@@ -892,3 +896,584 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+ *
+ * 0 uncompressed_chunk_id REGCLASS
+ * 1 if_not_compressed BOOL = false
+ */
+Datum
+tsl_recompress_chunk_segmentwise(PG_FUNCTION_ARGS)
+{
+	Oid uncompressed_chunk_id = PG_ARGISNULL(0) ? InvalidOid : PG_GETARG_OID(0);
+	bool if_not_compressed =
+		PG_ARGISNULL(1) ? false : PG_GETARG_BOOL(1); // probably don't need this
```
jnidzwetzki, line 1165: Please resolve this open question in the final version of the PR.

tsl/src/compression/api.c
konskov: The outcome was that I will look for formatting tools we can use and then create a poll for us to decide, but I haven't done this yet. Regarding the comments in this PR, I will fix them all to follow the `/**/` style

tsl/src/compression/api.c
konskov: right, removed it

sql/maintenance_utils.sql
konskov: changed it, if you'd like to take another look

tsl/src/compression/compression.c
konskov: sorry for the confusion here. I believe it is needed, so I removed the misleading comment. From `ExecStoreVirtualTuple`: 
```
 * The protocol for loading a slot with virtual tuple data is:
 *		* Call ExecClearTuple to mark the slot empty.
 *		* Store data into the Datum/isnull arrays.
 *		* Call ExecStoreVirtualTuple to mark the slot valid.
 ```

File: sql/maintenance_utils.sql (hunk 1 above)
Comment (konskov) [reply sql/maintenance_utils.sql]: Actually the changes I made wrt this broke the CI so I'm still working on it 😅

**Comment by konskov** [comment]
I have included a test case for prepared statements in `recompress_chunk_segmentwise.sql`

File: tsl/src/compression/api.c (hunk 10 above)
Comment (konskov) [reply tsl/src/compression/api.c]: It was redundant because I am not creating a new compressed chunk, and for the existing compressed chunk, this is already marked Invalid upon its creation. So I removed it.

File: sql/maintenance_utils.sql (hunk 1 above)
Comment (konskov) [reply sql/maintenance_utils.sql:102]: The reason I am using this Oid returning function is because this way I could use the existing `invoke_compression_func_remotely` function that is used for compress and decompress, which works with Oids. And that way did not have to write additional functions.

Comment (konskov) [reply sql/maintenance_utils.sql:102]: Alternatively, I would either write a similar function to `invoke_compression_func_remotely` that works with bool datums, or I would move what is currently `get_compressed_chunk_index_for_recompression` in an internal function. Which option do you think would be best?

File: sql/updates/reverse-dev.sql (hunk 4 above)
Comment (konskov) [reply sql/updates/reverse-dev.sql:7]: done

**Comment by konskov** [comment]
Due to the fact that we no longer decompress into the uncompressed chunk, we are not able to report the uncompressed chunk size accurately anymore. So the information on uncompressed_heap_size, uncompressed_toast_size, uncompressed_index_size stored in the catalog table `_timescaledb_catalog.compression_chunk_size` will not be up-to-date if the recompression has been performed segmentwise. 
We discussed this with the team and as a way forward, decided to document this and make a note of which informational views and size reporting functions will be affected. I will edit this comment to include the list of these. 
There is a workaround in order to get up-to-date information: manually decompress + compress the chunk.

**Comment by konskov** [comment]
@jnidzwetzki I have updated the commit message in the PR description. How does it look now?

File: tsl/src/compression/api.c (hunk 11 above)
Comment (konskov) [reply tsl/src/compression/api.c]: done

File: tsl/src/compression/api.c (hunk 15 above)
Comment (konskov) [reply tsl/src/compression/api.c]: I think it couldn't hurt to check this again, so decided to leave it there.

File: tsl/test/sql/recompress_chunk_segmentwise.sql (hunk 14 above)
Comment (konskov) [reply tsl/test/sql/recompress_chunk_segmentwise.sql:164]: Added, as suggested

Comment (konskov) [reply tsl/test/sql/recompress_chunk_segmentwise.sql:164]: I also need to find a way to remove/suppress some additional debug prints so I unresolved the conversation until then

File: tsl/src/compression/compression.c [hunk 16]
```diff
@@ -1180,15 +1091,19 @@ row_compressor_update_group(RowCompressor *row_compressor, TupleTableSlot *row)
...
 												  row_compressor
-													  ->sequence_num_metadata_column_offset));
+													  ->uncompressed_col_to_compressed_col,
+												  row_compressor->per_column,
+												  row_compressor->n_input_columns,
+												  AttrOffsetGetAttrNumber(
+													  row_compressor
+														  ->sequence_num_metadata_column_offset));
+	else
+		row_compressor->sequence_num = SEQUENCE_NUM_GAP;
```
svenklemm, line 1106: do we have a test that shows the sequence number is increasing in recompressed segments? dont we need to increment this?

tsl/test/expected/recompress_chunk_segmentwise.out
```diff
@@ -0,0 +1,419 @@
...
+alter table mytab set (timescaledb.compress = false);
+alter table mytab set (timescaledb.compress);
+select compress_chunk(show_chunks('mytab'));
+              compress_chunk              
+------------------------------------------
+ _timescaledb_internal._hyper_11_12_chunk
+(1 row)
+
+INSERT INTO mytab VALUES ('2023-01-01'::timestamptz, 2, 3, 2);
+set client_min_messages = DEBUG;
```
svenklemm, line 403: This will lead to a flaky tests as you will randomly get DEBUG messages from other postgres subsystems. Can we change the test to not use DEBUG log level?

konskov: I changed the test to verify which method is used for recompression by checking whether the compressed chunk is the existing one or a new one after the operation completes.

konskov:
TODO: properly handle NULL values

tsl/src/compression/compression.h
```diff
@@ -114,6 +191,63 @@ typedef struct CompressionStats
...
+	MemoryContext per_row_ctx;
+
+	/* the table we're writing the compressed data to */
+	Relation compressed_table;
+	BulkInsertState bistate;
+	/* segment by index Oid if any */
+	Oid index_oid;
+
+	/* in theory we could have more input columns than outputted ones, so we
+	   store the number of inputs/compressors seperately*/
```
jnidzwetzki, line 223: Nit:

```suggestion
	   store the number of inputs/compressors separately */
```

tsl/src/compression/compression.c
```diff
@@ -1761,11 +1623,34 @@ row_decompressor_decompress_row(RowDecompressor *decompressor)
...
+			{
+				heap_insert(decompressor->out_rel,
+							decompressed_tuple,
+							decompressor->mycid,
+							0 /*=options*/,
+							decompressor->bistate);
+			}
+			else
+			{
+				TupleTableSlot *slot =
```
jnidzwetzki, line 1636: I wonder if the re-creation of the slot is necessary. Couldn't you create it once and reuse it in the loop here?

tsl/src/compression/compression.h
konskov: fixed

tsl/src/compression/api.c
```diff
@@ -892,3 +976,649 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+						colinfo_array,
+						in_column_offsets,
+						compressed_rel_tupdesc->natts,
+						true /*need_bistate*/,
+						true /*recompressing*/);
+
+	if (OidIsValid(row_compressor.index_oid))
+	{
+		/* lock in least strict mode so it's not dropped by other sessions */
+		Relation index_rel = relation_open(row_compressor.index_oid, AccessExclusiveLock);
```
svenklemm, line 1133: Thats not least strict mode

svenklemm: Wouldnt we want that lock on the chunk and not the index, why are we locking the index specifically?

tsl/src/compression/api.c
```diff
@@ -892,3 +976,649 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	if (OidIsValid(row_compressor.index_oid))
+	{
+		PG_RETURN_OID(uncompressed_chunk_id);
+	}
+	else
+		PG_RETURN_NULL();
+}
+
+static void
+fetch_unmatched_uncompressed_chunk_into_tuplestore(Tuplesortstate *segment_tuplesortstate,
```
svenklemm, line 1151: can you add a comment here explaining what unmatched refers to

tsl/src/compression/api.c
```diff
@@ -892,3 +976,649 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+		pfree(null_segbycols);
+
+	if (scankey != NULL)
+		pfree(scankey);
+}
+
+/*
+ * This function will replace the procedure for recompress_chunk
+ * It will keep the compressed chunk around, only decompressing the
+ * batches that are affected by the addition of newer data.
```
svenklemm, line 1273:
```suggestion
 * Recompress an existing chunk by decompressing the
 * batches that are affected by the addition of newer data.
 * The existing compressed chunk will not be recreated but
 * modified in place.
```

tsl/src/compression/api.c
```diff
@@ -892,3 +976,649 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	{
+		current_segment[i] = palloc(sizeof(CompressedSegmentInfo));
+	}
+	bool current_segment_init = false;
+
+	/************** snapshot ****************************/
+	Snapshot snapshot = RegisterSnapshot(GetTransactionSnapshot());
+
+	/* Index scan */
+	elog(DEBUG1, "Index oid for recompress is %u", row_compressor.index_oid);
```
svenklemm, line 1443: do we need that debug message?

tsl/src/compression/api.c
```diff
@@ -892,3 +976,649 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	Relation index_rel = index_open(row_compressor.index_oid, AccessExclusiveLock);
```
svenklemm, line 1444: Why do we need AccessExclusiveLock here?

tsl/src/compression/api.c
konskov: Yes the comment is misleading, it was there before I changed the AccessShareLock to AccessExclusive and I forgot to adjust it. I'll fix that. 
The AccessExclusiveLock on the index is taken because eventually, we will reindex the compressed chunk relation, which will request an AccessExclusiveLock. Requesting it in the beginning of the transaction prevents a deadlock between two concurrent recompress operations. 

Then again maybe a deadlock there is not so unacceptable as one of the recompression operations should fail anyway.

tsl/src/compression/api.c
konskov: removed.

tsl/src/compression/api.c
konskov: applied suggestion

tsl/src/compression/api.c
konskov: basically I did this to avoid the deadlock between two concurrent recompress operations https://github.com/timescale/timescaledb/pull/5321#discussion_r1136639296 but that deadlock might be acceptable?

tsl/src/compression/api.c
konskov: done

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	HeapTuple uncompressed_tuple;
+	TupleDesc uncompressed_rel_tupdesc = RelationGetDescr(uncompressed_chunk_rel);
+
+	heapScan = table_beginscan(uncompressed_chunk_rel, GetLatestSnapshot(), 0, NULL);
+
+	for (uncompressed_tuple = heap_getnext(heapScan, ForwardScanDirection);
+		 uncompressed_tuple != NULL;
+		 uncompressed_tuple = heap_getnext(heapScan, ForwardScanDirection))
+	{
+		TupleTableSlot *heap_tuple_slot =
```
jnidzwetzki, line 1171: Couldn't this `heap_tuple_slot ` be cleared and reused in the entire loop? This could improve performance.

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	}
+
+	heapScan =
+		table_beginscan(uncompressed_chunk_rel, GetLatestSnapshot(), nsegbycols_nonnull, scankey);
+
+	for (uncompressed_tuple = heap_getnext(heapScan, ForwardScanDirection);
+		 uncompressed_tuple != NULL;
+		 uncompressed_tuple = heap_getnext(heapScan, ForwardScanDirection))
+	{
+		TupleTableSlot *heap_tuple_slot =
```
jnidzwetzki, line 1241: Same here. Can the slot be re-used in the loop?

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+			row_decompressor_decompress_row(&decompressor, segment_tuplesortstate);
+
+			simple_table_tuple_delete(compressed_chunk_rel, &(slot->tts_tid), snapshot);
+			changed_segment = false; // because this is the first tuple of the new segment
+			/* make changes visible */
+			CommandCounterIncrement();
+
+			if (should_free)
+				heap_freetuple(compressed_tuple);
+			ExecClearTuple(slot);
```
jnidzwetzki, line 1554: I am not sure if I follow the code flow here. We have a while loop over `index_getnext_slot`. The slot is cleared in the `if` branch (Line 1505) and here in the `if else` branch. Shouldn't it be cleared in every iteration of the while loop?

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	int i = 0, htcols_listlen;
+	ListCell *lc;
+
+	int32 srcht_id =
+		uncompressed_chunk->fd.hypertable_id; // need it to find the segby cols from the catalog
+	Chunk *compressed_chunk = ts_chunk_get_by_id(uncompressed_chunk->fd.compressed_chunk_id, true);
+
+	List *htcols_list = ts_hypertable_compression_get(srcht_id);
+	htcols_listlen = list_length(htcols_list);
+	// find segmentby columns (need to know their index)
```
jnidzwetzki, line 1340: Nit:

```suggestion
	/* find segmentby columns (need to know their index) */
```

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+			 "unexpected chunk status %d in chunk %s.%s",
+			 status,
+			 uncompressed_chunk->fd.schema_name.data,
+			 uncompressed_chunk->fd.table_name.data);
+
+	int i = 0, htcols_listlen;
+	ListCell *lc;
+
+	int32 srcht_id =
+		uncompressed_chunk->fd.hypertable_id; // need it to find the segby cols from the catalog
```
jnidzwetzki, line 1335: Nit

```suggestion
        /* need it to find the segmentby columns from the catalog */
	int32 srcht_id =
		uncompressed_chunk->fd.hypertable_id; 
```

tsl/src/compression/compression.c
jnidzwetzki: 👍  It seems it is needed here. So, could you remove the comment?

tsl/src/compression/compression.c
```diff
@@ -1752,14 +1612,36 @@ row_decompressor_decompress_row(RowDecompressor *decompressor)
...
+							decompressed_tuple,
+							decompressor->mycid,
+							0 /*=options*/,
+							decompressor->bistate);
+
+				ts_catalog_index_insert(decompressor->indexstate, decompressed_tuple);
+			}
+			else
+			{
+				// create the virtual tuple slot
```
jnidzwetzki, line 1629: Nit:

```suggestion
				/* create the virtual tuple slot */
```

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+	TupleDesc uncompressed_rel_tupdesc = RelationGetDescr(uncompressed_chunk_rel);
+	int index = 0;
+	int nsegbycols_nonnull = 0;
+	Bitmapset *null_segbycols = NULL;
+
+	for (int seg_col = 0; seg_col < nsegmentby_cols; seg_col++)
+	{
+		if (!current_segment[seg_col]->segment_info->is_null)
+			nsegbycols_nonnull++;
+		// also build a bms of null columns to check later on
```
jnidzwetzki, line 1203: Nit:

```suggestion
		/* also build a bms of null columns to check later on */
```

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+			fetch_matching_uncompressed_chunk_into_tuplestore(segment_tuplesortstate,
+															  nsegmentby_cols,
+															  uncompressed_chunk_rel,
+															  current_segment);
+
+			tuplesort_performsort(segment_tuplesortstate);
+
+			recompress_segment(segment_tuplesortstate, uncompressed_chunk_rel, &row_compressor);
+
+			tuplesort_end(segment_tuplesortstate); // now any pointers returned will be garbage
```
jnidzwetzki, line 1518: Nit:

```suggestion
                         /* now any pointers returned will be garbage */
			tuplesort_end(segment_tuplesortstate); 
```

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+
+			heap_deform_tuple(compressed_tuple,
+							  compressed_rel_tupdesc,
+							  decompressor.compressed_datums,
+							  decompressor.compressed_is_nulls);
+
+			row_decompressor_decompress_row(&decompressor, segment_tuplesortstate);
+
+			simple_table_tuple_delete(compressed_chunk_rel, &(slot->tts_tid), snapshot);
+			changed_segment = false; // because this is the first tuple of the new segment
```
jnidzwetzki, line 1548: Nit:

```suggestion
                         /* because this is the first tuple of the new segment */
			changed_segment = false; 
```

tsl/test/sql/recompress_chunk_segmentwise.sql
konskov: in order to test this, I removed the debug prints and check whether the compressed chunk after recompression is the same or not because this differs depending on the strategy. So marking conversation as resolved.

tsl/src/compression/api.c
konskov: Yes you are right, thank you for pointing this out. Fixed.

jnidzwetzki (approved):

tsl/src/compression/compression.h
```diff
@@ -11,7 +11,13 @@
 #include <executor/tuptable.h>
 #include <fmgr.h>
 #include <lib/stringinfo.h>
+#include <access/heapam.h>
 #include <utils/relcache.h>
+#include <utils/tuplesort.h>
```
sb230132, line 16: Do you need these 2 header files to be included ?
Adding unwanted header files will increase dependency between different c files and can cause increase in source code compilation times.

tsl/src/compression/compression.h
```diff
@@ -114,6 +194,63 @@ typedef struct CompressionStats
...
+	/* a unique monotonically increasing (according to order by) id for each compressed row */
+	int32 sequence_num;
+
+	/* cached arrays used to build the HeapTuple */
+	Datum *compressed_values;
+	bool *compressed_is_null;
+	int64 rowcnt_pre_compression;
+	int64 num_compressed_rows;
+	/* if recompressing, we must know this so we can reset the sequence number */
+	bool recompressing;
```
sb230132, line 251: If iam not wrong we reset sequence number only for old recompress ( decompress + compress).
May be rename it to is_segmentwise_recompress for better readability.

sb230132 (commented):
Thanks for all the efforts in getting this feature to closure. Its a great improvement to timescaledb.
Few minor comments and nit picks here and there, rest all looks perfect.

svenklemm (approved):

src/nodes/chunk_dispatch/chunk_insert_state.c
```diff
@@ -734,6 +735,8 @@ ts_chunk_insert_state_destroy(ChunkInsertState *state)
 		Oid chunk_relid = RelationGetRelid(state->result_relation_info->ri_RelationDesc);
 		Chunk *chunk = ts_chunk_get_by_relid(chunk_relid, true);
 		ts_chunk_set_partial(chunk);
+		/* changed chunk status, so invalidate any plans involving this chunk */
+		CacheInvalidateRelcacheByRelid(chunk_relid);
```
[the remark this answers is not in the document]
konskov: Consider the example in line 183 of recompress_chunk_segmentwise.sql where we use a prepared statement. Without invalidating the plan, we miss the newly inserted data when doing `select * from mytab_prep;` so the invalidation is required to return correct results with prepared statements.

tsl/src/compression/api.c
```diff
@@ -23,6 +23,9 @@
 #include <utils/elog.h>
 #include <utils/fmgrprotos.h>
 #include <libpq-fe.h>
+#include <utils/tuplestore.h>
```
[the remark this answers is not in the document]
konskov: no, this was a leftover from debugging and thank you for catching this.

tsl/src/compression/compression.h
konskov: <access/heapam> is needed for BulkInsertState bistate in RowCompressor

src/nodes/chunk_dispatch/chunk_insert_state.c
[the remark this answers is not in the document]
sb230132: Ok, thanks for the explanation.

tsl/src/compression/compression.h
sb230132: Ok.

tsl/src/compression/api.c
```diff
@@ -178,6 +182,86 @@ compression_chunk_size_catalog_update_merged(int32 chunk_id, const RelationSize
 	return updated;
 }
 
+static int
```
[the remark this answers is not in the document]
konskov: Added

tsl/src/compression/api.c
```diff
@@ -178,6 +182,86 @@ compression_chunk_size_catalog_update_merged(int32 chunk_id, const RelationSize
...
+		bool replIsnull[Natts_compression_chunk_size] = { false };
+		bool repl[Natts_compression_chunk_size] = { false };
+		bool should_free;
+		TupleInfo *ti = ts_scan_iterator_tuple_info(&iterator);
+		HeapTuple tuple = ts_scanner_fetch_heap_tuple(ti, false, &should_free);
+		HeapTuple new_tuple;
+		heap_deform_tuple(tuple, ts_scanner_get_tupledesc(ti), values, replIsnull);
+
+		/* Only update the information pertaining to the compressed chunk */
+		values[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_heap_size)] =
```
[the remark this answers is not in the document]
konskov: removed the redundant lines

tsl/src/compression/api.c
```diff
@@ -178,6 +182,86 @@ compression_chunk_size_catalog_update_merged(int32 chunk_id, const RelationSize
...
+			values[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_heap_size)];
+		repl[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_heap_size)] = false;
+
+		values[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_toast_size)] =
+			values[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_toast_size)];
+		repl[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_toast_size)] = false;
+
+		values[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_index_size)] =
+			values[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_index_size)];
+		repl[AttrNumberGetAttrOffset(Anum_compression_chunk_size_uncompressed_index_size)] = false;
```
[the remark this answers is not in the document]
konskov: removed

tsl/src/compression/compression.h
konskov: sorry I was a bit confused by this comment, (`we reset sequence number only for old recompress`). If I understand you correctly you mean that this `recompressing` only applies to the case where we are in fact *segmentwise* recompressing? If so, yes you are correct about that, and it makes it clearer if this is renamed as you suggested. I will rename this.

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+												  n_keys,
+												  sort_keys,
+												  sort_operators,
+												  sort_collations,
+												  nulls_first,
+												  maintenance_work_mem,
+												  NULL,
+												  false);
+
+	fetch_unmatched_uncompressed_chunk_into_tuplestore(segment_tuplesortstate,
```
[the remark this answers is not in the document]
konskov: done

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+{
+	TableScanDesc heapScan;
+	HeapTuple uncompressed_tuple;
+	TupleDesc uncompressed_rel_tupdesc = RelationGetDescr(uncompressed_chunk_rel);
+
+	heapScan = table_beginscan(uncompressed_chunk_rel, GetLatestSnapshot(), 0, NULL);
+	TupleTableSlot *heap_tuple_slot =
+		MakeTupleTableSlot(uncompressed_rel_tupdesc, &TTSOpsHeapTuple);
+
+	for (uncompressed_tuple = heap_getnext(heapScan, ForwardScanDirection);
```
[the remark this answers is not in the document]
konskov: done. agree looks better this way :)

tsl/src/compression/api.c
```diff
@@ -892,3 +976,655 @@ tsl_recompress_chunk_wrapper(Chunk *uncompressed_chunk)
...
+}
+
+static void
+decompress_segment_update_current_segment(CompressedSegmentInfo **current_segment,
+										  TupleTableSlot *slot, PerCompressedColumn *per_cols)
+{
+	Datum val;
+	bool is_null;
+	int seg_idx = 0;
+	for (int i = 0; i < slot->tts_tupleDescriptor->natts; i++)
```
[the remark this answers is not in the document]
konskov: Thank you once again for the careful review. This comment made me realize another thing as well, I might not be handling scenarios with add/drop column correctly, so I will add a test for that too as part of addressing this.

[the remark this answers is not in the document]
konskov: I updated this, only the segmentby columns are of interest here, so I use an array with their offsets instead. Does it look ok?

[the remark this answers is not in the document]
sb230132: Yes, thanks for handling this.

sb230132 (approved):
Nice Work !!

PR #5339, opened 2023-02-16
Support UPDATE/DELETE on compressed hypertables.

sb230132:
This patch does following:

1. Executor changes to parse qual ExprState to check if SEGMENTBY
   column is specified in WHERE clause.
2. Based on step 1, we build scan keys.  
3. Executor changes to do heapscan on compressed chunk based on
   scan keys and move only those rows which match the WHERE clause
   to staging are aka uncompressed chunk.
4. Mark affected chunk as partially compressed.
5. Perform regular UPDATE/DELETE operations on staging area aka
   uncompressed chunk.
6. Since there is no Custom Scan (HypertableModify) node for
   UPDATE/DELETE operations on PG versions < 14, we don't support this
   feature on PG12 and PG13.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ accidentally triggering the load of a previous DB version.**
 * #5454 Add support for ON CONFLICT DO UPDATE for compressed hypertables
 * #5344 Enable JOINS for Hierarchical Continuous Aggregates
 * #5417 Refactor and optimize distributed COPY
+* #5339 Support UPDATE/DELETE on compressed hypertables
 
 **Bugfixes**
 * #5396 Fix SEGMENTBY columns predicates to be pushed down
--- a/src/cross_module_fn.h
+++ b/src/cross_module_fn.h
@@ -139,6 +139,7 @@ typedef struct CrossModuleFunctions
 	PGFunction decompress_chunk;
 	void (*decompress_batches_for_insert)(ChunkInsertState *state, Chunk *chunk,
 										  TupleTableSlot *slot);
+	void (*decompress_batches_for_update_delete)(List *chunks, List *predicates);
 	/* The compression functions below are not installed in SQL as part of create extension;
 	 *  They are installed and tested during testing scripts. They are exposed in cross-module
 	 *  functions because they may be very useful for debugging customer problems if the sql
--- a/src/nodes/hypertable_modify.c
+++ b/src/nodes/hypertable_modify.c
@@ -33,7 +33,7 @@
 #if PG14_GE
 static void fireASTriggers(ModifyTableState *node);
 static void fireBSTriggers(ModifyTableState *node);
-static TupleTableSlot *ExecModifyTable(PlanState *pstate);
+static TupleTableSlot *ExecModifyTable(CustomScanState *cs_node, PlanState *pstate);
 static TupleTableSlot *ExecProcessReturning(ResultRelInfo *resultRelInfo, TupleTableSlot *tupleSlot,
 											TupleTableSlot *planSlot);
 static void ExecInitInsertProjection(ModifyTableState *mtstate, ResultRelInfo *resultRelInfo);
@@ -92,6 +92,53 @@ get_chunk_dispatch_states(PlanState *substate)
 
 	return NIL;
 }
+
+#if PG14_GE
+typedef struct ChunkScanNodes
+{
+	/* list of compressed chunks */
+	List *chunks;
+	/* list of conditions specified in WHERE */
+	List *predicates;
+} ChunkScanNodes;
+/*
+ * Traverse the plan tree to look for Scan nodes on uncompressed chunks.
+ * Once Scan node is found check if chunk is compressed, if so then save
+ * the chunk in HypertableModifyState to be used during plan execution.
+ * We also save the WHERE quals to get information about predicates.
+ */
+static bool
+collect_chunks_from_scan(PlanState *ps, ChunkScanNodes *sn)
+{
+	RangeTblEntry *rte = NULL;
+	Chunk *current_chunk;
+	if (ps == NULL || ts_guc_enable_transparent_decompression == false)
+		return false;
+
+	switch (nodeTag(ps))
+	{
+		case T_SeqScanState:
+		case T_SampleScanState:
+		case T_IndexScanState:
+		case T_IndexOnlyScanState:
+		case T_BitmapHeapScanState:
+		case T_TidScanState:
+		case T_TidRangeScanState:
+			rte = rt_fetch(((Scan *) ps->plan)->scanrelid, ps->state->es_range_table);
+			current_chunk = ts_chunk_get_by_relid(rte->relid, false);
+			if (current_chunk && ts_chunk_is_compressed(current_chunk))
+			{
+				sn->chunks = lappend(sn->chunks, current_chunk);
+				if (ps->plan->qual && !sn->predicates)
+					sn->predicates = ps->plan->qual;
+			}
+			break;
+		default:
+			break;
+	}
+	return planstate_tree_walker(ps, collect_chunks_from_scan, sn);
+}
+#endif
 
 /*
  * HypertableInsert (with corresponding executor node) is a plan node that
@@ -170,7 +217,7 @@ hypertable_modify_exec(CustomScanState *node)
 	return ExecProcNode(linitial(node->custom_ps));
 #else
 	ModifyTableState *mtstate = linitial_node(ModifyTableState, node->custom_ps);
-	return ExecModifyTable(&mtstate->ps);
+	return ExecModifyTable(node, &mtstate->ps);
 #endif
 }
 
@@ -675,8 +722,9 @@ internalGetUpdateNewTuple(ResultRelInfo *relinfo, TupleTableSlot *planSlot, Tupl
  * modified version of ExecModifyTable from executor/nodeModifyTable.c
  */
 static TupleTableSlot *
-ExecModifyTable(PlanState *pstate)
+ExecModifyTable(CustomScanState *cs_node, PlanState *pstate)
 {
+	HypertableModifyState *ht_state = (HypertableModifyState *) cs_node;
 	ModifyTableState *node = castNode(ModifyTableState, pstate);
 	ModifyTableContext context;
 	EState *estate = node->ps.state;
@@ -745,7 +793,31 @@ ExecModifyTable(PlanState *pstate)
 	context.mtstate = node;
 	context.epqstate = &node->mt_epqstate;
 	context.estate = estate;
-
+	/*
+	 * For UPDATE/DELETE on compressed hypertable, decompress chunks and
+	 * move rows to uncompressed chunks.
+	 */
+	if ((operation == CMD_DELETE || operation == CMD_UPDATE) && !ht_state->comp_chunks_processed)
+	{
+		ChunkScanNodes *sn = palloc0(sizeof(ChunkScanNodes));
+		collect_chunks_from_scan(pstate, sn);
+		if (sn->chunks && ts_cm_functions->decompress_batches_for_update_delete)
+		{
+			ts_cm_functions->decompress_batches_for_update_delete(sn->chunks, sn->predicates);
+			ht_state->comp_chunks_processed = true;
+			/*
+			 * save snapshot set during ExecutorStart(), since this is the same
+			 * snapshot used to SeqScan of uncompressed chunks
+			 */
+			ht_state->snapshot = estate->es_snapshot;
+			/* use current transaction snapshot */
+			estate->es_snapshot = GetTransactionSnapshot();
+			CommandCounterIncrement();
+			/* mark rows visible */
+			estate->es_output_cid = GetCurrentCommandId(true);
+		}
+		pfree(sn);
+	}
 	/*
 	 * Fetch rows from subplan, and execute the required table modification
 	 * for each row.
@@ -969,6 +1041,12 @@ ExecModifyTable(PlanState *pstate)
 	 * Insert remaining tuples for batch insert.
 	 */
 	relinfos = estate->es_opened_result_relations;
+
+	if (ht_state->comp_chunks_processed)
+	{
+		estate->es_snapshot = ht_state->snapshot;
+		ht_state->comp_chunks_processed = false;
+	}
 
 	foreach (lc, relinfos)
 	{
--- a/src/nodes/hypertable_modify.h
+++ b/src/nodes/hypertable_modify.h
@@ -27,6 +27,8 @@ typedef struct HypertableModifyState
 	CustomScanState cscan_state;
 	ModifyTable *mt;
 	List *serveroids;
+	bool comp_chunks_processed;
+	Snapshot snapshot;
 	FdwRoutine *fdwroutine;
 } HypertableModifyState;
 
--- a/tsl/src/compression/compression.c
+++ b/tsl/src/compression/compression.c
@@ -1994,3 +1994,330 @@ decompress_batches_for_insert(ChunkInsertState *cis, Chunk *chunk, TupleTableSlo
 
 	table_close(in_rel, NoLock);
 }
+
+#if PG14_GE
+/*
+ * Helper method which returns true if given column_name
+ * is configured as SEGMENT BY column in a compressed hypertable
+ */
+static bool
+is_segmentby_col(List *ht_compression_info, char *column_name)
+{
+	ListCell *lc;
+	foreach (lc, ht_compression_info)
+	{
+		FormData_hypertable_compression *fd = lfirst(lc);
+		if (namestrcmp(&fd->attname, column_name) == 0)
+		{
+			if (fd->segmentby_column_index > 0)
+				return true;
+			break;
+		}
+	}
+	return false;
+}
+
+/*
+ * This method will evaluate the predicates, extract
+ * left and right operands, check if one of the operands is
+ * a simple Var type. If its Var type extract its corresponding
+ * column name from hypertable_compression catalog table.
+ * If extracted column is a SEGMENT BY column then save column
+ * name, value specified in the predicate. This information will
+ * be used to build scan keys later.
+ */
+static void
+fill_predicate_context(Chunk *ch, List *predicates, List **segmentby_columns,
+					   List **segmentby_columns_value, List **is_null_check, List **is_null)
+{
+	List *ht_compression_info = ts_hypertable_compression_get(ch->fd.hypertable_id);
+	ListCell *lc;
+	foreach (lc, predicates)
+	{
+		Node *node = lfirst(lc);
+		if (node == NULL)
+			continue;
+
+		Var *var;
+		char *column_name;
+		switch (nodeTag(node))
+		{
+			case T_OpExpr:
+			{
+				OpExpr *opexpr = (OpExpr *) node;
+				Expr *leftop, *rightop;
+				Const *arg_value;
+
+				leftop = linitial(opexpr->args);
+				rightop = lsecond(opexpr->args);
+
+				if (IsA(leftop, RelabelType))
+					leftop = ((RelabelType *) leftop)->arg;
+				if (IsA(rightop, RelabelType))
+					rightop = ((RelabelType *) rightop)->arg;
+
+				if (IsA(leftop, Var) && IsA(rightop, Const))
+				{
+					var = (Var *) leftop;
+					arg_value = (Const *) rightop;
+				}
+				else if (IsA(rightop, Var) && IsA(leftop, Const))
+				{
+					var = (Var *) rightop;
+					arg_value = (Const *) leftop;
+				}
+				else
+					continue;
+
+				column_name = get_attname(ch->table_id, var->varattno, false);
+				if (is_segmentby_col(ht_compression_info, column_name))
+				{
+					TypeCacheEntry *tce = lookup_type_cache(var->vartype, TYPECACHE_BTREE_OPFAMILY);
+					int op_strategy = get_op_opfamily_strategy(opexpr->opno, tce->btree_opf);
+					if (op_strategy == BTEqualStrategyNumber)
+					{
+						/* save segment by column name and its corresponding value specified in
+						 * WHERE */
+						*segmentby_columns = lappend(*segmentby_columns, column_name);
+						*segmentby_columns_value = lappend(*segmentby_columns_value, arg_value);
+						/* this constraint is not IS [NOT] NULL, so mark is_null_check as 0 */
+						*is_null_check = lappend_int(*is_null_check, 0);
+					}
+				}
+			}
+			break;
+			case T_NullTest:
+			{
+				NullTest *ntest = (NullTest *) node;
+				if (IsA(ntest->arg, Var))
+				{
+					var = (Var *) ntest->arg;
+					column_name = get_attname(ch->table_id, var->varattno, false);
+					if (is_segmentby_col(ht_compression_info, column_name))
+					{
+						*segmentby_columns = lappend(*segmentby_columns, column_name);
+						*is_null_check = lappend_int(*is_null_check, 1);
+						if (ntest->nulltesttype == IS_NULL)
+							*is_null = lappend_int(*is_null, 1);
+						else
+							*is_null = lappend_int(*is_null, 0);
+					}
+				}
+			}
+			break;
+			default:
+				break;
+		}
+	}
+}
+
+/*
+ * This method will build scan keys for predicates including
+ * SEGMENT BY column with attribute number from compressed chunk
+ * if condition is like <segmentbycol> = <const value>, else
+ * OUT param null_columns is saved with column attribute number.
+ */
+static ScanKeyData *
+build_update_delete_scankeys(Chunk *chunk, List *segmentby_columns, List *segmentby_column_values,
+							 List *is_null_check, int *num_scankeys, Bitmapset **null_columns)
+{
+	Chunk *comp_chunk = NULL;
+	Relation comp_chunk_rel;
+	TupleDesc tupleDesc;
+	ListCell *col;
+	ListCell *is_null_or_not_null;
+	int key_index = 0;
+	int pos = 0;
+
+	ScanKeyData *scankeys = palloc0(segmentby_columns->length * sizeof(ScanKeyData));
+	comp_chunk = ts_chunk_get_by_id(chunk->fd.compressed_chunk_id, true);
+	comp_chunk_rel = table_open(comp_chunk->table_id, AccessShareLock);
+	tupleDesc = RelationGetDescr(comp_chunk_rel);
+
+	forboth (col, segmentby_columns, is_null_or_not_null, is_null_check)
+	{
+		char *column_name = lfirst(col);
+		AttrNumber attnum = get_attnum(RelationGetRelid(comp_chunk_rel), column_name);
+		if (attnum == InvalidAttrNumber)
+			ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_COLUMN),
+					 errmsg("column \"%s\" of relation \"%s\" does not exist",
+							column_name,
+							RelationGetRelationName(comp_chunk_rel))));
+		Form_pg_attribute attr = TupleDescAttr(tupleDesc, AttrNumberGetAttrOffset(attnum));
+		TypeCacheEntry *tce = lookup_type_cache(attr->atttypid, TYPECACHE_EQ_OPR_FINFO);
+		if (!OidIsValid(tce->eq_opr_finfo.fn_oid))
+			elog(ERROR, "no equality function for column \"%s\"", column_name);
+
+		if (lfirst_int(is_null_or_not_null) == 0)
+		{
+			Const *const_value = list_nth(segmentby_column_values, pos++);
+			Datum value = (const_value ? const_value->constvalue : 0);
+			ScanKeyEntryInitialize(&scankeys[key_index++],
+								   0, /* flags */
+								   attnum,
+								   BTEqualStrategyNumber,
+								   InvalidOid, /* No strategy subtype. */
+								   attr->attcollation,
+								   tce->eq_opr_finfo.fn_oid,
+								   value);
+		}
+		else
+		{
+			*null_columns = bms_add_member(*null_columns, attnum);
+		}
+	}
+	*num_scankeys = key_index;
+	table_close(comp_chunk_rel, AccessShareLock);
+	return scankeys;
+}
+
+/*
+ * This method will:
+ *  1.scan compressed chunk
+ *  2.decompress the row
+ *  3.insert decompressed rows to uncompressed chunk
+ *  4.delete this row from compressed chunk
+ */
+static bool
+decompress_batches(Chunk *ch, ScanKeyData *scankeys, int num_scankeys, Bitmapset *null_columns,
+				   List *is_nulls, bool *chunk_status_changed)
+{
+	Relation chunk_rel;
+	Relation comp_chunk_rel;
+	Chunk *comp_chunk;
+	HeapTuple compressed_tuple;
+	RowDecompressor decompressor;
+	Snapshot snapshot;
+
+	snapshot = GetTransactionSnapshot();
+	chunk_rel = table_open(ch->table_id, RowExclusiveLock);
+	comp_chunk = ts_chunk_get_by_id(ch->fd.compressed_chunk_id, true);
+	comp_chunk_rel = table_open(comp_chunk->table_id, RowExclusiveLock);
+	decompressor = build_decompressor(comp_chunk_rel, chunk_rel);
+
+	TableScanDesc heapScan = table_beginscan(comp_chunk_rel, snapshot, num_scankeys, scankeys);
+	while ((compressed_tuple = heap_getnext(heapScan, ForwardScanDirection)) != NULL)
+	{
+		bool skip_tuple = false;
+		int attrno = bms_next_member(null_columns, -1);
+		int pos = 0;
+		bool is_null_condition = 0;
+		bool seg_col_is_null = false;
+		for (; attrno >= 0; attrno = bms_next_member(null_columns, attrno))
+		{
+			is_null_condition = list_nth_int(is_nulls, pos);
+			seg_col_is_null = heap_attisnull(compressed_tuple, attrno, decompressor.in_desc);
+			if ((seg_col_is_null && !is_null_condition) || (!seg_col_is_null && is_null_condition))
+			{
+				/*
+				 * if segment by column in the scanned tuple has non null value
+				 * and IS NULL is specified, OR segment by column has null value
+				 * and IS NOT NULL is specified then skip this tuple
+				 */
+				skip_tuple = true;
+				break;
+			}
+			pos++;
+		}
+		if (skip_tuple)
+			continue;
+		heap_deform_tuple(compressed_tuple,
+						  decompressor.in_desc,
+						  decompressor.compressed_datums,
+						  decompressor.compressed_is_nulls);
+
+		row_decompressor_decompress_row(&decompressor, NULL);
+		TM_FailureData tmfd;
+		TM_Result result;
+		result = table_tuple_delete(comp_chunk_rel,
+									&compressed_tuple->t_self,
+									decompressor.mycid,
+									snapshot,
+									InvalidSnapshot,
+									true,
+									&tmfd,
+									false);
+
+		if (result == TM_Updated || result == TM_Deleted)
+		{
+			if (IsolationUsesXactSnapshot())
+				ereport(ERROR,
+						(errcode(ERRCODE_T_R_SERIALIZATION_FAILURE),
+						 errmsg("could not serialize access due to concurrent update")));
+			return false;
+		}
+		if (result == TM_Invisible)
+		{
+			elog(ERROR, "attempted to lock invisible tuple");
+			return false;
+		}
+		*chunk_status_changed = true;
+	}
+	if (scankeys)
+		pfree(scankeys);
+	table_endscan(heapScan);
+	ts_catalog_close_indexes(decompressor.indexstate);
+	FreeBulkInsertState(decompressor.bistate);
+
+	table_close(chunk_rel, NoLock);
+	table_close(comp_chunk_rel, NoLock);
+	return true;
+}
+
+/*
+ * This method will:
+ *  1. Evaluate WHERE clauses and check if SEGMENT BY columns
+ *     are specified or not.
+ *  2. Build scan keys for SEGMENT BY columns.
+ *  3. Move scanned rows to staging area.
+ *  4. Update catalog table to change status of moved chunk.
+ */
+void
+decompress_batches_for_update_delete(List *chunks, List *predicates)
+{
+	List *segmentby_columns = NIL;
+	List *segmentby_columns_value = NIL;
+	List *is_null_check = NIL;
+	List *is_null = NIL;
+	ListCell *ch = NULL;
+	if (predicates)
+		fill_predicate_context(linitial(chunks),
+							   predicates,
+							   &segmentby_columns,
+							   &segmentby_columns_value,
+							   &is_null_check,
+							   &is_null);
+	foreach (ch, chunks)
+	{
+		bool chunk_status_changed = false;
+		ScanKeyData *scankeys = NULL;
+		Bitmapset *null_columns = NULL;
+		int num_scankeys = 0;
+
+		if (segmentby_columns)
+		{
+			scankeys = build_update_delete_scankeys(lfirst(ch),
+													segmentby_columns,
+													segmentby_columns_value,
+													is_null_check,
+													&num_scankeys,
+													&null_columns);
+		}
+		if (decompress_batches(lfirst(ch),
+							   scankeys,
+							   num_scankeys,
+							   null_columns,
+							   is_null,
+							   &chunk_status_changed))
+		{
+			/*
+			 * tuples from compressed chunk has been decompressed and moved
+			 * to staging area, thus mark this chunk as partially compressed
+			 */
+			if (chunk_status_changed == true)
+				ts_chunk_set_partial(lfirst(ch));
+		}
+	}
+}
+#endif
--- a/tsl/test/expected/continuous_aggs-12.out
+++ b/tsl/test/expected/continuous_aggs-12.out
@@ -1552,13 +1552,11 @@ SELECT * from search_query_count_3 ORDER BY 1, 2, 3;
 -- more data
 -- ).
 insert into raw_data select '2000-05-01 00:00+0','Q3', 0, 0;
---this one fails now
-\set ON_ERROR_STOP 0
+
 CALL refresh_continuous_aggregate('search_query_count_3', NULL, '2000-06-01 00:00+0'::timestamptz);
-ERROR:  cannot update/delete rows from chunk "_hyper_41_79_chunk" as it is compressed
 CALL refresh_continuous_aggregate('search_query_count_3', '2000-05-01 00:00+0'::timestamptz, '2000-06-01 00:00+0'::timestamptz);
-ERROR:  cannot update/delete rows from chunk "_hyper_41_79_chunk" as it is compressed
-\set ON_ERROR_STOP 1
+NOTICE:  continuous aggregate "search_query_count_3" is already up-to-date
+
 --insert row
 insert into raw_data select '2001-05-10 00:00+0','Q3', 100, 100;
 --this should succeed since it does not refresh any compressed regions in the cagg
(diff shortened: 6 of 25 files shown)


svenklemm (changes requested):
Seems like you are completely missing the batch filtering. This needs way more tests especially isolation tests. Not sure why we need the modified DecompressChunkDML node since the tuples move into uncompressed chunk this should not be needed. Did you have a look at how this is handled in the INSERT path? Why did you choose a different approach here?

sb230132:
> Seems like you are completely missing the batch filtering. This needs way more tests especially isolation tests. Not sure why we need the modified DecompressChunkDML node since the tuples move into uncompressed chunk this should not be needed. Did you have a look at how this is handled in the INSERT path? Why did you choose a different approach here?

Fixed as per your suggestion.

tsl/test/sql/CMakeLists.txt
```diff
@@ -104,7 +101,11 @@ endif(CMAKE_BUILD_TYPE MATCHES Debug)
 if((${PG_VERSION_MAJOR} GREATER_EQUAL "14"))
   if(CMAKE_BUILD_TYPE MATCHES Debug)
     list(APPEND TEST_FILES chunk_utils_internal.sql)
+    list(APPEND TEST_TEMPLATES continuous_aggs.sql.in
+         continuous_aggs_deprecated.sql)
   endif()
+  list(APPEND TEST_FILES compression.sql compression_upd_del.sql
```
konskov, line 107: shouldn't these tests be executed also for pg versions < 14?

konskov, line 107: nit: suggest to rename the new test file to `compression_update_delete`

svenklemm (changes requested):
This needs to work with any scan not just SeqScan. I dont think your approach will work correctly with execution time chunk exclusion.

konskov:
would it be possible to add a test with NULL values in segmentby columns?

tsl/test/sql/CMakeLists.txt
sb230132: We don't support UPDATE/DELETE operation on PG version < 14, this is because planner does not generate CustomScan (HypertableModify) node on these versions.

These tests have several queries related to UPDATE/DELETE operations on compressed hypertables, if we want to run them we need different result (.out) file for version < 14
Instead i decided to run them only on version >= 14

sb230132: Done. Thanks.

sb230132:
> This needs to work with any scan not just SeqScan. I dont think your approach will work correctly with execution time chunk exclusion.

I tried several different testcases to check if there are other than SeqScan nodes being generated on compressed hypertables, and i dont see planner generated any other scan nodes.

However for same query if i decompress all chunks i see that Index Scan/ BitMap Index scan nodes are getting created.

tsl/src/compression/compression.c
```diff
@@ -2215,3 +2215,129 @@ decompress_batches_for_insert(ChunkInsertState *cis, Chunk *chunk, TupleTableSlo
...
+	{
+		scankeys = build_scankeys_for_update_delete(ch->fd.hypertable_id,
+													decompressor,
+													attname,
+													attval,
+													strategy);
+		num_scankeys = attname->length;
+	}
+
+	TableScanDesc heapScan = table_beginscan(comp_chunk_rel, snapshot, num_scankeys, scankeys);
```
konskov, line 2309: because, as Sven had pointed out, SK_SEARCHNULL and SK_SEARCHNOTNULL are not supported by heapscan, I think some logic has to be added for handling NULL values in segmentby columns. For example: 
```
create table exam (time timestamptz, a int, b int);
 select create_hypertable('exam', 'time');

insert INTO exam values (now(), 1, 1);
insert INTO exam values (now(), 2, 2);

insert INTO exam values (now(), NULL, 2);
insert INTO exam values (now(), NULL, 2);

ALTER TABLE exam set (timescaledb.compress, timescaledb.compress_segmentby='a');

select compress_chunk(show_chunks('exam'));

update exam set b = 4 where a = NULL;
```
In this example currently the rows where a = NULL will not get updated.

konskov: oh wait that was just stupid of me, sorry for the confusion... 😅 this should be `a IS NULL` and that does update the respective rows, so you can ignore this!

svenklemm (changes requested):
The way you currently handle constraints does not work correctly you dont consider the used operators at all. I think you should refactor passing of those attname lists and values lists this is a very unusual pattern.

E.g. the following query will not work: `DELETE FROM metrics WHERE device <= 'd1';`

sb230132:
> The way you currently handle constraints does not work correctly you dont consider the used operators at all. I think you should refactor passing of those attname lists and values lists this is a very unusual pattern.
> 
> E.g. the following query will not work: `DELETE FROM metrics WHERE device <= 'd1';` NULL values are skipped in scankeys but not handled in the iterator leading to unnecessary decompressions.

Ok, i took a different approach in handing WHERE predicates.

1. I parse WHERE to see if it has SEGMENTBY columns. If so i save these RestrictInfos to be used during heap scan.
2. I dont build scan keys. Instead i fetch a compressed row, and call ExecQual() to filter out the rows.

Added more tests to check for all kinds of predicates like:
=, <, <=, >, >=, IS NULL, IS NOT NULL.

svenklemm:
The constraint handling is still not optimal. why are we not generating scankeys as we do in other places? To enable future improvements in constraint handling we need to map chunk constraints from the uncompressed chunk to the compressed chunk similar to how DecompressChunk does it. This is required to add support for filtering by orderby columns down the line.

Example of inconsistency in DELETE:
```
postgres@updelete[4089760]=# BEGIN; delete from metrics where device IN ('d3','d2'); select * from metrics;EXPLAIN SELECT * FROM metrics; rollback;
BEGIN
Time: 1.755 ms
DELETE 1
Time: 3.567 ms
          time          | device | value
------------------------+--------+-------
 2000-01-01 00:00:00+01 | d1     |   0.1
 2000-01-01 00:00:00+01 | d2     |   0.1
(2 rows)

Time: 1.151 ms
```

The query after the delete should not include the `d2` tuple.

sb230132:
> The constraint handling is still not optimal. why are we not generating scankeys as we do in other places? To enable future improvements in constraint handling we need to map chunk constraints from the uncompressed chunk to the compressed chunk similar to how DecompressChunk does it. This is required to add support for filtering by orderby columns down the line.
> 
> Example of inconsistency in DELETE:
> 
> ```
> postgres@updelete[4089760]=# BEGIN; delete from metrics where device IN ('d3','d2'); select * from metrics; rollback;
> BEGIN
> Time: 1.755 ms
> DELETE 1
> Time: 3.567 ms
>           time          | device | value
> ------------------------+--------+-------
>  2000-01-01 00:00:00+01 | d1     |   0.1
>  2000-01-01 00:00:00+01 | d2     |   0.1
> (2 rows)
> 
> Time: 1.151 ms
> ```
> 
> The query after the delete should not include the `d2` tuple.

Regarding attribute mapping, i use tupledescriptor of compressed chunk to get correct index of SEGMENT BY columns.
Fixed row visibility issue.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	scankey = palloc0(sizeof(ScanKeyData));
+	value = (const_value ? const_value->constvalue : 0);
+	/*
+	 * make compiler happy by silencing this error
+	 * ‘attr’ may be used uninitialized in this function
+	 */
+	attr = TupleDescAttr(tupleDesc, 0);
+	for (attno = 1; attno <= tupleDesc->natts; attno++)
+	{
+		attr = TupleDescAttr(tupleDesc, attno - 1);
```
antekresic, line 213:
```suggestion
		attr = TupleDescAttr(tupleDesc, AttrNumberGetAttrOffset(attno));
```

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	 * physical layouts of compressed and uncompressed chunks. Thus we
+	 * need to retreive correct attrno from compressed chunk tuple descriptor.
+	 */
+	curr_chunk = linitial(context->chunks);
+	comp_chunk = ts_chunk_get_by_id(curr_chunk->fd.compressed_chunk_id, true);
+	comp_chunk_rel = relation_open(comp_chunk->table_id, AccessShareLock);
+	tupleDesc = RelationGetDescr(comp_chunk_rel);
+	for (attno = 1; attno <= tupleDesc->natts; attno++)
+	{
+		Form_pg_attribute attribute = TupleDescAttr(tupleDesc, attno - 1);
```
antekresic, line 148:
```suggestion
		Form_pg_attribute attribute = TupleDescAttr(tupleDesc, AttrNumberGetAttrOffset(attno));
```

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	tce = lookup_type_cache(attr->atttypid, TYPECACHE_EQ_OPR_FINFO | TYPECACHE_CMP_PROC_FINFO);
+	ScanKeyEntryInitialize(scankey,
+						   0, /* flags */
+						   attno,
+						   strategy,
+						   InvalidOid, /* No strategy subtype. */
+						   attr->attcollation,
+						   (strategy == BTEqualStrategyNumber ? tce->eq_opr_finfo.fn_oid :
+																tce->cmp_proc_finfo.fn_oid),
+						   value);
```
antekresic, line 226: Minor nit: formatting looks off here.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+		case T_TidScanState:
+		case T_TidRangeScanState:
+			rte = rt_fetch(((Scan *) ps->plan)->scanrelid, ps->state->es_range_table);
+			chunk = ts_chunk_get_by_relid(rte->relid, false);
+			if (chunk && ts_chunk_is_compressed(chunk))
+			{
+				context->chunks = lappend(context->chunks, chunk);
+				if (ps->plan->qual && !context->chunk_quals_processed)
+				{
+					/* check if perdicate has SEGMENT BY columns, then build scan keys */
```
antekresic, line 355:
```suggestion
					/* check if predicate has SEGMENT BY columns, then build scan keys */
```

antekresic (commented):
Did a pass on the code part of PR, left a couple of minor comments. 

Formatting looks weird in some places.

I'll take another pass at the test part of the PR.

tsl/test/expected/continuous_aggs-12.out
```diff
@@ -1555,9 +1555,8 @@ insert into raw_data select '2000-05-01 00:00+0','Q3', 0, 0;
 --this one fails now
 \set ON_ERROR_STOP 0
 CALL refresh_continuous_aggregate('search_query_count_3', NULL, '2000-06-01 00:00+0'::timestamptz);
-ERROR:  cannot update/delete rows from chunk "_hyper_41_79_chunk" as it is compressed
 CALL refresh_continuous_aggregate('search_query_count_3', '2000-05-01 00:00+0'::timestamptz, '2000-06-01 00:00+0'::timestamptz);
-ERROR:  cannot update/delete rows from chunk "_hyper_41_79_chunk" as it is compressed
+NOTICE:  continuous aggregate "search_query_count_3" is already up-to-date
 \set ON_ERROR_STOP 1
```
akuzm, line 1560: It works now, so let's update the comment and remove ON_ERROR_STOP.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+					build_scankeys_for_segmentby_qual(ps, context);
+					context->chunk_quals_processed = true;
+				}
```
akuzm, line 358: So we only do this for the first chunk, right? I think scankeys have the attribute numbers inside, and they might not match for different chunks. We have some examples in `compression_alter.sql`, basically you add a chunk, compress it, drop a column from table, add another chunk, compress it. And there you have two compressed chunks with different sequences of columns.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+}
+
+/*
+ * Traverse the plan tree to look for Scan nodes on uncompressed chunks.
+ * Once Scan node is found check if chunk is compressed, if so then save
+ * the chunk in HypertableModifyState to be used during plan execution.
+ * We also parse the WHERE quals to get information about predicates.
+ */
+static bool
+collect_chunks_from_scan(PlanState *ps, HypertableModifyState *context)
```
akuzm, line 332: We do have the list of chunks for the update somewhere, that we evaluate at the planning time, right? Can we use it here? This approach with traversing the plan tree looks unreliable, imagine you have a join with a different hypertable, then you'll find unrelated chunks and everything will break. Something like `explain update ht_metrics_compressed set value = 1 from ht_metrics_compressed t1 join ht_metrics_compressed t2 on t1.value = t2.value + 1;`

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	TupleDesc tupleDesc;
+	int attno = 0;
+	bool is_null = false;
+
+	chunk_rel = relation_open(context->chunk_rte->relid, AccessShareLock);
+	tupleDesc = RelationGetDescr(chunk_rel);
+	/*
+	 * check if SEGMENT BY column is NULLABLE or not based on uncompressed
+	 * chunk tuple descriptor. This is because compressed chunk column
+	 * attributes doesnot have NULLABLE properties.
```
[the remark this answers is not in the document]
antekresic: Actually, this is incorrect. You can use a NULLABLE column in SEGMENT BY and it will be NULLABLE in the compressed chunk.

In that case, NullTest won't work correctly:
```
postgres=# SELECT count(*) FROM sample_table WHERE sensor_id IS NOT NULL;
 count 
-------
     7
(1 row)

postgres=# UPDATE sample_table SET temperature = 34.21 WHERE sensor_id IS NOT NULL;
UPDATE 0
postgres=# 


```

src/nodes/hypertable_modify.c
akuzm: By the way, a similar query is already broken on main: https://github.com/timescale/timescaledb/issues/5475

src/nodes/hypertable_modify.c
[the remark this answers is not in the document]
sb230132: I mean NULLABLE property in column definition. Please refer to Nullable column in below definitions of compressed and uncompressed chunk.
```
postgres=# \d _timescaledb_internal._hyper_7_17_chunk
   Table "_timescaledb_internal._hyper_7_17_chunk"
  Column   |  Type   | Collation | Nullable | Default
-----------+---------+-----------+----------+---------
 time      | integer |           | not null |
 device_id | integer |           |          |
 val       | integer |           |          |
Indexes:
    "_hyper_7_17_chunk_sample_table_time_idx" btree ("time" DESC)
Check constraints:
    "constraint_9" CHECK ("time" >= 0 AND "time" < 10)
Inherits: sample_table

postgres=# \d _timescaledb_internal.compress_hyper_8_19_chunk
                    Table "_timescaledb_internal.compress_hyper_8_19_chunk"
        Column         |                 Type                  | Collation | Nullable | Default
-----------------------+---------------------------------------+-----------+----------+---------
 time                  | _timescaledb_internal.compressed_data |           |          |
 device_id             | integer                               |           |          |
 val                   | _timescaledb_internal.compressed_data |           |          |
 _ts_meta_count        | integer                               |           |          |
 _ts_meta_sequence_num | integer                               |           |          |
 _ts_meta_min_1        | integer                               |           |          |
 _ts_meta_max_1        | integer                               |           |          |
Indexes:
    "compress_hyper_8_19_chunk__compressed_hypertable_8_device_id__t" btree (device_id, _ts_meta_sequence_num)
Inherits: _timescaledb_internal._compressed_hypertable_8
```

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
 	return NIL;
 }
 
+#if PG14_GE
+/*
+ * Helper method which returns true if SEGMENT BY column
+ * is NULLABLE. This method also returns the index of
+ * this NULLABLE SEGMENT BY column in compressed chunk.
```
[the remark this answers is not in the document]
sb230132: Function return value is bool which tells if SEGMENT BY column is defined as IS NOT OR NOT NULL.
3rd param is an OUT param which returns attribute number of the SEGMENT BY column present in compressed chunk.

src/nodes/hypertable_modify.c
sb230132: Fixed.

src/nodes/hypertable_modify.c
sb230132: Regarding formatting iam not sure, there is a github action about code formatting which will fail if formatting was wrong. So i will correct it if github action complains it.

tsl/test/expected/continuous_aggs-12.out
sb230132: Fixed.

src/nodes/hypertable_modify.c
```diff
@@ -745,7 +1018,54 @@ ExecModifyTable(PlanState *pstate)
...
+	if (operation == CMD_DELETE || operation == CMD_UPDATE)
+		collect_chunks_from_scan(pstate, ht_state);
+	if ((operation == CMD_DELETE || operation == CMD_UPDATE) && !ht_state->comp_chunks_processed)
+	{
+		ListCell *ch;
+		foreach (ch, ht_state->chunks)
+		{
+			Chunk *chunk = lfirst(ch);
+			bool chunk_status_changed = false;
+			if (ts_cm_functions->decompress_batches_for_update_delete)
```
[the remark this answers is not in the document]
sb230132: Ok, now i handle by reporting a more general way. Please suggest a better error message.
```
if (result != TM_Ok)
		{
			ereport(ERROR,
					 errmsg("Could not perform this operation. Please retry again.")));
			return false;
		}
```

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	}
+	return false;
+}
+
+/*
+ * Helper method which will build scan keys for SEGMENT BY column
+ * with index attritbute number from compressed chunk.
+ */
+static void
+build_scankeys(HypertableModifyState *context, char *column_name, Const *const_value, int strategy)
```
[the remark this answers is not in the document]
sb230132: Renamed it to build_update_delete_scankeys() as this function is called for both UPDATE and DELETE.

src/nodes/hypertable_modify.c
sb230132: I had the same doubt, then i did not have a valid testcase to prove it. Thanks for finding this issue. 
Now for each chunk i do get scankeys.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,276 @@ get_chunk_dispatch_states(PlanState *substate)
...
+				return false;
+
+			column_name = get_attname(context->chunk_rte->relid, var->varattno, false);
+			if (is_segmentby_col(ht_compression_info, column_name))
+			{
+				tce = lookup_type_cache(var->vartype, TYPECACHE_BTREE_OPFAMILY);
+				op_strategy = get_op_opfamily_strategy(opexpr->opno, tce->btree_opf);
+				if (op_strategy == BTLessStrategyNumber || op_strategy == BTEqualStrategyNumber ||
+					op_strategy == BTGreaterStrategyNumber)
+					build_scankeys(context, column_name, arg_value, op_strategy);
```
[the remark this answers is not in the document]
sb230132: This patch of mine follows the approach of using ExecQual() 
https://github.com/timescale/timescaledb/commit/d46eba22f59d8d948626bd780c6ab5750b11204f

tsl/test/sql/compression_update_delete.sql
```diff
@@ -0,0 +1,656 @@
...
+DELETE FROM sample_table WHERE c4 IS NULL;
+-- report 0 row
+SELECT * FROM sample_table WHERE c4 IS NULL;
+
+-- report 1 row which ensure that only required row is moved and deleted
+SELECT COUNT(*) FROM :COMPRESS_CHUNK_1;
+-- report 0 row
+SELECT COUNT(*) FROM :COMPRESS_CHUNK_1 WHERE c4 IS NULL;
+
+DROP TABLE sample_table;
```
akuzm, line 656: Do we want to support this kind of updates on a single chunk, in the fashion of "transparent decompression"? That is, if you write `update _timescaledb_internal.chunk_1 set x = y ...`, it should also decompress the relevant parts  and so on. We do support selects like this, so it might make sense.

src/nodes/hypertable_modify.c
sb230132: In short i dont see any issue with this, however iam not sure what can break this.

Long story:

Challenge is that i need to access Filter clauses which are present at Scan node.
As you say we end up saving same chunks in the list and also chunks which represent some other relation specified in the query.
After some testing i see following behaviour w.r.t to below plan:

```
postgres=# explain update sample_table set temperature = 1 from sample_table t1 join sample_table1 t2 on t1.temperature > t2.temperature + 1;
                                                     QUERY PLAN
--------------------------------------------------------------------------------------------------------------------
 Custom Scan (HypertableModify)  (cost=0.26..1246.08 rows=65333 width=38)
   ->  Update on sample_table  (cost=0.26..1246.08 rows=65333 width=38)
         Update on _hyper_9_10_chunk sample_table_1
         ->  Nested Loop  (cost=0.26..1246.08 rows=65333 width=38)
               ->  Nested Loop  (cost=0.26..428.35 rows=9333 width=20)
                     Join Filter: (t1.temperature > (t2.temperature + '1'::double precision))
                     ->  Seq Scan on _hyper_9_10_chunk t1  (cost=0.00..1.07 rows=7 width=18)
                     ->  Custom Scan (DecompressChunk) on _hyper_7_8_chunk t2  (cost=0.26..1.04 rows=4000 width=18)
                           ->  Seq Scan on compress_hyper_8_9_chunk  (cost=0.00..1.04 rows=4 width=136)
               ->  Materialize  (cost=0.00..1.10 rows=7 width=10)
                     ->  Seq Scan on _hyper_9_10_chunk sample_table_1  (cost=0.00..1.07 rows=7 width=10)
(11 rows)
```
In above example i end up having 2 entries in chunk List, which are duplicates. Note Seq Scan on compress_hyper_8_9_chunk will be skipped as its compression status will be set to 0.

So if a chunk on which update/delete is already done, then 2nd call to perform the same operation on same chunk will be a no-op as the qualifying rows have already been moved from compressed chunk to uncompressed chunk.

src/nodes/hypertable_modify.c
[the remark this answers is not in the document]
akuzm: I think we can learn from the normal `ExecUpdate`. The correct thing to do is pretty complicated, and depends on the transaction isolation level. Maybe for starters just this simple check is OK. The usual error is:
```
						ereport(ERROR,
								(errcode(ERRCODE_T_R_SERIALIZATION_FAILURE),
								 errmsg("could not serialize access due to concurrent update")));
```
Maybe we could also add an errhint that prints out the tuple values, if it's easy to do. Just to give the users more context.

Let's add at least a very basic test for this, the concurrent modifications are a frequent scenario.

src/nodes/hypertable_modify.c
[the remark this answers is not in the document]
sb230132: Sven's reply:

ExecQual will always happen after the scan and never be incorporated into the scan, for SeqScan this does not matter as it cannot incorporate scankeys but for index scans it does matter and using ExecQual is not desirable. We need to derive ScanKeys from the quals here as we improve the batch filtering to also work with orderby metadata (in a followup PR)

https://iobeam.slack.com/archives/C032AS21X4P/p1679570721314599?thread_ts=1679563611.989269&cid=C032AS21X4P

src/nodes/hypertable_modify.c
sb230132: Sven's reply:

ExecQual will always happen after the scan and never be incorporated into the scan, for SeqScan this does not matter as it cannot incorporate scankeys but for index scans it does matter and using ExecQual is not desirable. We need to derive ScanKeys from the quals here as we improve the batch filtering to also work with orderby metadata (in a followup PR)

https://iobeam.slack.com/archives/C032AS21X4P/p1679570721314599?thread_ts=1679563611.989269&cid=C032AS21X4P

src/nodes/hypertable_modify.c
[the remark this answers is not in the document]
sb230132: Sven's reply:

ExecQual will always happen after the scan and never be incorporated into the scan, for SeqScan this does not matter as it cannot incorporate scankeys but for index scans it does matter and using ExecQual is not desirable. We need to derive ScanKeys from the quals here as we improve the batch filtering to also work with orderby metadata (in a followup PR)

https://iobeam.slack.com/archives/C032AS21X4P/p1679570721314599?thread_ts=1679563611.989269&cid=C032AS21X4P

src/nodes/hypertable_modify.c
akuzm: Well, I would prefer planning time because it's a more "usual" way, and probably it's even simpler? Once for hypertable, you would need to find the conditions that you will use for choosing the segments to decompress. Then, split them into the ScanKey/ExecQual parts (this might be optional at this point, we can start with just ExecQual I guess). Then, for each chunk, you'll need to translate the variables using `ts_make_inh_translation_list()`.

If you feel that redoing this is too much work, we can go with traversing the planner tree at execution time, but please limit it to the immediate children of the ModifyTable node (the ones that show as "Update chunk1" in explain). But I have a hunch that we'll have to redo it in the future anyway.

svenklemm: The reason why this has to happen at execution time is to make this work with executor startup exclusion otherwise chunk exclusion will only work with planner chunk exclusion resulting in potentially a lot of unnecessary decompressions. Also I dont think limiting to immediate children is the correct approach here as this has the potential to miss chunks. If we are worried we trigger decompression on other hypertables due to joins we should instead check the chunk belongs to the hypertable we are updating.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,287 @@ get_chunk_dispatch_states(PlanState *substate)
...
+		if (attribute->attisdropped)
+			continue;
+		char *attributeName = NameStr(attribute->attname);
+		if (strcmp(attributeName, column_name) == 0)
+		{
+			*attrno = attno;
+			break;
+		}
+	}
+	table_close(comp_chunk_rel, AccessShareLock);
```
svenklemm, line 156: This function should be rewritten to use the postgres catalog cache instead, this would simplify the code a lot. Have a look at get_attnum for an example.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,287 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	Relation comp_chunk_rel;
+	TupleDesc tupleDesc;
+	int attno = 0;
+	Form_pg_attribute attr;
+	Datum value;
+	ScanKeyData *scankey = NULL;
+	TypeCacheEntry *tce;
+
+	comp_chunk = ts_chunk_get_by_id(curr_chunk->fd.compressed_chunk_id, true);
+	comp_chunk_rel = relation_open(comp_chunk->table_id, AccessShareLock);
```
svenklemm, line 199: relation_open/relation_close and table_open/table_close should be used together and not mixed

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +95,287 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	Chunk *comp_chunk = NULL;
+	Relation comp_chunk_rel;
+	TupleDesc tupleDesc;
+	int attno = 0;
+	Form_pg_attribute attr;
+	Datum value;
+	ScanKeyData *scankey = NULL;
+	TypeCacheEntry *tce;
+
+	comp_chunk = ts_chunk_get_by_id(curr_chunk->fd.compressed_chunk_id, true);
```
svenklemm, line 198: why are we refetching the chunk for every scankey we build? shouldn't this be handled in caller? The same applies to opening/closing the relation

svenklemm (changes requested):
I dont see how the code will correctly deal with multiple scankeys per chunk. I think that code could use some refactoring and have the ScanKey generation closer to the batch decompression.

src/nodes/hypertable_modify.c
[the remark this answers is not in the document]
sb230132: I added a testcase in compression_ddl_iso.spec, however since we support this feature only in PG version >= 14
i see testcase fail for PG12 and PG13.

Should i skip this tests for  PG12 and PG13 ?
Or
Please suggest the where can i add concurrent tests ?

src/nodes/hypertable_modify.c
sb230132: Thanks, fixed it.

src/nodes/hypertable_modify.c
sb230132: Fixed.

src/nodes/hypertable_modify.c
sb230132: Right, now i pass compressed_chunk as parameter to avoid this.

sb230132:
> I dont see how the code will correctly deal with multiple scankeys per chunk. I think that code could use some refactoring and have the ScanKey generation closer to the batch decompression.

As discussed on slack, i do handle multiple scan keys, however i was maintaining scankeys as List of List, now i took a different approach to solve this.

I needed to maintain a list of list for scankeys because i have no context to know which all scankeys belong to which chunk.
I need to maintain all context related to predicates for each chunk, so i came up with an approach of having a new struct
called 
```
typedef struct PredicatesContext
{
	/* hold scan keys for each predicate clause */
	List *scankeys;
	/* save attribute no: of the null column */
	List *null_columns;
	/* save context above IS NULL or IS NOT NULL is specified in predicate clause */
	List *is_null_checks;
} PredicatesContext;
```

This struct will hold predicates context needed by each chunk, in case WHERE clause with SEGMENT BY column is specified in the UPDATE/DELETE query.

Once the chunk is processed (decompressed and data is moved to staging area) i free up all space occupied by PredicatesContext, this way there will be no leaks.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +110,286 @@ get_chunk_dispatch_states(PlanState *substate)
...
+	 * ‘attr’ may be used uninitialized in this function
+	 */
+	attr = TupleDescAttr(tupleDesc, 0);
+	for (attno = 1; attno <= tupleDesc->natts; attno++)
+	{
+		attr = TupleDescAttr(tupleDesc, AttrNumberGetAttrOffset(attno));
+		if (namestrcmp(&(attr->attname), column_name) == 0)
+			break;
+	}
+	tce = lookup_type_cache(attr->atttypid, TYPECACHE_EQ_OPR_FINFO);
```
svenklemm, line 215: While lookup_type_cache will always return something, the specific information requested might still be InvalidOid so we nned to check for that

sb230132: Ok, Added this additional check:
```
	if (NULL == tce || !OidIsValid(tce->eq_opr_finfo.fn_oid))
		elog(ERROR, "no equality function for column \"%s\"", column_name);
```

svenklemm: lookup_type_cache will never return NULL

src/nodes/hypertable_modify.c
```diff
@@ -745,7 +1027,75 @@ ExecModifyTable(PlanState *pstate)
...
+		collect_chunks_from_scan(pstate, ht_state);
+		ListCell *ch;
+		int pos = 0;
+		foreach (ch, ht_state->chunks)
+		{
+			bool chunk_status_changed = false;
+			PredicatesContext *cx = NULL;
+			List *scankeys = NULL;
+			List *null_columns = NULL;
+			List *is_null_checks = NULL;
```
svenklemm, line 1045:
```suggestion
			List *scankeys = NIL;
			List *null_columns = NIL;
			List *is_null_checks = NIL;
```
The correct macro for empty list is NIL

src/nodes/hypertable_modify.c
```diff
@@ -745,7 +1027,75 @@ ExecModifyTable(PlanState *pstate)
...
+				{
+					/*
+					 * tuples from compressed chunk has been decompressed and moved
+					 * to staging area, thus mark this chunk as partially compressed
+					 */
+					if (chunk_status_changed == true)
+						ts_chunk_set_partial(lfirst(ch));
+					/* free all chunks context */
+					if (scankeys)
+						list_free(scankeys);
```
svenklemm, line 1071: list_free does not free the entries of the list, so we are leaking memory here

src/nodes/hypertable_modify.c
sb230132: Fixed.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +111,269 @@ get_chunk_dispatch_states(PlanState *substate)
...
+						   InvalidOid, /* No strategy subtype. */
+						   attr->attcollation,
+						   tce->eq_opr_finfo.fn_oid,
+						   value);
+	table_close(comp_chunk_rel, AccessShareLock);
+	return scankey;
+}
+
+/*
+ * This method will parse the expression (node), extract
```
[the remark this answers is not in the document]
sb230132: Replaced parse with evaluate and renamed function name to check_predicates()

src/nodes/hypertable_modify.c
sb230132: Fixed as:
```
					if (scankeys)
					{
						foreach (cf, scankeys)
							pfree((ScanKeyData *) lfirst(cf));
					}
```

src/nodes/hypertable_modify.c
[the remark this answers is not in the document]
akuzm: Let's add a new separate file for compressed update/delete isolation tests, and skip it for older versions.

src/nodes/hypertable_modify.c
```diff
@@ -745,7 +981,82 @@ ExecModifyTable(PlanState *pstate)
...
+					/* free all chunks context */
+					if (scankeys)
+					{
+						foreach (cf, scankeys)
+							pfree((ScanKeyData *) lfirst(cf));
+					}
+					if (null_columns)
+					{
+						foreach (cf, null_columns)
+							pfree((Bitmapset *) linitial(null_columns));
```
akuzm, line 1033:
```suggestion
							pfree(lfirst(cf));
```

Probably should be like this? Also no need to cast because pfree takes `void *`. The list probably also should be emptied so that it doesn't contain the invalid elements. Maybe this can be replaced entirely by `list_free_deep`.

src/nodes/hypertable_modify.c
```diff
@@ -745,7 +981,82 @@ ExecModifyTable(PlanState *pstate)
...
+		}
+		if (ht_state->comp_chunks_processed)
+		{
+			/*
+			 * save snapshot set during ExecutorStart(), since this is the same
+			 * snapshot used to SeqScan of uncompressed chunks
+			 */
+			ht_state->snapshot = estate->es_snapshot;
+			/* use current transaction snapshot */
+			estate->es_snapshot = GetTransactionSnapshot();
```
akuzm, line 1049: Why is this required, isn't `CommandCounterIncrement()` enough? I think Postgres almost never calls `GetTransactionSnapshot()` inside the executor code.

src/nodes/hypertable_modify.c
```diff
@@ -93,6 +110,224 @@ get_chunk_dispatch_states(PlanState *substate)
...
+ * left and right operands, check if one of the operands is
+ * a simple Var type. If its Var type extract its corresponding
+ * column name from hypertable_compression catalogue table.
+ * If extracted column is a SEGMENT BY column then build
+ * scan keys.
+ */
+static bool
+check_predicate(Node *node, Chunk *curr_chunk, Chunk *curr_comp_chunk, ScanKeyData **scankey,
+				Bitmapset **null_columns, int *is_null_checks)
+{
```
akuzm, line 175: Is it possible to do this once per hypertable? This looks costly, because we're redoing mostly the same work for every chunk, and there can easily be thousands of chunks. Can we compute the nullable columns and scankeys once for hypertable, and then for each chunk just translate the attnos?

src/nodes/hypertable_modify.c
```diff
@@ -56,6 +59,20 @@ static bool ExecOnConflictUpdate(ModifyTableContext *context, ResultRelInfo *res
 static void ExecCheckTupleVisible(EState *estate, Relation rel, TupleTableSlot *slot);
 static void ExecCheckTIDVisible(EState *estate, ResultRelInfo *relinfo, ItemPointer tid,
 								TupleTableSlot *tempSlot);
+/*
+ * A placeholder to save all context required to batch filter
+ * compressed chunks during UPDATE or DELETE operation.
+ */
+typedef struct PredicatesContext
```
akuzm, line 66: `PredicatesContext` sounds very generic, maybe something more specific like `CompressedBatchPredicates`?

src/nodes/hypertable_modify.c
sb230132: This code is no longer valid.

src/nodes/hypertable_modify.c
sb230132: Sounds good. Now i only evaluate the predicates to check if there are any SEGMENT BY columns and what type of condition it is.
ex: <segmentby_col> = <Const> OR  <segmentby_col> IS NULL OR  <segmentby_col> IS NOT NULL

In later code where i iterate over the chunks, i build scan keys and decompress compressed chunk.

src/nodes/hypertable_modify.c
sb230132: Fixed.

src/nodes/hypertable_modify.c
sb230132: If i dont have this code then, later when SeqScan nodes tries to fetch rows from uncompressed chunk, it will not find any rows. Sven mentioned that transaction snapshot should be same for batch filtering and Scan node.

For more details please refer: 
https://iobeam.slack.com/archives/C04QFNN1H6Y/p1677483665605009

src/nodes/hypertable_modify.c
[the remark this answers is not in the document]
sb230132: Ok, added new test file compression_dml_iso.spec which will run only for PG versions >= 14

src/cross_module_fn.h
```diff
@@ -139,6 +139,12 @@ typedef struct CrossModuleFunctions
 	PGFunction decompress_chunk;
 	void (*decompress_batches_for_insert)(ChunkInsertState *state, Chunk *chunk,
 										  TupleTableSlot *slot);
+	bool (*decompress_batches_for_update_delete)(Chunk *ch, ScanKeyData *scankeys, int num_scankeys,
+												 Bitmapset *null_columns, List *is_nulls,
+												 bool *chunk_status_changed);
+	ScanKeyData *(*build_update_delete_scankeys)(Chunk *chunk, List *segmentby_columns,
+												 List *segmentby_columns_value, List *is_null_check,
+												 int *num_scankeys, Bitmapset **null_columns);
```
svenklemm, line 147: why would the ScanKeys ever need to cross the TSL/non-TSL boundary, all the dedcompression should happen localized in the TSL code

src/nodes/hypertable_modify.c
akuzm: The slack link doesn't work for me, probably because I wasn't in that conversation at that point. Anyway, let's add a more detailed explanation to the code comment.

sb230132: Oops, sorry, added you to the group, i hope now you can access the chat.

src/nodes/hypertable_modify.c
akuzm: But the current code still traverses the RestrictInfos once per chunk, right? We could do this once per hypertable, but I think we can leave this optimization for the future.

akuzm (approved):
There are still many improvements to do, but I think we should merge it soon to make progress. Probably makes sense to do the refactoring that Sven mentioned, to place the functions in proper modules.

src/cross_module_fn.h
sb230132: Move most of code related to constraints handling to compression.c. So above code will no longer be valid.

src/nodes/hypertable_modify.c
sb230132: Nope, please check the function call to 
```
static CompressedBatchPredicates *fill_predictate_context(Chunk *ch, List *predicates)
```
I pass 1 of the chunks to evaluate WHERE and save all context in CompressedBatchPredicates.
This saved context is used during processing of each chunk where i fill the attribute numbers for building scan keys.

So the optimisation you proposed is still in place.

src/nodes/hypertable_modify.c
```diff
@@ -745,7 +786,33 @@ ExecModifyTable(PlanState *pstate)
...
+	{
+		collect_chunks_from_scan(pstate, ht_state);
+		if (ht_state->chunks && ts_cm_functions->decompress_batches_for_update_delete)
+		{
+			ts_cm_functions->decompress_batches_for_update_delete(ht_state->chunks,
+																  ht_state->predicates);
+			ht_state->comp_chunks_processed = true;
+		}
+		if (ht_state->comp_chunks_processed)
+		{
```
svenklemm, line 803:
```suggestion
```

src/nodes/hypertable_modify.h
```diff
@@ -27,6 +27,12 @@ typedef struct HypertableModifyState
 	CustomScanState cscan_state;
 	ModifyTable *mt;
 	List *serveroids;
+	/* list of compressed chunks */
+	List *chunks;
+	/* list of conditions specified in WHERE */
+	List *predicates;
```
svenklemm, line 33: why do we need to keep this in the state we only need this information once to do the batch decompression

src/nodes/hypertable_modify.c
sb230132: Fixed, thanks.

src/nodes/hypertable_modify.h
sb230132: Right, removed it.

tsl/src/compression/compression.c
```diff
@@ -1994,3 +1994,349 @@ decompress_batches_for_insert(ChunkInsertState *cis, Chunk *chunk, TupleTableSlo
...
+{
+	/* list of segmentby column names specified in WHERE */
+	List *segmentby_columns;
+	/* list of segmentby column values */
+	List *segmentby_columns_value;
+	/* represents if condition is IS [NOT] NULL or something else */
+	List *is_null_check;
+	/* represents if condition is IS NULL or IS NOT NULL */
+	List *is_null;
+} CompressedBatchPredicates;
```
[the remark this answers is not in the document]
sb230132: This is to resolve 1 of the comments posted by @akuzm 
https://github.com/timescale/timescaledb/pull/5339#discussion_r1154445511

So what i do is split evaluation all predicates and initialising scan keys into 2 separate steps. This way i can avoid on performing same repetitive job of evaluating the WHERE for each chunk.

tsl/src/compression/compression.c
```diff
@@ -1994,3 +1994,330 @@ decompress_batches_for_insert(ChunkInsertState *cis, Chunk *chunk, TupleTableSlo
...
+ * This method will evaluate the predicates, extract
+ * left and right operands, check if one of the operands is
+ * a simple Var type. If its Var type extract its corresponding
+ * column name from hypertable_compression catalog table.
+ * If extracted column is a SEGMENT BY column then save column
+ * name, value specified in the predicate. This information will
+ * be used to build scan keys later.
+ */
+static void
+fill_predictate_context(Chunk *ch, List *predicates, List **segmentby_columns,
```
svenklemm, line 2030:
```suggestion
fill_predicate_context(Chunk *ch, List *predicates, List **segmentby_columns,
```

svenklemm: typo in function name

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ accidentally triggering the load of a previous DB version.**
 * #5454 Add support for ON CONFLICT DO UPDATE for compressed hypertables
 * #5344 Enable JOINS for Hierarchical Continuous Aggregates
 * #5417 Refactor and optimize distributed COPY
+* #5339 Support UPDATE/DELETE sql on compressed hypertables
```
svenklemm, line 19:
```suggestion
* #5339 Support UPDATE/DELETE on compressed hypertables
```

svenklemm (approved):

tsl/src/compression/compression.c
sb230132: Fixed.

CHANGELOG.md
sb230132: Fixed.

Micromegass:
@sb230132 Since this is a very important feature for me (and probably for many), is there any info when and/or in which version this feature will be released?

PR #5346, opened 2023-02-17
Decompress all rows in one go in Simple8b-RLE decompressor

akuzm:
Our rows are compressed in batches no more than 1000 rows, so it is optimal to process the entire batch in one go.

This improves the performance of some simple queries like "select min(value) from compressed_table" by 5-10%. Also serves as a building block for the future optimizations.

Part of #5282.


svenklemm (changes requested):
I dont think its a good idea to change the default decompressor like this. This has performance implications eg in the presence of LIMIT but also different memory requirements. If we want a way to decompress in a non-iterative way this should be separate functions.

akuzm:
I think in the LIMIT case the performance is dominated by other things. This function decompressing 1k rows max is really fast. I can make some benchmarks. Even if there's a small regression like 15%, I think this can be acceptable given the overall increase in bulk queries.

About the memory usage, of course it's going to increase. But if we don't have enough memory to cache 1k rows, we should forget about OLAP in general. Normally they use even larger blocks, like 65k or 1M rows. If we're going to use vectorized computation, there's no way around materializing batches of data in memory.

akuzm:
Benchmarked queries like `select distinct on (id) id, ts, value from metric_compressed order by id, ts limit 1;`
The break-even point is at LIMIT 2 :sweat_smile: I had to use pgbench with the prepared statements to get the planning time out of the way.

limit | row-by-row, ms | batch, ms | speedup
-- | -- | -- | --
1 | 0.497 | 0.699 | -1.41
2 | 3.105 | 3.055 | 1.02
10 | 36.23 | 33.49 | 1.08
100 | 85.147 | 75.772 | 1.12

@svenklemm do you think extra 200 microseconds for LIMIT 1 is acceptable? Technically it's not difficult to do a separate implementation for LIMIT 1, but I would prefer not to support a second variant of this code for such a narrow case.

akuzm:
> The break-even point is at LIMIT 2

The reason is that the row-by-row decompression takes noticeably longer for two rows than for one, and also MergeAppend kicks in with its slow `binaryheap_replace_first` thing.

akuzm:
> Technically, it's not difficult to do a separate implementation for LIMIT 1, but I would prefer not to support a second variant of this code for such a narrow case.

Although maybe we won't need an entire "second implementation", just passing the limit value through to the decompressors will be enough.

svenklemm:
> 1	0.497	0.699	-1.41
> 2	3.105	3.055	1.02
> 10	36.23	33.49	1.08
> 100	85.147	75.772	1.12

According to your table it increases by 40% for limit 1. Though I'm not sure what the time is measuring. Can you also add the speedup for 1000 rows to see what the maximum gain would be. I think we want both implementations. Compressing all tuples at once has higher memory requirements and we might want to switch implementation depending on memory constraints. There is also executor support for tuple limits (see ExecSetTupleBound) which we could use for the decission.

akuzm:
> According to your table it increases by 40% for limit 1. Though I'm not sure what the time is measuring.

Yes. This is query execution time without planning, I use pgbench with prepared statements. The absolute increase is small, ~200 microseconds.

>  I think we want both implementations. 

The problem is how we switch between them. One way is to push down the limit to the DecompressChunk node, so that it can choose the bulk vs single row algorithm. But in this example query we can't do that, it would give incorrect results. 

The second way is to automatically switch to batch decompression on the second row, or wherever the breakeven point is. This is going to be very complex -- maybe just for decompression it could be OK, but imagine if we also have vectorized quals and aggregates and whatnot. Having two completely separate pipelines for all that, and switching between them on the fly, will be very bugprone.

So I'd sacrifice these 200 ms on LIMIT 1 case to avoid this complexity and speed up the rest of queries.

akuzm:
Probably will merge this in a different form as part of other batch decompression changes.

PR #5351, opened 2023-02-20
Forcefully enable tiered data for CAGGs

zilder:
When cagg refresh occurs and is triggered by the background worker tiered data may be excluded if `timescaledb.enable_tiered_reads` is off on the server level. Also when refresh is triggered manually by a user whether tiered data appears in the cagg or not depends on the user's session settings despite of the server settings.
This PR fixes that by forcefully enabling the `timescaledb.enable_tiered_reads` GUC in the cagg refresh procedure.

### DIFF
--- a/tsl/src/continuous_aggs/refresh.c
+++ b/tsl/src/continuous_aggs/refresh.c
@@ -711,6 +711,7 @@ continuous_agg_refresh_internal(const ContinuousAgg *cagg,
 	int64 computed_invalidation_threshold;
 	int64 invalidation_threshold;
 	bool is_raw_ht_distributed;
+	bool old_enable_osm_reads = ts_guc_enable_osm_reads;
 	int rc;
 
 	/* Connect to SPI manager due to the underlying SPI calls */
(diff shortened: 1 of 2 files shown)


fabriziomello (changes requested):
Would be good add some tests demonstrating that even with the GUC disabled we refresh the caggs and the materialized data correspond also to tiered data

zilder:
I don't think it can be done without having the osm extension installed. I'm planning to add these tests in osm.
But also after having a chat with @gayyappan we might want to rework this PR, gotta discuss these changes further.

zilder:
...or maybe it can be done with `file_fdw` or `postgres_fdw`, will check.

gayyappan:
> ...or maybe it can be done with `file_fdw` or `postgres_fdw`, will check.

We have tests with fdw setup for testing osm. see chunk_utils_internal.sql

zilder:
Updating this old PR (thought it was merged). Added a test demonstrating that tiered data can be accessed when refreshing CAggs.

gayyappan:
> Updating this old PR (thought it was merged). Added a test demonstrating that tiered data can be accessed when refreshing CAggs.
Didn't we decide not to go ahead with this?

tsl/src/continuous_aggs/refresh.c
```diff
@@ -879,9 +880,14 @@ continuous_agg_refresh_internal(const ContinuousAgg *cagg,
 
 	cagg = ts_continuous_agg_find_by_mat_hypertable_id(mat_id, false);
 
+	/* Cagg should be able to see tiered data despite the server settings */
+	ts_guc_enable_osm_reads = true;
```
fabriziomello, line 884: Hmmm... I don't think this way is the best, maybe we should introduce a GUC like `enable_cagg_osm_reads` that do this trick that is ON by default... wondering if for some reason (performance or whatever) we should NOT read osm data during the cagg refresh.

svenklemm:
Closing this since the PR is over 2 years old

PR #5362, opened 2023-02-24
Make copy fetcher more async

erimatnor:
Make the copy fetcher more asynchronous by separating the sending of the request for data from the receiving of the response. By doing that, the async append node can send the request to each data node before it starts reading the first response. This can massively improve the performance because the response isn't returned until the remote node has finished executing the query and is ready to return the first tuple.


pmwkaa (approved):
Looks good

akuzm (approved):

PR #5368, opened 2023-02-28
Renamed size_utils.sql

sotirissl:
Renamed test/sql/size_utils.sql
to   test/sql/utils_size.sql
because conflicting with tsl/test/sql/size_utils.sql


akuzm (approved):
These names are kind of confusing :) I'd suggest `size_utils_apache` and `size_utils_tsl`.

fabriziomello:
> These names are kind of confusing :) I'd suggest `size_utils_apache` and `size_utils_tsl`.

Agree that the current renaming is a bit confusing but would like to simplify a bit it:  `size_utils.sql` and `size_utils_tsl.sql`

fabriziomello (approved):

jnidzwetzki (approved):

mkindahl (approved):

Issue #4846, opened 2022-10-18
Multinode HA: handle DML to chunks when target DN is down


nikkhils:
### What problem does the new feature solve?

https://github.com/timescale/product/issues/288 lays down the requirements for Datanode HA. We want to allow both read and write queries when a datanode (DN) goes down. 

One way to allow write queries when DN goes down is to track which chunk on that down DN is being written to and to mark that.

### What does the feature do?

Instead of tracking dirty chunks explicitly on AN, we will remove the metadata on the AN which associates this down DN with such a dirty chunk. Since we will remove metadata on the AN, any future queries will never go to this DN for this data avoiding the issue of stale data fetches completely.

When the DN comes back, the rebalancer microservice in TSC (Timescale Cloud) will kick in automatically and will refresh these chunks on the DN. That will need changes in [copy_chunk/move_chunk functionality](https://github.com/timescale/timescaledb/issues/4847). 

The DN can also when it comes back tell the AN about its chunk and AN can also cleanup [stale chunks](https://github.com/timescale/timescaledb/issues/4848) that ways on it.

Hat tip to @pmwkaa for this overall idea of removing metadata on the AN as the way to track the dirtied chunks. 

This issue will only track the AN handling of the removing metadata part.

jfjoly:
Note on the behavior: when a DN is unavailable, new chunks will be created on the "surviving" nodes. The replication factor may be below the desired one until all DNs come back online.

PR #5378, opened 2023-03-02
Fix perf regression due to DML HA

nikkhils:
We added checks via #4846 to handle DML HA when replication factor is
greater than 1 and a datanode is down. Since each insert can go to a different
chunk with a different set of datanodes, we added checks on every insert to check if DNs are unavailable. This increased CPU consumption on the AN leading to a performance regression for RF > 1 code paths.

This patch fixes this regression. We now track if any DN is marked as unavailable at the start of the transaction and use that information to reduce unnecessary checks for each inserted row.

### DIFF
--- a/tsl/src/remote/dist_copy.c
+++ b/tsl/src/remote/dist_copy.c
@@ -129,7 +129,8 @@ typedef struct RemoteCopyContext
 	List *attnums;
 	void *data_context; /* TextCopyContext or BinaryCopyContext */
 	bool binary_operation;
-	MemoryContext mctx; /* MemoryContext that holds the RemoteCopyContext */
+	MemoryContext mctx;	  /* MemoryContext that holds the RemoteCopyContext */
+	bool dns_unavailable; /* are some DNs marked as "unavailable"? */
 
 	/*
 	 * Incoming rows are batched before creating the chunks and sending them to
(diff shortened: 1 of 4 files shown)


pmwkaa (approved):

tsl/src/remote/dist_copy.c
```diff
@@ -1333,6 +1335,8 @@ remote_copy_process_and_send_data(RemoteCopyContext *context)
 	Hypertable *ht = context->ht;
 	const int n = context->batch_row_count;
 	Assert(n <= MAX_BATCH_ROWS);
+	static int32 chunk_id = INVALID_CHUNK_ID;
```
pdipesh02, line 1338: Why do we need this variables to be static as these are being reset when the function returns?

tsl/src/remote/dist_copy.c
```diff
@@ -1366,12 +1370,35 @@ remote_copy_process_and_send_data(RemoteCopyContext *context)
...
+		 *
+		 * The worse case scenario is one in which INSERT1 goes into CHUNK1, INSERT2 goes into
+		 * CHUNK2, INSERT3 goes into CHUNK1,... in which case we will end up refreshing the list
+		 * everytime
+		 *
+		 * We will also enter the below loop if we KNOW that any of the DNs has been marked
+		 * unavailable before we started this transaction. If not, then we know that every chunk's
+		 * datanode list is fine and no stale chunk metadata updates are needed.
+		 */
+		if (context->dns_unavailable && found && ht->fd.replication_factor > 1)
```
pdipesh02, line 1386: Shouldn't we make a call to reset `context->dns_unavailable` for the remaining rows in the batch after removing the data nodes for stale chunk?

tsl/src/remote/dist_copy.c
nikkhils: It needs to be static for the loop to cache the previous chunk_id. Assuming that nearby inserts will go to the same chunk, we this ways, avoid additional DN scans. @pdipesh02

tsl/src/remote/dist_copy.c
nikkhils: No. The `dns_unavailable` is a global state for the entire transaction. We just want to avoid doing these expensive checks if all DNs are available.

pdipesh02 (approved):

tsl/src/remote/dist_copy.c
nikkhils: Hmm, yes, good point @pdipesh02 will remove the statics!

PR #5388, opened 2023-03-03
Add hooks for hypertable deletion

zilder:
To properly clean up the OSM catalog we need a way to reliably track hypertables deletion (including internal hypertables for CAGGs). This PR adds two more hooks:
- `hypertable_drop_hook`
- `hypertable_drop_by_name_hook`

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -54,6 +54,7 @@
 #include "hypercube.h"
 #include "hypertable.h"
 #include "hypertable_cache.h"
+#include "osm_callbacks.h"
 #include "partitioning.h"
 #include "process_utility.h"
 #include "scan_iterator.h"
@@ -1156,27 +1157,15 @@ ts_chunk_create_only_table(Hypertable *ht, Hypercube *cube, const char *schema_n
 	return chunk;
 }
 
-#if PG14_GE
-#define OSM_CHUNK_INSERT_CHECK_HOOK "osm_chunk_insert_check_hook"
-typedef int (*ts_osm_chunk_insert_hook_type)(Oid ht_oid, int64 range_start, int64 range_end);
-static ts_osm_chunk_insert_hook_type
-get_osm_chunk_insert_hook()
-{
-	ts_osm_chunk_insert_hook_type *func_ptr =
-		(ts_osm_chunk_insert_hook_type *) find_rendezvous_variable(OSM_CHUNK_INSERT_CHECK_HOOK);
-	return *func_ptr;
-}
-#endif
-
 static Chunk *
 chunk_create_from_hypercube_after_lock(const Hypertable *ht, Hypercube *cube,
 									   const char *schema_name, const char *table_name,
 									   const char *prefix)
 {
 #if PG14_GE
-	ts_osm_chunk_insert_hook_type insert_func_ptr = get_osm_chunk_insert_hook();
+	OsmCallbacks *callbacks = ts_get_osm_callbacks();
 
-	if (insert_func_ptr)
+	if (callbacks)
 	{
 		/* OSM only uses first dimension . doesn't work with multinode tables yet*/
 		Dimension *dim = &ht->space->dimensions[0];
@@ -1185,7 +1174,10 @@ chunk_create_from_hypercube_after_lock(const Hypertable *ht, Hypercube *cube,
 			ts_internal_to_time_int64(cube->slices[0]->fd.range_start, dim->fd.column_type);
 		int64 range_end =
 			ts_internal_to_time_int64(cube->slices[0]->fd.range_end, dim->fd.column_type);
-		int chunk_exists = insert_func_ptr(ht->main_table_relid, range_start, range_end);
+
+		int chunk_exists =
+			callbacks->chunk_insert_check_hook(ht->main_table_relid, range_start, range_end);
+
 		if (chunk_exists)
 		{
 			Oid outfuncid = InvalidOid;
--- a/src/hypertable.c
+++ b/src/hypertable.c
@@ -63,6 +63,7 @@
 #include "cross_module_fn.h"
 #include "scan_iterator.h"
 #include "debug_assert.h"
+#include "osm_callbacks.h"
 
 Oid
 ts_rel_get_owner(Oid relid)
@@ -624,6 +625,20 @@ hypertable_tuple_delete(TupleInfo *ti, void *data)
 		if (compressed_hypertable != NULL)
 			ts_hypertable_drop(compressed_hypertable, DROP_RESTRICT);
 	}
+
+#if PG14_GE
+	OsmCallbacks *callbacks = ts_get_osm_callbacks();
+
+	/* Invoke the OSM callback if set */
+	if (callbacks)
+	{
+		Name schema_name =
+			DatumGetName(slot_getattr(ti->slot, Anum_hypertable_schema_name, &isnull));
+		Name table_name = DatumGetName(slot_getattr(ti->slot, Anum_hypertable_table_name, &isnull));
+
+		callbacks->hypertable_drop_hook(NameStr(*schema_name), NameStr(*table_name));
+	}
+#endif
 
 	ts_catalog_database_info_become_owner(ts_catalog_database_info_get(), &sec_ctx);
 	ts_catalog_delete_tid(ti->scanrel, ts_scanner_get_tuple_tid(ti));
--- a/src/osm_callbacks.c
+++ b/src/osm_callbacks.c
@@ -1,0 +1,18 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+#include "osm_callbacks.h"
+
+#include <fmgr.h>
+
+#define OSM_CALLBACKS_VAR_NAME "osm_callbacks"
+
+OsmCallbacks *
+ts_get_osm_callbacks(void)
+{
+	OsmCallbacks **ptr = (OsmCallbacks **) find_rendezvous_variable(OSM_CALLBACKS_VAR_NAME);
+
+	return *ptr;
+}
--- a/src/osm_callbacks.h
+++ b/src/osm_callbacks.h
@@ -1,0 +1,29 @@
+/*
+ * This file and its contents are licensed under the Apache License 2.0.
+ * Please see the included NOTICE for copyright information and
+ * LICENSE-APACHE for a copy of the license.
+ */
+#ifndef TIMESCALEDB_OSM_CALLBACKS_H
+#define TIMESCALEDB_OSM_CALLBACKS_H
+
+#include <postgres.h>
+#include <catalog/objectaddress.h>
+
+typedef int (*chunk_insert_check_hook_type)(Oid ht_oid, int64 range_start, int64 range_end);
+typedef void (*hypertable_drop_hook_type)(const char *schema_name, const char *table_name);
+
+/*
+ * Object Storage Manager callbacks.
+ *
+ * chunk_insert_check_hook - checks whether the specified range is managed by OSM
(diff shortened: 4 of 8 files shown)


src/osm_callbacks.h
```diff
@@ -0,0 +1,25 @@
...
+typedef void (*hypertable_drop_hook_type)(const ObjectAddress *object_address);
+typedef void (*hypertable_drop_by_name_hook_type)(const char *schema_name, const char *table_name);
+
+typedef struct {
+	chunk_insert_check_hook_type 		chunk_insert_check_hook;
+	hypertable_drop_hook_type			hypertable_drop_hook;
+	hypertable_drop_by_name_hook_type	hypertable_drop_by_name_hook;
+} OsmCallbacks;
+
+OsmCallbacks *get_osm_callbacks(void);
```
svenklemm, line 23: any exported function should have `ts_` as prefix

zilder: Thanks, fixed!

gayyappan (commented):
Need some tests.

tsl/test/sql/chunk_utils_internal.sql
```diff
@@ -267,6 +267,14 @@ SELECT  _timescaledb_internal.drop_chunk( :'CHNAME1');
...
 SELECT * FROM hyper1_cagg ORDER BY 1;
 
+-- check that dropping cagg triggers OSM callback
+SELECT ts_setup_osm_hook();
+BEGIN;
+DROP MATERIALIZED VIEW hyper1_cagg CASCADE;
+DROP TABLE test1.hyper1;
+ROLLBACK;
+SELECT ts_undo_osm_hook();
+
```
gayyappan, line 277: We should have 1 more test. Expect 2 calls to hooks : 1 for mat hypertable and 1 for the hypertable itself
BEGIN;
DROP TABLE test1.hyper1 CASCADE;
ROLLBACK;

src/ts_catalog/continuous_agg.c
```diff
@@ -1184,6 +1185,13 @@ drop_continuous_agg(FormData_continuous_agg *cadata, bool drop_user_view)
 
 	if (OidIsValid(mat_hypertable.objectId))
 	{
+#if PG14_GE
+		OsmCallbacks *callbacks = ts_get_osm_callbacks();
+
+		if (callbacks)
+			callbacks->hypertable_drop_hook(&mat_hypertable);
```
gayyappan, line 1192: why do we need a different hook type? Can't we just use the drop_by_name hook.
ts_hypertable_get_by_id() will fetch the hypertable info and we can get the schema_name + table_name from there.

src/osm_callbacks.h
```diff
@@ -0,0 +1,25 @@
...
+ * LICENSE-APACHE for a copy of the license.
+ */
+#ifndef TIMESCALEDB_OSM_CALLBACKS_H
+#define TIMESCALEDB_OSM_CALLBACKS_H
+
+#include <postgres.h>
+#include <catalog/objectaddress.h>
+
+typedef int (*chunk_insert_check_hook_type)(Oid ht_oid, int64 range_start, int64 range_end);
+typedef void (*hypertable_drop_hook_type)(const ObjectAddress *object_address);
```
gayyappan, line 13: Think we can get rid of this hook. See continuous_agg.c comment

src/hypertable.c
```diff
@@ -698,6 +699,17 @@ ts_hypertable_drop(Hypertable *hypertable, DropBehavior behavior)
 		/* Drop the postgres table */
 		performDeletion(&hypertable_addr, behavior, 0);
 	}
+
+#if PG14_GE
+	OsmCallbacks *callbacks = ts_get_osm_callbacks();
+
+	if (callbacks)
+	{
+		callbacks->hypertable_drop_by_name_hook(hypertable->fd.schema_name.data,
```
gayyappan, line 708:
```suggestion
		callbacks->hypertable_drop_by_name_hook(NameStr(hypertable->fd.schema_name), NameStr(hypertable->fd.table_name))
```

gayyappan (commented):
Minor changes requested

tsl/test/sql/chunk_utils_internal.sql
zilder: Added

src/ts_catalog/continuous_agg.c
zilder: Good point, thanks!

zilder: Removed one redundant hook

src/osm_callbacks.h
zilder: Done

src/hypertable.c
zilder: Just copy-pasted from few lines below. But I like your suggestion more : )

src/osm_callbacks.c
```diff
@@ -0,0 +1,18 @@
...
+#include "osm_callbacks.h"
+
+#include <fmgr.h>
+
+#define OSM_CALLBACKS_VAR_NAME "osm_callbacks"
+
+OsmCallbacks *
+ts_get_osm_callbacks(void)
+{
+	OsmCallbacks **ptr = (OsmCallbacks **) find_rendezvous_variable(OSM_CALLBACKS_VAR_NAME);
```
svenklemm, line 15: i think we should have a static variable holding the result to avoid doing the lookup on every call

svenklemm (approved):

src/ts_catalog/continuous_agg.c
```diff
@@ -1184,6 +1185,18 @@ drop_continuous_agg(FormData_continuous_agg *cadata, bool drop_user_view)
...
 	if (OidIsValid(mat_hypertable.objectId))
 	{
+#if PG14_GE
+		OsmCallbacks *callbacks = ts_get_osm_callbacks();
+
+		if (callbacks)
+		{
+			Hypertable *mat_ht = ts_hypertable_get_by_id(cadata->mat_hypertable_id);
+
+			callbacks->hypertable_drop_hook(NameStr(mat_ht->fd.schema_name),
```
kgyrtkirk, line 1195: this file is called `contious_agg.c` - these lines looks out-of-place here ; I think this hook should be invoked during `ts_hypertable_delete_by_id` or not?

pmwkaa:
I am curious why we needed it, is using event hooks for this was not enough?

src/osm_callbacks.h
```diff
@@ -0,0 +1,23 @@
...
+#include <catalog/objectaddress.h>
+
+typedef int (*chunk_insert_check_hook_type)(Oid ht_oid, int64 range_start, int64 range_end);
+typedef void (*hypertable_drop_hook_type)(const char *schema_name, const char *table_name);
+
+typedef struct
+{
+	chunk_insert_check_hook_type chunk_insert_check_hook;
+	hypertable_drop_hook_type hypertable_drop_hook;
+} OsmCallbacks;
```
kgyrtkirk, line 19: I think its a bit unfortunate that the API defines the number of calls directly - if you want to add a new method you will need a second version....

how about approaching differently:
* add just a single method
* an opcode why the fook was called
* a context object with the currently known infos (table cat+name)

this way later on
* adding a new call is quite simple - just add a new enum value
* passing new infos is trivial since the old context is the prefix of the new
* no need to recompile old hooks when rolling out a new version of the...it will work without issues

src/osm_callbacks.h
```diff
@@ -0,0 +1,23 @@
...
+ * LICENSE-APACHE for a copy of the license.
+ */
+#ifndef TIMESCALEDB_OSM_CALLBACKS_H
+#define TIMESCALEDB_OSM_CALLBACKS_H
+
+#include <postgres.h>
+#include <catalog/objectaddress.h>
+
+typedef int (*chunk_insert_check_hook_type)(Oid ht_oid, int64 range_start, int64 range_end);
+typedef void (*hypertable_drop_hook_type)(const char *schema_name, const char *table_name);
```
kgyrtkirk, line 13: if this is only about `deletion` as the PR title suggest; then what does `insert` doing here?

I feel that only declared `drop` + `insert` ; but I think other things may happen as well...what about: create, delete/etc ?

other parts of the PR makes me wonder if this is a general `hypertable` lifecycle hook...

src/chunk.c
```diff
@@ -1156,27 +1157,15 @@ ts_chunk_create_only_table(Hypertable *ht, Hypercube *cube, const char *schema_n
...
 chunk_create_from_hypercube_after_lock(const Hypertable *ht, Hypercube *cube,
 									   const char *schema_name, const char *table_name,
 									   const char *prefix)
 {
 #if PG14_GE
-	ts_osm_chunk_insert_hook_type insert_func_ptr = get_osm_chunk_insert_hook();
+	OsmCallbacks *callbacks = ts_get_osm_callbacks();
 
-	if (insert_func_ptr)
+	if (callbacks)
```
kgyrtkirk, line 1168: I feel that calling this `callbacks` while there might be only one registered handler is a bit misleading - but in the current concept it contains "2" handlers...but since those are both from the same source I wouldn't add the `s`....

src/chunk.c
```diff
@@ -1185,7 +1174,10 @@ chunk_create_from_hypercube_after_lock(const Hypertable *ht, Hypercube *cube,
 			ts_internal_to_time_int64(cube->slices[0]->fd.range_start, dim->fd.column_type);
 		int64 range_end =
 			ts_internal_to_time_int64(cube->slices[0]->fd.range_end, dim->fd.column_type);
-		int chunk_exists = insert_func_ptr(ht->main_table_relid, range_start, range_end);
+
+		int chunk_exists =
+			callbacks->chunk_insert_check_hook(ht->main_table_relid, range_start, range_end);
+
 		if (chunk_exists)
```
kgyrtkirk, line 1181: this block became conditional based on the presence of the state of `callbacks` above; and based on the result of the `hook` some check is being executed...I think the below block either belongs to the `new hook` ; or the old code should remain here....in its current form I feel its a bit odd..

src/osm_callbacks.h
```diff
@@ -0,0 +1,23 @@
...
+
+typedef struct
+{
+	chunk_insert_check_hook_type chunk_insert_check_hook;
```
kgyrtkirk, line 17: please document what the hook is for - looking at the if in the other file ; I'm a bit confused what its contract could be

src/ts_catalog/continuous_agg.c
zilder: Would you say `hypertable_tuple_delete` is a good place for it?

src/osm_callbacks.h
zilder: Insert callback is for checking for intersections with chunks managed by OSM. This check already was in the code, I'm not adding it, see code changes in `chunks.c`. We just decided it would be nicer to have one struct with all the callbacks OSM might need.

src/chunk.c
zilder: I can rename it into `osm_callbacks`. Would make more sense?

zilder: We're getting a struct containing multiple callbacks, we only use one in this particular situation.

zilder:
> I am curious why we needed it, is using event hooks for this was not enough?

@pmwkaa You mean event triggers? It didn't work for caggs (or rather for underlying materialized hypertable). As far as I can tell timescale uses an event trigger on `sql_drop` to track drops on mat views and calls `performDeletion` on the underlying materialized ht from the trigger function. In our case our own event trigger in OSM was firing before timescale's one and couldn't see subsequent events.

src/osm_callbacks.c
gayyappan: find rendezvous variable already allocs the var as a hash entry and returns that on subsequent calls. So I don't think there is a need for an additional static here.

gayyappan (approved):
Changes look good.
Please undo changes to continuous_agg.c and process_utility.c before merging.

PR #5395, opened 2023-03-06
Revert changing merge driver for CHANGELOG

svenklemm:
This is causing bad merges in the CHANGELOG often resulting in changes being added to previuous releases instead of at the top. I think it's better to error here and have a human review instead of having a CHANGELOG that has entries in random places.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,0 @@
-# to avoid frequent conflicts
-CHANGELOG.md text merge=union


akuzm:
Yeah, union merge doesn't work well when there are deletions. But this will basically disable the automated backports, because `CHANGELOG.md` never merges cleanly into the back branches I think.

svenklemm:
I guess I'll close this for now since we plan to change the way we handle changelog anyway. Have to take extra care when doing changelog in the meantime.

PR #5400, opened 2023-03-06
Look up compressed column metadata at planning time

akuzm:
Now we look them up at execution time for each chunk separately, which adds up for tables with a large number of chunks.

This gives about 15% speedup (100 mcs) on a small query on a table from tests with 50 chunks:
`select id, ts, value from metric_compressed order by id, ts limit 100;`

### DIFF
--- a/tsl/src/nodes/decompress_chunk/decompress_chunk.h
+++ b/tsl/src/nodes/decompress_chunk/decompress_chunk.h
@@ -49,6 +49,17 @@ typedef struct DecompressChunkPath
 	 * uncompressed chunk, but are still used for decompression.
 	 */
 	List *decompression_map;
+
+	/*
+	 * This Int list is parallel to the compressed scan targetlist, just like
+	 * the above one. The value is true if a given targetlist entry is a
+	 * segmentby column, false otherwise. Has the same length as the above list.
+	 * We have to use the parallel lists and not a list of structs, because the
+	 * Plans have to be copyable by the Postgres _copy functions, and we can't
+	 * do that for a custom struct.
+	 */
+	List *is_segmentby_column;
+
 	List *compressed_pathkeys;
 	bool needs_sequence_num;
 	bool reverse;
--- a/tsl/src/nodes/decompress_chunk/exec.c
+++ b/tsl/src/nodes/decompress_chunk/exec.c
@@ -71,6 +71,7 @@ typedef struct DecompressChunkState
 {
 	CustomScanState csstate;
 	List *decompression_map;
+	List *is_segmentby_column;
 	int num_columns;
 	DecompressChunkColumnState *columns;
 
@@ -112,6 +113,7 @@ decompress_chunk_state_create(CustomScan *cscan)
 	state->chunk_relid = lsecond_int(settings);
 	state->reverse = lthird_int(settings);
 	state->decompression_map = lsecond(cscan->custom_private);
+	state->is_segmentby_column = lthird(cscan->custom_private);
 
 	return (Node *) state;
 }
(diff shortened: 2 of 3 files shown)


svenklemm:
Did you test this with chunks with  different physical layouts?

akuzm:
> Did you test this with chunks with different physical layouts?

Yes, we have some tests like this, e.g. in `compression_alter.sql`. I think I phrased it incorrectly, the DecompressChunk plans are still per-chunk, but we can avoid a second lookup at the execution time.

tsl/src/nodes/decompress_chunk/decompress_chunk.h
```diff
@@ -49,6 +49,13 @@ typedef struct DecompressChunkPath
 	 * uncompressed chunk, but are still used for decompression.
 	 */
 	List *decompression_map;
+
+	/*
+	 * Whether a targetlist entry is a segmentby column. Parallel to the above
```
jnidzwetzki, line 54: It took me some time to understand what this list actually contains. Could you expand the comment a bit and mention that it has the same length as the list above and contains int values for each attribute to indicate if it is a `segment_by` column or not?

jnidzwetzki (commented):
I added two minor comments, the rest looks good to me.

svenklemm (approved):

tsl/src/nodes/decompress_chunk/planner.c
```diff
@@ -227,6 +227,9 @@ build_decompression_map(DecompressChunkPath *path, List *scan_tlist, Bitmapset *
 
 		path->decompression_map =
 			lappend_int(path->decompression_map, destination_attno_in_uncompressed_chunk);
+		path->is_segmentby_column =
+			lappend_int(path->is_segmentby_column,
+						compression_info && compression_info->segmentby_column_index != 0);
```
kgyrtkirk, line 232: I think instead of double-walking the 2 Lists in parallell a new struct should be created with 2 int fields; and store&pass those - it would reduce future confusion this might cause

it would be much clearer and would need less comments/etc (and would not need the assert @jnidzwetzki  is asking for a few lines above)
```
struct compressed_col_info {
  int  destination_attno;
  bool is_segmentby_column;
}
```

akuzm: Unfortunately for `custom_private` we have to stick with data types copyable by postgres `_copy*` functions, and this precludes introducing any structs.

tsl/src/nodes/decompress_chunk/decompress_chunk.h
akuzm: I changed the comment, does it look better? I have trouble formulating this in a comprehensible way...

tsl/src/nodes/decompress_chunk/exec.c
```diff
@@ -139,11 +139,13 @@ initialize_column_state(DecompressChunkState *state)
 
 	AttrNumber next_compressed_scan_attno = 0;
 	state->num_columns = 0;
-	foreach (lc, state->decompression_map)
+	ListCell *dest_cell;
+	ListCell *is_segmentby_cell;
+	forboth (dest_cell, state->decompression_map, is_segmentby_cell, state->is_segmentby_column)
```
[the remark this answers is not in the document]
akuzm: Added.

tsl/src/nodes/decompress_chunk/decompress_chunk.h
```diff
@@ -49,6 +49,17 @@ typedef struct DecompressChunkPath
 	 * uncompressed chunk, but are still used for decompression.
 	 */
 	List *decompression_map;
+
+	/*
+	 * This Int list is parallel to the compressed scan targetlist, just like
+	 * the above one. The value is true if a given targetlist entry is a
+	 * segmentby column, false otherwise. Has the same length as the above list.
+	 * We have to use the parallel lists and not a list of structs, because the
+	 * Plans have to by copyable by the Postgres _copy functions, and we can't
```
jnidzwetzki, line 58:
```suggestion
	 * Plans have to be copyable by the Postgres _copy functions, and we can't
```

jnidzwetzki (approved):
Found a typo in a comment, the rest looks good to me 👍

Issue #5394, opened 2023-03-05
Assertion failure in CopyFetcher


svenklemm:
This was found by sqlsmith. sqlsmith was running against the database from regresscheck-shared.

Query:
```
WITH
jennifer_0 AS (select
    pg_catalog.tsmultirange() as c0
  from
    public.metrics_int as ref_0
  where pg_catalog.line(
      cast(case when (select relation_name from timescaledb_experimental.policies limit 1 offset 5)
             !~~* pg_catalog.version() then cast(null as point) else cast(null as point) end
         as point),
      cast(cast(null as point) as point)) = cast(null as line)
  limit 114)
select
    _timescaledb_internal.tsl_loaded() as c0,
    70 as c1,
    sample_1.v3 as c2
  from
    public.matches_reference as sample_0 tablesample system (7.9)
      inner join public.metrics_compressed as sample_1 tablesample bernoulli (9.9)
      on (case when (sample_0.location is NULL)
              or (70 = cast(null as int8)) then cast(null as xid) else cast(null as xid) end
             = sample_1.device_id)
  where pg_catalog.pg_stat_get_bgwriter_buf_written_clean() <> (select time from public.mvcp_hyper limit 1 offset 98)

  limit 96

```

Stacktrace:
````
#0  0x00007fc5fd80864c in ?? () from /usr/lib/libc.so.6
#1  0x00007fc5fd7b8938 in raise () from /usr/lib/libc.so.6
#2  0x00007fc5fd7a253d in abort () from /usr/lib/libc.so.6
#3  0x00005591459455da in ExceptionalCondition (conditionName=0x7fc5f0dc0536 "fetcher->state.eof", errorType=0x7fc5f0dbffe1 "FailedAssertion", fileName=0x7fc5f0dc00c0 "/home/sven/projects/tsmaster/tsl/src/remote/copy_fetcher.c",
    lineNumber=494) at assert.c:69
#4  0x00007fc5f0d8fdb5 in copy_fetcher_complete (fetcher=0x5591462b6588) at /home/sven/projects/tsmaster/tsl/src/remote/copy_fetcher.c:494
#5  0x00007fc5f0d8ff0c in copy_fetcher_fetch_data (df=0x5591462b6588) at /home/sven/projects/tsmaster/tsl/src/remote/copy_fetcher.c:535
[...]

Issue #5323, opened 2023-02-14
[Bug]: Copy fetcher creates NULL values


nikolaps:
### What type of bug is this?

Incorrect result

### What subsystems and features are affected?

Distributed hypertable

### What happened?

We have a hypertable  with all fields NOT NULL constraint:
```
ticks=# \d schema.table_name
                            Table "schema.table_name"
    Column     |            Type             | Collation | Nullable | Default
---------------+-----------------------------+-----------+----------+---------
 col01         | integer                     |           | not null |
 col02         | integer                     |           | not null |
 col03         | character(1)                |           | not null |
 col04         | double precision            |           | not null |
 col05         | double precision            |           | not null |
 col06         | integer                     |           | not null |
 col07         | timestamp without time zone |           | not null |
 col08         | timestamp without time zone |           | not null |
 sol09         | integer                     |           |          |
Indexes:
    "table_name_col01_idx" btree (col01)
    "table_name_col07_idx" btree (col07 DESC)
Triggers:
    ts_insert_blocker BEFORE INSERT ON schema.table_name FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker()
Number of child tables: 9138 (Use \d+ to list them.)
```
However, with certain timestamp condition interval in where clause the query returns one row with all NULLs:

```
select count(col01), count(*)
 from schema.table_name
where col07 > '2023-01-24 14:31:08.009484'
and col07 <= '2023-01-24 14:31:10.009484'
 and col01 = 618;
 count | count
-------+-------
  9999 | 10000
(1 row)
```
and
```
select count(*)
from schema.table_name where col01 is null;
 count
-------
     0
(1 row)
```
![photo_2023-02-13_15-31-36](https://user-images.githubusercontent.com/44470692/218751743-ccbeadfe-e1e2-4fd4-bef6-e6ea858cfc1d.jpg)

col07 is the field we use as chunk partition key, If I change the interval (enlarge or shrink, whatever) - the null row disappears

### TimescaleDB version affected

2.9.1

### PostgreSQL version used

14.1

### What operating system did you use?

AlmaLinux release 8.7

### What installation method did you use?

RPM

### What platform did you run on?

On prem/Self-hosted

### How can we reproduce the bug?

```bash
I don't think this is possible, since data populated this table is sensitive.
```

kgyrtkirk:
* can we have an `explain analyze` for the problematic query? (I guess it will be the usual; but we never know)
```
[...]

PR #5410, opened 2023-03-07
Fix file trailer handling in the copy fetcher

jnidzwetzki:
The copy fetcher fetches tuples in batches. When the last element in the batch is the file trailer, the trailer was not handled correctly. The existing logic did not perform a PQgetCopyData in that case. Therefore the state of the fetcher was not set to EOF and the copy operation was not correctly finished at this point.

Fixes: #5323, #5394


svenklemm:
Might also fix #5394

akuzm (approved):
Nice. Maybe we can get rid of `file_trailer_received` now?

PR #5417, opened 2023-03-09
Refactor and optimize distributed COPY

erimatnor:
Refactor the code path that handles remote distributed COPY. The
main changes are:

* Use a hash table to lookup data node connections instead of a list.
* Reduce the number of data copies by accumulating bigger CopyData
  messages directly in the libpq connection buffer instead of doing it
  in a separate buffer which later requires copying the data. To
  achieve this, some internal libpq functions are imported that gives
  more fine-grain control over buffers.
* Add an foreign data wrapper option to set the number of rows to
  send in a CopyData message.

### DIFF
--- a/tsl/src/fdw/option.c
+++ b/tsl/src/fdw/option.c
@@ -133,7 +133,8 @@ option_validate(List *options_list, Oid catalog)
 			if (fetch_size <= 0)
 				ereport(ERROR,
 						(errcode(ERRCODE_SYNTAX_ERROR),
-						 errmsg("%s requires a non-negative integer value", def->defname)));
+						 errmsg("%s requires a non-zero and positive integer value",
+								def->defname)));
 		}
 		else if (strcmp(def->defname, "available") == 0)
 		{
@@ -144,6 +145,18 @@ option_validate(List *options_list, Oid catalog)
 		{
 			/* check and store list, warn about non existing tables */
 			(void) option_extract_join_ref_table_list(defGetString(def));
+		}
+		else if (strcmp(def->defname, "copy_rows_per_message") == 0)
+		{
+			int copy_rows_per_message;
+
+			copy_rows_per_message = strtol(defGetString(def), NULL, 10);
+
+			if (copy_rows_per_message <= 0)
+				ereport(ERROR,
+						(errcode(ERRCODE_SYNTAX_ERROR),
+						 errmsg("%s requires a non-zero and positive integer value",
+								def->defname)));
 		}
 	}
 }
@@ -170,6 +183,8 @@ init_ts_fdw_options(void)
 		{ "available", ForeignServerRelationId },
 		/* join reference tables */
 		{ "reference_tables", ForeignDataWrapperRelationId },
+		/* Rows per CopyData when ingesting with COPY */
+		{ "copy_rows_per_message", ForeignDataWrapperRelationId },
 		{ NULL, InvalidOid }
 	};
 
--- a/tsl/src/nodes/data_node_copy.c
+++ b/tsl/src/nodes/data_node_copy.c
@@ -225,7 +225,6 @@ data_node_copy_exec(CustomScanState *node)
 			ResultRelInfo *rri_chunk = cds->rri;
 			const ChunkInsertState *cis = rri_chunk->ri_FdwState;
 			MemoryContext oldmctx;
-			bool success;
 			const TupleDesc rri_desc = RelationGetDescr(rri_chunk->ri_RelationDesc);
 
 			if (NULL != rri_chunk->ri_projectReturning && rri_desc->constr &&
@@ -234,25 +233,20 @@ data_node_copy_exec(CustomScanState *node)
 
 			ResetPerTupleExprContext(estate);
 			oldmctx = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
-			success = remote_copy_send_slot(dncs->copy_ctx, slot, cis);
+			remote_copy_send_slot(dncs->copy_ctx, slot, cis);
 			MemoryContextSwitchTo(oldmctx);
 
-			if (!success)
-				slot = ExecClearTuple(slot);
-			else
+			if (has_returning)
 			{
-				if (has_returning)
-				{
-					ExprContext *econtext;
+				ExprContext *econtext;
 
-					Assert(NULL != rri_saved->ri_projectReturning);
-					econtext = rri_saved->ri_projectReturning->pi_exprContext;
-					econtext->ecxt_scantuple = slot;
-				}
-
-				if (dncs->set_processed)
-					estate->es_processed++;
+				Assert(NULL != rri_saved->ri_projectReturning);
+				econtext = rri_saved->ri_projectReturning->pi_exprContext;
+				econtext->ecxt_scantuple = slot;
 			}
+
+			if (dncs->set_processed)
+				estate->es_processed++;
 		}
 	} while (!has_returning && !TupIsNull(slot));
 
--- a/tsl/src/remote/dist_copy.c
+++ b/tsl/src/remote/dist_copy.c
@@ -3,21 +3,25 @@
  * Please see the included NOTICE for copyright information and
  * LICENSE-TIMESCALE for a copy of the license.
  */
-#include "dist_copy.h"
-
 #include <postgres.h>
 #include <access/tupdesc.h>
 #include <catalog/namespace.h>
+#include <commands/dbcommands.h>
 #include <executor/executor.h>
 #include <libpq-fe.h>
+#include <lib/stringinfo.h>
 #include <miscadmin.h>
 #include <parser/parse_type.h>
 #include <port/pg_bswap.h>
+#include <storage/latch.h>
 #include <utils/builtins.h>
 #include <utils/lsyscache.h>
+#include <utils/palloc.h>
 
+#include "dist_copy.h"
 #include "compat/compat.h"
 #include "chunk.h"
+#include "config.h"
 #include "copy.h"
 #include "data_node.h"
 #include "dimension.h"
@@ -28,6 +32,7 @@
 #include "nodes/chunk_dispatch/chunk_dispatch.h"
 #include "nodes/chunk_dispatch/chunk_insert_state.h"
 #include "partitioning.h"
+#include "remote/connection.h"
 #include "remote/connection_cache.h"
 #include "remote/dist_txn.h"
 #include "ts_catalog/chunk_data_node.h"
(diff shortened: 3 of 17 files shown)


akuzm:
> there seems little to gain from accumulating a bigger data buffer before calling PQputCopyData()

The reason to batch rows before PQputCopyData is that it is a pretty complex function that adds several kinds of overhead:
1) Overhead in the size of data that is sent to the remote server. Each buffer is sent as a separate message, which has overhead (header/length/trailer etc).
2) CPU overhead on data node to parse these extra messages.
3) CPU overhead on access node. This function does a lot of things.
4) Possibly network overhead because it tries to flush the pending output buffers.

Removing the batching increases all these kinds of overhead 1000 times (by current batch size).

Point 3 is probably the most critical, because COPY throughput is limited by CPU usage on the access node (Igor can comment on that).

erimatnor:
> > there seems little to gain from accumulating a bigger data buffer before calling PQputCopyData()
> 
> The reason to batch rows before PQputCopyData is that it is a pretty complex function that adds several kinds of overhead:
> 
>     1. Overhead in the size of data that is sent to the remote server. Each buffer is sent as a separate message, which has overhead (header/length/trailer etc).
> 
>     2. CPU overhead on data node to parse these extra messages.
> 
>     3. CPU overhead on access node. This function does a lot of things.
> 
>     4. Possibly network overhead because it tries to flush the pending output buffers.
> 
> 
> Removing the batching increases all these kinds of overhead 1000 times (by current batch size).
> 
> Point 3 is probably the most critical, because COPY throughput is limited by CPU usage on the access node (Igor can comment on that).

CPU overhead is actually what I am most concerned about. I am testing different things in this PR (which is why it is still in draft), and removing buffering actually is faster on my local machine. Presumably this is because the buffering done in the current code is actually very CPU intensive given how much string wrangling happens:  for each data node, we need to first construct the row index arrays, then loop through them to create a data node unique buffer/data string. That's potentially a lot of CPU to do that and it gets much worse with higher replication factors. Perhaps the network overhead is worth it if CPU/latency is the bottleneck and not network bandwidth. 

Now, my local (single machine) setup is not realistic so we should basically benchmark this code on a real cluster and see what effect it has. I might be wrong, but adding back some buffering to this PR is not difficult.

erimatnor:
@akuzm I added the ability to accumulate bigger CopyData messages directly in the libpq send buffer instead of maintaining our own buffers. The number of rows to accumulate before sending a CopyData message can be set using a foreign data wrapper option. This allows us to test and evaluate different buffering levels.

I had to import some libpq code because, although it is possible to include the internal libpq API (actually the header explicitly says apps can do this at their own risk), the internal functions aren't exported by the linker. But the send functions needed isn't much code and is likely to stay quite stable across PG versions.

tsl/src/remote/dist_copy.c
```diff
@@ -1248,355 +1223,388 @@ calculate_hyperspace_point_from_binary(Datum *values, bool *nulls, const Hypersp
...
+				 * mapping for the unavailable data nodes.
+				 *
+				 * Note that the metadata will only get updated once since we assign the chunk's
+				 * data_node list to the list of available DNs the first time this
+				 * dist_update_stale_chunk_metadata API gets called. So both chunk_data_nodes and
+				 * chunk->data_nodes will point to the same list and no subsequent metadata updates
+				 * will occur.
+				 */
+				if (ht->fd.replication_factor > list_length(chunk_data_nodes))
+					ts_cm_functions->dist_update_stale_chunk_metadata(chunk, chunk_data_nodes);
```
akuzm, line 1585: This is not changed by your PR, but can we update metadata once per batch instead of once per row?

erimatnor: I think @nikkhils already did some changes here, although I am not sure I fully follow what changed. 

Perhaps better to do those changes separately, with @nikkhils's input.

nikkhils: @akuzm @erimatnor this metadata update change is already in master and 2.10.x branch

erimatnor:
Here are some performance numbers for ingest with this change on a 5-DN cluster:

Screenshot 2023-03-28 at 2 07 37 PM

tsl/src/remote/dist_copy.c
```diff
@@ -1248,355 +1225,369 @@ calculate_hyperspace_point_from_binary(Datum *values, bool *nulls, const Hypersp
...
-		 * pqPutMsgEnd().
-		 */
-		int res = PQputCopyData(pg_conn, copy_data.data, copy_data.len);
+					if (ret == -1)
+						elog(ERROR,
+							 "error when flushing connection to data node %s: %s",
+							 remote_connection_node_name(dnc->connection),
+							 PQerrorMessage(pg_conn));
+				}
+			}
```
akuzm, line 1454: Probably should be `ereport`. The `elog` is for internal program errors, and this looks like a network error.

tsl/src/remote/dist_copy.c
```diff
@@ -1248,355 +1225,369 @@ calculate_hyperspace_point_from_binary(Datum *values, bool *nulls, const Hypersp
...
-	 * doesn't provide a way to know the outgoing buffer size. It also doesn't
-	 * provide any way to control the outgoing buffer size.
-	 * Don't do it if we have ended the COPY above to create new chunks.
-	 * The number 11 is an arbitrary prime, growing the output buffer to at
-	 * most 11ki rows sounds reasonable.
-	 */
-	if (!did_end_copy && context->batch_ordinal % 11 == 0)
-	{
-		flush_active_connections(&context->connection_state);
-	}
```
akuzm, line 1480: Do we still do the periodic flush somewhere? We need some way to control the outgoing queue size. I expect we're going to be bottlenecked at the data nodes in the normal case, so we need to avoid growing the output buffers on access node indefinitely.

akuzm (approved):
Good to see the improvement in the benchmarks.

tsl/src/fdw/option.c
```diff
@@ -145,6 +146,18 @@ option_validate(List *options_list, Oid catalog)
 			/* check and store list, warn about non existing tables */
 			(void) option_extract_join_ref_table_list(defGetString(def));
 		}
+		else if (strcmp(def->defname, "copy_rows_per_message") == 0)
+		{
+			int copy_rows_per_message;
+
+			copy_rows_per_message = strtol(defGetString(def), NULL, 10);
```
pmwkaa, line 153: nit: it could also return `LONG_MAX` on overflow and set errno as `ERANGE`, maybe it should be checked as well

tsl/src/remote/libpq-int.c
```diff
@@ -0,0 +1,153 @@
...
+ * This file contains functions that are available in the internal libpq API,
+ * but are not exported symbols in the shared library.
+ *
+ * In the internal/libpq-int.h header, it is stated that it is possible for
+ * applications to include the internal API at their own risk (the risk being
+ * potential breakage between PG versions). However, the library exludes those
+ * functions from the list of exported symbols for some reason.
+ */
+int
+pqCheckOutBufferSpace(size_t bytes_needed, PGconn *conn)
```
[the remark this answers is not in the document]
pmwkaa: Yeah, it also could lead to some weird linking issues depending on a way how PostgreSQL/libpq is built

pmwkaa (approved):

tsl/src/nodes/data_node_copy.c
```diff
@@ -234,25 +233,20 @@ data_node_copy_exec(CustomScanState *node)
 
 			ResetPerTupleExprContext(estate);
 			oldmctx = MemoryContextSwitchTo(GetPerTupleMemoryContext(estate));
-			success = remote_copy_send_slot(dncs->copy_ctx, slot, cis);
+			remote_copy_send_slot(dncs->copy_ctx, slot, cis);
 			MemoryContextSwitchTo(oldmctx);
 
-			if (!success)
-				slot = ExecClearTuple(slot);
```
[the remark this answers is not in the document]
erimatnor: remote_copy_send_slot() actually never returned failure. It always returned true, in case of failure it would throw an error.

tsl/src/remote/dist_copy.c
erimatnor: So, I removed this because I don't think it is necessary.

The reason I don't think it is necessary is because even if we'd use standard libpq functions, it actually flushes automatically once a msg end marker is written and the buffer has reached 8k. 

Further, now that I changed this to use the internal API, we have better control on a per socket basis. So, every socket is flushed once we write the CopyData end marker, which happens after N rows are buffered. So, basically, we accumulate up to N rows of data on a socket, then write msg end and flush. If the flush fails, the code will do a WaitEventSetWait on all sockets that could not be flushed.

Previously, we flushed all sockets after having sent N rows across all sockets, which meant that some sockets might not require a flush while some does.

I kept the `flush_active_connections()` at the end of the transfer, but it only does something if any of the sockets failed the their last flush for some reason. Then it will wait until the flush succeeds.

tsl/src/remote/libpq-int.c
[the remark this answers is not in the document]
erimatnor: I changed this to use the prefix.

tsl/src/fdw/option.c
erimatnor: If LONG_MAX is returned, I believe that just means we'd use the max number of rows per message even if the user input a bizarrely large number. Not sure we need to notify the user for that odd case.

I also just followed the same code as for other parameters, like fetch_size, which originally came from postgres_fdw.

tsl/src/remote/dist_copy.c
akuzm: > So, every socket is flushed once we write the CopyData end marker, which happens after N rows are buffered. So, basically, we accumulate up to N rows of data on a socket, then write msg end and flush. If the flush fails, the code will do a WaitEventSetWait on all sockets that could not be flushed.

Does this guarantee that we're flushing at the same rate we're receiving the new data, and the buffer size doesn't grow? Before it just flushed to zero outgoing buffers, now that you imported the internal header, probably we can just examine the buffer size.

akuzm: After reading the code more, looks like `send_row_to_data_nodes` flushes to zero output buffer, right?

akuzm: Thanks, yes, it's more clear.

erimatnor:
@akuzm @pmwkaa @nikkhils I had issues getting the use of internal APIs to play nice with the Windows builds and tests, so I reverted the use of those APIs. Multi-row message are now accumulated in per-datanode buffers, similar to before. This means additional copying, but it doesn't seem to affect the performance in my local tests. I think the rest of the refactor still has value, however.

PR #5418, opened 2023-03-09
Fix chunk status update in catalog table

### LINKED ISSUE #5389: [Bug]: Inconsistent chunk status for concurrent COMPRESS - INSERT
### What type of bug is this?

Incorrect result

Compression

### What happened?

Adding the following permutation in `compression_ddl_iso`
```
permutation "LockChunk1" "C1" "I1" "UnlockChunk" "Cc" "Ic" "SC1" "S1"
```
Notice that the row inserted by the Insert operation is not returned when querying the table, because its status is (1), when it should be partial (9), due to the new insert. 

2.11.0-dev

### PostgreSQL version used

13.6

Macos ventura 13.1

Source

Not applicable

_No response_

```bash
Following the steps of the compression_ddl_iso test
```

sb230132:
When INSERT and COMPRESS/RECOMPRESS are run in parallel, chunk status in CHUNK catalog table is not updated correctly. This patch fixes this issue.

Fixes #5389

### DIFF
--- a/src/nodes/chunk_dispatch/chunk_dispatch.c
+++ b/src/nodes/chunk_dispatch/chunk_dispatch.c
@@ -4,11 +4,13 @@
  * LICENSE-APACHE for a copy of the license.
  */
 #include <postgres.h>
+#include <access/xact.h>
 #include <nodes/nodes.h>
 #include <nodes/extensible.h>
 #include <nodes/makefuncs.h>
 #include <nodes/nodeFuncs.h>
 #include <parser/parsetree.h>
+#include <storage/lmgr.h>
 #include <utils/rel.h>
 #include <catalog/pg_type.h>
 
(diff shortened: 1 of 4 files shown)


src/nodes/chunk_dispatch/chunk_insert_state.c
```diff
@@ -737,12 +737,21 @@ ts_chunk_insert_state_destroy(ChunkInsertState *state)
...
+	 * rather than checking compressed status from cached chunks.
+	 * This is because once chunk which is retrieved from catalog
+	 * table is cached, an update to this chunk catalog information
+	 * is not updated in cached chunk, which leads to invalid
+	 * behaviour.
+	 */
+	MemoryContext oldmcxt = MemoryContextSwitchTo(state->mctx);
+	Chunk *chunk = ts_chunk_get_by_relid(chunk_relid, true);
+	/* set status to partial (9) only if chunk is compressed (1) */
+	if (ts_chunk_is_compressed(chunk))
```
konskov, line 752:
```suggestion
	if (ts_chunk_is_compressed(chunk) && !ts_chunk_is_partial(chunk))
```

src/nodes/chunk_dispatch/chunk_dispatch.c
```diff
@@ -305,6 +307,17 @@ chunk_dispatch_exec(CustomScanState *node)
 #endif
 	}
 
+	if (IsolationUsesXactSnapshot())
```
konskov, line 310:
```suggestion
	if (IsolationUsesXactSnapshot() || IsolationIsSerializable())
```

antekresic:
This change might address the issue with decompression as well: #5419 

We could just add another isolation tests with decompression to confirm.

svenklemm (changes requested):
This will make operations on a hypertable block operations on all other hypertables which we want to avoid.

antekresic:
This can be closed as duplicate of #5533

antekresic:
Closing since the issue is already fixed with different PR.

PR #5421, opened 2023-03-09
Simplify control flow in decompress_chunk_exec

akuzm:
No functional changes, mostly just reshuffles the code to prepare for batch decompression.


tsl/src/nodes/decompress_chunk/exec.c
```diff
@@ -416,101 +442,94 @@ static TupleTableSlot *
...
-					break;
+
+				slot->tts_isnull[attr] = result.is_null;
+				slot->tts_values[attr] = result.val;
 			}
 		}
 
-		if (batch_done)
+		/*
+		 * It's a virtual tuple slot, so no point in clearing/storing it
```
jnidzwetzki, line 522: It seems to work, but it might violate [the protocol](https://github.com/postgres/postgres/blob/a0137388cb642bcc1ba10fb906b88658653c589d/src/backend/executor/execTuples.c#L1544) for dealing with virtual tuples and we might miss [this free](https://github.com/postgres/postgres/blob/a0137388cb642bcc1ba10fb906b88658653c589d/src/backend/executor/execTuples.c#L109).

I wonder if that is intentional. If so, could you add an `Assert(TTS_IS_VIRTUAL(slot))` here?

tsl/src/nodes/decompress_chunk/exec.c
```diff
@@ -317,35 +322,56 @@ initialize_batch(DecompressChunkState *state, TupleTableSlot *slot)
...
-				}
-				else
+				value = slot_getattr(compressed_slot, column->compressed_scan_attno, &isnull);
+
+				if (isnull)
+				{
+					/*
+					 * The column will have a default value for the entire batch,
+					 * set it now.
+					 */
```
jnidzwetzki, line 334: Is the line below still needed? The iterator was already set to `NULL` in line 325.

tsl/src/nodes/decompress_chunk/exec.c
```diff
@@ -317,35 +322,56 @@ initialize_batch(DecompressChunkState *state, TupleTableSlot *slot)
...
 				break;
+			}
 			case COUNT_COLUMN:
-				value = slot_getattr(slot, column->compressed_scan_attno, &isnull);
-				state->counter = DatumGetInt32(value);
+				value = slot_getattr(compressed_slot, column->compressed_scan_attno, &isnull);
 				/* count column should never be NULL */
 				Assert(!isnull);
+				int count_value = DatumGetInt32(value);
+				Assert(count_value > 0);
```
jnidzwetzki, line 372: Is it guaranteed that no empty batches exist (e.g., created due to a bug in a previous version)? Otherwise, the Assert should be changed.

```suggestion
				Assert(count_value >= 0);
```

akuzm: I think this would mean data corruption. I'll change it to an `ereport`. We don't control the data, so this is technically an environment error.

tsl/src/nodes/decompress_chunk/exec.c
akuzm: Removed.

tsl/src/nodes/decompress_chunk/exec.c
```diff
@@ -416,101 +442,94 @@ static TupleTableSlot *
...
+			ResetExprContext(econtext);
+
 			TupleTableSlot *subslot = ExecProcNode(linitial(state->csstate.custom_ps));
 
 			if (TupIsNull(subslot))
 				return NULL;
 
-			batch_done = false;
-			initialize_batch(state, subslot);
+			initialize_batch(state, subslot, slot);
```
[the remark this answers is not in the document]
akuzm: Renamed.

tsl/src/nodes/decompress_chunk/exec.c
jnidzwetzki: I did some experiments and could not generate such chunks either. I think `ereport` is a good idea here.

tsl/src/nodes/decompress_chunk/exec.c
jnidzwetzki: Could you add a comment that we are aware of this? Someone who does not know the discussion in this PR might otherwise wonder about it.

jnidzwetzki (approved):

tsl/src/nodes/decompress_chunk/exec.c
akuzm: Added a mention of the standard protocol.

svenklemm (approved):

PR #5422, opened 2023-03-09
Track file trailer only in debug builds

jnidzwetzki:
The commit 96574a749757b20e995dd6fcd9be5140f34092c1 changes the handling of the `file_trailer_received` flag. It
is now only used in asserts and not in any other kind of logic. This
patch encapsulates the `file_trailer_received` in a `USE_ASSERT_CHECKING`
macro.

### DIFF
--- a/tsl/src/remote/copy_fetcher.c
+++ b/tsl/src/remote/copy_fetcher.c
@@ -19,8 +19,10 @@ typedef struct CopyFetcher
 	/* Data for virtual tuples of the current retrieved batch. */
 	Datum *batch_values;
 	bool *batch_nulls;
-	bool file_trailer_received;
 	AsyncRequest *req;
+#ifdef USE_ASSERT_CHECKING
+	bool file_trailer_received;
+#endif
 } CopyFetcher;
 
 static void copy_fetcher_send_fetch_request(DataFetcher *df);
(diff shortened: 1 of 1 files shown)


tsl/src/remote/copy_fetcher.c
```diff
@@ -449,7 +454,9 @@ copy_fetcher_complete(CopyFetcher *fetcher)
 				 * nor the expected number of columns. This provides an extra
 				 * check against somehow getting out of sync with the data.
 				 */
+#ifdef USE_ASSERT_CHECKING
 				fetcher->file_trailer_received = true;
```
akuzm, line 458: Below you have a comment saying that the loop will not be executed in some cases, so maybe we can not execute it always?

The reasons we needed this flag is that the final iteration of the loop was supposed to check for EOF. Now that you are doing this check in place, we don't need the final iteration at all, right?

svenklemm (approved):

tsl/src/remote/copy_fetcher.c
jnidzwetzki: This is correct. The final loop should check for EOF but it was not always executed. Now, we perform the read operation directly here and update the EOF state accordingly. I updated the comment to make this easier to understand.

